PR #8, opened 2024-01-15
Port C++ CAGRA runtime from RAFT

lowener:
This PR is porting the CAGRA runtime API from RAFT to cuVS.


cpp/include/cuvs/neighbors/cagra_types.hpp
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2023, NVIDIA CORPORATION.
+ * Copyright (c) 2023-2024, NVIDIA CORPORATION.
```
cjnolet, line 2: Since cuVS is not going to have a header-only component, we no longer need to separate the headers out like this. This is a change that should be okay for users to make- instead of having to include 2 different files, they can jsut include `cuvs/neighbors/cagra.hpp`. And there should only be declarations in the new headers. All defintions/implementations should be compiled into the source files.

cpp/include/cuvs/neighbors/cagra.cuh
```diff
@@ -29,397 +25,68 @@
...
-// TODO: Remove deprecated experimental namespace in 23.12 release
-namespace cuvs::neighbors::experimental::cagra {
-using cuvs::neighbors::cagra::build;
-using cuvs::neighbors::cagra::build_knn_graph;
-using cuvs::neighbors::cagra::optimize;
-using cuvs::neighbors::cagra::search;
-using cuvs::neighbors::cagra::sort_knn_graph;
-}  // namespace cuvs::neighbors::experimental::cagra
+// Using device and host_matrix_view avoids needing to typedef multiple mdspans based on accessors
+#define CUVS_INST_CAGRA_FUNCS(T, IdxT)                                             \
```
cjnolet, line 29: This file should be an `hpp` since it's no longer compiling any kernels. The reason we used `cuh` for raft is because it's header-only and any headers that compile kernels should be marked as such (by convention).

cjnolet (commented):
Looking great so far! I wanted to share some initial feedback early. 

I think we should go ahead and remove all of the headers from the `detail` namespace and any of the public headers that we're not using. I had originally copied the entire RAFT repo over thinking it might be easy just to rename the VSS bits but it proved harder than I had hoped.

cpp/test/CMakeLists.txt
```diff
@@ -152,15 +152,15 @@ if(BUILD_TESTS)
...
-    src/neighbors/detail/cagra/search_multi_cta_float_uint64_dim128_t8.cu
-    src/neighbors/detail/cagra/search_multi_cta_float_uint64_dim256_t16.cu
-    src/neighbors/detail/cagra/search_multi_cta_float_uint64_dim512_t32.cu
-    src/neighbors/detail/cagra/search_multi_cta_float_uint64_dim1024_t32.cu
-    src/neighbors/detail/cagra/search_single_cta_float_uint64_dim128_t8.cu
-    src/neighbors/detail/cagra/search_single_cta_float_uint64_dim256_t16.cu
-    src/neighbors/detail/cagra/search_single_cta_float_uint64_dim512_t32.cu
-    src/neighbors/detail/cagra/search_single_cta_float_uint64_dim1024_t32.cu
+    #test/neighbors/ann_cagra/test_float_int64_t.cu
+    #src/neighbors/detail/cagra/search_multi_cta_float_uint64_dim128_t8.cu
```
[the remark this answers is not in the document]
lowener: I kept `test/neighbors/ann_cagra/*` to test end-to-end CAGRA

cjnolet:
Changes have been merged into #27. Closing this one for now.

PR #28, opened 2024-01-30
CAGRA C-api compile fixes

benfred:
The cagra_c.h file was using in-class member initializers, which are only available in c++11 - and not in C. This resulted in compile failures when trying to build C libraries using CAGRA.

Fix and add a basic C smoke test that would have caught this.


cpp/include/cuvs/neighbors/cagra_c.h
```diff
@@ -46,15 +46,33 @@ enum cagraGraphBuildAlgo {
...
 
+typedef struct cagraIndexParams* cagraIndexParams_t;
+
+/**
+ * @brief Allocate CAGRA Index params, and populate with default values
+ *
+ * @param[in] params cagraIndexParams_t to allocate
+ * @return cuvsError_t
+ */
+cuvsError_t cagraIndexParamsCreate(cagraIndexParams_t* params);
```
cjnolet, line 66: Small nitpick, but can we please add cuvs tot he beginning? It makes it a lot easier to know where different functions are coming from when, for example, an application is making calls to other C libraries like cublas and cusparse.

cjnolet (approved):

PR #39, opened 2024-03-04
Expose exception text to C, Python and Rust API's

benfred:
Previously when something went wrong in the Rust or C API - all the end user would see is a `CUVS_ERROR` return code with no extra indication of what went wrong.

This change exposes the exception text to both the C and Rust api's, and provides a convenience method to automatically catch c++ exceptions, and convert the exception into an error code with the text set appropiately.


benfred:
Thanks @dantegd !  I've updated the python side to also have the error text - let me know what you think,

dantegd (approved):

PR #40, opened 2024-03-05
Fix 'cuvs' conda package to only include Python artifacts and clean build.sh 

dantegd:
PR does the following:

- [x] Stops the inclusion of C/C++ artifacts in the `cuvs` conda package
- [x] Cleans `build.sh` from code that was copied from RAFT
- [x] Adds `dlpack` to development environments
- [x] Other small fixes and cleanups


benfred:
I'm trying to fix the broken rust build here https://github.com/rapidsai/cuvs/pull/43 . My current guess is that I didn't pin the cuda version right when creating the conda environment, so it started picking up cuda 12.4 (which was released yesterday) and is failing since we don't support that yet.

benfred (approved):

cjnolet (approved):

AyodeAwe (approved):

PR #43, opened 2024-03-06
Attempt to fix broken rust CI

benfred:


bdice:
Thanks for this @benfred -- I was about to try the same thing in #40 after noticing the failure there. We may need to merge these two PRs together to get a working environment.

benfred:
@bdice if this works, lets just merge this fix into #40 (I'm hoping this fixes, but am not 100% sure)

benfred:
looks like it worked 😅  - pushed out the commit to #40

PR #48, opened 2024-03-11
Replace local copyright check with pre-commit-hooks verify-copyright

KyleFromNVIDIA:
The local `copyright.py` script is bug-prone. Replace it with a more robust centralized script from `pre-commit-hooks`.

Issue: https://github.com/rapidsai/build-planning/issues/30


.pre-commit-config.yaml
```diff
@@ -107,6 +101,22 @@ repos:
         rev: v4.5.0
         hooks:
               - id: check-json
+      - repo: https://github.com/rapidsai/pre-commit-hooks
+        rev: v0.0.1
+        hooks:
+          - id: verify-copyright
+            files: |
+              (?x)
+                  [.](cmake|cpp|cu|cuh|h|hpp|sh|pxd|py|pyx)$|
```
benfred, line 110: should we add in the rust files to this ?
```suggestion
                  [.](cmake|cpp|cu|cuh|h|hpp|sh|pxd|py|pyx|rs)$|
```

benfred (approved):
looks good - thanks!

.pre-commit-config.yaml
KyleFromNVIDIA: That was not present in the old file, but I'll go ahead and add it.

raydouglass (approved):

PR #57, opened 2024-03-18
Remove unused get_*.cmake files.

bdice:
This PR removes some unused `get_*.cmake` files. I know cuVS may need to add some of these back as functionality is migrated from RAFT. As that happens, I would like to copy them again from the latest versions in RAFT.

Specifically, files like `get_thrust.cmake` are outdated (we are moving to `rapids_cpm_cccl` and `rapids_cpm_thrust` is deprecated) so this will reduce the maintenance burden for CMake / build system developers in the interim period.


bdice:
We can retarget to 24.06 if desired. I think this PR should be low-risk, since there were no references to the deleted files and CI is passing.

lowener (approved):

PR #59, opened 2024-03-21
Add Brute Force KNN to Python and Rust API's

benfred:


python/cuvs/cuvs/neighbors/brute_force/brute_force.pyx
```diff
@@ -0,0 +1,216 @@
...
+from pylibraft.distance.pairwise_distance import DISTANCE_TYPES
+from pylibraft.neighbors.common import _check_input_array
+
+from cuvs.common.c_api cimport cuvsResources_t
+
+from cuvs.common.exceptions import check_cuvs
+
+
+cdef class FlatIndex:
+    """
```
cjnolet, line 43: I understand the reasoning behind changing this to flat index, but I don't think this is the right way to do this and I'm not sure if we need to follow FAISS' naming convention here. We call this index brute-force everywhere else, so we should stick that for consistency.

python/cuvs/cuvs/neighbors/brute_force/brute_force.pyx
```diff
@@ -0,0 +1,216 @@
...
+from cuvs.common cimport cydlpack
+from cuvs.distance_type cimport DistanceType
+
+from pylibraft.common import auto_convert_output, cai_wrapper, device_ndarray
+from pylibraft.common.cai_wrapper import wrap_array
+from pylibraft.common.interruptible import cuda_interruptible
+from pylibraft.distance.pairwise_distance import DISTANCE_TYPES
+from pylibraft.neighbors.common import _check_input_array
+
+from cuvs.common.c_api cimport cuvsResources_t
```
dantegd, line 37: thanks for the PR for fixing this! I _really_ wanted to avoid having to have specific cython code for this here and just use the one in pylibraft for the handle :/

python/cuvs/cuvs/neighbors/brute_force/brute_force.pyx
```diff
@@ -0,0 +1,216 @@
...
+    >>> n_samples = 50000
+    >>> n_features = 50
+    >>> n_queries = 1000
+    >>> k = 10
+    >>> dataset = cp.random.random_sample((n_samples, n_features),
+    ...                                   dtype=cp.float32)
+    >>> index = brute_force.build(dataset, metric="cosine")
+    >>> distances, neighbors = brute_force.search(index, dataset, k)
+    >>> distances = cp.asarray(distances)
+    >>> neighbors = cp.asarray(neighbors)
```
dantegd, line 100: why do we need to cp.asarray these? Also, for the example we could consider printing them (same comment applies to all examples)

python/cuvs/cuvs/neighbors/brute_force/brute_force.pyx
```diff
@@ -0,0 +1,216 @@
...
+    ...                                   dtype=cp.float32)
+    >>> # Build index
+    >>> index = brute_force.build(dataset, metric="sqeuclidean")
+    >>> # Search using the built index
+    >>> queries = cp.random.random_sample((n_queries, n_features),
+    ...                                   dtype=cp.float32)
+    >>> k = 10
+    >>> # Using a pooling allocator reduces overhead of temporary array
+    >>> # creation during search. This is useful if multiple searches
+    >>> # are performad with same query size.
```
dantegd, line 170: I know that I am one of the people that left this comment in the examples :), but was thinking it'd be neat if we add it instead to the docstring of relevant methods in a "widget" style, similar to what networkx does with cugraph algos, see the bottom of https://networkx.org/documentation/stable/reference/algorithms/generated/networkx.algorithms.centrality.betweenness_centrality.html

python/cuvs/cuvs/neighbors/brute_force/brute_force.pyx
```diff
@@ -0,0 +1,216 @@
...
+        return self.trained
+
+    def __repr__(self):
+        return "FlatIndex(type=BruteForce)"
+
+
+@auto_sync_resources
+def build(dataset, metric="sqeuclidean", metric_arg=2.0, resources=None):
+    """
+    Build the Brute Force index from the dataset for efficient search.
```
dantegd, line 70: Note: pending decision of naming convention, there's a few places where we use Brute or Flat, so it'd be good to do a quick check of consistency after the name is decide

python/cuvs/cuvs/test/test_brute_force.py
```diff
@@ -0,0 +1,90 @@
...
+from pylibraft.common import device_ndarray
+from scipy.spatial.distance import cdist
+
+from cuvs.neighbors import brute_force
+
+
+@pytest.mark.parametrize("n_index_rows", [32, 100])
+@pytest.mark.parametrize("n_query_rows", [32, 100])
+@pytest.mark.parametrize("n_cols", [40, 100])
+@pytest.mark.parametrize("k", [1, 5, 32])
```
dantegd, line 27: Why these values? Is 5 testing something that 32 isn't? Note that this is a general question for all the pytests including the ones I added from pylibraft, so we could do a follow up work for deciding this.

python/cuvs/cuvs/test/test_brute_force.py
```diff
@@ -0,0 +1,90 @@
...
+import pytest
+from pylibraft.common import device_ndarray
+from scipy.spatial.distance import cdist
+
+from cuvs.neighbors import brute_force
+
+
+@pytest.mark.parametrize("n_index_rows", [32, 100])
+@pytest.mark.parametrize("n_query_rows", [32, 100])
+@pytest.mark.parametrize("n_cols", [40, 100])
```
dantegd, line 26: Same question as why these particular values. Perhaps we're overparametrizing a bit, that given the number of `metrics`, is not an insignificant number.

dantegd (commented):
Review of the python code and it looks good, just added a few minor questions and comments

python/cuvs/cuvs/test/test_brute_force.py
benfred: I believe 5 was added for making sure we include a k value thats not a power of 2 - 

Fwiw, this test was copied from the the one in pylibraft - which is what originally included these parameters

python/cuvs/cuvs/neighbors/brute_force/brute_force.pyx
benfred: thanks! I couldn't find a way to use pylibraft for the handle here - so I made that other PR to change

python/cuvs/cuvs/test/test_brute_force.py
dantegd: That's fair enough, and the additional matrix will allow us to triage a failure more easily than a succint matrix. 

That said, maybe something we could do per-release could be a pytest time analysis. I've done it in cuML a few times (example; https://github.com/rapidsai/cuml/issues/3228) when pytests were slowing things down, it could be a practice that will allow us to be proactive before we accidentally slow things down when adding new pytests, but that is beyond the PR, but was wondering what do you think about that

python/cuvs/cuvs/neighbors/brute_force/brute_force.pyx
cjnolet: This is a great idea. Do one of you mind copy/pasting this comment into a Github issue just so we can keep it on our radar?

python/cuvs/cuvs/test/test_brute_force.py
benfred: I think thats a great idea - as we keep adding tests, keeping on top of the total execution time will be important. We could even add something like `pytest --durations=10`  to print out the top 10 slowest tests on each CI run (though, right now that is a little noisy because the first test run initializes cublas etc - which hits causes a bit of a slowdown)

cjnolet (approved):

PR #61, opened 2024-03-21
Add CAGRA-Q compression

achirkin:
Minimal code changes to add CAGRA-Q compression to the CAGRA index.
When built with compression, the index uses the compressed dataset automatically during search. At this moment, there's not new search parameters compared to standard CAGRA.

### DIFF
--- a/cpp/include/cuvs/neighbors/cagra.h
+++ b/cpp/include/cuvs/neighbors/cagra.h
@@ -40,39 +40,6 @@ enum cuvsCagraGraphBuildAlgo {
   NN_DESCENT
 };
 
-/**
- * @brief Supplemental parameters to build CAGRA Index
- *
- */
-struct cuvsCagraIndexParams {
-  /** Degree of input graph for pruning. */
-  size_t intermediate_graph_degree;
-  /** Degree of output graph. */
-  size_t graph_degree;
-  /** ANN algorithm to build knn graph. */
-  enum cuvsCagraGraphBuildAlgo build_algo;
-  /** Number of Iterations to run if building with NN_DESCENT */
-  size_t nn_descent_niter;
-};
-
-typedef struct cuvsCagraIndexParams* cuvsCagraIndexParams_t;
-
-/**
- * @brief Allocate CAGRA Index params, and populate with default values
- *
- * @param[in] params cuvsCagraIndexParams_t to allocate
- * @return cuvsError_t
- */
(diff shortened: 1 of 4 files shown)


cpp/include/cuvs/neighbors/cagra.h
```diff
@@ -73,6 +73,61 @@ cuvsError_t cuvsCagraIndexParamsCreate(cuvsCagraIndexParams_t* params);
...
+struct cuvsCagraCompressionParams {
+  /**
+   * The bit length of the vector element after compression by PQ.
+   *
+   * Possible values: [4, 5, 6, 7, 8].
+   *
+   * Hint: the smaller the 'pq_bits', the smaller the index size and the better the search
+   * performance, but the lower the recall.
+   */
+  uint32_t pq_bits = 8;
```
benfred, line 86: In class member initializers aren't valid in C (only are in c++11) . This has previously caused problems trying to compile with a C compiler :  https://github.com/rapidsai/cuvs/pull/28 .

We should probably just remove these default values ,  since the default values are going to get picked up from the c++ struct in `cuvsCagraCompressionParamsCreate`

benfred (commented):
thanks for this @achirkin !  Lets get this PR merged in and I can expose to rust (and python too maybe)

cpp/include/cuvs/neighbors/cagra.h
achirkin: Thanks, Ben! I totally missed this in the copy-paste process

benfred (approved):
lgtm! thanks @achirkin

achirkin:
Thanks, @benfred, for the review. I have a small opinion question about the C api: do you think it would be better to make the `cuvsCagraCompressionParams_t` a member of `cuvsCagraIndexParams_t`? That is, by default, it would be `nullptr` and would mimic the C++ api `optional<vpq_params>` behavior; then, the `cuvsCagraBuildCompressed` could be removed.

Also I'd like to add couple test cases for CAGRA to cover compression.

benfred:
>Thanks, @benfred, for the review. I have a small opinion question about the C api: do you think it would be better to make the cuvsCagraCompressionParams_t a member of cuvsCagraIndexParams_t? That is, by default, it would be nullptr and would mimic the C++ api optional<vpq_params> behavior; then, the cuvsCagraBuildCompressed could be removed.

I think thats a great idea! (also adding tests is a great idea - I'm maybe just jumping the gun a bit with the review because I'm excited to see this change in =) )

cjnolet (approved):

PR #66, opened 2024-04-03
ConfigureCUDA.cmake now sets CUVS_ prefixed variables

robertmaynard:
This makes sure that CUVS uses the compile flags that are required for RAPIDS C++ projects.

### DIFF
--- a/cpp/cmake/modules/ConfigureCUDA.cmake
+++ b/cpp/cmake/modules/ConfigureCUDA.cmake
@@ -1,5 +1,5 @@
 # =============================================================================
-# Copyright (c) 2018-2023, NVIDIA CORPORATION.
+# Copyright (c) 2018-2024, NVIDIA CORPORATION.
 #
 # Licensed under the Apache License, Version 2.0 (the "License"); you may not use this file except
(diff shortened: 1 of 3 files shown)


robertmaynard:
Ci failures fixed in https://github.com/rapidsai/raft/pull/2254

cpp/cmake/modules/ConfigureCUDA.cmake
```diff
@@ -13,45 +13,45 @@
...
 endif()
 
 if(OpenMP_FOUND)
-  list(APPEND RAFT_CUDA_FLAGS -Xcompiler=${OpenMP_CXX_FLAGS})
+  list(APPEND CUVS_CUDA_FLAGS -Xcompiler=${OpenMP_CXX_FLAGS})
 endif()
 
 # Debug options
 if(CMAKE_BUILD_TYPE MATCHES Debug)
   message(VERBOSE "RAFT: Building with debugging flags")
```
bdice, line 54:
```suggestion
  message(VERBOSE "cuVS: Building with debugging flags")
```

bdice (approved):

cjnolet (approved):

PR #69, opened 2024-04-05
Adds missing files to `update-version.sh`

AyodeAwe:

PR #75, opened 2024-04-10
Fix for 24.04 release

raydouglass:
Fixes some versions so `24.04` can be released.

#69 implements the changes in the script

This PR should replace #76

### DIFF
--- a/.devcontainer/cuda11.8-pip/devcontainer.json
+++ b/.devcontainer/cuda11.8-pip/devcontainer.json
@@ -5,16 +5,26 @@
     "args": {
       "CUDA": "11.8",
       "PYTHON_PACKAGE_MANAGER": "pip",
-      "BASE": "rapidsai/devcontainers:24.04-cpp-llvm16-cuda11.8-ubuntu22.04"
+      "BASE": "rapidsai/devcontainers:24.04-cpp-cuda11.8-ubuntu22.04"
     }
   },
   "hostRequirements": {"gpu": "optional"},
   "features": {
-    "ghcr.io/rapidsai/devcontainers/features/ucx:24.4": {"version": "1.14.1"},
+    "ghcr.io/rapidsai/devcontainers/features/ucx:24.4": {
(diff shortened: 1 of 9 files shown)


.devcontainer/cuda11.8-pip/devcontainer.json
```diff
@@ -10,8 +10,8 @@
   },
   "hostRequirements": {"gpu": "optional"},
   "features": {
-    "ghcr.io/rapidsai/devcontainers/features/ucx:24.2": {"version": "1.14.1"},
-    "ghcr.io/rapidsai/devcontainers/features/rapids-build-utils:24.2": {}
+    "ghcr.io/rapidsai/devcontainers/features/ucx:24.4": {"version": "1.14.1"},
```
bdice, line 13: We updated all of RAPIDS 24.04 to use ucx 1.15, so this is out of date. This isn't a blocker for the release because the same issue exists in other RAPIDS repos (I checked RAFT). However, we should make a pass in 24.06 to fix these to 1.15.0. I wasn't aware that devcontainers hard-coded this. @trxcllnt could you help with updating this?

https://github.com/rapidsai/build-planning/issues/27#issuecomment-1997868433
https://github.com/rapidsai/integration/pull/704
https://github.com/rapidsai/ucx-py/pull/1032
https://github.com/rapidsai/ucxx/pull/205

bdice (approved):

.devcontainer/cuda11.8-pip/devcontainer.json
trxcllnt: @bdice does this need to be done in this PR?

bdice: > This isn't a blocker for the release because the same issue exists in other RAPIDS repos (I checked RAFT). However, we should make a pass in 24.06 to fix these to 1.15.0.

No, please target 24.06 for the update.

benfred (approved):

PR #72, opened 2024-04-08
Update devcontainers to use cuda12.2

benfred:

PR #76, opened 2024-04-10
CMake patch

AyodeAwe:
PRs apply CMake fixes. See https://github.com/rapidsai/cuvs/pull/72#issuecomment-2044097811


benfred (commented):
To properly backport the cmake fixes in #72 to 24.04 - I think we might also need https://github.com/rapidsai/cuvs/commit/e7e3f5684509eca24fca187b30d56ba62f66d0aa

PR #80, opened 2024-04-11
Forward merge branch-24.04 to branch-24.06

benfred:


bdice (approved):

.github/workflows/pr.yaml
```diff
@@ -105,5 +105,5 @@ jobs:
       cuda: '["12.2"]'
       build_command: |
         sccache -z;
-        build-all -DBUILD_PRIMS_BENCH=ON -DBUILD_ANN_BENCH=ON --verbose;
+        build-all --verbose;
```
cjnolet, line 108: Oh- good find. I guess we don't need this in cuVS, do we?

benfred: Yeah, it was a good catch - @trxcllnt fixed this in https://github.com/rapidsai/cuvs/pull/75/commits/a6e1436a010e672230b1511d754d81a918b2a703

cjnolet (approved):

raydouglass (approved):

PR #82, opened 2024-04-15
Add python and rust bindings for Ivf-Flat

benfred:


python/cuvs/cuvs/neighbors/ivf_flat/CMakeLists.txt
```diff
@@ -0,0 +1,24 @@
+# =============================================================================
+# Copyright (c) 2023-2024, NVIDIA CORPORATION.
```
dantegd, line 2:
```suggestion
# Copyright (c) 2024, NVIDIA CORPORATION.
```

python/cuvs/cuvs/neighbors/ivf_flat/ivf_flat.pyx
```diff
@@ -0,0 +1,354 @@
...
+    uint8_t,
+    uint32_t,
+    uint64_t,
+    uintptr_t,
+)
+
+from cuvs.common.exceptions import check_cuvs
+
+
+cdef class IndexParams:
```
dantegd, line 48: Question that popped up in my mind as we add the other ANN methods: is having identical named `IndexParams` potentially confusing or leading to issues from the fact that they are named the same for every ANN variant? I might be overthinking this, but wanted to see your opinion @benfred

python/cuvs/cuvs/neighbors/ivf_flat/ivf_flat.pyx
```diff
@@ -0,0 +1,354 @@
...
+        "euclidean"], where
+            - sqeuclidean is the euclidean distance without the square root
+              operation, i.e.: distance(a,b) = \\sum_i (a_i - b_i)^2,
+            - euclidean is the euclidean distance
+            - inner product distance is defined as
+              distance(a, b) = \\sum_i a_i * b_i.
+    kmeans_n_iters : int, default = 20
+        The number of iterations searching for kmeans centers during index
+        building.
+    kmeans_trainset_fraction : int, default = 0.5
```
dantegd, line 67: I wonder if adding a suggestion of when to change this would be useful for users

python/cuvs/cuvs/neighbors/ivf_flat/ivf_flat.pyx
```diff
@@ -0,0 +1,354 @@
...
+        are used for training.
+    add_data_on_build : bool, default = True
+        After training the coarse and fine quantizers, we will populate
+        the index with the dataset if add_data_on_build == True, otherwise
+        the index is left empty, and the extend method can be used
+        to add new vectors to the index.
+    adaptive_centers : bool, default = False
+        By default (adaptive_centers = False), the cluster centers are
+        trained in `ivf_flat::build`, and and never modified in
+        `ivf_flat::extend`. The alternative behavior (adaptive_centers
```
dantegd, line 79:
```suggestion
        trained in `ivf_flat.build`, and and never modified in
        `ivf_flat.extend`. The alternative behavior (adaptive_centers
```

I presume the `::` were just inheritted from the C++ side

dantegd (commented):
Quick review of the python side, things look good overall there!

python/cuvs/cuvs/neighbors/ivf_flat/ivf_flat.pyx
benfred: I think since they are in different modules its probably ok - like we have `ivf_flat.IndexParams` vs `cagra.IndexParams` which will differentiate the two.

python/cuvs/cuvs/neighbors/ivf_flat/ivf_flat.pyx
benfred: good catch!

python/cuvs/cuvs/neighbors/ivf_flat/ivf_flat.pyx
benfred: I think it could be very helpful!  Though - I'm not sure of when to change this myself, and have just copied this comment from the c++ api. Do you have any thoughts here?

benfred: @cjnolet that sounds good! I've added in the last commit https://github.com/rapidsai/cuvs/pull/82/commits/3c9efd3832a4edc312e390dd2cb33267f802c487 (which also adds this into the docs site)

dantegd (approved):
looks great from my side!

cjnolet (approved):
LGTM!

PR #85, opened 2024-04-18
Moving and renaming distance namespaces from raft -> cuvs

cjnolet:
This PR moves the `raft::distance` APIs over to cuVS. We're ripping the band-aid off quickly here so moving all over at once. 

All of the non-public APIs are being moved to `src/distance` and only the public APIs being moved to `include/cuvs/distance`.


benfred (approved):

benfred (approved):

raydouglass (approved):
`ops-codeowners` changes look good

PR #90, opened 2024-04-19
Add python and rust bindings for Ivf-Pq

benfred:


python/cuvs/cuvs/neighbors/ivf_pq/ivf_pq.pyx
```diff
@@ -0,0 +1,430 @@
...
+
+@auto_sync_resources
+def build(IndexParams index_params, dataset, resources=None):
+    """
+    Build the IvfPq index from the dataset for efficient search.
+
+    Parameters
+    ----------
+    index_params : IndexParams object
+    dataset : CUDA array interface compliant matrix shape (n_samples, dim)
```
cjnolet, line 225: Now that we're not just exposing a simple device-level primitive, I think it's time we start supporting host-based (e.g. numpy) arrays as well, even if that means automatically allocating device memory and copying from host to device internally in the python layer like we do in cuML. I think we can push that change to a future PR, but do you mind creating an issue for it at least?

python/cuvs/cuvs/neighbors/ivf_pq/ivf_pq.pyx
```diff
@@ -0,0 +1,430 @@
...
+    >>> n_queries = 1000
+    >>> k = 10
+    >>> dataset = cp.random.random_sample((n_samples, n_features),
+    ...                                   dtype=cp.float32)
+    >>> build_params = ivf_pq.IndexParams(metric="sqeuclidean")
+    >>> index = ivf_pq.build(build_params, dataset)
+    >>> distances, neighbors = ivf_pq.search(ivf_pq.SearchParams(),
+    ...                                        index, dataset,
+    ...                                        k)
+    >>> distances = cp.asarray(distances)
```
cjnolet, line 249: I think we can clean this up now too and maybe follow the semantics that have been successful in cuml- if a numpy array is passed in, a numpy array should be returned, etc... 

cuml abstracts these conversations behind a "CumlArray" object. I suspect we should consider doing something similar for cuVS.

python/cuvs/cuvs/neighbors/ivf_pq/ivf_pq.pyx
```diff
@@ -0,0 +1,430 @@
...
+           k,
+           neighbors=None,
+           distances=None,
+           resources=None):
+    """
+    Find the k nearest neighbors for each query.
+
+    Parameters
+    ----------
+    search_params : SearchParams
```
cjnolet, line 351: Does this automatically to the ivf_pq.SearchParas in the docs? If not, we might want to fully-qualify these

cjnolet (approved):
LGTM, though there's a couple user quality of life improvements I think we should capture in Github issues and follow-up with soon

python/cuvs/cuvs/neighbors/ivf_pq/ivf_pq.pyx
benfred: https://github.com/rapidsai/cuvs/issues/96

python/cuvs/cuvs/neighbors/ivf_pq/ivf_pq.pyx
benfred: good point about the docs! We were missing the RST files for ivf-pq and bfknn, so it didn't show up at all =( - I've added those in.

For linking to the SearchParams struct - I had to update to `:py:class:cuvs.neighbors.ivf_pq.SearchParams` (like both add the fully qualified class name, and the `py:class` marker to create the hyper link) - but with that change we get a clickable link in the sphinx docs

cjnolet (approved):

PR #100, opened 2024-04-30
Moving CAGRA from RAFT

cjnolet:


cpp/src/core/nvtx.hpp
achirkin, line 1: I assume, this is supposed to be used within cuVS only, while the users would be encouraged to use raft's public headers?
If so, I'd suggest (1) putting this into detail, (2) also adding a something like `using range = raft::common::nvtx::range<cuvs>` and using that alias across cuVS codebase.

cjnolet: Please note that this file was introduced in an upstream PR (moving ivfpq) which itself is still under review. At the moment, we are more concerned with getting things to compile and tests to run successfully before we focus on smaller details like this. 

To answer your question, yes, the ultimate goal will be use (public) utilities/wrappers in RAFT to mark the NVTX, but have them set cuvs name spacing in cuvs (the utility should still not be in detail if used across namespaces, but it should not be in the user's public include headers. 

This move exposed many many cases where detail namespaces being used across namespaces, and this is going to need to be fixed. Going forward, I'm going to start enforcing this more broadly- it leads to code that becomes unnecessarily complicated to maintain and it creates multiple entry points for code that should otherwise be very straightforward to follow end-to-end (and that's not even including the compile implications on said code).

PR #111, opened 2024-05-09
Pass through raft static CMake var

vyasr:
cuvs wheel builds were deviating from the behavior of the rest of RAPIDS before this PR by not linking statically to libraft.so but instead building the full dynamic and static lib (the default behavior) and then presumably bundling libraft.so. This change should fix that linking.

### DIFF
--- a/ci/test_wheel_cuvs.sh
+++ b/ci/test_wheel_cuvs.sh
@@ -7,6 +7,9 @@ mkdir -p ./dist
 RAPIDS_PY_CUDA_SUFFIX="$(rapids-wheel-ctk-name-gen ${RAPIDS_CUDA_VERSION})"
 RAPIDS_PY_WHEEL_NAME="cuvs_${RAPIDS_PY_CUDA_SUFFIX}" rapids-download-wheels-from-s3 ./dist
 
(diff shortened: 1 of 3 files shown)


ci/test_wheel_cuvs.sh
```diff
@@ -7,6 +7,9 @@ mkdir -p ./dist
 RAPIDS_PY_CUDA_SUFFIX="$(rapids-wheel-ctk-name-gen ${RAPIDS_CUDA_VERSION})"
 RAPIDS_PY_WHEEL_NAME="cuvs_${RAPIDS_PY_CUDA_SUFFIX}" rapids-download-wheels-from-s3 ./dist
 
+# Force installing a pylibraft before the libraft wheel
+python -m pip install pylibraft-${RAPIDS_PY_CUDA_SUFFIX}==24.06.00a51
+
```
vyasr, line 12:
```suggestion
```

dantegd (approved):
lgtm

raydouglass:
Removing `ops-codeowners` from the required reviews since it doesn't seem there are any file changes that we're responsible for. Feel free to add us back if necessary.

PR #114, opened 2024-05-10
hide RAFT #pragma deprecation warnings

trxcllnt:
Define `-DRAFT_HIDE_DEPRECATION_WARNINGS` added in https://github.com/rapidsai/raft/pull/2271.


trxcllnt:
[Seems like](https://github.com/rapidsai/cuvs/actions/runs/9036982176/job/24835026440?pr=114#step:7:1803) raft headers have code that triggers warnings. Since the `pip` builds add raft as a CMake submodule, the raft includes are added as `-I` instead of `-isystem`, so raft warnings are failing the `-Werror` cuvs build.

Doesn't raft build with `-Werror`? It seems like any raft warnings should have been caught in the raft CI before being merged?

cjnolet (approved):

PR #115, opened 2024-05-13
Migrate `raft::cluster` to `cuvs::cluster`

cjnolet:


cpp/src/cluster/detail/kmeans_balanced.cuh
```diff
@@ -0,0 +1,1096 @@
...
+
+  // Training clusters in each meso-cluster
+  IdxT n_clusters_done = 0;
+  for (IdxT i = 0; i < n_mesoclusters; i++) {
+    IdxT k = 0;
+    for (IdxT j = 0; j < n_rows && k < mesocluster_size_max; j++) {
+      if (labels_mptr[j] == LabelT(i)) { mc_trainset_ids[k++] = j; }
+    }
+    if (k != static_cast<IdxT>(mesocluster_sizes[i]))
+      RAFT_LOG_WARN("Incorrect mesocluster size at %d. %zu vs %zu",
```
tfeher, line 889:
```suggestion
      RAFT_LOG_DEBUG("Incorrect mesocluster size at %d. %zu vs %zu",
```

cpp/src/cluster/detail/kmeans_balanced.cuh
```diff
@@ -0,0 +1,1096 @@
...
+  raft::resource::sync_stream(handle, stream);
+
+  // build fine clusters
+  auto [mesocluster_size_max, fine_clusters_nums_max, fine_clusters_nums, fine_clusters_csum] =
+    arrange_fine_clusters(n_clusters, n_mesoclusters, n_rows, mesocluster_sizes);
+
+  const IdxT mesocluster_size_max_balanced = raft::div_rounding_up_safe<size_t>(
+    2lu * size_t(n_rows), std::max<size_t>(size_t(n_mesoclusters), 1lu));
+  if (mesocluster_size_max > mesocluster_size_max_balanced) {
+    RAFT_LOG_WARN(
```
tfeher, line 1038:
```suggestion
    RAFT_LOG_DEBUG(
```

cpp/src/cluster/detail/kmeans_balanced.cuh
```diff
@@ -0,0 +1,1096 @@
...
+ *   2. Predict fine cluster
+ *   3. Refince the fine cluster centers
+ *
+ *  As a result, the fine clusters are what is returned by `build_hierarchical`;
+ *  this function returns the total number of fine clusters, which can be checked to be
+ *  the same as the requested number of clusters.
+ *
+ *  Note: this function uses at most `fine_clusters_nums_max` points per mesocluster for training;
+ *  if one of the clusters is larger than that (as given by `mesocluster_sizes`), the extra data
+ *  is ignored and a warning is reported.
```
tfeher, line 838:
```suggestion
 *  is ignored.
```

tfeher (commented):
While we are here, we could remove the confusing warning message.

cpp/include/cuvs/cluster/agglomerative.hpp
```diff
@@ -0,0 +1,107 @@
...
+#pragma once
+
+#include <cuvs/distance/distance_types.hpp>
+#include <optional>
+#include <raft/core/device_mdspan.hpp>
+#include <raft/core/resources.hpp>
+
+namespace cuvs::cluster::agglomerative {
+
+constexpr int DEFAULT_CONST_C = 15;
```
divyegala, line 26: Can you add a comment for what this is?

cpp/include/cuvs/cluster/kmeans.hpp
```diff
@@ -0,0 +1,332 @@
...
+
+/**
+ * @brief Predict the closest cluster each sample in X belongs to.
+ *
+ * @code{.cpp}
+ *   #include <raft/core/resources.hpp>
+ *   #include <cuvs/cluster/kmeans.hpp>
+ *   using namespace  cuvs::cluster;
+ *   ...
+ *   raft::raft::resources handle;
```
divyegala, line 171:
```suggestion
 *   raft::resources handle;
```

cpp/include/cuvs/cluster/kmeans.hpp
```diff
@@ -0,0 +1,332 @@
...
+/**
+ * @brief Compute k-means clustering and predicts cluster index for each sample
+ * in the input.
+ *
+ * @code{.cpp}
+ *   #include <raft/core/resources.hpp>
+ *   #include <cuvs/cluster/kmeans.hpp>
+ *   using namespace  cuvs::cluster;
+ *   ...
+ *   raft::raft::resources handle;
```
divyegala, line 242:
```suggestion
 *   raft::resources handle;
```

cpp/src/cluster/detail/kmeans.cuh
```diff
@@ -0,0 +1,1269 @@
...
+      (pams.metric == cuvs::distance::DistanceType::L2Expanded ||
+       pams.metric == cuvs::distance::DistanceType::L2SqrtExpanded)) {
+    RAFT_LOG_DEBUG(
+      "batch_centroids=%d was passed, but batch_centroids=%d will be used (reason: "
+      "batch_centroids has no impact on the memory footprint when FusedL2NN can be used)",
+      pams.batch_centroids,
+      pams.n_clusters);
+  }
+
+  //  raft::logger::get(RAFT_NAME).set_level(pams.verbosity);
```
divyegala, line 880: There are many more comments like this one. Should we keep and uncomment or remove?

cpp/src/cluster/detail/single_linkage.cuh
```diff
@@ -0,0 +1,125 @@
...
+ * @param[in] m number of rows in X
+ * @param[in] n number of columns in X
+ * @param[in] metric distance metrix to use when constructing connectivities graph
+ * @param[out] out struct containing output dendrogram and cluster assignments
+ * @param[in] c a constant used when constructing connectivities from knn graph. Allows the indirect
+ control
+ *            of k. The algorithm will set `k = log(n) + c`
+ * @param[in] n_clusters number of clusters to assign data samples
+ */
+template <typename value_idx, typename value_t, Linkage dist_type>
```
divyegala, line 52: I have followed the stack up and down, and I don't think `Linkage` needs to be a template parameter, we can get away with it as a function argument and reduce the number of instantiations. While I don't think we are invoking any kernels twice so the savings might not be extraordinary, I think it's worth reducing number of templates anyway.

I'm okay with doing this as a follow-on if you consider this worth refactoring later on.

cpp/src/cluster/single_linkage_float.cu
```diff
@@ -0,0 +1,42 @@
...
+                    cuvs::cluster::agglomerative::Linkage linkage,
+                    std::optional<int> c)
+{
+  if (linkage == Linkage::KNN_GRAPH) {
+    single_linkage<float, int, Linkage::KNN_GRAPH>(
+      handle, X, dendrogram, labels, metric, n_clusters, c);
+  } else {
+    single_linkage<float, int, Linkage::PAIRWISE>(
+      handle, X, dendrogram, labels, metric, n_clusters, c);
+  }
```
divyegala, line 40: Here's where we could pass `Linkage` straight as a function argument as we are already receiving it as a function argument.

cpp/src/distance/fused_distance_nn-ext.cuh
```diff
@@ -0,0 +1,84 @@
...
+#pragma once
+
+#include "fused_distance_nn_helpers.cuh"  // include initialize and reduce operations
+#include <raft/core/kvp.hpp>              // raft::KeyValuePair
+#include <raft/core/resources.hpp>        // raft::resources
+#include <raft/util/raft_explicit.hpp>    // RAFT_EXPLICIT
+
+#include <cstdint>  // int64_t
+
+#ifdef RAFT_EXPLICIT_INSTANTIATE_ONLY
```
divyegala, line 26: Do we define this macro in cuvs? Why are we keeping the -ext and -inl structure?

cpp/src/cluster/detail/kmeans.cuh
cjnolet: This was left over from when I was trying to figure out compile errors. Those have been fixed so I can uncomment.

divyegala: Oh I see. Do we have to use `RAFT_NAME` or are we able to initialize the logger to "cuvs"?

cpp/src/cluster/detail/single_linkage.cuh
cjnolet: There was a strict reason behind this, actually. If you recall, initially we had half of single linkage in cuml and half in raft. We had the pairwise distance primitive in raft but the knn primitives hadn't been moved over yet. This is essentially doing an extensible strategy pattern that allowed the raft version to plug into the cuml version. 

I agree it could go away now. I don't think this migration is the place to do that, though. I'm trying to change as little in the internals as possible so that we have a baseline point for the move. The goal is to have a heavy focus on continuing to improve these internals over the next few versions.

I created a blanket issue for removing the unneeded templates. I'll add this one to that issue (and please also add things as you see / think of them just so we don't lose sight of them.

cpp/src/distance/fused_distance_nn-ext.cuh
cjnolet: This is just to get the code over and a public API exposed. We can and should continue to clean these up over time. Main concern right now is clean public api, proper source file etiquette, and build time that's not horrendous.

cjnolet: One thing I probably should do is search and replace for all the template instantiation macros that have _raft_ in the name. Don't need to keep those raft.

divyegala: What we should do right now is just keep the -inl file then and call it a day. It will work as-is without needing any of the extra fluff that comes in the -ext and the main file, and we also won't have to worry about macros.

cpp/src/cluster/detail/kmeans.cuh
cjnolet: For now we do. We need to get the logger issue figured out soon, but I think that's a lower priority. (Raft name is unfortunately what's hardcoded in raft/core/logger.hpp)

cjnolet: Oh my bad. That's what I get for responding on my phone. Yeah we probably don't have to use raft name here, but unfortunately the way the logger.hpp is structure in raft it forces us to use RAFT_NAME everywhere else... thus changing it here doesn't really do much for us. I can change it in this file though just for the heck of it.

divyegala: So we need to do `RAFT_NAME = "cuvs"` for it to work?

divyegala: Or `CUVS_NAME = "raft"`?

cjnolet: Unfortunately, today we need to have RAFT_NAME defined for the logger itself to work. We don't have time for a better solution for 24.06 (there's still a lot of work to do and we need to get the guts migrated first).

divyegala (approved):

PR #116, opened 2024-05-14
Fix IVF-PQ helper functions

lowener:
Some IVF-PQ helper functions needed for FAISS were missing.


cpp/include/cuvs/neighbors/ivf_pq_helpers.hpp
```diff
@@ -39,65 +39,337 @@ namespace cuvs::neighbors::ivf_pq::helpers {
 
 namespace codepacker {
```
cjnolet, line 40: Can we keep the namespace separation but go ahead and add these definitions to `ivf_pq.hpp` please? Now that we're no longer header-only, there is near zero cost to just putting all the API elements for ivf-pq together.

cpp/src/neighbors/ivf_pq/ivf_pq_build_common.cu
```diff
@@ -1,9 +1,9 @@
 /*
- * raft::copyright (c) 2022-2024, NVIDIA CORPORATION.
+ * Copyright (c) 2022-2024, NVIDIA CORPORATION.
```
cjnolet, line 2: I didn't notice this one before. That's a great search and replace snafu. This one just made my day.

docs/source/cpp_api/neighbors_ivf_pq.rst
```diff
@@ -66,3 +66,11 @@ Index serialize
     :project: cuvs
     :members:
     :content-only:
+
+Helper Methods
```
cjnolet, line 70: Can you add a small description for this section? For example, so that a user knows when they might want to use these functions.

docs/source/cpp_api/neighbors_ivf_pq.rst
```diff
@@ -66,3 +66,15 @@ Index serialize
...
+---------------
+
+Those methods are used to manipulate the underlying data of an IVF-PQ index, and perform operations like unpacking records, or writing PQ codes into an existing IVF list.
```
cjnolet, line 73:
```suggestion
Additional helper functions for manipulating the underlying data of an IVF-PQ index, unpacking records, and writing PQ codes into an existing IVF list.
```

cpp/include/cuvs/neighbors/ivf_pq.hpp
```diff
@@ -1064,4 +1064,673 @@ void deserialize(raft::resources const& handle,
...
+  uint32_t pq_dim,
+  uint32_t pq_bits,
+  uint32_t offset,
+  raft::device_mdspan<uint8_t, list_spec<uint32_t, uint32_t>::list_extents, raft::row_major>
+    list_data);
+
+/**
+ * @}
+ */
+};  // namespace codepacker
```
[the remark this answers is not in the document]
cjnolet: We should do this as a follow-on just to make sure we can get the main bits merged.

cjnolet (approved):

cjnolet:
@lowener I was otherwise thinking we should move the additional codepacker functions into codepacker namespace in a follow up, but since this PR requires fixing merge conflicts anyways, I think we should just go ahead and do that move here.

PR #122, opened 2024-05-15
Enable random subsampling

tfeher:
Random sampling of training set for IVF methods was reverted in https://github.com/rapidsai/raft/pull/2144 due to the large memory usage of the subsample method.

Since then, PR https://github.com/rapidsai/raft/pull/2155 has implemented a new random sampling method with improved memory utilization.  Using that we can now enable random sampling of IVF methods (https://github.com/rapidsai/raft/pull/2052 and https://github.com/rapidsai/raft/pull/2077).

Random subsampling has measurable overhead for IVF-Flat, therefore it is only enabled for IVF-PQ.


cjnolet (approved):
LGTM! Thanks @tfeher

PR #130, opened 2024-05-17
ANN_BENCH

achirkin:
Porting the ANN benchmarks from RAFT.
- [x] Make it build

Sanity check that benchmarks work (runs and gives reasonable recall for Deep-1M dataset)
- [x] cuVS brute force kNN 
- [x] cuVS IVF-Flat
- [x] cuVS IVF-PQ (+ refinement)
- [x] cuVS CAGRA
- [x] cuVS CAGRA-Q (+refinement)
- [x] Faiss GPU/CPU IVF-Flat & IVF-PQ
- [x] HNSW
- [x] CAGRA + HNSW 
- [x] GGNN

NB: the indices built using the old ANN_BENCH in raft tend to crash in cuvs search benchmarks during index deserialization - don't forget to build the indexes anew when testing.


tfeher:
Moved additional tasks to #160.

tfeher:
Note the PR contains changes from #154 and #148. We will need #131 as well (due to CAGRA API options for fine tuning index build), but that is not yet included here.

tfeher:
Most of the benchmarks work now (except linker issue for serialization for CAGRA + HNSWLIB). Half precision benchmarks are disabled since cuVS does not support it yet.

build.sh
```diff
@@ -18,8 +18,8 @@ ARGS=$*
 # scripts, and that this script resides in the repo dir!
 REPODIR=$(cd $(dirname $0); pwd)
 
-VALIDARGS="clean libcuvs python rust docs tests examples --uninstall  -v -g -n --compile-static-lib --allgpuarch --no-nvtx --show_depr_warn --incl-cache-stats --time -h"
-HELP="$0 [<target> ...] [<flag> ...] [--cmake-args=\"<args>\"] [--cache-tool=<tool>] [--limit-tests=<targets>] [--build-metrics=<filename>]
+VALIDARGS="clean libcuvs python rust docs tests bench-ann examples --uninstall  -v -g -n --compile-static-lib --allgpuarch --no-nvtx --show_depr_warn --incl-cache-stats --time -h"
```
cjnolet, line 21: Can we rename this to benchmarks, maybe? I'm thinking maybe we could start building one benchmarks suite but maybe have different entrypoints for the ANN grid searches vs clustering benchmarking, for example.

cpp/bench/ann/src/common/ann_types.hpp
```diff
@@ -0,0 +1,168 @@
...
+  }
+}
+
+struct AlgoProperty {
+  MemoryType dataset_memory_type;
+  // neighbors/distances should have same memory type as queries
+  MemoryType query_memory_type;
+};
+
+class AnnBase {
```
cjnolet, line 74: Can we maybe refer to this as an "Index"?

build.sh
achirkin: This sounds like the benchmark infrastructure differs significantly between vector search vs clustering.
What about having multiple entry points but renaming them to better reflect what we're benchmarking? e.g. ~bench-search~, bench-clustering.
Update: bench-ann currently handles search and indexing, so probably "bench-search" is not very appropriate. Maybe keep this `bench-ann` and add the `bench-clustering`, `bench-dists` later?

cpp/bench/ann/src/common/ann_types.hpp
achirkin: For now I renamed this to `algo_base` and `algo<T>`. I hesitate changing this to `index`, because this would clash with actual `index` types from other namespaces, which are members of these wrappers. What do you think about this?

achirkin:
I've just realized the benchmarks are not compiled during `conda-cpp-build` CI. @cjnolet  what's the best way to add the benchmark component to CI build?

achirkin:
NB: some algorithms are likely to fail with OOM in the limiting_resource_adapter; the fix is in https://github.com/rapidsai/cuvs/pull/181

achirkin:
I tried to add benchmarks to the test build in CI, but got the error from faiss not being able to find BLAS libraries, even though I added openblas as a conda dependency. @cjnolet, @benfred, could you please have a look?

tfeher (approved):
Thanks Artem for the updates. This PR ports the existing infrastructure from raft and enables us to run all the existing the benchmarks. I believe it would be useful to have this merged, and work o follow up improvements in separate PRs. I have added open discussion points to the tracker issue https://github.com/rapidsai/cuvs/issues/160#issuecomment-2173310776.

The PR looks good to me. I have also contributed to the PR, so I shall not be the only approver.

conda/recipes/libcuvs/build_libcuvs_tests.sh
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
 # Copyright (c) 2022-2024, NVIDIA CORPORATION.
 
-./build.sh tests --allgpuarch --no-nvtx --build-metrics=tests_bench --incl-cache-stats
+./build.sh tests bench-ann --allgpuarch --no-nvtx --build-metrics=tests_bench --incl-cache-stats
```
jameslamb, line 4: Over in RAFT, `bench-ann` is its own package with its own dependencies, build scripts, etc.:

https://github.com/rapidsai/raft/tree/877644a423c0268746af62cecb7150afa65d8386/python/raft-ann-bench

https://github.com/rapidsai/raft/blob/877644a423c0268746af62cecb7150afa65d8386/conda/recipes/raft-ann-bench/meta.yaml

For my own understanding in reviewing this... why is this being added to the `libcuvs-tests` package here instead of creating a new standalone package as exists in RAFT?

Having these separate things be their own packages can be helpful for parallelizing development, limiting the potential impact of packaging changes, and speeding up debugging.

jameslamb (approved):
I've reviewed on behalf of `packaging-codeowners`. Left 2 small comments, neither needs to block merging this.

cpp/CMakeLists.txt
```diff
@@ -55,6 +55,7 @@ option(BUILD_SHARED_LIBS "Build cuvs shared libraries" ON)
 option(BUILD_TESTS "Build cuvs unit-tests" ON)
 option(BUILD_C_LIBRARY "Build raft C API library" OFF)
 option(BUILD_C_TESTS "Build raft C API tests" OFF)
+option(BUILD_ANN_BENCH "Build cuVS ann benchmarks" OFF)
```
benfred, line 58: Can we add an option to build this in CI? 

It looks to me that this is off by default (which is fine), but that does mean that we don't have any guarantee that any of this is compiling.

achirkin: We enable this currently together with tests in build.sh (see https://github.com/rapidsai/cuvs/pull/130#discussion_r1642994596)

conda/recipes/libcuvs/build_libcuvs_tests.sh
achirkin: I agree that benchmarks should probably be a separate package. I'm only hesitating to add this because I'm not very familiar with this conda/CI setup. I hoped to postpone this question till a follow-on PR together with the naming decision (https://github.com/rapidsai/cuvs/issues/160#issuecomment-2173310776). @benfred, @cjnolet , what do you think?

cjnolet (approved):

PR #131, opened 2024-05-17
CAGRA API update and allow async host refinement 

mfoerste4:
This PR updates the CAGRA public API, changes defaults, and improves refinement during IVF-PQ build step.

Updated defaults:
- By default CAGRA would select NN descent. We fall back to IVF-PQ build algorithm if there is not enough memory for NN descent.
- For the IVF-PQ build algo, the search params were updated to use `n_probe = 0.01*nlist`, and both LUT and internal distance type is set to `fp16` (as opposed to `fp8/fp32` previously)

- By default build would create the index that contains both the `graph` and the `dataset` on GPU. If the `dataset` does not fit gpu, then the returned index will only contain the graph (on device). In such case the user is expected to call `index.update_dataset()`  (for example with dataset in managed memory) before we can `search` the index. 

API changes:
- We can specify IVF-PQ build algo parameters the following way:
```c++
cagra::index_params params;
params.graph_degree = 32;
params.intermediate_graph_degree = 48;
auto pq_params = cagra::graph_build_algo::ivf_pq_params(dataset.extents());
// This sets reasonable defaults, but can be updated by the user, e.g.:
pq_params.pq_dim = 32;
// Select IVF-PQ algorithm by passing ivf_pq_params as graph_build_algo
params.graph_build_algo = pq_params;
```

- We can specify NN descent algo parameters a similar way
```c++
cagra::index_params params;
params.graph_degree = 32;
params.intermediate_graph_degree = 48;
// Select NN descent algo by passing nn_descent_params
params.graph_build_algo = cagra::graph_build_params::nn_descent_params(intermediate_degree) ;
```

Additionally, this PR optimizes the IVF-PQ refinement step within the CAGRA graph creation in case the dataset is in host-memory.

* depending on hardware, a certain amount of refinement comes for free
* the change also has a positive effect without refinement as the graph creation is done on host either way


cpp/src/neighbors/detail/cagra/cagra_build.cuh
```diff
@@ -155,32 +230,59 @@ void build_knn_graph(
 
     cuvs::neighbors::ivf_pq::search(
       res, *search_params, index, queries_view, neighbors_view, distances_view);
-    if constexpr (raft::is_host_mdspan_v<decltype(dataset)>) {
+
+    if (!device_refinement) {
```
tfeher, line 234: For managed arrays (expected to be large, and residing in host memory) we should use host refinement.

tfeher (changes requested):
Thanks Malte for the PR, it looks good overall, just two small comments.

Public API is needed to allow fine control of CAGRA IVF-PQ build options.

cpp/include/cuvs/neighbors/cagra.hpp
```diff
@@ -385,35 +387,65 @@ struct index : cuvs::neighbors::index {
...
+  bool construct_index_with_dataset = true) -> cuvs::neighbors::cagra::index<int8_t, uint32_t>;
+
+auto build(
+  raft::resources const& handle,
+  const cuvs::neighbors::cagra::index_params& params,
+  raft::host_matrix_view<const int8_t, int64_t, raft::row_major> dataset,
+  std::optional<cuvs::neighbors::nn_descent::index_params> nn_descent_params = std::nullopt,
+  std::optional<float> refine_rate                                           = std::nullopt,
+  std::optional<cuvs::neighbors::ivf_pq::index_params> pq_build_params       = std::nullopt,
+  std::optional<cuvs::neighbors::ivf_pq::search_params> search_params        = std::nullopt,
```
tfeher, line 427: These options are useful for fine tuning index construction. One thing we should consider: we could make these member of `cagra::index_params`, similarly to the `compression `struct which is part of `index_params`.

cpp/include/cuvs/neighbors/cagra.hpp
```diff
@@ -385,35 +387,65 @@ struct index : cuvs::neighbors::index {
...
+  bool construct_index_with_dataset = true) -> cuvs::neighbors::cagra::index<int8_t, uint32_t>;
```
tfeher, line 428: We are discussing with @cjnolet whether we need this parameter in the public API. Alternative is that the user runs separate `build_knn_graph` and `optimize` steps, and constructs the index as per their requirements (e.g. moving graph / data to desired memory space).

tfeher (commented):
Thanks Malte, for the updates! Let's discuss API questions with @cjnolet, otherwise, the refinement improvement part looks good!

cpp/src/neighbors/detail/cagra/cagra_build.cuh
```diff
@@ -143,6 +212,12 @@ void build_knn_graph(
   size_t next_report_offset = 0;
   size_t d_report_offset    = dataset.extent(0) / 100;  // Report progress in 1% steps.
 
+  bool host_refinement   = raft::is_host_mdspan_v<decltype(dataset)> && top_k != gpu_top_k;
+  bool device_refinement = !raft::is_host_mdspan_v<decltype(dataset)> && top_k != gpu_top_k;
```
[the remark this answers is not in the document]
cjnolet: Yes- we have a managed mdspan which is expressed as an mdspan with a host_device_accessor that allows for access from both host and device.

cpp/include/cuvs/neighbors/cagra.hpp
```diff
@@ -385,35 +387,65 @@ struct index : cuvs::neighbors::index {
...
-auto build(raft::resources const& handle,
-           const cuvs::neighbors::cagra::index_params& params,
-           raft::device_matrix_view<const uint8_t, int64_t, raft::row_major> dataset)
-  -> cuvs::neighbors::cagra::index<uint8_t, uint32_t>;
-
-auto build(raft::resources const& handle,
-           const cuvs::neighbors::cagra::index_params& params,
-           raft::host_matrix_view<const uint8_t, int64_t, raft::row_major> dataset)
-  -> cuvs::neighbors::cagra::index<uint8_t, uint32_t>;
+auto build(
```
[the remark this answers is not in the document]
cjnolet: Correct- we should document every function / prototype exposed in `include/cuvs/**.hpp`

cpp/include/cuvs/neighbors/cagra.hpp
```diff
@@ -385,35 +387,65 @@ struct index : cuvs::neighbors::index {
...
+  std::optional<float> refine_rate                                           = std::nullopt,
+  std::optional<cuvs::neighbors::ivf_pq::index_params> pq_build_params       = std::nullopt,
+  std::optional<cuvs::neighbors::ivf_pq::search_params> search_params        = std::nullopt,
+  bool construct_index_with_dataset = true) -> cuvs::neighbors::cagra::index<uint8_t, uint32_t>;
+
+auto build(
+  raft::resources const& handle,
+  const cuvs::neighbors::cagra::index_params& params,
+  raft::host_matrix_view<const uint8_t, int64_t, raft::row_major> dataset,
+  std::optional<cuvs::neighbors::nn_descent::index_params> nn_descent_params = std::nullopt,
```
cjnolet, line 444: This is awkward- a user is going to have to set std::nullopt for at least one of nn-descent_params or ivf_pq params. I think we should make these build_params for cagra. Actually, I think what we should do is take away the `graph_build_algo` option from `build_params` and add a `graph_build_params` option that accepts the `cuvs::neighbors::build_params` base class, documents it approprirately with what's allowed, and then dispatches accordingly. Also remember to raise an exception if anything other than `ivf_pq` or `nn_descent` is used. This allows us to add more graph build algos in the future without too much additional effort.

cjnolet: On second though- since you need refine_ratio on the `ivf_pq_build_params`, why not have a new type `cagra::graph_build_params` which can extend `cuvs::neighbors::build_params`? Then you can have a `cagra::graph_build_params::ivf_pq` which extends `ivf_pq::build_params` and `cagra::graph_build_params` and adds the `refine ratio`. Then you can have `cagra__graph_build_params::nn_descent` which does the same for nn-descent. This way you can have `cagra::build_params.graph_build_params` which accepts anything that extends `cagra::graph_build_params` and dispatches the build algo accordingly? This would get all this additional stuff out of `cagra::build` and keep the API clean and easy to follow, while minimizing duplication.

cjnolet (changes requested):

cpp/include/cuvs/neighbors/cagra.hpp
```diff
@@ -37,31 +40,45 @@ namespace cuvs::neighbors::cagra {
...
+ * @brief ANN parameters used by CAGRA to build knn graph
  *
  */
-enum class graph_build_algo {
-  /* Use IVF-PQ to build all-neighbors knn graph */
-  IVF_PQ,
-  /* Experimental, use NN-Descent to build all-neighbors knn graph */
-  NN_DESCENT
+struct graph_build_params {
+  bool construct_index_with_dataset = true;
```
tfeher, line 47: This should be moved out to `cagra::index_params`.

cpp/include/cuvs/neighbors/cagra.hpp
```diff
@@ -37,31 +40,45 @@ namespace cuvs::neighbors::cagra {
...
+  std::optional<cuvs::neighbors::ivf_pq::index_params> build_params   = std::nullopt;
+  std::optional<cuvs::neighbors::ivf_pq::search_params> search_params = std::nullopt;
+  float refinement_rate                                               = 2.0f;
+};
+}  // namespace ivf_pq
+
+namespace nn_descent {
+/** Specialized parameters utilizing NN_DESCENT to build knn graph */
+struct graph_build_params : cuvs::neighbors::cagra::graph_build_params {
+  std::optional<cuvs::neighbors::nn_descent::index_params> nn_descent_params = std::nullopt;
```
tfeher, line 63: Does this need to be optional? After we move the `construct_index_with_dataset` member out, we can just use the `std::variant` to keep an `nn_descent::index_params` struct optionally.

cpp/include/cuvs/neighbors/cagra.hpp
```diff
@@ -37,31 +40,45 @@ namespace cuvs::neighbors::cagra {
...
+};
+}  // namespace nn_descent
```
tfeher, line 65: An alternative idea is to group everything into `graph_build_params` nampspace and have the structures different names. 
```suggestion
namespace graph_build_params {
/** Specialized parameters utilizing IVF-PQ to build knn graph */
struct ivf_pq_params :  {
  std::optional<cuvs::neighbors::ivf_pq::index_params> build_params   = std::nullopt;
  std::optional<cuvs::neighbors::ivf_pq::search_params> search_params = std::nullopt;
  float refinement_rate                                               = 2.0f;
};
using nn_descent_params =  cuvs::neighbors::nn_descent::index_params;
}  // namespace graph_build_params
```

tfeher (changes requested):

tfeher (changes requested):
Thanks Malte, just one nitpick.

cpp/include/cuvs/neighbors/cagra.hpp
```diff
@@ -39,31 +40,45 @@ namespace cuvs::neighbors::cagra {
...
-  NN_DESCENT
+struct graph_build_params {
+  bool construct_index_with_dataset = true;
+  virtual ~graph_build_params() {}
 };
 
+namespace ivf_pq {
+/** Specialized parameters utilizing IVF-PQ to build knn graph */
+struct graph_build_params : cuvs::neighbors::cagra::graph_build_params {
+  std::optional<cuvs::neighbors::ivf_pq::index_params> build_params   = std::nullopt;
```
cjnolet, line 49: This is a bit awkward to have optional here. A struct that could be potentially empty? I think we should remove the std::optional and make these reasonable defaults.

cpp/include/cuvs/neighbors/cagra.hpp
```diff
@@ -418,31 +649,157 @@ auto build(raft::resources const& handle,
...
+ *
+ * The following distance metrics are supported:
+ * - L2
+ *
+ * @param[in] res
+ * @param[in] params parameters for building the index
+ * @param[in] dataset a matrix view (device) to a row-major matrix [n_rows, dim]
+ * @param[out] idx the constructed cagra index
+ */
 void build_device(raft::resources const& handle,
```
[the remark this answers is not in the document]
cjnolet: Oversight- should be removed from the public APIs (note that these were copied directly from RAFT)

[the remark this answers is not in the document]
cjnolet: If you have a moment, it would be helpful to remove build_host and build_device. That was an oversight. Those are invoked internally to `build()` but shouldn't be part of the public APIs. It also creates confusion for users as to what APIs are available for them. `hpp` files should only include APIs we expect users to interact with.

cpp/include/cuvs/neighbors/cagra.hpp
cjnolet: I think your last idea looks great. It's clean, explicit, and tells the user right away what graph_build_params variants are supported.

cpp/include/cuvs/neighbors/cagra.hpp
```diff
@@ -385,65 +393,342 @@ struct index : cuvs::neighbors::index {
...
+ *
+ * The following distance metrics are supported:
+ * - L2
+ *
+ * @param[in] res
+ * @param[in] params parameters for building the index
+ * @param[in] dataset a matrix view (device) to a row-major matrix [n_rows, dim]
+ * @param[out] idx the constructed cagra index
+ */
+void build_device(raft::resources const& res,
```
cjnolet, line 631: I think we should remove these definitions from the `include/cuvs/neighbors/cagra.hpp`

cpp/src/neighbors/cagra_c.cpp
```diff
@@ -38,12 +38,23 @@ void* _build(cuvsResources_t res, cuvsCagraIndexParams params, DLManagedTensor*
...
+  auto index_params                      = cuvs::neighbors::cagra::index_params();
+  index_params.intermediate_graph_degree = params.intermediate_graph_degree;
+  index_params.graph_degree              = params.graph_degree;
+
+  switch (params.build_algo) {
+    case cuvsCagraGraphBuildAlgo::IVF_PQ:
+      index_params.graph_build_params = cuvs::neighbors::cagra::graph_build_params::ivf_pq_params{};
+      break;
+    case cuvsCagraGraphBuildAlgo::NN_DESCENT:
+      cuvs::neighbors::cagra::graph_build_params::nn_descent_params nn_descent_params{};
```
cjnolet, line 50: Can we consolidate the C++ and C structs so that we don't need to duplicate all of these values, please? It's going to be painful to manage two complete copies of these structs because changes are going to get made to one without being made to the other. 

I did this with the `DIstanceType` enum in `cuvs::distance`, for example. I think we should ultimately strive to write these once (in C++ or C) and then expose them through the other.

cpp/include/cuvs/neighbors/cagra.hpp
```diff
@@ -37,31 +40,45 @@ namespace cuvs::neighbors::cagra {
...
-  graph_build_algo build_algo = graph_build_algo::IVF_PQ;
-  /** Number of Iterations to run if building with NN_DESCENT */
-  size_t nn_descent_niter = 20;
   /**
    * Specify compression parameters if compression is desired.
    *
    * NOTE: this is experimental new API, consider it unsafe.
    */
   std::optional<cuvs::neighbors::vpq_params> compression = std::nullopt;
+
```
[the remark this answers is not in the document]
tfeher: Removed, will try to put dataset on device. In case of OOM, catch error and return index without added dataset.

cpp/src/neighbors/detail/cagra/cagra_build.cpp
```diff
@@ -0,0 +1,35 @@
...
+namespace cuvs::neighbors::cagra::graph_build_params {
+ivf_pq_params::ivf_pq_params(raft::matrix_extent<int64_t> dataset_extents,
+                             cuvs::distance::DistanceType metric)
+{
+  build_params = cuvs::neighbors::ivf_pq::index_params::from_dataset(dataset_extents, metric);
+
+  search_params                         = cuvs::neighbors::ivf_pq::search_params{};
+  search_params.n_probes                = std::max<uint32_t>(10, build_params.n_lists * 0.01);
+  search_params.lut_dtype               = CUDA_R_16F;
+  search_params.internal_distance_dtype = CUDA_R_16F;
```
tfeher, line 31: Updated defaults.

cpp/src/neighbors/detail/cagra/cagra_build.cuh
```diff
@@ -330,20 +409,45 @@ index<T, IdxT> build(
   std::optional<raft::host_matrix<IdxT, int64_t>> knn_graph(
     raft::make_host_matrix<IdxT, int64_t>(dataset.extent(0), intermediate_degree));
 
-  if (params.build_algo == graph_build_algo::IVF_PQ) {
-    build_knn_graph(res, dataset, knn_graph->view(), refine_rate, pq_build_params, search_params);
+  // Set default value in case knn_build_params is not defined.
+  auto knn_build_params = params.graph_build_params;
+  if (std::holds_alternative<std::monostate>(params.graph_build_params)) {
```
tfeher, line 414: Select default algorithm

cpp/src/neighbors/detail/cagra/cagra_build.cuh
```diff
@@ -355,25 +459,27 @@ index<T, IdxT> build(
...
+
+    return idx;
+  }
+  try {
     return index<T, IdxT>(res, params.metric, dataset, raft::make_const_mdspan(cagra_graph.view()));
-  } else {
-    // We just add the graph. User is expected to update dataset separately. This branch is used
-    // if user needs special control of memory allocations for the dataset.
+  } catch (std::bad_alloc& e) {
+    RAFT_LOG_DEBUG("Insufficient GPU memory to construct CAGRA index with dataset on GPU");
```
tfeher, line 480: Automatically construct index without dataset if dataset would not fit GPU memory.

tfeher (commented):
Highlighting recent changes.

cjnolet (approved):

cpp/include/cuvs/neighbors/cagra.hpp
```diff
@@ -37,31 +40,36 @@ namespace cuvs::neighbors::cagra {
...
+  bool construct_index_with_dataset = true;
```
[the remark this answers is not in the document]
tfeher: Now this is set automatically: https://github.com/rapidsai/cuvs/pull/131/files#diff-f3ca3714f8c6d118b61d369e6edc085a4ff20ef17a5ec804beb53432e58f9940R479-R484

cpp/include/cuvs/neighbors/cagra.hpp
tfeher: removed.

cpp/include/cuvs/neighbors/cagra.hpp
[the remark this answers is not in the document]
tfeher: removed.

cpp/include/cuvs/neighbors/nn_descent.hpp
```diff
@@ -353,7 +362,17 @@ auto build(raft::resources const& res,
...
 /**
+ * @brief Test if we have enough GPU memory to run NN descent algorithm.
+ *
+ * @param res
+ * @param dataset shape of the dataset
+ * @param idx_size the size of index type in bytes
+ * @return true if enough GPU memory can be allocated
+ * @return false otherwise
+ */
+bool has_enough_device_memory(raft::resources const& res,
```
cjnolet, line 373: Small nitpick- can we move this to the `cuvs::neighbors::nn_descent::helpers` namespace please? 

We can do it in a follow-on PR for 24.08 so we can get this merged ASAP.

cjnolet (approved):

tfeher (approved):

PR #132, opened 2024-05-17
Removing `libraft.so` from libcuvs dependencies

cjnolet:
Unfortunately, libraft.so is still a dependency of pylibraft and so it's needed in order to use pylibraft. We'll work on that during the 24.08 release cycle. For now, libraft.so is no longer a required dependency to build `libcuvs` or any of its wrappers.


cpp/include/cuvs/neighbors/common.hpp
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2024, NVIDIA CORPORATION.
+ * Copyright (c) 2022-2023, NVIDIA CORPORATION.
```
divyegala, line 2: I'm not sure if the copyright check is working correctly in cuvs. Can we file an issue to address it later?

cpp/include/cuvs/neighbors/common.hpp
```diff
@@ -209,7 +280,7 @@ auto make_aligned_dataset(const raft::resources& res, const SrcT& src, uint32_t
 }
 
 /** Parameters for VPQ compression. */
-using vpq_params = cuvs::neighbors::ann::vpq_params;
+using vpq_params = cuvs::neighbors::vpq_params;
```
divyegala, line 283: Does not look like this line is needed anymore as the class definition is in the same file and same namespace now

divyegala (commented):
Minor comments

cpp/include/cuvs/neighbors/common.hpp
cjnolet: Good catch!

divyegala (approved):

PR #142, opened 2024-05-21
Add pairwise_distance api's for C, Python and Rust

benfred:


cpp/include/cuvs/distance/pairwise_distance.h
```diff
@@ -0,0 +1,62 @@
...
+ * @endcode
+ *
+ * @param[in] res cuvs resources object for managing expensive resources
+ * @param[in] x first set of points (size n*k)
+ * @param[in] y second set of points (size m*k)
+ * @param[out] dist output distance matrix (size n*m)
+ * @param[in] metric distance to evaluate
+ * @param[in] metric_arg metric argument (used for Minkowski distance)
+ */
+cuvsError_t cuvsPairwiseDistance(cuvsResources_t res,
```
cjnolet, line 54: This looks so simple and elegant to use!

cpp/src/distance/pairwise_distance_c.cpp
```diff
@@ -0,0 +1,72 @@
...
+
+#include <raft/core/error.hpp>
+#include <raft/core/mdspan_types.hpp>
+#include <raft/core/resources.hpp>
+
+#include <cuvs/core/c_api.h>
+#include <cuvs/core/exceptions.hpp>
+#include <cuvs/core/interop.hpp>
+#include <cuvs/distance/distance.h>
+#include <cuvs/distance/distance.hpp>
```
cjnolet, line 29: Do we need both of these? I refactored `distance.hpp` to basically include `distance.h` so we didn't have two definitions of `DistanceType`.

cpp/src/distance/pairwise_distance_c.cpp
```diff
@@ -0,0 +1,72 @@
...
+  auto res_ptr = reinterpret_cast<raft::resources*>(res);
+
+  using mdspan_type           = raft::device_matrix_view<T const, int64_t, raft::row_major>;
+  using distances_mdspan_type = raft::device_matrix_view<T, int64_t, raft::row_major>;
+
+  auto x_mds         = cuvs::core::from_dlpack<mdspan_type>(x_tensor);
+  auto y_mds         = cuvs::core::from_dlpack<mdspan_type>(y_tensor);
+  auto distances_mds = cuvs::core::from_dlpack<distances_mdspan_type>(distances_tensor);
+
+  cuvs::distance::pairwise_distance(*res_ptr, x_mds, y_mds, distances_mds, metric, metric_arg);
```
cjnolet, line 50: Do we know that is always going to use the `float` (instantiated) version?

cpp/test/distance/pairwise_distance_c.cu
```diff
@@ -0,0 +1,64 @@
...
+#include <raft/core/device_mdarray.hpp>
+#include <raft/core/handle.hpp>
+#include <raft/random/rng.cuh>
+
+#include <cuvs/distance/distance.h>
+
+extern "C" void run_pairwise_distance(int64_t n_rows,
+                                      int64_t n_queries,
+                                      int64_t n_dim,
+                                      float* index_data,
```
cjnolet, line 29: What happens if someone were to compile these tests against a `double` pointer? It wouldn't cause the C++ to compile a new specialization would it? Ideally it would fail gracefull and state that it doesn't support the given type combinations.

cpp/test/distance/run_pairwise_distance_c.c
```diff
@@ -0,0 +1,72 @@
...
+  distances_tensor.dl_tensor.ndim               = 2;
+  distances_tensor.dl_tensor.dtype.code         = kDLFloat;
+  distances_tensor.dl_tensor.dtype.bits         = 32;
+  distances_tensor.dl_tensor.dtype.lanes        = 1;
+  int64_t distances_shape[2]                    = {n_rows, n_queries};
+  distances_tensor.dl_tensor.shape              = distances_shape;
+  distances_tensor.dl_tensor.strides            = NULL;
+
+  // run pairwise distances
+  cuvsPairwiseDistance(res, &dataset_tensor, &queries_tensor, &distances_tensor, metric, 2.0);
```
cjnolet, line 69: If you haven't done it before, can you try this w/ `double`? (Just making sure we don't get any unnexpected surprises from the UX or the compile times / binary size).

python/cuvs/cuvs/neighbors/ivf_flat/ivf_flat.pyx
```diff
@@ -30,9 +30,10 @@ from cuvs.distance_type cimport cuvsDistanceType
 from pylibraft.common import auto_convert_output, cai_wrapper, device_ndarray
 from pylibraft.common.cai_wrapper import wrap_array
 from pylibraft.common.interruptible import cuda_interruptible
-from pylibraft.distance.pairwise_distance import DISTANCE_TYPES
 from pylibraft.neighbors.common import _check_input_array
```
cjnolet, line 33: We'll probably need to bring this over at some point too (no rush for 24.06 since we already depend on `pylibraft`.

cjnolet (approved):
I think overall this PR looks great. I'm pre-approving, but do want to make sure we verify the behavior of the C api for unsupported type combinations. Maybe you've already done this, but I just want to avoid any unintended behaviors.

cpp/src/distance/pairwise_distance_c.cpp
benfred: removed the distance.h include in last commit

cpp/test/distance/pairwise_distance_c.cu
benfred: In cases where we are passing a tensor with an invalid dtype (float16 etc) - we should fail with an informative error message. (no new specializations will be compiled - will just use the existing float/double ones).

PR #145, opened 2024-05-22
Use rapids-build-backend

KyleFromNVIDIA:
Contributes to rapidsai/build-planning#31.


dantegd (approved):

python/cuvs/cuvs/_version.py
```diff
@@ -13,10 +13,22 @@
...
+    __git_commit__ = (
+        importlib.resources.files(__package__)
+        .joinpath("GIT_COMMIT")
+        .read_text()
+        .strip()
+    )
+except FileNotFoundError:
+    __git_commit__ = ""
+
+__all__ = ["__version__", "__git_commit__"]
```
jameslamb, line 34:
```suggestion
__all__ = ["__git_commit__", "__version__"]
```

If you end up pushing another commit here, could you please alphabetize this list? Doesn't matter *that* much when it's only 2 items, but that'd make it easier to view at a glance if it grows to more elements in the future.

jameslamb (approved):
I left 2 very very very minor comments... do whatever you want with them, and I don't need to review again when they're addressed.

.pre-commit-config.yaml
```diff
@@ -93,7 +93,7 @@ repos:
                 args: ["--toml", "pyproject.toml"]
                 exclude: (?x)^(^CHANGELOG.md$)
       - repo: https://github.com/rapidsai/dependency-file-generator
-        rev: v1.8.0
+        rev: v1.13.9
```
bdice, line 96:
```suggestion
        rev: v1.13.11
```

dependencies.yaml
```diff
@@ -88,6 +97,17 @@ channels:
...
+    common:
+      - output_types: [conda]
+        packages:
+          - scikit-build-core>=0.7.0
+      - output_types: [requirements, pyproject]
+        packages:
+          - scikit-build-core[pyproject]>=0.7.0
+      - output_types: [requirements, pyproject, conda]
+        packages:
+          - rapids-build-backend>=0.3.0,<0.4.0.dev0
```
bdice, line 109: For other RBB PRs, we've used an order more like:
- rapids-build-backend (conda, requirements, pyproject)
- scikit-build-core (conda)
- scikit-build-core (requirements, pyproject)

bdice (approved):
A couple small comments, but otherwise looks good! Resolve those as you see fit.

PR #148, opened 2024-05-23
Accept host_mdspan for IVF-PQ build and extend

### LINKED ISSUE #120: [BUG] IVF-PQ build / extend API does not support host mdspan
The pointer based APIs in RAFT for `ivf_pq::build` and `ivf_pq::extend` do accept host and device arrays (as well es pinned, and managed arrays). Accepting host arrays is an important feature: IVF-PQ algorithm allows us to build index from dataset that is larger than GPU memory (it can be even larger than host memory, if the data is `mmap`-ed). The cuVS interface should also support host arrays.

### LINKED ISSUE #143: [FEA] consistent serialization API
- IVF-Flat has `serialize` and `serialize_file `, where the first one serializes into a string.
- IVF-PQ has only `serialize`, which serializes into file
- CAGRA has `serialize` and `serialize_file` methods, similar to IVF-Flat

To ensure consistency, IVF-PQ should also have `serialize_file `(and corresponding `deserialize_file`) method.

tfeher:
This PR enables host input arrays for `ivf_pq::build` and `ivf_pq::extend`.

closes #120 
closes #143


cpp/include/cuvs/neighbors/ivf_pq.hpp
```diff
@@ -594,6 +594,146 @@ void build(raft::resources const& handle,
...
+ *   auto index = ivf_pq::build(handle, index_params, dataset);
+ * @endcode
+ *
+ * @param[in] handle
+ * @param[in] index_params configure the index building
+ * @param[in] dataset a host_matrix_view to a row-major matrix [n_rows, dim]
+ *
+ * @return the constructed ivf-pq index
+ */
+auto build(raft::resources const& handle,
```
cjnolet, line 615: Do we need the extend expose here too?

cjnolet (approved):

PR #151, opened 2024-05-24
CAGRA new vector addition

enp1s0:
This PR introduces the new vector addition feature to CAGRA.

Rel: https://github.com/rapidsai/raft/issues/1775
Original PR: https://github.com/rapidsai/raft/pull/2157

CAGRA-Q is not supported

## Usage
```cpp
auto additional_dataset = raft::make_host_matrix<float, int64_t>(res,updated_dataset_size, dim);
cuvs::neighbors::cagra::extend(handle, raft::make_const_mdspan(additiona_dataset.view()), cagra_index);
```

## Algorithm

Graph degree: d

The algorithm consists of two stages: rank-based reordering and reverse edge addition.
1. Rank-based reordering
1-1. Obtain d' (=2d) nearest neighbor vectors (V) of a given new vector using the CAGRA search
1-2. Count the number of detourable edges using the result of step 1 and the neighbor list of the input index. Then we prune (3*d/2) edges in the same way as the CAGRA graph optimization. Through this operation, we decide d/2 neighbors.
2. Reverse edge addition
2-1. Count the number of incoming edges for all nodes.
2-2. Add d/2 reverse edges from the nodes added to the neighbor list in Step 1 by replacing a node with a new node. To prevent the connection to the replaced node from being lost, we add the node to the neighbor list of the new node. This allow us to make a detour connection. The replaced nodes are the largest number of incoming edge nodes in the 2/d nodes from the back of the neighbor list without duplication with the nodes already in the neighbor list.

## Performance
In this experiment, we first split the dataset into two parts: the initial and the additional part. Then, we extend the CAGRA index built by the initial part to include the additional part.
search-eval

We can see a larger recall drop compared to the baseline by increasing the number of added vectors.
Therefore, rebuilding the CAGRA index is recommended when one wants to add a lot of vectors.


tfeher (approved):
Thanks @enp1s0 for the PR! The implementation was already reviewed in https://github.com/rapidsai/raft/pull/2157. The PR looks good to me.

tfeher:
Tests were failing on CI (CAGRA test hanging), restarting tests. According to @enp1s0 locally test were passing.

tfeher:
`AnnCagraAddNodesTest` hangs on V100, I could reproduce it locally.

tfeher:
Some `raft` -> `cuvs` namespace change were missing and the instantiated `cagra::extend` code was calling itself recursively due to a missing `detail `scope. The latter lead to the test hanging. These issues are fixed now.

cpp/include/cuvs/neighbors/cagra.hpp
```diff
@@ -448,6 +445,92 @@ void build_host(raft::resources const& handle,
                 raft::host_matrix_view<const uint8_t, int64_t, raft::row_major> dataset,
                 cuvs::neighbors::cagra::index<uint8_t, uint32_t>& idx);
 
+/**
+ * @defgroup cagra_cpp_index_add_nodes CAGRA graph node addition functions
+ * @{
+ */
+void add_graph_nodes(
```
cjnolet, line 452: If the intended usage of this feature is going to be for the user to call extend() then I think we should move add_nodes to the internal APIs to simplify things. Is there any other reason to keep this part of the public APIs?

cpp/include/cuvs/neighbors/cagra.hpp
```diff
@@ -448,6 +445,92 @@ void build_host(raft::resources const& handle,
...
+    updated_dataset_view,
+  const cuvs::neighbors::cagra::index<uint8_t, uint32_t>& idx,
+  raft::host_matrix_view<uint32_t, std::int64_t> updated_graph_view,
+  uint32_t batch_size = 0);
+
+/**
+ * @defgroup cagra_cpp_index_extend CAGRA extend functions
+ * @{
+ */
+void extend(raft::resources const& handle,
```
cjnolet, line 504: These all need to be documented with descriptions and copy/paste usage examples. Please also document all of them so that any function that shows up in the API docs will contain the proper documentation.

cpp/include/cuvs/neighbors/cagra.hpp
```diff
@@ -448,6 +445,92 @@ void build_host(raft::resources const& handle,
...
+            raft::device_matrix_view<const float, int64_t, raft::row_major> additional_dataset,
+            cuvs::neighbors::cagra::index<float, uint32_t>& idx,
+            uint32_t batch_size = 0);
+
+void extend(raft::resources const& handle,
+            raft::host_matrix_view<const float, int64_t, raft::row_major> additional_dataset,
+            cuvs::neighbors::cagra::index<float, uint32_t>& idx,
+            uint32_t batch_size = 0);
```
cjnolet, line 512: I'm a little torn about this batch size parameter- is this something we can infer? Without docs, it's hard for me to know what it means or how to set it so maybe that will help. In the absence of this knowledge, though, we are generally trying to keep the "build", "extend", and "search" APIs as consistent as possible across all the algorithms so anything we can do to minimize differences is helpful.

cpp/src/neighbors/cagra.cuh
```diff
@@ -394,6 +395,188 @@ void search(raft::resources const& res,
...
+      stride);
+
+    neighbors::cagra::detail::add_node_core<T, IdxT>(
+      handle, internal_index, additional_dataset_view, updated_graph);
+    raft::resource::sync_stream(handle);
+  }
+}
+
+/**
+ * @brief Add new vectors to a CAGRA index
```
cjnolet, line 484: Perhaps this comment should be moved to the hpp file because that is what is going to be visible to end users.

cjnolet (changes requested):
This is definitely a very welcome feature and it would be great to get this into 24.06. I haven't gone through the guts of the implementation yet but I wanted to leave some initial feedback about the public APIs to start.

cpp/include/cuvs/neighbors/cagra.hpp
enp1s0: Yes, this function is for users who do not want to use host memory a lot. In the new vector addition operation, we need a concatenated dataset of the existing and additional parts. So, the `extend` function allocates memory space for the concatenated dataset inside and copies both parts to it. However, if the dataset is very large, this method is not user-friendly. In this situation, users can add vectors without additional host memory by allocating a larger memory area in anticipation of adding vectors later and then calling this function to update the graph.

cpp/src/neighbors/cagra.cuh
enp1s0: Is it better to move the other comments to the hpp in this PR? or separate PR?

cpp/include/cuvs/neighbors/cagra.hpp
enp1s0: I replaced the `chunk_size` param with `extend_params`.

cpp/src/neighbors/cagra.cuh
enp1s0: I only moved the comment of the functions available in the include/hpp file. Some functions in the cuh file are not in the hpp file, and I didn't move their comments. I'll make a separate PR to make them public APIs and port the tests for them from RAFT for the next release.

cpp/include/cuvs/neighbors/cagra.hpp
```diff
@@ -644,11 +657,136 @@ auto build(raft::resources const& res,
...
+ *
+ * See [cagra::extend](#cagra::extend) for usage example
+ *
+ * @tparam T data element type
+ * @tparam IdxT type of the indices
+ *
+ * @param[in] handle raft resources
+ * @param[in] input_updated_dataset_view updated dataset (initial + additional dataset)
+ * @param[in] index CAGRA index
+ * @param[out] updated_graph_view updated graph
```
cjnolet, line 677: This seems awkward to return to the user- what are we expecting them to do with this parameter? From the user's perspective, this should be getting updated inside the index itself, shouldn't it?

cpp/include/cuvs/neighbors/cagra.hpp
```diff
@@ -644,11 +657,136 @@ auto build(raft::resources const& res,
...
+ * If there are no specific requirements for dataset allocation, then it is recommended
+ * to use cagra::extend to add new vectors.
+ *
+ * See [cagra::extend](#cagra::extend) for usage example
+ *
+ * @tparam T data element type
+ * @tparam IdxT type of the indices
+ *
+ * @param[in] handle raft resources
+ * @param[in] input_updated_dataset_view updated dataset (initial + additional dataset)
```
cjnolet, line 675: I'm really struggling with the need to expose this function through the public APIs, especially when the cagra index contains methods for the dataset to be attached or reattached after the fact. 

End users generally interact with these index APIs as a black box- that is, they invoke cuVS APIs to build and populate an index object and then they invoke functions to act upon that index object without ever having to care what's actually being done in the index object.

I understand the want to provide helper functions for power users but this isn't the correct way to provide these types of functions. If the only reason for exposing this function is because we want to provide a way for the user to give the dataset with the additional vectors appended then I think we need to put some additional thought into the existing "update_dataset" methods on the index and add an option to the extend params that affords more control to the user. 

The problem is that this function assumes the user knows or cares about a graph object, which is intentionally being hidden inside the index. If we remove the graph output, we are left with the extend() function and what I think should be an option for the user to specify whether they're attaching the whole detaset with the new vectors appended. Does that make sense?

cjnolet (changes requested):
Thanks again for the recent updates @enp1s0 and please accept my sincere apology that it's taken some time for me to get back to this PR. This past release turned out to require a significant amount of attention since it's the first release with the Ann code officially migrated from RAFT. 

I think these changes are shaping up but they aren't yet ready to be merged. It's very important that we focus on the user experience in the public APIs as a first and highest priority. That means making sure the docs usable and easily searchable, and that the APIs are consistent and easy to use.

cpp/include/cuvs/neighbors/cagra.hpp
enp1s0: Yes, I agree that providing the APIs as a black box would be better for most users. And providing an option to specify whether the passed dataset is an entire one sounds good. An alternative way is to provide a function to expand the memory space (num vectors) for the dataset and graph in the CAGRA index. We allocate the host memory for the new dataset because we need to replace the old dataset with the new one on the device memory. To do this, we 1) copy the old dataset on the host memory, 2) deallocate the device memory, 3) allocate the new memory space on the device memory, and 4) copy the old dataset and additional dataset to the device memory (I noticed that we only need to allocate the host memory for the old dataset). It would be beneficial if there is a functionality to expand the memory space for new vectors because we don't need to reallocate the dataset memory space as above every time as long as the memory space is left.

cjnolet:
@enp1s0 from what I can tell, my review feedback has not yet been addressed. This will need to be addressed before this feature is merged. 

We need to make sure our public APIs are as consistent as possible across algorithms. Each index type provides 3 different methods- build(), extend(), and search(). the arguments are also intentionally ordered in the same way across all of them. 

Further, ALL of the functions exposed in the public APIs need to be documented individually- that means not documenting only the first function. 

We need to make this new vector addition API fit within these 3 methods. This allows us to build on top of these APIs for features like sharding and multi-gpu scaling, irrespective of index type.

This new vector addition feature is important, but it needs to fit within the API constraints we have been establishing. Once my review comments are addressed, this feature can be merged.

enp1s0:
@cjnolet Yes, I understand. However, the first thing we need to do is to decide on the API design we are discussing [here](https://github.com/rapidsai/cuvs/pull/151#discussion_r1622567572). If we don't expose the `add_graph_nodes` function, I don't need to address the documentation issue about it. If we change the arguments of the `extend` function, it is better to address the issues related to it. That's why I haven't addressed these issues yet.

cjnolet:
I'm not opposed to the feature that add_graph_nodes is providing. We should find a way to enable it through extend(), so that the user has a unified experience.

cjnolet (changes requested):
I think your most recent changes look much better with some fairly minor things at this point. We need some more documentation around the why/when/how. I also think I'd prefer to keep the two additional optional args, rather than combine them into a new vocab item.

cpp/include/cuvs/neighbors/cagra.hpp
```diff
@@ -644,11 +682,172 @@ auto build(raft::resources const& res,
...
+ *
+ *   cagra::extend_params params;
+ *   cagra::extend(res, params, raft::make_const_mdspan(additional_dataset.view()), index);
+ * @endcode
+ *
+ * @param[in] handle raft resources
+ * @param[in] params extend params
+ * @param[in] additional_dataset additional dataset on device memory
+ * @param[in,out] idx CAGRA index
+ * @param[in] new_memory_buffers (Optional) user-allocated memory buffers for the extended dataset
```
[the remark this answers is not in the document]
enp1s0: I changed the arguments of the extend function to pass the dataset and graph optional views.

cpp/include/cuvs/neighbors/cagra.hpp
```diff
@@ -173,6 +174,43 @@ struct search_params : cuvs::neighbors::search_params {
 /**
  * @}
  */
+
+/**
+ * @defgroup cagra_cpp_extend_params CAGRA index extend parameters
```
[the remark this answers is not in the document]
enp1s0: I fixed the explanation of the parameter. Please let me know if the updated explanation is not sufficient.

enp1s0:
@cjnolet Thank you for reviewing the code. I have fixed it according to your comment. Can you check it again?

tfeher (approved):
The recent changes look good to me: `extend` now allows fine control of dataset and graph placement if needed.

cjnolet (approved):
The changes look great, @enp1s0! So happy to have this feature.

raydouglass:
Removing `ops-codeowners` from the required reviews since it doesn't seem there are any file changes that we're responsible for. Feel free to add us back if necessary.

PR #154, opened 2024-05-24
Add `refine` to public API

lowener:
Add `cuvs::neighbors::refine` to public API, with it's test


tfeher (changes requested):
Thanks Mickael for the PR! Looks good, just a few small comments.

cpp/include/cuvs/neighbors/refine.hpp
```diff
@@ -0,0 +1,149 @@
...
+#include <raft/core/host_mdarray.hpp>
+#include <raft/core/mdspan_types.hpp>
+#include <raft/core/resources.hpp>
+#include <raft/util/integer_utils.hpp>
+
+namespace cuvs::neighbors {
+/**
+ * @defgroup ann_refine Approximate Nearest Neighbors Refinement
+ * @{
+ */
```
cjnolet, line 32: I know it seems exhaustive and repetitive, but for documentation's sake, please copy and update the docs for each individual prototype below. We want to make sure what any function in the docs a user looks up is going to carry along the full doc string with it, not matter the type.

cjnolet (changes requested):
Looks great- just one thing wrt docs (and we should really make sure we're doing it across the public APIs).

tfeher (approved):
Thanks @lowener for the updates, the PR looks good to me!

cjnolet (approved):

PR #159, opened 2024-05-27
 Select k instantiations

benfred:
Since we're now using raft in header only mode, we don't have the compiled select_k instantiations in raft available to us anymore. Instead instantiate inside cuvs so we don't recompile in multiple spots.


cpp/include/cuvs/selection/select_k.hpp
```diff
@@ -0,0 +1,95 @@
...
+ * @param[in] select_min
+ *   whether to select k smallest (true) or largest (false) keys.
+ * @param[in] sorted
+ *   whether to make sure selected pairs are sorted by value
+ * @param[in] algo
+ *   the selection algorithm to use
+ * @param[in] len_i
+ *  optional array of size (batch_size) providing lengths for each individual row
+ */
+template <typename T, typename IdxT>
```
cjnolet, line 82: Sorry I missed this earlier- we don't want templates in the public APIs for the computational functions. (Really we don't want them for anything, but the classes/structs that don't end up calling CUDA kernels are impacted much less. 

We want to make sure only the wrappers for all concrete types are exposed here and the template function is exposed as a cuh file in src/, which the source file includes and invokes.

tfeher (approved):
Apart from the issue mentioned by Corey, the PR looks good to me.

cjnolet (approved):
I think at some point we might want to consider a params struct for the options just to make them easier to use, but for now I think the direct wrap is fine.

PR #173, opened 2024-06-04
Allow serialization on streams

benfred:
This change allows serializing to a std::ostream and deserializaing from a std::istream. This also fixes some minor docstring issues in the C++ serialization api's.

### DIFF
--- a/cpp/include/cuvs/neighbors/cagra.hpp
+++ b/cpp/include/cuvs/neighbors/cagra.hpp
@@ -727,55 +727,321 @@ void search(raft::resources const& res,
  * @defgroup cagra_cpp_serialize CAGRA serialize functions
  * @{
  */
-void serialize_file(raft::resources const& handle,
-                    const std::string& filename,
-                    const cuvs::neighbors::cagra::index<float, uint32_t>& index,
-                    bool include_dataset = true);
 
-void deserialize_file(raft::resources const& handle,
-                      const std::string& filename,
-                      cuvs::neighbors::cagra::index<float, uint32_t>* index);
+/**
+ * Save the index to file.
+ *
+ * Experimental, both the API and the serialization format are subject to change.
+ *
+ * @code{.cpp}
+ * #include <raft/core/resources.hpp>
+ * #include <cuvs/neighbors/cagra.hpp>
+ *
+ * raft::resources handle;
+ *
+ * // create a string with a filepath
+ * std::string filename("/path/to/index");
+ * // create an index with `auto index = cuvs::neighbors::cagra::build(...);`
+ * cuvs::neighbors::cagra::serialize(handle, filename, index);
+ * @endcode
+ *
+ * @param[in] handle the raft handle
+ * @param[in] filename the file name for saving the index
+ * @param[in] index CAGRA index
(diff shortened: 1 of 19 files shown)


cpp/include/cuvs/neighbors/cagra.hpp
```diff
@@ -727,55 +727,321 @@ void search(raft::resources const& res,
...
+ * cuvs::neighbors::cagra::deserialize(handle, is, &index);
+ * @endcode
+ *
+ * @param[in] handle the raft handle
+ * @param[in] is input stream
+ * @param[out] index the cagra index
+ */
 void deserialize(raft::resources const& handle,
-                 const std::string& str,
+                 std::istream& is,
```
divyegala, line 834: Should this be 
```suggestion
                 const std::istream& is,
```

cpp/src/neighbors/cagra_serialize.cuh
```diff
@@ -19,212 +19,36 @@
...
- * @param[in] index CAGRA index
- *
- */
-template <typename T, typename IdxT>
-void serialize_to_hnswlib(raft::resources const& handle,
-                          const std::string& filename,
-                          const cuvs::neighbors::cagra::index<T, IdxT>& index)
-{
-  detail::serialize_to_hnswlib<T, IdxT>(handle, filename, index);
-}
```
divyegala, line 162: Where is this function now?

divyegala (commented):
Were there no changes to Python wrappers?

cpp/include/cuvs/neighbors/cagra.hpp
benfred: I think this is correct as is.  Reading from the stream advances its location - so all of the istream methods to read from the stream aren't `const` . 

This also matches the deserialize behaviour from the corresponding raft api https://github.com/rapidsai/raft/blob/5cbda5db3dcaaaabc361917e0ef64e3bf2d76b4c/cpp/include/raft/neighbors/ivf_pq_serialize.cuh#L110

cpp/src/neighbors/cagra_serialize.cuh
benfred: This function wasn't being publicly exposed (this is in an internal header file in the src/ directory), and wasn't being called anywhere - and since it was dead code I deleted it.

If we need to publicly expose this we can add that as part of another PR that adds this functionality

benfred:
Thanks for the review @divyegala !

> Were there no changes to Python wrappers?

The python wrappers are using the c-api - so don't need updating (and the only python wrapper we have for serialization is  CAGRA which has been updated here, I'll expose the ivf-* methods in a different PR).

cpp/src/neighbors/cagra_serialize.cuh
divyegala: Can we keep it so we don't have to re-add it later? Very soon, we'll be adding CAGRA+HNSW interop to cuVS

benfred: re-added in last commit

divyegala (approved):

PR #179, opened 2024-06-06
[FEA] Support for Cosine distance in IVF-Flat

lowener:


tfeher (changes requested):
Thanks @lowener for this work! It looks good overall, I have left some comments. I still need to take a closer look at the modified distance kernels

cpp/src/cluster/detail/kmeans_balanced.cuh
```diff
@@ -163,6 +214,14 @@ inline std::enable_if_t<std::is_floating_point_v<MathT>> predict_core(
                          distances.data(),
                          n_clusters,
                          stream);
+      if (dataset_norm) {
```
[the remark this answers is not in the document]
lowener: No it is left over from some tests I ran. I removed that code.

cpp/test/neighbors/ann_ivf_flat.cuh
```diff
@@ -252,6 +254,124 @@ class AnnIVFFlatTest : public ::testing::TestWithParam<AnnIvfFlatInputs<IdxT>> {
     }
   }
 
+  void testIVFFlatCosine()
```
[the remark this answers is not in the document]
lowener: Initially because the naive knn used in the tests didn't support cosine so I switched to brute-force to test.
I now added cosine to the naive knn so the regular test code is now used.

cpp/src/neighbors/ivf_flat/ivf_flat_interleaved_scan.cuh
```diff
@@ -134,6 +137,10 @@ struct loadAndComputeDist {
 #pragma unroll
       for (int k = 0; k < Veclen; ++k) {
         compute_dist(dist, queryRegs[k], encV[k]);
+        if constexpr (ComputeNorm) {
+          norm_query += queryRegs[k] * queryRegs[k];
+          norm_data += encV[k] * encV[k];
+        }
```
tfeher, line 143: We should ask the question whether it is more efficient to compute the norms here or precompute them: 
- we could store norm of database vectors in the index
- query norm is recalculated every time we process vector in the cluster. I expect that it would be better to use precalculated query norm, but we should confirm that by measurement. Please open an issue to track this, and improve this in a follow-up PR.

tfeher (approved):
Thanks @lowener for the updates. The PR looks good overall. There might be a potential for improving the norm computations, but that need not stop this PR.

python/cuvs/cuvs/neighbors/ivf_flat/ivf_flat.pyx
```diff
@@ -58,12 +58,14 @@ cdef class IndexParams:
     metric : str, default = "sqeuclidean"
         String denoting the metric type.
         Valid values for metric: ["sqeuclidean", "inner_product",
-        "euclidean"], where
+        "euclidean", "cosine"], where
```
cjnolet, line 61: The pytest should be updated to run the cosine metric

cpp/test/neighbors/ann_ivf_flat.cuh
```diff
@@ -546,12 +562,14 @@ const std::vector<AnnIvfFlatInputs<int64_t>> inputs = {
   {20, 100000, 16, 10, 20, 1024, cuvs::distance::DistanceType::L2Expanded, true},
   {1000, 100000, 16, 10, 20, 1024, cuvs::distance::DistanceType::L2Expanded, true},
   {10000, 131072, 8, 10, 20, 1024, cuvs::distance::DistanceType::L2Expanded, false},
+  {3, 20, 1024, 3, 3, 4, cuvs::distance::DistanceType::CosineExpanded, false},
```
cjnolet, line 565: Why are we testing such a limited set of cases here? Ideally we would remove the metric from these params and always run the given params for all supported metrics.

cpp/src/cluster/detail/kmeans_balanced.cuh
```diff
@@ -141,6 +142,53 @@ inline std::enable_if_t<std::is_floating_point_v<MathT>> predict_core(
                         raft::compose_op<raft::cast_op<LabelT>, raft::key_op>());
       break;
     }
+    case cuvs::distance::DistanceType::CosineExpanded: {
```
cjnolet, line 145: We should update the doc strings in the public API also. If the doc strings don't currently specify which metrics are supported in ivf flat then they should.

cjnolet (changes requested):
The implementation is looking good. We need to make sure we are testing all cases (for all supported languages) and properly conveying the support metrics to the users (also for all supported languages).

cjnolet (approved):

PR #181, opened 2024-06-06
Scaling workspace resources

achirkin:
Use raft's large workspace resource for large temporary allocations during ANN index build.
This is the port of https://github.com/rapidsai/raft/pull/2194, which didn't make into raft before the algorithms were ported to cuVS.

### DIFF
--- a/cpp/src/neighbors/detail/cagra/cagra_build.cuh
+++ b/cpp/src/neighbors/detail/cagra/cagra_build.cuh
@@ -34,6 +34,7 @@
 
 // TODO: Fixme- this needs to be migrated
 #include "../../ivf_pq/ivf_pq_build.cuh"
+#include "../../ivf_pq/ivf_pq_search.cuh"
 #include "../../nn_descent.cuh"
 
 // TODO: This shouldn't be calling spatial/knn APIs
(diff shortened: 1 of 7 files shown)


cpp/src/neighbors/ivf_pq/ivf_pq_build.cuh
```diff
@@ -1694,12 +1712,22 @@ auto build(raft::resources const& handle,
...
+      raft::resource::get_large_workspace_resource(handle);
+    if (sizeof(float) * n_rows_train * index.dim() * kTolerableRatio <
+        raft::resource::get_workspace_free_bytes(handle)) {
+      big_memory_resource = device_memory;
+    }
 
     // Besides just sampling, we transform the input dataset into floats to make it easier
     // to use gemm operations from cublas.
-    rmm::device_uvector<float> trainset(n_rows_train * index.dim(), stream, device_memory);
+    rmm::device_uvector<float> trainset(n_rows_train * index.dim(), stream, big_memory_resource);
```
tfeher, line 1730: While we are here, we could improve the error message:
```suggestion
    rmm::device_uvector<float> trainset(0, stream, big_memory_resource);
    try {
      trainset.resize(n_rows_train * index.dim(), stream);
    } catch (raft::logic_error& e) {
      RAFT_LOG_ERROR("Insufficient memory for kmeans training set allocation. Please decrease kmeans_trainset_fraction, or set large_workspace_resource appropriately.");
      throw; 
   }
  // catch (std::bad_alloc& e) { } // In bad_alloc always wrapped as logic_error?
```

cpp/src/neighbors/detail/cagra/cagra_build.cuh
```diff
@@ -162,11 +162,26 @@ void build_knn_graph(
...
+  constexpr size_t kMaxBatchSize     = 4096;  // No more perf beyond this
+  constexpr size_t kMinBatchSize     = 128;   // Too slow if smaller
+  // Heuristic: how much of the workspace we can spare for the queries.
+  // The rest is going to be used by ivf_pq::search
+  const auto workspace_queries_bytes = raft::resource::get_workspace_free_bytes(res) / 5;
+  auto max_batch_size =
+    std::min<size_t>(workspace_queries_bytes / sizeof(DataT) / dataset.extent(1), kMaxBatchSize);
+  // Heuristic: if the workspace is too small for a decent batch size, switch to use the large
+  // resource with a default batch size.
+  if (max_batch_size < kMinBatchSize) {
```
tfeher, line 181: If `max_batch_size = 128`, then we would be running with that batch size using workspace memory. Is that batch size already large enough that it is faster than keeping `defaultBatchSize` and using `large_workspace_resource`?

cpp/src/neighbors/detail/cagra/cagra_build.cuh
```diff
@@ -162,11 +162,26 @@ void build_knn_graph(
...
+    max_batch_size = kDefaultBatchSize;
+    workspace_mr   = raft::resource::get_large_workspace_resource(res);
+  }
```
tfeher, line 184: While the intention is good, I am not sure if adding more heuristics is the way to go. The code looks good to me, but it adds complexity, and will make it more difficult to reproduce future memory errors.

Since IVF-PQ would anyways adjust its internal batch size according to the available workspace size, couldn't we just fix here `max_batch_size = 4096`.

cpp/src/neighbors/ivf_pq/ivf_pq_build.cuh
```diff
@@ -1796,7 +1824,7 @@ auto build(raft::resources const& handle,
                          trainset.data(),
                          labels.data(),
                          params.kmeans_n_iters,
-                         &managed_memory_upstream);
+                         &managed_memory);
```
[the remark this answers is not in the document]
achirkin: Not exactly. Originally, these two functions were accessing the data from both host and device to spare some boilerplate coding time on writing explicit data transfers.
As I see today, `train_per_subset` does not use this resource at all, while `train_per_cluster` does use it and indeed accesses the data on both device and host. Thanks for noting this! I'll remove the argument where possible.

cpp/src/neighbors/ivf_pq/ivf_pq_build.cuh
```diff
@@ -1515,12 +1517,22 @@ void extend(raft::resources const& handle,
...
+  rmm::device_async_resource_ref batches_mr = device_memory;
+  if (n_rows * (index->dim() * sizeof(T) + index->pq_dim() + sizeof(IdxT) + sizeof(uint32_t)) >
+      free_mem) {
+    labels_mr = large_memory;
+  }
   // Allocate a buffer for the new labels (classifying the new data)
-  rmm::device_uvector<uint32_t> new_data_labels(n_rows, stream, device_memory);
-  free_mem -= sizeof(uint32_t) * n_rows;
+  rmm::device_uvector<uint32_t> new_data_labels(n_rows, stream, labels_mr);
+  if (labels_mr == device_memory) { free_mem -= sizeof(uint32_t) * n_rows; }
```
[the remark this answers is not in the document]
achirkin: I believe the value returned by `get_workspace_free_bytes` does not take into account the alignment/padding. The reason is that the workspace resource is a limiting adapter around the (possibly pooled, padded, aligned) memory resource; it counts the bytes requested by the user, not the bytes allocated on the device. But this is ok: this whole construct does not represent the actual device memory anyway; the workspace resource is supposed to be limited to a fraction of the device memory, so it should allow extra padding bytes with no problem.

So I think the original and your suggested code should result in identical update to `free_mem`. Do you think I still better apply the suggestion for readability?

cpp/src/neighbors/detail/cagra/cagra_build.cuh
achirkin: The trade-off is that taking more memory here will limit the memory available to IVF-PQ internal batching.

If we fix the batch size here and there's not enough memory, we could end up with a few nasty situations depending on the size:
  a) (very large batch) OOM throws here 
  b) (large batch) OOM throws further in IVF-PQ search (not sure if this one is possible)
  c) (a bit too large batch) IVF-PQ internally selects very small batch and runs slowly
  d) (small batch) IVF-PQ is limited by the given batch size and runs slowly

So I don't see how we can easily go here without a heuristics.

cpp/src/neighbors/detail/cagra/cagra_build.cuh
achirkin: Hard to say at this point. I did some benchmarking back in February hence the value, but probably we should test it more. I'd suggest we can do more thorough benchmarking later and update the constants in a separate PR.

cpp/src/neighbors/detail/cagra/cagra_build.cuh
tfeher: Assuming `max_batch_size = 4096`, we are talking about `4096*O(graph_degree) + O(dataset_dim)` memory allocated by CAGRA build. This is negligible compared to the internal memory usage of IVF-PQ. Therefore I would just leave the constant large batch here, and leave it to IVF-PQ to do any internal batching if necessary. Please correct me if I am missing something.

cpp/src/neighbors/ivf_pq/ivf_pq_build.cuh
```diff
@@ -64,6 +64,7 @@
 #include <variant>
 
 namespace cuvs::neighbors::ivf_pq::detail {
+using raft::RAFT_NAME;                       // TODO: this is required for RAFT_LOG_XXX messages.
```
tfeher, line 67: https://github.com/rapidsai/raft/issues/2357

cpp/src/neighbors/ivf_pq/ivf_pq_build.cuh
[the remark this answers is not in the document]
tfeher: It does  take [alignment into account](https://github.com/rapidsai/rmm/blob/dcca09818f1c5b5b1936f684b0ff9cca5683e63f/include/rmm/mr/device/limiting_resource_adaptor.hpp#L125-L126). But in general you are right, it does not consider all possible factors. So yes, I accept that for our purpose it would be fine to track size in `free_mem` variable manually. 

I think the suggested code is simpler to read, please apply the suggestions.

tfeher (changes requested):

cpp/src/neighbors/detail/cagra/cagra_build.cuh
achirkin: Hmm, looking through the code one more time, to me it seems that `build_knn_graph` itself uses the workspace resource only to batch the dataset. This is the asymptotic:
  1. build_knn_graph (vec_batches) :
      `max_batch_size * O(dataset_dim)`
  2. ivf_pq::search():
      `max_batch_size * (O(dataset_dim) + O(max_samples) + O(k * n_probes))`

Normally we only have problems with the non-fused version of IVF-PQ and the large probe count, so that the max_batch_size * max_samples can reach gigabytes-scale. Then the heuristics for the `build_knn_graph` seems indeed unnecessary.

achirkin: I've updated the code to leave just one extra constant `kMinWorkspaceRatio` and greatly simplify the logic. @tfeher  please have a look.

tfeher (approved):
Thanks Artem for the update, LGTM.

cpp/src/neighbors/detail/cagra/cagra_build.cuh
tfeher: ok, thanks!

PR #191, opened 2024-06-16
DOC: update notebook link

raybellwaves:
This updates the notebook link from https://github.com/rapidsai/cuvs/tree/HEAD/exmples (doesn't exist) to https://github.com/rapidsai/cuvs/tree/branch-24.08/notebooks

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -6,7 +6,7 @@ Useful Resources
 
 .. _cuvs_reference: https://docs.rapids.ai/api/cuvs/stable/
 
-- `Example Notebooks <https://github.com/rapidsai/cuvs/tree/HEAD/exmples>`_: Self-contained Code examples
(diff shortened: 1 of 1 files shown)


docs/source/index.rst
```diff
@@ -6,7 +6,7 @@ Useful Resources
 
 .. _cuvs_reference: https://docs.rapids.ai/api/cuvs/stable/
 
-- `Example Notebooks <https://github.com/rapidsai/cuvs/tree/HEAD/exmples>`_: Self-contained Code examples
+- `Example Notebooks <https://github.com/rapidsai/cuvs/tree/branch-24.08/notebooks>`_: Self-contained Code examples
```
cjnolet, line 9: We should probably use `HEAD` here instead of `branch-24.08`. If we use HEAD then we only need to update this link when/if the notebooks directory should change, which won't be often. If we use a specific branch name, we'll need to update this link for each and every release and history tells me it will get stale very quickly.

cjnolet (approved):

PR #192, opened 2024-06-17
Fix a CAGRA graph opt bug

enp1s0:
There is a bug in the current CAGRA graph rank-based neighbor reordering process. A low recall or illegal memory access can occur if there are many detourable nodes from a node to its neighbors, e.g. there is a small subgraph in the initial kNN graph. This PR fixes this.


tfeher (approved):
Thanks @enp1s0 for the fix. The PR looks good to me.

PR #193, opened 2024-06-17
fix library name in docs ('cuvs' not 'pycuvs')

jameslamb:
The Python library is distributed as conda package `cuvs`, but the installation docs say `pycuvs`. This fixes that.

Looked for other uses like this:

```shell
git grep -i pycuvs
```

Didn't find any.

### DIFF
--- a/README.md
+++ b/README.md
@@ -34,9 +34,9 @@ cuVS contains state-of-the-art implementations of several algorithms for running
 
 cuVS comes with pre-built packages that can be installed through [conda](https://conda.io/projects/conda/en/latest/user-guide/getting-started.html#managing-python). Different packages are available for the different languages supported by cuVS:
 
-| Python | C++ | C |
-|--------|-----|---|
-| `cuvs`| `libcuvs` | `libcuvs_c` |
+| Python | C/C++                       |
+|--------|-----------------------------|
+| `cuvs` | `libcuvs`, `libcuvs-static` |
 
 ### Stable release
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -36,19 +36,21 @@ cuVS comes with pre-built packages that can be installed through [conda](https:/
 
 | Python | C++ | C |
 |--------|-----|---|
-| `pycuvs`| `libcuvs` | `libcuvs_c` |
+| `cuvs`| `libcuvs` | `libcuvs_c` |
```
raydouglass, line 39: I don't think there is a `libcuvs_c` package, but there is a `libcuvs-static`: https://github.com/rapidsai/cuvs/blob/branch-24.08/conda/recipes/libcuvs/meta.yaml#L18

jameslamb: There is a `libcuvs_c` shared library, with a C API over `libcuvs`:

https://github.com/rapidsai/cuvs/blob/9dc3a4dc5085cb5eb92d18ca1eafd5eef3dc405e/cpp/CMakeLists.txt#L556

https://github.com/rapidsai/cuvs/blob/9dc3a4dc5085cb5eb92d18ca1eafd5eef3dc405e/cpp/CMakeLists.txt#L588-L592

But you're right, it doesn't appear there's a conda package being distributed under that name.

I just pushed a commit updating the relevant docs, but maybe we should get another review from @benfred or @cjnolet to be sure the docs are representing these packages the way they'd like.

dantegd (approved):
changes lgtm

README.md
dantegd: I don't think there would be much point of a `libcuvs_c`, since it just contains the C-API that **requires** `lubcuvs`, so this change makes sense to me. Would like confirmation from @benfred that I'm not misremembering something.

PR #200, opened 2024-06-25
User flag to construct CAGRA index with dataset

### LINKED ISSUE #199: [FEA] Expose CAGRA construct_index_with_dataset flag to public API

tarang-jain:
Currently we would always construct the CAGAR index with dataset if the dataset fits into GPU mem. If it does not fit, then we fall back to constructing index with graph only. In this PR a simple flag is added to `index_params` to allow the user a choice if they want to disable constructing the index with the dataset (i.e. only construct with the only graph instead).

Closes #199


divyegala (approved):

cpp/include/cuvs/neighbors/cagra.hpp
```diff
@@ -104,6 +105,16 @@ struct index_params : cuvs::neighbors::index_params {
...
+
+  /**
+   * Whether to add the dataset content to the index, i.e.:
+   *
+   *  - `true` means the index is filled with the dataset vectors and ready to search after calling
+   * `build`.
+   *  - `false` means `build` only builds the graph, but
+   * the user is expected to update dataset separately.
+   */
+  bool add_data_on_build = true;
```
achirkin, line 117: I think we need a different parameter name here. `add_data_on_build` is [already defined for the index_params](https://github.com/rapidsai/cuvs/blob/581c9cc9ca23c1d828d28f1ad84e1c53fe5077c6/cpp/include/cuvs/neighbors/common.hpp#L101), and I believe it has a slightly different meaning. It's meant to be used together with `extend` function; in IVF-* methods we allow adding any new data after clustering this way. In CAGRA in the current form, the user would only be allowed to add the data that is present in the graph (and not via `extend` function).

Or, at the very least, use explicit `using cuvs::neighbors::index_params::add_data_on_build;` instead of shadowing the parent definition.

achirkin (changes requested):

tfeher:
@tarang-jain could you clarify the PR title and description? Currently the word "save" is used, and that makes me associate to serialization, but for serialization we already have the `include_dataset` flag.

I assume that problem you are addressing is index construction: whether to construct the index with the dataset. In that case the description could be updated like: "currently we would always construct the CAGAR index with dataset if the dataset fits into GPU mem. If it does not fit, then we fall back to constructing index with graph only. In this PR a simple flag is added to index_params to allow the user a choice if they want to disable constructing the index with the dataset (i.e. only construct with the only graph instead)."

cpp/include/cuvs/neighbors/cagra.hpp
tarang-jain: I have moved `add_data_on_build` to the ivf `index_params` and added a new arg for cagra -- `populate_data`. The PR is also green on CI right now.

cpp/include/cuvs/neighbors/cagra.hpp
```diff
@@ -104,6 +105,16 @@ struct index_params : cuvs::neighbors::index_params {
...
+  /**
+   * Whether to add the dataset content to the index, i.e.:
+   *
+   *  - `true` means the index is filled with the dataset vectors and ready to search after calling
+   * `build` provided there is enough memory available.
+   *  - `false` means `build` only builds the graph and the user is expected to
+   * update the dataset using cuvs::neighbors::cagra::update_dataset. CAGRA does not have `extent`
+   * API.
+   */
+  bool populate_data = true;
```
cjnolet, line 117: I think we should use a more descriptive naming here- this doesn't say much at all. What does it mean to populate data? I'd take that to mean we have an array and we're filling it up with something, but I believe what we are really doing is attaching a dataset to the index, right? 

I propose we name this `attach_dataset_on_build` so there's no confusion here. That's also more inline with the naming used in the IVF methods: `add_data_on_build`.

cpp/include/cuvs/neighbors/cagra.hpp
```diff
@@ -448,6 +463,23 @@ struct index : cuvs::neighbors::index {
  *   auto distances = raft::make_device_matrix<float>(res, n_queries, k);
  *   cagra::search(res, search_params, index, queries, neighbors, distances);
  * @endcode
+ *   In the above example, the index is populated with the dataset and is ready for search. In
```
cjnolet, line 466: Again, I think we should be careful about the nomenclature here. When I see "populate", I think of a chunk of memory being materialized into something. In reality, I think what we're doing is attaching a view of the dataset to the index. We should be very explicit here allthewhile attempting to be as consistent with prior naming standards as possible (e.g. "attach" vs "add").

cjnolet (changes requested):

cpp/include/cuvs/neighbors/common.hpp
```diff
@@ -90,15 +90,6 @@ struct index_params {
   cuvs::distance::DistanceType metric = cuvs::distance::DistanceType::L2Expanded;
   /** The argument used by some distance metrics. */
   float metric_arg = 2.0f;
-  /**
```
tfeher, line 93: Why is it removed from here? Is it only that we can repeat the docstring in each subclass? If that is the case, why not keep this and add the following with the docstring in the subclasses?
```c++
/** docstring */
using cuvs::neighbors::index_params;
```

tfeher (changes requested):

cpp/include/cuvs/neighbors/common.hpp
tarang-jain: `add_data_on_build` is added separately to the index params for IVF-Flat and IVF-PQ. It is not used in other algorithms.

tarang-jain:
Closing in favor of #213

PR #213, opened 2024-07-05
Complete Migration of IVF Helpers / Features from RAFT

tarang-jain:
Utils / Helpers to enable FAISS migration to cuVS from RAFT.


cpp/include/cuvs/neighbors/cagra.hpp
```diff
@@ -105,6 +106,30 @@ struct index_params : cuvs::neighbors::index_params {
                graph_build_params::ivf_pq_params,
                graph_build_params::nn_descent_params>
     graph_build_params;
+  /**
+   * Whether to add the dataset content to the index, i.e.:
+   *
+   *  - `true` means the index is filled with the dataset vectors and ready to search after calling
+   * `build` provided there is enough memory available.
+   *  - `false` means `build` only builds the graph and the user is expected to
+   * update the dataset using cuvs::neighbors::cagra::update_dataset. CAGRA does not have `extent`
```
tfeher, line 115: We do have now `extend`. I would extend the description as 
```suggestion
   * update the dataset using cuvs::neighbors::cagra::update_dataset.
   *
   * Regardless of the value of `attach_dataset_on_build`, the search graph is created using all
   * the vectors in the dataset.  Setting `attach_dataset_on_build = false` can be useful if the 
   * the user needs to build only the search graph but does not intend to search it using CAGRA
   * (e.g. search using another graph search algorithm), or if specific memory placement options 
   * need to be applied on the dataset before it is attached to the index using `update_dataset`.
```

cpp/include/cuvs/neighbors/cagra.hpp
```diff
@@ -468,6 +497,23 @@ struct index : cuvs::neighbors::index {
  *   auto distances = raft::make_device_matrix<float>(res, n_queries, k);
  *   cagra::search(res, search_params, index, queries, neighbors, distances);
  * @endcode
+ *   In the above example, the dataset is attached to the index and the index is ready for
```
tfeher, line 500: I think it is sufficient to keep the example in the docstring of `attach_dataset_on_build`.

cpp/include/cuvs/neighbors/ivf_flat.hpp
```diff
@@ -61,6 +61,18 @@ struct index_params : cuvs::neighbors::index_params {
...
+  /**
+   * Whether to add the dataset content to the index, i.e.:
+   *
+   *  - `true` means the index is filled with the dataset vectors and ready to search after calling
+   * `build`.
+   *  - `false` means `build` only trains the underlying model (e.g. quantizer or clustering), but
+   * the index is left empty; you'd need to call `extend` on the index afterwards to populate it.
+   * NB: for CAGRA index, `false` means `build` only builds the graph and the user is expected to
+   * update the dataset using cuvs::neighbors::cagra::update_dataset. CAGRA does not have `extent`
+   * API.
```
tfeher, line 73: delete

cpp/src/neighbors/ivf_pq/ivf_pq_list.cuh
```diff
@@ -13,7 +13,6 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-#pragma once
```
tfeher, line 16: Why is this removed?

tfeher (changes requested):
Thanks @tarang-jain for the PR, looks good overall, just few small changes.

tarang-jain:
rerun tests

cpp/src/neighbors/ivf_pq/ivf_pq_list.cuh
tarang-jain: This header is removed in this PR

tfeher (approved):
Thanks Tarang for the update, LGTM!

cjnolet (approved):

PR #217, opened 2024-07-11
Add col-major support for brute force knn

benfred:


tarang-jain (approved):
Thanks for this PR. Looks good to me.

cjnolet (approved):

cjnolet:
I'm going to merge this now just to unblock @tarang-jain so we can finalize the faiss+cubs updates. We should follow up with the tile-by-tile solution to improve memory usage.

Issue #125, opened 2024-05-16
[FEA] Use `warp_select` inside of fusedL2KNN and haversine_knn


benfred:
We should upgrade the fusedL2KNN code and haversine_knn code to use our warp_select from raft (aka GridSelect from https://dl.acm.org/doi/10.1145/3581784.3607062 ), instead of using the existing faiss_select code.

Our warp_select code has shown itself to be substantially faster than faiss select, especially on smaller numbers of input rows.

Issue #110, opened 2024-05-08
[FEA] Reduce template parameters


cjnolet:
Currently, there's a lot of template params that could easily become runtime parameter. Every template param causes the need to compile more instantiations. If we reduce template params, we will improve compile time and binary size. 

- [ ] Pairwise distances distance metric arg 
- [ ] ANN filter type (make polymorphic)
- [ ] Reduce CAGRA / CAGRA-Q templates (make more runtime hyper-params) 

@tfeher @divyegala @achirkin @benfred please add to this list as opportunities arise.

cjnolet:
- [ ] Remove `Linkage` template parameter from the single linkage internals

cjnolet:
Linking https://github.com/rapidsai/cuvs/issues/1254

PR #218, opened 2024-07-11
Moving random ball cover

cjnolet:
This PR brings in RBC implementation from RAFT while also reducing the number of templates that are instantiated by moving the following templates to runtime parameters:
1. `dims`
2. booleans
3. distance functor

### Benefits of these changes

Allows for a reduction in cuML binary sizes, once cuML switches from RAFT's implementation to this one.

See:

https://github.com/rapidsai/cuml/issues/6626#issuecomment-2864936395


cpp/src/neighbors/faiss_select/DistanceUtils.h
```diff
@@ -0,0 +1,52 @@
+/**
+ * Copyright (c) Facebook, Inc. and its affiliates.
```
benfred, line 2: I'm a little surprised that this didn't cause an error from our copyright-verify script https://github.com/rapidsai/cuvs/blob/43ed8ccd9bf2a7f6297d396db36f1a0795a66ba7/.pre-commit-config.yaml#L102-L112 (iirc, I had to manually exclude these files in RAFT with the previous copyright checker). @KyleFromNVIDIA is this what you'd expect to happen here?

cpp/src/neighbors/ball_cover/registers-ext.cuh
```diff
@@ -0,0 +1,205 @@
+/*
```
benfred, line 1: nitpick: since we're no longer header only - I'd prefer to get rid of having `-ext.cuh` and `-inl.cuh` header files.

cpp/include/cuvs/neighbors/ball_cover.hpp
```diff
@@ -0,0 +1,362 @@
...
+ * the index and query are the same array. This function will
+ * build the index and assumes rbc_build_index() has not already
+ * been called.
+ *
+ * Usage example:
+ * @code{.cpp}
+ *
+ *  #include <raft/core/resources.hpp>
+ *  #include <raft/neighbors/ball_cover.cuh>
+ *  #include <raft/distance/distance_types.hpp>
```
benfred, line 210: we should update the example to use the cuvs headers here

cpp/include/cuvs/neighbors/ball_cover.hpp
```diff
@@ -0,0 +1,362 @@
...
+ * function does not build the index and assumes rbc_build_index() has
+ * already been called. Use this function when the index and
+ * query arrays are different, otherwise use rbc_all_knn_query().
+ *
+ * Usage example:
+ * @code{.cpp}
+ *
+ *  #include <raft/core/resources.hpp>
+ *  #include <raft/neighbors/ball_cover.cuh>
+ *  #include <raft/distance/distance_types.hpp>
```
benfred, line 319:
```suggestion
 *  #include <cuvs/neighbors/ball_cover.hpp>
 *  #include <cuvs/distance/distance_types.hpp>
```

cpp/src/neighbors/faiss_select/Comparators.cuh
```diff
@@ -0,0 +1,29 @@
+/**
```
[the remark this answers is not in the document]
benfred: I totally agree that we should move to our own select-k in a future release,   probably at the same time as the updating the fusedL2KNN code (which we are tracking in #125).  

Sorry if I wasn't clear above there: I was trying to say that I'd like to keep the faiss_select code in raft instead of moving to cuvs, since we eventually should stop using faiss_select - rather than we should stop using faiss_select in this PR.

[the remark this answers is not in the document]
cjnolet: Oh I see what you mean. I think we can do that for sure.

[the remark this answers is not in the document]
benfred: thanks! I've updated #125 to also mention updating RBC - (and I might take this on for the 24.10 release if I can find a bit of time then)

cjnolet:
Linking https://github.com/rapidsai/cuvs/issues/110

cpp/src/neighbors/ball_cover/registers-ext.cuh
cjnolet: Do you mean just consolidate them into one file or do you mean remove the intantiations / ignored extern templates?

cjnolet:
Metrics are showing this change adds ~78mb to the overall binary size. This is not acceptable.  

Screenshot from 2024-08-29 12-45-14

divyegala:
As of commit https://github.com/rapidsai/cuvs/pull/218/commits/8cc63b1f0105fdb5c91442fb987b867a50aa1fb3, the total size added is now ~20 MB

Screenshot 2025-05-07 at 4 01 58 PM

cpp/include/cuvs/neighbors/ball_cover.hpp
```diff
@@ -0,0 +1,352 @@
...
+ *  auto metric = cuvs::distance::DistanceType::L2Expanded;
+ *  ball_cover::index index(handle, X, metric);
+ *  ball_cover::build_index(handle, index);
+ * @endcode
+ *
+ * @param[in] handle library resource management handle
+ * @param[inout] index an empty (and not previous built) instance of
+ * cuvs::neighbors::ball_cover::index
+ */
+void build(raft::resources const& handle, index<int64_t, float>& index);
```
cjnolet, line 181: Just a note- we should have the factory function here too. We can do as a follow-on at some point. RBC isn't a heavily used API today.

cpp/src/neighbors/ball_cover.cu
```diff
@@ -0,0 +1,71 @@
+/*
+ * Copyright (c) 2021-2025, NVIDIA CORPORATION.
```
cjnolet, line 2: These are new files and I just realized we have 2021 in the date here. Mind changing these to just 2025?

notebooks/ivf_flat_example.ipynb
```diff
@@ -21,7 +21,7 @@
   },
```
cjnolet, line 21: Not sure if this change is valid. Do you mind looking at this notebook real quick just to verify this isn't stale?

notebooks/tutorial_ivf_pq.ipynb
```diff
@@ -14,16 +14,37 @@
   },
   {
```
cjnolet, line 15: This one too

notebooks/ivf_flat_example.ipynb
divyegala: I just reverted the changes.

divyegala (approved):

jameslamb (approved):
Approving, with the understanding that while this increases the binary size for cuVS a bit, it leads to a much larger compensating reduction in cuML binary sizes: https://github.com/rapidsai/cuml/issues/6626#issuecomment-2864936395

Issue #127, opened 2024-05-16
[BUG] Flakey rust unittest `brute_force::tests::test_l2`


benfred:
The `brute_force::tests::test_l2` unittest sometimes fails randomly with an error like

```
thread 'brute_force::tests::test_l2' panicked at cuvs/src/brute_force.rs:164:9:
assertion `left == right` failed
  left: 13
 right: 0
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace

failures:
    brute_force::tests::test_l2
```

Re-running the tests usually resolves, but we should find out why this test isn't reliable and fix.

PR #226, opened 2024-07-17
Mark the rust brute force unittest as flaky

benfred:
The rust `brute_force::tests::test_l2` test can sometimes randomly fail (https://github.com/rapidsai/cuvs/issues/127).

We should figure out whats causing the test failure, but in the meantime - this PR marks the test as flaky using the `mark_flaky_tests` crate, which will retry the test 3 times and pass it if any of those runs succeed.


cjnolet (approved):

jakirkham:
Thanks Ben and Corey! 🙏

PR #227, opened 2024-07-17
CAGRA: reduce argument count in select_and_run() kernel wrappers

achirkin:
A small change that reduces the number of arguments in one of the wrapper layers in the detail namespace of CAGRA. The goal is twofold:
  1) Simplify the overly long signature of `selet_and_run` (which has many instances) 
  2) Give access to all search parameters for future upgrades of the search kernel

This is to simplify the integration (and review) of the persistent kernel (https://github.com/rapidsai/cuvs/pull/215).
No performance or functional changes expected.


tfeher (commented):
Thanks Artem for the PR! The changes so far look good, but don't we need to change `search_*_cta_kernel-ext.cuh` files too?

achirkin:
Thanks for noticing this, Tamas! I've updated the `*-ext.cuh` files and filed a bug for these files not being used https://github.com/rapidsai/cuvs/issues/242

tfeher (approved):
Thanks Artem for updating the PR and filing the issue. LGTM.

PR #228, opened 2024-07-17
CAGRA bench: use device-side refinement when the data is on device

achirkin:
Fix CAGRA build benchmark always assuming the dataset is on host.
Before the fix, this would result in the refinement step being performed on host even if the data is on device; this would lead to a segfault.


tfeher (approved):
Thanks Artem for the fix, LGTM!

PR #229, opened 2024-07-17
Moving over C++ API of CAGRA+hnswlib from RAFT

divyegala:


cpp/include/cuvs/neighbors/hnsw.hpp
```diff
@@ -0,0 +1,386 @@
...
+ *   index<uint8_t>* hnsw_index = nullptr;
+ *   hnsw::deserialize(res, "index.bin", index->dim(), index->metric(), &hnsw_index);
+ *
+ *   // Delete index after use
+ *   delete hnsw_index;
+ * @endcode
+ */
+void deserialize(raft::resources const& res,
+                 const std::string& filename,
+                 int dim,
```
benfred, line 342: Rather than have the caller specify the dimension and distance metric - can we read this from the serialized cagra index instead ? 

This means the user doesn't have to track extra metadata about a serialized index separately - and also will match the api of our other serialization methods

divyegala: There's 2 facets to this:
1. The serialized index is not a CAGRA index anymore, it's serialized in hnswlib format
2. Creating an hnswlib index and offloading deserialization to their APIs needs us to know the dimension beforehand

So, we could probably read this information from the index but it comes at the cost of abstraction. We would have to write our wrapper to read the index and information and then somehow plug that into hnswlib to avoid a second read.

cpp/src/neighbors/detail/hnsw.hpp
```diff
@@ -0,0 +1,183 @@
...
+struct index_impl : index<T> {
+ public:
+  /**
+   * @brief load a base-layer-only hnswlib index originally saved from a built CAGRA index
+   *
+   * @param[in] filepath path to the index
+   * @param[in] dim dimensions of the training dataset
+   * @param[in] metric distance metric to search. Supported metrics ("L2Expanded", "InnerProduct")
+   */
+  index_impl(std::string filepath, int dim, cuvs::distance::DistanceType metric)
```
benfred, line 58: nitpick: pass by reference
```suggestion
  index_impl(const std::string & filepath, int dim, cuvs::distance::DistanceType metric)
```

cpp/include/cuvs/neighbors/hnsw.hpp
benfred: thanks for the explanation!

I took a look and it seems like while we could get the dim from the index (with some hacks), it isn't possible to get the distance metric from the hnswlib serialized index at all =(.

benfred (approved):
lgtm!

cpp/include/cuvs/neighbors/hnsw.hpp
divyegala: Exactly, the serialized format is not intuitive in terms of the information that is written out!

cpp/src/neighbors/detail/hnsw.hpp
benfred: gah - I messed the style up there , sorry about that

PR #230, opened 2024-07-17
Enable kernel & memcpy overlapping in IVF index building

abc99lr:
Currently, in IVF index building (both IVF-Flat and IVF-PQ), large dataset is usually in pageable host memory or mmap-ed file. In both case, after the cluster centers are trained, the entire dataset needs to be copied twice to the GPU -- one for assigning vectors to clusters, the other for copying vectors to the corresponding clusters. Both copies are done using `batch_load_iterator` in a chunk-by-chunk fashion. Since the source buffer is in pageable memory, the current `batch_load_iterator` implementation doesn't support kernel and memcopy overlapping. This PR adds support on prefetching with `cudaMemcpyAsync` on pageable memory. We achieve kernel copy overlapping by launching kernel first following by the prefetching of the next chunk. 

We benchmarked the change on L40S. The results show 3%-21% speedup on index building, without impacting the search recall (about 1-2%, similar to run-to-run variance). 
algo | dataset | model | with prefetching (s) | without prefetching (s) | speedup
-- | -- | -- | -- | -- | --
IVF-PQ | deep-100M | d64b5n50K | 97.3547 | 100.36 | 1.03
IVF-PQ | wiki-all-10M | d64-nlist16K | 14.9763 | 18.1602 | 1.21
IVF-Flat | deep-100M | nlist50K | 78.8188 | 81.4461 | 1.03

This PR is related to the issue submitted to RAFT: https://github.com/rapidsai/raft/issues/2106


abc99lr:
@tfeher , do you have any suggestions on how we should expose the overlapping feature to users? Currently, user need to create a stream pool with one stream in it (separate from the main stream), so we could achieve kernel copy overlapping. But this model is not intuitive. Supporting overlapping in ANN benchmark is not difficult, but for end-user, do we want to add a flag to our API (specifically `extend``)? Or do you have other suggestions?

cpp/src/neighbors/detail/ann_utils.cuh
```diff
@@ -471,12 +485,38 @@ struct batch_load_iterator {
...
+        "batch_load_iterator::prefetch(offset = %zu, size = %zu, row_width = %zu)",
+        size_t(prefetch_offset),
+        size_t(prefetch_size),
+        size_t(row_width()));
+      raft::copy(prefetch_dev_ptr_,
+                 source_ + prefetch_offset * row_width(),
+                 prefetch_size * row_width(),
+                 stream_);
+      raft::common::nvtx::pop_range();
+      stream_.synchronize();
```
tfeher, line 517: I am wondering whether this is the most intuitive place to sync. Before looking into the implementation details, I would have expected, that prefetch, like most of our memory operations are asynchronous, but it is apparently not. If we want to keep this, then please mention it in the docstring. Alternatively, one could consider syncing before load swaps pointers, or we could leave it up to the user. Any thoughts @achirkin?

tfeher (changes requested):
Thanks Rui for the PR, this is an important improvement for the build time of IVF methods! I In general it looks good, but there are a few small things that we shall clarify.

tfeher:
> currently, user need to create a stream pool [...] so we could achieve kernel copy overlapping. But this model is not intuitive. [...]  but for end-user, do we want to add a flag to our API?

This is a good question. In other words, to enable overlapping of clustering and H2D copies, we have to do the following:
```python
handle = DeviceResources(n_streams=1)
index = ivf_pq.build(build_params, dataset, resources=handle)
```

The question is, whether we would like to have an extra build option like `overlap_h2d_copies`, and spawn a extra stream in the background if that is set:
```python
build_params.overlap_hd2_copies = True
index = ivf_pq.build(build_params, dataset)
```

I would prefer that we use the first option, if we have the resources (stream pool) then we use that automatically. But we would need to update the docstring of `build`, so that the user is aware that we need a handle with a stream pool for the fastest build. What do you think @cjnolet?

cpp/src/neighbors/detail/ann_utils.cuh
tfeher: One thing to keep in mind that we are currently using pageable buffers, so the copy is synchronous anyways, the extra sync call just makes that more explicit.

cpp/src/neighbors/ivf_flat/ivf_flat_build.cuh
```diff
@@ -195,13 +196,24 @@ void extend(raft::resources const& handle,
...
+  // Determine if a stream pool exist and make sure there is at least one stream in it so we
+  // could use the stream for kernel/copy overlapping.
+  auto copy_stream = raft::resource::get_cuda_stream(handle);  // Using the main stream by default
+  if (handle.has_resource_factory(raft::resource::resource_type::CUDA_STREAM_POOL)) {
+    if (raft::resource::get_stream_pool_size(handle) >= 1) {
+      copy_stream = raft::resource::get_stream_from_stream_pool(handle);
+    }
+  }
   // Predict the cluster labels for the new data, in batches if necessary
+  bool enable_prefetch = false;
```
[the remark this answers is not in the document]
abc99lr: I didn't change it back to `true` after a benchmarking. After we decide how to expose, I'll modify it.

cpp/src/neighbors/detail/ann_utils.cuh
abc99lr: Pageable buffer copies is not entirely sync -- see [here, Synchronous point 1](https://docs.nvidia.com/cuda/cuda-driver-api/api-sync-behavior.html), the copy is sync until the copy to staging buffer is done. 

For this particular case, I think we could leave it to users. Users could add stream record/event wait APIs to ensure the pipeline is correct. What do you think?

abc99lr:
> > currently, user need to create a stream pool [...] so we could achieve kernel copy overlapping. But this model is not intuitive. [...]  but for end-user, do we want to add a flag to our API?
> 
> This is a good question. In other words, to enable overlapping of clustering and H2D copies, we have to do the following:
> 
> ```python
> handle = DeviceResources(n_streams=1)
> index = ivf_pq.build(build_params, dataset, resources=handle)
> ```
> 
> The question is, whether we would like to have an extra build option like `overlap_h2d_copies`, and spawn a extra stream in the background if that is set:
> 
> ```python
> build_params.overlap_hd2_copies = True
> index = ivf_pq.build(build_params, dataset)
> ```
> 
> I would prefer that we use the first option, if we have the resources (stream pool) then we use that automatically. But we would need to update the docstring of `build`, so that the user is aware that we need a handle with a stream pool for the fastest build. What do you think @cjnolet?

Hi @cjnolet , could you comment on the two choices Tamas summarized here?

cpp/src/neighbors/detail/ann_utils.cuh
abc99lr: We decided to keep the API as it is and add proper documentation.

cpp/src/neighbors/ivf_pq/ivf_pq_build.cuh
```diff
@@ -1573,9 +1574,18 @@ void extend(raft::resources const& handle,
...
+  // Determine if a stream pool exist and make sure there is at least one stream in it so we
+  // could use the stream for kernel/copy overlapping.
+  auto copy_stream = raft::resource::get_cuda_stream(handle);  // Using the main stream by default
+  if (handle.has_resource_factory(raft::resource::resource_type::CUDA_STREAM_POOL)) {
+    if (raft::resource::get_stream_pool_size(handle) >= 1) {
+      copy_stream = raft::resource::get_stream_from_stream_pool(handle);
+    }
+  }
   // Predict the cluster labels for the new data, in batches if necessary
+  bool enable_prefetch = false;
```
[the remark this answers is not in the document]
abc99lr: Fixed

cpp/src/neighbors/ivf_flat/ivf_flat_build.cuh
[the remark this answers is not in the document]
abc99lr: Fixed

cpp/src/neighbors/detail/ann_utils.cuh
```diff
@@ -503,19 +543,30 @@ struct batch_load_iterator {
                       size_type row_width,
```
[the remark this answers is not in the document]
abc99lr: Added one example

abc99lr:
Re-requesting reviews. 

Added documentations. Exposed the overlapping option to users. Users need to setup a stream pool in resource and pass that to the `build` or `extend` APIs.

cjnolet:
@abc99lr just to add my response here quickly- I do agree with Tamas' suggestion that we should utilize the infrastructure already available in the handle to inform the user-configurable behavior of algorithms as much as possible.

That being said, the expectations of these configurations do need to be communicated more clearly to the users through the docs, I +1 to improving the docstring.

abc99lr:
Thanks for the review @cjnolet . Could you let me know if there is anything specific we should add to the docstring? I have added more documentations in my changes last Friday.

cjnolet:
@abc99lr the doc description itself looks okay, but, but it would be helpful to users to add a line or two to the usage example below the description that sets a stream pool on the handle. I would also add a small code comment to the snippet that says something like “create a cuda stream pool to …”. 

Thanks again for the contribution!

cpp/include/cuvs/neighbors/ivf_flat.hpp
```diff
@@ -436,6 +460,10 @@ void build(raft::resources const& handle,
 /**
  * @brief Build the index from the dataset for efficient search.
  *
+ * Note, if dataset is in host memory and index_params.add_data_on_build = true, user could set a
+ * stream pool in the input raft::resource with (at least) one stream to enable kernel/copy
+ * overlapping.
+ *
  * Usage example:
  * @code{.cpp}
  *   using namespace cuvs::neighbors;
```
tfeher, line 469: I agree with Corey's last comment: it would be great to show here how to create the stream pool
```suggestion
 *   using namespace cuvs::neighbors;
 *   raft::resources handle;
 *   // Add an extra stream to overlap dataset copies with computation during index building.
 *   raft::resource::set_cuda_stream_pool(handle, std::make_shared<rmm::cuda_stream_pool>(1));
```

tfeher (changes requested):
Thanks @abc99lr  for the updates! It looks good, let's just add a small improvement to the docstring code examples.

cpp/include/cuvs/neighbors/ivf_flat.hpp
```diff
@@ -295,6 +295,10 @@ struct index : cuvs::neighbors::index {
 /**
  * @brief Build the index from the dataset for efficient search.
  *
+ * Note, if dataset is in host memory and index_params.add_data_on_build = true, user could set a
```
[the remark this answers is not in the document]
abc99lr: Removed. Good catch!

cpp/include/cuvs/neighbors/ivf_flat.hpp
abc99lr: Thanks for the suggestion. I added the examples

tfeher (approved):
Thanks Rui for the update, the PR looks good to me!

cjnolet:
@abc99lr there's some errors in the style checker. Can you please install and use pre-commit to guarantee that each commit will pass the style checker? It takes only a few seconds to install and it's automatic each time you commit. Instructions [here](https://docs.rapids.ai/api/cuvs/nightly/contributing/#python-pre-commit-hooks).

abc99lr:
> @abc99lr there's some errors in the style checker. Can you please install and use pre-commit to guarantee that each commit will pass the style checker? It takes only a few seconds to install and it's automatic each time you commit. Instructions [here](https://docs.rapids.ai/api/cuvs/nightly/contributing/#python-pre-commit-hooks).

Thanks, my pre-commit was installed in my previous conda env. I had to rebuild an env so forgot to add it. Thanks for the instruction. I think we should be good now.

PR #231, opened 2024-07-18
SNMG ANN

viclafargue:
This PR implements a distributed (single-node-multiple-GPUs) implementation of ANN indexes. It allows to build, extend and search an index on multiple GPUs.

Before building the index, the user has to choose between two modes :

**Sharding mode** : The index dataset is split, each GPU trains its own index with its respective share of the dataset. This is intended to both increase the search throughput and the maximal size of the index.
**Index duplication mode** : The index is built once on a GPU and then copied over to others. Alternatively, the index dataset is sent to each GPU to be built there. This intended to increase the search throughput.

SNMG indexes can be serialized and de-serialized. Local models can also be deserialized and deployed in index duplication mode.

bench

Migrated from https://github.com/rapidsai/raft/pull/1993


cpp/CMakeLists.txt
```diff
@@ -198,6 +194,21 @@ endif()
 # ##################################################################################################
 # * cuvs ---------------------------------------------------------------------
 
+if(BUILD_MG_ALGOS)
```
divyegala, line 197: I don't see this added as an option in the diff

cpp/CMakeLists.txt
```diff
@@ -487,11 +499,16 @@ if(NOT BUILD_CPU_ONLY)
...
+  if(BUILD_MG_ALGOS)
+    set(CUVS_COMMS_DEPENDENCY ucp ucs ucxx nccl)
+  endif()
+
   # Keep cuVS as lightweight as possible. Only CUDA libs and rmm should be used in global target.
   target_link_libraries(
     cuvs
     PUBLIC rmm::rmm raft::raft ${CUVS_CTK_MATH_DEPENDENCIES}
     PRIVATE nvidia::cutlass::cutlass $<TARGET_NAME_IF_EXISTS:OpenMP::OpenMP_CXX>
+            ${CUVS_COMMS_DEPENDENCY}
```
divyegala, line 511: Have you tried building with `BUILD_MG_ALGOS=OFF`? I wonder if CMake will throw an error saying that this variable is undefined.

Also, we want to add `if (BUILD_MG_ALGOS) target_compile_options(cuvs CUVS_BUILD_MG_ALGOS) endif()` so that we can optionally compile the header and throw an error if a user tries to include `ann_mg.hpp` header with `BUILD_MG_ALGOS=OFF`

cpp/include/cuvs/neighbors/ann_mg.hpp
```diff
@@ -0,0 +1,295 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include <nccl.h>
```
divyegala, line 19: This leaks NCCL dependency to the end-user. You can define the macro `RAFT_NCCL_TRY` in `cpp/src`

cpp/include/cuvs/neighbors/ann_mg.hpp
```diff
@@ -0,0 +1,295 @@
...
+
+struct nccl_clique {
+  nccl_clique(const std::vector<int>& device_ids);
+  const raft::device_resources& set_current_device_to_root_rank() const;
+  ~nccl_clique();
+
+  int root_rank_;
+  int num_ranks_;
+  std::vector<int> device_ids_;
+  std::vector<ncclComm_t> nccl_comms_;
```
divyegala, line 81: We will have to forward declare `ncclComm_t` so that we can avoid including NCCL header:
```
class ncclComm_t {};

```suggestion
  std::vector<std::unique_ptr<ncclComm_t>> nccl_comms_;
```

cpp/include/cuvs/neighbors/ann_mg.hpp
```diff
@@ -0,0 +1,295 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
```
divyegala, line 18: Going by my previous comment on passing a macro to enable this feature, this whole file can be guarded by:
```
#ifdef CUVS_BUILD_MG_ALGOS

... // all the file info
#else

... // throw an error
#endif

File: cpp/CMakeLists.txt (hunk 2 above)
Comment (viclafargue) [reply cpp/CMakeLists.txt:515]: Setting `BUILD_MG_ALGOS=OFF` does not seem to throw an error. Yes, just added a compile definition and exception check.

File: cpp/src/neighbors/ann_mg/ann_mg.cuh [hunk 6]
```diff
@@ -741,3 +747,11 @@ ann_mg_index<cagra::index<T, IdxT>, T, IdxT> distribute_cagra(
 }
 
 }  // namespace cuvs::neighbors::mg::detail
+
+#else
+
+static_assert(false,
+              "FORBIDEN_MG_ALGORITHM_IMPORT\n\n"
+              "Please recompile the cuVS library with MG algorithms.\n");
```
divyegala, line 755: `#error "FORBIDEN_MG_ALGORITHM_IMPORT\n\n"
              "Please recompile the cuVS library with CMake option `BUILD_MG_ALGOS=ON.\n"

divyegala: Also this error based ifdef should be in `cpp/include`, not `cpp/src` as users will include the former

viclafargue: `CUVS_BUILD_MG_ALGOS` is only defined during the build of `libcuvs.so` not during the build of user's target binary, right? Do we want CMake to edit the header file before installation? I think this can be done with a header template, and the `configure_file` command. But, is this worth it?

viclafargue: I agree that if there's a safeguard it only makes sense to add it to the header file in the include directory.
However, in my mind there's two goals :
1) Preventing a cuVS developer from including something that wasn't compiled. The linking process should fail already, but the solution you propose allows the developer to get an instructive error message.
2) Preventing a cuVS user from including something that wasn't compiled. Don't know if we want to do this.

The problem is that with the current solution a cuVS user would have to define `CUVS_BUILD_MG_ALGOS` to be able to read the declarations for SNMG ANN.

divyegala: Why would you prefer a user be able to read the declarations? Declaration without definitions will just lead to errors down the road, because compilation will succeed for the user but linking will fail if they accidentally include the header.

divyegala: cuVS user doesn't have to define CUVS_BUILD_SNMG_ALGOS in their build system, they just have to turn it on in cuVS CMake. By setting CUVS_BUILD_SNMG_ALGOS to a PUBLIC definition, the user will automatically inherit it.

viclafargue: Just noticed the existence of the `$CONDA_PREFIX/lib/cmake/cuvs/cuvs-targets.cmake` file. CMake should indeed be able to apply the compile definitions to the downstream projects thanks to it. Sorry, not very familiar with all of this.

cpp/test/neighbors/ann_mg.cuh
```diff
@@ -0,0 +1,450 @@
...
+    cudaGetDeviceCount(&n_devices);
+    std::cout << n_devices << " GPUs detected" << std::endl;
+    std::vector<int> device_ids(n_devices);
+    std::iota(device_ids.begin(), device_ids.end(), 0);
+
+    uint64_t n_rows_per_batch = 3000;  // [3000, 3000, 1000] == 7000 rows
+    cuvs::neighbors::mg::nccl_clique clique(device_ids);
+
+    // IVF-Flat
+    for (parallel_mode d_mode : {parallel_mode::REPLICATED, parallel_mode::SHARDED}) {
```
tfeher, line 88: This feels very verbose to manually repeat the test blocks for each algo. Additionally you are combining multiple test into a single test instance: parallel_modes x [vector search algo]. Ideally these should go into test params, define test params as tensor product of various options, and each test should run a single algo. This need not block the PR, but would be a nice to have.

cpp/src/neighbors/ann_mg/ann_mg.cuh
```diff
@@ -0,0 +1,747 @@
...
+      IdxT translation_offset  = 0;
+      for (int rank = 0; rank < num_ranks_; rank++) {
+        h_trans[rank] = translation_offset;
+        translation_offset += ann_interfaces_[rank].size();
+      }
+      auto d_trans = raft::make_device_vector<IdxT, IdxT>(root_handle_, num_ranks_);
+      raft::copy(
+        d_trans.data_handle(), h_trans.data(), num_ranks_, resource::get_cuda_stream(root_handle_));
+
+      cuvs::neighbors::detail::knn_merge_parts(in_distances.data_handle(),
```
tfeher, line 493: Do I understand correctly, that all threads participate here, each merge a part of the data?

tfeher (changes requested):
Thanks Victor for the PR, it looks really great! I have just a few smaller comments.

cpp/src/neighbors/ann_mg/ann_mg.cuh
```diff
@@ -0,0 +1,747 @@
...
+    RAFT_LOG_INFO("REPLICATED BUILD: %d*%drows", num_ranks_, n_rows);
+
+    ann_interfaces_.resize(num_ranks_);
+#pragma omp parallel for
+    for (int rank = 0; rank < num_ranks_; rank++) {
+      int dev_id                            = clique.device_ids_[rank];
+      const raft::device_resources& dev_res = clique.device_resources_[rank];
+      RAFT_CUDA_TRY(cudaSetDevice(dev_id));
+      auto& ann_if = ann_interfaces_[rank];
+      ann_if.build(dev_res, index_params, index_dataset);
```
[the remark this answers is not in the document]
viclafargue: The SNMG ANN API allows to distribute a pre-build and serialized local ANN index in REPLICATED fashion as demonstrated in the tests. Btw, serializing and deserializing the full SNMG ANN index is also possible. Finally, even if probably not ideal an SNMG ANN index can be be built in replicated mode : one build per GPU (as shown here).

cpp/src/neighbors/ann_mg/ann_mg.cuh
viclafargue: Unless mistaken, this section of code is supposed to run on the main thread aside of OpenMP. The merge should happen on the GPU designated as "root".

cpp/src/neighbors/ann_mg/ann_mg.cuh
```diff
@@ -0,0 +1,747 @@
...
+  } else if constexpr (std::is_same<AnnIndexType, ivf_pq::index<IdxT>>::value) {
+    auto idx = cuvs::neighbors::ivf_pq::build(
+      handle, *static_cast<const ivf_pq::index_params*>(index_params), index_dataset);
+    index_.emplace(std::move(idx));
+  } else if constexpr (std::is_same<AnnIndexType, cagra::index<T, IdxT>>::value) {
+    auto extents = raft::make_extents<int64_t>(index_dataset.extent(0), index_dataset.extent(1));
+    const bool host_acc   = decltype(index_dataset)::accessor_type::is_host_type::value;
+    const bool device_acc = decltype(index_dataset)::accessor_type::is_device_type::value;
+    auto dataset          = raft::make_mdspan<const T, int64_t, row_major, host_acc, device_acc>(
+      index_dataset.data_handle(), extents);
```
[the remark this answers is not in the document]
viclafargue: Yes, this is solely because of the matrix extent type. Unfortunately, because the IVF-Flat and IVF-PQ APIs were built against RAFT they also had to have a variable matrix extent type. I agree that we should in the end switch all matrix extent types to `int64_t`, but this should probably be done in an other PR. What do you think?

cpp/src/neighbors/ann_mg/ann_mg.cuh
tfeher: You are right. I have misread the code and did not see that the omp parallel region is already closed.

cpp/src/neighbors/ann_mg/ann_mg.cuh
```diff
@@ -0,0 +1,750 @@
...
+                            resource::get_cuda_stream(dev_res));
+          comms.device_send(d_distances.data_handle(),
+                            n_rows_of_current_batch * n_neighbors,
+                            clique.root_rank_,
+                            resource::get_cuda_stream(dev_res));
+          RAFT_NCCL_TRY(ncclGroupEnd());
+          resource::sync_stream(dev_res);
+        }
+      }
+#pragma omp barrier
```
tfeher, line 483: The parallel region ended a line above. The barrier construct has no effect outside a parallel region.  Please remove this, and all other stray barriers, it makes the code confusing to read.

For ref [OpenMP 5.2](https://www.openmp.org/spec-html/5.2/openmpse57.html): "An implicit barrier occurs at the end of a parallel region."

cpp/src/neighbors/ann_mg/ann_mg.cuh
```diff
@@ -0,0 +1,750 @@
...
+}
+
+template <typename AnnIndexType, typename T, typename IdxT>
+void ann_mg_index<AnnIndexType, T, IdxT>::search(
+  const cuvs::neighbors::mg::nccl_clique& clique,
+  const cuvs::neighbors::search_params* search_params,
+  raft::host_matrix_view<const T, IdxT, row_major> queries,
+  raft::host_matrix_view<IdxT, IdxT, row_major> neighbors,
+  raft::host_matrix_view<float, IdxT, row_major> distances,
+  IdxT n_rows_per_batch) const
```
tfeher, line 354: Do we really need this argument? Internally our ANN methods do batching if necessary, why do we need another level of batching here?

tfeher (changes requested):
A few additional comments

cpp/src/neighbors/ann_mg/ann_mg.cuh
viclafargue: This is to avoid getting an OOM error on the GPU running the merges. This value is set very high as a default, but if there's any issue, it will be as simple as lowering this value. In the end, when scaling on very large setups (multi nodes), we might opt for a tree merge which is less hungry for memory, but requires more steps. But this would be for a follow-up PR.

cpp/src/neighbors/ann_mg/ann_mg.cuh
viclafargue: Thanks for noticing this, was a mistake.

cpp/include/cuvs/neighbors/ann_mg.hpp
```diff
@@ -0,0 +1,1367 @@
...
+ * @param[in] n_rows_per_batch search batch size
+ *
+ */
+void search(const raft::resources& handle,
+            const cuvs::neighbors::mg::nccl_clique& clique,
+            const ann_mg_index<cagra::index<float, uint32_t>, float, uint32_t>& index,
+            const cagra::search_params& search_params,
+            raft::host_matrix_view<const float, uint32_t, row_major> queries,
+            raft::host_matrix_view<uint32_t, uint32_t, row_major> neighbors,
+            raft::host_matrix_view<float, uint32_t, row_major> distances,
```
tfeher, line 918:
```suggestion
            raft::host_matrix_view<const float, int64_t, row_major> queries,
            raft::host_matrix_view<uint32_t, int64_t, row_major> neighbors,
            raft::host_matrix_view<float, int64_t, row_major> distances,
```

cpp/include/cuvs/neighbors/ann_mg.hpp
```diff
@@ -0,0 +1,1367 @@
...
+ * @param[in] n_rows_per_batch search batch size
+ *
+ */
+void search(const raft::resources& handle,
+            const cuvs::neighbors::mg::nccl_clique& clique,
+            const ann_mg_index<cagra::index<int8_t, uint32_t>, int8_t, uint32_t>& index,
+            const cagra::search_params& search_params,
+            raft::host_matrix_view<const int8_t, uint32_t, row_major> queries,
+            raft::host_matrix_view<uint32_t, uint32_t, row_major> neighbors,
+            raft::host_matrix_view<float, uint32_t, row_major> distances,
```
tfeher, line 951:
```suggestion
            raft::host_matrix_view<const int8_t, int64_t, row_major> queries,
            raft::host_matrix_view<uint32_t, int64_t, row_major> neighbors,
            raft::host_matrix_view<float, int64_t, row_major> distances,
```

cpp/include/cuvs/neighbors/ann_mg.hpp
```diff
@@ -0,0 +1,1367 @@
...
+ * @param[in] n_rows_per_batch search batch size
+ *
+ */
+void search(const raft::resources& handle,
+            const cuvs::neighbors::mg::nccl_clique& clique,
+            const ann_mg_index<cagra::index<uint8_t, uint32_t>, uint8_t, uint32_t>& index,
+            const cagra::search_params& search_params,
+            raft::host_matrix_view<const uint8_t, uint32_t, row_major> queries,
+            raft::host_matrix_view<uint32_t, uint32_t, row_major> neighbors,
+            raft::host_matrix_view<float, uint32_t, row_major> distances,
```
tfeher, line 984:
```suggestion
            raft::host_matrix_view<const uint8_t, int64_t, row_major> queries,
            raft::host_matrix_view<uint32_t, int64_t, row_major> neighbors,
            raft::host_matrix_view<float, int64_t, row_major> distances,
```

tfeher (changes requested):
We are almost there, but the matrix extent type has to be fixed: we should distinguish the neighbor index type (IdxT, usually int64), and the mdspan extent type which is always int64. In a previous review I have marked such changes in `ann_mp.cuh`, but actually the public API in `ann_mg.hpp` has also a few places where this need to be fixed

cpp/src/neighbors/ann_mg/ann_mg.cuh
tfeher: Ok, thanks for the explanation!

cpp/src/neighbors/ann_mg/ann_mg.cuh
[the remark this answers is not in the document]
tfeher: Matrix extent type should be always int64_t. See my suggestions for the public API.

cpp/include/cuvs/neighbors/ann_mg.hpp
```diff
@@ -0,0 +1,1367 @@
...
+ * index_params.metric_arg = 2.0f;
+ * index_params.add_data_on_build = true;
+ * @endcode
+ *
+ */
+struct mg_index_params : public cuvs::neighbors::ivf_flat::index_params {
+  /** Distribution mode */
+  cuvs::neighbors::mg::distribution_mode mode;
+};
+}  // namespace cuvs::neighbors::ivf_flat
```
achirkin, line 66: What do you think about generalizing the mg_index_params into something like the following?
```C++
namespace cuvs::neighbors::mg {
template <typename Upstream>
struct mg_params :: public Upstream {
  /** Distribution mode */
  distribution_mode mode;
};
}
```
Then, you could use this in the search/build functions directly, with two benefits:
  1. No need to create a new type for every existing algorithm
  2. Potentially, you could make some of the functions in the detail namespace use the templated version of the struct rather than duplicating code
```C++
void search(const raft::resources& handle,
            const cuvs::neighbors::mg::nccl_clique& clique,
            const ann_mg_index<ivf_flat::index<T, IdxT>, T, IdxT>& index,
            const mg_params<ivf_flat::search_params>& search_params,
            raft::host_matrix_view<const T, IdxT, row_major> queries,
            raft::host_matrix_view<IdxT, IdxT, row_major> neighbors,
            raft::host_matrix_view<float, IdxT, row_major> distances)
```
(In this example, one could also modify ivf_flat  `index` type to add `using search_params_type = ivf_flat::search_params`, and then the whole MG search could be templated over the index type (i.e. `mg_params<typename Upstream::search_params_type>`))

achirkin: Here's an example of how I get around this with a similar approach in a dynamic batching prototype (but the API is surely simpler than this multi-gpu PR, because I only need the search methods) https://github.com/rapidsai/cuvs/pull/261/files#diff-89b13a31203fa009604a3a2bdb0aee01668bb68969d9193a2b31d6bf6dd3894aR29-R48

viclafargue: I like the idea of improving the API through templating (or even dynamic batching) and keeping everything under the `cuvs::neighbors::mg` namespace. The `distribution_mode` is only necessary during build time and the `merge_mode` during search time though. The `merge_mode` could be incorporated in the search parameter. However, there's way too many things to consider to modify this in time for the release. I think we could allow ourselves a one-time update of the public API if necessary during the second release as not too many people (we could hope) would be using it yet.

cpp/include/cuvs/neighbors/ann_mg.hpp
```diff
@@ -0,0 +1,1417 @@
...
+  ann_mg_index(const ann_mg_index&)                    = delete;
+  ann_mg_index(ann_mg_index&&)                         = default;
+  auto operator=(const ann_mg_index&) -> ann_mg_index& = delete;
+  auto operator=(ann_mg_index&&) -> ann_mg_index&      = default;
+
+  void deserialize_and_distribute(const raft::resources& handle,
+                                  const cuvs::neighbors::mg::nccl_clique& clique,
+                                  const std::string& filename);
+
+  void deserialize_mg_index(const raft::resources& handle,
```
cjnolet, line 214: Maybe just `deserialize()`? We know it's an mg index already.

cpp/include/cuvs/neighbors/ann_mg.hpp
```diff
@@ -0,0 +1,1417 @@
...
+  distribution_mode mode_;
+  int num_ranks_;
+  std::vector<ann_interface<AnnIndexType, T, IdxT>> ann_interfaces_;
+};
+
+/// \defgroup ann_mg_cpp_index_build ANN MG index build
+
+/// \ingroup ann_mg_cpp_index_build
+/**
+ *
```
cjnolet, line 270: Please write descriptions for all public API functions. The usage examples are great, but docs look much cleaner when there's nice descriptions of the expected behavior to accompany them.

cpp/include/cuvs/neighbors/ann_mg.hpp
```diff
@@ -0,0 +1,1417 @@
...
+ * @param[in] handle
+ * @param[in] clique configure the NCCL clique
+ * @param[in] index_params configure the index building
+ * @param[in] index_dataset a row-major matrix on host [n_rows, dim]
+ *
+ * @return the constructed IVF-Flat MG index
+ */
+auto build(const raft::resources& handle,
+           const cuvs::neighbors::mg::nccl_clique& clique,
```
cjnolet, line 335: If we can remove this argument then we can match the single gpu APIs. I think I'd prefer to put this argument on the resource handle. Please create a new resource type for the nccl clique like we do in cuml.

cpp/include/cuvs/neighbors/ann_mg.hpp
```diff
@@ -0,0 +1,1417 @@
...
+  std::vector<raft::device_resources> device_resources_;
+};
+
+using namespace raft;
+
+template <typename AnnIndexType, typename T, typename IdxT>
+class ann_interface {
+ public:
+  template <typename Accessor>
+  void build(raft::resources const& handle,
```
cjnolet, line 171: I'm not in love with having the build, search, extend functions as methods on the index. We have intentionally kept these as free functions so that the APIs are easier to handle and wrap in other languages. Please pull these back into free functions that accept the index object and have the index object only maintain simple needed state about the index (opaque to the user). 

I think we should be able to keep the free functions but still make the underlying APIs smart. We will need to build similar APIs for the tiered index also.

cpp/src/neighbors/ann_mg/ann_mg.cuh
```diff
@@ -0,0 +1,912 @@
...
+void check_omp_threads(const int requirements);
+
+template <typename AnnIndexType, typename T, typename IdxT>
+template <typename Accessor>
+void ann_interface<AnnIndexType, T, IdxT>::build(
+  raft::resources const& handle,
+  const cuvs::neighbors::index_params* index_params,
+  raft::mdspan<const T, matrix_extent<int64_t>, row_major, Accessor> index_dataset)
+{
+  if constexpr (std::is_same<AnnIndexType, ivf_flat::index<T, IdxT>>::value) {
```
cjnolet, line 43: We should probably centralize this logic somewhere so that's not only in the mg namespace. We will need to do this for the tiered index also (and we still shouldn't be doing this as methods, but rather free functions).

cpp/src/neighbors/ann_mg/ann_mg.cuh
```diff
@@ -0,0 +1,912 @@
...
+
+template <typename AnnIndexType, typename T, typename IdxT>
+void ann_interface<AnnIndexType, T, IdxT>::search(
+  raft::resources const& handle,
+  const cuvs::neighbors::search_params* search_params,
+  raft::host_matrix_view<const T, int64_t, row_major> h_queries,
+  raft::device_matrix_view<IdxT, int64_t, row_major> d_neighbors,
+  raft::device_matrix_view<float, int64_t, row_major> d_distances) const
+{
+  int64_t n_rows = h_queries.extent(0);
```
cjnolet, line 88: We should probaly centralize all of these proxy functions in the main `cuvs::neighbors` namespace. I'm not sure the best way to do this with free functions.

cpp/src/neighbors/ann_mg/ann_mg_cagra_int8_t_uint32_t.cu
```diff
@@ -0,0 +1,93 @@
...
+ *
+ * > python generate_ann_mg.py
+ *
+ */
+
+#include "ann_mg.cuh"
+
+namespace cuvs::neighbors::mg {
+
+#define CUVS_INST_ANN_MG_CAGRA(T, IdxT)                                                          \
```
cjnolet, line 30: Can you please verify how much your changes increase the binary size and compile times compared to cuvs in branch-24.10? You changes aren't creating any new crazy complex kernels so ideally they shouldn't drastically increase binary size or compile time. However it's worth checking so we're not surprised once we merge it over.

docs/source/cpp_api/neighbors_ann_mg.rst
```diff
@@ -0,0 +1,76 @@
...
+Index deserialize
+-----------------
+
+.. doxygengroup:: ann_mg_cpp_deserialize
+    :project: cuvs
+    :members:
+    :content-only:
+
+Distribute pre-built local index
+--------------------------------------
```
cjnolet, line 71: This usually throws a doc error when the underline doesn't have the exact same number of characters as the title.

cjnolet (changes requested):
Hey Victor. I'm so sorry for the delay but i've finally looked through all of this (it's a huge PR so thanks for being patient with me!). Most of my focus was on public API design and docs. More specifically, we need to make sure we're writing doc strings in a way where users understand when, why, and how. I also think we should stick to the free function style of the rest of cuVS, but i like that you've written proxy functions to generalize them. 

Hopefully we can get this feature into 24.10!

viclafargue:
By default, the replicated mode splits the query and gives an equal share to each GPU. However as discussed, to serve "online" use-cases we could instead detect single row queries and make the index run as a load balancer in this case. I think that most things should run fine for this as long as the user calls the search function in different threads. However, the search as implemented right now will systematically give the query to the first GPU in the clique. We could make this a random attribution instead. However, what would happen if an local index is used twice in parallel? We could maybe have a thread safe (mutex protected) structure in the MG index to ensure secure access of local indexes. Do you guys have insights on this?

cc @cjnolet @tfeher @achirkin

cjnolet:
> We could make this a random attribution instead. However, what would happen if an local index is used twice in parallel? We could maybe have a thread safe (mutex protected) structure in the MG index to ensure secure access of local indexes. Do you guys have insights on this?

Honestly, I think rather than making a special case for single-query batches, it'd be fine to just use an atomic counter and always send the next query to the next device. You could do this, for example, by incrementing a number to a particular max (e.g. number of gpus) and resetting it back to 0 each time it hits that max. This should be well documented, of course, but it would give users the ability to batch queries themselves and the behavior would be consistent across all batch sizes.

build.sh
```diff
@@ -18,7 +18,7 @@ ARGS=$*
 # scripts, and that this script resides in the repo dir!
 REPODIR=$(cd $(dirname $0); pwd)
 
-VALIDARGS="clean libcuvs python rust docs tests bench-ann examples --uninstall  -v -g -n --compile-static-lib --allgpuarch --no-nvtx --show_depr_warn --incl-cache-stats --time -h"
+VALIDARGS="clean libcuvs python rust docs tests bench-ann examples --uninstall  -v -g -n --compile-static-lib --allgpuarch --mg_algos_off --no-nvtx --show_depr_warn --incl-cache-stats --time -h"
```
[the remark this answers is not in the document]
viclafargue: I thought that it is an important feature that we would like to have by default. Even though it can be interesting to have an option to get a smaller build with fewer dependencies. Do we want to change this?

cpp/include/cuvs/neighbors/ann_mg.hpp
```diff
@@ -0,0 +1,1417 @@
...
+  std::vector<int> device_ids_;
+  std::vector<ncclComm_t> nccl_comms_;
+  std::vector<std::unique_ptr<pool_mr>> per_device_pools_;
+  std::vector<raft::device_resources> device_resources_;
+};
+
+using namespace raft;
+
+template <typename AnnIndexType, typename T, typename IdxT>
+class ann_interface {
```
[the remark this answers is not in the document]
viclafargue: I put the code into a new `iface.hpp` file instead for now as to avoid unnecessarily including all ANN types into the `common.hpp` file. Please tell me if you want things to be done otherwise.

cpp/src/neighbors/ann_mg/ann_mg.cuh
cjnolet: Honestly, I'd prefer to put it all in `common.hpp` to make it brutally simple for users. cuVS' public APIs are not header-only and they are just declaring functions and classes who's symbols are largely defined in the shared object so there's little to no overhead for having large headers in `include/`.

viclafargue: There's a circular dependency issue that makes things very messy. The iface struct depends on the definitions of `ivf_flat`, `ivf_pq` and `cagra`, which themselves depends on what is defined in `common.hpp`. I would have to either put include statements in the middle of the file or do so many forward declarations that it would render the code unreadable. Because of this, I believe that the best solution is to keep the iface in a separate file for now. Hope that's ok, else we can still discuss this more offline.

cjnolet: I don’t think bad smells in code should be mitigated this way. I think it’s worth expending a little extra effort to simplify the experience and layout of the public API. All headers exposed in the public API are intended to be included directly by users, and since cuVS is only strictly compiled, we should only be exposing a minimal set of includes that the user absolutely needs. 

Also, forward declaration is pretty much par for the course in traditional c++ libraries. Is there a reason the interface of the struct can’t be declared in common.h and the implementation defined in a separate source file? We really should be doing that everywhere, but I’ve been more lenient on that thus far since it doesn’t really impact compile time all that much. The more symbols we compile into the cuvs binary; the less users need to compile into their apps (so overall compile times are reduced) .

sorry if this seems nitpicky but cuVS is now in a different category of software than both cuml and raft- we are heavily focused on the ease of integration and we can do that now since it’s no longer header only.

cpp/include/cuvs/neighbors/common.hpp
```diff
@@ -615,5 +616,56 @@ enable_if_valid_list_t<ListT> deserialize_list(const raft::resources& handle,
...
+template <typename AnnIndexType, typename T, typename IdxT, typename Accessor1, typename Accessor2>
+void extend(
+  const raft::device_resources& handle,
+  cuvs::neighbors::iface<AnnIndexType, T, IdxT>& interface,
+  raft::mdspan<const T, matrix_extent<int64_t>, row_major, Accessor1> new_vectors,
+  std::optional<raft::mdspan<const IdxT, vector_extent<int64_t>, layout_c_contiguous, Accessor2>>
+    new_indices);
+
+template <typename AnnIndexType, typename T, typename IdxT>
+void search(const raft::device_resources& handle,
```
tfeher, line 649: Do we need this function as part of the public API? Isn't this just used internally in iface.hpp?

Here we have a search API that takes `queries` as host_matrix, and returns `neighbors` as device matrix. I see that this interface is used internally in the implementation of the MG algorithm. In contrast the public API in mg.hpp defines methods that takes host `queries` and return host `neighbors`. I might be missing something, but I do not see the role of this in the public API.

cpp/bench/ann/src/cuvs/cuvs_mg_ivf_pq_wrapper.h
```diff
@@ -0,0 +1,139 @@
...
+#include "cuvs_ann_bench_utils.h"
+#include "cuvs_ivf_pq_wrapper.h"
+#include <cuvs/neighbors/mg.hpp>
+#include <raft/core/resource/nccl_clique.hpp>
+
+namespace cuvs::bench {
+using namespace cuvs::neighbors;
+
+template <typename T, typename IdxT>
+class cuvs_mg_ivf_pq : public algo<T>, public algo_gpu {
```
tfeher, line 28: Can't we have a generic mg wrapper for all mg algos?

cpp/bench/ann/src/cuvs/cuvs_benchmark.cu
```diff
@@ -113,6 +174,29 @@ auto create_search_param(const std::string& algo_name, const nlohmann::json& con
...
+  if (algo_name == "raft_mg_ivf_pq" || algo_name == "cuvs_mg_ivf_pq") {
+    auto param = std::make_unique<typename cuvs::bench::cuvs_mg_ivf_pq<T, int64_t>::search_param>();
+    parse_search_param<T, int64_t>(conf, *param);
+    add_merge_mode(&param->merge_mode, conf);
+    return param;
+  }
+
+  if (algo_name == "raft_mg_cagra" || algo_name == "cuvs_mg_cagra") {
+    auto param = std::make_unique<typename cuvs::bench::cuvs_mg_cagra<T, uint32_t>::search_param>();
+    parse_search_param<T, uint32_t>(conf, *param);
```
tfeher, line 195: Here (and above in the mg cases) can't we just parse the single GPU param struct, and construct the parallel param struct from that?

cpp/bench/ann/src/cuvs/cuvs_benchmark.cu
```diff
@@ -71,6 +108,30 @@ auto create_algo(const std::string& algo_name,
...
+  if (algo_name == "raft_mg_ivf_pq" || algo_name == "cuvs_mg_ivf_pq") {
+    typename cuvs::bench::cuvs_mg_ivf_pq<T, int64_t>::build_param param;
+    parse_build_param<T, int64_t>(conf, param);
+    add_distribution_mode(&param.mode, conf);
+    a = std::make_unique<cuvs::bench::cuvs_mg_ivf_pq<T, int64_t>>(metric, dim, param);
+  }
+
+  if (algo_name == "raft_mg_cagra" || algo_name == "cuvs_mg_cagra") {
+    typename cuvs::bench::cuvs_mg_cagra<T, uint32_t>::build_param param;
+    parse_build_param<T, uint32_t>(conf, param);
```
tfeher, line 130: Here (and above in the mg cases) can't we just parse the single GPU param struct, and construct the parallel param struct from that?

tfeher (changes requested):
Thanks Victor, this looks great. I have a few additional comment.

For benchmarks it would be great if we can reduce code duplication. But I am fine with pushing that to next release.

cpp/bench/ann/src/cuvs/cuvs_ann_bench_param_parser.h
```diff
@@ -176,7 +188,11 @@ nlohmann::json collect_conf_with_prefix(const nlohmann::json& conf,
 
 template <typename T, typename IdxT>
 void parse_build_param(const nlohmann::json& conf,
+#if defined(CUVS_ANN_BENCH_USE_CUVS_MG)
+                       typename cuvs::bench::cuvs_mg_cagra<T, IdxT>::build_param& param)
```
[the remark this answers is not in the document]
viclafargue: This function parses the single GPU parameters and puts them either in a single GPU or multiple GPU bench struct. This function should indeed work for both cases. [The MG benchmarks CUDA files are parameterized with `CUVS_ANN_BENCH_USE_CUVS_MG`](https://github.com/rapidsai/cuvs/blob/bcf97c9e8f35f8d8e7830ae8ac8f3cff99532d56/cpp/bench/ann/CMakeLists.txt#L259-L261).

cpp/bench/ann/src/cuvs/cuvs_benchmark.cu
viclafargue: The MG bench struct inherits from the SG bench struct. Therefore, in the MG case the full MG struct should be allocated which is not necessary for the SG case. The main parameters (the SG ones) are filled with the `parse_search_param` function, and the `add_merge_mode` adds the specifically MG parameter. In my opinion, it is relatively clean, but I might be missing some bits to improve the code even further though.

build.sh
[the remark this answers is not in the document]
cjnolet: I think maybe we should consider not having this on by default, but we should make sure we add `--mg` (I propose we rename it to be simpler) to all the conda recipes. Please also add this to the docs both in the "help" for this file and in the build and install docs (docs/source/build.md).

cpp/bench/ann/CMakeLists.txt
```diff
@@ -32,6 +32,7 @@ option(CUVS_ANN_BENCH_USE_CUVS_BRUTE_FORCE "Include cuVS brute force knn in benc
 option(CUVS_ANN_BENCH_USE_CUVS_CAGRA_HNSWLIB "Include cuVS CAGRA with HNSW search in benchmark" ON)
 option(CUVS_ANN_BENCH_USE_HNSWLIB "Include hnsw algorithm in benchmark" ON)
 option(CUVS_ANN_BENCH_USE_GGNN "Include ggnn algorithm in benchmark" OFF)
+option(CUVS_ANN_BENCH_USE_CUVS_MG "Include cuVS ann mg algorithm in benchmark" ON)
```
cjnolet, line 35: What's going to happen if we build with this argument ON but BUILD_MG_ALGOS is OFF? I think we should just default this to whatever $BUILD_MG_ALGOS is set to. That makes it easier and puts the control in a single place.

cpp/bench/ann/src/cuvs/cuvs_mg_ivf_pq_wrapper.h
viclafargue: I considered doing this, but it looks like most commonalities are not MG-specific, but rather ANN-specific.

cpp/bench/ann/src/cuvs/cuvs_ann_bench_param_parser.h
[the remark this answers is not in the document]
tfeher: Thanks for the clarification! Now I see why this can work.

I still think this is an unfortunate design. This is added complexity that is not necessary. The function here reads only `cuvs_cagra<T, IdxT::build_param` and not `cuvs_mg_cagra<T, IdxT>::build_param`. We could just leave it as is with `cuvs_cagra<T, IdxT::build_param`.

cpp/bench/ann/src/cuvs/cuvs_mg_ivf_pq_wrapper.h
tfeher: Let's think about in the next release how to make this less verbose.

cpp/bench/ann/src/cuvs/cuvs_benchmark.cu
tfeher: I agree. Except to fill the main parameters, I would not have a specific overload of `parse_search_param` for the mg case (see my comment [here](https://github.com/rapidsai/cuvs/pull/231/files#r1777733285)). Why does the cagra param reader have separate variants for the SG and MG bench case, while other methods do not?

cpp/bench/ann/CMakeLists.txt
tfeher: Agree, we get compilation error if the flags do not align.

cpp/include/cuvs/neighbors/common.hpp
tfeher: @cjnolet please confirm that this is the case, and that we indeed want a search interface with _host_ input queries and _device_ output.

tfeher (changes requested):
Thanks Viktor for the updates, we are almost there. Issues regarding ann_bench can be pushed to a follow up.

cpp/src/neighbors/ann_mg/ann_mg_cagra_int8_t_uint32_t.cu
viclafargue: The PR adds 7M, from 130M to 137M.

cpp/include/cuvs/neighbors/ann_mg.hpp
```diff
@@ -0,0 +1,1417 @@
...
+
+namespace cuvs::neighbors::mg {
+using pool_mr = rmm::mr::pool_memory_resource<rmm::mr::device_memory_resource>;
+
+/// \defgroup ann_mg_cpp_nccl_clique NCCL clique utility
+
+/// \ingroup ann_mg_cpp_nccl_clique
+struct nccl_clique {
+  /**
+   * Instantiates a NCCL clique with all available GPUs
```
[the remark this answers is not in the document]
viclafargue: I could test the benchmarks by selecting different GPUs, and everything worked fine.

cpp/include/cuvs/neighbors/common.hpp
viclafargue: As discussed offline with @tfeher, the search function used in the mg namespace will not be exposed in the public `iface` declaration. Instead a search function with inputs and outputs on device will take place.

cpp/bench/ann/src/cuvs/cuvs_benchmark.cu
viclafargue: I wrote this code quite some time ago. Actually, for CAGRA bench, the build parameters for the MG bench did not inherit from the SG bench. I could fix this with a number of changes.

cpp/bench/ann/src/cuvs/cuvs_ann_bench_param_parser.h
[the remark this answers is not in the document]
viclafargue: Could fix this with my latest commit.

tfeher (approved):
Thanks Victor for the updates, the PR looks good to me!

One thing to note: we are actually adding two features here: one is SNMG, other is the general `iface` to use ANN. In the next release we need to continue working on the interface to make it easy to use.

jameslamb (approved):
Giving you a `ci-codeowners` / `packaging-codeowners` approval, now that this is getting so close to code freeze. The packaging changes here (CMake and `build*.sh`) look fine to me. I limited my review only to those.

cpp/CMakeLists.txt
```diff
@@ -57,6 +57,7 @@ option(BUILD_C_LIBRARY "Build cuVS C API library" OFF)
 option(BUILD_C_TESTS "Build cuVS C API tests" OFF)
 option(BUILD_CUVS_BENCH "Build cuVS ann benchmarks" OFF)
 option(BUILD_CAGRA_HNSWLIB "Build CAGRA+hnswlib interface" ON)
+option(BUILD_MG_ALGOS "Build with multi-GPU support" ON)
```
jameslamb, line 60: Summarizing our offline conversation here (somewhere threaded, that could be marked `Resolved` and collapsed).

I just pushed https://github.com/rapidsai/cuvs/pull/231/commits/55fbb3646fb2e85cf0e01631ebd15ea1efeab536, which does the following:

* makes multi-GPU support on by default for wheel and conda builds
* adds a `build.sh` option, `--no-mg`, to run a build with it not enabled
* adds a CMake option here, to make the default setting here in CMake explicit (so that, for example, wheels will be built with this support enabled by default)

cjnolet (approved):

conda/recipes/libcuvs/meta.yaml
```diff
@@ -65,6 +65,7 @@ outputs:
       host:
         - librmm ={{ minor_version }}
         - libraft-headers ={{ minor_version }}
+        - nccl {{ nccl_version }}
```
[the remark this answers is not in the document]
jakirkham: Makes sense. Thanks James! 🙏

To reflect this, updated this comment: https://github.com/rapidsai/build-planning/issues/102#issuecomment-2377616789

Please feel free to edit that further

dantegd:
CI error here seemed to have been a connection error: 

```
[rapids-conda-retry] Exiting, no retryable mamba errors detected: 'ChecksumMismatchError:', 'ChunkedEncodingError:', 'CondaHTTPError:', 'CondaMultiError:', 'Connection broken:', 'ConnectionError:', 'DependencyNeedsBuildingError:', 'EOFError:', 'JSONDecodeError:', 'Multi-download failed', 'Timeout was reached', segfault exit code 139
```

conda/recipes/libcuvs/meta.yaml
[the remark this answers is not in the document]
vyasr: Thanks for the heads-up James.

cpp/src/neighbors/iface/iface.hpp
```diff
@@ -0,0 +1,198 @@
...
+
+using namespace raft;
+
+template <typename AnnIndexType, typename T, typename IdxT, typename Accessor>
+void build(const raft::device_resources& handle,
+           cuvs::neighbors::iface<AnnIndexType, T, IdxT>& interface,
+           const cuvs::neighbors::index_params* index_params,
+           raft::mdspan<const T, matrix_extent<int64_t>, row_major, Accessor> index_dataset)
+{
+  interface.mutex_->lock();
```
rhdong, line 19: Hi @viclafargue @achirkin, I'm studying the code for the multi-GPU index feature—great work! I noticed a potential issue: the lock system does not prevent data races on device memory since the API does not explicitly synchronize the stream (at least, I found this to be the case with deserialize). If this behavior is intentional, please feel free to disregard my comments.

viclafargue: Hi @rhdong, thank you for noticing this! It is indeed probably safer to synchronize before unlocking to unsure that all `cudaMemcpy` calls made during serialization have completed. Will update this.

viclafargue: Thanks again! Did an edit to fix this issue : https://github.com/rapidsai/cuvs/pull/454/commits/4e0f512a31b5abef3267f35b7c60d5291924ac38

PR #235, opened 2024-07-19
Fix pinning to a different RAFT tag

benfred:
Attempting to pin the version of a raft to a custom fork wasn't working, and it was still using the version installed by conda. Fix by mirroing the `CUML_RAFT_CLONE_ON_PIN` logic found in the cuml cmake files.

### DIFF
--- a/python/cuvs/CMakeLists.txt
+++ b/python/cuvs/CMakeLists.txt
@@ -37,7 +37,6 @@ project(
 option(FIND_CUVS_CPP "Search for existing CUVS C++ installations before defaulting to local files"
        OFF
 )
-set(CUVS_RAFT_CLONE_ON_PIN OFF)
 
 message(
   "CUVS_PY: Searching for existing cuVS C/C++ installations before defaulting to local files: ${FIND_CUVS_CPP}"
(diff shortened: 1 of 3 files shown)


python/cuvs/CMakeLists.txt
```diff
@@ -37,6 +37,7 @@ project(
 option(FIND_CUVS_CPP "Search for existing CUVS C++ installations before defaulting to local files"
        OFF
 )
+set(CUVS_RAFT_CLONE_ON_PIN OFF)
```
divyegala, line 40: Why does this need to be set off?

benfred: I copied the logic here from cuml https://github.com/rapidsai/cuml/blob/99b20f67c6dea4aba80da5787f64557dba30a1f8/python/cuml/CMakeLists.txt#L41 - but now I'm wondering if this might have been accidentally applied in cuml in https://github.com/rapidsai/cuml/pull/4990 , and not meant to be included there permanently.

Removed in latest commit

divyegala: Thanks for tracking that down, definitely looks to be accidental.

divyegala (approved):

PR #237, opened 2024-07-22
Add MST optimization to guarantee the connectivity of CAGRA graphs

anaruse:
This PR allows us to guarantee the connectivity of the CAGRA search graph using approximate MST.

It has been empirically shown that the graph indexes generated by CAGRA for search provide comparable search accuracy to other libraries, but reachability from any node to all nodes is not guaranteed. In fact, it has been confirmed that the number of strongly connected components (SCC) of graph indexes created by CAGRA is not 1 in some 100M scale datasets.

This problem can be alleviated by increasing the number of degrees in the search graph, but this would increase the size of the graph index. It is desirable to address this problem without increasing the number of degrees of the search graph.

Prior study has shown that this can be solved by using a Minimum Spanning Tree (MST)-like approach, but in general, MST calculation takes a long time. However, what is needed here is not an exact MST, but, for example, an approximate MST in which the total number of edges is not necessarily minimum. Such an approximate MST could be computed quickly on GPUs.

This PR contains implementation to create a approximate MST on the GPU at high speed based on the above policy and use it to guarantee the connectivity of the search graph.

This functionality is not always required, so it is considered an opt-in feature. A member variable named `guarantee_connectivity` is added to `index_params`, so set this variable to `true` if you wish to use this featgure.

> cuvs::neighbors::cagra::index_params index_params;
> index_params.guarantee_connectivity = true;
> auto index = cuvs::neighbors::cagra::build(res, index_params, dataset_view);


cpp/src/neighbors/detail/cagra/graph_core.cuh
```diff
@@ -324,14 +557,516 @@ void sort_knn_graph(
...
+
+    // 4. Calculate the number of clusters and the size of each cluster
+    num_clusters = 0;
+    if (use_gpu) {
+      stats_ptr[0] = num_clusters;
+      raft::copy(d_stats_ptr, stats_ptr, 1, raft::resource::get_cuda_stream(res));
+
+      constexpr uint64_t n_threads = 256;
+      const dim3 threads(n_threads, 1, 1);
+      const dim3 blocks((graph_size + n_threads - 1) / n_threads, 1, 1);
```
tfeher, line 919:
```suggestion
      const dim3 blocks(raft::ceildiv<uint64_t>(graph_size, n_threads), 1, 1);
```

tfeher (changes requested):
Thanks Akira for porting the PR to cuVS! The original PR https://github.com/rapidsai/raft/pull/2218 was already reviewed, and I see that most of the issues are addressed here. I have a marked a few smaller issues below.

There is some scope for improving code readability by relying more on mdspan, and adding a helper function to check duplicates (e.g. [here](https://github.com/rapidsai/raft/pull/2218/files#r1567965346)). @anaruse I could easily apply the mdspan changes on the PR, if you are fine with that.

anaruse:
Thanks for the review, Tamas. I have fixed the points you pointed out, and it would be helpful if you could make the mdspan related changes.

tfeher (approved):
Thanks Akira for the update. The PR looks good to me.

cjnolet:
Before this is merged can we please have a GitHub issue created to expose this through the public APIs? We already have a  use-case for this API in HDBSCAN and single linkage clustering.

cjnolet:
Also, given that both of you have made changes to the PR, I’d like to have a 3rd person do a quick look over. I’ll take a look tomorrow.

cjnolet:
Just noticed this issue https://github.com/rapidsai/cuvs/issues/256 was created, so disregard my first comment.

cjnolet (approved):
LGTM. It would be nice to have a little more detail in the description of the `guarantee_connectivity` param, but i don't think that needs to hold up this PR.

cjnolet:
@anaruse @tfeher it looks like a small failure in the style checker

PR #240, opened 2024-07-22
C API for CAGRA+HNSW

divyegala:


cjnolet (changes requested):
The changes look good to me, however I think we need to make clear in the docs that at the present time, only the base graph is used in HNSW and so the index is immutable.

Also, please add this and the other C APIs to the docs.

cpp/include/cuvs/neighbors/cagra.h
```diff
@@ -416,6 +419,32 @@ cuvsError_t cuvsCagraSerialize(cuvsResources_t res,
                                cuvsCagraIndex_t index,
                                bool include_dataset);
 
+/**
+ * Save the CAGRA index to file in hnswlib format.
+ *
+ * Experimental, both the API and the serialization format are subject to change.
```
cjnolet, line 425: Please describe the assumptions wrt cuVS and hnswlib so that it's blatantly clear for users. For example, this requires the user to use cuVS as a proxy for hnswlib because we've patched hnswlib to make this possible, even though we're writing to the hnswlib format.

cpp/include/cuvs/neighbors/hnsw.h
```diff
@@ -0,0 +1,204 @@
...
+ */
+
+/**
+ * @defgroup hnsw_c_serialize HNSW C-API serialize functions
+ * @{
+ */
+
+/**
+ * Load hnswlib index from file which was serialized from a HNSW index.
+ * NOTE: The loaded hnswlib index is immutable.
```
cjnolet, line 164: This is a start, but it still needs to clearly describe that this is meant to support a cagra graph that's been serialized to hnswlib format only through cuVS, since it's been patched to support this.

cjnolet (approved):

PR #243, opened 2024-07-23
Add distance epilogue functionality for bruteforce knn

benfred:
The mutual reachability code inside of cuml needs to be able to specify a distance epilogue, to modify the distances after the distance computation but before the select-k step.  This change adds an optional function to do this.


cpp/src/neighbors/brute_force.cu
```diff
@@ -145,54 +145,59 @@ void index<T>::update_dataset(raft::resources const& res,
...
+  void search(raft::resources const& res,                                                    \
+              const cuvs::neighbors::brute_force::index<T>& idx,                             \
+              raft::device_matrix_view<const T, int64_t, raft::row_major> queries,           \
+              raft::device_matrix_view<int64_t, int64_t, raft::row_major> neighbors,         \
+              raft::device_matrix_view<T, int64_t, raft::row_major> distances,               \
+              std::optional<cuvs::core::bitmap_view<const uint32_t, int64_t>> sample_filter, \
+              std::optional<DistanceEpilogue<T>> distance_epilogue)                          \
+  {                                                                                          \
+    if (sample_filter.has_value() && distance_epilogue.has_value()) {                        \
+      RAFT_FAIL("can't specify both a sample_filter and a distance_epilogue");               \
```
lowener, line 175: Is there a plan to support both in the future?

cpp/test/neighbors/brute_force.cu
```diff
@@ -335,6 +348,36 @@ class RandomBruteForceKNNTest : public ::testing::TestWithParam<RandomKNNInputs>
                                                     float(0.001),
                                                     stream_,
                                                     true));
+
+    if (params_.row_major) {
+      // test out distance epilogue functionality, by squaring distances
```
lowener, line 353:
```suggestion
      // test out distance epilogue functionality, by adding one to distances
```

cjnolet:
Discussed offline but I want to mention it publicly- we are discussing moving over the mutual reachabiility computation from cuML and making it a first-class distance within bfknn (and nn-descent as that will be needed also) so that we can avoid having to expose any type of distance epilogue.

benfred:
mutual reachability was directly added in #323 -

PR #244, opened 2024-07-23
Add cuvs_bench python folder, config files and constraints

dantegd:


python/cuvs_bench/cuvs_bench/config/datasets.yaml
```diff
@@ -0,0 +1,127 @@
...
+  query_file: nytimes-256-angular/query.fbin
+  groundtruth_neighbors_file: nytimes-256-angular/groundtruth.neighbors.ibin
+  distance: euclidean
+
+- name: nytimes-256-inner
+  dims: 256
+  base_file: nytimes-256-inner/base.fbin
+  query_file: nytimes-256-inner/query.fbin
+  groundtruth_neighbors_file: nytimes-256-inner/groundtruth.neighbors.ibin
+  distance: euclidean
```
cjnolet, line 99: Probably my fault, but there's several places in the file that say Euclidean when they should say inner.

python/cuvs_bench/cuvs_bench/config/deep-image-96-inner.yaml
```diff
@@ -0,0 +1,712 @@
...
+- algo: raft_ivf_pq
+  build_param:
+    niter: 25
+    nlist: 1024
+    pq_dim: 64
+    ratio: 1
+  file: index/deep-image-96-inner/raft_ivf_pq/dimpq64-cluster1024-float-half
+  name: raft_ivf_pq.dimpq64-cluster1024-float-half
+  search_params:
+  - internalDistanceDtype: float
```
cjnolet, line 523: Does this work like? Is the intention here to group these 3 args together?

python/cuvs_bench/cuvs_bench/config/deep-1B.yaml
```diff
@@ -0,0 +1,26 @@
+dataset:
```
cjnolet, line 1: I don't think we want to couple the param sweeps dirdftly to the datasets. The whole reason for separating them is that the sweeps should generalize across all datasets (and because it's more manageable to specify the sweeps once, instead of per dataset)

cjnolet (approved):
My comments are somewhat minore, but relevant. I think we went to keep the params separate from the datasets. I'm also okay removing the files that tried to combine everything if they are only being left around for reference.

cjnolet (approved):

cjnolet (approved):

PR #246, opened 2024-07-24
Python API for CAGRA+HNSW

divyegala:


jameslamb:
@divyegala I observed that same pip devcontainer build failure on an unrelated PR: https://github.com/rapidsai/cuvs/pull/247#issuecomment-2248744164

So I suspect it's unrelated to your changes here.

divyegala:
@jameslamb I suspect it's related to this PR https://github.com/rapidsai/raft/pull/2346

What I cannot figure out is why conda-cpp builds pass (new conda packages released?) but wheel builds fail. cuVS wheel builds pick up libraft wheel builds?

vyasr:
Maybe conda packages weren't out yet when the builds failed? Wheel builds will pull the source of raft to rebuild the static lib (unless you changed something in how cuvs wheels specifically are built, but I don't see any indication of that in the CMake).

cpp/include/cuvs/neighbors/cagra.h
```diff
@@ -416,6 +428,32 @@ cuvsError_t cuvsCagraSerialize(cuvsResources_t res,
                                cuvsCagraIndex_t index,
                                bool include_dataset);
 
+/**
+ * Save the CAGRA index to file in hnswlib format.
```
cjnolet, line 432: Should we add a note here that this must be loaded by the (patched) hnswlib wrappers inside cuVS and it can't just be loaded with hnswlib?

python/cuvs/cuvs/neighbors/hnsw/hnsw.pyx
```diff
@@ -0,0 +1,364 @@
...
+                                                 index.index))
+
+
+@auto_sync_resources
+def load(filename, dim, dtype, metric="sqeuclidean", resources=None):
+    """
+    Loads base-layer-only hnswlib index from file, which was originally
+    saved as a built CAGRA index.
+
+    Saving / loading the index is experimental. The serialization format is
```
cjnolet, line 147: I know you are alluding to it in this description but we should also be very explicit and mention that the serialized index can only be loaded with the functions in cuVS because the hnswlib version underneath is patched to be able to search the converted CAGRA graph. 


We recently had 2 different partners try to do this conversion themselves and they didn't realixe they couldn't just load this into hnswlib proper.

cjnolet (commented):
Looks good overall. Mostly a couple small documentation-related things to polish this off and then I think it's ready to merge.

cjnolet (approved):

python/cuvs/cuvs/neighbors/hnsw/hnsw.pyx
```diff
@@ -0,0 +1,371 @@
...
+    """
+    HNSW search parameters
+
+    Parameters
+    ----------
+    ef: int, default = 200
+        Maximum number of candidate list size used during search.
+    num_threads: int, default = 0
+        Number of CPU threads used to increase search parallelism.
+        When set to 0, the number of threads is automatically determined.
```
cjnolet, line 52: Can you briefly specify how the number of threads is determined? Does it use the number hyper hreads avaialble in the system? Number of physical cores?

python/cuvs/cuvs/neighbors/hnsw/hnsw.pyx
```diff
@@ -0,0 +1,371 @@
...
+
+@auto_sync_resources
+def save(filename, cagra.Index index, resources=None):
+    """
+    Saves the CAGRA index to a file as an hnswlib index.
+    The saved index is immutable and can only be searched by the hnswlib
+    wrapper in cuVS, as the format is not compatible with the original
+    hnswlib.
+
+    Saving / loading the index is experimental. The serialization format is
```
cjnolet, line 113: Please include the note here about how even though this format closely matches hnswlib, it is only intended to be used by cuVS hnsw apis.

python/cuvs/cuvs/neighbors/hnsw/hnsw.pyx
```diff
@@ -0,0 +1,371 @@
...
+    Parameters
+    ----------
+    filename : string
+        Name of the file.
+    dim : int
+        Dimensions of the training dataest
+    dtype : np.dtype of the saved index
+        Valid values for dtype: [np.float32, np.byte, np.ubyte]
+    metric : string denoting the metric type, default="sqeuclidean"
+        Valid values for metric: ["sqeuclidean"], where
```
cjnolet, line 165: Should the valid values example include inner product as well? 

```suggestion
        Valid values for metric: ["sqeuclidean"], where
```

python/cuvs/cuvs/neighbors/hnsw/hnsw.pyx
```diff
@@ -0,0 +1,371 @@
...
+            - sqeuclidean is the euclidean distance without the square root
+              operation, i.e.: distance(a,b) = \\sum_i (a_i - b_i)^2,
+            - inner product distance is defined as
```
cjnolet, line 168: Perhaps this should be worded so that the actual required string is in the sentence like the sqeuclidean? 
```suggestion
            - inner_product distance is defined as
```

python/cuvs/cuvs/neighbors/hnsw/hnsw.pyx
divyegala: That note is already present right above
```
    The saved index is immutable and can only be searched by the hnswlib
    wrapper in cuVS, as the format is not compatible with the original
    hnswlib.
```

python/cuvs/cuvs/neighbors/hnsw/hnsw.pyx
divyegala: Right above, the note is quite clear:
```
The loaded index is immutable and can only
    be searched by the hnswlib wrapper in cuVS, as the format is not
    compatible with the original hnswlib.
```

> can only be searched by the hnswlib wrapper in cuVS, as the format is not compatible with the original hnswlib.

python/cuvs/cuvs/neighbors/hnsw/hnsw.pyx
```diff
@@ -0,0 +1,371 @@
...
+        idx.index
+    ))
+    idx.trained = True
+    return idx
+
+
+@auto_sync_resources
+def from_cagra(cagra.Index index, resources=None):
+    """
+    Returns an hnswlib base-layer-only index from a CAGRA index.
```
cjnolet, line 228: Maybe since we are wrapping hnsw into its own APIs here, we could just say `Returns an hnsw ...` to make it a little easier. Most users don't need to know or care that it's hnswlib. 

The note below is great! Can you add another couple words to this description just to define what `base-layer-only` means and how it impacts the user?

cpp/include/cuvs/neighbors/hnsw.hpp
```diff
@@ -303,7 +303,7 @@ void search(raft::resources const& res,
 void search(raft::resources const& res,
             const search_params& params,
             const index<int8_t>& idx,
-            raft::host_matrix_view<const int, int64_t, raft::row_major> queries,
+            raft::host_matrix_view<const int8_t, int64_t, raft::row_major> queries,
             raft::host_matrix_view<uint64_t, int64_t, raft::row_major> neighbors,
             raft::host_matrix_view<float, int64_t, raft::row_major> distances);
```
cjnolet, line 308: We are ultimatley going to need to support filtered search here. Can you create an issue for this and reference it in a todo in this file? Beginning of the file will be fine or just before all the search functions.

python/cuvs/cuvs/neighbors/hnsw/hnsw.pyx
```diff
@@ -0,0 +1,375 @@
...
+    return idx
+
+
+@auto_sync_resources
+def from_cagra(cagra.Index index, resources=None):
+    """
+    Returns an hnsw base-layer-only index from a CAGRA index.
+
+    NOTE: This method uses the filesystem to write the CAGRA index in
+          `/tmp/<random_number>.bin` before reading it as an hnsw index,
```
cjnolet, line 232: Can we please make this path configurable? It's not safe to assume we always have access to this directory (not that this directory is always where the user wants to store this intermediate file). We should provide an argument to control this path but allow the user to override this path if they need/want to.

python/cuvs/cuvs/neighbors/hnsw/hnsw.pyx
```diff
@@ -0,0 +1,375 @@
...
+@auto_sync_resources
+@auto_convert_output
+def search(SearchParams search_params,
+           Index index,
+           queries,
+           k,
+           neighbors=None,
+           distances=None,
+           resources=None):
+    """
```
cjnolet, line 281: Will need to add HNSW to the docs for C, C++ and Python versions. We want to make sure this info is available for users (and that we can point to it in our getting started materials).

cjnolet (changes requested):
The implementation is shaping up great. Mostly minor things, but important nonetheless.

cjnolet (approved):

Issue #160, opened 2024-05-28
[FEA] ANN bench improvements


tfeher:
During porting the ANN benchmarks from RAFT (#130), a list of potential improvements was created: 

- [ ] clang-tidy: cleanup naming conventions
- [ ] cleanup/unify logging (including CPU-only builds)
- [ ] fix/unify create_algo + create_search_param syntax
- [ ] actualize cuda_stub.hpp for cpu-only builds
- [ ] make sure there's no direct cuda dependencies in benchmark.hpp
- [ ] add the memory interconnect info to benchmark context
- [ ] add command line string which used for running the benchmarking to the benchmark context
- [ ] add more information about CPU cores and RAM to the benchmark context
- [ ] etc...

tfeher:
Suggestions for improved naming
- [ ] benchmark name (in build.sh) https://github.com/rapidsai/cuvs/pull/130#discussion_r1619512483
- [ ] name of index class (in ann_type.hpp) https://github.com/rapidsai/cuvs/pull/130#discussion_r1619515306

PR #248, opened 2024-07-24
Add more info to ANN_BENCH context

achirkin:
Add extra information to benchmark context for better reproducibility and performance analysis:

  1. Full command line used to call the executable (so you can copy-paste and run again).
  2. More CUDA device information: whether HMM, AST, or host atomics are available (how GPU can efficiently communicate with CPU).
  3. Host information: min/max frequences, used virtual processors and cores, available physical memory and swap (does the benchmark segfault due to not enough host memory? is SMT enabled? etc).

Addresses parts of https://github.com/rapidsai/cuvs/issues/160


tfeher (approved):
Thanks Artem, LGTM, it is useful to extend the context information. I do not see any issues with adding the hostname.

tfeher:
For env vars, I am not sure what would be relevant.
For OpenMP context, you could add [omp_get_max_threads()](https://www.openmp.org/spec-html/5.0/openmpsu112.html)

achirkin:
We could get the openmp-related env vars to see if the limit to the number of threads is set explicitly.
I wouldn't like to add `omp_get_max_threads()` and other OpenMP functions to avoid introducing openmp dependency in the benchmark (for the algorithms that do not use OpenMP and for the ANN_BENCH executable in the single-exe mode). I'm also not sure if cuVS team decides to move away from using OpenMP in favor of standard C++ threads in future.

PR #260, opened 2024-07-29
Diskann Benchmarking Wrapper

tarang-jain:
Brings DiskANN into cuvs-bench
- [x] Build and search in-memory DiskANN index
- [x] Build and search SSD DiskANN index
- [x] Build a cuvs Vamana index on GPU and serialize it in DiskANN format. Search on CPU using in-memory DiskANN search API.


achirkin (changes requested):
Thanks for the PR, it's nice to see the list of supported algorithms growing.

I feel unsettled though about adding new functionality to the common `benchmark.hpp` header just for one algorithm - this looks unsustainable. I see you probably want to avoid having the user to specify the path to the index file / dataset manually, but I think adding the new functionality in such an ad-hoc way is not a solution.
I think, the easiest solution could be just to pass both paths as algo parameters.
Maybe you can automate this in the python layer of the benchmark. Alternatively, I think it's acceptable to modify the `parse_build_param` api to give it access to the full json config file; that way, you could get the path to the dataset and fill-in the algo parameters inside that function.

cpp/bench/ann/src/cuvs/cuvs_vamana_wrapper.h
```diff
@@ -0,0 +1,123 @@
...
+{
+  diskann_memory_search_->load(file);
+}
+
+template <typename T, typename IdxT>
+void cuvs_vamana<T, IdxT>::search(
+  const T* queries, int batch_size, int k, algo_base::index_type* neighbors, float* distances) const
+{
+  diskann_memory_search_->search(queries, batch_size, k, neighbors, distances);
+}
```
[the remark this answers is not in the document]
tarang-jain: The GPU build requires the stream, whereas the search is done on CPU. So we would still need the stream for the build part. So we would still have to set `uses_stream()` to true for the whole algorithm (which is the default)? I believe we are currently doing it the same way in CAGRA + HNSWLIB.

conda/recipes/cuvs_bench_cpu/meta.yaml
```diff
@@ -47,6 +47,9 @@ requirements:
     - benchmark
     - fmt {{ fmt_version }}
     - glog {{ glog_version }}
+    - libaio
+    - libboost-devel
+    - mkl-devel
```
bdice, line 52:
```suggestion
    - mkl-devel  # [linux64]
```

cpp/bench/ann/src/cuvs/cuvs_vamana_wrapper.h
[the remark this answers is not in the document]
achirkin: Good point! Then you can set it by default to `true` and override this value in `set_search_param` function.

cpp/bench/ann/src/cuvs/cuvs_cagra_diskann_wrapper.h
```diff
@@ -0,0 +1,108 @@
...
+}
+
+template <typename T, typename IdxT>
+void cuvs_cagra_hnswlib<T, IdxT>::set_search_param(const search_param_base& param_)
+{
+  hnswlib_search_.set_search_param(param_);
+}
+
+template <typename T, typename IdxT>
+void cuvs_cagra_hnswlib<T, IdxT>::save(const std::string& file) const
```
bkarsin, line 89: Why is HNSW converter part of the diskann search benchmark?

cpp/bench/ann/src/diskann/diskann_wrapper.h
```diff
@@ -0,0 +1,335 @@
...
+    " " + std::string(std::to_string(false)) + " " + std::string(std::to_string(0)) + " " +
+    std::string(std::to_string(param.QD));
+  base_file_         = param.dataset_file;
+  index_path_prefix_ = param.path_to_index;
+}
+
+template <typename T>
+void diskann_ssd<T>::build(const T* dataset, size_t nrow)
+{
+  diskann::build_disk_index<float>(base_file_.c_str(),
```
bkarsin, line 273: Where is this defined? Is it using the CPU DiskANN repo to build the ssd verson of the index?

cpp/bench/ann/CMakeLists.txt
```diff
@@ -32,6 +32,12 @@ option(CUVS_ANN_BENCH_USE_CUVS_BRUTE_FORCE "Include cuVS brute force knn in benc
 option(CUVS_ANN_BENCH_USE_CUVS_CAGRA_HNSWLIB "Include cuVS CAGRA with HNSW search in benchmark" ON)
 option(CUVS_ANN_BENCH_USE_HNSWLIB "Include hnsw algorithm in benchmark" ON)
 option(CUVS_ANN_BENCH_USE_GGNN "Include ggnn algorithm in benchmark" OFF)
+option(CUVS_ANN_BENCH_USE_DISKANN "Include DISKANN search in benchmark" ON)
+option(CUVS_ANN_BENCH_USE_CUVS_VAMANA "Include cuVS Vamana with DiskANN search in benchmark" ON)
+if(CMAKE_SYSTEM_PROCESSOR MATCHES "(ARM|arm|aarch64)")
```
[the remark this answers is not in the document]
tarang-jain: Yes, they have `mkl-devel` as a dependency, which is not meant to be installed in aarch64.

cpp/bench/ann/src/cuvs/cuvs_cagra_diskann_wrapper.h
tarang-jain: Removed it!

cpp/bench/ann/src/diskann/diskann_wrapper.h
tarang-jain: Yes, that is right. It is just to benchmark the CPU build

cjnolet (approved):

bkarsin (approved):

cpp/bench/ann/src/cuvs/cuvs_vamana_wrapper.h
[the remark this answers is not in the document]
achirkin: It doesn't look like this has been addressed yet, right?
(feel free to re-resolve if I'm wrong though)

achirkin (changes requested):
Please avoid algorithm-specific changes to `cpp/bench/ann/src/common/benchmark.hpp` as per our discussion.
I'm not opposed to exposing dataset/index filepaths to algorithms in general though.
But maybe it would make sense to hide (filter out) them when dumping all build parameters to avoid cluttering the console/reports and exposing private user filepaths?

tarang-jain:
@achirkin are you suggesting that we expose the dataset / index filepaths to all algorithms, but only use them for diskann ssd indexes?

cpp/bench/ann/src/common/benchmark.hpp
```diff
@@ -144,7 +150,8 @@ void bench_build(::benchmark::State& state,
 
   const auto algo_property = parse_algo_property(algo->get_preference(), index.build_param);
 
-  const T* base_set      = dataset->base_set(algo_property.dataset_memory_type);
+  const T* base_set = nullptr;
+  if (index.algo != "diskann_ssd") base_set = dataset->base_set(algo_property.dataset_memory_type);
```
tarang-jain, line 154: @achirkin if we do not have this line, the entire dataset will be read into the `base_set` variable needlessly for DiskANN ssd based indexes. For these indexes, only the path to the dataset needs to be known and the DiskANN index building APIs will read the data from the path. In fact, it is important we do it this way because those SSD indexes are designed to let the data remain on disk if there are extremely large datasets.

achirkin: You can just set the `dataset_memory_type` to `kHostMmap` by default for diskann index. It will lazily map the file content, so it won't read anything until you actually use the pointer.
It will still open the file and read the first two words anyway - independently of this change. That is because the benchmark reads and reports the dataset dimensionality.

cpp/bench/ann/src/common/benchmark.hpp
```diff
@@ -135,6 +135,12 @@ void bench_build(::benchmark::State& state,
     }
   }
 
+  if (index.algo == "diskann_ssd") {
+    make_sure_parent_dir_exists(index.file);
+    index.build_param["dataset_file"]  = dataset->base_filename();
+    index.build_param["path_to_index"] = index.file;
+  }
+
```
achirkin, line 143: I think you shouldn't need `path_to_index` parameters, because the index member `file` is accessible from within the index in your algorithm-specific code.
If you need the dataset file path during index build,  you can either ingest it as a build param in the `conf.hpp`, similar to how we inject some search parameters in [parse_index(...)](https://github.com/rapidsai/cuvs/blob/11ecbb2a263e3f6591d0a2390453ca0356b845c0/cpp/bench/ann/src/common/conf.hpp#L106-L141), or create one more member in the `configuration::index` - all in `conf.hpp`; I think it's ok to have it available to all algorithm at that point.

cpp/bench/ann/src/common/benchmark.hpp
```diff
@@ -223,7 +230,12 @@ void bench_search(::benchmark::State& state,
 
   const T* query_set = nullptr;
 
-  if (!file_exists(index.file)) {
+  std::string filename;
+  if (index.algo != "diskann_ssd")
+    filename = index.file;
+  else
+    filename = index.file + "_disk.index";
```
achirkin, line 237: Same problem, the algorithm-specific code should go to the algorithm headers.

achirkin (changes requested):
Please try to always adhere to these two rules:
 - No algorithm-specific code in the common header files, such as `benchmark.hpp`
 - All configuration-specific code should go into `conf.hpp`, not into `benchmark.hpp`.

cpp/bench/ann/src/common/benchmark.hpp
tarang-jain: Unfortunately, the `cuvs::bench::configuration::index` object's attributes are not directly passed down to the specific algorithm's wrapper class. So we would need the `path_to_index`. The algorithm specific structs use only the parameters that are explicitly passed down from the configuration object.

cpp/bench/ann/src/common/benchmark.hpp
tarang-jain: This is just for checking if the index file exists. The diskann ssd index is saved with the suffix `_disk.index` along with the filename.

cjnolet: This is algorithm-specific code in common benchmarking code, though. This should be moved into the diskann portions of the code. You have to understand that we create generalized abstractions for this very purpose- anything in `common/benchmark.hpp` should be applied to all index algorithms and any specific things should be propagated down to the respective index types. Polluting the common code like this moves up the abstraction tree, rather than down, and it results in hard to maintain spaghetti code.

cpp/bench/ann/src/common/benchmark.hpp
cjnolet: My question is: CAN we pass the `cuvs::bench::configuration::Index` down into the wrappers? Or rather, why shouldn't we?

cpp/bench/ann/src/common/benchmark.hpp
cjnolet: This is another reason to propagate this down to the individual wrappers. You could at least specify a method on the wrapper class to allow any algorithms to override and adjust the filename as needed. That would be more robust than one-offs like this.

cjnolet:
@tarang-jain i noticed some changes from @achirkin to create a singleton out of the configuration object, so this means you should be able to access it down in the individual algorithm wrappers now.

cpp/bench/ann/src/diskann/diskann_benchmark.cpp
```diff
@@ -0,0 +1,136 @@
...
+void parse_build_param(const nlohmann::json& conf,
+                       typename cuvs::bench::diskann_ssd<T>::build_param& param)
+{
+  param.R = conf.at("R");
+  if (conf.contains("L_build")) { param.L_build = conf.at("L_build"); }
+  if (conf.contains("alpha")) { param.num_threads = conf.at("alpha"); }
+  if (conf.contains("num_threads")) { param.num_threads = conf.at("num_threads"); }
+  if (conf.contains("QD")) { param.QD = conf.at("QD"); }
+  param.dataset_base_file = cuvs::bench::configuration::singleton().get_dataset_conf().base_file;
+  for (auto index : cuvs::bench::configuration::singleton().get_indices()) {
```
achirkin, line 54: Nitpick: avoid copying the index during iteration.
```suggestion
  for (const auto& index : cuvs::bench::configuration::singleton().get_indices()) {
```

cpp/bench/ann/src/diskann/diskann_benchmark.cpp
```diff
@@ -0,0 +1,136 @@
...
+    if (index.build_param == param) {
+      param.index_file = index.file;
```
achirkin, line 56: Nitpick: stop iterating as soon as we find the item.
```suggestion
      param.index_file = index.file;
      break;
```

cpp/bench/ann/src/cuvs/cuvs_vamana.cu
```diff
@@ -0,0 +1,93 @@
+/*
+ * Copyright (c) 2024, NVIDIA CORPORATION.
```
achirkin, line 2: Please check the copyright years.
@cjnolet isn't this checked in style checks in CI?

achirkin (approved):
Thanks, the PR looks good to go (just a couple minor nitpicks below)

cpp/bench/ann/src/common/benchmark.hpp
achirkin: Fixed via configuration singleton

gforsyth:
`CMakeLists.txt` from the upstream DiskANN repo is setting `CMAKE_CXX_COMPILER` explicitly to `g++` which is skipping the `g++` compiler in the conda environment and as a result, using the wrong `ld`.

These lines should be added to the `diff` so they can be removed when that file gets patched:
https://github.com/tarang-jain/DiskANN/blob/cmake/CMakeLists.txt#L26-L28

tarang-jain:
The diff is looking too big because https://github.com/rapidsai/cuvs/pull/675 still needs to be merged.

conda/recipes/cuvs-bench-cpu/meta.yaml
```diff
@@ -62,6 +65,9 @@ requirements:
     - pyyaml
     - python
     - requests
+    - libaio  # [linux64]
```
jakirkham, line 68: The [`libaio` package has `run_exports`]( https://github.com/conda-forge/libaio-feedstock/blob/b59670c44db55fae31ea8db9994888eb7c963447/recipe/meta.yaml#L14-L15 ). Meaning that once `libaio` is in `requirements/host`, it will automatically add itself to `requirements/run` with proper version constraints. So we can drop this from `requirements/run`

```suggestion
```

jakirkham (commented):
Thanks Tarang! 🙏

Had a couple questions below

Note: There are other occurrences like those commented on. Just picked one of each to simplify the discussion

conda/recipes/cuvs-bench-cpu/meta.yaml
```diff
@@ -62,6 +65,9 @@ requirements:
...
+    - libboost-devel=1.87  # [linux64]
+    - mkl-devel=2023  # [linux64]
```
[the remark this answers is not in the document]
tarang-jain: Currently we don't need them. @dantegd has recently been adding tests to CI for cuvs-bench, and so we would eventually need these in the run environment (when tests for diskann wrapper are added). But for the time being we can certainly drop them.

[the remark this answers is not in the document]
tarang-jain: @jakirkham I think it makes sense to drop these for now. We can add them later.

cjnolet (approved):

PR #261, opened 2024-07-30
Dynamic Batching

achirkin:
Non-blocking / stream-ordered dynamic batching as a new index type.

## API

This PR implements dynamic batching as a new index type, mirroring the API of other indices.

  * [_building is wrapping_] Building the index means creating a lightweight wrapper on top of an existing index and initializing necessary components, such as IO batch buffers and synchronization primitives.
  * [_type erasure_] The underlying/upstream index type is erased once the dynamic_batching wrapper is created, i.e. there's no way to recover the original search index type or parameters.
  * [_explicit control over batching_] To allow multiple user requests group into a dynamic batch request, the users must use copies of the same dynamic batching index (the user-facing index type is a thin wrapper on top of a shared pointer, hence the copy is shallow and cheap). The search function is thread-safe.

## Feature:  stream-ordered dynamic batching

Non-blocking / stream-ordered dynamic batching means the batching does not involve synchronizing with a GPU stream. The control is returned to the user as soon as the necessary work is submitted to the GPU. This entails a few good-to-know features:

1. The dynamic batching index has the same blocking properties as the upstream index: if the upstream index does not involve stream sync during search, that the dynamic batching index does not involve it as well (otherwise, the dynamic batching search obviously waits till the upstream search synchronizes under the hood).
2. It's responsibility of the user to synchronize the stream before getting the results back - even if the upstream index search does not need it (the batch results are scattered back to the request threads in a post-processing kernel).
3. If the upstream index does not synchronize during search, the dynamic batching index can group the queries even in a single-threaded application (_try it with --no-lap-sync option in the ann-bench benchmarks_).

Overall, stream-ordered dynamic batching makes it easy to modify existing cuVS indexes, because the wrapped index has the same execution behavior as the upstream index.

## Work-in-progress TODO

- [x] Add dynamic batching option to more indices in ann-bench
- [x] Add tests
- [x] **(postponed to 25.02)** Do proper benchmarking and possibly fine-tune the inter-thread communication
- [x] Review the API side (`cpp/include/cuvs/neighbors/dynamic_batching.hpp`) [ready for review CC @cjnolet]
- [x] Review the algorithm side (`cpp/src/neighbors/detail/dynamic_batching.cuh`) [ready for preliminary review: requests for algorithm docsting/clarifications are especially welcome]


achirkin:
Sneak-peek into performance (single-query benchmarks on a workstation):
CAGRA multi-cta original vs dynamically batched
CAGRA single-cta original vs dynamically batched

achirkin:
Multi-cta benchmarks with `--no-lap-sync` option:

CAGRA multi-cta original vs dynamically batched _ --no-lap-sync

Interestingly, the disabled event synchronization between benchmark iterations makes the requests coming from a single thread group together and saturate the dynamic batching. The CPU submits all work significantly earlier than the GPU completes it; as a result the GPU always has enough full batches and never stays idle. Note, this is only possible because the results are not transferred to the host until the entire benchmark loop finishes.

examples/cpp/src/dynamic_batching_example.cu
```diff
@@ -0,0 +1,279 @@
...
+  // preparing IO buffers)
+  dynamic_batching::index_params<decltype(orig_index)> dynb_index_params{
+      cuvs::distance::DistanceType::L2Expanded, 2.0, orig_index,
+      /* We need to know the search parameters in advance for batching */
+      orig_search_params, nullptr, topk, queries.extent(1),
+      /* Configure the number and the size of IO buffers */
+      64, kNumWorkerStreams};
+
+  // "build" the index
+  dynamic_batching::index<float, uint32_t> dynb_index(res, dynb_index_params);
```
tfeher, line 156: Right now `dynb_index_params` holds a reference to the original index, and the new dynamic index is constructed using this params struct. Alternatively, one could pass the `orig_index` to the constructor of the dynamic index:

```c++
dynamic_batching::index_params<decltype(orig_index)> dynb_index_params{
      orig_search_params, nullptr, topk, queries.extent(1),
      /* Configure the number and the size of IO buffers */
      64, kNumWorkerStreams};

  // "build" the index
  dynamic_batching::index<float, uint32_t> dynb_index(res, dynb_index_params, orig_index);
```

I would slightly prefer this, because in my mind the `orig_index` is not a parameter, but the object around which we are adding a wrapper to do the dynamic batching (the same way as `dataset` is not a parameter of `index_params`). 

Let's wait for other opinions of this topic.

examples/cpp/src/dynamic_batching_example.cu
```diff
@@ -0,0 +1,279 @@
...
+  cagra::search_params orig_search_params;
+  // get a decent recall by increasing the internal topk list
+  orig_search_params.itopk_size = 512;
+  orig_search_params.algo = cagra::search_algo::SINGLE_CTA;
+
+  // Dynamic batching index takes the upstream index as a build parameter
+  // ("building" dynamic_batching::index means just wrapping the upstream and
+  // preparing IO buffers)
+  dynamic_batching::index_params<decltype(orig_index)> dynb_index_params{
+      cuvs::distance::DistanceType::L2Expanded, 2.0, orig_index,
```
tfeher, line 149: The metric is already part of the `orig_index`. Can we provide a constructor that sets the metric and metric arg automatically from `orig_index`?

cpp/include/cuvs/neighbors/dynamic_batching.hpp
```diff
@@ -0,0 +1,134 @@
...
+  /** The index, to which the dynamic batches are dispatched. */
+  Upstream& upstream;
+  /** Search paramerets for all requests within a batch. */
+  typename Upstream::search_params_type& upstream_params;
+  /** Filtering function, if any, must be the same for all requests in a batch. */
+  cuvs::neighbors::filtering::base_filter* sample_filter = nullptr;
+  /** The number of neighbors to search is fixed at construction time. */
+  int64_t k;
+  /** Input data (queries) dimensionality. */
+  int64_t dim;
```
tfeher, line 43: Could we remove `dim` and query the original index for `dim` while constructing the batched index?

tfeher (changes requested):
Hi @achirkin, thank you for this PR. I have started to have a look, here is my first batch of comments.

examples/cpp/src/dynamic_batching_example.cu
achirkin: I very much agree with this. I find it also particularly clumsy to use the aggregate constructor in the index params (in benchmarks and in the example file) and a little unsafe to keep the reference there (a user must keep the original index alive).

In here, I attempted to keep the same build API as the other indexes, but failed (`::build(...)` function arguments don't fit here at all). I'd be happy to hear further suggestions and adapt the code once we have a consensus.

examples/cpp/src/dynamic_batching_example.cu
achirkin: We've been avoiding any non-default constructors/destructors in the index and search params and I hoped to keep it the same for the dynamic batching.
Perhaps, if we remove all parameters that do not have defaults (in particular, upstream reference), this problem goes away, because we could use the default constructor again.

cpp/include/cuvs/neighbors/dynamic_batching.hpp
achirkin: Hmm, probably yes. I'd need to check if all our indices have a matching getter method with exactly the same signature.

cpp/bench/ann/src/cuvs/cuvs_ann_bench_param_parser.h
```diff
@@ -56,6 +56,24 @@ extern template class cuvs::bench::cuvs_cagra<int8_t, uint32_t>;
...
+    param.dynamic_batching_max_batch_size = conf.at("dynamic_batching_max_batch_size");
+  }
+  if (conf.contains("dynamic_batching_soft_deadline_ms")) {
+    param.dynamic_batching_soft_deadline_ms = conf.at("dynamic_batching_soft_deadline_ms");
+  }
+  if (conf.contains("dynamic_batching_n_queues")) {
+    param.dynamic_batching_n_queues = conf.at("dynamic_batching_n_queues");
+  }
+  param.dynamic_batching_k =
+    uint32_t(uint32_t(conf.at("k")) * float(conf.value("refine_ratio", 1.0f)));
```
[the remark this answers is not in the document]
achirkin: That's because `k` is surprisingly not a search parameter in `ann_bench`. Instead, it's an argument of the `search` function. The config still provides this value for the algorithms, but they need custom logic to parse and store it (for access in `set_search_param` function which doesn't have it as an argument otherwise). In this case, the batcher needs to know the maximum number of neighbors it will need to return (to allocate the buffer).

achirkin:
### Current progress note

At the moment, all PRs, upon which this PR depends, are merged - there are no pending fixes/blockers in cuVS or raft.

There's one known bug that leads to a deadlock in rare cases:
When many concurrent threads submit their work, they subscribe for the incoming batch slots in the `batch_queue_t` (ahead of the free buffers pushed back by the GPU). If they do this fast enough and loop over the whole ring buffer, they may eventually skip the slot from the previous round over the buffer, which makes it impossible to push to that slot at a later point. Currently, the hardcoded buffer size is [256 slots](https://github.com/rapidsai/cuvs/blob/3603e2f78790125766ff1db5829a3542a082d8c8/cpp/src/neighbors/detail/dynamic_batching.cuh#L552). For benchmarks, I workaround this by increasing this size to 65536 or more, but a proper fix is planned (coming early next week probably).

achirkin:
### Current progress note

 - The implementation is ready for review; all known problems are resolved.
 - The API waits for more reviewers (see https://github.com/rapidsai/cuvs/pull/261#discussion_r1830659249)
 - There are a few problems with the benchmark setup, which prevent dynamic batching from reaching good speedups. I suggest we merge this PR and leave further performance studies to a follow-on.

achirkin:
For the record, the list of identified benchmark problems, which require a further study:

1. CAGRA produces `uint32_t` indices, which are mapped in a post-processing kernel to `int64_t`. This is not dynamically-batched; as a result all threads in the benchmark launch this kernel and cause CUDA context locks degrading the performance.
2. IVF-PQ only makes sense with refinement, but suffers from the same problem as (1), because the refinement is performed in every thread rather than once per batch. Interestingly, the problem persists even if the refinement is performed on CPU (maybe it's the intermediate results copy using CUDA, maybe it's just CPU overuse).
3. The benchmarks runs equal number of iterations in every thread in `throughput` mode. The dynamic batching loads the threads not equally and thus causes significant difference in the per-thread run time. This, in turn, leads to a long tail effect, where one or few threads run significantly longer than the others. The dynamic batcher is never able to get enough threads for the full batch and is forced to wait longer (full dispatch time on every iteration). As a result, the total average time/QPS can be hurt as well.

examples/cpp/src/dynamic_batching_example.cu
achirkin: I updated the API to pass more parameters via the index constructor and made the index parameters non-template.

cpp/src/neighbors/dynamic_batching.cu
```diff
@@ -0,0 +1,89 @@
...
+              cuvs::neighbors::dynamic_batching::search_params const& params,      \
+              cuvs::neighbors::dynamic_batching::index<T, IdxT> const& index,      \
+              raft::device_matrix_view<const T, int64_t, raft::row_major> queries, \
+              raft::device_matrix_view<IdxT, int64_t, raft::row_major> neighbors,  \
+              raft::device_matrix_view<float, int64_t, raft::row_major> distances) \
+  {                                                                                \
+    return index.runner->search(res, params, queries, neighbors, distances);       \
+  }
+
+CUVS_INST_DYNAMIC_BATCHING_INDEX(float, uint32_t, cuvs::neighbors::cagra, index<float, uint32_t>);
```
cjnolet, line 54: It's really unfortunate that we'll need to instantiate these individually for each index type.  For example, Vamana is not included here. Is there any way we can remove this constaint? Can we just tie this to the search_params super class?

cpp/include/cuvs/neighbors/dynamic_batching.hpp
```diff
@@ -0,0 +1,164 @@
...
+ * One lightweight dynamic batching index manages a single index and a single search parameter set.
+ * Upon creation, it starts a separate thread to serve the incoming requests.
+ * The server thread is terminated on the destruction of the index.
+ */
+template <typename T, typename IdxT>
+struct index : cuvs::neighbors::index {
+  std::shared_ptr<detail::batch_runner<T, IdxT>> runner;
+
+  /**
+   * @brief Construct a dynamic batching index by wrapping the upstream index.
```
cjnolet, line 87: Please include a usage example and add this to the API docs.

cpp/include/cuvs/neighbors/dynamic_batching.hpp
```diff
@@ -0,0 +1,164 @@
...
+   */
+  template <typename Upstream>
+  index(const raft::resources& res,
+        const cuvs::neighbors::dynamic_batching::index_params& params,
+        const Upstream& upstream_index,
+        const typename Upstream::search_params_type& upstream_params,
+        const cuvs::neighbors::filtering::base_filter* sample_filter = nullptr);
+};
+
+void search(raft::resources const& res,
```
cjnolet, line 108: Please document these functions. I know it seems redundant, but it's important users can look these functions up in the docs. Also please add usage examples.

cpp/include/cuvs/neighbors/ivf_flat.hpp
```diff
@@ -138,6 +138,10 @@ using list_data = ivf::list<list_spec, SizeT, ValueT, IdxT>;
  */
 template <typename T, typename IdxT>
 struct index : cuvs::neighbors::index {
+  using index_params_type  = ivf_flat::index_params;
```
cjnolet, line 141: Please also include the other index types (e.g. bfknn, vamana, etc...).

cpp/test/CMakeLists.txt
```diff
@@ -175,6 +175,19 @@ if(BUILD_TESTS)
     100
   )
 
+  ConfigureTest(
+    NAME
+    NEIGHBORS_DYNAMIC_BATCHING_TEST
+    PATH
+    neighbors/dynamic_batching/test_cagra.cu
```
cjnolet, line 182: Please include the other index types here as well (e.g. bfknn, vamana, etc...).

cpp/src/neighbors/dynamic_batching.cu
achirkin: I agree and I see couple solutions to this.

One is to go with the class-based polymorphism.
Then we'd have to make the search parameters `neighbors::search_params` and the index type `neighbohrs::index` virtual, by adding the virtual destructor type. We will also need a virtual `clone()` method, so we can copy implementation search parameters via the base class. This goes slightly against our initial design of keeping the search parameters a POD. This also means it would be dangerous to pass search parameters struct to kernels (but I think we haven't been doing this so far).
Then we would also need to add virtual `search` method to the index (and also `dim()` which is currently used by the dynamic batching), which goes slightly against our initial design of having search/build functions as plain functions.
Then there will be only one, non-templated constructor taking the abstract upstream index and search parameters.

Another solution is to go with the template-based polymorphism.
I could define a template constructor in the public header file (similar to what I have in the `dynamic_batching_test` at the moment). It would take the index, search params, and the search function as the template parameters, so that users can instantiate it on the user side. I think I would have to slightly rework the `detail::batch_runner` and expose at least one of its constructors in the public header for that.
This obviously goes against our design of not instantiating anything on the user side (but it doesn't involve any cuda-specific code and should be fast).

cpp/test/neighbors/dynamic_batching.cuh
```diff
@@ -0,0 +1,286 @@
...
+      if (i >= ps.max_concurrent_threads) { futures[j].wait(); }
+      // submit a new job
+      if (i < ps.n_queries) {
+        futures[j] =
+          std::async(std::launch::async,
+                     [&res       = resource_pool[j],
+                      &params    = search_params_dynb,
+                      index      = index_dynb.value(),
+                      query_view = raft::make_device_matrix_view<data_type, int64_t>(
+                        queries->data_handle() + i * ps.dim, 1, ps.dim),
```
tfeher, line 153: Here we submit one query at a time, right? Could we test with more than one queries as well?

tfeher (changes requested):
I have reviewed the implementation details. Thanks Artem for the additional documentation, overall the code looks great.

To achieve high throughput and low latency, one has to watch out for intricate details for queuing and synchronization, which makes the implementation complex. I have left a few comments that requests additional explanation, and suggests potential refactoring to make the logic easier to follow.

cpp/include/cuvs/neighbors/dynamic_batching.hpp
tfeher: Please also explain how do we know if the batch is finished. Is it just a sync with the stream in `res`?

cpp/bench/ann/src/cuvs/cuvs_ann_bench_param_parser.h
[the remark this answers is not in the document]
achirkin: Good point! I've added a check that issues a warning and falls back to a non-batched behavior for the given request.

cpp/src/neighbors/detail/dynamic_batching.cuh
```diff
@@ -0,0 +1,1073 @@
...
+                                  const cuvs::neighbors::filtering::base_filter&);
+
+template <typename T, typename IdxT>
+using function_search_type = void(raft::resources const&,
+                                  raft::device_matrix_view<const T, int64_t, raft::row_major>,
+                                  raft::device_matrix_view<IdxT, int64_t, raft::row_major>,
+                                  raft::device_matrix_view<float, int64_t, raft::row_major>);
+
+/** State of the batch token slot. */
+enum struct slot_state : int32_t {
```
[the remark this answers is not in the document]
achirkin: You're absolutely right. The crucial detail here is that the state depends not on the token only, but also on the thread/actor looking at it (see `batch_queue_t::batch_status`) it is computed as a difference between (function of) sequential counter ids of the actor that puts the value and the actor that reads the value (this is all to avoid the ABA atomic update problem).
I'll write this down in the docstring.

cpp/src/neighbors/detail/dynamic_batching.cuh
```diff
@@ -0,0 +1,1073 @@
...
+          : std::nullopt}
+  {
+    tail_.store(0, kMemOrder);
+    head_.store(0, kMemOrder);
+    for (uint32_t i = 0; i < kSize; i++) {
+      rem_time_us_(i).store(std::numeric_limits<int32_t>::max(), kMemOrder);
+      if (batch_sizes_.has_value()) { batch_sizes_.value()(i).store(0, kMemOrder); }
+      dispatch_sequence_id_[i].store(uint32_t(-1), kMemOrder);
+      tokens_(i).store(
+        batch_token{static_cast<uint32_t>(slot_state::kEmpty) * kSize + kCounterLocMask},
```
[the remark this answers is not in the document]
achirkin: `kSize` determines the number of bits used for the "round id" part of the `seq_order_id`, such that the remaining (`log2(kSize)`) bits are used for the IO buffer id when the token is full (or ignored otherwise).

I'll use a helper function here to make the intent clear and add the explanation in the helper function.

cpp/src/neighbors/detail/dynamic_batching.cuh
```diff
@@ -0,0 +1,1073 @@
...
+  /** Offset of the `id()` value in the token if it's interpreted as uint32_t[2]. */
+  static constexpr inline uint32_t kOffsetOfId = CUVS_SYSTEM_LITTLE_ENDIAN;
+  /** Offset of the `size_committed()` value in the token if it's interpreted as uint32_t[2]. */
+  static constexpr inline uint32_t kOffsetOfSC = 1 - kOffsetOfId;
+};
+static_assert(sizeof(batch_token) == sizeof(uint64_t));
+static_assert(cuda::std::atomic<batch_token>::is_always_lock_free);
+
+template <uint32_t Size>
+struct batch_queue_t {
```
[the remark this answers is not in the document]
achirkin: - Yes
- It means it's full or deadlined, but not necessarily submitted yet. There may be more concurrent requesters filling in (committing to) the batches faster than they are submitted.
- No, it's the batches available for use
- `push()` reserves a place in the queue for the GPU to fill it in later
- You're right. The primary 1D states are `full` and `empty`; the rest is for consistency to avoid ABA atomic updates.
I've spread this answer to relevant declarations. There's quite a lot :)

cpp/src/neighbors/detail/dynamic_batching.cuh
```diff
@@ -0,0 +1,1073 @@
...
+      timeout = true;
+      return false;
+    }
+    return true;
+  }
+
+ private:
+  cuda::atomic<int32_t, cuda::thread_scope_system>* cpu_provided_remaining_time_us_;
+  uint64_t timestamp_ns_           = 0;
+  int32_t local_remaining_time_us_ = std::numeric_limits<int32_t>::max();
```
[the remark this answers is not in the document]
achirkin: Sort of, but not exactly :)

On the one hand, we initialize the timestamp at the construction. But the timestamp is updated when `has_time()` is called first time and it's compared to int_max at that point, so effectively it's ignored.

On the other hand, the participating host threads also provide the remaining time in their "private coordinate systems"  - where the remaining time is computed w.r.t. the point at the beginning of the search call (even before they commit to a batch).

So the first timestamp and the first local check here are just to provide sensible initialization for the counter.

cpp/src/neighbors/detail/dynamic_batching.cuh
```diff
@@ -0,0 +1,1073 @@
...
+ * Grid size: (max_batch_size, 1, 1) - one block per query
+ */
+template <typename T, typename IdxT>
+RAFT_KERNEL gather_inputs(
+  raft::device_matrix_view<T, uint32_t, raft::row_major> batch_queries,
+  raft::pinned_vector_view<request_pointers<T, IdxT>, uint32_t> request_ptrs,
+  /* The remaining time may be updated on the host side: a thread with a tighter deadline may reduce
+     it (but not increase). */
+  cuda::atomic<int32_t, cuda::thread_scope_system>* remaining_time_us,
+  /* The token contains the current number of queries committed and is cleared in this kernel. */
```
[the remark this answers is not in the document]
achirkin: A query is committed means the host thread reserved a slot in a batch for it. A batch is fully committed when all slots are reserved (or timeout exceeded).

cpp/src/neighbors/detail/dynamic_batching.cuh
```diff
@@ -0,0 +1,1073 @@
...
+      // This branch checks if the token was recently filled or dispatched.
+      // This means the head counter of the ring buffer is slightly outdated.
+      if (token_status == slot_state::kEmptyPast || token_status == slot_state::kFullPast ||
+          batch_token_observed.size_committed() >= max_batch_size_) {
+        batch_queue_.pop(seq_id);
+        return false;
+      }
+      batch_token_updated = batch_token_observed;
+      batch_token_updated.size_committed() =
+        std::min(batch_token_observed.size_committed() + n_queries, max_batch_size_);
```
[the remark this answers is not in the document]
achirkin: I think, you're right, we can do this, but there's been two things preventing me from doing that so far:
 - The queue is not aware of the `max_batch_size`, but it's needed for the commit logic
 - We still need the states when we're waiting for the IO buffer after committing to the batch (see Note: waiting for batch IO buffers)

None of the two seem to be complete blockers though.

cpp/src/neighbors/detail/dynamic_batching.cuh
```diff
@@ -0,0 +1,1073 @@
...
+      // TODO: it could potentially be more efficient to first commit everything and only then
+      //        submit the work/wait for the event
+      local_io_offset += queries_committed;
+      to_commit.reset(
+        local_waiter::kNonSleepIterations);  // reset the waiter for the next iteration.
+    }
+  }
+
+ private:
+  raft::resources res_;  // Sic! Store by value to copy the resource.
```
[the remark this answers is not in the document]
achirkin: We need a separate resource for the `batch_runner` to not rely on the lifetime of the argument resource (a user can easily pass a short-lived temporary resource).
We allocate a few device objects using this resource, so we need to make sure they are not invalidated, and that the memory resource stays in the same place.

cpp/src/neighbors/detail/dynamic_batching.cuh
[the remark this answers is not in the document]
achirkin: (Added this to the function docstring as well)

cpp/include/cuvs/neighbors/ivf_flat.hpp
achirkin: I've checked all available indexes and, unfortunately, I cannot add any more instances at this point:
 - bfknn would be the best candidate, but its build/search methods do not adhere to the expected api (missing index/search params as function arguments).
 - vamana and nn_descend do not have corresponding `search` functions, which are needed for instantiating the dynamic batching index.

We can adapt the code to allow bfknn, but it would require more changes. I suggest we postpone this for 25.02 together with https://github.com/rapidsai/cuvs/pull/261#discussion_r1851601263

tfeher (approved):
Thanks Artem for the added documentation. The PR looks good to me. Please open an issue for refactoring ideas that are feasible to implement as a follow up.

cjnolet (approved):

Issue #144, opened 2024-05-22
[FEA] Support fp16 input type for vector search


tfeher:
The `ivf_pq` and `cagra` algorithms in RAFT do support `half` input type. cuVS should also support index building using `fp16` input. This is a useful feature for index building: using `fp16` is sufficient to several datasets, and it reduces the size of the index. Moreover, embedding models might generate vectors in `fp16`, and it would be unfortunate if we need to convert that to fp32 for index building.

xiayuqing0622:
Agree!

PR #264, opened 2024-07-30
FP16 API for CAGRA and IVF-PQ

tfeher:
This PR adds public API to CAGRA and IVF-PQ ANN search using FP16 input data.

Note that the fp16 kernels are already compiled in libcuvs. This PR just adds the missing public API declarations into the C++ headers, and restores the instantiations of public API functions.

This PR partially fixes https://github.com/rapidsai/cuvs/issues/144 (the IVF-Flat API is not yet added here).


cjnolet:
We also need to measure the impact of binary size and compile time for adding these new types to the public API. 

We can't keep increasing without figuring out ways we can consolidate what's there. These two thjngs are the number 1 complaint from users currently. (It's not just this PR. This is also holding up the half precision bfknn and RBC PRs).

cjnolet:
Linking https://github.com/rapidsai/cuvs/issues/110

tfeher:
The kernels are already compiled and included in `libcuvs.so`. The additional instantiations of the API entry points shall have negligible size. I will confirm this once CI finishes. 

It was a mistake during porting the code from RAFT, that the public API was not defined for fp16, therefore I labelled this PR as a bugfix.

cpp/bench/ann/src/cuvs/cuvs_ivf_pq.cu
```diff
@@ -17,7 +17,7 @@
 
 namespace cuvs::bench {
 template class cuvs_ivf_pq<float, int64_t>;
-// template class cuvs_ivf_pq<half, int64_t>;
+template class cuvs_ivf_pq<half, int64_t>;
```
achirkin, line 20: I assume, it's ok to keep all `cuvs_ivf_pq` in one file here, because the wrapper instances are fast to compile, right?
Do you know, why this is not the case with cagra (since we put wrapper instances for cagra in separate files)?

cpp/include/cuvs/neighbors/cagra.hpp
```diff
@@ -966,6 +1038,30 @@ void search(raft::resources const& res,
             raft::device_matrix_view<uint32_t, int64_t, raft::row_major> neighbors,
             raft::device_matrix_view<float, int64_t, raft::row_major> distances);
 
+/**
+ * @brief Search ANN using the constructed index.
+ *
+ * See the [cagra::build](#cagra::build) documentation for a usage example.
+ *
+ * @tparam T data element type
+ * @tparam IdxT type of the indices
```
achirkin, line 1047:
```suggestion
```
No template params in the public api.

cpp/include/cuvs/neighbors/cagra.hpp
```diff
@@ -1390,6 +1591,58 @@ void serialize_to_hnswlib(raft::resources const& handle,
...
+ * @endcode
+ *
+ * @param[in] handle the raft handle
+ * @param[in] os output stream
+ * @param[in] index CAGRA index
+ *
+ */
+void serialize_to_hnswlib(raft::resources const& handle,
+                          std::ostream& os,
+                          const cuvs::neighbors::cagra::index<half, uint32_t>& index);
```
achirkin, line 1618: Just curious, does HNSW support half data type or we convert it internally to fp32? If the latter, perhaps it makes sense to document this.

cpp/src/neighbors/cagra_search_half.cu
```diff
@@ -0,0 +1,37 @@
+/*
+ * Copyright (c) 2023-2024, NVIDIA CORPORATION.
```
achirkin, line 2: Here and in other files newly moved from raft, shall we have the copyright starting year copied from raft, or reset the counter to 2024?

cpp/src/neighbors/ivf_pq/detail/ivf_pq_search_half_int64_t.cu
```diff
@@ -0,0 +1,47 @@
...
+  void search(raft::resources const& handle,                                    \
+              const cuvs::neighbors::ivf_pq::search_params& params,             \
+              cuvs::neighbors::ivf_pq::index<IdxT>& index,                      \
+              raft::device_matrix_view<const T, IdxT, raft::row_major> queries, \
+              raft::device_matrix_view<IdxT, IdxT, raft::row_major> neighbors,  \
+              raft::device_matrix_view<float, IdxT, raft::row_major> distances) \
+  {                                                                             \
+    cuvs::neighbors::ivf_pq::detail::search(                                   \
+      handle, params, index, queries, neighbors, distances);  \
+  }
```
achirkin, line 42: I know it's a bit out of the scope of this PR, but I think it would make sense to remove the `#define CUVS_INST_IVF_PQ_SEARCH` in the generated code and put it in a separate header. Like we do with `CUVS_INST_IVF_PQ_BUILD_EXTEND` and a few others: for consistency and to reduce code duplication.

achirkin (approved):
Thanks Tamas for the PR! The changes are straightforward and it looks good to me as-is, yet a few nitpicks below (if the time permits).

cpp/src/neighbors/cagra_search_half.cu
bdice: If an earlier start year exists, by copying from RAFT, don’t reset it. The copyright script only enforces that the end year is current. Leave start years at the earlier of (whatever git log says) and (a year in which we know the code existed in RAFT or other NVIDIA libraries).

tfeher: Thanks @bdice for the clarification!

cpp/src/neighbors/ivf_pq/detail/ivf_pq_search_half_int64_t.cu
tfeher: I might prefer to remove the macro. Created #269 to track this.

tfeher (commented):
Thanks Artem for the review, I have addressed the issues.

achirkin:
Binary size: [679](https://downloads.rapids.ai/ci/cuvs/branch/branch-24.10/8e02150/cuda12_x86_64.compile_lib.html) -> [661](https://downloads.rapids.ai/ci/cuvs/pull-request/215/0bfb6be/cuda12_x86_64.compile_lib.html) MB
The size is actually reduced, because I removed a few unused template instances and most of FP16-related instances were already in the binary (unused until this PR).

A quick check with bench-ann shows the FP16 benchmarks seem to work. I've checked that CAGRA is a little bit faster on FP16 vs FP32 on the glove dataset.

cpp/src/neighbors/detail/cagra/cagra_build.cuh
```diff
@@ -357,8 +353,7 @@ void build_knn_graph(
   raft::host_matrix_view<IdxT, int64_t, raft::row_major> knn_graph,
   cuvs::neighbors::nn_descent::index_params build_params)
 {
-  auto nn_descent_idx = cuvs::neighbors::nn_descent::index<IdxT>(res, knn_graph);
-  cuvs::neighbors::nn_descent::build<DataT, IdxT>(res, build_params, dataset, nn_descent_idx);
+  auto nn_descent_idx = cuvs::neighbors::nn_descent::build(res, build_params, dataset);
```
achirkin, line 356: We have to undo this migration to public API, because the public overload produces an index that owns its knn_graph memory, whereas the current design assumes the graph comes in the arguments of this function.

docs/source/developer_guide.md
```diff
@@ -296,20 +296,20 @@ Sometimes, we need to temporarily change the log pattern (eg: for reporting deci
 
 RAFT is a heavily templated library. Several core functions are expensive to compile and we want to prevent duplicate compilation of this functionality. To limit build time, RAFT provides a precompiled library (libraft.so) where expensive function templates are instantiated for the most commonly used template parameters. To prevent (1) accidental instantiation of these templates and (2) unnecessary dependency on the internals of these templates, we use a split header structure and define macros to control template instantiation. This section describes the macros and header structure.
 
-**Macros.** We define the macros `RAFT_COMPILED` and `RAFT_EXPLICIT_INSTANTIATE_ONLY`. The `RAFT_COMPILED` macro is defined by `CMake` when compiling code that (1) is part of `libraft.so` or (2) is linked with `libraft.so`. It indicates that a precompiled `libraft.so` is present at runtime.
+**Macros.** We define the macros `CUVS_COMPILED` and `CUVS_EXPLICIT_INSTANTIATE_ONLY`. The `CUVS_COMPILED` macro is defined by `CMake` when compiling code that (1) is part of `libraft.so` or (2) is linked with `libraft.so`. It indicates that a precompiled `libraft.so` is present at runtime.
```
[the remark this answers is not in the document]
achirkin: You're right, perhaps this whole section is obsolete by now. But also I've just noticed these changes are not relevant to any of the changed code in this PR, so I'll just remove the changeset, so we can fix this in a separate PR.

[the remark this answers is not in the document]
cjnolet: I don't mean to make this your problem, btw. I think at some point we need to basicaly rewrite this developer guide so that it matches more closely to cuVS. RAFT is header only and we designed things a certain way for reusable primitives. cuVS is not header only and we've established some different programming paradigms as a result. I'll create an issue to revamp the developer guide eventually.

cjnolet (approved):

PR #268, opened 2024-07-31
Remove fp16 kernels that have no public entry point

tfeher:
`libcuvs.so` contains fp16 kernels that are not accessible (missing headers and missing public entry points). This PR removes the unused kernel.

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -448,11 +448,11 @@ add_library(
   src/neighbors/nn_descent_int8.cu
   src/neighbors/nn_descent_uint8.cu
   src/neighbors/refine/detail/refine_device_float_float.cu
-  # src/neighbors/refine/detail/refine_device_half_float.cu
+  src/neighbors/refine/detail/refine_device_half_float.cu
   src/neighbors/refine/detail/refine_device_int8_t_float.cu
   src/neighbors/refine/detail/refine_device_uint8_t_float.cu
   src/neighbors/refine/detail/refine_host_float_float.cpp
-  # src/neighbors/refine/detail/refine_host_half_float.cpp
+  src/neighbors/refine/detail/refine_host_half_float.cpp
   src/neighbors/refine/detail/refine_host_int8_t_float.cpp
   src/neighbors/refine/detail/refine_host_uint8_t_float.cpp
   src/neighbors/sample_filter.cu
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -458,7 +458,7 @@ add_library(
   src/neighbors/sample_filter.cu
   src/selection/select_k_float_int64_t.cu
   src/selection/select_k_float_uint32_t.cu
-  # src/selection/select_k_half_uint32_t.cu
+  src/selection/select_k_half_uint32_t.cu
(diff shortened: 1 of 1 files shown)


benfred (approved):

cpp/CMakeLists.txt
```diff
@@ -268,14 +268,14 @@ add_library(
...
   src/neighbors/detail/cagra/q_search_multi_cta_float_uint32_dim1024_t32_8pq_4subd_half.cu
-  src/neighbors/detail/cagra/q_search_multi_cta_half_uint32_dim128_t8_8pq_2subd_half.cu
-  src/neighbors/detail/cagra/q_search_multi_cta_half_uint32_dim128_t8_8pq_4subd_half.cu
-  src/neighbors/detail/cagra/q_search_multi_cta_half_uint32_dim256_t16_8pq_2subd_half.cu
-  src/neighbors/detail/cagra/q_search_multi_cta_half_uint32_dim256_t16_8pq_4subd_half.cu
-  src/neighbors/detail/cagra/q_search_multi_cta_half_uint32_dim512_t32_8pq_2subd_half.cu
-  src/neighbors/detail/cagra/q_search_multi_cta_half_uint32_dim512_t32_8pq_4subd_half.cu
-  src/neighbors/detail/cagra/q_search_multi_cta_half_uint32_dim1024_t32_8pq_2subd_half.cu
-  src/neighbors/detail/cagra/q_search_multi_cta_half_uint32_dim1024_t32_8pq_4subd_half.cu
+  # src/neighbors/detail/cagra/q_search_multi_cta_half_uint32_dim128_t8_8pq_2subd_half.cu
```
cjnolet, line 271: For good form, can we remove these please and then introduce them back in when we make the future changes?

cjnolet (changes requested):

cpp/CMakeLists.txt
tfeher: Fixed

cjnolet (approved):
LGTM!

tfeher:
Moving comments from PR description here:

Supporting FP16 is beneficial for decreasing memory footprint of large dataset, therefore it is planned to revert this PR in the future and add the missing interfaces.

#264 adds the missing pieces to run vector search with FP16 data, but it comes with ~ 50 MB increase in binary size:

Filename | build time | binary size
-- | -- | --
CMakeFiles/cuvs.dir/src/neighbors/cagra_search_half.cu.o | 15:03 min | 33.5 MB
CMakeFiles/cuvs.dir/src/neighbors/cagra_build_half.cu.o | 6:50 min | 11.5 MB
CMakeFiles/cuvs.dir/src/neighbors/ivf_pq/detail/ivf_pq_build_extend_half_int64_t.cu.o | 4:41 min | 9.0 MB
CMakeFiles/cuvs.dir/src/neighbors/cagra_serialize_half.cu.o | 60.411 s | 1.3 MB
CMakeFiles/cuvs.dir/src/neighbors/ivf_pq/detail/ivf_pq_search_half_int64_t.cu.o | 93.104 s | 1.7 MB

It is somewhat surprising that we have so large binary size increase, all these components shall be just thin wrappers around functions compiled in separate object files.

PR #271, opened 2024-08-01
Fix python ivf-pq for int8/uint8 dtypes

benfred:
The ivf-pq python bindings weren't working appropriately for int8/uint8 dtypes. Fix.


lowener (approved):
LGTM

PR #275, opened 2024-08-05
Add cuvs-bench to dependencies and conda environments

dantegd:
Adding conda environments and a few extra files needed for the upcoming cuvs-bench.

cc @tarang-jain @cjnolet could use a look at the correct version of packages like hnsw


conda/environments/bench_ann_cuda-125_arch-aarch64.yaml
```diff
@@ -0,0 +1,42 @@
...
+- cuda-profiler-api
+- cuda-python>=12.0,<13.0a0
+- cuda-version=12.5
+- cxx-compiler
+- cython>=3.0.0
+- dlpack>=0.8,<1.0
+- gcc_linux-aarch64=11.*
+- glog>=0.6.0
+- h5py>=3.8.0
+- hnswlib=0.7.0
```
tarang-jain, line 27: cuvs' cpp cmake patch has the hnswlib version as 0.6.2 so there is a mismatch. @divyegala can weigh in on this.

divyegala: We'll need to match versions for now until we can upgrade and create a new patch.

cjnolet (approved):

cjnolet (approved):

python/cuvs_bench/LICENSE
raydouglass, line 1: This should be a symlink to the `LICENSE` file in the root directory.

python/cuvs_bench/cuvs_bench/VERSION
raydouglass, line 1: This should be a symlink to the `VERSION` file in the root directory, so it will be updated automatically.

raydouglass (approved):

PR #276, opened 2024-08-06
Expose `extend()` in C API 

ajit283:
Add functionality to add additional vectors after build to C API


cpp/src/neighbors/cagra_c.cpp
```diff
@@ -265,6 +316,19 @@ extern "C" cuvsError_t cuvsCagraCompressionParamsDestroy(cuvsCagraCompressionPar
   return cuvs::core::translate_exceptions([=] { delete params; });
 }
 
+extern "C" cuvsError_t cuvsCagraExtendParamsCreate(cuvsCagraExtendParams_t* params)
+{
+  return cuvs::core::translate_exceptions([=] {
+    // auto ps = cuvs::neighbors::cagra::extend_params();
```
benfred, line 322:
```suggestion
```

cpp/src/neighbors/cagra_c.cpp
```diff
@@ -83,6 +83,34 @@ void* _build(cuvsResources_t res, cuvsCagraIndexParams params, DLManagedTensor*
...
+             cuvsCagraExtendParams params,
+             cuvsCagraIndex index,
+             DLManagedTensor* additional_dataset_tensor)
+{
+  auto dataset   = additional_dataset_tensor->dl_tensor;
+  auto index_ptr = reinterpret_cast<cuvs::neighbors::cagra::index<T, uint32_t>*>(index.addr);
+  auto res_ptr   = reinterpret_cast<raft::resources*>(res);
+
+  auto extend_params           = cuvs::neighbors::cagra::extend_params();
+  extend_params.max_chunk_size = params.max_chunk_size;
```
benfred, line 97: This doesn't need to be changed in this PR (especially since the rest of cagra/ivf-* doesn't follow this right now) - but I think we should start re-using the C-api parameter structs in the C++ api , like we are doing in the distances here https://github.com/rapidsai/cuvs/blob/72154b0b806c106300b52870f6113fdda3f87f0b/cpp/include/cuvs/distance/distance.hpp#L27. This means we don't have to have two different identical structs in the C and C++ api, and copy between them you are doing here

benfred (commented):
thanks for adding this functionality! This change looks good to me -

cpp/test/neighbors/ann_cagra_c.cu
```diff
@@ -66,6 +80,11 @@ TEST(CagraC, BuildSearch)
   cuvsCagraIndexParamsCreate(&build_params);
   cuvsCagraBuild(res, build_params, &dataset_tensor, index);
 
+  cuvsCagraExtendParams_t extend_params;
+  cuvsCagraExtendParamsCreate(&extend_params);
```
benfred, line 84: We should probably cleanup the memory here w/ `cuvsCagraExtendParamsDestroy` at the end

cpp/test/neighbors/ann_cagra_c.cu
```diff
@@ -66,6 +80,11 @@ TEST(CagraC, BuildSearch)
...
+  extend_params->max_chunk_size = 100;
+  cuvsCagraExtend(res, extend_params, &additional_dataset_tensor, index);
```
benfred, line 86: for debugging the segfault - does this still occur when called with the original dataset - like :
```suggestion
  cuvsCagraExtend(res, extend_params, &dataset_tensor, index);
```

ajit283: still segfaults unfortunately

cjnolet: @ajit283 I think @benfred might be right about this being from the number of vectors in the dataset. Can we generate a matrix using the make_blobs from raft::random ? Then we can just generate a larger dataset and bypass the issue.

ajit283:
👍 will look into this tomorrow

cjnolet:
@ajit283 can you fix the conflicts in this PR?

ajit283:
Done! We should look into the tests for this one.

- The previous approach was to test against a set of precomputed queries, expected neighbors and distances.
- This does not work anymore as we generate the dataset dynamically using `make_blobs()`. Right now I am adding a static set of vectors to the dataset using `extend()`, then query these vectors and check if I get them back with distance 0. This shows that `extend()` works, but is generally a regression compared to the previous tests.
- My ideas for improving the tests would be either to remove `make_blobs()` and use a larger static dataset, or to calculate expected neighbors and distances at runtime using brute-force.

What do you think?

ajit283:
@cjnolet I added a test using the pairwise distance matrix as discussed in the call

cpp/src/neighbors/cagra_c.cpp
cjnolet: +1 to this. It increases the maintenance burden when we have parallel objects like this.

cjnolet: @ajit283 as Ben mentioned, it doesn't have to be changed in this PR, but can you create a Github issue for this and reference the issue number in a TODO here in the code? This way we won't lose sight of it.

cpp/test/neighbors/ann_cagra_c.cu
```diff
@@ -126,3 +138,214 @@ TEST(CagraC, BuildSearch)
...
+  int32_t main_data_size       = 1024;
+  int32_t additional_data_size = 64;
+  int32_t num_queries          = 4;
+
+  // create random data for datasets and queries
+  rmm::device_uvector<float> random_data_d(
+    (main_data_size + additional_data_size + num_queries) * dimensions, stream);
+  rmm::device_uvector<int32_t> random_labels_d(
+    (main_data_size + additional_data_size + num_queries) * dimensions, stream);
+  raft::random::make_blobs(random_data_d.data(),
```
cjnolet, line 162: Very nice to see this! And i"m glad this helped get past the bug, at least for now. Do we have a Github issue for the problem you encountered here? If not, do you mind creating one and referencing the issue number in a TODO comment in the code?

cjnolet:
Overall, I think this PR is taking shape and looking great! I think the remaining items are minimal- mostly just doing the diligence of capturing Github issues.

cpp/test/neighbors/ann_cagra_c.cu
ajit283: Good idea - just created an issue and added it in a comment!

cpp/src/neighbors/cagra_c.cpp
divyegala: One problem that arises when we do this is that we have to include the C header file in the C++ header file. While it works for distances enum, here the issue is that we also have dlpack headers which we don't link to the C++ target.

So we'll have to pull out all the structs into another file to make it work.

@cjnolet @benfred

ajit283:
any idea why the style-check failed? 

```
include-check FAILED! See below for errors...
File: cpp/test/neighbors/ann_cagra_c.cu
  Line:18 use #include <...>
```

I can't see any issue with the include statement

cjnolet:
@ajit283 looks like we have a couple style check failures. We will need to fix those to run ci to completion

cpp/test/neighbors/ann_cagra_c.cu
```diff
@@ -15,6 +15,8 @@
  */
 
 #include "../test_utils.cuh"
+#include "cuvs/distance/distance.hpp"
```
lowener, line 18:
```suggestion
#include <cuvs/distance/distance.hpp>
```

ajit283:
it seems like the style check fails due to this:

```
The rapids-cmake cmake-format configuration file was not found at any of the default search locations: 

/__w/cuvs/cuvs/cpp/build/_deps/rapids-cmake-src/cmake-format-rapids-cmake.json

Try setting the environment variable RAPIDS_CMAKE_FORMAT_FILE to the path to the config file.
```

is this an issue with this PR specifically?

cjnolet:
Looks like we have a couple test failures for the C extend() API of CAGRA

cpp/src/neighbors/cagra_c.cpp
cjnolet: @divyegala in that case, I think we should consider different ways that we can consolidate the structs. The goal here is that we reudce duplication (and potential divergence) of trivial code when we could be defining them once and reusing. Whether we ultimately define them in a C++ header, include in a C header or vice versa, the final goal should be consolidation.

cpp/include/cuvs/neighbors/cagra.h
```diff
@@ -141,6 +141,42 @@ cuvsError_t cuvsCagraCompressionParamsCreate(cuvsCagraCompressionParams_t* param
...
+ * @{
+ */
+/**
+ * @brief Supplemental parameters to extend CAGRA Index
+ *
+ */
+struct cuvsCagraExtendParams {
+  /** Degree of input graph for pruning. */
+  uint32_t max_chunk_size;
+  /** Degree of output graph. */
```
benfred, line 159:
```suggestion
```

cpp/include/cuvs/neighbors/cagra.h
```diff
@@ -141,6 +141,42 @@ cuvsError_t cuvsCagraCompressionParamsCreate(cuvsCagraCompressionParams_t* param
...
+/**
+ * @defgroup cagra_c_extend_params C API for CUDA ANN Graph-based nearest neighbor search
+ * @{
+ */
+/**
+ * @brief Supplemental parameters to extend CAGRA Index
+ *
+ */
+struct cuvsCagraExtendParams {
+  /** Degree of input graph for pruning. */
```
benfred, line 157: Maybe should use the docstring from the c++ class 

```suggestion
  /** The additional dataset is divided into chunks and added to the graph. This is the knob to
   * adjust the tradeoff between the recall and operation throughput. Large chunk sizes can result
   * in high throughput, but use more working memory (O(max_chunk_size*degree^2)). This can also
   * degrade recall because no edges are added between the nodes in the same chunk. Auto select when
   * 0. */
```

benfred (commented):
lgtm! just a minor issue with the c docstrings that I think should get fixed up - but aside from that I think this is good to go

benfred (approved):

PR #279, opened 2024-08-06
Add cuvs_bench.run python code and build

dantegd:
Builds on top of #275


divyegala (approved):

cjnolet (approved):

divyegala:
@rapidsai/packaging-codeowners could we get a review here please?

dependencies.yaml
```diff
@@ -38,6 +38,7 @@ files:
       - develop
       - bench
       - bench_python
+      - rapids_build_setuptools
```
bdice, line 41: Typically we only use these lists as build-system extras in a pyproject.toml. Why is it going in a conda environment here?

python/cuvs_bench/pyproject.toml
```diff
@@ -64,7 +61,7 @@ skip = [
 version = { file = "cuvs_bench/VERSION" }
 
 [tool.rapids-build-backend]
-build-backend = "scikit_build_core.build"
+build-backend = "setuptools.build_meta"
```
bdice, line 64: Is this being managed by dependencies.yaml? See the entries on build-system extras for other projects.

bdice (commented):
A couple questions on dependencies.yaml usage. I assume this is basically the same as the RAFT code but ported to cuVS so I didn’t inspect it all super closely.

build.sh
```diff
@@ -419,6 +419,11 @@ if (( ${NUMARGS} == 0 )) || hasArg python; then
         python -m pip install --no-build-isolation --no-deps --config-settings rapidsai.disable-cuda=true ${REPODIR}/python/cuvs
 fi
 
+# Build and (optionally) install the cuvs_bench Python package
+if (( ${NUMARGS} == 0 )) || hasArg bench-ann; then
+    python -m pip install --no-build-isolation --no-deps --config-settings rapidsai.disable-cuda=true ${REPODIR}/python/cuvs_bench -vvv
```
bdice, line 424: Is the `-vvv` needed? We removed it in some builds somewhere in RAPIDS recently, I think we can drop it.
```suggestion
    python -m pip install --no-build-isolation --no-deps --config-settings rapidsai.disable-cuda=true ${REPODIR}/python/cuvs_bench
```

conda/recipes/libcuvs/build_libcuvs_tests.sh
```diff
@@ -1,5 +1,5 @@
 #!/usr/bin/env bash
 # Copyright (c) 2022-2024, NVIDIA CORPORATION.
 
-./build.sh tests bench-ann --allgpuarch --no-nvtx --build-metrics=tests_bench --incl-cache-stats
+./build.sh tests --allgpuarch --no-nvtx --build-metrics=tests_bench --incl-cache-stats
```
bdice, line 4: Does the `--build-metrics` flag need to change if benchmarks are not being built here?
```suggestion
./build.sh tests --allgpuarch --no-nvtx --build-metrics=tests --incl-cache-stats
```

python/cuvs_bench/cuvs_bench/run/__main__.py
```diff
@@ -0,0 +1,216 @@
...
+    "--dry-run",
+    is_flag=True,
+    help="Dry-run mode will convert the yaml config for the specified "
+    "algorithms and datasets to the json format that’s consumed "
+    "by the lower-level c++ binaries and then print the command to "
+    "run execute the benchmarks but will not actually execute "
+    "the command.",
+)
+@click.option(
+    "--raft-log-level",
```
bdice, line 142: Do references to RAFT need to be changed here, or do these still pass through to RAFT? This applies in several places.

python/cuvs_bench/cuvs_bench/run/__main__.py
```diff
@@ -0,0 +1,216 @@
...
+    "--groups",
+    default="base",
+    show_default=True,
+    prompt="Enter the comma separated groups of parameters",
+    help="Run only comma separated groups of parameters",
+)
+@click.option(
+    "--algo-groups",
+    help="Add comma separated <algorithm>.<group> to run. Example usage: "
+    ' "--algo-groups=raft_cagra.large,hnswlib.large".',
```
bdice, line 102: Does this need to be updated to cuvs?
```suggestion
    ' "--algo-groups=cuvs_cagra.large,hnswlib.large".',
```

python/cuvs_bench/pyproject.toml
```diff
@@ -33,9 +37,6 @@ classifiers = [
 [project.urls]
 Homepage = "https://github.com/rapidsai/raft"
```
bdice, line 38: This homepage should be updated.
```suggestion
Homepage = "https://github.com/rapidsai/cuvs"
```

bdice (changes requested):
Some necessary changes, and some questions for developers to determine whether changes are needed.

python/cuvs_bench/cuvs_bench/run/__main__.py
divyegala: Yes this still passes through to RAFT

.gitignore
```diff
@@ -79,3 +79,6 @@ cagra_index
 ivf_flat_index
 ivf_pq_index
 
+# cuvs_bench
+datasets/
+*.json
```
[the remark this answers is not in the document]
divyegala: Good point, I'll remove this. Thanks!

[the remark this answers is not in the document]
divyegala: Sorry, I meant I'll make it more specific to only ignore `.json` files from root directory

.gitignore
```diff
@@ -79,3 +79,6 @@ cagra_index
 ivf_flat_index
 ivf_pq_index
 
+# cuvs_bench
+datasets/
+/*.json
```
bdice, line 84: Can we add back the trailing newline? And is there a name pattern that we can attach to this? Like “/bench*.json”? It’s very possible that we might want top level configuration in JSON.

bdice (approved):
Thanks @divyegala.

.gitignore
divyegala: We'll rework this if we add top-level JSON configurations later on. For now, this tool sometimes create top-level JSON files so we'll refactor it, if needed, so it doesn't do that

PR #285, opened 2024-08-11
Add RMM Pool memory resource to C API

ajit283:
Enables switching to the pool allocator for better performance


cpp/include/cuvs/core/c_api.h
```diff
@@ -127,6 +127,24 @@ cuvsError_t cuvsRMMAlloc(cuvsResources_t res, void** ptr, size_t bytes);
...
+/**
+ * @brief Switches the memory resource to use the pool memory resource
+ *
+ * @param[in] initial_pool_size_percent The initial pool size as a percentage of the total
+ * available memory
+ * @param[in] max_pool_size_percent The maximum pool size as a percentage of the total
+ * available memory
+ * @return cuvsError_t
+ */
+cuvsError_t cuvsRMMEnablePoolMemoryResource(int initial_pool_size_percent,
```
cjnolet, line 139: RMM's APIs also allow for a managed pool (a memory pool in managed memory) by accepting a boolean argument `managed`. I think we should add that argument here too.

cpp/include/cuvs/core/c_api.h
```diff
@@ -127,6 +127,24 @@ cuvsError_t cuvsRMMAlloc(cuvsResources_t res, void** ptr, size_t bytes);
...
+                                            int max_pool_size_percent);
+/**
+ * @brief Resets the memory resource to use the default
+ * memory resource (cuda_memory_resource)
+ * @return cuvsError_t
+ */
+cuvsError_t cuvsRMMResetMemoryResource();
```
cjnolet, line 146: The verb always comes att the end of our other C apis. `cuvsRMMMemoryResourceRest()` would be more consistent with current naming.

cpp/include/cuvs/core/c_api.h
cjnolet: `cuvsRMMPoolMemoryResourceEnable` would be more consistent with our C-API naming elsewhere.

cpp/include/cuvs/core/c_api.h
```diff
@@ -127,6 +127,24 @@ cuvsError_t cuvsRMMAlloc(cuvsResources_t res, void** ptr, size_t bytes);
  */
 cuvsError_t cuvsRMMFree(cuvsResources_t res, void* ptr, size_t bytes);
 
+/**
+ * @brief Switches the memory resource to use the pool memory resource
```
cjnolet, line 131:
```suggestion
 * @brief Switches the working memory resource to use the RMM pool memory resource, which will bypass unnecessary synchronizations by allocating a chunk of device memory up front and carving that up for temporary memory allocations within algorithms. Be aware that this function will change the memory resource for the whole process and the new memory resource will be used until explicitly changed.
```

cjnolet (changes requested):
Thanks for adding this @ajit283. A few minor things, but overall this is an important feature for supporting the extended API wrappers.

cpp/include/cuvs/core/c_api.h
ajit283: > RMM's APIs also allow for a managed pool (a memory pool in managed memory) by accepting a boolean argument `managed`. I think we should add that argument here too.

I cannot find this argument in the [rmm documentation](https://docs.rapids.ai/api/rmm/stable/librmm_docs/memory_resources/#classrmm_1_1mr_1_1pool__memory__resource), but there is a seperate [managed_memory_resource](https://docs.rapids.ai/api/rmm/stable/librmm_docs/memory_resources/#classrmm_1_1mr_1_1managed__memory__resource). Should I add this as well?

cjnolet: You are right, they changed the APIs more recently so they no longer include the “managed” argument in the pool resource API. However, they can get away with this in the c++ layer since the memory resources are compostable… we can’t get away with this in the C layer, unfortunately, and so I’d prefer to still provide the “managed” argument and smartly wrap a ManagedMemoryResource around the pool memory resource when it is set to true.

cjnolet (approved):
Changes look great, thanks @ajit283!

PR #297, opened 2024-08-19
Remove NumPy <2 pin

seberg:
This PR removes the NumPy<2 pin which is expected to work for
RAPIDS projects once CuPy 13.3.0 is released (CuPy 13.2.0 had
some issues preventing the use with NumPy 2).


python/cuvs/pyproject.toml
```diff
@@ -32,7 +32,7 @@ license = { text = "Apache 2.0" }
 requires-python = ">=3.9"
 dependencies = [
     "cuda-python",
-    "numpy>=1.23,<2.0a0",
+    "numpy>=1.23,<3.0a0",
```
jakirkham, line 35: Looks like we require NumPy here and the library does use it. However not seeing a dependency in the Conda recipe. Could you please add this same constraint there?

jakirkham (commented):
Thanks Sebastian! 🙏

Had a suggestion below

python/cuvs/pyproject.toml
seberg: Done, please have a look.  Will mark as ready for review as well.

jakirkham: Looks good. Thanks! 🙏

conda/recipes/cuvs/meta.yaml
```diff
@@ -69,6 +69,7 @@ requirements:
     - libcuvs {{ version }}
     - python x.x
     - rmm ={{ minor_version }}
+    - numpy >=1.23,<3.0a0
```
jakirkham, line 72: Somewhat tangential, but see `cuda-python` in the wheel as well. So think we can add it here too

Not seeing any other differences

```suggestion
    - cuda-python
    - numpy >=1.23,<3.0a0
```

jakirkham:
Pushed a small change to line up the dependencies across conda and wheels. Hope that is ok

As this depends on RAFT, CI may fail until we get PR ( https://github.com/rapidsai/raft/pull/2414 ) through. We can restart after that is done

jameslamb:
Some jobs were failing with what looked like temporary network errors, so I restarted CI here.

jakirkham:
Restarting CI now that RAFT has been rebuilt with NumPy 2 support: https://github.com/rapidsai/raft/pull/2414

seberg:
Nice, seems like this is passing now and using NumPy 2.0.1 (I also briefly scanned the code to see if the most likely NumPy 2.1 change can't affect it -- related to `np.full/copyto` with python integers).

jameslamb (approved):
Looks good, I double-checked that there's no need for a build dependency on `numpy` here, that the `cuda-python` dependency added is necessary, and that we're seeing `numpy` 2.x in tests.

jakirkham:
Thanks Sebastian and James! 🙏

PR #305, opened 2024-08-27
Add `managed` option for RMM Pool memory resource to C API

ajit283:


ajit283:
Hi @cjnolet I think this implementation using [`make_owning_wrapper`](https://docs.rapids.ai/api/rmm/stable/librmm_docs/memory_resources/#classrmm_1_1mr_1_1owning__wrapper) is more extensible and doesn't throw an error when the current memory resource is not a `cuda_memory_resource`

cpp/include/cuvs/core/c_api.h
```diff
@@ -141,7 +141,8 @@ cuvsError_t cuvsRMMFree(cuvsResources_t res, void* ptr, size_t bytes);
  * @return cuvsError_t
  */
 cuvsError_t cuvsRMMPoolMemoryResourceEnable(int initial_pool_size_percent,
-                                            int max_pool_size_percent);
+                                            int max_pool_size_percent,
+                                            int managed);
```
benfred, line 145: I think this should be a bool - and we should also add this managed parameter to the docstring above

```suggestion
                                            bool managed);
```

cpp/src/core/c_api.cpp
```diff
@@ -83,10 +86,14 @@ extern "C" cuvsError_t cuvsRMMFree(cuvsResources_t res, void* ptr, size_t bytes)
...
-thread_local std::unique_ptr<rmm::mr::pool_memory_resource<rmm::mr::cuda_memory_resource>> pool_mr;
+thread_local std::shared_ptr<
+  rmm::mr::owning_wrapper<rmm::mr::pool_memory_resource<rmm::mr::device_memory_resource>,
+                          rmm::mr::device_memory_resource>>
+  pool_mr;
 
 extern "C" cuvsError_t cuvsRMMPoolMemoryResourceEnable(int initial_pool_size_percent,
-                                                       int max_pool_size_percent)
+                                                       int max_pool_size_percent,
+                                                       int managed)
```
benfred, line 96:
```suggestion
                                                       bool managed)
```

cpp/src/core/c_api.cpp
```diff
@@ -83,10 +86,14 @@ extern "C" cuvsError_t cuvsRMMFree(cuvsResources_t res, void* ptr, size_t bytes)
   });
 }
 
-thread_local std::unique_ptr<rmm::mr::pool_memory_resource<rmm::mr::cuda_memory_resource>> pool_mr;
+thread_local std::shared_ptr<
```
benfred, line 89: I'm curious as to why this needs to be a shared_ptr instead of a unique_ptr like it was before?

cpp/test/core/c_api.c
```diff
@@ -33,34 +33,49 @@ int main()
...
+  if (error == CUVS_ERROR) { exit(EXIT_FAILURE); }
 
   // Reset pool memory resource
-  cuvsError_t reset_error = cuvsRMMMemoryResourceReset();
-  if (reset_error == CUVS_ERROR) { exit(EXIT_FAILURE); }
+  error = cuvsRMMMemoryResourceReset();
+  if (error == CUVS_ERROR) { exit(EXIT_FAILURE); }
+
+  // Enable pool memory resource (managed)
+  error = cuvsRMMPoolMemoryResourceEnable(10, 100, 1);
```
benfred, line 61:
```suggestion
  error = cuvsRMMPoolMemoryResourceEnable(10, 100, true);
```

cpp/test/core/c_api.c
```diff
@@ -33,34 +33,49 @@ int main()
...
   // Free memory
-  cuvsError_t free_error = cuvsRMMFree(res, ptr, bytes);
-  if (free_error == CUVS_ERROR) { exit(EXIT_FAILURE); }
+  error = cuvsRMMFree(res, ptr, bytes);
+  if (error == CUVS_ERROR) { exit(EXIT_FAILURE); }
 
   // Enable pool memory resource
-  cuvsError_t pool_error = cuvsRMMPoolMemoryResourceEnable(10, 100);
-  if (pool_error == CUVS_ERROR) { exit(EXIT_FAILURE); }
+  error = cuvsRMMPoolMemoryResourceEnable(10, 100, 0);
```
benfred, line 45:
```suggestion
  error = cuvsRMMPoolMemoryResourceEnable(10, 100, false);
```

benfred (commented):
thanks for adding this! 

This mostly looks good to me, left some more comments below:

cpp/src/core/c_api.cpp
ajit283: `rmm::mr::make_owning_wrapper` returns a `shared_ptr`. 
We could call the constructor directly to get a `unique_ptr` but not sure it's worth the extra complexity:
```
    using pool_wrapper =
      rmm::mr::owning_wrapper<rmm::mr::pool_memory_resource<rmm::mr::device_memory_resource>,
                              rmm::mr::device_memory_resource>;
    pool_mr = std::make_unique<pool_wrapper>(pool_wrapper(std::make_tuple(std::move(mr)),
                                                          std::forward<int>(initial_size),
                                                          std::forward<int>(max_size)));
```

benfred: makes sense - thanks for the explanation!

benfred (approved):

cpp/include/cuvs/core/c_api.h
```diff
@@ -17,6 +17,7 @@
 #pragma once
 
 #include <cuda_runtime.h>
+#include <stdbool.h>
```
bdice, line 20: I'm not a C wizard -- but is this include really needed? It sounds like it's a compatibility shim for super old-school C code. https://stackoverflow.com/questions/47374356/why-use-stdbool-h-instead-of-bool

I'd be happy to learn more -- but this might have just been an include automatically added by an IDE?

cpp/src/core/c_api.cpp
```diff
@@ -95,10 +102,22 @@ extern "C" cuvsError_t cuvsRMMPoolMemoryResourceEnable(int initial_pool_size_per
...
-    pool_mr = std::make_unique<rmm::mr::pool_memory_resource<rmm::mr::cuda_memory_resource>>(
-      cuda_mr_casted, initial_size, max_size);
+
+    auto mr = std::shared_ptr<rmm::mr::device_memory_resource>();
+    if (managed) {
+      mr = std::static_pointer_cast<rmm::mr::device_memory_resource>(
+        std::make_shared<rmm::mr::managed_memory_resource>());
+    } else {
+      mr = std::static_pointer_cast<rmm::mr::device_memory_resource>(
+        std::make_shared<rmm::mr::cuda_memory_resource>());
```
bdice, line 115: Should we use an async memory resource by default?

bdice (approved):
I left two nonblocking comments. This PR did not require a packaging/CI review (which was requested). This PR had some complicated merge history and so it may have needed a review from some changes that were ultimately not in the final changeset. In the future you can try to deselect the packaging/CI review groups. If they are required, you won't be able to deselect them. If you can deselect them, the review wasn't really needed.

PR #309, opened 2024-08-29
Use raft current resource wrappers

### LINKED ISSUE #308: [FEA] Use RAFT get/set_current_device_resource() wrappers in cuVS
To reduce thrash due to RMM refactoring, RAFT is adding wrappers around RMM `rmm::get_current_device_resource()`, `rmm::set_current_device_resource()` and aliases for RMM `resource_ref` types.  

Using these will allow us to change them in a single location in RAFT without having to touch every file that uses memory resources.

**Additional context**
See https://github.com/rapidsai/raft/issues/2423
and
https://github.com/rapidsai/raft/pull/2424

harrism:
Closes #308
Depends on rapidsai/raft#2424

### DIFF
--- a/examples/cpp/src/cagra_example.cu
+++ b/examples/cpp/src/cagra_example.cu
@@ -17,11 +17,11 @@
 #include <cstdint>
 #include <raft/core/device_mdarray.hpp>
 #include <raft/core/device_resources.hpp>
+#include <raft/core/resource/device_memory_resource.hpp>
 #include <raft/random/make_blobs.cuh>
 
 #include <cuvs/neighbors/cagra.hpp>
 
-#include <rmm/mr/device/device_memory_resource.hpp>
 #include <rmm/mr/device/pool_memory_resource.hpp>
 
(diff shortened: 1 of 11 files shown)


examples/cpp/src/cagra_example.cu
achirkin, line 1: I'm sorry, but this looks like pushing an RMM problem onto an unrelated 3-rd project. Why raft? With the same effect, you could create a new project `rmm-legacy` with its own `device_memory_resource.hpp` and refer to that instead.

achirkin (changes requested):

examples/cpp/src/cagra_example.cu
harrism: I'm attempting to shield RAPIDS from upcoming churn in RMM.

The real problem is that RAFT, cuML and cuGRAPH have painfully long compilation time whenever RMM changes. It takes me a long time to test all of RAPIDS for minor changes in RMM.

PR #315, opened 2024-09-06
Euclidean distance example

abner-ma:
Add euclidean distance calculation example in c.

### DIFF
--- a/examples/c/src/L2_c_example.c
+++ b/examples/c/src/L2_c_example.c
@@ -1,0 +1,105 @@
+#include <cuvs/core/c_api.h>
+#include <cuvs/distance/pairwise_distance.h>
+#include <cuvs/neighbors/cagra.h>
+
+#include <dlpack/dlpack.h>
+
+#include <cuda_runtime.h>
+#include <stdio.h>
+#include <stdlib.h>
+
+#define DIM 4
+#define N_ROWS 1
+
+float PointA[N_ROWS][DIM] = {1.0,2.0,3.0,4.0};
+float PointB[N_ROWS][DIM] = {2.0,3.0,4.0,5.0};
+
+void outputVector(float * Vec) {
+  printf("Vector is ");
+  for (int i = 0; i < DIM; ++i){
+    printf(" %f",Vec[i]);
+  }
+  printf("\n");
+}
+
+cuvsResources_t res;
+
+int initResources(void){
+  // Create a cuvsResources_t object
+  cuvsResourcesCreate(&res);
+  return 0;
+}
+
+int freeResources(void){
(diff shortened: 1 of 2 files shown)


examples/c/src/L2_c_example.c
```diff
@@ -0,0 +1,105 @@
+#include <cuvs/core/c_api.h>
```
cjnolet, line 1: Need the copyright here. You can copy this over from other files.

examples/c/src/L2_c_example.c
```diff
@@ -0,0 +1,105 @@
...
+  x_tensor->dl_tensor.ndim = 2;
+  x_tensor->dl_tensor.dtype.code = kDLFloat;
+  x_tensor->dl_tensor.dtype.bits = 32;
+  x_tensor->dl_tensor.dtype.lanes = 1;
+  x_tensor->dl_tensor.shape = x_shape;
+  x_tensor->dl_tensor.strides = NULL;
+}
+/* Ended by AICoder, pid:96cd0bcb6ad27b1149870b5f20ef16170f62a328 */
+
+int calcL2Distance(int64_t n_cols,float x[], float y[], float *ret) {
```
cjnolet, line 51: Can you provide some docs in this file, that briefly describe what the different steps are doing? Bonus points for function-level descriptions also (though not required).

examples/c/src/L2_c_example.c
```diff
@@ -0,0 +1,105 @@
...
+    printf(" %f",Vec[i]);
+  }
+  printf("\n");
+}
+
+cuvsResources_t res;
+
+int initResources(void){
+  // Create a cuvsResources_t object
+  cuvsResourcesCreate(&res);
```
cjnolet, line 29: Since these are one-liners anyways, it reduces the size of the code and clarity of the example to keep these inline in the example. Another small nitpick: we have intentionally named the C API functions consistently- `cuvsNounVerb` and changing that to `verbNoun` breaks this consistently. I'd keep `initializeTensor` below but rename it to `tensor_initialize` just to separate the casing from the C API. I'd also rename `calcL2Distance` to `l2_distance_calc`

cjnolet (changes requested):
Thanks for the contribution @abner-ma! Just a couple brief and minor things. Overall, contributions like this are very welcome!

examples/c/src/L2_c_example.c
```diff
@@ -0,0 +1,123 @@
+/*
+ * Copyright (c) 2022-2024, NVIDIA CORPORATION.
```
cjnolet, line 2: Please only use the current year for new files.

cjnolet (approved):
Thanks for your contribution!

Issue #277, opened 2024-08-06
Brute force knn tile size selection


tfeher:
**Describe the bug**

Heuristics incorrectly selects tiled execution of brute force-knn when the output tile could still fit the memory. This makes knn search slower than torch matmul + topk.

** Additional context **
 
https://github.com/rapidsai/cuvs/blob/72154b0b806c106300b52870f6113fdda3f87f0b/cpp/src/neighbors/detail/faiss_distance_utils.h#L50

**Steps/Code to reproduce bug**

Run brute force vector search using input with 1Mx128 input matrix and small number of queries. (The example below uses pylibraft, python wrappers, which currently has the same code as cuvs).

```python
import rmm
mr = rmm.mr.PoolMemoryResource(rmm.mr.CudaMemoryResource(), initial_pool_size=2**30)
rmm.mr.set_current_device_resource(mr)
import torch
import numpy as np
import pylibraft
from pylibraft.common import Handle
from pylibraft.neighbors.brute_force import knn
import cupy as cp
import time
device = torch.device("cuda")

class BenchmarkTimer:
    """Provides a context manager that runs a code block `reps` times
    and records results to the instance variable `timings`. Use like:
    .. code-block:: python
        timer = BenchmarkTimer(rep=5)
        for _ in timer.benchmark_runs():
            ... do something ...
        print(np.min(timer.timings))

        This class is part of the rapids/cuml benchmark suite
    """

    def __init__(self, reps=1, warmup=0):
        self.warmup = warmup
        self.reps = reps
        self.timings = []

    def benchmark_runs(self):
        for r in range(self.reps + self.warmup):
            t0 = time.time()
            yield r
            t1 = time.time()
            if r >= self.warmup:
                self.timings.append(t1 - t0)

rows = 1000000
cols = 128
n_queries = 8
k = 10
dataset = torch.randn(rows, cols, device=device)
queries = torch.randn(n_queries, cols, device=device)
dataset_cp = cp.asarray(dataset)
queries_cp = cp.asarray(queries)
timer = BenchmarkTimer(reps=100, warmup=5)
for rep in timer.benchmark_runs():
    distance = torch.matmul(queries, dataset.T)
    distances, indices = torch.topk(distance, k, dim=1, largest=True)

timings = np.asarray(timer.timings)
avg_time = timings.mean() * 1000
std_time = timings.std() * 1000
print("Average search time: {0:7.3f} +/- {1:7.3} ms".format(avg_time, std_time))

[...]

PR #316, opened 2024-09-06
Brute force knn tile size heuristic

mfoerste4:
This PR modifies the tile size heuristic for brute force knn as mentioned in (#277). 

It also removes some unneeded cuda calls to save a couple of microseconds which might be relevant when running smaller batches.

CC @tfeher


cpp/src/neighbors/detail/knn_brute_force.cuh
```diff
@@ -81,12 +81,13 @@ void tiled_brute_force_knn(const raft::resources& handle,
...
-  auto stream        = raft::resource::get_cuda_stream(handle);
-  auto device_memory = raft::resource::get_workspace_resource(handle);
-  auto total_mem     = rmm::available_device_memory().second;
+  size_t tile_rows = 0;
+  size_t tile_cols = 0;
+  auto stream      = raft::resource::get_cuda_stream(handle);
 
+  // total memory is not relevant in the heuristic for data below 512 MB
+  auto total_mem =
+    (sizeof(DistanceT) * m * n < 1 << 29) ? (1ul << 36) : rmm::available_device_memory().second;
```
benfred, line 90: Rather than tweak the total memory being passed to `chooseTileSize` - can we just adjust the chooseTileSize function instead to handle smaller inputs properly?  This will consolidate all the logic into a single function, rather than requiring some of the logic to be handled by the caller of chooseTileSize

cpp/src/neighbors/detail/knn_brute_force.cuh
```diff
@@ -356,27 +357,26 @@ void brute_force_knn_impl(
 
   ASSERT(input.size() == sizes.size(), "input and sizes vectors should be the same size");
 
-  std::vector<IdxType>* id_ranges;
-  if (translations == nullptr) {
+  std::vector<IdxType> id_ranges;
```
benfred, line 360: longer term - we can probably remove the  code that handles translations entirely. Its not being used in the public api anymore, and is just left over from the RAFT version. (doesn't need to change in this PR though)

benfred (changes requested):
thanks for fixing this @mfoerste4 !

benfred (approved):
lgtm - thanks @mfoerste4 !

tfeher (approved):
Thanks Malte for the updates, LGTM!

PR #321, opened 2024-09-10
Refactor dependencies.yaml to use depends-on pattern.

bdice:
This refactors cuvs's `dependencies.yaml` to follow a convention we've established in other repos. For RAPIDS packages we declare a `depends_on_X` dependency list, and use that rather than specifying packages multiple times. This enables more reuse and easier automation for version migrations.


python/cuvs/pyproject.toml
```diff
@@ -38,6 +38,7 @@ dependencies = [
     "nvidia-cusolver",
     "nvidia-cusparse",
     "pylibraft==24.10.*,>=0.0.0a0",
+    "rmm==24.10.*,>=0.0.0a0",
```
jameslamb, line 41:
```suggestion
```

I'm not sure about this... it seems to me that `cuvs` doesn't have a direct runtime dependency on the `rmm` Python package.

```shell
git grep -i rmm python/
```

If this wasn't intentional, it should be removed.

jameslamb (approved):
I love this pattern generally. Left one question for your consideration, I don't need to review again after you've addressed it.

python/cuvs/pyproject.toml
bdice: Agreed. I'll clean this up.

bdice: I ended up doing a partial cleanup in this PR and a partial cleanup in #322, to avoid merge conflicts.

jameslamb: it looks great, thank you for that

jameslamb (approved):
re-approving. Really glad we could cut some unnecessary build dependencies here!

PR #322, opened 2024-09-10
Simplify libcuvs conda recipe.

bdice:
This PR refactors the `libcuvs` conda recipe to use `libraft-headers-only` instead of `libraft`.

I removed dependencies on `rmm` from the `cuvs` build -- it is not imported.

I included a couple small fixes in CMake and the `cuvs` recipe.


conda/recipes/libcuvs/meta.yaml
```diff
@@ -129,7 +129,8 @@ outputs:
         - ninja
         - {{ stdlib("c") }}
       host:
-        - libraft ={{ minor_version }}
+        - librmm ={{ minor_version }}
```
bdice, line 132: We need `librmm` here because we `#include` it in libcuvs.

conda/recipes/cuvs/meta.yaml
```diff
@@ -54,10 +54,8 @@ requirements:
     - pylibraft {{ minor_version }}
     - libcuvs {{ version }}
     - python x.x
-    - rmm ={{ minor_version }}
```
bdice, line 57: `rmm` (Python) is not used directly in `cuvs` Python code. It is not needed at build or run time.

conda/recipes/libcuvs/meta.yaml
```diff
@@ -63,7 +63,8 @@ outputs:
...
+        - libraft-headers-only ={{ minor_version }}
```
bdice, line 67: We can migrate `libraft` to `libraft-headers-only` because there is no longer a requirement for `libraft` shared objects. This reduces the total dependency tree size for `libcuvs` because users do not need to download `libraft` binaries.

conda/recipes/libcuvs/meta.yaml
```diff
@@ -248,6 +248,9 @@ outputs:
         - {{ compiler('cuda') }}
         - cuda-cudart-dev
         - libcublas-dev
+        - libcurand-dev
```
bdice, line 251: These changes align `libcuvs-examples` with the `-tests` and other packages above. These are needed to build the examples, now that we're removing `libraft`. Previously `libraft` was supplying the CUDA math library dependencies.

conda/recipes/libcuvs/meta.yaml
cjnolet: I understand the change to remove libraft binaries, but I’d prefer to use libraft-headers instead of libraft-headers-only. Users of raft should be able to get the math libs transitively so they aren’t stuck having to redefine those all in the recipe.

bdice: cuVS links directly to the CUDA math libraries so we should not remove them here: https://github.com/rapidsai/cuvs/blob/c40abaeb0e9bfa3f7e13db35cc30105574412952/cpp/CMakeLists.txt#L453-L459

I can switch to `libraft-headers` if we are still using math libraries _through_ RAFT (I don't know whether that is true) but I don't know if the header-only pieces of libraft are using math libraries.

cjnolet: Usage of all math libs in cuvs are dome through RAFT APIs. I suspect there might be 1 or 2 places in legacy code that might still be calling math libs directly, but those ultimately will be updated to use raft. RAFT intentionally provides abstractions around the math libs so that users have a single cohesive API available for writing algorithms. Does that make sense?

bdice:
```suggestion
        - libraft-headers ={{ minor_version }}
```

conda/recipes/libcuvs/meta.yaml
```diff
@@ -129,7 +129,8 @@ outputs:
         - ninja
         - {{ stdlib("c") }}
       host:
-        - libraft ={{ minor_version }}
+        - librmm ={{ minor_version }}
+        - libraft-headers-only ={{ minor_version }}
```
bdice, line 133:
```suggestion
        - libraft-headers ={{ minor_version }}
```

conda/recipes/libcuvs/meta.yaml
```diff
@@ -195,7 +195,8 @@ outputs:
...
```
bdice, line 199:
```suggestion
        - libraft-headers ={{ minor_version }}
```

conda/recipes/libcuvs/meta.yaml
```diff
@@ -263,17 +266,27 @@ outputs:
...
```
bdice, line 270:
```suggestion
        - libraft-headers ={{ minor_version }}
```

conda/recipes/libcuvs/meta.yaml
bdice: Yes, I think that makes sense. Changes applied.

cjnolet: Thanks!

cjnolet (approved):
LGTM. Thanks for these changes Bradley!

PR #323, opened 2024-09-11
Add function for calculating the mutual_reachability_graph

benfred:


cpp/include/cuvs/neighbors/reachability.hpp
```diff
@@ -0,0 +1,79 @@
...
+ * @param[in] alpha weight applied when internal distance is chosen for
+ *            mutual reachability (value of 1.0 disables the weighting)
+ */
+void mutual_reachability_graph(
+  const raft::resources& handle,
+  raft::device_matrix_view<const float, int64_t, raft::row_major> X,
+  int min_samples,
+  raft::device_vector_view<int64_t> indptr,
+  raft::device_vector_view<float> core_dists,
+  raft::sparse::COO<float, int64_t>& out,
```
cjnolet, line 73: Small nitpick- this sparse::COO class is going away completely and it's really not designed well. Can we swap this out for the `raft::sparsity_owning_coo_matrix`? I think we can do a view since we already know the size, right?

cjnolet: It's also okay if the answer is "time is running out, let's change that later"

benfred: One challenge here is that we'd have to add support in RAFT for  `device_coo_matrix_view` / `device_sparsity_owning_coo_matrix` to some sparse algorithms like `raft::sparse::linalg::symmetrize`  and`raft::sparse::convert::sorted_coo_to_csr` - which only accept the `raft::sparse::COO` class.  This shouldn't be too hard to do though -

cpp/include/cuvs/selection/select_k.hpp
```diff
@@ -87,6 +87,16 @@ void select_k(
...
 
+void select_k(raft::resources const& handle,
+              raft::device_matrix_view<const float, int64_t, raft::row_major> in_val,
+              std::optional<raft::device_matrix_view<const int, int64_t, raft::row_major>> in_idx,
+              raft::device_matrix_view<float, int64_t, raft::row_major> out_val,
+              raft::device_matrix_view<int, int64_t, raft::row_major> out_idx,
+              bool select_min,
+              bool sorted                                                       = false,
+              SelectAlgo algo                                                   = SelectAlgo::kAuto,
+              std::optional<raft::device_vector_view<const int, int64_t>> len_i = std::nullopt);
```
cjnolet, line 98: It's saddening to see having to add this in here just for `int` type. Can you create an issue or maybe even a larger tracker issue with some follow-up tasks and add consolidating these type instantiations to that? Similar to the other types, it would be great if we could establish a single set of common integral types and maybe even consolidate float/double just into float.

cpp/src/neighbors/detail/knn_brute_force.cuh
```diff
@@ -78,7 +81,8 @@ void tiled_brute_force_knn(const raft::resources& handle,
                            size_t max_col_tile_size                  = 0,
                            const DistanceT* precomputed_index_norms  = nullptr,
                            const DistanceT* precomputed_search_norms = nullptr,
-                           const uint32_t* filter_bitmap             = nullptr)
+                           const uint32_t* filter_bitmap             = nullptr,
+                           DistanceEpilogue distance_epilogue        = raft::identity_op())
```
cjnolet, line 85: Unfortunate we have to instantiate bfknn twice now :-( but great that we don't have to expose the eilogue through the public APIs. Hopefully at some point soon we'll establish a good way to specify these (maybe as JIT compiled functions) through the public APIs where we don't need to instanaite all the kernels end to end for each one.

cjnolet (approved):
Approving because my only comment is asking to create an issue or tracker issue for consolidating templates.

cpp/include/cuvs/neighbors/reachability.hpp
cjnolet: Underneath the public API, we can just use the arrays / existing types. But I’d like to get to a point where we are using the new types at least for new public APIs. Eventually we need to scrape through all the sparse APIs and use the new types everywhere. However new APIs could at least use the new types in the meantime.

Can you create an issue for this? I think we are running out of time to do it for 24.10.

benfred: Created an issue here https://github.com/rapidsai/cuvs/issues/369

cpp/include/cuvs/selection/select_k.hpp
benfred: I've added an issue here https://github.com/rapidsai/cuvs/issues/370 for consolidating the template params -

PR #324, opened 2024-09-16
CAGRA - runtime dispatch of distance functions

achirkin:
Move the TEAM_SIZE and DATASET_BLOCK_DIM template parameters of CAGRA search kernels (single_cta and multi_cta versions) to runtime by introducing a switch-case statement inside the compute distance components (hence the name: runtime dispatch of the distance implementation). As a result:
  - 4x less search kernel instances against `branch-24.10`, but each instance is slightly larger
  - There's a small performance penalty due to larger kernel code (and the register usage is bounded by the most register-expensive branch)
  - [implementation limitation] only one `DATASET_BLOCK_DIM` per `TEAM_SIZE` is possible, i.e. the team size defines both (though this is how it currently is anyway).


achirkin:
This is an alternative to #296 for reducing the binary size with minimal changes to the code. Here's how they compare:

### enh-cagra-runtime-distance-dispatch (#324)
performance: up to 12% QPS slowdown on deep-100M and 5% QPS slowdown on wiki-all
libcuvs.so size: [791](https://downloads.rapids.ai/ci/cuvs/branch/branch-24.10/a6b71d7/cuda12_x86_64.compile_lib.html) -> [657](https://downloads.rapids.ai/ci/cuvs/pull-request/324/1da5d9d/cuda12_x86_64.compile_lib.html) MB
parallel build time in CI (NB: caching is not disabled):
    - total: 709 seconds
    - neighbors/cagra_search_xxx: 9 min
    - neighbors/detail/cagra_xxx (kernels): 2-3 mins

### enh-cagra-separable-compilation (#296)

performance: up to 13% QPS slowdown on deep-100M and up to 8% QPS slowdown on wiki-all (low-itopk CAGRA-Q cases being the worst, with others seemingly slightly better than (#326)
libcuvs.so size: [791](https://downloads.rapids.ai/ci/cuvs/branch/branch-24.10/a6b71d7/cuda12_x86_64.compile_lib.html) -> [624](https://downloads.rapids.ai/ci/cuvs/pull-request/296/267902e/cuda12_x86_64.compile_lib.html) MB
parallel build time in CI (NB: caching is not disabled):
    - total: 478 seconds
    - neighbors/cagra_search_xxx: 4 min
    - neighbors/detail/cagra_xxx (kernels): 1-3 mins
    - neighbors/detail/compute_distance_xxx: 50 sec (many of them)

achirkin:
To sum it up:
  - the QPS of this PR is actually slightly better due to some VPQ cases being particularly bad in #296
  - this PR reduces the compile size almost as good as #296, but kernels are still large; this means adding #215 on top of this PR will have larger impact on binary size, because it duplicates the single-cta kernel (perhaps 20-30MB?.. I don't know)
  - In contrast to #296, this PR **does not** reduce the compile time
  - This PR introduces minimal changes, which is much easier for review
  - #296 adds extra functionality (multi-kernel CAGRA-Q) and (:in my opinion:) makes it easier to do further changes (e.g. flexible code generation for instances and more team_size, dataset_block_dim combinations can be supported).

achirkin:
Performance update:
After a few optimizations for the worst cases #296, its performance is generally better than this PR on deep-100M dataset (dim = 96) and slightly worse on wiki-all dataset (dim = 768, only VPQ distance tested due to the index size). In the worst case, #296 leads to QPS drop of 6-7%, with plenty of cases actually getting a decent speedup (against both this PR and base-24.10). The caveat here is that some optimizations in #296 are potentially transferable to the base branch, so this benchmark is not entirely apples-to-apples anymore.

Latest benchmarks (includes k = 10 and k = 100 for both datasets):
https://docs.google.com/spreadsheets/d/191f1sYsAUwPidncV4xDzgNtOhKx9sp6jJFvOfpB-nxs

PR #325, opened 2024-09-17
Recommend `miniforge` for conda install

bdice:
Recommending `miniforge` for conda install in installation docs.

I also cleaned up a few other small pieces in this document.


bdice:
Style checks are blocked by #327.

cjnolet (approved):

PR #328, opened 2024-09-17
fix style checks on Python 3.12

jameslamb:
After merging https://github.com/rapidsai/ci-imgs/pull/188, style checks here started breaking like this:

```text
python/cuvs/cuvs/test/test_doctests.py:103:24: E231 missing whitespace after ':'
python/cuvs/cuvs/test/test_doctests.py:124:27: E231 missing whitespace after ':'
```

([build link](https://github.com/rapidsai/cuvs/actions/runs/10905609739/job/30264867507?pr=325))

Looks like both of those are false positives (we really do intend to not have a space after the `:` on those lines). This proposes ignoring them with `#noqa` comments.

### DIFF
--- a/python/cuvs/cuvs/test/test_doctests.py
+++ b/python/cuvs/cuvs/test/test_doctests.py
@@ -100,7 +100,7 @@ DOC_STRINGS.extend(_find_doctests_in_obj(cuvs.distance))
 
 def _test_name_from_docstring(docstring):
     filename = Path(docstring.filename).name.split(".")[0]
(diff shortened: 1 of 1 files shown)


jameslamb:
`check-style` is passing here: https://github.com/rapidsai/cuvs/actions/runs/10907091447/job/30269909264?pr=328

I'll ask for an admin merge.

jameslamb:
For those finding this from search.... the better fix was just to upgrade `flake8` to 7.1.1. A bug in earlier versions led to these false positives.

That was done here: #327

PR #341, opened 2024-09-20
remove NCCL pins in build and test environments

jameslamb:
## Description

Contributes to https://github.com/rapidsai/build-planning/issues/102

Some RAPIDS libraries are using `ncclCommSplit()`, which was introduced in `nccl==2.18.1.1`. This is part of a series of PRs across RAPIDS updating libraries' pins to `nccl>=2.18.1.1` to ensure they get a new-enough version that supports that.

`cuvs` doesn't have any *direct* uses of NCCL... it only uses it via raft. This PR proposes removing `cuvs`'s dependency pinnings on NCCL, in favor of just using whatever it gets transitively via raft.

### DIFF
--- a/conda/environments/all_cuda-118_arch-aarch64.yaml
+++ b/conda/environments/all_cuda-118_arch-aarch64.yaml
@@ -37,7 +37,6 @@ dependencies:
 - libcusparse=11.7.5.86
 - librmm==24.10.*,>=0.0.0a0
 - make
-- nccl>=2.18.1.1
 - ninja
 - numpy>=1.23,<3.0a0
 - numpydoc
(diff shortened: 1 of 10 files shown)


vyasr (changes requested):
Does cuvs use nccl directly, or only transitively via raft? I notice that it's not specified in the nccl recipes, and if I search the repo I don't see other references which makes me think it's just via raft. If so, we should just drop it here.

jameslamb:
> Does cuvs use nccl directly, or only transitively via raft? I notice that it's not specified in the nccl recipes, and if I search the repo I don't see other references which makes me think it's just via raft. If so, we should just drop it here.

I searched around and it seems like it's only used transitively through raft.

```shell
git grep -i nccl
```

Looked through the diff of #231 too, and I only see stuff like this:

```cpp
const raft::comms::nccl_clique& clique = handle.get_nccl_clique_handle();
```

vyasr:
Yeah that matches my expectation. Let's see if tests pass. If they do, please update the title and description accordingly and I can approve.

jameslamb:
Tests look happy! I've updated the title and description here.

Thanks for the nudge, removing these pins will make the next such update a little easier 😁

jakirkham:
Thanks all! 🙏

PR #346, opened 2024-09-24
CosineExpanded Metric for IVF-PQ (normalize inputs)

tarang-jain:


cpp/src/neighbors/ivf_common.cuh
```diff
@@ -254,7 +254,14 @@ void postprocess_distances(ScoreOutT* out,      // [n_queries, topk]
...
     } break;
-    case distance::DistanceType::CosineExpanded:
+    case distance::DistanceType::CosineExpanded: {
+      raft::linalg::unaryOp(
+        out,
+        in,
+        len,
+        raft::compose_op(raft::add_const_op<ScoreOutT>{1.0}, raft::cast_op<ScoreOutT>{}),
+        stream);
+    } break;
```
achirkin, line 264: I don't get it, why do we need `f(x) = x + 1` mapping here? I'd expect it to be the same as for inner product, or with extra diving by the query norm.

cpp/src/neighbors/ivf_pq/ivf_pq_build.cuh
```diff
@@ -1632,9 +1660,24 @@ void extend(raft::resources const& handle,
...
+      raft::make_device_vector_view<const T, internal_extents_t>(vec_batch.data(),
+                                                                 vec_batch.size() * index->dim()),
+      raft::make_device_vector_view<float, internal_extents_t>(float_vec_batch.data_handle(),
+                                                               vec_batch.size() * index->dim()),
+      [=] __device__(internal_extents_t idx, T i) { return utils::mapping<float>{}(i); });
+    if (index->metric() == cuvs::distance::DistanceType::CosineExpanded) {
+      raft::linalg::row_normalize(handle,
+                                  raft::make_const_mdspan(float_vec_batch.view()),
+                                  float_vec_batch.view(),
+                                  raft::linalg::NormType::L2Norm);
```
achirkin, line 1676: Does row_normalize work in-place?

cpp/src/neighbors/ivf_pq/ivf_pq_build.cuh
```diff
@@ -1632,9 +1660,24 @@ void extend(raft::resources const& handle,
...
     const auto& idx_batch = *idx_batches++;
+    auto float_vec_batch =
+      raft::make_device_matrix<float, internal_extents_t>(handle, vec_batch.size(), index->dim());
+    raft::linalg::map_offset(
+      handle,
+      raft::make_device_vector_view<const T, internal_extents_t>(vec_batch.data(),
+                                                                 vec_batch.size() * index->dim()),
+      raft::make_device_vector_view<float, internal_extents_t>(float_vec_batch.data_handle(),
+                                                               vec_batch.size() * index->dim()),
+      [=] __device__(internal_extents_t idx, T i) { return utils::mapping<float>{}(i); });
```
achirkin, line 1671: If you don't need offset as an argument, you can use just `linalg::map` in place of `linalg::map_offset`. In future, it would also be nice to fuse this with row_normalize in one op to avoid multiple passes through the data.

cpp/src/neighbors/ivf_pq/ivf_pq_build.cuh
```diff
@@ -1754,10 +1797,22 @@ auto build(raft::resources const& handle,
...
     cuvs::cluster::kmeans::balanced_params kmeans_params;
     kmeans_params.n_iters = params.kmeans_n_iters;
-    kmeans_params.metric  = static_cast<cuvs::distance::DistanceType>((int)index.metric());
+    if (index.metric() == distance::DistanceType::CosineExpanded) {
+      raft::print_device_vector(
+        "non_normalized_build", trainset.data_handle(), index.dim(), std::cout);
+      raft::linalg::row_normalize(
+        handle, trainset_const_view, trainset.view(), raft::linalg::NormType::L2Norm);
+      raft::print_device_vector("normalized_build", trainset.data_handle(), index.dim(), std::cout);
+      kmeans_params.metric = distance::DistanceType::InnerProduct;
```
achirkin, line 1806: If you fool kmeans into using inner product instead of cosine metric, it probably produces non-normalized cluster centers. This messes up with the cluster selection stage for the cosine metric and thus likely reduces the recall.
Maybe try to normalize cluster centers after you do the training, or adjust index extend and search code to take this into account?

cpp/src/neighbors/ivf_pq/ivf_pq_search.cuh
```diff
@@ -156,6 +172,35 @@ void select_clusters(raft::resources const& handle,
                      n_lists,
                      stream);
 
+  if (metric == distance::DistanceType::CosineExpanded) {
+    // TODO: store dataset norms in a different manner for the cosine metric to avoid the copy here
```
achirkin, line 176: I don't get it. What's the difference to the inner product here?

achirkin (changes requested):
Thanks for the PR!
Couple nitpicks and a more substantial question about the cluster center handling here.

cpp/src/neighbors/ivf_common.cuh
tarang-jain: x  is equal to the negative of the normalized inner product. We add one in the end to get the cosine distance as
`1 - (a.b) / ||a||||b||`

cpp/src/neighbors/ivf_pq/ivf_pq_build.cuh
tarang-jain: yeah it is working

cpp/src/neighbors/ivf_pq/ivf_pq_build.cuh
tarang-jain: The problem is that then cluster center + residual would certainly have a norm greater than 1 and not close to 1.  Since residuals are trained with L2 as the distance metric.

achirkin: The residuals shouldn't be a problem, but centers need to be normalized (or their norms accounted for) during search, so that correct clusters are selected for the fine search.

cpp/src/neighbors/ivf_pq/ivf_pq_build.cuh
```diff
@@ -1754,7 +1796,13 @@ auto build(raft::resources const& handle,
       cluster_centers, index.n_lists(), index.dim());
     cuvs::cluster::kmeans::balanced_params kmeans_params;
     kmeans_params.n_iters = params.kmeans_n_iters;
-    kmeans_params.metric  = static_cast<cuvs::distance::DistanceType>((int)index.metric());
+    if (index.metric() == distance::DistanceType::CosineExpanded) {
```
lowener, line 1799: `cuvs::cluster::kmeans_balanced::fit` Already supports Cosine metric, so there is no need to add normalization + switch to inner product.

cpp/src/neighbors/ivf_pq/ivf_pq_search.cuh
```diff
@@ -137,6 +141,18 @@ void select_clusters(raft::resources const& handle,
       alpha = -1.0;
       beta  = 0.0;
     } break;
+    case cuvs::distance::DistanceType::CosineExpanded: {
```
lowener, line 144: As @achirkin noted, the norms of the centers and of the queries should be accounted for when computing the cosine distance. Right now only the norms of the queries is used, and this can result in the wrong clusters getting selected.
In IVF-Flat: https://github.com/rapidsai/cuvs/blob/branch-24.10/cpp/src/neighbors/ivf_flat/ivf_flat_search.cuh#L166

lowener (changes requested):

cpp/src/neighbors/ivf_pq/ivf_pq_build.cuh
```diff
@@ -1576,13 +1579,33 @@ void extend(raft::resources const& handle,
...
       cuvs::cluster::kmeans::balanced_params kmeans_params;
-      kmeans_params.metric = static_cast<cuvs::distance::DistanceType>((int)index->metric());
-      cuvs::cluster::kmeans_balanced::predict(handle,
-                                              kmeans_params,
-                                              batch_data_view,
-                                              centers_view,
-                                              batch_labels_view,
-                                              utils::mapping<float>{});
+
+      if (index->metric() == cuvs::distance::DistanceType::CosineExpanded) {
```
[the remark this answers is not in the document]
tarang-jain: Yes I tried that. I also tried normalizing the cluster centers, but that does not give good recall. I get the best recall when I normalize the inputs and use inner product.

cpp/src/neighbors/ivf_pq/ivf_pq_search.cuh
tarang-jain: I tried to do that, but that gives poorer recall. Simply using inner product to select the clusters to probe gives better recall in the tests.

cpp/src/neighbors/ivf_pq/ivf_pq_search.cuh
```diff
@@ -137,6 +149,18 @@ void select_clusters(raft::resources const& handle,
...
+      alpha = -1.0;
+      beta  = 0.0;
+
+      auto float_queries_matrix_view =
+        raft::make_device_matrix_view<float, uint32_t>(float_queries, n_queries, dim_ext);
+
+      raft::linalg::row_normalize(handle,
```
achirkin, line 159: I feel like normalizing the queries twice shouldn't be necessary. As long as we use the `qc_distances` only for selecting the clusters for one query, its norm shouldn't affect search results at all. You normalize the rotated queries further below, that should be enough.

achirkin (changes requested):
Thanks @tarang-jain for the updates. I think this is a long-awaited feature, but it's not so urgent to squeeze it in 24.10. I'd suggest we take a bit more time to make sure it has good, well understood performance from day one in the main branch.
If we decide to push this to 24.12, it would be nice to run a few benchmarks to see how cosine metric fares against other metrics and against the cuVS main branch.

cpp/src/neighbors/ivf_pq/ivf_pq_search.cuh
```diff
@@ -363,8 +387,9 @@ void ivfpq_search_worker(raft::resources const& handle,
       // stores basediff (query[i] - center[i])
       precomp_data_count = index.rot_dim();
     } break;
+    case distance::DistanceType::CosineExpanded:
     case distance::DistanceType::InnerProduct: {
-      // stores two components (query[i] * center[i], query[i] * center[i])
+      // stores two components (query[i], query[i] * center[i])
```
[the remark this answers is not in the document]
tarang-jain: Yes, that is right. It was incorrect before.

cpp/src/neighbors/ivf_pq/ivf_pq_build.cuh
```diff
@@ -1632,9 +1635,24 @@ void extend(raft::resources const& handle,
...
     const auto& idx_batch = *idx_batches++;
+    auto float_vec_batch  = raft::make_device_mdarray<float, internal_extents_t>(
+      handle,
+      device_memory,
+      raft::make_extents<internal_extents_t>(vec_batch.size(), index->dim()));
+    raft::linalg::map(handle,
+                      float_vec_batch.view(),
+                      utils::mapping<float>{},
+                      raft::make_device_matrix_view<const T, internal_extents_t>(
+                        vec_batch.data(), vec_batch.size(), index->dim()));
```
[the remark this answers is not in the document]
tarang-jain: @achirkin for float datatype, we can normalize in place. We just need the extra memory for the float batch when the data type is uint8 or int8.

cpp/src/neighbors/ivf_pq/ivf_pq_search.cuh
tarang-jain: Done. We do not normalize when computing qc_distances.

cjnolet (approved):

lowener (approved):

PR #347, opened 2024-09-25
cuvs_bench plotting functions 

dantegd:
Builds on top of #279, will change from draft after that merge, or we can merge into one single PR if reviewers prefer so.


.gitignore
```diff
@@ -81,4 +81,5 @@ ivf_pq_index
 
 # cuvs_bench
 datasets/
-/*.json
\ No newline at end of file
+/*.json
+*.png
```
cjnolet, line 85: I don't think we want to do a blanket ignore on png, especialy since we have png images in our docs. I'm okay ignoring json files, since those are assumed to be generated.

python/cuvs_bench/cuvs_bench/plot/__main__.py
```diff
@@ -0,0 +1,617 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# This script is inspired by
+# 1: https://github.com/erikbern/ann-benchmarks/blob/main/plot.py
+# 2: https://github.com/erikbern/ann-benchmarks/blob/main/ann_benchmarks/plotting/utils.py  # noqa: E501
+# 3: https://github.com/erikbern/ann-benchmarks/blob/main/ann_benchmarks/plotting/metrics.py  # noqa: E501
+# Licence: https://github.com/erikbern/ann-benchmarks/blob/main/LICENSE
```
cjnolet, line 20: We should put this license in our repository and point to it.  We also need to move the `get_dataset` script over.

cjnolet (commented):
Leaving some initial comments for now. I'll work on porting over our benchmarks documentation. 

One thing I'd like to try and do for 24.12 is to gather some pareto plots for each of our algorithms on cloud GPUs and provide them in our docs for different data scales. Today, we usually show this info off in slide decks but providing it in our docs will reach a much wider audience (especially since we are rebanding to reach a wider audience than RAPIDS).

python/cuvs_bench/cuvs_bench/plot/__main__.py
```diff
@@ -17,7 +17,7 @@
 # 1: https://github.com/erikbern/ann-benchmarks/blob/main/plot.py
 # 2: https://github.com/erikbern/ann-benchmarks/blob/main/ann_benchmarks/plotting/utils.py  # noqa: E501
 # 3: https://github.com/erikbern/ann-benchmarks/blob/main/ann_benchmarks/plotting/metrics.py  # noqa: E501
-# Licence: https://github.com/erikbern/ann-benchmarks/blob/main/LICENSE
+# Licence: https://github.com/rapidsai/cuvs/blob/branch-24.10/thirdparty/LICENSES/LICENSE.ann-benchmark # noqa: E501
```
cjnolet, line 20:
```suggestion
# License: https://github.com/rapidsai/cuvs/blob/branch-24.10/thirdparty/LICENSES/LICENSE.ann-benchmark # noqa: E501
```

cjnolet (approved):

PR #348, opened 2024-09-25
Add multigpu kmeans fit function

benfred:
Changes to support using kmeans clustering inside of cuml, so we can transition cuml off of the RAFT kmeans code

* Add a multigpu kmeans fit function
* Adds instantiations for kmeans on int64_t indicies, which unfortunately also requires int64_t indices for the PW distance functions
* Add support for `double` precision kmeans


cpp/include/cuvs/cluster/kmeans_mg.hpp
```diff
@@ -0,0 +1,75 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#pragma once
+#include <cuvs/cluster/kmeans.hpp>
+#include <raft/core/device_mdspan.hpp>
+#include <raft/core/host_mdspan.hpp>
+#include <raft/core/resources.hpp>
+
+namespace cuvs::cluster::kmeans::mg {
```
cjnolet, line 22: One of the nice benefits to using `raft::resources` and further using a PIMPL design for the `raft::comms` is that we shouldn't be coupled to NCCL at all in the API layer. For this reason, I think we should expose a single unified API for both the single-gpu and multi-gpu kmeans and just route to the proper implementation based on the presence of the `raft::comms` object in the `raft::resources` object. Does that make sense?

cpp/src/cluster/kmeans_fit_double.cu
```diff
@@ -0,0 +1,45 @@
+/*
+ * Copyright (c) 2022-2023, NVIDIA CORPORATION.
```
cjnolet, line 2: New files should just use the current year.

cpp/include/cuvs/cluster/kmeans_mg.hpp
cjnolet: The other benefit to this is that we still expose the same symbols from the API side and can consolidate everything down (since the MNMG version really just uses primitives from the SG version, we don't need different instantiations for those).

cpp/CMakeLists.txt
```diff
@@ -344,6 +350,32 @@ add_library(
   src/distance/detail/pairwise_matrix/dispatch_russel_rao_float_float_float_int.cu
   src/distance/detail/pairwise_matrix/dispatch_russel_rao_half_float_float_int.cu
   src/distance/detail/pairwise_matrix/dispatch_russel_rao_double_double_double_int.cu
+  src/distance/detail/pairwise_matrix/dispatch_canberra_double_double_double_int64_t.cu
```
cjnolet, line 353: OH man this is so sad to see :-( I think we definitely should update all of our APIs to be int64_t (and maybe even instantiate the int32_t but have them all go through the same template instantiations in the end conditionally).

cpp/include/cuvs/cluster/kmeans.hpp
```diff
@@ -191,6 +191,56 @@ void fit(raft::resources const& handle,
...
+ *   clusters are reinitialized by choosing new centroids with
+ *   k-means++ algorithm.
+ *
+ * @code{.cpp}
+ *   #include <raft/core/resources.hpp>
+ *   #include <cuvs/cluster/kmeans.hpp>
+ *   using namespace  cuvs::cluster;
+ *   ...
+ *   raft::resources handle;
+ *    cuvs::cluster::kmeans::params params;
```
cjnolet, line 206: Minor nitpick but maybe this should be moved back 1 space.

cpp/include/cuvs/cluster/kmeans.hpp
```diff
@@ -515,6 +594,77 @@ void fit_predict(raft::resources const& handle,
...
+ * @brief Compute k-means clustering and predicts cluster index for each sample
+ * in the input.
+ *
+ * @code{.cpp}
+ *   #include <raft/core/resources.hpp>
+ *   #include <cuvs/cluster/kmeans.hpp>
+ *   using namespace  cuvs::cluster;
+ *   ...
+ *   raft::resources handle;
+ *    cuvs::cluster::kmeans::params params;
```
cjnolet, line 616: I think maybe in all of these, this should be moved back 1 space.

cpp/src/distance/detail/pairwise_matrix/dispatch_canberra_double_double_double_int64_t.cu
```diff
@@ -0,0 +1,55 @@
+/*
+ * Copyright (c) 2021-2023, NVIDIA CORPORATION.
```
cjnolet, line 2: New files should only have the current year

cpp/test/cluster/kmeans_mg.cu
```diff
@@ -0,0 +1,199 @@
+/*
+ * Copyright (c) 2022-2024, NVIDIA CORPORATION.
```
cjnolet, line 2: If it's going to take too much to get this into 24.10, I'm okay forgoing this just for this release so long as cuML mnmg (Dask) tests are running properly on multiple gpus. We should create an issue to come back tot his, though , and reference the issue in the code if you can.

cpp/include/cuvs/cluster/kmeans_mg.hpp
benfred: I added this in the last commit https://github.com/rapidsai/cuvs/pull/348/commits/d78dc8c2b9281b685de014b787f51c0b26f83d7e - I like how this simplifies the public api a bunch here =)

cpp/src/distance/detail/pairwise_matrix/dispatch_canberra_double_double_double_int64_t.cu
benfred: I've updated that script that generated these files to set the copyright year to 2024 only -

cpp/test/cluster/kmeans_mg.cu
benfred: fwiw - the dask kmeans in cuml test all are working for me with this change

```
ben@ben-Precision-7920-Tower:~/code/cuml$ pytest python/cuml/cuml/tests/dask/test_dask_kmeans.py 
========================================================================= test session starts =========================================================================
platform linux -- Python 3.12.6, pytest-7.4.4, pluggy-1.5.0
benchmark: 4.0.0 (defaults: timer=time.perf_counter disable_gc=False min_rounds=5 min_time=0.000005 max_time=1.0 calibration_precision=10 warmup=False warmup_iterations=100000)
rootdir: /home/ben/code/cuml/python/cuml
configfile: pyproject.toml
plugins: benchmark-4.0.0, xdist-3.6.1, hypothesis-6.112.1, cov-5.0.0, cases-3.8.5
collected 491 items                                                                                                                                                   

python/cuml/cuml/tests/dask/test_dask_kmeans.py .ss.ssssssssssssssssssssssssssssssssssssssssssssssssss.ss.ssssssssssssssssssssssssssssssssssssssssssssssssss.ss [ 22%]
.ssssssssssssssssssssssssssssssssssssssssssssssssss.ss.sssssssssssssssssssssssssssssssssssssssssssssssssss.....ssssssss.ssssssss.ssssssssssssssssssssssssssssss [ 54%]
ssssssssssssssssssssssssssssssss.ssssssss.ssssssss.ssssssssssssssssssssssssssssssssssssssssssssssssssssssssssssss.ss.ssssssssssssssssssssssssssssssssssssssssss [ 87%]
ssssssss.ss.ssssssssssssssssssssssssssssssssssssssssssssssssss                                                                                                  [100%]

================================================================== 22 passed, 469 skipped in 39.48s ===================================================================
```

I will create an issue for the test here

benfred: the dask kmeans tests work locally on my workstation - but seem to be failing in CI on the cuml PR =(
https://github.com/rapidsai/cuml/actions/runs/11100367838/job/30840143557?pr=6085

looking into this

benfred: I am reasonably sure that the last commit here fixes - re-running CI on the cuml pr to test this out

benfred: dask kmeans tests passing cuml ci now  https://github.com/rapidsai/cuml/actions/runs/11114409824/job/30882703741?pr=6085

cpp/CMakeLists.txt
```diff
@@ -341,6 +347,32 @@ add_library(
...
+  src/distance/detail/pairwise_matrix/dispatch_canberra_float_float_float_int64_t.cu
+  src/distance/detail/pairwise_matrix/dispatch_correlation_double_double_double_int64_t.cu
```
cjnolet, line 352: Question- do we need all of these to support int64_t or just the euclidean? Kmeans is inherently coupled to euclidean (and cosine, though that's not an option yet in our kmeans).

cpp/test/cluster/kmeans_mg.cu
cjnolet: I think we should consider eventually exposing a rudimentary Dask API in cuVS for kmeans (and ann/knn when that's ready) so that we can test this stuff right inside cuVS. I'll create an issue for this.

cpp/CMakeLists.txt
benfred: pushed out a change to limit to just l2expanded distance w/ int64_t indices here https://github.com/rapidsai/cuvs/pull/348/commits/63c11ce771f2b6b319d846e1e3c154b7baa6cdf8 - lmk what you think

cjnolet (approved):

PR #350, opened 2024-09-25
[Feat] Relative change with `bitset` API feature #2439 in raft

rhdong:


cpp/cmake/thirdparty/get_raft.cmake
```diff
@@ -61,8 +61,8 @@ endfunction()
 # To use a different RAFT locally, set the CMake variable
 # CPM_raft_SOURCE=/path/to/local/raft
 find_and_configure_raft(VERSION  ${RAFT_VERSION}.00
-        FORK                     ${RAFT_FORK}
-        PINNED_TAG               ${RAFT_PINNED_TAG}
+        FORK                     rhdong
```
lowener, line 64: Switch that back before merging

rhdong: Sure, just for running CI in advance. I will revert it after success.

lowener (approved):
LGTM

PR #358, opened 2024-09-27
Remove EXPLICIT_INSTANTIATE_ONLY macros

achirkin:
`CUVS_EXPLICIT_INSTANTIATE_ONLY` macro carried over from raft; it is not very useful in the fully compiled library, such as cuVS, hence is removed.


achirkin:
I've checked the build artifacts of [this branch](https://downloads.rapids.ai/ci/cuvs/pull-request/358/4cd81e7/) against [the latest state of branch-24.10](https://downloads.rapids.ai/ci/cuvs/branch/branch-24.10/c616a22/).
The cuda12_x86_x64 libcuvs.so file is 40 bytes larger in the PR version than the base-24.10, I'm not sure why. I've checked the contents with nm and diff tools; after stripping the symbol addresses, the outputs of nm are identical (hence no extra instances defined).
With that in mind, I am rather confident that the PR does not indeed affect any functionality or compiling behavior. @cjnolet , I'd suggest we have an option to merge this in 24.10 if you're feeling like doing so.

cjnolet:
@achirkin I missed your comment here for some reason. I'm not opposed to merging this for 24.10, so long as we can get it through CI in time. I slated it for 24.10 in the meantime.

cjnolet (approved):

PR #372, opened 2024-10-02
Put the ann-bench large_workspace_resource in managed memory

achirkin:
Setup the `raft::resource::large_workspace_resource` to an `rmm::mr::managed_memory_resource` for the benchmarks. 

This change was first introduced in https://github.com/rapidsai/raft/pull/2322 but somehow went missing in cuVS. It allows to run large (not fitting in GPU memory)  ANN problemsto completion at the cost to speed.


cjnolet:
I see some c++-level cusolver include build errors in the Python package.

tfeher (approved):
Thanks Artem, LGTM!

PR #382, opened 2024-10-03
Add a static library for cuvs

benfred:
For the cuml integration, we need to be able to statically link to cuvs in order build the python wheels. This change adds a static target that lets us do that

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -485,7 +485,7 @@ target_include_directories(
 )
 
 # ensure CUDA symbols aren't relocated to the middle of the debug build binaries
-target_link_options(cuvs_static PRIVATE "${CMAKE_CURRENT_BINARY_DIR}/fatbin.ld")
+target_link_options(cuvs_static PRIVATE $<HOST_LINK:${CMAKE_CURRENT_BINARY_DIR}/fatbin.ld>)
 
 target_include_directories(
   cuvs_static PUBLIC "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>"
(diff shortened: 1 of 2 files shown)


cpp/CMakeLists.txt
```diff
@@ -465,11 +520,17 @@ if(NOT BUILD_CPU_ONLY)
     PUBLIC rmm::rmm raft::raft ${CUVS_CTK_MATH_DEPENDENCIES}
     PRIVATE nvidia::cutlass::cutlass $<TARGET_NAME_IF_EXISTS:OpenMP::OpenMP_CXX> cuvs-cagra-search
   )
+
+  target_link_libraries(
+    cuvs_static
+    PUBLIC rmm::rmm raft::raft ${CUVS_CTK_MATH_DEPENDENCIES}
+    PRIVATE nvidia::cutlass::cutlass $<TARGET_NAME_IF_EXISTS:OpenMP::OpenMP_CXX>
```
achirkin, line 527: Shouldn't this target also link `cuvs-cagra-search`?

cpp/CMakeLists.txt
```diff
@@ -430,12 +430,67 @@ add_library(
...
+)
+
+target_include_directories(
+  cuvs_static
+  PUBLIC "$<BUILD_INTERFACE:${DLPACK_INCLUDE_DIR}>"
+  INTERFACE "$<INSTALL_INTERFACE:include>"
+)
+
+# ensure CUDA symbols aren't relocated to the middle of the debug build binaries
+target_link_options(cuvs_static PRIVATE "${CMAKE_CURRENT_BINARY_DIR}/fatbin.ld")
```
achirkin, line 488: If you get a linker error here due to `cuvs-cagra-search` component, consider replacing this same as for the shared lib target:
```suggestion
target_link_options(cuvs_static PRIVATE $<HOST_LINK:${CMAKE_CURRENT_BINARY_DIR}/fatbin.ld>)
```

cpp/CMakeLists.txt
benfred: yeah - I agree this should link cuvs-cagra-search (and tried in the previous commit) .   The error I was hitting was : 

```
  CMake Error in CMakeLists.txt:
    export called with target "cuvs_static" which requires target
    "cuvs-cagra-search" that is not in any export set.
```

https://github.com/rapidsai/cuvs/actions/runs/11155746120/job/31007185860

Will try your suggestion - thanks for this!

benfred: works great - thanks! applied in latest commit

achirkin: Do you think we still need the `cuvs-cagra-search` in the install export set? Theoretically, we shouldn't be exposing that as a library and nobody outside cuVS should need to link against it.

cjnolet (approved):

PR #388, opened 2024-10-03
Remove cuvs-cagra-search from cuvs_static link

benfred:
Seeing build errors in the cuml wheels, and the only difference afaict was adding the cuvs-cagra-search to link to the static library.

Removing this since cuml is the only consumer of the static library this release and doesn't use cagra (will revisit in 24.12)


benfred:
wheels are building with this change https://github.com/rapidsai/cuml/actions/runs/11166718328/job/31041370942?pr=6085

PR #400, opened 2024-10-07
Fix NVTX annotations

achirkin:
1) Replace the domain name from `raft` to `cuvs` to avoid confusion when using tools such as NSYS to analyze the program timeline.
2) Use C++17 feature `__has_include` instead of a CMake script to find out if NVTX available in the benchmark executable. It turns out our CMake check has been not reliable due to not finding include directories correctly.


lowener (approved):
LGTM

PR #408, opened 2024-10-10
Add `click` package to `cuvs-bench` conda recipe

divyegala:
This package is available in `dependencies.yaml`, but due to an oversight was not added to conda metas.

### DIFF
--- a/conda/recipes/cuvs_bench/meta.yaml
+++ b/conda/recipes/cuvs_bench/meta.yaml
@@ -82,6 +82,7 @@ requirements:
 
   run:
     - benchmark
+    - click
     - {{ pin_compatible('cuda-version', max_pin='x', min_pin='x') }}
     {% if cuda_major == "11" %}
     - cudatoolkit
--- a/conda/recipes/cuvs_bench_cpu/meta.yaml
+++ b/conda/recipes/cuvs_bench_cpu/meta.yaml
@@ -55,6 +55,7 @@ requirements:
 
   run:
     - benchmark
+    - click
     - glog {{ glog_version }}
     - h5py {{ h5py_version }}
     - matplotlib


dantegd (approved):

jameslamb (approved):
Approving this so you can merge this if it's time-sensitive, but I do have a suggestion.

Seeing a run-time dependency missed made me think "how could this not have been caught by tests"? And it looks like the answer is "there aren't any tests run in CI".

Would you please consider adding some tests on this package in https://github.com/rapidsai/cuvs/blob/branch-24.12/ci/test_python.sh?

I know this is a little different than other libraries because it's for benchmarking, but you could for example add a test like this one that all the other RAPIDS libraries have:

https://github.com/rapidsai/raft/blob/907edb8261a94a06f8857d0a10f2ec9abcd1ae42/python/raft-ann-bench/src/raft_ann_bench/_version.py

and then a test like this:

https://github.com/rapidsai/cuvs/blob/f62b217f97c9e14b340f11bcbfe556fcad9ed816/python/cuvs/cuvs/test/test_version.py

That's be a cheap and lightweight way to at least minimally test that the package is installable and importable. It might not have caught this specific issue because `click` is probably somewhere in the dependency tree of other packages, but it would catch this type of problem for other dependencies, and give us higher confidence in releases.

dantegd:
@jameslamb fully agree with your comment, adding tests is in the backlog and something I've wanted to get to as bandwidth permits. [I even added unit code pytests](https://github.com/rapidsai/cuvs/blob/branch-24.12/python/cuvs_bench/cuvs_bench/tests/test_run.py), but haven't had time to fully check them and enable them, it something we plan to do as soon as we can alongside as benchmarks like you suggest in a follow up :)

divyegala:
@jameslamb we have a test suite that we'll be focussing on getting ready this release https://github.com/rapidsai/cuvs/tree/branch-24.12/python/cuvs_bench/cuvs_bench/tests. Until then, if it's okay with you, let's merge this PR.

jameslamb:
> if it's okay with you, let's merge this PR.

Yeah totally fine.

> we have a test suite that we'll be focussing on getting ready this release

Sure, I understand. I think it'd be worth adding something lightweight like the `test_version.py` I mentioned in https://github.com/rapidsai/cuvs/pull/408#pullrequestreview-2373821599 sooner (instead of waiting to add the full test suite you want). Having some minimal validation that the conda packages are at least installable and importable would be useful.

You're not getting that kind of coverage from `conda mambabuild` because all the conda builds here are done with `--no-tests`.

https://github.com/rapidsai/cuvs/blob/f62b217f97c9e14b340f11bcbfe556fcad9ed816/ci/build_python.sh#L35-L39

divyegala:
@jameslamb yes, appreciate your suggestion here. I can do a quick follow on to add the minimal test and that will act as a placeholder in CI until the full test suite is ready.

PR #414, opened 2024-10-16
Test Ubuntu 24.04 in CI workflows.

bdice:
This PR tests Ubuntu 24.04 in CI workflows. See https://github.com/rapidsai/shared-workflows/pull/248.

Contributes to https://github.com/rapidsai/build-planning/issues/74.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -28,7 +28,7 @@ concurrency:
 jobs:
   cpp-build:
     secrets: inherit
-    uses: rapidsai/shared-workflows/.github/workflows/conda-cpp-build.yaml@branch-24.12
+    uses: rapidsai/shared-workflows/.github/workflows/conda-cpp-build.yaml@ubuntu24.04
     with:
       build_type: ${{ inputs.build_type || 'branch' }}
       branch: ${{ inputs.branch }}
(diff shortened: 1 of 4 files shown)


bdice:
It appears there is a failure due to the change to `rockylinux8` for conda C++ builds (we want to build on a system with the oldest supported glibc).

```
Formatting build metrics
awk: cmd. line:1: error: division by zero attempted
```

I'll look into this (a bit later).

bdice:
Testing is complete. https://github.com/rapidsai/shared-workflows/pull/248 has been merged, so we can close this.

bdice:
Following up: I forgot that the fix in `build.sh` needed to be merged for CI to pass. That was corrected in https://github.com/rapidsai/raft/pull/2474 and https://github.com/rapidsai/cuvs/pull/420.

The decision to use `python` here is pragmatic: there wasn't a good alternative that I could see.

The root cause of the problem was that `awk` throws an error for expressions with divide-by-zero even if that branch isn't taken!

```bash
$ echo - | awk "{ if (1 > 0) { print \"yes\" } else { printf \"%.2f\", 0 / 0 } }"
awk: cmd. line:1: error: division by zero attempted
```

We are using shell variables to provide the `$COMPILE_REQUESTS` and `$CACHE_HITS` so the `awk` command sees a literal `0 / 0` and fails, even though the conditional never goes down that path. It's like a compile-time error.

Other tools like `bc` also fail here, due to lacking a concept of `NaN`. Using `python` is the next best choice to avoid overcomplication, and it can be done in a one-line command like the previous `awk` command.

```bash
python3 -c "print(f'{${CACHE_HITS} / ${COMPILE_REQUESTS}:.2f}' if ${COMPILE_REQUESTS} else 'nan')"
```

PR #415, opened 2024-10-16
No variable named _CUVS_TEST_C_LIB exists

robertmaynard:
No variable named `_CUVS_TEST_C_LIB` exists, or any variant of that. So removing this dead code.


benfred (approved):
thanks for the fix!

Is there any way to catch things like this automatically ?

PR #417, opened 2024-10-17
Use dashes in cuvs-bench package name.

bdice:
This attempts to rename `cuvs_bench` to `cuvs-bench` and similarly for the CPU package. This follows from this thread: https://github.com/rapidsai/docker/pull/715#discussion_r1805232894


bdice:
I am hoping that CI covers this well. A close eye from reviewers would be helpful!

jameslamb (approved):
I looked around too (`git grep cuvs_bench`) and don't see any other references that I think need to be updated.

dantegd (approved):

Issue #419, opened 2024-10-21
Pending items for NN Descent


divyegala:
- [ ] Consolidate code in cuVS by migrating over diff of features from RAFT
- [ ] Triage CI wheel test failures
- [ ] Triage additional distance failures
- [ ] Triage HDBSCAN numerical instabilities when using NN Descent (https://github.com/rapidsai/cuml/pull/5939)

PR #421, opened 2024-10-21
Migrate feature diff for NN Descent from RAFT to cuVS

divyegala:
This PR is an amalgamation of the diff of 3 PRs in RAFT:

1. https://github.com/rapidsai/raft/pull/2345
2. https://github.com/rapidsai/raft/pull/2380
3. https://github.com/rapidsai/raft/pull/2403

This PR also addresses part 1 of #419 and making CAGRA use the compiled headers of NN Descent, which seemed to have been a pending TODO https://github.com/rapidsai/cuvs/blob/009bb8de03ce9708d4d797166187250f77a59a36/cpp/src/neighbors/detail/cagra/cagra_build.cuh#L36

### DIFF
--- a/cpp/include/cuvs/neighbors/nn_descent.hpp
+++ b/cpp/include/cuvs/neighbors/nn_descent.hpp
@@ -55,6 +55,8 @@ struct index_params : cuvs::neighbors::index_params {
   size_t intermediate_graph_degree = 128;     // Degree of input graph for pruning.
   size_t max_iterations            = 20;      // Number of nn-descent iterations.
   float termination_threshold      = 0.0001;  // Termination threshold of nn-descent.
+  bool return_distances            = false;   // return distances if true
+  size_t n_clusters                = 1;       // defaults to not using any batching
 
   /** @brief Construct NN descent parameters for a specific kNN graph degree
    *
@@ -101,13 +103,18 @@ struct index : cuvs::neighbors::index {
    * @param n_rows number of rows in knn-graph
    * @param n_cols number of cols in knn-graph
    */
-  index(raft::resources const& res, int64_t n_rows, int64_t n_cols)
+  index(raft::resources const& res, int64_t n_rows, int64_t n_cols, bool return_distances = false)
     : cuvs::neighbors::index(),
       res_{res},
       metric_{cuvs::distance::DistanceType::L2Expanded},
       graph_{raft::make_host_matrix<IdxT, int64_t, raft::row_major>(n_rows, n_cols)},
-      graph_view_{graph_.view()}
+      graph_view_{graph_.view()},
+      return_distances_{return_distances}
   {
+    if (return_distances) {
+      distances_      = raft::make_device_matrix<float, int64_t>(res_, n_rows, n_cols);
+      distances_view_ = distances_.value().view();
+    }
   }
 
   /**
@@ -121,12 +128,17 @@ struct index : cuvs::neighbors::index {
    * @param graph_view raft::host_matrix_view<IdxT, int64_t, raft::row_major> for storing knn-graph
    */
   index(raft::resources const& res,
-        raft::host_matrix_view<IdxT, int64_t, raft::row_major> graph_view)
+        raft::host_matrix_view<IdxT, int64_t, raft::row_major> graph_view,
+        std::optional<raft::device_matrix_view<float, int64_t, row_major>> distances_view =
+          std::nullopt,
+        bool return_distances = false)
     : cuvs::neighbors::index(),
       res_{res},
       metric_{cuvs::distance::DistanceType::L2Expanded},
       graph_{raft::make_host_matrix<IdxT, int64_t, raft::row_major>(0, 0)},
-      graph_view_{graph_view}
+      graph_view_{graph_view},
+      distances_view_{distances_view},
+      return_distances_{return_distances}
   {
   }
 
@@ -155,6 +167,13 @@ struct index : cuvs::neighbors::index {
     return graph_view_;
   }
 
+  /** neighborhood graph distances [size, graph-degree] */
+  [[nodiscard]] inline auto distances() noexcept
+    -> std::optional<device_matrix_view<float, int64_t, row_major>>
+  {
+    return distances_view_;
+  }
+
   // Don't allow copying the index for performance reasons (try avoiding copying data)
   index(const index&)                    = delete;
   index(index&&)                         = default;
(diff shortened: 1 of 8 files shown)


cpp/include/cuvs/neighbors/nn_descent.hpp
```diff
@@ -427,6 +451,270 @@ auto build(raft::resources const& res,
...
+ * @param[in] params an instance of nn_descent::index_params that are parameters
+ *               to run the nn-descent algorithm
+ * @param[in] dataset raft::device_matrix_view input dataset expected to be located
+ *                in device memory
+ * @param[out] index<IdxT> index containing all-neighbors knn graph in host memory
+ */
+void build(raft::resources const& res,
+           index_params const& params,
+           raft::device_matrix_view<const float, int64_t, raft::row_major> dataset,
+           index<uint32_t>& index);
```
cjnolet, line 483: Why doesn't nn-descent return the built index like all the other index types?

cpp/src/neighbors/detail/nn_descent.cuh
```diff
@@ -44,12 +50,13 @@
...
 #include <limits>
+#include <optional>
 #include <queue>
 #include <random>
 
 namespace cuvs::neighbors::nn_descent::detail {
-static const std::string RAFT_NAME = "raft";
-using pinned_memory_resource       = thrust::universal_host_pinned_memory_resource;
+
+using pinned_memory_resource = thrust::universal_host_pinned_memory_resource;
```
cjnolet, line 59: We should work to minimize the direct thrust calls in this algorithm. Not somehting that needs to be done in this PR, but can you create an issue to do this as tech debt in some future version? 

The more direct calls we make to libraries we don't control, the more things can break as those libraries evolve. We have a lot of pinned stuff that we do within RAFT, and we should try to use RAFT calls as much as possible.

cpp/src/neighbors/detail/nn_descent.cuh
```diff
@@ -1161,18 +1174,23 @@ GNND<Data_t, Index_t>::GNND(raft::resources const& res, const BuildConfig& build
...
-  thrust::fill(thrust::device,
-               dists_buffer_.data_handle(),
-               dists_buffer_.data_handle() + dists_buffer_.size(),
-               std::numeric_limits<float>::max());
-  thrust::fill(thrust::device,
-               reinterpret_cast<Index_t*>(graph_buffer_.data_handle()),
-               reinterpret_cast<Index_t*>(graph_buffer_.data_handle()) + graph_buffer_.size(),
-               std::numeric_limits<Index_t>::max());
-  thrust::fill(thrust::device, d_locks_.data_handle(), d_locks_.data_handle() + d_locks_.size(), 0);
+  raft::matrix::fill(res, dists_buffer_.view(), std::numeric_limits<float>::max());
```
cjnolet, line 1177: Very nice!

cpp/include/cuvs/neighbors/nn_descent.hpp
divyegala: It does, there's an API for that as well. We need this API as well especially for CAGRA because it needs to own the knn graph that it sends to NN Descent. For that, it needs to construct an index first and that's why we need this API.

cpp/include/cuvs/neighbors/nn_descent.hpp
```diff
@@ -119,14 +127,22 @@ struct index : cuvs::neighbors::index {
...
+   * @param distances_view optional raft::device_matrix_view<float, int64_t, row_major> for storing
+   * distances
+   * @param return_distances whether to return distances
    */
   index(raft::resources const& res,
-        raft::host_matrix_view<IdxT, int64_t, raft::row_major> graph_view)
+        raft::host_matrix_view<IdxT, int64_t, raft::row_major> graph_view,
+        std::optional<raft::device_matrix_view<float, int64_t, row_major>> distances_view =
+          std::nullopt,
+        bool return_distances = false)
```
cjnolet, line 138: Please infer this based on whether the optional contains a value so that we can avoid adding another argument here.

cpp/include/cuvs/neighbors/nn_descent.hpp
```diff
@@ -427,6 +451,270 @@ auto build(raft::resources const& res,
...
+ *                in device memory
+ * @param[out] index index containing all-neighbors knn graph in host memory
+ */
+void build(raft::resources const& res,
+           index_params const& params,
+           raft::device_matrix_view<const float, int64_t, raft::row_major> dataset,
+           index<uint32_t>& index);
+
+/**
+ * @brief Build nn-descent Index with dataset in host memory
```
cjnolet, line 486: Make sure all of these are exposed through the docs.

cpp/include/cuvs/neighbors/nn_descent.hpp
```diff
@@ -427,6 +451,270 @@ auto build(raft::resources const& res,
...
+ * @param[in] params an instance of nn_descent::index_params that are parameters
+ *               to run the nn-descent algorithm
+ * @param[in] dataset raft::host_matrix_view input dataset expected to be located
+ *                in host memory
+ * @param[out] index index containing all-neighbors knn graph in host memory
+ */
+void build(raft::resources const& res,
+           index_params const& params,
+           raft::host_matrix_view<const float, int64_t, raft::row_major> dataset,
+           index<uint32_t>& index);
```
cjnolet, line 516: This too- see above.

cpp/src/neighbors/detail/nn_descent.cuh
```diff
@@ -310,16 +311,17 @@ struct GnndGraph {
 
   raft::host_matrix<DistData_t, size_t, raft::row_major> h_dists;
 
-  thrust::host_vector<Index_t, pinned_memory_allocator<Index_t>> h_graph_new;
-  thrust::host_vector<int2, pinned_memory_allocator<int2>> h_list_sizes_new;
+  raft::pinned_matrix<Index_t, size_t> h_graph_new;
```
cjnolet, line 314: Nice to see!

cpp/test/neighbors/ann_nn_descent.cuh
```diff
@@ -146,6 +189,125 @@ class AnnNNDescentTest : public ::testing::TestWithParam<AnnNNDescentInputs> {
   rmm::device_uvector<DataT> database;
 };
 
+template <typename DistanceT, typename DataT, typename IdxT>
+class AnnNNDescentBatchTest : public ::testing::TestWithParam<AnnNNDescentBatchInputs> {
```
cjnolet, line 193: Do we have this consistently passing yet?

divyegala: No, that's a follow-on item. The tests are disabled for now both here and in RAFT.

cpp/include/cuvs/neighbors/nn_descent.hpp
divyegala: Yes, they are part of the doxygen group that is already in the docs source.

cpp/include/cuvs/neighbors/nn_descent.hpp
divyegala: Done in latest commit

PR #434, opened 2024-10-29
Add ci run_ scripts needed for build infra

robertmaynard:
These `run_*` scripts are needed by the build infra team and bring the cuvs project in line with the rest of RAPIDS

### DIFF
--- a/ci/run_ctests.sh
+++ b/ci/run_ctests.sh
@@ -1,0 +1,9 @@
+#!/bin/bash
+# Copyright (c) 2024, NVIDIA CORPORATION.
+
+set -euo pipefail
+
+# Support customizing the ctests' install location
+cd "${INSTALL_PREFIX:-${CONDA_PREFIX:-/usr}}/bin/gtests/libcuvs/"
+
+ctest --output-on-failure --no-tests=error "$@"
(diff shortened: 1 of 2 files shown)


jameslamb (approved):
If the goal of this PR is consistency with the rest of RAPIDS, I think the pytests script should be named `run_cuvs_pytests.sh`.

As most other projects do:

* raft example: https://github.com/rapidsai/raft/tree/branch-24.12/ci
* cudf example: https://github.com/rapidsai/cudf/tree/branch-24.12/ci
* cuml example: https://github.com/rapidsai/cuml/tree/branch-24.12/ci

Giving you a `ci-codeowners` approval so you don't have to request another review, but I do think you should make that change.

robertmaynard:
> If the goal of this PR is consistency with the rest of RAPIDS, I think the pytests script should be named `run_cuvs_pytests.sh`.
> 
> As most other projects do:
> 
> * raft example: https://github.com/rapidsai/raft/tree/branch-24.12/ci
> * cudf example: https://github.com/rapidsai/cudf/tree/branch-24.12/ci
> * cuml example: https://github.com/rapidsai/cuml/tree/branch-24.12/ci
> 
> Giving you a `ci-codeowners` approval so you don't have to request another review, but I do think you should make that change.

Happy to do so, I was copying kvikio ( https://github.com/rapidsai/kvikio/blob/branch-24.12/ci/run_pytests.sh ) which also only has a single python package. But we can rename now in expection that cuvs will have multiple packages

PR #435, opened 2024-10-30
Fix index overflow in edge cases of CAGRA graph optimize

achirkin:
Force `input_graph_degree`, `output_graph_degree`, and `graph_size` variables to `uint64_t`.
Before the PR, they've been `uint32_t`, and the product of them could overflow. This would lead to `cudaMemsetAsync` not filling in a large fraction of the graph.

It's not known whether this bug has surfaced for anyone until now, but it's better to be safe than sorry.
The change shouldn't have any impact on performance.


lowener (approved):
LGTM

PR #438, opened 2024-10-31
BUG: CAGRA multi-cta illegal access with bad queries

achirkin:
CAGRA search kernel errors with `cudaErrorIllegalAddress` in some conditions, specified in the new test case.

According to compute-sanitizer, the illegal access happens in [compute_distance_to_child_nodes(...) function](https://github.com/rapidsai/cuvs/blob/b422cbeec92fe925ee59f5f966ac9834440200e2/cpp/src/neighbors/detail/cagra/device_common.hpp#L185) accessing the graph.
The `parent_id` variable in that function sometimes appears to be out-of-bounds (larger than the graph size / number of records); it's invalid and seems to be the same reported by multiple threads, yet it's not an `invalid_index`, neither `index_msb_1_mask`, neither any derivative of the two.
Further observations:
  - I've checked the graph just before calling the search kernel; it does not contain any invalid indices.
  - One should disable any fancy pool memory resources to make it easier to reproduce the error (so that `parent_id` does not hit other user allocations in the pool)
  - It seems important that the query yields infinite distance to the dataset.
  - Running the search with a newly created `raft::resources` seems to increase the chance to hit the error
  - Even with all conditions satisfied, the error does not reproduce every time...

### Reproducer
```
./build.sh -n tests --limit-tests=NEIGHBORS_ANN_CAGRA_TEST && ./cpp/build/gtests/NEIGHBORS_ANN_CAGRA_TEST --gtest_filter=AnnCagraBugMultiCTACrash*
```

### DIFF
--- a/cpp/src/neighbors/detail/cagra/device_common.hpp
+++ b/cpp/src/neighbors/detail/cagra/device_common.hpp
@@ -181,8 +181,20 @@ RAFT_DEVICE_INLINE_FUNCTION void compute_distance_to_child_nodes(
     const IndexT smem_parent_id = parent_indices[i / knn_k];
     IndexT child_id             = invalid_index;
     if (smem_parent_id != invalid_index) {
+      // NOTE: BUG internal_topk_list contains out-of-range values (but not `invalid_index`) in rare
+      // cases this doesn't fail, indicating that parent_indices buffer is OK
+      // assert(smem_parent_id < IndexT(result_child_indices_ptr - internal_topk_list));
       const auto parent_id = internal_topk_list[smem_parent_id] & ~index_msb_1_mask;
-      child_id             = knn_graph[(i % knn_k) + (static_cast<int64_t>(knn_k) * parent_id)];
+      // NOTE: Find the root cause for the incorrect value?
+      // This condition fails, indicating the internal_topk_list content is NOT OK
+      // if (parent_id < dataset_desc.size) {
+      child_id = knn_graph[(i % knn_k) + (static_cast<int64_t>(knn_k) * parent_id)];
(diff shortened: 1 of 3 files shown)


achirkin:
CC @enp1s0

enp1s0:
@achirkin The out-of-range index seems to slip into the internal topk list in the random node pickup step. Can you change the following line and run the test again?
https://github.com/achirkin/cuvs/blob/e7496886792f7f409ce54d6015417c39604fa9f0/cpp/src/neighbors/detail/cagra/device_common.hpp#L148
```diff
- if (hashmap::insert(visited_hash_ptr, hash_bitlen, best_index_team_local)) {
+ if (best_index_team_local < dataset_desc.size && hashmap::insert(visited_hash_ptr, hash_bitlen, best_index_team_local)) {
```
The problem is that `result_distances_ptr[i]` is set to an uninitialized variable `best_norm2_team_local` if the norm is inf or nan. An alternative solution is to add `std::isfinite(best_norm2_team_local)` instead of `best_index_team_local < dataset_desc.size`.

achirkin:
Many thanks, @enp1s0, for finding the root cause of the problem!
The solution indeed works, though I slightly modified it:
  1. Initialize the  `best_index_team_local` with the invalid index value on the declaration site <- this should already be enough
  2. Check for the index value <- just to avoid costly hashmap check.

I think it's better to check the index than the distance, because the distance can be infinite on valid inputs. Also it's better not to touch `dataset_desc` without need, because it's in shared memory (not in registers) in the current implementation.

cjnolet (approved):

PR #441, opened 2024-11-01
Don't presume pointers location infers usability.

robertmaynard:
Here is the results of looking at the cudaPointerGetAttributes of different allocation types on Grace + Hopper. Allocations of `malloc` are still usable on the GPU.
```
ccudaPointerGetAttributes attributes malloc ptr
  is_dev_ptr  -> 1
  is_host_ptr -> 1
  memory loc  -> unregistered

cudaPointerGetAttributes attributes cudaMalloc ptr
  is_dev_ptr  -> 1
  is_host_ptr -> 0
  memory loc  -> device

cudaPointerGetAttributes attributes cudaMallocManaged cudaMemAttachGlobal ptr
  is_dev_ptr  -> 1
  is_host_ptr -> 1
  memory loc  -> managed

```

### DIFF
--- a/cpp/src/neighbors/detail/ann_utils.cuh
+++ b/cpp/src/neighbors/detail/ann_utils.cuh
@@ -63,14 +63,9 @@ struct pointer_residency_count<Type, Types...> {
     auto [on_device, on_host] = pointer_residency_count<Types...>::run(ptrs...);
     cudaPointerAttributes attr;
     RAFT_CUDA_TRY(cudaPointerGetAttributes(&attr, ptr));
-    switch (attr.type) {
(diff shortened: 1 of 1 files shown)


benfred (approved):

cjnolet:
@robertmaynard I just wanted to point out that with our mdspan APIs, we classify pointers as to their accessability instead of a direct classification of host vs device. That is, a `device_mdspan` only means the pointer can be read accessed from device and a `host_mdspan` means only that the pointer can be accessed from host. This way, we really didn't much need a `managed_msdpan` (though we create a `managed_mdarray` since that needs to know to use `cudaMallocManaged`.

PR #448, opened 2024-11-06
Adding tech stack to docs

cjnolet:


docs/source/index.rst
```diff
@@ -26,6 +15,67 @@ Useful Resources
 - `Issue tracker <https://github.com/rapidsai/cuvs/issues>`_: Report issues or request features.
 
 
+
+What is cuVS?
+#############
+
+cuVS contains state-of-the-art implementations of several algorithms for running approximate nearest neighbors and clustering on the GPU. It can be used directly or through the various databases and other libraries that have integrated it. The primary goal of cuVS is to simplify the use of GPUs for vector similarity search and clustering.
```
divyegala, line 22:
```suggestion
cuVS contains state-of-the-art implementations of several algorithms for running approximate and exact nearest neighbors and clustering on the GPU. It can be used directly or through the various databases and other libraries that have integrated it. The primary goal of cuVS is to simplify the use of GPUs for vector similarity search and clustering.
```

docs/source/index.rst
```diff
@@ -26,6 +15,67 @@ Useful Resources
...
+
+Vector search is an information retrieval method that has been growing in popularity over the past few  years, partly because of the rising importance of multimedia embeddings created from unstructured data and the need to perform semantic search on the embeddings to find items which are semantically similar to each other.
```
divyegala, line 24: Does it make sense to both words of "Vector Search" as a key phrase?
```suggestion
Vector Search is an information retrieval method that has been growing in popularity over the past few  years, partly because of the rising importance of multimedia embeddings created from unstructured data and the need to perform semantic search on the embeddings to find items which are semantically similar to each other.
```

docs/source/index.rst
```diff
@@ -26,6 +15,67 @@ Useful Resources
...
+Semantic search
+~~~~~~~~~~~~~~~
+- Generative AI & Retrieval augmented generation (RAG)
+- Recommender systems
+- Computer vision
+- Image search
+- Text search
+- Audio search
+- Molecular search
+- Model training
```
divyegala, line 41: Same question as above about capitalizing words of key phrases. "Retrieval Augmented Generation (RAG)" reads better to me than "Retrieval augmented generation (RAG)"

docs/source/index.rst
```diff
@@ -26,6 +15,67 @@ Useful Resources
...
+
+1. Fast index build
+2. Latency critical and high throughput search
+3. Parameter tuning
+4. Cost savings
+5. Interoperability (build on GPU, deploy on CPU)
+6. Multiple language support
+7. Building blocks for composing new or accelerating existing algorithms
+
+In addition to the items above, cuVS takes on the burden of keeping non-trivial accelerated code up to date as new NVIDIA architectures and CUDA versions are released. This provides a deslightful development experimence, guaranteeing that any libraries, databases, or applications built on top of it will always be getting the best performance and scale.
```
divyegala, line 66: I don't like the negative connotation of the word "burden" here. This is an optional request open to suggestions or rejection:
```suggestion
In addition to the items above, cuVS shoulders the responsibility of keeping non-trivial accelerated code up to date as new NVIDIA architectures and CUDA versions are released. This provides a deslightful development experimence, guaranteeing that any libraries, databases, or applications built on top of it will always be getting the best performance and scale.
```

divyegala (approved):

PR #450, opened 2024-11-08
Initial cut for a cuVS Java API

chatman:
A Java API for cuVS for easy integration into Apache Lucene or other Java based projects.

Try:
```
./build.sh libcuvs
./build.sh java
```

For generating docs, ```mvn javadoc:javadoc```

Prerequisites:
* JDK 22
* Maven 3.9.6+

Todo:
* Generate project panama classes using jextract on every build
* Algorithms other than Cagra 
* Prefiltering in cagra


chatman:
FYI @cjnolet ^
An ExampleApp.java is added as a starting point for the review.

chatman:
@naramgvivek10 Let's move CuVSResources to the cuvs package instead of common? That way we can abstract out the internals of Panama out of sight of the users.

java/README.md
```diff
@@ -0,0 +1,12 @@
+Prerequisites
+-------------
+
+* JDK 22
+* Maven 3.9.6 or later
+
+Please build libcuvs (`./build.sh libcuvs` from top level directory) before building the Java API with `./build.sh` from this directory.
```
cjnolet, line 7: As per comment above, I think we should mention this, but also mention that calling `build.sh java` from the top level will automatically build libcuvs.

java/cuvs-java/src/main/java/com/nvidia/cuvs/CagraIndex.java
```diff
@@ -0,0 +1,352 @@
+/*
+ * Copyright (c) 2023-2024, NVIDIA CORPORATION.
```
cjnolet, line 2: All new files should only include the current year.

java/cuvs-java/pom.xml
```diff
@@ -0,0 +1,163 @@
...
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-jar-plugin</artifactId>
+                <version>2.2</version>
+                <configuration>
+                    <archive>
+                        <manifest>
+                            <addClasspath>true</addClasspath>
+                            <mainClass>
+                                com.nvidia.cuvs.ExampleApp</mainClass>
```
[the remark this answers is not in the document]
chatman: Fixed, thanks!

.gitignore
```diff
@@ -1,5 +1,6 @@
 ## common
 __pycache__
+.gitignore
```
cjnolet, line 3: Please don't check .gitignore into .gitignore. We need to know when this file changes.

java/build.sh
```diff
@@ -0,0 +1,7 @@
+export CMAKE_PREFIX_PATH=`pwd`/../cpp/build
```
cjnolet, line 1: Can we put this logic in the top-level build.sh please? We do that with all of the other language support.

java/cuvs-java/src/main/java/com/nvidia/cuvs/panama/cuvsCagraSearchParams.java
```diff
@@ -0,0 +1,697 @@
+/*
```
cjnolet, line 1: I know we've discussed having this automatically generate on each build in a future iteraiton. Can you please create a cuVS Github issue to do that?

java/cuvs-java/src/test/java/com/nvidia/cuvs/CagraBuildAndSearchTest.java
```diff
@@ -0,0 +1,117 @@
...
+  /**
+   * A basic test that checks the whole flow - from indexing to search.
+   * 
+   * @throws Throwable
+   */
+  @Test
+  public void testIndexingAndSearchingFlow() throws Throwable {
+
+    // Sample data and query
+    float[][] dataset = {{ 0.74021935f, 0.9209938f }, { 0.03902049f, 0.9689629f }, { 0.92514056f, 0.4463501f }, { 0.6673192f, 0.10993068f }};
```
cjnolet, line 50: Hardcoding tests makes it very hard to debug and troubleshoot problems, and they also become very brittle because we've found changes in the compute architectures can introduce flakiness and downright failures from hardcoded tests. Please generate the data instead of hardcoding the values and use bfknn to validate.

.gitignore
narangvivek10: Fixed.

java/README.md
narangvivek10: Updated.

java/build.sh
```diff
@@ -0,0 +1,10 @@
+VERSION="25.02"
```
cjnolet, line 1: Make sure this version gets updated when `ci/release/update_versions.sh` is called.

java/cuvs-java/src/main/java/com/nvidia/cuvs/panama/CuVSCagraIndexParams.java
```diff
@@ -0,0 +1,354 @@
+/*
+ * Copyright (c) 2025, NVIDIA CORPORATION.
```
cjnolet, line 2: Just a small nitpick: I notice this file is prefixed with "CuVS" while most of the other files are prefixed with "Cuvs". I personally prefer "CuVS" (or even "cuVS" if that doesn't break convention too badly) but I would at least make it consistent.

narangvivek10: Thanks for this note. I will fix this soon.

java/cuvs-java/src/main/java/com/nvidia/cuvs/BruteForceIndex.java
```diff
@@ -0,0 +1,364 @@
...
+
+  private final float[][] dataset;
+  private final CuVSResources resources;
+  private MethodHandle indexMethodHandle;
+  private MethodHandle searchMethodHandle;
+  private MethodHandle destroyIndexMethodHandle;
+  private MethodHandle serializeMethodHandle;
+  private MethodHandle deserializeMethodHandle;
+  private IndexReference bruteForceIndexReference;
+  private BruteForceIndexParams bruteForceIndexParams;
```
ChrisHegarty, line 54: Can we make these MethodHandles static final - since they optimise much nicer by the JVM? I'm happy to rework some of the code flow here, if possible (and it that would be helpful) - though I'm not overly familiar with the development environment.

java/cuvs-java/src/main/java/com/nvidia/cuvs/BruteForceIndex.java
```diff
@@ -0,0 +1,364 @@
...
+ */
+public class BruteForceIndex {
+
+  private final float[][] dataset;
+  private final CuVSResources resources;
+  private MethodHandle indexMethodHandle;
+  private MethodHandle searchMethodHandle;
+  private MethodHandle destroyIndexMethodHandle;
+  private MethodHandle serializeMethodHandle;
+  private MethodHandle deserializeMethodHandle;
```
ChrisHegarty, line 52: Can we make these MethodHandles static final - since they optimise much nicer by the JVM? I'm happy to rework some of the code flow here, if possible (and it that would be helpful) - though I'm not overly familiar with the development environment.

narangvivek10: Thanks for your valuable feedback. I will try to incorporate this change. Please don't hesitate to suggest and commit to this branch. @chatman please send the invitation. Thanks!

The dev environment is as follows:
- Nvidia GPU RTX 2080 or above
- Ubuntu 22.04 
- JDK 22+
- Install CUDA 12.6
- Install cmake 3.26+
- apt install libnccl-dev ninja-build
- clone this repository and this branch
- `./build.sh libcuvs   # To build the libcuvs C library`
- `cd java && ./build.sh  # To build the java API (cuvs-java)`

cjnolet (approved):

java/cuvs-java/src/main/java/com/nvidia/cuvs/BruteForceIndex.java
ChrisHegarty: @chatman @narangvivek10 I'm getting compile errors. Before trying to describe them, lemme explain a little the environment that I'm using. I thought it might be most straightforward to use an AWS cloud instance, g4dn.xlarge, which has Tesla T4, then install the relevant dev tools. Should that instance be suitable to use, or is there another cloud instance or similar that would be more appropriate?

ci/release/update-version.sh
```diff
@@ -96,3 +96,9 @@ find .devcontainer/ -type f -name devcontainer.json -print0 | while IFS= read -r
     sed_runner "s@rapidsai/devcontainers/features/rapids-build-utils:[0-9.]*@rapidsai/devcontainers/features/rapids-build-utils:${NEXT_SHORT_TAG_PEP440}@" "${filename}"
     sed_runner "s@rapids-\${localWorkspaceFolderBasename}-${CURRENT_SHORT_TAG}@rapids-\${localWorkspaceFolderBasename}-${NEXT_SHORT_TAG}@g" "${filename}"
 done
+
+# Update Java API version
+sed_runner "s/VERSION=\".*\"/VERSION=\"${NEXT_FULL_TAG}\"/g" java/build.sh
+for FILE in java/*/pom.xml; do
+  sed_runner "/<!--CUVS_JAVA#VERSION_UPDATE_MARKER_START-->.*<!--CUVS_JAVA#VERSION_UPDATE_MARKER_END-->/s//<!--CUVS_JAVA#VERSION_UPDATE_MARKER_START--><version>${NEXT_FULL_TAG}<\/version><!--CUVS_JAVA#VERSION_UPDATE_MARKER_END-->/g" "${FILE}"
```
bdice, line 103: This is currently inconsistent with our normal use of the script. We usually run `./ci/release/update-version.sh 25.02.00` which would assign a value of `25.02.00`. However, the `pom.xml` files use `25.02` without the patch version. I would guess that we do want to have a patch version here, but perhaps we want to normalize it to `25.02.0`. That's close to what we do in cuDF's Java code.

Please look at https://github.com/rapidsai/cudf/blob/133e0c869531af94474e0bbb66cb22c5f8ba80f2/ci/release/update-version.sh#L87-L91 and https://github.com/rapidsai/cudf/blob/133e0c869531af94474e0bbb66cb22c5f8ba80f2/java/pom.xml#L24 and see whether we should adopt the same patterns here.

java/README.md
```diff
@@ -0,0 +1,14 @@
...
+* Maven 3.9.6 or later
+
+To build this API, please do `./build.sh java` in the top level directory. Since this API is dependent on `libcuvs` it must be noted that `libcuvs` gets built automatically before building this API.
+
+Alternatively, please build libcuvs (`./build.sh libcuvs` from top level directory) before building the Java API with `./build.sh` from this directory.
+ 
+Building
+--------
+
+`./build.sh` will generate the libcuvs_java.so file in internal/ directory, and then build the final jar file for the cuVS Java API in cuvs-java/ directory.
```
bdice, line 14:
```suggestion
`./build.sh` will generate the `libcuvs_java.so` file in the `internal/` directory, and then build the final jar file for the cuVS Java API in the `cuvs-java/` directory.
```

java/build.sh
```diff
@@ -0,0 +1,14 @@
+VERSION="25.02" # Note: The version is updated automatically when ci/release/update-version.sh is invoked
```
bdice, line 1: See comment in `update-version.sh` about inconsistency -- this should probably have a patch version.

java/cuvs-java/pom.xml
```diff
@@ -0,0 +1,159 @@
...
+ */
+--> 
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+    xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+    xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <groupId>com.nvidia.cuvs</groupId>
+    <artifactId>cuvs-java</artifactId>
+    <!-- NOTE: The version automatically gets updated when ci/release/update-version.sh is invoked. -->
+    <!--CUVS_JAVA#VERSION_UPDATE_MARKER_START--><version>25.02</version><!--CUVS_JAVA#VERSION_UPDATE_MARKER_END-->
```
bdice, line 25: See comment in `update-version.sh` about inconsistency -- this should probably have a patch version.

java/examples/pom.xml
```diff
@@ -0,0 +1,105 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <groupId>com.nvidia.cuvs.examples</groupId>
+  <artifactId>cuvs-java-examples</artifactId>
+  <!-- NOTE: The version automatically gets updated when ci/release/update-version.sh is invoked. -->
+  <!--CUVS_JAVA#VERSION_UPDATE_MARKER_START--><version>25.02</version><!--CUVS_JAVA#VERSION_UPDATE_MARKER_END-->
```
bdice, line 8: See comment in `update-version.sh` about inconsistency -- this should probably have a patch version.

rapids_config.cmake
bdice, line 1: No material changes were made in this file, and the beginning year for the copyright was removed. Let's revert this.

bdice (commented):
I skimmed the packaging details at @cjnolet's request. There are some merge conflicts with #593. ~I will push a couple small packaging-related changes to this PR to help it move along.~ I do not have power to push to this fork, so I can't help here. Please merge in `branch-25.02` and run `pre-commit`.

The biggest work item needed on the packaging side is to clarify what versioning should look like. I think a patch version (`25.02.0`) is important to add here, as we often publish patch releases. Currently the `update-version.sh` script uses a full tag like `25.02.00` but that doesn't correspond to what the version strings look like right now. Also, we should probably normalize it to `25.02.0` (drop the extra leading zero), as we do for several packaging systems (all Python wheels are normalized, as is cuDF's Java layer). I would look at how cuDF versions its Java layer for more inspiration, which I linked in a comment.

java/cuvs-java/src/main/java/com/nvidia/cuvs/BruteForceIndex.java
ChrisHegarty: Any advise on emulation, setup, environment, configuration, etc, would be gratefully appreciated.

chatman: We use local workstations with NVIDIA GPUs. One thing we realized is that gcc 11 works out of the box for compiling libcuvs (./build.sh libcuvs). Since it comes out of the box with Ubuntu 22.04 LTS, it is our development environment of choice. I got things to work on Fedora 39 as well, but with a lot of trickery to make CUDA pick up an older version of GCC (11), since it comes with GCC 14, which doesn't work in my experience. @cjnolet FYI.

In your case, I don't see which GCC is being used. However, there's a step where HNSWLib patch is applied, which is failing for some reason. Can you try to "build.sh libcuvs" on branch-25.02? If that doesn't work at the same step (patch apply for HNSWLib), there's maybe something broken in the build.

chatman:
(FYI @ChrisHegarty, @narangvivek10), Last working commit in this branch is:

```commit a197d5212146a3660ef260f6d84eae9ac06eec82 (HEAD)
Author: Vivek Narang <123010842+narangvivek10@users.noreply.github.com>
Date:   Fri Jan 24 14:45:44 2025 -0500

    Add examples, update GPU info API methods, and update CI script
    
    * Update CI script to update version in pom.xml files
    * Add and update examples
    * Add GPU info API methods
    ---------
    Co-authored-by: Vivek Narang <vivek@searchscale.com>
```

build.sh
```diff
@@ -445,6 +447,14 @@ if (( ${NUMARGS} == 0 )) || hasArg rust; then
     cargo test
 fi
 
+# Build the cuvs Java bindings
+if (( ${NUMARGS} == 0 )) || hasArg java; then
+    # build libcuvs first as the Java API depends on it
+    ./$0 libcuvs
```
bdice, line 453: I'm not sure if it is good practice to call this script recursively to build the C++ code. It seems like it will lead to unexpected behavior. For example, how do you ensure that the CLI arguments such as `--allgpuarch` or `--no-mg` are being passed through?

I would recommend removing this line, and users must call `./build.sh libcuvs java` if the C++ has not yet been built. Perhaps a warning can be issued if `hasArg libcuvs` is false. I believe this is how we handle the Python builds, which also depend on C++ libraries being built.
```suggestion
```

cpp/cmake/thirdparty/get_dlpack.cmake
bdice, line 1: Please revert any files with only copyright changes.

ci/release/update-version.sh
bdice: Do you think we should normalize it to `25.02.0` instead of `25.02.00`? That would align with what we do for Java code elsewhere in RAPIDS, including [cuDF](https://github.com/rapidsai/cudf/blob/03e1f64a3678b8b24c0390a781ca99d8c2234c97/java/pom.xml#L24) and [KvikIO](https://github.com/rapidsai/kvikio/blob/716a99f212ce2e0f9862bad1be81a12c56d03dc6/java/pom.xml#L10).

narangvivek10: Hey @bdice I will update the versioning following your suggestion above and push this change soon.

msarahan (approved):
Minor comment. You may want to run shellcheck on your .sh scripts. Mostly pedantic, but sometimes it catches annoying bugs.

bdice (approved):

PR #452, opened 2024-11-08
Add filtering for CAGRA to C API

ajit283:
Adds the CAGRA filtering feature to the C API using DLPack Tensor as blocklist


cpp/include/cuvs/neighbors/cagra.h
```diff
@@ -392,6 +392,22 @@ cuvsError_t cuvsCagraSearch(cuvsResources_t res,
...
+ * @param[out] neighbors DLManagedTensor* output `k` neighbors for queries
+ * @param[out] distances DLManagedTensor* output `k` distances for queries
+ */
+cuvsError_t cuvsCagraFilteredSearch(cuvsResources_t res,
+                                    cuvsCagraSearchParams_t params,
+                                    cuvsCagraIndex_t index,
+                                    DLManagedTensor* queries,
+                                    DLManagedTensor* neighbors,
+                                    DLManagedTensor* distances,
+                                    DLManagedTensor* filter);
```
benfred, line 410: Instead of having a different method for the filtered search - we are trying to just have one search method, with the 'filter' being an optional argument. 

As an example check out the bfknn code here 

https://github.com/rapidsai/cuvs/blob/5062594138a40231475299c7bac61083b0669fd1/cpp/include/cuvs/neighbors/brute_force.h#L159-L164

Also rather than passing a `DLManagedTensor* filter` object - can we pass the same `cuvsFilter prefilter`  that the bfknn code is doing ? I'd like to keep the API's consistent between the various search algorithms.

benfred (changes requested):
thanks for the PR! It'd be great to get this functionality into the cagra c-api

cpp/include/cuvs/neighbors/cagra.h
ajit283: Good idea, I just modified it to be more similar to the bfknn API. One potential issue is that every search call now needs to have a filter, even if no filter is used (same as the bfknn API).
Keep in mind that bfknn only supports bitmaps, and cagra only supports bitsets as filters.

cpp/src/neighbors/cagra_c.cpp
```diff
@@ -117,8 +119,27 @@ void _search(cuvsResources_t res,
   auto queries_mds            = cuvs::core::from_dlpack<queries_mdspan_type>(queries_tensor);
   auto neighbors_mds          = cuvs::core::from_dlpack<neighbors_mdspan_type>(neighbors_tensor);
   auto distances_mds          = cuvs::core::from_dlpack<distances_mdspan_type>(distances_tensor);
-  cuvs::neighbors::cagra::search(
-    *res_ptr, search_params, *index_ptr, queries_mds, neighbors_mds, distances_mds);
+  if (filter.type == NO_FILTER) {
+    cuvs::neighbors::cagra::search(
+      *res_ptr, search_params, *index_ptr, queries_mds, neighbors_mds, distances_mds);
+  } else if (filter.type == BITSET) {
+    using filter_mdspan_type    = raft::device_vector_view<int64_t, int64_t, raft::row_major>;
```
lowener, line 126:
```suggestion
    using filter_mdspan_type    = raft::device_vector_view<std::uint32_t, int64_t, raft::row_major>;
```
Use the same data type as below.

lowener (approved):

benfred (approved):
lgtm!

cpp/src/neighbors/cagra_c.cpp
```diff
@@ -117,8 +119,27 @@ void _search(cuvsResources_t res,
   auto queries_mds            = cuvs::core::from_dlpack<queries_mdspan_type>(queries_tensor);
   auto neighbors_mds          = cuvs::core::from_dlpack<neighbors_mdspan_type>(neighbors_tensor);
   auto distances_mds          = cuvs::core::from_dlpack<distances_mdspan_type>(distances_tensor);
-  cuvs::neighbors::cagra::search(
-    *res_ptr, search_params, *index_ptr, queries_mds, neighbors_mds, distances_mds);
+  if (filter.type == NO_FILTER) {
+    cuvs::neighbors::cagra::search(
+      *res_ptr, search_params, *index_ptr, queries_mds, neighbors_mds, distances_mds);
+  } else if (filter.type == BITSET) {
+    using filter_mdspan_type    = raft::device_vector_view<std::uint32_t, int64_t, raft::row_major>;
```
ajit283, line 126: I think the build fails because of the `std::uint32_t` as the first type argument. The argument must be of type `<index_t, index_t>`, see raft `bitset.hpp`:

```
template <typename bitset_t = uint32_t, typename index_t = uint32_t>
struct bitset {
  static constexpr index_t bitset_element_size = sizeof(bitset_t) * 8;

  /**
   * @brief Construct a new bitset object with a list of indices to unset.
   *
   * @param res RAFT resources
   * @param mask_index List of indices to unset in the bitset
   * @param bitset_len Length of the bitset
   * @param default_value Default value to set the bits to. Default is true.
   */
  bitset(const raft::resources& res,
         raft::device_vector_view<const index_t, index_t> mask_index,
         index_t bitset_len,
         bool default_value = true);
```

lowener: Hmm actually I missed this. The function you are using here is creating a bitset from a list of indices and I don't think it is the workflow that we expect.
The C++ function accepts a `bitset_view`, not a `bitset`, so at this point the memory for the bitset should already allocated and we just need to transfer the pointer and the length of the bitset. The C function should also assume that the filter given in input is a bitset already allocated and filled, instead of a list of neighbors to filter. So the `filter` taken as a parameter in this function should be manipulated as a `bitset_view` object.

cpp/test/neighbors/ann_cagra_c.cu
```diff
@@ -34,9 +34,14 @@ float queries[4][2] = {{0.48216683, 0.0428398},
                        {0.51260436, 0.2643005},
                        {0.05198065, 0.5789965}};
 
+int64_t filter[2] = {1, 2};
```
lowener, line 37: Adapt this test input in consequence.

lowener (changes requested):

cpp/test/neighbors/ann_cagra_c.cu
ajit283: I've adapted the `_search` function and the test so that it doesn't create a new bitset

lowener (approved):
LGTM

cpp/include/cuvs/neighbors/cagra.h
```diff
@@ -385,13 +386,16 @@ cuvsError_t cuvsCagraBuild(cuvsResources_t res,
...
+              to filter queries and neighbors based on the given bitset.
  */
 cuvsError_t cuvsCagraSearch(cuvsResources_t res,
                             cuvsCagraSearchParams_t params,
                             cuvsCagraIndex_t index,
                             DLManagedTensor* queries,
                             DLManagedTensor* neighbors,
-                            DLManagedTensor* distances);
+                            DLManagedTensor* distances,
+                            cuvsFilter filter);
```
lowener, line 398: This API change needs to be propagated to:
- the python package
- the example C project (`cuvs/example/c`)
- probably the rust package

lowener (changes requested):

cpp/include/cuvs/neighbors/cagra.h
ajit283: done👍

python/cuvs/cuvs/neighbors/cagra/cagra.pyx
```diff
@@ -480,7 +482,8 @@ def search(SearchParams search_params,
            k,
            neighbors=None,
            distances=None,
-           resources=None):
+           resources=None,
+           filter=None):
```
lowener, line 486: Add this parameter to the python documentation

python/cuvs/cuvs/neighbors/filters/filters.pyx
```diff
@@ -95,3 +95,52 @@ def from_bitmap(bitmap):
...
 
     return Prefilter(filter, parent=bitmap)
+
+def from_bitset(bitset):
+    """
+    Create a pre-filter from an array with type of uint32.
+
+    Parameters
+    ----------
+    bitmap : numpy.ndarray
```
lowener, line 105: Update this docstring for bitset instead of bitmap

lowener (approved):
LGTM

cjnolet (approved):

PR #454, opened 2024-11-08
Moving MG functions into unified API + `raft::device_resources_snmg` as device resource type for MG functions

viclafargue:
- Moving MG functions into unified API
- Using `raft::device_resources_snmg` as device resource type for MG functions
- Removing instantiation of `raft::comms::build_comms_nccl_only` solving https://github.com/rapidsai/raft/issues/2465

`raft::device_resources_snmg` is introduced in this PR : https://github.com/rapidsai/raft/pull/2549


cpp/include/cuvs/neighbors/mg.hpp
```diff
@@ -124,18 +124,18 @@ struct index {
...
  *
- * @param[in] handle
+ * @param[in] clique
  * @param[in] index_params configure the index building
  * @param[in] index_dataset a row-major matrix on host [n_rows, dim]
  *
  * @return the constructed IVF-Flat MG index
  */
-auto build(const raft::device_resources& handle,
+auto build(const raft::device_resources_snmg& clique,
```
cjnolet, line 138: One of the main goals with having the "snmg" resources object match the single-gpu object is that we wanted to be able to remove this additional MG. We should now be able to accept `device_resources` and then check to see if a nccl clique has been set on it (which would imply that it's a multi-gpu resources object and not a single-gpu object). The whole goal with doing this was to consolidate code paths.

viclafargue: The NCCL clique is not set as a resource anymore, but we should still be able to implement the dispatching by checking the dynamic type of the `device_resources`. The real question then is, do we truly want dispatching on both the regular API (`cuvs::neighbors::build`) and the `mg` namespace (`cuvs::neighbors::mg::build`)? It kind of make sense that a user providing a `device_resources_snmg` instance to the regular API (`cuvs::neighbors::build`) would want things to be deployed on multiple GPUs. However, the reverse is not necessarily true. A user who explicitly chose the `mg` namespace, but did not provide the adequate `device_resources_snmg` would fallback to single GPU, potentially unintentionally. Is this what we want?

I propose that we implement the dispatching mechanism solely on the regular API (`cuvs::neighbors::build`) in a dedicated follow-up PR? What do you think?

cjnolet:
Not sure why CI is still failing. The RAFT PR was merged quite awhile ago but I'm still seeing errors like this:

```
/home/coder/cuvs/cpp/include/cuvs/neighbors/mg.hpp:24:10: fatal error: raft/core/device_resources_snmg.hpp: No such file or directory
     24 | #include <raft/core/device_resources_snmg.hpp>
        |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

bdice:
This fell back to old 24.12 RAFT packages for unclear reasons:
```
    libraft-headers:             24.12.00-cuda12_241211_geaf9cc72_0 rapidsai   
    libraft-headers-only:        24.12.00-cuda12_241211_geaf9cc72_0 rapidsai   
```
https://github.com/rapidsai/cuvs/actions/runs/12818869804/job/35749392606?pr=454#step:8:521

cjnolet:
> This fell back to old 24.12 RAFT packages for unclear reasons:

Oh fooey! Thanks for noticing that @bdice, I just realized the darn target branch was never updated!

cpp/bench/ann/src/cuvs/cuvs_mg_cagra_wrapper.h
```diff
@@ -87,7 +86,7 @@ class cuvs_mg_cagra : public algo<T>, public algo_gpu {
   std::unique_ptr<algo<T>> copy() override;
 
  private:
-  raft::device_resources handle_;
+  raft::device_resources_snmg clique_;
```
cjnolet, line 89: `device_resources_snmg` is a type of `raft::resource`. The pattern here is to accept the `raft::resources` instance and call the free function inside `raft::resource::nccl_resource` underneath to get the clique out of the resource. This way the APIs don't need to look any different.

cpp/include/cuvs/neighbors/cagra.hpp
```diff
@@ -1752,4 +1756,469 @@ void serialize_to_hnswlib(raft::resources const& handle,
...
+ * auto index = cuvs::neighbors::cagra::build(clique, index_params, index_dataset);
+ * @endcode
+ *
+ * @param[in] clique a `raft::device_resources_snmg` object specifying the NCCL clique configuration
+ * @param[in] index_params configure the index building
+ * @param[in] index_dataset a row-major matrix on host [n_rows, dim]
+ *
+ * @return the constructed CAGRA MG index
+ */
+auto build(const raft::device_resources_snmg& clique,
```
cjnolet, line 1780: Please don't couple the standard headers to nccl. This is no good. Instead, please just accept a `raft::device_resources` here but provide a function in `raft::resource::nccl_resource` namespace to introspect the resources for a nccl clique (otherwise fail). The snmg build params should be anough of a difference

cpp/src/neighbors/mg/mg_cagra_float_uint32_t.cu
```diff
@@ -25,68 +25,68 @@
...
+  template <>                                                                                      \
+  cuvs::neighbors::mg::index<cagra::index<T, IdxT>, T, IdxT> deserialize<T, IdxT>(                 \
+    const raft::device_resources_snmg& clique, const std::string& filename)                        \
+  {                                                                                                \
+    auto idx = cuvs::neighbors::mg::index<cagra::index<T, IdxT>, T, IdxT>(clique, filename);       \
+    return idx;                                                                                    \
+  }                                                                                                \
+                                                                                                   \
+  template <>                                                                                      \
+  cuvs::neighbors::mg::index<cagra::index<T, IdxT>, T, IdxT> distribute<T, IdxT>(                  \
```
cjnolet, line 81: I thought we were removing the mg namespace completely from the public APIs.

cjnolet (changes requested):
I think this is nearly there. Just a couple very minor things at this point. 

Note- cuVS is not header-only and we shouldn't be putting constants in include files (that further couples cuVS to our exact build policies and makes it harder to navigate the build). 

Also, we should not be introducing ANY dependencies in the public header files. That's why we're using raft::resources as a facade so that we don't couple device_resources or device_resources_snmg to our public APIs. In other words, the user can use those objects if they want for convenience, but none of the other APIs in cuVS (nor in RAFT) use those objects because raft::resource is a generic container for these types of objects. That way the user can couple themselves to the dependencie sbut it doesn't force the dependencies on all other users. Does that make sense?

cpp/CMakeLists.txt
```diff
@@ -306,7 +306,6 @@ if(BUILD_SHARED_LIBS)
         src/neighbors/mg/mg_cagra_int8_t_uint32_t.cu
         src/neighbors/mg/mg_cagra_uint8_t_uint32_t.cu
         src/neighbors/mg/omp_checks.cpp
-        src/neighbors/mg/nccl_comm.cpp
```
cjnolet, line 309: Nice to see this removed.

cpp/bench/ann/src/cuvs/cuvs_mg_cagra_wrapper.h
cjnolet: We should be able to store `raft::resources` here now, right? The idea being that `has_resource_factory` for the `NCCL_CLIQUE` will be `true` when it's an SNMG type? (else, we'll need to add some other type of state, along with an accessor in the `resources/nccl_clique.hpp` for `is_multi_gpu()` and `is_multi_node()`

cjnolet: Maybe it's not a huge deal since this file is specific to mg. Could we at least rename the file to `snmg`, though? (Just in case we end up adding an mnmg then we won't have a clash).

cpp/include/cuvs/neighbors/common.hpp
```diff
@@ -756,40 +756,123 @@ struct iface {
...
-void build(const raft::device_resources& handle,
+void build(const raft::resources& handle,
            cuvs::neighbors::iface<AnnIndexType, T, IdxT>& interface,
            const cuvs::neighbors::index_params* index_params,
            raft::mdspan<const T, matrix_extent<int64_t>, row_major, Accessor> index_dataset);
 
 template <typename AnnIndexType, typename T, typename IdxT, typename Accessor1, typename Accessor2>
 void extend(
-  const raft::device_resources& handle,
+  const raft::resources& handle,
```
cjnolet, line 766: Very nice!

cpp/include/cuvs/neighbors/common.hpp
```diff
@@ -756,40 +756,123 @@ struct iface {
...
-void deserialize(const raft::device_resources& handle,
+void deserialize(const raft::resources& handle,
                  cuvs::neighbors::iface<AnnIndexType, T, IdxT>& interface,
                  const std::string& filename);
 
 };  // namespace cuvs::neighbors
+
+/// \defgroup mg_cpp_index_params ANN MG index build parameters
+
+namespace cuvs::neighbors {
```
cjnolet, line 799: Why restart/redefine the namespace here? Can't we just extend the existing namespace?

cpp/tests/neighbors/mg.cuh
```diff
@@ -19,8 +19,10 @@
 #include "ann_utils.cuh"
 #include "naive_knn.cuh"
 
-#include <cuvs/neighbors/mg.hpp>
```
cjnolet, line 22: Can we rename this to `snmg.cuh` (and the corresponding namespace to `snmg`)?

cjnolet (approved):

KyleFromNVIDIA (approved):

viclafargue:
/ok to test https://github.com/rapidsai/cuvs/pull/454/commits/1e1d97e83e1595a77a726021b4b1265778720080

PR #457, opened 2024-11-09
Migrate sparse knn and distances code from raft

benfred:


cpp/include/cuvs/sparse/neighbors/brute_force.hpp
```diff
@@ -0,0 +1,132 @@
...
+
+#pragma once
+
+#include <cuvs/distance/distance.hpp>
+
+#include <raft/core/device_csr_matrix.hpp>
+#include <raft/core/device_mdspan.hpp>
+#include <raft/core/handle.hpp>
+
+namespace cuvs::sparse::neighbors::brute_force {
```
cjnolet, line 25: Just a reminder, we are going to put these in `distance/distance.hpp` alongside the dense APIs. Since cuVS is strictly compiled, there's zero overhead for having large headers of definitions.

cpp/include/cuvs/sparse/neighbors/brute_force.hpp
```diff
@@ -0,0 +1,132 @@
...
+ * @{
+ */
+/**
+ * @brief Sparse Brute Force index.
+ *
+ * @tparam T Data element type
+ * @tparam IdxT Index element type
+ */
+template <typename T, typename IdxT>
+struct index {
```
cjnolet, line 38: Since we're consolidating, we could probable call this `sparse_index`. In practice, it's opaque to the user, since they are using a factory function to build this anyways. Thus, it doesn't matter to them whether the underlying implementation differs from the dense, just that they can call the APIs in the same fashion.

cpp/src/sparse/distance/detail/bin_distance.cuh
```diff
@@ -0,0 +1,231 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include "common.hpp"
```
cjnolet, line 19: Can we also expose the pairwise distances in `distance/distance.hpp` please?

cpp/test/sparse/neighbors/brute_force.cu
```diff
@@ -0,0 +1,177 @@
...
+ * limitations under the License.
+ */
+
+#include "../../test_utils.cuh"
+
+#include <cuvs/sparse/neighbors/brute_force.hpp>
+#include <raft/core/resource/cuda_stream.hpp>
+#include <raft/util/cudart_utils.hpp>
+
+#include <cusparse_v2.h>
```
cjnolet, line 23: I think we need the tests moved over for the distances too.

cpp/src/sparse/distance/detail/bin_distance.cuh
benfred: sure! should be done (with the tests too) in the latest commit

cpp/include/cuvs/sparse/neighbors/brute_force.hpp
benfred: moved over to `sparse_index` and `sparse_search_params` (the `search` and `build` functions are still the same names)

cjnolet (approved):

PR #461, opened 2024-11-11
Add serialization API to brute-force

lowener:
I noticed it was missing while switching Milvus to cuVS


python/cuvs/cuvs/neighbors/brute_force/brute_force.pyx
```diff
@@ -256,3 +257,73 @@ def search(Index index,
         ))
 
     return (distances, neighbors)
+
+
+@auto_sync_resources
+def save(filename, Index index, bool include_dataset=True, resources=None):
```
cjnolet, line 263: Can you please add the Python and C++ brute force serialization APIs to the docs?

cpp/include/cuvs/neighbors/brute_force.h
```diff
@@ -166,6 +166,51 @@ cuvsError_t cuvsBruteForceSearch(cuvsResources_t res,
...
+
+/**
+ * Load index from file.
+ *
+ * Experimental, both the API and the serialization format are subject to change.
+ *
+ * @param[in] res cuvsResources_t opaque C handle
+ * @param[in] filename the name of the file that stores the index
+ * @param[out] index BRUTEFORCE index loaded disk
+ */
```
cjnolet, line 206: Please provide a quick usage example here too.

cpp/include/cuvs/neighbors/brute_force.hpp
```diff
@@ -375,4 +383,230 @@ void search(raft::resources const& handle,
...
  */
 
+/**
+ * @defgroup bruteforce_cpp_index_serialize Bruteforce index serialize functions
+ * @{
+ */
+/**
+ * Save the index to file.
+ *
+ * Experimental, both the API and the serialization format are subject to change.
```
cjnolet, line 393: We may be able to remove this now (everywhere). This API has been around for awhile in RAFT and in the other indexes.

python/cuvs/cuvs/neighbors/brute_force/brute_force.pyx
```diff
@@ -256,3 +257,73 @@ def search(Index index,
...
+    """
+    cdef string c_filename = filename.encode('utf-8')
+    cdef cuvsResources_t res = <cuvsResources_t>resources.get_c_obj()
+    check_cuvs(cuvsBruteForceSerialize(res,
+                                       c_filename.c_str(),
+                                       index.index))
+
+
+@auto_sync_resources
+def load(filename, resources=None):
```
cjnolet, line 300: Please provide a usage example here as well.

cpp/include/cuvs/neighbors/brute_force.hpp
lowener: I removed the mention of experimental, but I still kept the mention that the serialization format can be subject to changes to not have users expect retro-compatibility.

cjnolet (approved):

PR #463, opened 2024-11-12
Moving spectral embedding and kernel gramm APIs to cuVS

### LINKED ISSUE #455: [FEA] Remaining APIs to migrate from RAFT
- [ ] Move random ball cover (knn and epsnn)
- [ ] Kernel Gramm
- [ ] Spectral clustering
- [ ] Expose needed APIs for HDBSCAN's single-linkage (build_dendrogram_host, build_sorted_mst)

cjnolet:
Partially addresses #455


benfred (approved):

cpp/include/cuvs/distance/grammian.hpp
```diff
@@ -0,0 +1,665 @@
...
+using dense_input_matrix_view_t = raft::device_matrix_view<const math_t, int, raft::layout_stride>;
+template <typename math_t>
+using dense_output_matrix_view_t = raft::device_matrix_view<math_t, int, raft::layout_stride>;
+template <typename math_t>
+using csr_input_matrix_view_t = raft::device_csr_matrix_view<const math_t, int, int, int>;
+
+/**
+ * Base class for general Gram matrices
+ * A Gram matrix is the Hermitian matrix of inner probucts G_ik = <x_i, x_k>
+ * Here, the  inner product is evaluated for all elements from vectors sets X1,
```
bdice, line 38:
```suggestion
 * Here, the inner product is evaluated for all elements from vectors sets X1,
```

KyleFromNVIDIA (approved):

cpp/include/cuvs/distance/grammian.hpp
```diff
@@ -0,0 +1,665 @@
...
+   * @param [in] n2 number vectors in x2
+   * @param [out] out device buffer to store the Gram matrix, size [n1*n2]
+   * @param [in] is_row_major whether the input and output matrices are in row
+   *        major format
+   * @param [in] stream cuda stream
+   * @param ld1 leading dimension of x1 (usually it is n1)
+   * @param ld2 leading dimension of x2 (usually it is n2)
+   * @param ld_out leading dimension of out (usually it is n1)
+   */
+  [[deprecated]] virtual void evaluate(const math_t* x1,
```
bdice, line 176: Are all these new methods supposed to be marked as deprecated? I am guessing this was copied from RAFT, and it was deprecated there. Is it still deprecated?

cpp/src/sparse/cluster/cluster_solvers.cuh
```diff
@@ -0,0 +1,100 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#ifndef __CLUSTER_SOLVERS_H
+#define __CLUSTER_SOLVERS_H
+
+#pragma once
```
bdice, line 20: Do we need both `#pragma once` and the header define `#define __CLUSTER_SOLVERS_H`? Typically those are used for the same purpose and only one is needed. I would favor `#pragma once` between the two.

bdice (approved):
Packaging / CMake looks fine, so approving. But I have some reservations about C++ here.

cpp/include/cuvs/distance/grammian.hpp
bdice: This applies to many APIs / many files.

cpp/src/sparse/cluster/cluster_solvers.cuh
cjnolet: This stuff was copied from raft so that we can move towards finishing the migration. We will continue to improve in cuVSz the grammian stuff specifically is complicated right now- the APIs need a significant update (which means removing the complex classes altogether and using free functions to match the rest of cuVs). we aren't going to bother making those changes yet because we want minimal impact to cuml up front (eg we just want to be able to update the namespaces in cuml for 25.02).

cjnolet: Also this whole cluster solvers API is going away. The spectral clustering and embedding APIs in general do not match the rest of cuVS (nor raft).

PR #465, opened 2024-11-13
HNSW CPU Hierarchy 

divyegala:
This PR adds an option to build the full HNSW hierarchy on the CPU when converting a CAGRA index to an hnswlib index. This lets us enable an `extend()` API. 

For hnswlib:
1. Update to `v0.7.0`
2. Remove dependency as symbols are compiled within DSO


jakirkham:
Does this need to be added to the Conda recipes as well?

divyegala:
@jakirkham no actually we don't want to use the hnswlib conda package because their python distribution messes up with our python version requirements. I will be removing it from our conda dev environments as well.

We'll rely on CMake instead to get us hnswlib. There's no double compilation issue anyway since it's a header-only library.

jakirkham:
Oh ok. Hadn't not looked into the nuances of this package or our usage of it

Indeed if we don't use the Python package, we should drop it

There doesn't appear to be a header-only C++ package currently. So just fetching it from source seems like the thing to do

Are the headers just used in the build to produce a library? Or are we shipping their headers too? If the latter, we may want to look at siloing it somehow to avoid clobbering or affecting other installs a user may have

divyegala:
@jakirkham we don't need to ship the headers, they are self-contained within the libcuvs shared object.

cpp/cmake/modules/ConfigureCUDA.cmake
```diff
@@ -22,8 +22,12 @@ endif()
...
 # clang)
 if(CMAKE_COMPILER_IS_GNUCXX)
-  list(APPEND CUVS_CXX_FLAGS -Wall -Werror -Wno-unknown-pragmas -Wno-error=deprecated-declarations)
-  list(APPEND CUVS_CUDA_FLAGS -Xcompiler=-Wall,-Werror,-Wno-error=deprecated-declarations)
+  list(APPEND CUVS_CXX_FLAGS -Wall -Werror -Wno-unknown-pragmas -Wno-error=deprecated-declarations
+       -Wno-reorder
+  )
+  list(APPEND CUVS_CUDA_FLAGS
+       -Xcompiler=-Wall,-Werror,-Wno-error=deprecated-declarations,-Wno-reorder
+  )
```
divyegala, line 30: `no-reorder` was added because of hnswlib `v0.7.0`. It is weird because the order warning that the compiler complained about in hnswlib was not actually out of order.

bdice (approved):
Approving build/packaging changes. (I also skimmed the C++ and Python and nothing seemed noteworthy.)

PR #466, opened 2024-11-14
Modify cuvs-bench to be able to generate ground truth in CPU systems

dantegd:
PR allows calculating ground truth for cuvs-bench on CPU systems. Current version uses a simple NumPy brute force, perhaps we should consider using faiss? cc @cjnolet @divyegala


cjnolet:
Thanks for opening this. I think it looks good. Should we also add the cuvs and cupy dependencies to the gpu cuvs-bench python package? Right now users get a module not found error because those aren't installed w/ the gpu-enabled package.

python/cuvs_bench/cuvs_bench/generate_groundtruth/__main__.py
```diff
@@ -15,82 +15,206 @@
...
+    elif metric == "inner_product":
+        similarities = xp.dot(
+            queries, dataset.T
+        )  # Shape: (n_queries, n_samples)
+
+        neg_similarities = -similarities
+        indices = xp.argpartition(neg_similarities, kth=k - 1, axis=1)[:, :k]
+        distances = xp.take_along_axis(similarities, indices, axis=1)
+
+        sorted_idx = xp.argsort(-distances, axis=1)
```
divyegala, line 152: Have you been able to verify that this works as expected?

cjnolet (approved):

KyleFromNVIDIA (approved):

PR #467, opened 2024-11-14
Fix include errors, header, and unsafe locks in iface.hpp

achirkin:
Fix a few issues with the internal header `neighbors/iface/iface.hpp`  leading to compile time errors and dangerous runtime behavior:

  - Add missing includes
  - Use `std::lock_guard` to avoid a deadlock on exception
  - Add NVIDIA header
  - Avoid an extra stream sync during search.


viclafargue (approved):
LGTM, thanks a lot for looking into this @achirkin!

benfred (approved):

PR #469, opened 2024-11-14
use different wheel-size thresholds based on CUDA version

jameslamb:
`cuvs-cu11` wheels are significantly larger than `cuvs-cu12` wheels, because (among other reasons) they are not able to dynamically link to CUDA math library wheels.

In #464, I proposed a size limit for CI checks of "max CUDA 11 wheel size + a buffer".

This PR proposes using different thresholds based on CUDA major version, following these discussions:

* https://github.com/rapidsai/cugraph/pull/4754#discussion_r1842526907
* https://github.com/rapidsai/cuml/pull/6136#discussion_r1841774811

### DIFF
--- a/ci/validate_wheel.sh
+++ b/ci/validate_wheel.sh
@@ -6,12 +6,26 @@ set -euo pipefail
 package_dir=$1
 wheel_dir_relative_path=$2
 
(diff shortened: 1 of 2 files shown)


msarahan (approved):

jameslamb:
Thanks @msarahan !

Issue #472, opened 2024-11-18
[BUG] The recall rate significantly drops by 20% when using CAGRA with a full 1s prefilter (nothing filtered out).


rhdong:
**Describe the bug**
When using CAGRA with a full ones prefilter, (nothing filtered out), the recall rate significantly drop by 20%

```shell
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
Benchmark                                            Time             CPU   Iterations        GPU    Latency     Recall end_to_end items_per_second      itopk          k  n_queries total_queries
--------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
raft_cagra.dim32/0/0/process_time/real_time       1.17 ms         1.17 ms        11941   1.16104m   1.17158m     0.9385    13.9899        853.548/s        128         10          1       11.941k algo="single_cta"
raft_cagra.dim32/1/0/process_time/real_time       2.14 ms         2.14 ms         6554   2.12471m   2.13548m   **0.789747**    13.9959        468.279/s        256         10          1        6.554k algo="single_cta"
raft_cagra.dim32/0/1/process_time/real_time       2.42 ms         2.42 ms         5768   2.41063m   2.42168m     0.9385    13.9682       412.937k/s        128         10       1000        5.768M algo="single_cta"
raft_cagra.dim32/1/1/process_time/real_time       4.59 ms         4.59 ms         3045   4.58215m   4.59372m    **0.79002**    13.9879       217.689k/s        256         10       1000        3.045M algo="single_cta"

```

**Steps/Code to reproduce bug**
[CUVS_CAGRA_ANN_BENCH reproduce branch](https://github.com/rhdong/cuvs/tree/cagra-filter-recall-rate-drop)

**Expected behavior**
The recall rate should be very close to the no prefilter case.

**Environment details (please complete the following information):**
 - Any benchmark environment.

**Additional context**

enp1s0:
@rhdong Thank you for the bug report. I have reproduced the problem when the itopk size is equal to or larger than 128. I'll investigate the cause, so please wait a while.

enp1s0:
@rhdong Can you apply this change and check the recall again? https://github.com/rapidsai/cuvs/pull/489

PR #489, opened 2024-11-23
[BUG] Fix CAGRA filter

enp1s0:
Ref : https://github.com/rapidsai/cuvs/issues/472

## The cause of the bug
The bitonic sort was used on an array that was not a power of 2 long. In the current search implementation, the bitonic sort is used to move the invalid elements to the end of the buffer as:
https://github.com/rapidsai/cuvs/blob/5062594138a40231475299c7bac61083b0669fd1/cpp/src/neighbors/detail/cagra/search_single_cta_kernel-inl.cuh#L758-L763
https://github.com/rapidsai/cuvs/blob/5062594138a40231475299c7bac61083b0669fd1/cpp/src/neighbors/detail/cagra/search_single_cta_kernel-inl.cuh#L644-L649

The problem is that the (max) array length (=`MAX_ITOPK + MAX_CANDIDATES`) is not always the power of two.
These bitonic sorts are called even if no elements are filtered out unless `cuvs::neighbors::filtering::none_sample_filter` is specified as the filter, so #472 occurs.

## Fix
This PR changes the filtering process so that the bitonic sort is not used to move the invalid elements to the end of the buffer.


lowener:
Can you add a test that would prevent regression?

achirkin (changes requested):
Thanks for the PR, @enp1s0!
I'm a little bit confused with the description. Do I understand it right that this PR contains two fixes: (1) make the bitonic sort array always a power-of-two, (2) move filtered elements to the end of the topk buffer?
The big chunk of the PR addresses (2), but that should be irrelevant for #472, because in that bug no elements are filtered out.
Therefore, I think, it would be really beneficial to construct a reproducer for #472 as a test case in this PR and make sure it's fixed with the introduced change.

achirkin:
Also, (1) did you have a chance to check if this reduces performance? (2) do we need a similar fix for multi-cta and multi-kernel versions of CAGRA?

enp1s0:
@achirkin, thank you for your comment, and I'm sorry for the bad PR description. I updated it.

> Do I understand it right that this PR contains two fixes: (1) make the bitonic sort array always a power-of-two, (2) move filtered elements to the end of the topk buffer?

No, this PR changes the filtering process so that the bitonic sort is not used to move the invalid elements to the end of the buffer. In the current search implementation, the bitonic sort is used to move the invalid elements as:
https://github.com/rapidsai/cuvs/blob/5062594138a40231475299c7bac61083b0669fd1/cpp/src/neighbors/detail/cagra/search_single_cta_kernel-inl.cuh#L758-L763
https://github.com/rapidsai/cuvs/blob/5062594138a40231475299c7bac61083b0669fd1/cpp/src/neighbors/detail/cagra/search_single_cta_kernel-inl.cuh#L644-L649

The problem is that the (max) array length (=`MAX_ITOPK + MAX_CANDIDATES`) is not always the power of two.
These bitonic sorts are called even if no elements are filtered out unless `cuvs::neighbors::filtering::none_sample_filter` is specified as the filter, so #472 occurs.

Also, this bug is the cause of a problem in the CAGRA filtering unit test:
https://github.com/rapidsai/cuvs/blob/5062594138a40231475299c7bac61083b0669fd1/cpp/test/neighbors/ann_cagra.cuh#L762
When `itop_k` is not specified, the default value, 64, is used. The graph degree is also 64. Therefore, `MAX_ITOPK (64) + MAX_CANDIDATES (64)` equals 128, and the bitonic sort works correctly in this case. However, if itopk size is set to another value, the bitonic sort does not work.

Although, as you mentioned, making the bitonic sort array always a power-of-two is an alternative way to fix this issue, I didn't do it because 1) the array elements except the filtered-out nodes are already sorted, and 2) more registers are required that will not be used but required to make the bitonic sort array power-of-two.

> did you have a chance to check if this affects the QPS?

I measured the performance of no filtering out search (the same situation as #472 )

filtering-bug


> do we need a similar fix for multi-cta and multi-kernel versions of CAGRA?

No.
* In the case of multi-CTA, a bitonic sort for the power-of-2 array is used to move the invalid elements, so there is no need to change. (We use a bitonic sort here because the array size is relatively small (32+graph_degree), which would not increase the register usage pressure.)
* In the case of multi-kernel, the `_find_topk` routine is called, and this bug is not related.

cpp/src/neighbors/detail/cagra/search_single_cta_kernel-inl.cuh
```diff
@@ -453,6 +453,48 @@ RAFT_DEVICE_INLINE_FUNCTION void topk_by_bitonic_sort(
...
+      const auto index    = index_array[i];
+      const auto distance = distance_array[i];
+
+      if (found_invalid) {
+        index_array[i - 1]    = index;
+        distance_array[i - 1] = distance;
+      } else {
+        // Check if the index is invalid
+        const auto I_found_invalid = (index == invalid_index);
+        const auto who_has_invalid = __ballot_sync(~0u, I_found_invalid);
```
achirkin, line 481: A nitpick
```suggestion
        const auto who_has_invalid = raft::ballot(I_found_invalid);
```

cpp/src/neighbors/detail/cagra/search_single_cta_kernel-inl.cuh
```diff
@@ -627,28 +669,27 @@ __device__ void search_core(
...
-          internal_topk + search_width * graph_degree,
-          internal_topk,
-          false);
+      if (!(std::is_same<SAMPLE_FILTER_T, cuvs::neighbors::filtering::none_sample_filter>::value ||
+            *filter_flag == 0)) {
+        // Move the filtered out index to the end of the itopk list
+        for (unsigned i = 0; i < search_width; i++) {
+          move_invalid_to_end_of_list(
+            result_indices_buffer, result_distances_buffer, internal_topk);
+        }
```
achirkin, line 678: Do I understand that right, that this algorithm moves one index at a time, and repeats this for each candidate in the list? That's `O(search_width*(parent_list_buffer + search_width))` complexity?

Maybe we'd better do a prefix scan (to get the indices of the valid items) followed by a shift of all elements (e.g. we can use `cub::WarpScan` or `cub::BlockScan` for that)? Or do you think the performance difference would be negligible?

cpp/src/neighbors/detail/cagra/search_single_cta_kernel-inl.cuh
```diff
@@ -755,12 +796,65 @@ __device__ void search_core(
...
     __syncthreads();
-    topk_by_bitonic_sort_1st<MAX_ITOPK + MAX_CANDIDATES>(
-      result_distances_buffer,
-      result_indices_buffer,
-      internal_topk + search_width * graph_degree,
-      top_k,
-      false);
+    // Move invalid index items to the end of the buffer without sorting the entire buffer
+    using scan_op_t    = cub::WarpScan<unsigned>;
+    auto& temp_storage = *reinterpret_cast<typename scan_op_t::TempStorage*>(smem_work_ptr);
```
achirkin, line 801: Is there always enough dynamically allocated shared memory by this pointer?

cpp/src/neighbors/detail/cagra/search_single_cta_kernel-inl.cuh
```diff
@@ -755,12 +796,65 @@ __device__ void search_core(
...
+        if (is_valid_index) {
+          const auto dst_position               = num_found_valid + (new_position - 1);
+          result_indices_buffer[dst_position]   = result_indices_buffer[src_position];
+          result_distances_buffer[dst_position] = result_distances_buffer[src_position];
+        }
+
+        // Calculate the largest valid position within a warp and bcast it for the next iteration
+        num_found_valid += new_position;
+        for (std::uint32_t offset = (warp_size >> 1); offset > 0; offset >>= 1) {
+          const auto v = __shfl_xor_sync(~0u, num_found_valid, offset);
```
achirkin, line 823: A nitpick
```suggestion
          const auto v = raft::shfl_xor( num_found_valid, offset);
```

cpp/src/neighbors/detail/cagra/search_single_cta_kernel-inl.cuh
```diff
@@ -755,12 +796,65 @@ __device__ void search_core(
...
+
+      if (num_found_valid < top_k) {
+        // Fill the remaining buffer with invalid values so that `topk_by_bitonic_sort` is usable in
+        // the next step
+        for (std::uint32_t i = num_found_valid + threadIdx.x; i < internal_topk; i += warp_size) {
+          result_indices_buffer[i]   = invalid_index;
+          result_distances_buffer[i] = utils::get_max_value<DISTANCE_T>();
+        }
+      }
+    }
```
achirkin, line 839: I've just realized you do here exactly what I wanted to suggest in the comment above. Could you please put this in a separate function for better readability? And then consider if it makes sense to re-use that in place of the `move_invalid_to_end_of_list` loop above?

cpp/src/neighbors/detail/cagra/search_single_cta_kernel-inl.cuh
```diff
@@ -755,12 +796,65 @@ __device__ void search_core(
...
+
+    // If the sufficient number of valid indexes are not in the internal topk, pick up from the
+    // candidate list.
+    if (top_k > internal_topk || result_indices_buffer[top_k - 1] == invalid_index) {
```
achirkin, line 843: Do I understand it right, that we need this because we filter the result indices buffer after we move candidates from the internal workspace, which is larger?
If so, why shouldn't we first filter the internal workspace and only then copy the results instead?

achirkin (changes requested):
Thanks @enp1s0 for the PR and the comprehensive answer. Now I understand the logic of the change it looks good to me overall.
Nevertheless, I have a few questions to the design below.

cpp/src/neighbors/detail/cagra/search_single_cta_kernel-inl.cuh
enp1s0: The reason I didn't use the scan method is that the maximum number of filtered-out elements here is `search_width`, which is typically 1. In our experience, increasing `search_width` does not help improve the search performance much, so it is likely to be a small number. So I use the specialized function for 1 here and run it multiple times if `search_width` is not 1.

cpp/src/neighbors/detail/cagra/search_single_cta_kernel-inl.cuh
enp1s0: What the code here does is 1) pick up valid nodes (=not filtered-out nodes) from the candidate list by the bitonic sort and 2) concatenate the resulting list and the itopk valid nodes list. This operation is needed when enough valid indices for topk are not obtained only from the itopk valid node list.

I think the function name `topk_by_bitonic_sort` is problematic because it is not a simple sort but takes a sorted list A and an unsorted list B and outputs merge_sort(A, bitonic_sort(B)). I'll change the name.

cpp/src/neighbors/detail/cagra/search_single_cta_kernel-inl.cuh
enp1s0: Fixed the smem size calculation in 3d676e2a45e14bfa5686cb1c92f560ed34ec28cc

enp1s0:
@achirkin Thank you for your review! I fixed the code, so can you check it again?

cpp/src/neighbors/detail/cagra/search_single_cta_kernel-inl.cuh
```diff
@@ -755,12 +797,66 @@ __device__ void search_core(
...
+
+      if (num_found_valid < top_k) {
+        // Fill the remaining buffer with invalid values so that `topk_by_bitonic_sort_and_merge` is
+        // usable in the next step
+        for (std::uint32_t i = num_found_valid + threadIdx.x; i < internal_topk; i += warp_size) {
+          result_indices_buffer[i]   = invalid_index;
+          result_distances_buffer[i] = utils::get_max_value<DISTANCE_T>();
+        }
+      }
+    }
```
achirkin, line 840: If if we use this routine only once, I still think it would be nice to move it out as a separate function alongside the` move_invalid_to_end_of_list` (so that we'd have two: `move_first_invalid_to_end_of_list` and `move_all_invalid_to_end_of_list`).
But given that we're very close to the code freeze and this PR is very important, I'd say we can postpone this grooming to the next release.

achirkin (approved):
Thank you for the updates! LGTM

cpp/src/neighbors/detail/cagra/search_single_cta_kernel-inl.cuh
achirkin: Makes sense, thanks for the explanation. For some reason I thought we could have more than `search_width` new  candidates each iteration.

PR #493, opened 2024-11-25
Update example code fetching rapids-cmake to use CUVS instead of RAFT

bdice:
Small update to CMake example code to use cuVS instead of RAFT.

### DIFF
--- a/examples/cmake/thirdparty/fetch_rapids.cmake
+++ b/examples/cmake/thirdparty/fetch_rapids.cmake
@@ -14,8 +14,8 @@
 # Use this variable to update RAPIDS and RAFT versions
 set(RAPIDS_VERSION "25.02")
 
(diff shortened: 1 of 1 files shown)


examples/cmake/thirdparty/fetch_rapids.cmake
```diff
@@ -14,8 +14,8 @@
 # Use this variable to update RAPIDS and RAFT versions
```
bdice, line 14: Should this comment be updated, too?
```suggestion
# Use this variable to update RAPIDS and cuVS versions
```

lowener: I think it should be updated, since changing this variable will change the cuVS version being used.

lowener (approved):

PR #494, opened 2024-11-25
Add C++ API scalar quantization

mfoerste4:
First draft for scalar quantization.

WIP status:
* only int8_t target type
* quantile computation inefficient (via sampling & sorting)


cpp/src/neighbors/detail/quantization.cuh
```diff
@@ -0,0 +1,164 @@
...
+  // conditional: search for quantiles / min / max
+  if (!params.is_computed) {
+    ASSERT(params.quantile > 0.5 && params.quantile <= 1.0,
+           "quantile for scalar quantization needs to be within (.5, 1]");
+
+    double quantile_inv = 1.0 / params.quantile;
+
+    // select subsample
+    int seed                     = 137;
+    constexpr size_t num_samples = 10000;
```
tfeher, line 53: For a dataset that has dim=1k , this would correspond to 10 vectors. I suggest to increase this significantly.

cpp/src/neighbors/detail/quantization.cuh
```diff
@@ -0,0 +1,164 @@
...
+    params.min         = double(minmax_h[0]);
+    params.max         = double(minmax_h[1]);
+    params.scalar      = double(range_q_type) / (params.max - params.min + 1.0);
+    params.is_computed = true;
+  }
+
+  // allocate target
+  auto out = raft::make_device_matrix<QuantI, int64_t>(res, dataset.extent(0), dataset.extent(1));
+
+  // raft unary op or raft::linalg::map?
```
tfeher, line 84: `map` is recommended I believe.

cpp/src/neighbors/detail/quantization.cuh
```diff
@@ -0,0 +1,164 @@
...
+    size_t subset_size           = std::min(num_samples, n_elements);
+    std::vector<T> subset(subset_size);
+    std::sample(dataset.data_handle(),
+                dataset.data_handle() + n_elements,
+                std::back_inserter(subset),
+                subset_size,
+                rng);
+
+    // quantile / sort and pick for now
+    thrust::sort(thrust::omp::par, subset.data(), subset.data() + subset_size);
```
tfeher, line 137: Need to confirm whether sorting on CPU side would be still preferred or not when we increase num_samples.

cpp/include/cuvs/neighbors/quantization.hpp
```diff
@@ -0,0 +1,66 @@
...
+#include <cuvs/neighbors/common.hpp>
+#include <raft/core/device_mdarray.hpp>
+#include <raft/core/device_mdspan.hpp>
+#include <raft/core/handle.hpp>
+#include <raft/core/host_mdarray.hpp>
+#include <raft/core/host_mdspan.hpp>
+
+namespace cuvs::neighbors::quantization {
+struct params {
+  float quantile = 0.99;
```
tfeher, line 28: Add docstring, mention value range.

cpp/include/cuvs/neighbors/quantization.hpp
```diff
@@ -0,0 +1,66 @@
...
+
+  bool is_computed = false;
+  double min;
+  double max;
+  double scalar;
+};
+
+raft::device_matrix<int8_t, int64_t> scalar_quantize(
```
tfeher, line 36: As discussed offline: it would be preferred to divide this into two step:
1. Train the quantizer, that derives the `[min, max]` range from `dataset` and `quantile`.
2. Quantize the dataset using the [min, max] range.

cpp/include/cuvs/neighbors/quantization.hpp
```diff
@@ -0,0 +1,66 @@
...
+#include <raft/core/device_mdspan.hpp>
+#include <raft/core/handle.hpp>
+#include <raft/core/host_mdarray.hpp>
+#include <raft/core/host_mdspan.hpp>
+
+namespace cuvs::neighbors::quantization {
+struct params {
+  float quantile = 0.99;
+
+  bool is_computed = false;
```
tfeher, line 30: Alternatively consider to have a `ScalarQuantizer` object, that would store the computed parameters. Then the `params` struct could remain a simple input argument.

tfeher (changes requested):
Thanks Malte for this PR, here are my first set of comments.

tfeher (changes requested):
Thanks Malte for the update, here are my next batch of comments.

tfeher (changes requested):
Hi Malte, this is already in a good shape, please take it out of draft state. A few more comments below.

cpp/src/neighbors/detail/quantization.cuh
tfeher: Alternative would be to gather the points on the CPU side, and then call the GPU to do the quantization. Assuming we have a streaming quantile estimation, that would not be faster to copy data to the device. But with the current sorting approach, it is not clear to me what is the better solution.

tfeher (approved):
Thanks Malte for the updates. The PR looks good to me.

cpp/include/cuvs/neighbors/quantization.hpp
```diff
@@ -0,0 +1,195 @@
...
+   * quantizer.train(handle, params, dataset);
+   * auto quantized_dataset = quantizer.transform(handle, dataset);
+   * @endcode
+   *
+   * @param[in] res raft resource
+   * @param[in] dataset a row-major matrix view on device
+   *
+   * @return device matrix with quantized dataset
+   */
+  raft::device_matrix<QuantI, int64_t> transform(
```
lowener, line 111: Most of the functions in cuvs usually take mdspan as inputs/outputs, but here you are using device arrays as return values. Is it possible to use mdspan as in/out to be as consistent as possible with the other functions of the cuvs API?

cpp/src/neighbors/detail/quantization.cuh
```diff
@@ -0,0 +1,205 @@
...
+  return static_cast<float>(a) == static_cast<float>(b);
+}
+
+template <typename T, typename QuantI, typename TempT = double>
+struct quantize_op {
+  const T min_;
+  const T max_;
+  const QuantI q_type_min_ = std::numeric_limits<QuantI>::min();
+  const QuantI q_type_max_ = std::numeric_limits<QuantI>::max();
+  const TempT a_;
```
lowener, line 48: Can another name be used for these variables? Like `scale_` for `a_`?

lowener (commented):
Add the documentation to doxygen

cpp/src/neighbors/detail/quantization.cuh
```diff
@@ -0,0 +1,208 @@
...
+
+  cudaStream_t stream = raft::resource::get_cuda_stream(res);
+
+  // select subsample
+  raft::random::RngState rng(seed);
+  size_t n_elements  = dataset.extent(0) * dataset.extent(1);
+  size_t subset_size = std::min(max_num_samples, n_elements);
+  auto subset        = raft::make_device_vector<T>(res, subset_size);
+  auto dataset_view  = raft::make_device_vector_view<const T>(dataset.data_handle(), n_elements);
+  raft::random::sample_without_replacement(
```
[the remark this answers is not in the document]
cjnolet: i think this depends on whether we need the min and max for each dimension separately or whether we are only talking about a single global min and max for the entire dataset.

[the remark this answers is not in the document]
tfeher: We talk about a single global min and max. We will keep the current implementation then. We will return to this question later, when we improve the quantile estimation method.

cpp/include/cuvs/neighbors/quantization.hpp
```diff
@@ -0,0 +1,195 @@
...
+   * cuvs::neighbors::quantization::ScalarQuantizer<float, int8_t> quantizer;
+   * cuvs::neighbors::quantization::sq_params params;
+   * quantizer.train(handle, params, dataset);
+   * @endcode
+   *
+   * @param[in] res raft resource
+   * @param[in] params configure scalar quantizer, e.g. quantile
+   * @param[in] dataset a row-major matrix view on device
+   */
+  void train(raft::resources const& res,
```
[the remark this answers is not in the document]
cjnolet: For example, `transform()`, `inverse_transform()` should be free functions that accept a `cuvs::neighbors::quantization::scalar_quantizer`. `train()` should return a new `scalar_quantizer()`.

cpp/include/cuvs/neighbors/quantization.hpp
```diff
@@ -0,0 +1,195 @@
...
+#include <cuvs/neighbors/common.hpp>
+#include <raft/core/device_mdarray.hpp>
+#include <raft/core/device_mdspan.hpp>
+#include <raft/core/handle.hpp>
+#include <raft/core/host_mdarray.hpp>
+#include <raft/core/host_mdspan.hpp>
+
+#include <cuda_fp16.h>
+
+namespace cuvs::neighbors::quantization {
```
cjnolet, line 28: We should move this outside of `cuvs::neighbors`. If we put everything that was even remotely nearest neighbors related into `cuvs::neighbors` then everything in cuvs would be in this namespace. 

Let's use `cuvs::preprocessing::quantize`

cjnolet:
Thanks for this feature, @mfoerste4. Do you know yet whether and by how much this improves over the corresponding CPU versions?

cpp/include/cuvs/preprocessing/quantization.hpp
```diff
@@ -0,0 +1,216 @@
...
+ * dataset); auto quantized_dataset = cuvs::preprocessing::quantization::transform(handle,
+ * quantizer, dataset);
+ * @endcode
+ *
+ * @tparam T data element type
+ * @tparam QuantI quantized type of data after transform
+ *
+ * @param[in] res raft resource
+ * @param[in] quantizer a scalar quantizer
+ * @param[in] dataset a row-major matrix view on device
```
lowener, line 124: Add `@param[out]` for all these functions. And add them to doxygen (`docs/source/cpp_api`)

cpp/include/cuvs/preprocessing/quantization.hpp
```diff
@@ -0,0 +1,222 @@
...
+ * @tparam QuantI quantized type of data after transform
+ *
+ * @param[in] res raft resource
+ * @param[in] params configure scalar quantizer, e.g. quantile
+ * @param[in] dataset a row-major matrix view on device
+ *
+ * @return ScalarQuantizer
+ */
+template <typename T, typename QuantI>
+ScalarQuantizer<T, QuantI> train_scalar(raft::resources const& res,
```
cjnolet, line 77: I think I prefer cuvs::preprocessing::quantize::scalar::train for consistency work the other APIs.

cpp/include/cuvs/preprocessing/quantization.hpp
```diff
@@ -0,0 +1,222 @@
...
+#include <raft/core/host_mdspan.hpp>
+
+#include <cuda_fp16.h>
+
+namespace cuvs::preprocessing::quantization {
+
+/**
+ * @brief ScalarQuantizer parameters.
+ */
+struct sq_params {
```
cjnolet, line 32: Please namespace to cuvs::preprocessing::quantize::scalar. Then this struct just becomes "params" nested in the scalar namespace. We are going to have several algorithms for quantizing and this will make it more straightforward for users.

Also please put create a directory for "quantize" and call this scalar.hpp.

cpp/include/cuvs/preprocessing/quantization.hpp
```diff
@@ -0,0 +1,222 @@
...
+ *
+ * The quantization is performed by a linear mapping of an interval in the
+ * float data type to the full range of the quantized int type.
+ *
+ * @tparam T data element type
+ * @tparam QuantI quantized type of data after transform
+ *
+ */
+template <typename T, typename QuantI>
+struct ScalarQuantizer {
```
cjnolet, line 51: This can just be "scalar_quantizer" (lower cased, for consistency with the other APIs).

cpp/include/cuvs/preprocessing/quantization.hpp
```diff
@@ -0,0 +1,222 @@
...
+ * @tparam QuantI quantized type of data after transform
+ *
+ * @param[in] res raft resource
+ * @param[in] params configure scalar quantizer, e.g. quantile
+ * @param[in] dataset a row-major matrix view on host
+ *
+ * @return ScalarQuantizer
+ */
+template <typename T, typename QuantI>
+ScalarQuantizer<T, QuantI> train_scalar(raft::resources const& res,
```
cjnolet, line 102: cuVS is strictly compile only and not header-only, so we don't expose any templates through public APIs, except for the occasional trivial class template for simple object types. All exposed free functions should be concrete types only.

cpp/include/cuvs/preprocessing/quantization.hpp
```diff
@@ -0,0 +1,222 @@
...
+ * @param[in] res raft resource
+ * @param[in] quantizer a scalar quantizer
+ * @param[in] dataset a row-major matrix view on device
+ * @param[out] out a row-major matrix view on device
+ *
+ */
+template <typename T, typename QuantI>
+void transform(raft::resources const& res,
+               const ScalarQuantizer<T, QuantI>& quantizer,
+               const raft::device_matrix_view<const T, int64_t> dataset,
```
cjnolet, line 131: We don't usually use const device_matrix_view because it is already a simple immutable view type. Instead please just use device_matrix_view<const T...>

cjnolet (changes requested):
I think we are almost there! Remaining changes from my side are mostly topical- hiding templates, naming and namespacing.

cjnolet (approved):

PR #502, opened 2024-11-29
[FEA] Serialize Vamana Dataset

tarang-jain:
Serialize the dataset in addition to the Vamana graph through `cuvs::neighbors::experimental::vamana::serialize`. This is required so that the saved index can be consumed by DiskANN / BANG for NN search.


tarang-jain:
Closing in favor of https://github.com/rapidsai/cuvs/pull/260

PR #509, opened 2024-12-04
Automatic adjustment of itopk size according to filtering rate

anaruse:
This PR is based on #492.

The new multi-CTA algorithm proposed in #492 can be used to obtain good recall even with high filtering rates. However, good recall cannot be obtained unless the number of search iterations, or itopk size, one of CAGRA's search parameters, is appropriately increased according to the filtering rate. Therefore, users need to find the appropriate itopk size according to the filtering rate by trial and error, which is a pain.

This PR is intended to alleviate this problem by internally calculating the filtering rate and automatically adjusting the itopk size accordingly.


cpp/include/cuvs/neighbors/cagra.hpp
```diff
@@ -229,6 +229,13 @@ struct search_params : cuvs::neighbors::search_params {
    * impact on the throughput.
    */
   float persistent_device_usage = 1.0;
+
+  /**
+   * A parameter indicating the rate of nodes to be filtered-out, when filtering is used.
+   * The value must be equal to or greater than 0.0 and less than 1.0. Default value is
+   * negative, in which case the filtering rate is automatically set.
+   */
+  float filtering_rate = -1.0;
```
cjnolet, line 238: This is interesting. So is specifying this value an alternative to explicitly computing the sparsity of the pre-filtering bitset (which would introduce additional compute into the search)?

cpp/src/neighbors/cagra.cuh
```diff
@@ -349,9 +351,18 @@ void search(raft::resources const& res,
     auto& sample_filter =
       dynamic_cast<const cuvs::neighbors::filtering::bitset_filter<uint32_t, int64_t>&>(
         sample_filter_ref);
+    search_params params_copy = params;
+    if (params.filtering_rate < 0.0) {
```
cjnolet, line 355: Oh I see- the filtering rate set to "auto" will automatically calculate the sparsity of the bitset while a user can skip having to compute this if they set the filtering rate directly. I like this.

cjnolet:
@anaruse it looks like there's some failed style checkers in this PR. Can you run the git pre-commit hooks to fix them? [Here's](https://docs.rapids.ai/api/cuvs/nightly/contributing/#python-pre-commit-hooks) the docs to enable them.

cpp/include/cuvs/neighbors/cagra.hpp
anaruse: It would be clearer to write "automatically computed" rather than "automatically set" here. I will fix it.

cpp/src/neighbors/cagra.cuh
anaruse: That is correct. By default, it calculates the filtering rate, but if the user specifies the filtering rate, that calculation is skipped.

cpp/include/cuvs/neighbors/cagra.hpp
cjnolet: I like that. "automatically compted" is perfect.

cjnolet:
@anaruse I think the changes look great. Can you also add the new search param to the main indexing docs for completeness? That would be the file in `docs/source/indexes/cagra.rst`

cjnolet (approved):

PR #516, opened 2024-12-04
Update cuvs to match raft's cutlass changes

vyasr:
Due to the tight integration between cuvs and raft, we need to ensure that cuvs is updated for rapidsai/raft#2503 or builds of cuvs that rely on cloning raft will get an incompatible version of cutlass due to raft's update.

### DIFF
--- a/cpp/cmake/thirdparty/get_cutlass.cmake
+++ b/cpp/cmake/thirdparty/get_cutlass.cmake
@@ -18,13 +18,12 @@ function(find_and_configure_cutlass)
   set(multiValueArgs)
   cmake_parse_arguments(PKG "${options}" "${oneValueArgs}" "${multiValueArgs}" ${ARGN})
 
-  # if(RAFT_ENABLE_DIST_DEPENDENCIES OR RAFT_COMPILE_LIBRARIES)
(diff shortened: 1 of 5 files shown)


cpp/cmake/thirdparty/get_cutlass.cmake
```diff
@@ -22,7 +24,7 @@ function(find_and_configure_cutlass)
       CACHE BOOL "Enable only the header library"
   )
   set(CUTLASS_NAMESPACE
-      "cuvs_cutlass"
+      "raft_cutlass"
```
KyleFromNVIDIA, line 27: Was the change from `cuvs` to `raft` intentional?

KyleFromNVIDIA, line 27: Was the change from `cuvs` to `raft` intentional?

vyasr: Nope, good catch thanks. The export sets were wrong too, fixed those.

KyleFromNVIDIA (approved):

benfred (approved):

PR #532, opened 2024-12-12
Fix cagra::extend error message

enp1s0:
When extending a CAGRA index that is built with `index_param.attach_dataset_on_build = false`, an error message "Only uncompressed dataset is supported" is displayed even if the dataset used to build the graph is not compressed. This problem occurs since the extend function does not check whether the dataset is empty. This PR fixes it.


achirkin (approved):
LGTM Thanks!

PR #541, opened 2024-12-17
Fix CI for python cuvs_bench

benfred:
I'm seeing CI failures due to a missing 'setuptools' on a recent PR https://github.com/rapidsai/cuvs/actions/runs/12363479650/job/34545223389?pr=536 in building the cuvs_bench code.

This seems to be because we were missing some requires in dependencies.yaml. Fix


bdice:
This seems not quite right. I’ll have a closer look.

bdice:
The [CI failure says](https://github.com/rapidsai/cuvs/actions/runs/12363479650/job/34545223389?pr=536#step:8:2649):
```
ValueError: Could not import build backend specified in pyproject.toml's tool.rapids-build-backend table. Make sure you specified the right optional dependency in your build-system.requires entry for rapids-build-backend.
```

but we already specified that here, I think? https://github.com/rapidsai/cuvs/blob/b859bc5edd41009604219416356cb6c96b189c2f/python/cuvs_bench/pyproject.toml#L7

bdice:
Aha. This occurs in conda package builds, so the `dependencies.yaml` / `pyproject.toml` are not used. This is occurring because we are missing `setuptools` in `conda/recipes/cuvs-bench/meta.yaml`. The proper fix should be the same as this PR from earlier today: https://github.com/rapidsai/dask-cuda/pull/1418

bdice (approved):

benfred:
Ahh - thanks @bdice ! Thats a great find - I've applied the same fix in the last commit

Issue #545, opened 2024-12-18
[FEA] Inner Product distance for CAGRA in python package


martinloretzzz:
**Is your feature request related to a problem? Please describe.**
For a project I need a GPU based index with the inner product distance, I looked a bit into the code, it looks like it's already implemented in the C++ API, but isn't exposed in the python library.

**Describe the solution you'd like**
Allow users of the python library to use CAGRA with the inner product distance.
Also error messages would be nice if distances are used in the python lib that aren't supported.

**Describe alternatives you've considered**
Compiling the lib myself, but my C++ skills are ~.

tarang-jain:
@martinloretzzz I have pushed a PR that fixes this and automatically closes the issue. If you see any more issues, feel free to re-open this and I'll look into it.
Thanks,
~Tarang

martinloretzzz:
Thanks a lot 🎉✨

PR #546, opened 2024-12-19
Additional Distances for CAGRA C and Python API

tarang-jain:
Add InnerProduct metric to CAGRA C and Python API + updates to CAGRA pytests.
Closes https://github.com/rapidsai/cuvs/issues/545

### DIFF
--- a/python/cuvs/cuvs/neighbors/cagra/cagra.pyx
+++ b/python/cuvs/cuvs/neighbors/cagra/cagra.pyx
@@ -136,6 +136,8 @@ cdef class IndexParams:
         Valid values for metric: ["sqeuclidean", "inner_product"], where
             - sqeuclidean is the euclidean distance without the square root
               operation, i.e.: distance(a,b) = \\sum_i (a_i - b_i)^2
+            - inner_product distance is defined as
+              distance(a, b) = \\sum_i a_i * b_i.
     intermediate_graph_degree : int, default = 128
 
     graph_degree : int, default = 64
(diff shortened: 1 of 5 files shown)


python/cuvs/cuvs/neighbors/cagra/cagra.pyx
```diff
@@ -170,10 +173,8 @@ cdef class IndexParams:
                  nn_descent_niter=20,
                  compression=None):
 
-        # todo (dgd): enable once other metrics are present
-        # and exposed in cuVS C API
-        # self.params.metric = _get_metric(metric)
-        # self.params.metric_arg = 0
+        self._metric = metric
+        self.params.metric = <cuvsDistanceType>DISTANCE_TYPES[metric]
```
divyegala, line 177: Can you check if the dictionary has an entry for "sqeuclidean" and not "euclidean"? I'm not sure how the tests were working until now.

tarang-jain: The tests until now were using the `L2Expanded` metric for CAGRA irrespective of what was being passed as the metric. "sqeuclidean" translates to `L2Expanded` and "euclidean" translates to `L2SqrtExpanded`. The min recall bound in the pytests so far was quite low (=0.7) which is why even if the groundtruth was created using inner product, the tests were passing. I've checked precisely what cuvs distance type is being dispatched to the C++ level.

divyegala (approved):

Issue #393, opened 2024-10-03
[FEA] Improve Vamana/DiskANN build


bkarsin:
Initial Vamana build has limited features and several items to improve performance or usability. Once these items are addressed, we can consider moving it out of the "experimental" namespace. They include:

- [ ] Reduce global memory footprint - this limits dataset and graph construction size. Simple ways include batching reverse edge generation, but will still be limited by requiring the entire dataset and graph be resident in device memory. Need to investigate storing the graph in host memory and the potential performance impacts as well.
- [ ] Add support for any dimension dataset (current have alignment issues with odd or < 16 for uint8 / int8.
- [ ] Auto-select and optimize `queue_size` for different `visited_size` values. Also, add support for any `visited_size` value (currently only poewers of 2.
- [ ] Improve performance for high-degree graph build. This seems limited by GreedySearch, which becomes increasingly costly. Things to investigate include reducing shared memory/registers to improve occupancy, improving priority queue efficiency, or trying other data structures.
- [ ] Add additional distance metrics - currently only L2 is supported. At least inner product and cosine are needed.

bkarsin:
Fixed alignment issue for datasets with dimensions not a multiple of 16, but this introduced a bug that I am currently working on. Resulting delay means we need to push this issue off to the next release.

bkarsin:
PR #558 addresses dataset dimension support and documentation

bkarsin:
Replacing with the newer and more detailed #1757.

PR #558, opened 2025-01-07
 Vamana build improvement and added docs

bkarsin:
Includes several fixes and improvements to Vamana, primarily:
- Edge case and bug fixes for Vamana index build (details below)
- Documentation added for Vamana
- experimental namespace removed
- Reduce device memory usage by splitting reverse edge work into batches

The edge case fix adds padding to all shared memory size and offset calculations so any dataset dimension is supported (tests added that verify this). A bug was also fixed with the L2 distance metric causing incorrect results in some rare cases. 

This PR addresses the most pressing items in #393 and stabilize the index construction sufficiently to remove the experimental namespace.


docs/source/indexes/vamana.rst
```diff
@@ -0,0 +1,76 @@
...
+
+The 2 hyper-parameters that are most often tuned are `graph_degree` and `visited_size`. The time needed to create a graph increases dramatically when increasing `graph_degree`, in particular. However, larger graphs may be needed to achieve very high recall search, especially for large datasets.
+
+Memory footprint
+----------------
+
+Vamana builds a graph that is stored in device memory. However, in order to serialize the index and write it to a file for later use, it must be moved into host memory. If the `include_dataset` parameter is also set, then the dataset must be resident in host memory when calling serialize as well. 
+
+Device memory usage
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
```
cjnolet, line 67: I'm surprised the docs built without error. I'm sure there's a warning in there about having the underlining for headings be longer the text.

docs/source/indexes/vamana.rst
```diff
@@ -0,0 +1,76 @@
...
+
+The built index represents the graph as fixed degree, storing a total of :math:`graph_degree * n_index_vectors` edges. Graph construction also requires the dataset be in device memory (or it copies it to device during build). In addition, device memory is used during construction to sort and create the reverse edges. Thus, the amount of device memory needed depends on the dataset itself, but it is bounded by a maximum sum of:
+
+- vector dataset: :math:`n_index_vectors * n__dims * sizeof(T)`
```
cjnolet, line 71: I think you might need to escape all of the _'s in here. Have you checked to see how this formats?

bkarsin: Did a quick scan but didn't look at all the formatting details. Thanks for checking, I'll do a pass over the format of the final built docs.

cpp/src/neighbors/detail/vamana/vamana_build.cuh
```diff
@@ -253,6 +237,15 @@ void batched_insert_vamana(
       int total_edges;
       raft::copy(&total_edges, d_total_edges.data_handle(), 1, stream);
```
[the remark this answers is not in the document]
bkarsin: I thought that because the next line allocates on the same stream, it should be serialized? 
Will add a stream (kinda prefer not to have rmm calls directly in the code, since everything else uses `raft`).

cpp/src/neighbors/detail/vamana/robust_prune.cuh
```diff
@@ -145,9 +145,11 @@ __global__ void RobustPruneKernel(
   // Dynamic shared memory used for blocksort, temp vector storage, and neighborhood list
   extern __shared__ __align__(alignof(ShmemLayout)) char smem[];
 
-  T* s_coords = reinterpret_cast<T*>(&smem[sort_smem_size]);
-  DistPair<IdxT, accT>* new_nbh_list =
-    reinterpret_cast<DistPair<IdxT, accT>*>(&smem[dim * sizeof(T) + sort_smem_size]);
+  int align_padding = (((dim - 1) / alignof(ShmemLayout)) + 1) * alignof(ShmemLayout) - dim;
```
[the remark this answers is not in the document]
bkarsin: Thanks, done.

docs/source/indexes/vamana.rst
bkarsin: Fixed all the DiskANN/vamana docs.

docs/source/indexes/vamana.rst
bkarsin: Fixed.

cpp/src/neighbors/detail/vamana/vamana_build.cuh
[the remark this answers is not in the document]
tfeher: That is a good point, but without going through the implementation details of `mdspan`, I it is not clear to me when the value of the `total_edges` variable is read while we prepare the extents of the matrix. Therefore I prefer the explicit sync, thanks for adding it!

Regarding `rmm`, it is legal to use it in cuvs, but it is also fine to stick to `mdarray`.

tfeher (approved):
Thanks @bkarsin for the updates, the PR looks good to me!

PR #560, opened 2025-01-08
[Feat] Support `bitset` filter for Brute Force

rhdong:


cpp/cmake/thirdparty/get_raft.cmake
```diff
@@ -60,8 +60,8 @@ endfunction()
 # To use a different RAFT locally, set the CMake variable
 # CPM_raft_SOURCE=/path/to/local/raft
 find_and_configure_raft(VERSION  ${RAFT_VERSION}.00
-        FORK                     ${RAFT_FORK}
-        PINNED_TAG               ${RAFT_PINNED_TAG}
+        FORK                     rhdong
```
rhdong, line 63: Before reverting this customized raft, please don't merge this PR.

cpp/src/neighbors/brute_force_c.cpp
```diff
@@ -67,8 +67,8 @@ void _search(cuvsResources_t res,
   using queries_mdspan_type   = raft::device_matrix_view<T const, int64_t, QueriesLayoutT>;
   using neighbors_mdspan_type = raft::device_matrix_view<int64_t, int64_t, raft::row_major>;
   using distances_mdspan_type = raft::device_matrix_view<float, int64_t, raft::row_major>;
-  using prefilter_mds_type    = raft::device_vector_view<const uint32_t, int64_t>;
```
cjnolet, line 70: I think we want to keep the filter immutable, don' we?

cpp/include/cuvs/neighbors/common.hpp
```diff
@@ -456,8 +457,11 @@ inline constexpr bool is_vpq_dataset_v = is_vpq_dataset<DatasetT>::value;
 
 namespace filtering {
 
+enum class FilterType { None, Bitmap, Bitset };
+
 struct base_filter {
-  virtual ~base_filter() = default;
+  virtual ~base_filter()                     = default;
```
cjnolet, line 463: I notice no changes have been made to `brute_force.hpp`. Ideally, we'll at at least be listing out in the docs which filters are supported, right? Otherwise this is going to be very confusing for users. Also, can we set the default to `bitset` filter? I suspect most users will want bitset.

cpp/src/neighbors/brute_force_c.cpp
rhdong: This change is to be compatible with the `bitset` filter utilized in CAGRA and other algorithms. The immutable is implemented on the `search` API by restricting the filter to be `const cuvs::neighbors::filtering::base_filter`. Please refer to [cagra](https://github.com/rapidsai/cuvs/blob/branch-25.02/cpp/src/neighbors/cagra.cuh#L350), here

cpp/include/cuvs/neighbors/common.hpp
rhdong: I’ve just added the comments. I believe using bitset as the default setting might not be ideal if we don't have enough input from end-users. Perhaps we should discuss this in the team group, as I noticed that the none filter is also set as the default in CAGRA.

cjnolet: > I believe using bitset as the default setting might not be ideal if we don't have enough input from end-users.

I think you may have misunderstood me. The none filter is fine as the default, but for the code example, we should make sure we use a bitset and leave bitmap to users who need it. FAISS doesn't even support a bitmap and users aren't asking for it generally but it's good that we have it for those who might need it.

cpp/src/neighbors/brute_force_c.cpp
cjnolet: We are using obj<const type> everywhere in the codebase so we should be doing that here as well. If the CAGRA API is not doing this, it should be.

rhdong:
Hey, @cjnolet @benfred @lowener,  I've addressed all comments. I'd appreciate it if you could take a moment to re-review and approve. Thanks!

cpp/include/cuvs/neighbors/brute_force.hpp
```diff
@@ -421,15 +464,51 @@ void search(raft::resources const& handle,
...
  * @param[out] neighbors a device pointer to the indices of the neighbors in the source dataset
  * [n_queries, k]
  * @param[out] distances a device pointer to the distances to the selected neighbors [n_queries, k]
- * @param[in] sample_filter an optional device bitmap filter function that greenlights samples for a
- * given query
+ * @param[in] sample_filter An optional device filter that restricts which dataset elements should
+ * be considered for each query.
+ *
+ * - Supports two types of filters:
+ *   1. **Bitmap Filter**: A per-query filter with a logical shape of `[n_queries, index->size()]`,
```
cjnolet, line 505: Can we reorder these please? I think we should be pushing bitset as the first option.

cpp/src/neighbors/detail/knn_brute_force.cuh
```diff
@@ -56,9 +56,13 @@
 
 #include <cstdint>
 #include <iostream>
+#include <optional>
 #include <set>
+#include <variant>
 
 namespace cuvs::neighbors::detail {
+
+using namespace cuvs::neighbors::filtering;
```
cjnolet, line 65: Never use `using namespace` in a header file. It cascades to all header files.

rhdong: All done! Thanks!

cjnolet (approved):

PR #561, opened 2025-01-08
FAISS with cuVS enabled in cuvs-bench

tarang-jain:
Mostly adapted from https://github.com/rapidsai/raft/pull/2026


cpp/bench/ann/src/common/benchmark.hpp
```diff
@@ -483,7 +483,14 @@ void register_search(std::shared_ptr<const dataset<T>> dataset,
                   ->MeasureProcessCPUTime()
                   ->UseRealTime();
 
-      if (metric_objective == Mode::kThroughput) { b->ThreadRange(threads[0], threads[1]); }
+      if (metric_objective == Mode::kThroughput) {
```
cjnolet, line 486: Let's not put algorithm-specific stuff in the common files. I agree with @achirkin about this one. If needed, you could propoagate this down to the specific algorithms, but the common stuff should be agnostic of he algoithmmic specific.

achirkin: Thanks, @cjnolet for spotting this. @tarang-jain , could you please move the warning into the faiss-related header (perhaps into the set_search_params)? There, you should be able to query the `cuvs::bench::benchmark_n_threads > 1` from `common/util.hpp`.

cpp/bench/ann/src/faiss/faiss_gpu_benchmark.cu
```diff
@@ -138,5 +153,18 @@ REGISTER_ALGO_INSTANCE(std::uint8_t);
...
+  // and is initially sized to half of free device memory.
+  rmm::mr::pool_memory_resource<rmm::mr::cuda_memory_resource> pool_mr{
+    &cuda_mr, rmm::percent_of_free_device_memory(50)};
+  // Updates the current device resource pointer to `pool_mr`
+  auto old_mr = rmm::mr::set_current_device_resource(&pool_mr);
+  auto ret = cuvs::bench::run_main(argc, argv);
+  // Restores the current device resource pointer to its previous value
+  rmm::mr::set_current_device_resource(old_mr);
+  return ret;
+}
```
achirkin, line 169: We've run in many issues in ann-bench in the past doing the program-wide setting of memory resource. This also changes the behavior depending on whether `ANN_BENCH_BUILD_MAIN` flag is enabled. Therefore, it's best if these lines stay unmodified and same for all algorithms.
Please move this resource setting to the place where the index is constructed/destructed  in the algo header file.

achirkin (changes requested):
Here are couple suggestions on my side to keep the faiss behavior in line with other algorithms.

cpp/bench/ann/src/common/benchmark.hpp
```diff
@@ -483,6 +483,14 @@ void register_search(std::shared_ptr<const dataset<T>> dataset,
                   ->MeasureProcessCPUTime()
                   ->UseRealTime();
 
+      if (metric_objective == Mode::kThroughput) {
+        if (index.algo.find("faiss_gpu") != std::string::npos) {
```
cjnolet, line 487: This is, again, putting algorithm specific stuff in the common files. Please put this in the FAISS benchmark files.

cpp/bench/ann/src/faiss/faiss_gpu_benchmark.cu
tarang-jain: Solved this. I'm now using the configured_raft_resources object similar to how the other cuvs algos are doing.

tarang-jain: Marking as resolved.

cpp/bench/ann/src/common/benchmark.hpp
tarang-jain: Fixed this! Marking as resolved.

cpp/bench/ann/src/common/benchmark.hpp
tarang-jain: Fixed this! Marking as resolved.

cpp/bench/ann/src/common/benchmark.hpp
```diff
@@ -484,6 +484,7 @@ void register_search(std::shared_ptr<const dataset<T>> dataset,
                   ->UseRealTime();
 
       if (metric_objective == Mode::kThroughput) { b->ThreadRange(threads[0], threads[1]); }
+      if (metric_objective == Mode::kThroughput) { b->ThreadRange(threads[0], threads[1]); }
```
achirkin, line 487: I guess, this can be removed now?
```suggestion
```

cjnolet (approved):

cpp/bench/ann/CMakeLists.txt
```diff
@@ -124,6 +125,7 @@ function(ConfigureAnnBench)
   target_link_libraries(
     ${BENCH_NAME}
     PRIVATE ${ConfigureAnnBench_LINKS}
+            $<$<BOOL:${CUVS_FAISS_ENABLE_GPU}>:raft::raft>
```
achirkin, line 128: Please move this dependency down to the algo-specific configs.

tarang-jain: Done!

achirkin (approved):
Thanks for the updates, no more complaints on my side :)

tarang-jain:
.merge

PR #563, opened 2025-01-10
Add support for refinement with `uint32_t` index type

### LINKED ISSUE #537: [BUG] refine_device fails to compile for idx_t != matrix_idx
Calling `raft::neighbors::refine` on input which uses a different indexing type for the input matrices than for the vector ids results in a compilation failure in `raft::neighbors::ivf_flat::detail::fill_refinement_index`. This comes up in attempting to refine output from some of our supported CAGRA [instantiations](https://github.com/rapidsai/raft/blob/branch-23.12/cpp/include/raft/neighbors/cagra.cuh#L78), which use `int64_t` for the input matrices' indexes but some other type for the vector ids.

To trace through the problem:

1. `refine` can be instantiated with different `idx_t` and `matrix_idx` [here](https://github.com/rapidsai/raft/blob/branch-23.12/cpp/include/raft/neighbors/refine-inl.cuh#L69).
2. This calls `refine_device` with the same type instantiations [here](https://github.com/rapidsai/raft/blob/branch-23.12/cpp/include/raft/neighbors/detail/refine_device.cuh#L39).
3. `refine_device` calls `fill_refinement_index` [here](https://github.com/rapidsai/raft/blob/branch-23.12/cpp/include/raft/neighbors/detail/refine_device.cuh#L83).
4. `fill_refinement_index` has only a [single](https://github.com/rapidsai/raft/blob/branch-23.12/cpp/include/raft/neighbors/detail/ivf_flat_build.cuh#L369) indexing type in its template. This is deduced to be `idx_t` from `refine_device` due to [this argument](https://github.com/rapidsai/raft/blob/branch-23.12/cpp/include/raft/neighbors/detail/refine_device.cuh#L80), but `matrix_idx` due to [this argument](https://github.com/rapidsai/raft/blob/branch-23.12/cpp/include/raft/neighbors/detail/refine_device.cuh#L49). Because these two do not match, no instantiation matches the template, and we fail to compile.

lowener:
Closes #537.
Needed change for the transition from Raft to cuVS.

### DIFF
--- a/cpp/src/neighbors/refine/refine_device.cuh
+++ b/cpp/src/neighbors/refine/refine_device.cuh
@@ -84,12 +84,13 @@ void refine_device(
   cuvs::neighbors::ivf_flat::index<data_t, idx_t> refinement_index(
     handle, cuvs::distance::DistanceType(metric), n_queries, false, true, dim);
 
-  cuvs::neighbors::ivf_flat::detail::fill_refinement_index(handle,
-                                                           &refinement_index,
-                                                           dataset.data_handle(),
-                                                           neighbor_candidates.data_handle(),
-                                                           n_queries,
-                                                           n_candidates);
+  cuvs::neighbors::ivf_flat::detail::fill_refinement_index<data_t, idx_t>(
+    handle,
+    &refinement_index,
+    dataset.data_handle(),
+    neighbor_candidates.data_handle(),
+    (idx_t)n_queries,
+    (uint32_t)n_candidates);
   uint32_t grid_dim_x = 1;
 
   // the neighbor ids will be computed in uint32_t as offset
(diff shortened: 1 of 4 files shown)


cpp/src/neighbors/refine/refine_device.cuh
```diff
@@ -84,12 +84,13 @@ void refine_device(
...
-                                                           neighbor_candidates.data_handle(),
-                                                           n_queries,
-                                                           n_candidates);
+  cuvs::neighbors::ivf_flat::detail::fill_refinement_index<data_t, idx_t>(
+    handle,
+    &refinement_index,
+    dataset.data_handle(),
+    neighbor_candidates.data_handle(),
+    (idx_t)n_queries,
+    (uint32_t)n_candidates);
```
benfred, line 93: nitpick: I think we should be using c++ style casts over c-style ones:

```suggestion
    static_cast<idx_t>(n_queries),
    static_cast<uint32_t>(n_candidates));
```

benfred (approved):
lgtm! 

left one minor comment, but pre-approving since its not a huge deal

PR #565, opened 2025-01-11
[BUG] Fix CAGRA graph optimization bug

enp1s0:
The merging process of the pruned and revedge graphs is inappropriate, resulting in neighbor index duplication. This PR fixes the issue and adds the validation of graph indices at the end of the `graph::optimze` function.

The simplest solution to fix this issue is 1151511b5e0c11675986dfe314fa525777d7849b, but it ignores the memory space optimization by https://github.com/rapidsai/raft/commit/12480cf795196cf487dda25aa77599ac5683627f. The code changes in this PR apply the memory space optimization to the simplest solution to remove the memory space for the pruned graph (`pruned_graph`).

This PR also fixes a problem that throwing an exception from an OMP loop does not work as expected.

This is the search performance comparison between the graph optimization that has a bug (branch-25.02) and fixed (fix-cagra-graph-optimization-bug). Each point represents the search performance for itopk=32, 64, ... 512. By this PR, the recall becomes slightly higher when searching with a small itopk size. Because the number of duplicated nodes by this bug is not so large (typically less than 100 in a 1M dataset), the performances are almost the same when searching with a large itopk size and traversing the graph sufficiently. 
fix-cagra-graph-optimization-bug


enp1s0:
The new `optimize` function gets stuck in `cagra_extreme_inputs_oob_test` because there are some invalid index (`~0u`) nodes in the initial knn graph. It would be better to check the invalid nodes first in the function and abort the process if there are any.

cjnolet (approved):

cjnolet:
Approving, but waiting to merge until you confirm it's ready @enp1s0

enp1s0:
@cjnolet Thank you for approving. I think it's ready to be merged.

anaruse:
> @anaruse Can you review the code again?

It looks good to me.

PR #568, opened 2025-01-13
Improve filtering documentation

lowener:
This PR add a dedicated documentation page for filtering in the `Getting started` tab, and add the `cuvs::neighbors::filtering` namespace to the C++ documentation


docs/source/filtering.rst
```diff
@@ -0,0 +1,75 @@
+.. _filtering:
+
+~~~~~~~~~~~~~~~~~~~~~~~~
+Filtering vector indexes
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+CuVS supports different type of filtering depending on the vector index being used. The main method used in all of the vector indexes
```
cjnolet, line 7: Nitpick: Please write cuVS with lowercase 'c', even at the beginning of a sentence.

docs/source/filtering.rst
```diff
@@ -0,0 +1,75 @@
...
+A bitset is an array of bits where each bit can have two possible values: `0` and `1`, which signify in the context of filtering whether
+a sample should be filtered or not. `0` means that the corresponding vector will be filtered, and will therefore not be present in the results of the search.
+This mechanism is optimized to take as little memory space as possible, and is available through the RAFT library
+(check out RAFT's `bitset API documentation <https://docs.rapids.ai/api/raft/stable/cpp_api/core_bitset/>`). When calling a search function of an ANN index, the
+bitset length should match the number of vectors present in the database.
+
+Bitmap
+======
+
+A bitmap is based on the same principle as a bitset, but in two dimensions. This allow users to give different bitset for each query that
```
cjnolet, line 23:
```suggestion
A bitmap is based on the same principle as a bitset, but in two dimensions. This allows users to give different bitset for each query that
```

docs/source/filtering.rst
```diff
@@ -0,0 +1,75 @@
...
+is getting searched. Check out RAFT's `bitmap API documentation <https://docs.rapids.ai/api/raft/stable/cpp_api/core_bitmap/>`.
```
cjnolet, line 24:
```suggestion
being searched. Check out RAFT's `bitmap API documentation <https://docs.rapids.ai/api/raft/stable/cpp_api/core_bitmap/>`.
```

docs/source/filtering.rst
```diff
@@ -0,0 +1,75 @@
...
+
+Using filters in cuVS
+=====================
+
+Example
```
cjnolet, line 29: Rust doc makes different sized heading for each different type of heading symbol used. Please follow other docs and use different symbols for these. based on the order it traverses the heading tree, it'll figure out which ones whould be H1, H2, H3, and so on, so long as you are consistent about where they are used.

cjnolet, line 29:
```suggestion
Examples
```

docs/source/filtering.rst
```diff
@@ -0,0 +1,75 @@
...
+=======
+
+Using a Bitmap filter on a Brute-force index
+--------------------------------------------
```
cjnolet, line 33: This one seems to be missing.

docs/source/filtering.rst
```diff
@@ -0,0 +1,75 @@
...
+
+
+Using a Bitset filter on a CAGRA index
```
cjnolet, line 36: Since you present the bitset and bitmap in that order above, perhaps you should also present the examples in the same order for consistency.

docs/source/filtering.rst
```diff
@@ -0,0 +1,75 @@
...
+--------------------------------------
+
+.. code-block:: c++
+
+    #include <cuvs/neighbors/cagra.hpp>
```
cjnolet, line 41: Can you add an example `cu` file in `examples/cpp` as well (assuming it doesn't exist already). That would also be a great thing to link to from these docs.

docs/source/filtering.rst
```diff
@@ -0,0 +1,75 @@
+.. _filtering:
+
+~~~~~~~~~~~~~~~~~~~~~~~~
+Filtering vector indexes
+~~~~~~~~~~~~~~~~~~~~~~~~
+
+CuVS supports different type of filtering depending on the vector index being used. The main method used in all of the vector indexes
+is pre-filtering, which is a technique that will into account the filtering of the vectors before computing it's closest neighbors, saving
+some computation from calculating distances.
```
cjnolet, line 9: Given that this is a supposed to be included in the getting started guide, I think we should provide a little more info to beginner users who might not be familiar with the "when" and "why". Adding some use-cases here could be helpful for users.

docs/source/indexes/cagra.rst
```diff
@@ -134,16 +134,20 @@ IVF-PQ Build:
 
 .. math::
 
-   n_vectors / train_set_ratio * dim * sizeof(float)   // trainset, may be in managed mem
-   + n_vectors / train_set_ratio * sizeof(uint32_t)    // labels, may be in managed mem
-   + n_clusters * n_dim * sizeof(float)                // cluster centers
+   n\_vectors / train\_set\_ratio * dim * sizeof(float)   // trainset, may be in managed mem
+
+   + n\_vectors / train\_set\_ratio * sizeof(uint32_t)    // labels, may be in managed mem
```
cjnolet, line 139: Very nice! Thanks for noticing and adding these things!

docs/source/indexes/ivfflat.rst
```diff
@@ -86,7 +86,7 @@ Memory footprint
 ----------------
 
 Each cluster is padded to at least 32 vectors (but potentially up to 1024). Assuming uniform random distribution of vectors/list, we would have
-:math:`cluster\_overhead = (conservative\_memory\_allocation ? 16 : 512 ) * dim * sizeof_{float})`
```
cjnolet, line 89: This was intentional here- we want the `float` to be a subscript for ease of readership.

cjnolet (approved):

PR #573, opened 2025-01-15
Reduce the recall threshold for IVF-PQ low-precision LUT inner product  tests

achirkin:
IVF-PQ allows to use low-precision for the lookup table during search to improve QPS. When used for the inner product distance, this has extra tall on recall. This PR reduces our expectation of the recall in this case as an answer to occasional test failures in CI.

### DIFF
--- a/cpp/test/neighbors/ann_ivf_pq.cuh
+++ b/cpp/test/neighbors/ann_ivf_pq.cuh
@@ -879,7 +879,7 @@ inline auto enum_variety_ip() -> test_cases_t
         // InnerProduct score is signed,
         // thus we're forced to used signed 8-bit representation,
         // thus we have one bit less precision
-        y.min_recall = y.min_recall.value() * 0.90;
+        y.min_recall = y.min_recall.value() * 0.88;
       } else {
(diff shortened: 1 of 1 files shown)


cpp/test/neighbors/ann_ivf_pq.cuh
```diff
@@ -879,7 +879,7 @@ inline auto enum_variety_ip() -> test_cases_t
         // InnerProduct score is signed,
         // thus we're forced to used signed 8-bit representation,
         // thus we have one bit less precision
-        y.min_recall = y.min_recall.value() * 0.90;
+        y.min_recall = y.min_recall.value() * 0.88;
```
cjnolet, line 882: Thanks for opening this PR. Can you also open a GIthub issue for this so that we don' tlose sight of it?

cjnolet (approved):

jakirkham:
Thanks Artem and Corey! 🙏

cpp/test/neighbors/ann_ivf_pq.cuh
achirkin: https://github.com/rapidsai/cuvs/issues/581

PR #574, opened 2025-01-15
Run cuvs-bench pytests and end-to-end tests in CI 

dantegd:
PR does the following:

- [x] Modifies CI to run pytest and e2e test of cuvs-bench
    - [x] We need to test the additional time needed to run the tests. They should be fast, but if they are not, then we can add an additional job to run them in parallel.
- [x] Adds synthetic test-data generation so the CI jobs don't depend on downloading datasets, and users can have easy testing locally. 
    - [ ] Few improvements to be done to docs, yaml and other things to make it easy for users.
- [x] Check in some additional pytests that hadn't been checked in before.


ci/test_python.sh
```diff
@@ -54,5 +55,21 @@ pytest \
...
+ --cache-clear \
+ --junitxml="${RAPIDS_TESTS_DIR}/junit-cuvs.xml" \
+ --cov-config=../.coveragerc \
+ --cov=cuvs \
+ --cov-report=xml:"${RAPIDS_COVERAGE_DIR}/cuvs-bench-coverage.xml" \
+ --cov-report=term \
+ tests
+
+rapids-logger "cuvs-bench e2e test with synthetic data"
+python -m cuvs_bench.get_dataset --dataset test-data --dataset-path datasets
```
dantegd, line 70: Just found out that `click` has a `Clirunner` for testing the CLI in pytest in a more robust manner, so changing these calls to that

python/cuvs_bench/cuvs_bench/config/algos/cuvs_cagra.yaml
```diff
@@ -11,3 +11,11 @@ groups:
     search:
       itopk: [32, 64, 128, 256, 512]
       search_width: [1, 2, 4, 8, 16, 32, 64]
+  test:
```
cjnolet, line 14: Love this idea!

conda/environments/bench_ann_cuda-118_arch-aarch64.yaml
```diff
@@ -37,7 +37,7 @@ dependencies:
 - libcusparse=11.7.5.86
 - libcuvs==25.2.*,>=0.0.0a0
 - librmm==25.2.*,>=0.0.0a0
-- matplotlib
+- matplotlib-base
 - nccl>=2.19
```
cjnolet, line 41: Does Scikit-learn also need to be added here?

python/cuvs_bench/cuvs_bench/get_dataset/__main__.py
```diff
@@ -80,35 +89,141 @@ def download(name, normalize, ann_bench_data_path):
...
+        # Datasets
+        f.create_dataset("train", data=train_data)
+        f.create_dataset("test", data=test_data)
+        f.create_dataset("neighbors", data=neighbors)
+        f.create_dataset("distances", data=distances)
 
-    download(args.dataset, args.normalize, args.dataset_path)
+        f.attrs["distance"] = metric
+
+    convert_hdf5_to_fbin(full_path, normalize=True)
```
cjnolet, line 140: I wonder- should we add debug prints before `download` and `convert_hdf5_to_fbin` as well, just in case they fail?

python/cuvs_bench/cuvs_bench/get_dataset/__main__.py
```diff
@@ -80,35 +89,141 @@ def download(name, normalize, ann_bench_data_path):
...
+    print(f" - distances shape = {distances.shape}")
+    print(f" - metric = {metric}")
+    print(f" - neighbors per test sample = {actual_k}")
+
+
+def get_default_dataset_path():
+    if "RAPIDS_DATASET_ROOT_DIR" in os.environ:
+        return os.getenv("RAPIDS_DATASET_ROOT_DIR")
+    else:
+        return os.path.join(os.getcwd(), "datasets")
```
jameslamb, line 155:
```suggestion
    return os.getenv("RAPIDS_DATASET_ROOT_DIR", os.path.join(os.getcwd(), "datasets"))
```

This could be simplified. `os.getenv()` exists specifically for this purpose where you want to return either the value of an environment variable or some fallback if that environment variable isn't defined.

dependencies.yaml
```diff
@@ -540,9 +540,10 @@ dependencies:
         packages:
           - click
           - cuvs==25.2.*,>=0.0.0a0
-          - matplotlib
+          - matplotlib-base
           - pandas
           - pyyaml
+          - requests
```
jameslamb, line 546: I can't quite in-browser suggest this because of the separation between the lines... but `matplotlib-base` is a conda-specific concept. There's no project by that name on PyPI (https://pypi.org/project/matplotlib-base/) and I don't think there is on other `pip` indices referenced in this build.

That dependency probably be expressed like this:

```yaml
- output_types: [conda]
  packages:
    - matplotlib-base
- output_types: [requirements, pyproject]
  packages:
    - matplotlib
```

That'll keep `matplotlib-base` out of the `cuvs-bench` `pyproject.toml`.

jameslamb (approved):
I left some minor, but they're minor so giving you a `ci-codeowners` / `packaging-codeowners` approval so you have what you need to merge. Please do consider them.

cjnolet (approved):

PR #582, opened 2025-01-16
Fix the use of constexpr in the dynamic batching header

achirkin:
Remove the `constexpr` in a function that is non-constexpr according to the C++17 rules.


bdice (approved):

cjnolet (approved):

jakirkham:
Thanks Artem and Corey! 🙏

Issue #553, opened 2024-12-30
Is it do batches of vector search?


Wongboo:
Is it do batches of vector search? Each batch contains multiple queries and multiple databases, and queries only search for corresponding database in the batch. For example:
```
import cupy as cp

batch_size = 16
n_samples = 5000
n_features = 50
n_queries = 1000
dataset = cp.random.random_sample((batch_size, n_samples, n_features),
                                  dtype=cp.float32)
# Build index
index = cagra.build(cagra.IndexParams(), dataset)
# Search using the built index
queries = cp.random.random_sample((batch_size, n_queries, n_features),
                                  dtype=cp.float32)
# doing some indexing and searchs, queries only search for corresponding database
```
I've searched whole doc and issues. Really appreciate your answer!

cjnolet:
Hi @Wongboo thanks for your patience as most of the team was on holiday last week. We have separate API docs for the Python API. Does this help? https://docs.rapids.ai/api/cuvs/nightly/python_api/neighbors_cagra/#cuvs.neighbors.cagra.search

Wongboo:
> Hi [@Wongboo](https://github.com/Wongboo) thanks for your patience as most of the team was on holiday last week. We have separate API docs for the Python API. Does this help? https://docs.rapids.ai/api/cuvs/nightly/python_api/neighbors_cagra/#cuvs.neighbors.cagra.search

Thanks, but this has a significant difference. Note my example contains an extra batch size dimension.

cjnolet:
@Wongboo, I'm not sure what you mean here, but you can pass a 2d array to the search method to query multiple vectors at a time. If you have multiple such arrays, they would need to be passed into multiple calls of search(). If you use different `device_resources` with different CUDA streams in the calls to search, you can overlap them across batches. You can also take a look at the `persistent=True` option if you'd like to improve overlap further across searches.

Wongboo:
> [@Wongboo](https://github.com/Wongboo), I'm not sure what you mean here, but you can pass a 2d array to the search method to query multiple vectors at a time. If you have multiple such arrays, they would need to be passed into multiple calls of search(). If you use different `device_resources` with different CUDA streams in the calls to search, you can overlap them across batches. You can also take a look at the `persistent=True` option if you'd like to improve overlap further across searches.
> 
[...]

PR #583, opened 2025-01-16
Adding persistent kernel params to cagra python

cjnolet:
Closes https://github.com/rapidsai/cuvs/issues/553


python/cuvs/cuvs/neighbors/cagra/cagra.pyx
```diff
@@ -363,6 +363,7 @@ cdef class SearchParams:
         more.
     rand_xor_mask: int, default = 0x128394
         Bit mask used for initial random seed node selection.
+
```
lowener, line 366: Add a description of those new parameters to the docstring

PR #586, opened 2025-01-17
Fixing small typo in cuvs bench docs

cjnolet:


benfred (approved):
changes lgtm - 

I think removing the output of `--help` from the python scripts from these docs improves readability and flow here

PR #591, opened 2025-01-20
Fix cagra_hnsw serialization when dataset is not part of index

tfeher:
After calling `build()`, ideally the CAGRA index contains both the dataset and the graph. But when we do not have sufficient device memory, then only the graph is returned. In such case we need to pass the dataset explicitly to the serialization routines.

For serialization in HNSW format, in case we have flat hierarchy, the dataset was not passed. This PR fixes this problem by adding an optional `dataset` argument to `cagra::serialize_to_hnswlib`.

Furthermore, to improve execution time, we change from writing a single element to writing a single row of the graph and dataset at time. 

Additionally, debug messages for tracking data saving time are added.


divyegala (approved):

cjnolet:
@tfeher changes look good but there's a docs build failure.

Issue #585, opened 2025-01-16
[BUG] cuvs-bench bugs


tarang-jain:
Issue to track some of the recent bugs I have seen in cuvs-bench.

1. Python API `cuvs_bench.run` builds indexes for all parameter groups despite a user selecting a particular group using the `--groups` argument
2. `raft_log_level` argument needs to be removed / updated. Currently, it is an unrecognized argument.

PR #592, opened 2025-01-20
fix cuvs_bench.run --groups options

jiangyinzuo:
issue: #585


divyegala (approved):

dantegd (approved):
Thanks for the fix @jiangyinzuo ! The patch looks good to me

PR #595, opened 2025-01-21
Add CAGRA InnerProduct test and fix a bug

enp1s0:
There is no CAGRA InnerProduct test in cuVS while there was in RAFT. This PR adds the test.

Also, this PR
- fixes an error that occurs when building a CAGRA graph with NN Descent + InnerProduct (`cpp/src/neighbors/nn_descent_*.cu`)
  - Could not pass the `idx.metric() == params.metric` validation in the `nn_descent::detail::build` function.
- updates `RAFT_EXPECT` in the `nn_descent::detail::build` function for better error understanding
- fixes the NN Descent API sample codes


cjnolet (approved):
Thanks for this fix @enp1s0!

PR #596, opened 2025-01-21
[Fix] l2_exp random fail in half-float32 mixed precision on self-neighboring

rhdong:


python/cuvs/cuvs/test/test_distance.py
```diff
@@ -21,6 +21,7 @@
 from cuvs.distance import pairwise_distance
 
 
+@pytest.mark.parametrize("times", range(20))
```
benfred, line 24: Whats this times parameter used for? I don't see it used in the test it self - 

Are you just trying to run this test multiple times here to stress test it?

python/cuvs/cuvs/test/test_distance.py
```diff
@@ -79,7 +80,5 @@ def test_distance(n_rows, n_cols, inplace, order, metric, dtype):
     actual = output_device.copy_to_host()
 
     tol = 1e-3
-    if np.issubdtype(dtype, np.float16):
-        tol = 1e-1
```
benfred, line 83: I think I added this reduced tolerance because I was seeing failures - is this no longer needed?

benfred (approved):
PR looks good to me - had a couple of minor questions though

python/cuvs/cuvs/test/test_distance.py
rhdong: Yeah, I tried it successfully at local machines. I think this change can help us block potential actual failures in the future, so I made it.

python/cuvs/cuvs/test/test_distance.py
rhdong: Yeah, it's just for testing multiple times and to guarantee the reproducing on one going because the possibility is close to ~10% empirically.

bdice:
Looks like Python tests are now passing. We are waiting on one more C++ test job. I will go ahead and trigger a merge once CI finishes to unblock CI, since several PRs depend on this.

PR #602, opened 2025-01-22
Reduce CAGRA test runtime

bdice:
Currently, running `NEIGHBORS_ANN_CAGRA_TEST` takes:
[0.96 hours on CUDA 11.8, V100 (x86)](https://github.com/rapidsai/cuvs/actions/runs/12913409417/job/36012418022?pr=596#step:8:1718)
[1.59 hours on CUDA 12.5, V100 (x86)](https://github.com/rapidsai/cuvs/actions/runs/12913409417/job/36012418329?pr=596#step:8:492)
[0.28 hours on CUDA 12.0, A100 (ARM)](https://github.com/rapidsai/cuvs/actions/runs/12913409417/job/36012418741?pr=596#step:8:1729)

Individual tests should be able to complete in less than an hour. Ideally, less than 10 minutes.

This PR proposes some changes to CAGRA tests:
- Each CAGRA type is now its own test executable (e.g. `NEIGHBORS_ANN_CAGRA_FLOAT_UINT32_TEST`)
- Some parameter combinations were trimmed by ~50%


bdice:
@tfeher @achirkin, I'd appreciate your thoughts on the parameter reductions here. We don't want to lose anything that is important for coverage, and @cjnolet suggested you would be good reviewers to ensure we retain good coverage.

If we can be even more aggressive in our parameter trimming, that is also useful input!

cpp/test/neighbors/ann_cagra.cuh
```diff
@@ -1006,21 +1014,22 @@ inline std::vector<AnnCagraInputs> generate_inputs()
...
-    {1},
-    {cuvs::distance::DistanceType::L2Expanded},
-    {false},
-    {false},
-    {0.995});
+  // Varying dim, adding non_owning_memory_buffer_flag
+  inputs2 =
+    raft::util::itertools::product<AnnCagraInputs>({100},
+                                                   {1000},
+                                                   {1, 5, 8, 64, 137, 256, 619, 1024},  // dim
```
achirkin, line 1021: Here we test a few edge cases where the dimensionality is `2ⁿ` or `2ⁿ ± 1`. These are important to check whether we have any problems with padding the data during build and search.

cpp/test/neighbors/ann_cagra.cuh
```diff
@@ -987,11 +994,12 @@ inline std::vector<AnnCagraInputs> generate_inputs()
...
   }
 
-  // refinement options
+  // Refinement options
+  // Varying host_dataset, ivf_pq_search_refine_ratio
   inputs2 =
     raft::util::itertools::product<AnnCagraInputs>({100},
                                                    {5000},
-                                                   {32, 64},
+                                                   {64},
```
achirkin, line 1002: I'm not sure about this one either. Without this, we'd only be checking refinement in a single index configuration (host/device dataset + several refinement variants).

cpp/test/neighbors/ann_cagra.cuh
```diff
@@ -876,14 +876,15 @@ class AnnCagraFilterTest : public ::testing::TestWithParam<AnnCagraInputs> {
 inline std::vector<AnnCagraInputs> generate_inputs()
 {
   // TODO(tfeher): test MULTI_CTA kernel with search_width > 1 to allow multiple CTA per queries
+  // Varying dim, k, graph_build_algo, search_algo, max_queries
```
achirkin, line 879: I love the extra annotations on what are we varying. Thanks!

achirkin (changes requested):
Thanks, @bdice, for the PR! The long CAGRA test times has long been a problem and I totally agree we have to do something with it.

I have nothing against splitting the CAGRA test into multiple executables.

However, I'm a bit hesitating to reduce the test cases via the `generate_inputs()`. Most of these were set during development for various (legit, but sometimes forgotten) reasons.

I think, instead of reducing the base set of tests, we can go couple other ways and have a greater impact on test times without significantly reducing the coverage:

1. We can drastically reduce the number of cases in the `AnnCagraAddNodesTest` / `AnnCagraFilterTest`, while keeping the base `AnnCagraTest` as is. For example, we can remove all variants changing the build algorithm and refine ratios for the two suites as these parameters has no effect on them. We can copy-paste the `generate_inputs()` to `generate_inputs_limited()` and manually reduce the set by more than 50% of cases. Or we can simply filter out every second variant from the vector (knowing we've already tested the same combination in the base test).
2. A more involved refactoring (perhaps for a follow-up): I believe we can share some of the state between the tests rather than generating everything from scratch. E.g. generate a blob of data big enough only once and slice/reuse it for all tests in a suite. Or keep an index between the cases varying only search parameters.

cpp/test/neighbors/ann_cagra.cuh
bdice: I figured that was the case. There were a couple places that varied `dim`. I kept different values in each test so that the full range would be covered. Perhaps we can cover the full range in only one set of inputs instead, and use a small range for the other cases?

cpp/test/neighbors/ann_cagra.cuh
bdice: Happy to revert.
```suggestion
                                                   {32, 64},
```

tfeher:
@bdice Thank you for your proposal. I am preparing the changes, and I will push them soon.

cpp/test/neighbors/ann_cagra.cuh
```diff
@@ -876,14 +876,15 @@ class AnnCagraFilterTest : public ::testing::TestWithParam<AnnCagraInputs> {
 inline std::vector<AnnCagraInputs> generate_inputs()
 {
   // TODO(tfeher): test MULTI_CTA kernel with search_width > 1 to allow multiple CTA per queries
+  // Charge graph dim, search algo and max_query parameter
   std::vector<AnnCagraInputs> inputs = raft::util::itertools::product<AnnCagraInputs>(
     {100},
     {1000},
     {1, 8, 17},
-    {1, 16},  // k
```
tfeher, line 883: Dim and build algo combinations are tested below, therefor we focus on dim and search algo and max_query parameter value here.

cpp/test/neighbors/ann_cagra.cuh
```diff
@@ -892,11 +893,30 @@ inline std::vector<AnnCagraInputs> generate_inputs()
     {true},
     {0.995});
 
+  // Fixed dim, and changing neighbors and query size (output matrix size)
   auto inputs2 = raft::util::itertools::product<AnnCagraInputs>(
+    {1, 100},
```
tfeher, line 898: Above we tested with different `max_query` parameter, here we test with batch size 1 and 100.

cpp/test/neighbors/ann_cagra.cuh
```diff
@@ -1006,29 +1032,167 @@ inline std::vector<AnnCagraInputs> generate_inputs()
...
+  inputs.insert(inputs.end(), inputs2.begin(), inputs2.end());
+
+  // a few PQ configurations
   inputs2 = raft::util::itertools::product<AnnCagraInputs>(
     {100},
-    {1000},
-    {1, 3, 5, 7, 8, 17, 64, 128, 137, 192, 256, 512, 619, 1024},  // dim
-    {10},
+    {10000},
+    {192, 1024},  // dim
```
tfeher, line 1100: When testing AddNode, I assumed it is fine to limit to a smaller set of dimensions.

tfeher (commented):
Running the AddNode and Filtering tests on a subset of input makes a large difference. 

I have also applied a refactoring on the basic parameter combinations. This also gave a smaller extra improvement. @achirkin please have a look, if you prefer to keep some of the combinations then let me know.

Testing on H100, this runs 3.7x faster than the previously.

cpp/test/neighbors/ann_cagra/test_float_uint32_t.cu
bdice, line 1: Please apply these changes to `ann_cagra/test_half_uint32_t.cu` as well.

bdice (commented):
@tfeher Thanks so much! Your changes look fine to me, aside from one comment. It looks like `ann_cagra/test_half_uint32_t.cu` might have been missed in the refactor.

cpp/test/neighbors/ann_cagra/test_float_uint32_t.cu
bdice: Oh, nevermind! I see there are no AddNode / Filtering tests in that file. My apologies.

achirkin (approved):
Thanks, @tfeher, for the updates! Nice to see such a big speedup. Did you have a chance to check the difference in terms of the total number of test cases out of curiosity? 

I'd personally prefer to keep the original `generate_inputs()` unchanged and see how other improvements reduce the test time - just to be on the safe side. But the changes look reasonable, so I don't see a point in holding up this PR.

bdice:
@tfeher @achirkin I merged in the upstream to resolve merge conflicts from #595. I tried to reflect that PR's intent (adding a bunch of `InnerProduct` tests) but I would like your eyes one more time to make sure that I did what you expected w.r.t. this PR and #595. I'll also request a build codeowner review.

divyegala (approved):

cpp/test/neighbors/ann_cagra.cuh
```diff
@@ -1173,7 +1173,7 @@ inline std::vector<AnnCagraInputs> generate_filtering_inputs()
     {0},
     {64},
     {1},
-    {cuvs::distance::DistanceType::L2Expanded},
+    {cuvs::distance::DistanceType::L2Expanded, cuvs::distance::DistanceType::InnerProduct},
```
achirkin, line 1176: This test fails due to PQ compression not yet supporting InnerProduct distance.
```suggestion
    {cuvs::distance::DistanceType::L2Expanded},
```

achirkin (approved):
LGTM

bdice:
Thanks so much @achirkin!

PR #612, opened 2025-01-24
Iteratively build graph index

anaruse:
This PR is about how CAGRA's search() and optimize() can be used to iteratively create and improve graph index.

Currently, IVFPQ and NND are used to create the initial kNN graph, which is then optimized to create the CAGRA search graph. So, for example, if you want to support a new data type in CAGRA, you need to create an initial kNN graph with that data type, and IVFPQ or NND must also support that new data type. This is a bit of hassle.

This PR is one solution to that problem. With functionality of this PR, once the CAGRA search supports the new data type, it can be used to create a graph index with it.


anaruse:
This PR is related to https://github.com/rapidsai/cuvs/pull/610

anaruse:
Conflicts with branch-25.02 have been fixed.

cjnolet:
@anaruse can you install precommit and run your changes through the hooks? That'll get your formatting passing the style checker. 

Guide to install the commit hooks is [here](https://docs.rapids.ai/api/cuvs/nightly/contributing/#python-pre-commit-hooks).

anaruse:
> @anaruse can you install precommit and run your changes through the hooks? That'll get your formatting passing the style checker.
> 
> Guide to install the commit hooks is [here](https://docs.rapids.ai/api/cuvs/nightly/contributing/#python-pre-commit-hooks).

I've already install and set up the precommit. I'm not sure how this happened. I'll fix it anyway.

cpp/src/neighbors/detail/cagra/cagra_build.cuh
```diff
@@ -445,38 +442,183 @@ index<T, IdxT> build(
...
+        raft::copy(batch_neighbors_view.data_handle(),
+                   batch_dev_neighbors_view.data_handle(),
+                   batch_neighbors_view.size(),
+                   raft::resource::get_cuda_stream(res));
+      }
+
+      // Optimize graph
+      bool flag_last  = (curr_graph_size == final_graph_size);
+      curr_graph_size = curr_query_size;
+      cagra_graph     = raft::make_host_matrix<IdxT, int64_t>(curr_graph_size, curr_graph_degree);
```
tfeher, line 581: There are cases when we work with limited host mem, therefore it might make sense to delete the existing graph before we allocate the new one. 
```suggestion
cagra_graph = raft::make_host_matrix<IdxT, int64_t>(0, 0); // delete existing graph
cagra_graph = raft::make_host_matrix<IdxT, int64_t>(curr_graph_size, curr_graph_degree);
```
Alternatively, it might be even better to allocate the final graph in advance before the while loop, and here just construct a view for the actual graph size. This way host OOM could be caught earlier in the process.

cpp/src/neighbors/detail/cagra/cagra_build.cuh
```diff
@@ -445,38 +442,183 @@ index<T, IdxT> build(
...
+    // CAGRA's search() and optimize(). As for the size of the graph, instead
+    // of targeting all nodes from the beginning, the number of nodes is
+    // initially small, and the number of nodes is doubled with each iteration.
+    RAFT_LOG_INFO(
+      "Iteratively creating/improving graph index using CAGRA's search() and optimize()");
+
+    // If dataset is a host matrix, change it to a device matrix. Also, if the
+    // dimensionality of the dataset does not meet the alighnemt restriction,
+    // add extra dimensions and change it to a strided matrix.
+    auto dev_aligned_dataset      = make_aligned_dataset(res, dataset);
```
tfeher, line 460: This will OOM for large dataset, could we throw an informative error message in that case?
```suggestion
  std::unique_ptr<strided_dataset<T, int64_t>> dev_aligned_dataset;
try {
  dev_aligned_dataset = make_aligned_dataset(res, dataset);
} catch (raft::logic_error& e) {
  RAFT_LOG_ERROR("Iterative CAGRA build requires the dataset to fit GPU memory");
  throw e;
}
```

cpp/include/cuvs/neighbors/cagra.hpp
```diff
@@ -104,7 +108,8 @@ struct index_params : cuvs::neighbors::index_params {
    */
   std::variant<std::monostate,
                graph_build_params::ivf_pq_params,
-               graph_build_params::nn_descent_params>
+               graph_build_params::nn_descent_params,
```
cjnolet, line 111: We have explicit docs for each of these arguments. I understand the iterative search params are experimental to start, but can we at least add them to the docs?

tfeher (changes requested):
Thanks Akira for this PR, it is very nice to see this feature in cuvs. The PR looks great overall, I just have a few smaller suggestion.

cjnolet:
@anaruse given my comment above about the added comment in the code, I'll go ahead and merge this PR for now and perhaps @enp1s0 can add the comment in his bitwise hamming PR.

cpp/include/cuvs/neighbors/cagra.hpp
anaruse: Thanks, I will add comments about iterative_search_params here.

cpp/src/neighbors/detail/cagra/cagra_build.cuh
```diff
@@ -445,38 +442,183 @@ index<T, IdxT> build(
...
-  if (std::holds_alternative<cagra::graph_build_params::ivf_pq_params>(knn_build_params)) {
-    auto ivf_pq_params =
-      std::get<cuvs::neighbors::cagra::graph_build_params::ivf_pq_params>(knn_build_params);
-    build_knn_graph(res, dataset, knn_graph->view(), ivf_pq_params);
-  } else {
-    auto nn_descent_params =
-      std::get<cagra::graph_build_params::nn_descent_params>(knn_build_params);
+  if (std::holds_alternative<cagra::graph_build_params::iterative_search_params>(
+        knn_build_params)) {
+    // Iteratively improve the accuracy of the graph by repeatedly running
```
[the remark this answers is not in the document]
anaruse: OK, I will create function `iterative_build_graph`, and move the related code there.

cpp/src/neighbors/detail/cagra/cagra_build.cuh
```diff
@@ -445,38 +442,183 @@ index<T, IdxT> build(
...
+    // Determine graph degree and number of search results while increasing
+    // graph size.
+    auto small_graph_degree = std::max(graph_degree / 2, std::min(graph_degree, (uint64_t)32));
+    auto small_topk         = topk * small_graph_degree / graph_degree;
+    RAFT_LOG_DEBUG("# graph_degree = %lu", (uint64_t)graph_degree);
+    RAFT_LOG_DEBUG("# small_graph_degree = %lu", (uint64_t)small_graph_degree);
+    RAFT_LOG_DEBUG("# topk = %lu", (uint64_t)topk);
+    RAFT_LOG_DEBUG("# small_topk = %lu", (uint64_t)small_topk);
+
+    // Create an initial graph. This can be inaccurate.
```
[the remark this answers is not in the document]
anaruse: Yes, connectivity is guaranteed in the initial graphs created here. I will add a comment to the source code as well.

cpp/src/neighbors/detail/cagra/cagra_build.cuh
anaruse: Thanks, will fix in that way.

cpp/src/neighbors/detail/cagra/cagra_build.cuh
anaruse: Thanks, will fix in that way as well.

cpp/include/cuvs/neighbors/cagra.hpp
```diff
@@ -72,6 +72,10 @@ struct ivf_pq_params {
 };
 
 using nn_descent_params = cuvs::neighbors::nn_descent::index_params;
+
+// **** Experimental ****
+using iterative_search_params = cuvs::neighbors::search_params;
```
achirkin, line 77: As I understand, `cuvs::neighbors::search_params` is being re-used here as an empty tag for the `std::variant` index parameters below.
While this is a perfectly valid approach, I wonder  whether it would look confusing to a user and whether it would cause us any troubles in future (if at some point we decide to add any members to `cuvs::neighbors::search_params`).
An alternative would be to just replace it with `struct iterative_search_params {}`.
What do you think, @cjnolet ?

achirkin (approved):
Thanks, the changes look very good and clean to me

tfeher (approved):
Thanks Akira for the updates, the PR looks good to me.

PR #615, opened 2025-01-24
add runtime dependency on libcuvs in cuvs wheels

jameslamb:
`cuvs-cu{11,12}` wheels don't currently have a runtime dependency on `libcuvs-cu{11,12}`. They need one, for library-loading:

https://github.com/rapidsai/cuvs/blob/e9983e17408e6bec6f2558f9df49be97a7255417/python/cuvs/cuvs/__init__.py#L19-L25

This was missed in #594. This PR adds it.

## Notes for Reviewers

Adding for searchability... this bug can result in issues like this at runtime when using `cuvs` installed from wheels:

> ImportError: libcuvs_c.so: cannot open shared object file: No such file or directory


vyasr (approved):
Whoops...

PR #617, opened 2025-01-24
Add Scalar Quantization to the c and python apis

benfred:


cpp/include/cuvs/preprocessing/quantize/scalar.h
```diff
@@ -0,0 +1,130 @@
...
+/**
+ * @brief Perform inverse quantization step on previously quantized dataset
+ *
+ * Note that depending on the chosen data types train dataset the conversion is
+ * not lossless.
+ *
+ * @param[in] res raft resource
+ * @param[in] quantizer a scalar quantizer
+ * @param[in] dataset a row-major matrix
+ * @param[out] out a row-major matrix
```
cjnolet, line 120: Can we specify whether this is on device or host (or both)?

cpp/include/cuvs/preprocessing/quantize/scalar.h
```diff
@@ -0,0 +1,130 @@
...
+                                     DLManagedTensor* dataset,
+                                     cuvsScalarQuantizer_t quantizer);
+
+/**
+ * @brief Applies quantization transform to given dataset
+ *
+ * @param[in] res raft resource
+ * @param[in] quantizer a scalar quantizer
+ * @param[in] dataset a row-major matrix to transform
+ * @param[out] out a row-major matrix to store transformed data
```
cjnolet, line 104: Same here- we should be specifying both for the inputs and the outputs whether or not it's on device or host (or either/both).

python/cuvs/cuvs/preprocessing/quantize/scalar/scalar.pyx
```diff
@@ -0,0 +1,232 @@
...
+
+@auto_sync_resources
+def train(QuantizerParams params, dataset, resources=None):
+    """
+    Initializes a scalar quantizer to be used later for quantizing the dataset.
+
+    Parameters
+    ----------
+    params : QuantizerParams object
+    dataset : row major dataset
```
cjnolet, line 93: Here too we should be specifying whether it can exist on decive, host, or both (ideally we should be able to support both).

python/cuvs/cuvs/preprocessing/quantize/scalar/scalar.pyx
```diff
@@ -0,0 +1,232 @@
...
+@auto_sync_resources
+@auto_convert_output
+def transform(Quantizer quantizer, dataset, output=None, resources=None):
+    """
+    Applies quantization transform to given dataset
+
+    Parameters
+    ----------
+    quantizer : trained Quantizer object
+    dataset : row major dataset to transform
```
cjnolet, line 140: Here too- device, host, or both.

python/cuvs/cuvs/preprocessing/quantize/scalar/scalar.pyx
```diff
@@ -0,0 +1,232 @@
...
+    """
+    Perform inverse quantization step on previously quantized dataset
+
+    Note that depending on the chosen data types train dataset the conversion
+    is not lossless.
+
+    Parameters
+    ----------
+    quantizer : trained Quantizer object
+    dataset : row major dataset to transform
```
cjnolet, line 200: device/host/both

cpp/include/cuvs/preprocessing/quantize/scalar.h
benfred: we accept both host or device matrices here - I've added docs for this in the last commit

cjnolet (approved):
Looks great!

PR #620, opened 2025-01-27
Fix indexing bug when using parallelism to build CPU hierarchy in HNSW

divyegala:
hnswlib uses an internal indexing system which assigns an ID to points, atomically, in-order that they are added to the index. When using parallelism to add points to the index, the internal ID may be different than the "label" of the point (label, for us, is just the index of the row in the dataset) as a consequence of adding points in-parallel in no deterministic order.

The bug was that I was using the label itself to write out the CPU hierarchy, when I should have been using hnswlib's internal ID for the point associated with that label.


divyegala:
Build times before bug was fixed:
image

Built times with this PR:
image

cjnolet (approved):

navneet1v:
@divyegala does this similar problem exist with Cagra index getting converted to Faiss enabled HNSW?

divyegala:
@navneet1v no, this bug is not present in FAISS

cjnolet (approved):

cjnolet (approved):

PR #635, opened 2025-01-31
Expose NN-Descent to C and Python

benfred:


python/cuvs/cuvs/neighbors/nn_descent/nn_descent.pxd
```diff
@@ -0,0 +1,68 @@
...
+        float termination_threshold
+        bool return_distances
+        size_t n_clusters
+
+    ctypedef cuvsNNDescentIndexParams* cuvsNNDescentIndexParams_t
+
+    ctypedef struct cuvsNNDescentSearchParams:
+        uint32_t n_probes
+
+    ctypedef cuvsNNDescentSearchParams* cuvsNNDescentSearchParams_t
```
tfeher, line 43: Do we really have search params?

tfeher (changes requested):
I had a look at the PR, it looks nice overall, just a few comment.

python/cuvs/cuvs/neighbors/nn_descent/nn_descent.pxd
benfred: good catch - not needed here

python/cuvs/cuvs/neighbors/nn_descent/nn_descent.pyx
```diff
@@ -0,0 +1,232 @@
...
+    """
+    Build the NN-Descent index from the dataset
+
+    Parameters
+    ----------
+    index_params : :py:class:`cuvs.neighbors.nn_descent.IndexParams`
+    dataset : Array interface compliant matrix, on either host or device memory
+        Supported dtype [float, int8, uint8]
+    {resources_docstring}
+
```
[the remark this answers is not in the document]
benfred: thanks! the graph here can only be on the host - and I've updated the doctring (and updated the python unittest to test this out)

tfeher (approved):
Thanks Ben for the update, the PR looks good to me.

cpp/src/neighbors/nn_descent_c.cpp
```diff
@@ -0,0 +1,163 @@
...
+    } else {
+      RAFT_FAIL("Unsupported nn-descent dataset dtype: %d and bits: %d", dtype.code, dtype.bits);
+    }
+  });
+}
+
+extern "C" cuvsError_t cuvsNNDescentIndexParamsCreate(cuvsNNDescentIndexParams_t* params)
+{
+  return cuvs::core::translate_exceptions([=] {
+    *params = new cuvsNNDescentIndexParams{.metric                    = L2Expanded,
```
cjnolet, line 129: If we're going to duplicate the params structs for now, can we at least use constants to specify the defaults? I can see a lot of confusion and long debugging sessions happening as a result of these not getting synchronized between the C++ and C APIs.

benfred: how about we have the canonical defaults in the c++ struct - and then pull out of there ?  Something like : https://github.com/rapidsai/cuvs/pull/635/commits/19deed988e2c2b8474d076f4c867ed5210f7a0e7

cjnolet (approved):
LGTM!

PR #636, opened 2025-01-31
WIP: disallow fallback to Make in wheel builds

jameslamb:
Contributes to https://github.com/rapidsai/build-planning/issues/146

Proposes:

* setting `[tool.scikit-build].ninja.make-fallback = false`, so `scikit-build-core` will not silently fallback to using GNU Make if `ninja` is not available
* including all groups in `dependencies.yaml` with `output_types: pyproject` in `output_types: requirements` as well.
  - *So that, for example, they show up in the solved environment in RAPIDS devcontainers (https://github.com/rapidsai/devcontainers)*


dependencies.yaml
```diff
@@ -397,7 +397,7 @@ dependencies:
               - *libcusparse114
   cuda_wheels:
     specific:
-      - output_types: pyproject
+      - output_types: [requirements, pyproject]
```
jameslamb, line 400: This should be fine, for the reasons mentioned in https://github.com/rapidsai/raft/pull/2563/files#r1938056656

bdice (approved):

PR #637, opened 2025-02-01
Add float16 support in python for cagra/brute_force/ivf_pq and scalar quantizer

benfred:


benfred:
note that:
* ivf_flat doesn't support float16 in the c++ api - so hasn't been added here
* cagra doesn't support float16 in the 'extend' or 'serializeToHnswLIb' functionality in c++ (which aren't currently exposed to python)(
* refine/pairwise_distance already supported float16 in python ( https://github.com/rapidsai/cuvs/pull/547 etc)

cjnolet (approved):

PR #639, opened 2025-02-01
Fix std::lock_guard use for gcc 14 support

enp1s0:
The constructor of `std::lock_guard` in gcc 14 <mutex>, which is supported by CUDA 12.8, has the `nodiscard` attribute. This PR fixes the error occurring with gcc 14.


enp1s0:
@achirkin Could you review the code?

cjnolet (approved):

achirkin (approved):
Looks fine to me

achirkin (approved):

PR #644, opened 2025-02-03
Use new rapids-logger library

vyasr:
Contributes to https://github.com/rapidsai/build-planning/issues/104.


cjnolet (approved):

docs/source/developer_guide.md
```diff
@@ -227,61 +227,6 @@ Keep in mind that this only applies to files tracked by git that have been modif
 ## Error handling
 Call CUDA APIs via the provided helper macros `RAFT_CUDA_TRY`, `RAFT_CUBLAS_TRY` and `RAFT_CUSOLVER_TRY`. These macros take care of checking the return values of the used API calls and generate an exception when the command is not successful. If you need to avoid an exception, e.g. inside a destructor, use `RAFT_CUDA_TRY_NO_THROW`, `RAFT_CUBLAS_TRY_NO_THROW ` and `RAFT_CUSOLVER_TRY_NO_THROW`. These macros log the error but do not throw an exception.
 
-## Logging
```
jameslamb, line 230: 🎉 thanks for removing this stuff carried over from RAFT.

cpp/include/cuvs/cluster/kmeans.hpp
```diff
@@ -20,6 +20,7 @@
 #include <raft/core/logger.hpp>
 #include <raft/core/resources.hpp>
 #include <raft/random/rng_state.hpp>
+#include <rapids_logger/logger.hpp>
```
jameslamb, line 23: Should this usage mean `cuvs` should pick up build and runtime dependencies on `rapids-logger`? Or are we ok with transitively getting it from RAFT because this is the only place and it's in a tightly-coupled-to-RAFT part of the codebase?

jameslamb (approved):
Left one question, but pre-emptively approving so you can merge when you've addressed it + remove the CI-only testing stuff, without needing another review from me.

cpp/include/cuvs/cluster/kmeans.hpp
vyasr: Yeah that was pretty much my reasoning. The include is itself guaranteed to be pulled in by the raft logger header, which we also have above, and we're passing the verbosity to raft's logger, so I felt OK with this.

bdice (approved):

PR #647, opened 2025-02-04
Expose configuration singleton as a global context for ann-bench algos

achirkin:
Expose the `bench::configuration` as a global context (singleton), so that algorithms can inspect it during setup if the need arises.

This is a proposal for https://github.com/rapidsai/cuvs/pull/260 to avoid mixing algorithm-specific logic into the common benchmarking infrastructure.


tfeher (approved):
LGTM

PR #658, opened 2025-02-05
[WIP][Fix] duplicated node issue on Cagra NN-descent

rhdong:
- Try to fix issue: https://github.com/rapidsai/cuvs/issues/626

### DIFF
--- a/cpp/src/neighbors/detail/nn_descent.cuh
+++ b/cpp/src/neighbors/detail/nn_descent.cuh
@@ -972,7 +972,13 @@ int insert_to_ordered_list(InternalID_t<Index_t>* list,
   int idx_insert      = width;
   bool position_found = false;
   for (int i = 0; i < width; i++) {
-    if (list[i].id() == neighb_id.id()) { return width; }
+    if (list[i].id() == neighb_id.id()) {
+      if (dist_list[i] == std::numeric_limits<DistData_t>::max() && dist != dist_list[i]) {
+        idx_insert   = i;
+        dist_list[i] = dist;
+      }
+      return idx_insert;
+    }
     if (!position_found && dist_list[i] > dist) {
       idx_insert     = i;
       position_found = true;
@@ -1044,6 +1050,7 @@ void GnndGraph<Index_t>::sample_graph_new(InternalID_t<Index_t>* new_neighbors, 
 template <typename Index_t>
 void GnndGraph<Index_t>::init_random_graph()
 {
+  const bool small_dataset = (nrow <= 4 * segment_size);
   for (size_t seg_idx = 0; seg_idx < static_cast<size_t>(num_segments); seg_idx++) {
     // random sequence (range: 0~nrow)
     // segment_x stores neighbors which id % num_segments == x
@@ -1057,14 +1064,17 @@ void GnndGraph<Index_t>::init_random_graph()
       size_t base_idx      = i * node_degree + seg_idx * segment_size;
       auto h_neighbor_list = h_graph + base_idx;
       auto h_dist_list     = h_dists.data_handle() + base_idx;
+      size_t idx           = base_idx;
       for (size_t j = 0; j < static_cast<size_t>(segment_size); j++) {
-        size_t idx = base_idx + j;
         Index_t id = rand_seq[idx % rand_seq.size()] * num_segments + seg_idx;
         if ((size_t)id == i) {
-          id = rand_seq[(idx + segment_size) % rand_seq.size()] * num_segments + seg_idx;
+          idx = small_dataset ? (idx + 1) : (idx + segment_size);
+          id  = rand_seq[idx % rand_seq.size()] * num_segments + seg_idx;
         }
+
         h_neighbor_list[j].id_with_flag() = id;
         h_dist_list[j]                    = std::numeric_limits<DistData_t>::max();
+        idx++;
       }
     }
   }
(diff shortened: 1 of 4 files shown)


cpp/src/neighbors/detail/nn_descent.cuh
```diff
@@ -1057,14 +1064,17 @@ void GnndGraph<Index_t>::init_random_graph()
       size_t base_idx      = i * node_degree + seg_idx * segment_size;
       auto h_neighbor_list = h_graph + base_idx;
       auto h_dist_list     = h_dists.data_handle() + base_idx;
+      size_t idx           = base_idx;
       for (size_t j = 0; j < static_cast<size_t>(segment_size); j++) {
-        size_t idx = base_idx + j;
         Index_t id = rand_seq[idx % rand_seq.size()] * num_segments + seg_idx;
         if ((size_t)id == i) {
-          id = rand_seq[(idx + segment_size) % rand_seq.size()] * num_segments + seg_idx;
+          idx = small_dataset ? (idx + 1) : (idx + segment_size);
```
rhdong, line 1071: Hi @cjnolet, @achirkin , @anaruse @tfeher , the second issue I'm trying to resolve here, particularly in a small dataset where the dataset size is less than 32 (the segment size), is minimizing the likelihood of self-nodes in the initial graph.

cpp/tests/neighbors/ann_cagra.cuh
```diff
@@ -944,8 +944,26 @@ inline std::vector<AnnCagraInputs> generate_inputs()
     {true},
     {0.995});
 
+  // Corner case for small dataset
+  auto inputs = raft::util::itertools::product<AnnCagraInputs>(
+    {2},
+    {3, 31, 32, 64, 101},
```
rhdong, line 950: After these changes, the cases can't passed thoroughly: on the size=64, the log is below(but better than original imp):
```
Couldn't find the output_graph_degree (61m vs pk 54) smallest detourable count nodes for node 48 in the rank-based node reranking process

Couldn't find the output_graph_degree (61m vs pk 57) smallest detourable count nodes for node 37 in the rank-based node reranking process

Couldn't find the output_graph_degree (61m vs pk 59) smallest detourable count nodes for node 4 in the rank-based node reranking process

Couldn't find the output_graph_degree (61m vs pk 58) smallest detourable count nodes for node 5 in the rank-based node reranking process

Couldn't find the output_graph_degree (61m vs pk 60) smallest detourable count nodes for node 20 in the rank-based node reranking process

Couldn't find the output_graph_degree (61m vs pk 59) smallest detourable count nodes for node 61 in the rank-based node reranking process

Couldn't find the output_graph_degree (61m vs pk 58) smallest detourable count nodes for node 39 in the rank-based node reranking process

Couldn't find the output_graph_degree (61m vs pk 59) smallest detourable count nodes for node 31 in the rank-based node reranking process
```

enp1s0:
CC: @RayWang96

cpp/src/neighbors/detail/nn_descent.cuh
tfeher: In the previous indexing, `id` was dependent on the loop variable `j`, this is not the case anymore. Therefore I do not think this is correct.

I do not understand why this would help.

tfeher (changes requested):
Thanks @rhdong for working on this!

The NN descent algorithm requires a minimum graph_degree of 32. This also implies other constraints on the minimum number of vectors for which NN descent is expected to work.

One way to fix this is to update the internals of NN descent, to remove these restrictions. This PR goes into this direction. But I believe it would take some more work to iron out all the issues. Therefore I would create a bug to track that:  Fix NN descent for small graphs.

According to the PR description, we are trying to fix #626. Instead of fixing NN descent, we could just switch to other build method to fix that.  E.g. If n_rows <=64, then we could just switch to IVF_PQ build method. But in practice it would be even better to just create a fully connected graph in that case, sort it and optimize:
``` c++
if (n_rows <= 64) { // need to test what is the right limit
 knn_graph = make_host_matrix<uint32_t, int64_t>(n_rows, n_rows-1);
  // init dense graph
  for (int i=0; i<n_rows; i++)
    for (int j=0; j<n_rows-1, j++)
      knn_graph(i, j) = (i + j + 1) % n_rows; // omit self connection
  // The optimize method expects a sorted graph.
  sort_knn_graph(res, dataset, knn_graph.view());
}
...
optimize(res, knn_graph, cagra_graph);
```

cpp/src/neighbors/detail/nn_descent.cuh
rhdong: It partially worked because the duplicated nodes will not be unique in the sequent code, which will cause the duplicated node in the output graph in small dataset scenarios. Now, I know this change is not the last resolution, and it only resolves a super small dataset problem in the new cases added. But the conclusion is not changed: I'm pretty sure this `init_random_graph` has a bug, because it assumes the `nrow >> node_degree`:
```
std::vector<Index_t> rand_seq(nrow / num_segments); // when nrow=34, num_segments=2, the 2 segments will repeat in a `0-17`, full of repeated nodes as neighbors.
```
. I'm trying to provide the right change.

rhdong: The result of an improper initial graph will affect the consequence sample process and the last output graph.

PR #659, opened 2025-02-05
Use large workspace resource for balanced kmeans

csadorf:
Switch to using `get_large_workspace_resource` instead of `get_workspace_resource` in balanced KMeans code to ensure that we can run it on very large datasets. Previously, large datasets would run into `limited_adaptor_resource_limitation` errors.

**This change may lead to a performance regression!**

Closes https://github.com/rapidsai/cuvs/issues/682 .

### DIFF
--- a/cpp/src/cluster/detail/kmeans_balanced.cuh
+++ b/cpp/src/cluster/detail/kmeans_balanced.cuh
@@ -1044,6 +1044,7 @@ void build_hierarchical(const raft::resources& handle,
 
   // Need to use explicit managed_memory here since corresponding allocations
   // must be both host and device accessible.
+  // TODO: Remove the explicit managed memory- we shouldn't be creating this on the user's behalf.
   rmm::mr::managed_memory_resource managed_memory;
   rmm::device_async_resource_ref device_memory = raft::resource::get_large_workspace_resource(handle);
   auto [max_minibatch_size, mem_per_row] =
--- a/cpp/src/cluster/detail/kmeans_balanced.cuh
+++ b/cpp/src/cluster/detail/kmeans_balanced.cuh
@@ -1042,9 +1042,10 @@ void build_hierarchical(const raft::resources& handle,
   IdxT n_mesoclusters = std::min(n_clusters, static_cast<IdxT>(std::sqrt(n_clusters) + 0.5));
   RAFT_LOG_DEBUG("build_hierarchical: n_mesoclusters: %u", n_mesoclusters);
 
-  // TODO: Remove the explicit managed memory- we shouldn't be creating this on the user's behalf.
+  // Need to use explicit managed_memory here since corresponding allocations
+  // must be both host and device accessible.
   rmm::mr::managed_memory_resource managed_memory;
-  rmm::device_async_resource_ref device_memory = raft::resource::get_workspace_resource(handle);
+  rmm::device_async_resource_ref device_memory = raft::resource::get_large_workspace_resource(handle);
   auto [max_minibatch_size, mem_per_row] =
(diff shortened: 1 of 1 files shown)


cpp/src/cluster/detail/kmeans_balanced.cuh
```diff
@@ -1042,9 +1042,10 @@ void build_hierarchical(const raft::resources& handle,
   IdxT n_mesoclusters = std::min(n_clusters, static_cast<IdxT>(std::sqrt(n_clusters) + 0.5));
   RAFT_LOG_DEBUG("build_hierarchical: n_mesoclusters: %u", n_mesoclusters);
 
-  // TODO: Remove the explicit managed memory- we shouldn't be creating this on the user's behalf.
```
cjnolet, line 1045: Please don't remove this line. We do still need to fix this, but it's going to be a longer-term thing.

csadorf: Fix it by replacing the array allocations with `make_managed_vector` as mentioned [here](https://github.com/rapidsai/raft/pull/2570#discussion_r1943334811)? Or do you have something else in mind?

cjnolet: I mean just simply to leave the TODO in place. THough your change improves things, this is still a problem, even with your change. We shouldn't be forcing managed memory on users EVER. They should be full control over it. This was one spot that wasn't trivial to fix, and I'd like to still make sure it's marked as future tech debt and not remove the TODO.

csadorf: Done in https://github.com/rapidsai/cuvs/pull/659/commits/0e881e48d48e2e2068ba005ba6d9f10980cdfb45 .

tfeher (commented):
Thanks @csadorf for the PR!

I do not think this is the right fix. Just a few lines below your change, we calculate a minibatch size. We process the data in batches, therefore this should not go OOM.

I think the `workspace_allocator `is conceptually right allocator to use. But we should make sure that `calc_minibatch_size` considers the available workspace memory and calculates memory usage correctly.

As a workaround this PR is fine. I do not think there would be a perf impact. But if we decide to go ahead with this, then please create an issue to revise this in next release.

Do you have a reproducer for the problem?

csadorf:
> Thanks @csadorf for the PR!
> 
> I do not think this is the right fix. Just a few lines below your change, we calculate a minibatch size. We process the data in batches, therefore this should not go OOM.
> 
> I think the `workspace_allocator `is conceptually right allocator to use. But we should make sure that `calc_minibatch_size` considers the available workspace memory and calculates memory usage correctly.
> 
> As a workaround this PR is fine. I do not think there would be a perf impact. But if we decide to go ahead with this, then please create an issue to revise this in next release.
> 

@tfeher Thank you for your comments. The issue is that we do not only use this workspace for allocations related to the `minibatch_size`, but also within the [`build_fine_clusters()` function](https://github.com/rapidsai/cuvs/blob/branch-25.04/cpp/src/cluster/detail/kmeans_balanced.cuh#L912) (passed as `device_memory`) which allocates `mc_trainset_buf [mesoscluster_size_max x dim]` which is on the order of _dataset size / n_clusters_.

For very large datasets, the user must either increase the number of (mesoscale) clusters or use managed memory. However, in the current implementation, even using managed memory will not suffice since the current workspace would run into the resource_allocator limitation at _total vram / 4_.

Increasing the number of clusters commensurate with the dataset size is generally advisable, but I believe that even if we do not generally remove the resource limiter on the workspace, we should at least remove it specifically for the `mc_trainset_buf` allocation since there is no expectation that it should be on the order of `minibatch_size`.

> Do you have a reproducer for the problem?

The reproducer is in https://github.com/rapidsai/cuml/issues/6204 .

PR #662, opened 2025-02-05
Add `shellcheck` to pre-commit and fix warnings

gforsyth:
`shellcheck` is a fast, static analysis tool for shell scripts. It's good at
flagging up unused variables, unintentional glob expansions, and other potential
execution and security headaches that arise from the wonders of `bash` (and
other shlangs).

This PR adds a `pre-commit` hook to run `shellcheck` on all of the `sh-lang`
files in the `ci/` directory, and the changes requested by `shellcheck` to make
the existing files pass the check.

xref: rapidsai/build-planning#135

### DIFF
--- a/ci/build_python.sh
+++ b/ci/build_python.sh
@@ -15,9 +15,6 @@ rapids-print-env
 
 rapids-logger "Begin py build"
 
-package_name="cuvs"
-package_dir="python"
-
 CPP_CHANNEL=$(rapids-download-conda-from-s3 cpp)
 
 version=$(rapids-generate-version)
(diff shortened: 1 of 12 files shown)


ci/build_python.sh
```diff
@@ -15,9 +15,6 @@ rapids-print-env
 
 rapids-logger "Begin py build"
 
-package_name="cuvs"
-package_dir="python"
```
jameslamb, line 19: 🎉

ci/validate_wheel.sh
```diff
@@ -6,18 +6,17 @@ set -euo pipefail
 package_dir=$1
 wheel_dir_relative_path=$2
 
-RAPIDS_CUDA_MAJOR="${RAPIDS_CUDA_VERSION%%.*}"
```
jameslamb, line 9: Definitely looks like something I left behind in #594 , thanks for fixing it.

jameslamb (approved):

PR #667, opened 2025-02-06
Fix ann-bench deadlocking on HNSW destruction due to task locks

achirkin:
The ann-bench tool has been observed to deadlock on thread.join() due to unnecessary mutex lock.
The problem is that the destructor doesn't release the thread mutex and thus doesn't allow the thread to escape the condition_variable.wait() function.
The fix is to just remove the lock in the destructor (which doesn't modify the state of the task anyway).


tfeher (approved):
Thanks Artem, the fix works for me! LGTM.

Without this I get deadlock on the platforms I have tested.

PR #671, opened 2025-02-07
Fix ann-bench dataset blob integer overflow leading to incorrect data copy beyond 4GB

achirkin:
ann-bench keeps data dimensions as `uint32_t`. We use `std::fread` to copy the data from a file to the host memory and pass `n_rows * n_cols` there, which gets casted to size_t only after the multiplication. This leads to integer overflow for the datasets larger than 4B elements and a partial data copy.

This PR fixes the bug by casting the dimensions before the multiplication.
The bug only affects the benchmark cases where the data is requested in the host memory not backed by a file.


tfeher (approved):
Thanks Artem for the fix, it works!

PR #672, opened 2025-02-07
CuVS-Java: Update readme files

narangvivek10:
PR to update:

- Readme files
- Change logging in examples


java/README.md
```diff
@@ -1,14 +1,42 @@
-Prerequisites
--------------
+# CuVS Java API
 
-* JDK 22
-* Maven 3.9.6 or later
+CuVS Java API attempts to provide a Java based simple, efficient, and a robust vector search API.
```
cjnolet, line 3:
```suggestion
CuVS Java API provides a Java based simple, efficient, and a robust vector search API.
```

We don't attempt here, we just do.

java/README.md
```diff
@@ -1,14 +1,42 @@
...
 
-To build this API, please do `./build.sh java` in the top level directory. Since this API is dependent on `libcuvs` it must be noted that `libcuvs` gets built automatically before building this API.
+> [!CAUTION]
+> This is an experimental release and updates to this API should be expected in the coming days.
```
cjnolet, line 6: I've learned that timeboxing like this is not a good idea. We *should* be clear that APIs are experimental, but they'll be this way for the 25.02 release. Great if we can remove them in the 25.04 release, but this stable release will have experimental APIs. 

```suggestion
> cuVS 25.02 contains an experimental version and updates to this API are expected in the coming release.
```

java/README.md
```diff
@@ -1,14 +1,42 @@
...
 
-Alternatively, please build libcuvs (`./build.sh libcuvs` from top level directory) before building the Java API with `./build.sh` from this directory.
 
-Building
---------
+## Prerequisites
 
-`./build.sh` will generate the `libcuvs_java.so` file in the `internal/` directory, and then build the final jar file for the cuVS Java API in the `cuvs-java/` directory.
+- gcc 11.4
```
cjnolet, line 11: Is this the result of cuVS being dependent upong gcc 11.4? Is gcc 11.4 really a requirement or can users still use GCC 9.x (which is the requirement for buildign cuVS). Let's link to the cuVS build from source page here if these dependencies are only for building cuVS itself. Having these types of constraints duplicated across the codebas becomes hard to manage and maintain.

java/README.md
```diff
@@ -1,14 +1,42 @@
...
+- [nvcc 12.4 or above](https://developer.nvidia.com/cuda-downloads)
```
cjnolet, line 12: cuVS supports 11.x and box. Is there a reason the java API only supports 12.4? If not, let's just point to the cuVS `build from source` docs here and not duplicate these things.

java/README.md
```diff
@@ -1,14 +1,42 @@
...
+- [cmake 3.28 or above](https://cmake.org/download/)
```
cjnolet, line 13: Same here- I'm going to assume the only differences are the maven version and JDK. Why ubuntu 22.04? it doesn't look good to only support a very specific version of linux. I propose removing it, since users can get the toolchain they need from building the proper conda environment.

java/README.md
```diff
@@ -1,14 +1,42 @@
...
+- [maven 3.9.6 or above](https://maven.apache.org/download.cgi)
+- [JDK 22](https://openjdk.org/projects/jdk/22/)
+- [Ubuntu 22.04](https://releases.ubuntu.com/jammy/)
+
+
+## Building
+
+libcuvs libraries are needed for this API. If libcuvs libraries are already not built please do `./build.sh libcuvs java` in the top level directory to build this API.
```
cjnolet, line 21: I'm really not a fan of having this embedded in the java directory. I'd much prefer that this information is provided directly in the `build cuVS from source` docs and this README just links to that. It just simply makes things hard to find, maintain (and update) when this info is scattered around the codebase.

Let's keep it here for the moment while this API is in experimental state, but plan to move into the docs in 25.04 release cycle. 

```suggestion
The libcuvs C and C++ libraries are needed for this API. If libcuvs libraries have not been built and installed, use `./build.sh libcuvs java` in the top level directory to build this API.
```

java/README.md
```diff
@@ -1,14 +1,42 @@
...
+
+Alternatively, if libcuvs libraries are already built and you just want to build this API, please
+do `./build.sh java` in the top level directory or just do `./build.sh` in this directory.
+
+:warning: If you notice the tests failing please replace `mvn verify` with `mvn clean package` in the `build.sh` script found in this directory and try again. This should build the API (and skip the tests).
```
cjnolet, line 26: This seems like a `notice` to me, not a warning.

java/README.md
```diff
@@ -1,14 +1,42 @@
...
+
+
+## Examples
+
+For easy understanding we have provided starter examples for CAGRA, HNSW, and Bruteforce and these can be found in the `examples` directory.
```
cjnolet, line 31:
```suggestion
A few starter examples of CAGRA, HNSW, and Bruteforce index are provided in the `examples` directory.
```

cjnolet: Please make note that we need to move these to the `examples` directory at the top-level of this repo before we remove the experimental designator. All of the other code examples in cuVS are in that directory, so we expect users can look in a single place for them.

java/README.md
```diff
@@ -1,14 +1,42 @@
...
+
+
+## Javadocs
+
+To generate javadocs, in this directory (after building the API) please do:
```
cjnolet, line 36: Before we remove the experimental designator, we need to get these building in CI and deployed.

java/examples/README.md
```diff
@@ -1,8 +1,55 @@
-Building and Running
---------------------
+# CuVS Java API Examples
 
-Make sure to have JDK 22 and Maven 3.9.6+.
+This maven project contains simple examples for CAGRA, HNSW, and Bruteforce algorithms.
 
-    mvn clean compile assembly:single
+## Prerequisites
+- All the prerequisites in the CuVS Java API readme
```
cjnolet, line 6: This would be where you'd point to the cuVS build and install guide, which lists the dependencies for libcuvs.

java/examples/README.md
```diff
@@ -1,8 +1,55 @@
...
 
-    java --enable-native-access=ALL-UNNAMED -jar ./target/cagra-sample-1.0-SNAPSHOT-jar-with-dependencies.jar
+## Run Examples
+
+### CAGRA Example
+Doing the following in the current directory:
+```
+java --enable-native-access=ALL-UNNAMED -cp target/cuvs-java-examples-25.02.0.jar:$HOME/.m2/repository/com/nvidia/cuvs/cuvs-java/25.02.0/cuvs-java-25.02.0.jar com.nvidia.cuvs.examples.CagraExample
+```
+Should output the following (with different timestamps):
```
cjnolet, line 17: This seems unecessary- I would just provide the command-line to run the example. Let the unit tests verify the results.

narangvivek10:
Creating a fresh PR instead. I will include all the suggestions in the new one. Closing this now. Thanks!

PR #689, opened 2025-02-13
Optimize euclidean distance in host refine phase

anstellaire:
### Issue

Original code (below) generated serial assembly and used strictly-ordered `fadda` instruction on ARM with gcc & clang. That resulted in suboptimal performance.
```c++
for (size_t k = 0; k < dim; k++) {
  distance += DC::template eval<DistanceT>(query[k], row[k]);
}
```

### Proposed solution

This PR provides euclidean distance optimized with partial vector sum (below), that helps vectorization but loses strcictly-ordered compliance.
```c++
template <typename DC, typename DistanceT, typename DataT>
DistanceT euclidean_distance_squared_generic(DataT const* a, DataT const* b, size_t n) {
  size_t constexpr max_vreg_len = 512 / (8 * sizeof(DistanceT));

  // max_vreg_len is a power of two
  size_t n_rounded = n & (0xFFFFFFFF ^ (max_vreg_len - 1));
  DistanceT distance[max_vreg_len] = {0};

  for (size_t i = 0; i < n_rounded; i += max_vreg_len) {
    for (size_t j = 0; j < max_vreg_len; ++j) {
      distance[j] += DC::template eval<DistanceT>(a[i + j], b[i + j]);
    }
  }

  for (size_t i = n_rounded; i < n; ++i) {
    distance[i] += DC::template eval<DistanceT>(a[i], b[i]);
  }

  for (size_t i = 1; i < max_vreg_len; ++i) {
    distance[0] += distance[i];
  }

  return distance[0];
}
```
In addition, it has an implementation with NEON intrinsics which provides further speedup on certain test cases (can be removed if arch-specific code is undesired).

### Results
image


tfeher (commented):
Thanks @anstellaire for the PR! It is a clean implementation and it looks good overall. 

You have changed the distance computation for the large batch size case, but did not change for the small batch case (which is handled in a [separate branch here](https://github.com/rapidsai/cuvs/blob/2c3df758ee62a941f0e859ca29d294db66b12e96/cpp/src/neighbors/refine/refine_host.hpp#L54)). Is this because your benchmarks have shown no improvement for the small batch case? Or is it the other way around, and we do not see improvement for small batch cases, because the new distance computation routines are not used there?

(In any case, we can limit the scope of this PR to the large batch case, but please clarify the question above.)

tfeher (approved):
Thank Anna for the PR. The changes look good to me! The remaining question about small batch refinement can be discussed separately.

Issue #697, opened 2025-02-14
cuvs-java: Fix, document, and test BruteForceQuery.Builder::withPrefilter


ChrisHegarty:
This issue has been filed to evaluate the pre-filter for `BruteForceQuery`.

I have been unsuccessful in getting it to work, and am unsure how it is supposed to work. For example,

```
var queryBuilder = new BruteForceQuery.Builder()
   .withPrefilter(new long[] { 0b1001L });
```

Should this filter out vectors at ordinal 1 and 2? In my testing it does not.

chatman:
Fixed in #706

cjnolet:
This should be fixed by #870

PR #706, opened 2025-02-18
Java: Fixing the broken pre-filtering for Brute Force

chatman:
Changed the signature from `long[] prefilter` to `BitSet[] prefilters`.
Added a very simple test.

### DIFF
--- a/java/internal/src/cuvs_java.c
+++ b/java/internal/src/cuvs_java.c
@@ -266,26 +266,20 @@ cuvsBruteForceIndex_t build_brute_force_index(float *dataset, long rows, long di
  * @param[in] n_rows number of rows in the dataset
  */
 void search_brute_force_index(cuvsBruteForceIndex_t index, float *queries, int topk, long n_queries, int dimensions,
-    cuvsResources_t cuvs_resources, int64_t *neighbors_h, float *distances_h, int *return_value, long *prefilter_data,
-    long prefilter_data_length, long n_rows) {
+    cuvsResources_t cuvs_resources, int64_t *neighbors_h, float *distances_h, int *return_value, unsigned char *prefilter_data,
+    long prefilter_data_length) {
 
   cudaStream_t stream;
   cuvsStreamGet(cuvs_resources, &stream);
 
   int64_t *neighbors;
   float *distances, *queries_d;
-  long *prefilter_data_d;
-
-  long prefilter_data_32_size = sizeof(uint32_t) * prefilter_data_length * 2;
-  uint32_t *prefilter_data_32 = (uint32_t *)malloc(prefilter_data_32_size);
 
   cuvsRMMAlloc(cuvs_resources, (void**) &queries_d, sizeof(float) * n_queries * dimensions);
   cuvsRMMAlloc(cuvs_resources, (void**) &neighbors, sizeof(int64_t) * n_queries * topk);
   cuvsRMMAlloc(cuvs_resources, (void**) &distances, sizeof(float) * n_queries * topk);
-  cuvsRMMAlloc(cuvs_resources, (void**) &prefilter_data_d, prefilter_data_32_size);
 
   cudaMemcpy(queries_d, queries, sizeof(float) * n_queries * dimensions, cudaMemcpyDefault);
-  cudaMemcpy(prefilter_data_d, prefilter_data_32, prefilter_data_32_size, cudaMemcpyDefault);
 
   int64_t queries_shape[2] = {n_queries, dimensions};
   DLManagedTensor queries_tensor = prepare_tensor(queries_d, queries_shape, kDLFloat, 32, 2, kDLCUDA);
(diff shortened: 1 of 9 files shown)


chatman:
FYI @ChrisHegarty @narangvivek10.

ChrisHegarty:
relates #697

java/internal/src/cuvs_java.c
```diff
@@ -296,20 +290,22 @@ void search_brute_force_index(cuvsBruteForceIndex_t index, float *queries, int t
...
-
   cuvsFilter prefilter;
   if (prefilter_data == NULL) {
     prefilter.type = NO_FILTER;
     prefilter.addr = (uintptr_t)NULL;
   } else {
-    int64_t prefilter_shape[1] = {(n_queries * n_rows + 31) / 32};
-    DLManagedTensor prefilter_tensor = prepare_tensor(prefilter_data_d, prefilter_shape, kDLUInt, 32, 1, kDLCUDA);
+    // Parse the filters data
+    uint32_t *prefilters = (uint32_t*) malloc(prefilter_data_length);
```
ChrisHegarty, line 299: I don't understand why it is necessary to malloc and copy the pre filter data. Is it not possible to access it directly from the passed pointer?

chatman: I had to change the byte ordering to make it work. Just type casting it to `uint32_t *` didn't work. I'll explore if doing some in-place transformation without allocating a new array is feasible.

ChrisHegarty: It should be possible to write the expected byte order at the java level,

cjnolet (approved):

rhdong:
Hi @chatman , the style checking seems to fail, may you have time to take a look? thanks!

chatman:
FYI, for larger dataset sizes, this fix is incorrect. Working on the proper fix.

java/internal/src/cuvs_java.c
lowener: On the C++ side I added last month an option to specify the byte order used at the time of the creation of the prefilter data: https://github.com/rapidsai/raft/blob/branch-25.04/cpp/include/raft/core/bitmap.hpp#L64.
I think in your case it was created with `uint8_t`? So specifying `original_nbits = sizeof(uint8_t)` should prevent you from that extra allocation/copy. It is not present yet in the C API though.

chatman: @ChrisHegarty fixed, it was unnecessary to change the order. I was doing things wrong. Fixed now.

chatman:
I think this is ready for merge. Please review @rhdong @ChrisHegarty @narangvivek10.

chatman:
@rhdong Upon testing, I'm finding that on some systems, this is not working. On such systems, search call is returning non success return value, with invalid memory access. I copied the prefilter data into a device array and passed that array, and it works again. Wondering why it is working on one system with just the host array, but not on others. I'll continue the investigation and fix this by tomorrow (hopefully, before the freeze for the next release). FYI @cjnolet.

rhdong:
> @rhdong Upon testing, I'm finding that on some systems, this is not working. On such systems, search call is returning non success return value, with invalid memory access. I copied the prefilter data into a device array and passed that array, and it works again. Wondering why it is working on one system with just the host array, but not on others. I'll continue the investigation and fix this by tomorrow (hopefully, before the freeze for the next release). FYI @cjnolet.

Hi @chatman , the API always uses the filter memory pointer as a device accessible one, a little weird that `malloc` can work. :thinking: May I have your reproduction code? or code snippet?

java/internal/src/cuvs_java.c
```diff
@@ -296,20 +290,16 @@ void search_brute_force_index(cuvsBruteForceIndex_t index, float *queries, int t
...
     prefilter.type = NO_FILTER;
     prefilter.addr = (uintptr_t)NULL;
   } else {
-    int64_t prefilter_shape[1] = {(n_queries * n_rows + 31) / 32};
-    DLManagedTensor prefilter_tensor = prepare_tensor(prefilter_data_d, prefilter_shape, kDLUInt, 32, 1, kDLCUDA);
+    // Parse the filters data
+    int num_integers = (prefilter_data_length+63)/64 * 2;
+    int extraPaddingByteExists = prefilter_data_length % 64 > 32? 0: 1;
+    int64_t prefilter_shape[1] = {(prefilter_data_length + 31) / 32};
+    DLManagedTensor prefilter_tensor = prepare_tensor(prefilter_data, prefilter_shape, kDLUInt, 32, 1, kDLCUDA);
```
rhdong, line 302: I suppose the `DLManagedTensor prefilter_tensor` should be declared outside of `if` branch

jakirkham:
Closing as fixed by PR: https://github.com/rapidsai/cuvs/pull/695

chatman:
This closes #697.

PR #707, opened 2025-02-18
Replace `cub::TransformInputIterator` with `thrust::transform_iterator`

miscco:
The former is deprecated and will be removed in a future release


miscco:
@cjnolet it seems to be passing with the latest fix. Can you by change rerun the failed nightly run once nightly is green again?

cjnolet (approved):

PR #713, opened 2025-02-20
[Feat] Add support of logical merge in Cagra

rhdong:
- https://github.com/rapidsai/cuvs/issues/712


cpp/include/cuvs/neighbors/cagra.hpp
```diff
@@ -1207,7 +1288,105 @@ void search(raft::resources const& res,
...
+ * @param[in] index composite cagra index
+ * @param[in] queries a device matrix view to a row-major matrix [n_queries, index->dim()]
+ * @param[out] neighbors a device matrix view to the indices of the neighbors in the source dataset
+ * [n_queries, k]
+ * @param[out] distances a device matrix view to the distances to the selected neighbors [n_queries,
+ * k]
+ * @param[in] sample_filter an optional device filter function object that greenlights samples
+ * for a given query. (none_sample_filter for no filtering)
+ */
+void search(raft::resources const& res,
```
rhdong, line 1307: Hi @cjnolet @achirkin , I had to add a new search for `composite_index`, the root cause is the [`cuve::neighbors::index` ](https://github.com/rapidsai/cuvs/blob/branch-25.04/cpp/include/cuvs/neighbors/common.hpp#L87)is not designed to be an abstract class, which makes many pathways impossible,  like inheriting from `index` can avoid to declare a `composible_index` and a new search API. I was trying to refactor it, but I found it was used widely in different algo codes; maybe we can make it in the long term.

achirkin:
I may have missed some prior discussion, but the cagra-specific logical merge seems a little bit artificial for me.
If I understand this correctly, you essentially implement a multi-index search. Then, why don't we go one step further and make it independent of a particular index implementation? Moreover, it looks like much of this functionality is implemented already in [multi-gpu index](https://github.com/rapidsai/cuvs/blob/6b5b472ea479ecb11675283ddd43d28c4f821302/cpp/include/cuvs/neighbors/mg.hpp). Maybe we can generalize that one a bit to decouple "multi-index" from "multi-gpu" aspect? Optionally, one can also combine different index types and erase the upstream index type like we do in [dynamic batching index](https://github.com/rapidsai/cuvs/blob/6b5b472ea479ecb11675283ddd43d28c4f821302/cpp/include/cuvs/neighbors/dynamic_batching.hpp).

cpp/include/cuvs/neighbors/cagra.hpp
cjnolet: I think it's okay that we don't require a strict class hierarchy initially, so long as we have proper overloads so that from a user's perspective, they can call `auto index = cuvs::neighbors::cagra::merge(...)` and get back a proper object from which they can later call `cuvs::neighbors::cagra::search(..., index,..)`

cjnolet:
> I may have missed some prior discussion, but the cagra-specific logical merge seems a little bit artificial for me.

@achirkin you and I haven't discussed this yet but this feature is critical for certain database architectures like Lucene, which are variants of the log-structured-merge pattern but build indexes immediately after segments (flat files containing vectors) are created, rather than merging the segments together before building the indexes. This causes Lucene to have to first merge many tiny CAGRA indexes together, but eventually it'll end up merging very large indexes together. It is this latter case thart we care about doing a logical merge. It's more efficient to create a single cagra index with many (potentially hundreds) of tiny cagra indexes but when they reach a certain size, it's actually more efficient to logically merge them; meaning we essentially wrap them as if they were shards and broadcast the query to all the shards during search.

I agree this is similar in theory to the multi-gpu sharding and perhaps there is some code to be reused there. The next step for the `merge()` API is to be able to implement a `SMART` merge strategy, whereby we can more efficiently merge two cagra graphs together without having to rebuild from scratch, and that's ultimately the strategy we have discussed offline in more detail. The problem is that Lucene can make use of this today, so it gives us time to work towards the SMART option.

achirkin:
I totally agree and don't doubt the usefulness of logical merge.
I've just pointed out that I think we can accomplish a more general solution (supporting any index type) with actually less code by re-using/adjusting what we already have in multi-index/dynamic batching.

rhdong:
> I totally agree and don't doubt the usefulness of logical merge. I've just pointed out that I think we can accomplish a more general solution (supporting any index type) with actually less code by re-using/adjusting what we already have in multi-index/dynamic batching.

rhdong:
> I may have missed some prior discussion, but the cagra-specific logical merge seems a little bit artificial for me. If I understand this correctly, you essentially implement a multi-index search. Then, why don't we go one step further and make it independent of a particular index implementation? Moreover, it looks like much of this functionality is implemented already in [multi-gpu index](https://github.com/rapidsai/cuvs/blob/6b5b472ea479ecb11675283ddd43d28c4f821302/cpp/include/cuvs/neighbors/mg.hpp). Maybe we can generalize that one a bit to decouple "multi-index" from "multi-gpu" aspect? Optionally, one can also combine different index types and erase the upstream index type like we do in [dynamic batching index](https://github.com/rapidsai/cuvs/blob/6b5b472ea479ecb11675283ddd43d28c4f821302/cpp/include/cuvs/neighbors/dynamic_batching.hpp).

Hi @achirkin , sorry for the late response to your comments here. I studied the code of SNMG, which is a super useful feature. When I tried to reuse some code in it, I found the caller has to work with locker, nccl, which is unnecessary in the logical merge of CAGRA. I'd like to keep the simple implementation if you agree. Many thanks! cc: @cjnolet

rhdong:
Hi @cjnolet @achirkin, could you please merge it if there are no more comments? Many thanks!

cjnolet:
> When I tried to reuse some code in it, I found the caller has to work with locker, nccl, which is unnecessary in the logical merge of CAGRA. I'd like to keep the simple implementation if you agree. Many thanks! cc: @cjnolet

@rhdong I don't think the ask here is that you use the multi-gpu apis, but rather that we have a more consistent and general means of being able to support wrapper types that can work with any index type and avoid the need to, for example, implement a separate `composite_index` type for each index type within cuVS. Dynamic batching and the multi-gpu apis are two examples of where we have a single index that can be used by any other index. This is ultimately where we should strive to be.

rhdong:
> > When I tried to reuse some code in it, I found the caller has to work with locker, nccl, which is unnecessary in the logical merge of CAGRA. I'd like to keep the simple implementation if you agree. Many thanks! cc: @cjnolet
> 
> @rhdong I don't think the ask here is that you use the multi-gpu apis, but rather that we have a more consistent and general means of being able to support wrapper types that can work with any index type and avoid the need to, for example, implement a separate `composite_index` type for each index type within cuVS. Dynamic batching and the multi-gpu apis are two examples of where we have a single index that can be used by any other index. This is ultimately where we should strive to be.

Hi @cjnolet @achirkin , just want to clarify before moving forward, what's you suggesting is lifting up the [composite_index](https://github.com/rapidsai/cuvs/pull/713/files#diff-205397266487b3f6e703c950361c05df17c4c5d3540def0af301e83ea34775d2R642) to the upper level, like by moving it to `common.hpp`, and using the `cuvs::neighbour::index` as array item type, am I right? Many thanks!

cjnolet:
> Hi @cjnolet @achirkin , just want to clarify before moving forward, what you're suggesting is lifting up the [composite_index](https://github.com/rapidsai/cuvs/pull/713/files#diff-205397266487b3f6e703c950361c05df17c4c5d3540def0af301e83ea34775d2R642) to the upper level, like by moving it to common.hpp, and using the cuvs::neighbour::index as array item type, am I right? Many thanks!

Yes, though I think we could address this in a follow-up PR and merge this change in the meantime. We really need CAGRA merge capability initially so that we can unblock our Lucene friends. @rhdong can you create a new github issue to follow-up with the first-class formal generalized composite_index? 

The other big reason why centralizing this composite index is important is because we want it to work out of the box with the other APIs that work with general indexes (such as the snmg apis).

rhdong:
> > Hi @cjnolet @achirkin , just want to clarify before moving forward, what you're suggesting is lifting up the [composite_index](https://github.com/rapidsai/cuvs/pull/713/files#diff-205397266487b3f6e703c950361c05df17c4c5d3540def0af301e83ea34775d2R642) to the upper level, like by moving it to common.hpp, and using the cuvs::neighbour::index as array item type, am I right? Many thanks!
> 
> Yes, though I think we could address this in a follow-up PR and merge this change in the meantime. We really need CAGRA merge capability initially so that we can unblock our Lucene friends. @rhdong can you create a new github issue to follow-up with the first-class formal generalized composite_index?
> 
> The other big reason why centralizing this composite index is important is because we want it to work out of the box with the other APIs that work with general indexes (such as the snmg apis).

@cjnolet @achirkin I created an issue for tracking our discussion with this in one place: https://github.com/rapidsai/cuvs/issues/882

cpp/include/cuvs/neighbors/cagra.hpp
```diff
@@ -557,6 +556,82 @@ struct index : cuvs::neighbors::index {
...
+
+  constexpr inline auto graph_degree() const noexcept -> uint32_t
+  {
+    return sub_indices.front()->graph_degree();
+  }
+
+  constexpr inline auto num_indices() const noexcept -> uint32_t { return sub_indices.size(); }
+
+ public:
+  std::vector<cuvs::neighbors::cagra::index<T, IdxT>*> sub_indices;
```
cjnolet, line 627: I know we are planning a potential overhaul of composite index for the next iteration of merging, but in the meantime, it seems like a fairly straightforward change to introduce a template type to capture the `neighbors::xxx::index` type for the `composite_index`. Can we do that, and move this to `common.hpp`? That would be a good start anyways. Then the `cagra::composite_index` can extend `neighbors::composite_index` and provide the additional accessor methods for the cagra-specific things.

cpp/include/cuvs/neighbors/cagra.hpp
```diff
@@ -1117,7 +1192,6 @@ void extend(
  * @param[in] sample_filter an optional device filter function object that greenlights samples
  * for a given query. (none_sample_filter for no filtering)
  */
-
 void search(raft::resources const& res,
```
cjnolet, line 1195: Not necessarily needed for this PR, but it would be nice if we could work towards having a more agnostic API for the `composite::index` that could accept any `composite::index` and be able to search it appropriately (aka dispatch to the appropriate index type). Especially for the next step, when we move towards merging the`tiered::index` together and then searching it.

cpp/include/cuvs/neighbors/cagra.hpp
cjnolet: I know we are pushing this to a follow-on, but just want to make sure we're clear about the expectations for that design change.

cpp/include/cuvs/neighbors/cagra.hpp
```diff
@@ -2145,6 +2419,23 @@ auto merge(raft::resources const& res,
            const cuvs::neighbors::cagra::merge_params& params,
            std::vector<cuvs::neighbors::cagra::index<uint8_t, uint32_t>*>& indices)
   -> cuvs::neighbors::cagra::index<uint8_t, uint32_t>;
+
+auto make_composite_index(const cagra::merge_params& params,
```
cjnolet, line 2423: Should a user ever need to call this? The `merge()` is really the `make_composite_index` isn't it? And the guts of the merge is intentionally kept as implementaiton details. To be clear- I think the factory functions for constructing complex objects are important, but I thought `merge()` was the factory function to accomplish that.

cpp/src/neighbors/detail/cagra/cagra_search.cuh
```diff
@@ -41,8 +41,187 @@
...
 #include <rmm/cuda_stream_view.hpp>
 
+#include <cub/cub.cuh>
+#include <thrust/copy.h>
+#include <thrust/sequence.h>
+
 namespace cuvs::neighbors::cagra::detail {
 
+template <typename T, typename IdxT>
+void merge_top_k_gpu(raft::resources const& res,
```
cjnolet, line 51: We have routines to do this top-k merge already. We should use the same centralized routines so that we aren't reinventing this. I know the multi-gpu (sharded) search implementation should be using the same centralized routines also. cc @benfred @viclafargue

cpp/src/neighbors/detail/cagra/cagra_search.cuh
```diff
@@ -41,8 +41,187 @@
...
+
+  for (size_t i = 0; i < idx.num_indices(); i++) {
+    const index<T, IdxT>* sub_index = idx.sub_indices[i];
+    cagra::search(res, params, *sub_index, queries, neighbors, distances);
+    if (offset != 0) {
+      raft::linalg::addScalar(
+        neighbors.data_handle(), neighbors.data_handle(), offset, neighbors.size(), stream);
+    }
+
+    RAFT_CUDA_TRY(cudaMemcpy2DAsync(neighbors_buffer.data() + i * K,
```
cjnolet, line 194: Is there a reason we have to use 2d memcpy here instead of treating this as a single array? In general, we should be using `raft::copy` instead of invoking the cuda runtime API directly (I understand we're doing this in other place sin the codebase, but we shouldn't be). I'm wondering if we need a `raft::copy_matrix` utility in RAFT's `cudart_utils`

cpp/include/cuvs/neighbors/cagra.hpp
```diff
@@ -557,6 +556,82 @@ struct index : cuvs::neighbors::index {
...
+ * it holds non-owning pointers to them.
+ *
+ * All sub-indices within the composite index **must share the same distance metric
+ * and dimensionality**.
+ *
+ * @tparam T    Data element type.
+ * @tparam IdxT Index type representing dataset.extent(0), used for vector indices.
+ */
+
+template <typename T, typename IdxT>
```
cjnolet, line 583: Have you done any benchmarking to see how search times of logically merged indexes compare? It would help to know that the additional merges aren't introducing a ton of overhead that might not be acceptable in practice. Also good to know that we're getting the best possible perf out of it and there aren't any unexpected bottlenecks.

cpp/src/neighbors/detail/cagra/cagra_search.cuh
rhdong: Yeah, makes sense in general — in this case we're copying per-query top-k results with `stride` of multi-K between rows, so `cudaMemcpy2DAsync` fits pretty well.
Agree we could use a raft::copy_matrix or something similar long term to wrap this — would definitely help with consistency.

cjnolet: > Yeah, makes sense in general — in this case we're copying per-query top-k results with stride of multi-K between rows, so cudaMemcpy2DAsync fits pretty well.

But why the 2d instead of just `raft::copy` for size=n_rows * k? That's the part i don't understand. Why doesn't it have to be a 2d? If you just use `raft::copy` then we don't need the raft::copy_matrix and we can centralize all the copy calls today.

rhdong: > > Yeah, makes sense in general — in this case we're copying per-query top-k results with stride of multi-K between rows, so cudaMemcpy2DAsync fits pretty well.
> 
> But why the 2d instead of just `raft::copy` for size=n_rows * k? That's the part i don't understand. Why doesn't it have to be a 2d? If you just use `raft::copy` then we don't need the raft::copy_matrix and we can centralize all the copy calls today.

For each query, we have n_indices * k items to copy, and each time a copy processes k items that are searched from the i-th index. There's a stride of `n_indices * k` between each query, so we need a 2D copy here, and by following this, a ranking will get K from the `n_indices * k` as the last result.

cjnolet: If we need to use 2d copy then please add a quick utility to raft for `raft::copy_matrix` and have it call `cudaMemCpy2D`. I'd like to centralize this call.

rhdong: @cjnolet here is the `raft::copy_matrix` : https://github.com/rapidsai/raft/pull/2674, whould you please take a moment for a quick review? Thanks!

rhdong:
Depends on the raft PR: https://github.com/rapidsai/raft/pull/2674

cpp/src/neighbors/detail/cagra/cagra_search.cuh
rhdong: Can I have the code link for the existing tok-k merging?

cpp/src/neighbors/detail/cagra/cagra_search.cuh
rhdong: Done

cpp/src/neighbors/detail/cagra/cagra_search.cuh
rhdong: Hi @cjnolet @benfred @viclafargue , I suppose the existing tok-k merge API that Corey mentioned here is the `knn_merge_parts`, If that's correct, the `knn_merge_parts` seems to be different from the `merge_top_k_gpu `. Anyway, I will move the `merge_top_k_gpu ` to a centralized location.

rhdong:
```shell

- Build

---------------------------------------------------------------------------------------------------------------------------
Benchmark                                        Time             CPU   Iterations  Split_num    MergeStrategy   UserCounters...
---------------------------------------------------------------------------------------------------------------------------
raft_cagra.dim32/process_time/real_time       7.51 s           212 s             1          1         No merge     GPU=7.51419 graph_degree=32 index_size=1000k intermediate_graph_degree=48 ivf_pq_build_niter=10 ivf_pq_build_nlist=16.384k ivf_pq_build_pq_bits=8 ivf_pq_build_pq_dim=32 ivf_pq_build_ratio=10 ivf_pq_search_nprobe=30 ivf_pq_search_refine_ratio=1 split_num=1 dataset_memory_type="device"#graph_build_algo="NN_DESCENT"#ivf_pq_search_internalDistanceDtype="float"#ivf_pq_search_smemLutDtype="float"#merge_type="LOGICAL"
raft_cagra.dim32/process_time/real_time       7.55 s           211 s             1          4         Physical     GPU=7.54935 graph_degree=32 index_size=1000k intermediate_graph_degree=48 ivf_pq_build_niter=10 ivf_pq_build_nlist=16.384k ivf_pq_build_pq_bits=8 ivf_pq_build_pq_dim=32 ivf_pq_build_ratio=10 ivf_pq_search_nprobe=30 ivf_pq_search_refine_ratio=1 split_num=4 dataset_memory_type="device"#graph_build_algo="NN_DESCENT"#ivf_pq_search_internalDistanceDtype="float"#ivf_pq_search_smemLutDtype="float"#merge_type="PHYSICAL"
raft_cagra.dim32/process_time/real_time       8.18 s           254 s             1          4          Logical     GPU=8.17527 graph_degree=32 index_size=1000k intermediate_graph_degree=48 ivf_pq_build_niter=10 ivf_pq_build_nlist=16.384k ivf_pq_build_pq_bits=8 ivf_pq_build_pq_dim=32 ivf_pq_build_ratio=10 ivf_pq_search_nprobe=30 ivf_pq_search_refine_ratio=1 split_num=4 dataset_memory_type="device"#graph_build_algo="NN_DESCENT"#ivf_pq_search_internalDistanceDtype="float"#ivf_pq_search_smemLutDtype="float"#merge_type="LOGICAL"
                                                                                                               
                                                                                                               
- Search
----------------------------------------------------------------------------------------------------------------------------------
Benchmark                                        Time             CPU   Iterations   Split_num    MergeStrategy    UserCounters...
----------------------------------------------------------------------------------------------------------------------------------
raft_cagra.dim32/process_time/real_time       4.90 ms         4.90 ms        11462           1         No merge    GPU=4.8907m Latency=4.89552m Recall=0.99854 end_to_end=56.1125 items_per_second=204.268k/s itopk=256 k=10 n_queries=1000 total_queries=11.462M algo="single_cta"#dataset_memory_type="device"
raft_cagra.dim32/process_time/real_time       4.90 ms         4.90 ms        11453           4         Physical    GPU=4.89376m Latency=4.89864m Recall=0.99843 end_to_end=56.1041 items_per_second=204.138k/s itopk=256 k=10 n_queries=1000 total_queries=11.453M algo="single_cta"#dataset_memory_type="device"
raft_cagra.dim32/process_time/real_time       18.1 ms         18.1 ms         3109           4          Logical    GPU=0.0180525 Latency=0.0180575 Recall=0.99928 end_to_end=56.1409 items_per_second=55.3785k/s itopk=256 k=10 n_queries=1000 total_queries=3.109M algo="single_cta"#dataset_memory_type="devce"
```

cpp/include/cuvs/neighbors/cagra.hpp
rhdong: Hi @cjnolet, I made an initial benchmarking. With logical merge, the search delay increases roughly linearly with split_num. Although each graph is smaller, overall search time grows due to multiple searches, and there's no noticeable speedup.
```shell

- Build

---------------------------------------------------------------------------------------------------------------------------
Benchmark                                        Time             CPU   Iterations  Split_num    MergeStrategy   UserCounters...
---------------------------------------------------------------------------------------------------------------------------
raft_cagra.dim32/process_time/real_time       7.51 s           212 s             1          1         No merge     GPU=7.51419 graph_degree=32 index_size=1000k intermediate_graph_degree=48 ivf_pq_build_niter=10 ivf_pq_build_nlist=16.384k ivf_pq_build_pq_bits=8 ivf_pq_build_pq_dim=32 ivf_pq_build_ratio=10 ivf_pq_search_nprobe=30 ivf_pq_search_refine_ratio=1 split_num=1 dataset_memory_type="device"#graph_build_algo="NN_DESCENT"#ivf_pq_search_internalDistanceDtype="float"#ivf_pq_search_smemLutDtype="float"#merge_type="LOGICAL"
raft_cagra.dim32/process_time/real_time       7.55 s           211 s             1          4         Physical     GPU=7.54935 graph_degree=32 index_size=1000k intermediate_graph_degree=48 ivf_pq_build_niter=10 ivf_pq_build_nlist=16.384k ivf_pq_build_pq_bits=8 ivf_pq_build_pq_dim=32 ivf_pq_build_ratio=10 ivf_pq_search_nprobe=30 ivf_pq_search_refine_ratio=1 split_num=4 dataset_memory_type="device"#graph_build_algo="NN_DESCENT"#ivf_pq_search_internalDistanceDtype="float"#ivf_pq_search_smemLutDtype="float"#merge_type="PHYSICAL"
raft_cagra.dim32/process_time/real_time       8.18 s           254 s             1          4          Logical     GPU=8.17527 graph_degree=32 index_size=1000k intermediate_graph_degree=48 ivf_pq_build_niter=10 ivf_pq_build_nlist=16.384k ivf_pq_build_pq_bits=8 ivf_pq_build_pq_dim=32 ivf_pq_build_ratio=10 ivf_pq_search_nprobe=30 ivf_pq_search_refine_ratio=1 split_num=4 dataset_memory_type="device"#graph_build_algo="NN_DESCENT"#ivf_pq_search_internalDistanceDtype="float"#ivf_pq_search_smemLutDtype="float"#merge_type="LOGICAL"
                                                                                                               
                                                                                                               
- Search
----------------------------------------------------------------------------------------------------------------------------------
Benchmark                                        Time             CPU   Iterations   Split_num    MergeStrategy    UserCounters...
----------------------------------------------------------------------------------------------------------------------------------
raft_cagra.dim32/process_time/real_time       4.90 ms         4.90 ms        11462           1         No merge    GPU=4.8907m Latency=4.89552m Recall=0.99854 end_to_end=56.1125 items_per_second=204.268k/s itopk=256 k=10 n_queries=1000 total_queries=11.462M algo="single_cta"#dataset_memory_type="device"
raft_cagra.dim32/process_time/real_time       4.90 ms         4.90 ms        11453           4         Physical    GPU=4.89376m Latency=4.89864m Recall=0.99843 end_to_end=56.1041 items_per_second=204.138k/s itopk=256 k=10 n_queries=1000 total_queries=11.453M algo="single_cta"#dataset_memory_type="device"
raft_cagra.dim32/process_time/real_time       18.1 ms         18.1 ms         3109           4          Logical    GPU=0.0180525 Latency=0.0180575 Recall=0.99928 end_to_end=56.1409 items_per_second=55.3785k/s itopk=256 k=10 n_queries=1000 total_queries=3.109M algo="single_cta"#dataset_memory_type="devce"
```

cpp/src/neighbors/detail/cagra/cagra_search.cuh
rhdong: Done

cpp/include/cuvs/neighbors/cagra.hpp
rhdong: Hi @cjnolet @achirkin , I made an index refactoring. Could you take a moment to review? Many thanks!

cpp/include/cuvs/neighbors/cagra.hpp
rhdong: It was resolved by this commit: https://github.com/rapidsai/cuvs/pull/713/commits/eca72839bb85abd1cfb60c3dc5da526273e53a08
The setup: A6000 * 1, dataset: sift-128-euclideanl, `batch-size` is set to `1`.

```shell

- Build(keep no changed)

---------------------------------------------------------------------------------------------------------------------------
Benchmark                                        Time             CPU   Iterations  Split_num    MergeStrategy   UserCounters...
---------------------------------------------------------------------------------------------------------------------------
raft_cagra.dim32/process_time/real_time       7.51 s           212 s             1          1         No merge     GPU=7.51419 graph_degree=32 index_size=1000k intermediate_graph_degree=48 ivf_pq_build_niter=10 ivf_pq_build_nlist=16.384k ivf_pq_build_pq_bits=8 ivf_pq_build_pq_dim=32 ivf_pq_build_ratio=10 ivf_pq_search_nprobe=30 ivf_pq_search_refine_ratio=1 split_num=1 dataset_memory_type="device"#graph_build_algo="NN_DESCENT"#ivf_pq_search_internalDistanceDtype="float"#ivf_pq_search_smemLutDtype="float"#merge_type="LOGICAL"
raft_cagra.dim32/process_time/real_time       7.55 s           211 s             1          4         Physical     GPU=7.54935 graph_degree=32 index_size=1000k intermediate_graph_degree=48 ivf_pq_build_niter=10 ivf_pq_build_nlist=16.384k ivf_pq_build_pq_bits=8 ivf_pq_build_pq_dim=32 ivf_pq_build_ratio=10 ivf_pq_search_nprobe=30 ivf_pq_search_refine_ratio=1 split_num=4 dataset_memory_type="device"#graph_build_algo="NN_DESCENT"#ivf_pq_search_internalDistanceDtype="float"#ivf_pq_search_smemLutDtype="float"#merge_type="PHYSICAL"
raft_cagra.dim32/process_time/real_time       8.18 s           254 s             1          4          Logical     GPU=8.17527 graph_degree=32 index_size=1000k intermediate_graph_degree=48 ivf_pq_build_niter=10 ivf_pq_build_nlist=16.384k ivf_pq_build_pq_bits=8 ivf_pq_build_pq_dim=32 ivf_pq_build_ratio=10 ivf_pq_search_nprobe=30 ivf_pq_search_refine_ratio=1 split_num=4 dataset_memory_type="device"#graph_build_algo="NN_DESCENT"#ivf_pq_search_internalDistanceDtype="float"#ivf_pq_search_smemLutDtype="float"#merge_type="LOGICAL"
                                                                                                               
                                                                                                               
- Search
----------------------------------------------------------------------------------------------------------------------------------
Benchmark                                        Time             CPU   Iterations   Split_num    MergeStrategy    UserCounters...
----------------------------------------------------------------------------------------------------------------------------------
raft_cagra.dim32/process_time/real_time       1.26 ms         1.26 ms        44323 GPU=1.25917m Latency=1.26365m Recall=0.99851 end_to_end=56.009 items_per_second=791.355/s itopk=256 k=10 n_queries=1 total_queries=44.323k algo="single_cta"#dataset_memory_type="device"
raft_cagra.dim32/process_time/real_time       1.26 ms         1.26 ms        44651 GPU=1.25266m Latency=1.25714m Recall=0.99856 end_to_end=56.1326 items_per_second=795.455/s itopk=256 k=10 n_queries=1 total_queries=44.651k algo="single_cta"#dataset_memory_type="device"
raft_cagra.dim32/process_time/real_time       1.68 ms         1.68 ms        33382 GPU=1.67523m Latency=1.67982m Recall=0.99928 end_to_end=56.0757 items_per_second=595.302/s itopk=256 k=10 n_queries=1 total_queries=33.382k algo="single_cta"#dataset_memory_type="device"```

File: cpp/bench/ann/src/cuvs/cuvs_cagra_wrapper.h [hunk 8]
```diff
@@ -377,8 +481,40 @@ void cuvs_cagra<T, IdxT>::search_base(
...
+          wrapped_indices;
+        wrapped_indices.reserve(sub_indices_.size());
+        for (auto& ptr : sub_indices_) {
+          auto index_wrapper =
+            cuvs::neighbors::cagra::make_index_wrapper<T, IdxT, algo_base::index_type>(ptr.get());
+          wrapped_indices.push_back(index_wrapper);
+        }
+
+        raft::resources composite_handle(handle_);
+        size_t n_streams = std::min(wrapped_indices.size(), size_t(8));
```
cjnolet, line 503: out of curiosity, why 8 here? I don't think we need to hold up the release for this, but can you at least create a Github issue to make this configurable (or to find a good reasonable default which can be overridden)?

cjnolet (approved):

PR #716, opened 2025-02-24
Add mean and median threshold mode to binary quantization

enp1s0:

### DIFF
--- a/cpp/include/cuvs/preprocessing/quantize/binary.hpp
+++ b/cpp/include/cuvs/preprocessing/quantize/binary.hpp
@@ -34,13 +34,15 @@ namespace cuvs::preprocessing::quantize::binary {
 /**
  * @brief quantizer algorithms
  */
-enum class threshold_mode { zero, mean, sampling_median };
+enum class set_bit_threshold { zero, mean, sampling_median };
 
 /**
  * @brief quantizer parameters.
  */
 struct params {
-  threshold_mode threshold = threshold_mode::zero;
+  set_bit_threshold threshold = set_bit_threshold::zero;
+
+  float sampling_ratio = 0.1;
 };
 
 /**
--- a/cpp/include/cuvs/preprocessing/quantize/binary.hpp
+++ b/cpp/include/cuvs/preprocessing/quantize/binary.hpp
@@ -32,23 +32,31 @@ namespace cuvs::preprocessing::quantize::binary {
  */
 
 /**
+ * @brief quantizer parameters.
+ */
+struct params {};
+
+/**
  * @brief Applies binary quantization transform to given dataset. If a dataset element is positive,
  * set the corresponding bit to 1.
  *
  * Usage example:
  * @code{.cpp}
  * raft::handle_t handle;
+ * cuvs::preprocessing::quantize::binary::params params;
  * auto quantized_dataset = raft::make_device_matrix<uint8_t, int64_t>(handle, samples,
- * features); cuvs::preprocessing::quantize::binary::transform(handle, dataset,
+ * features); cuvs::preprocessing::quantize::binary::transform(handle, params, dataset,
  * quantized_dataset.view());
  * @endcode
  *
  * @param[in] res raft resource
+ * @param[in] params quantization params
  * @param[in] dataset a row-major matrix view on device
  * @param[out] out a row-major matrix view on device
  *
  */
 void transform(raft::resources const& res,
+               const params params,
                raft::device_matrix_view<const double, int64_t> dataset,
                raft::device_matrix_view<uint8_t, int64_t> out);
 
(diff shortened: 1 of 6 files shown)


cjnolet:
@enp1s0 this sounds super useful! Thanks for opening the PR.  Do you have a sense for how much this improves recall?

cpp/src/preprocessing/quantize/detail/binary.cuh
```diff
@@ -128,6 +178,77 @@ void transform(raft::resources const& res,
...
+    const uint32_t max_dim_chunk =
+      std::min(std::max(1lu, raft::resource::get_workspace_free_bytes(res) / data_size_per_vector),
+               static_cast<std::size_t>(dataset_dim));
+
+    auto sampled_dataset_chunk = raft::make_device_mdarray<T, int64_t>(
+      res, mr, raft::make_extents<int64_t>(max_dim_chunk, num_sampls));
+    for (uint32_t dim_offset = 0; dim_offset < dataset_dim; dim_offset += max_dim_chunk) {
+      const auto dim_chunk = std::min(max_dim_chunk, dataset_dim - dim_offset);
+
+      sample_and_transpose_dataset<T>(dataset.data_handle() + dim_offset,
```
tfeher, line 229: It would be great if we could reuse [raft::matrix::sample_rows()](https://github.com/rapidsai/raft/blob/3ff17a0efc29e96da2db7147549fcbafc4c1f267/cpp/include/raft/matrix/sample_rows.cuh#L39)

tfeher (changes requested):
Thanks @enp1s0 for the PR, this is a useful feature!

After we have quantized the dataset, during search we would need to quantize the queries as well. Assuming they have same distribution as the dataset, I would use the same quantization parameters. Therefore it would be important to actually store the threshold values that were found.

Please consider how we do this for [scalar quantization](https://github.com/rapidsai/cuvs/blob/ff8682798536b5975f47ac86b3f3d2e2c801904b/cpp/include/cuvs/preprocessing/quantize/scalar.hpp#L55), where a `quantizer` object is returned by the `train` method, that can be reused. Can we do something similar for binary quantization?

cpp/include/cuvs/preprocessing/quantize/binary.hpp
```diff
@@ -31,24 +31,41 @@ namespace cuvs::preprocessing::quantize::binary {
...
  * @endcode
  *
  * @param[in] res raft resource
+ * @param[in] params quantization params
  * @param[in] dataset a row-major matrix view on device
  * @param[out] out a row-major matrix view on device
  *
  */
 void transform(raft::resources const& res,
+               const params params,
```
tfeher, line 68: This is a breaking change. Please consider keeping still keeping the previous API as a deprecated overload.

cpp/include/cuvs/preprocessing/quantize/binary.hpp
```diff
@@ -59,17 +76,20 @@ void transform(raft::resources const& res,
  * Usage example:
  * @code{.cpp}
  * raft::handle_t handle;
+ * cuvs::preprocessing::quantize::binary::params params;
  * auto quantized_dataset = raft::make_host_matrix<uint8_t, int64_t>(handle, samples,
- * features); cuvs::preprocessing::quantize::binary::transform(handle, dataset,
+ * features); cuvs::preprocessing::quantize::binary::transform(handle, params, dataset,
```
tfeher, line 81: Please consider the following potential use case:
```c++
cuvs::preprocessing::quantize::binary::transform(handle, params, dataset, quantized_dataset.view());
auto index = cagra::build(handle, build_params, quantized_dataset);

auto quantized_queries = raft::make_device_matrix<uint8_t, int64_t>(handle, queries.extent(0), queries.extent(1));
cuvs::preprocessing::quantize::binary::transform(handle, params, queries, quantized_queries.view());

cagra::search(handle, search_params, index, quantized_queries, neighbors, distances);
```

How do we quantize the `queries` without storing the quantization parameters?

tfeher (changes requested):
Thanks enp1s0 for the updates, Some issues still need to be addressed.

cpp/src/preprocessing/quantize/detail/binary.cuh
enp1s0: It is possible to use raft::matrix::sample_rows() here, but would it be okay to keep the current implementation in this PR to reduce memory usage and update it to use the function later, after I add support for strided_matrix as its input in a separate PR for raft?

cpp/include/cuvs/preprocessing/quantize/binary.hpp
enp1s0: I kept the previous API.

enp1s0:
Thank you @tfeher for the review. I fixed the code. Can you review it again?

cpp/src/preprocessing/quantize/detail/binary.cuh
tfeher: That would make sense.

cjnolet (approved):

navneet1v:
@enp1s0 what is the performance of this type of quantization with GPU vs similar quantization happening on CPU nodes? Do you have benchmarking results?

enp1s0:
@navneet1v Sorry for the late reply. I don't have any CPU/GPU comparison numbers, but the performance gain by GPU for binary quantized datasets can be lower than for non-quantized datasets when the dataset dimension is not large enough (e.g., less than 1024~2048). This is because one of the performance limiters of the ANNS search is the memory bandwidth for dataset vector loading. We can get high search performance by using GPUs because one advantage of GPUs over CPUs is their high memory bandwidth. However, we can't utilize the hardware memory bandwidth efficiently if the dataset dimension is too small. Binary quantization reduces the dataset dimension's bit length from D*sizeof(T) bits to D bits, where D is the dimension. This makes it more difficult to utilize the hardware memory bandwidth efficiently. If we can't utilize the memory bandwidth efficiently, the search performance gain over CPUs may be lower.

PR #720, opened 2025-02-25
Update cuvs to properly create a NCCL::NCCL target

robertmaynard:
Previously cuvs was relying on the fact that nccl would be found when passed as `-lnccl` and be found on one of the implicit link directories.

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -583,10 +583,10 @@ if(BUILD_SHARED_LIBS)
         NCCL
         HEADER_NAMES nccl.h
         LIBRARY_NAMES nccl
-        )
+      )
       find_package(NCCL REQUIRED)
       target_link_libraries(cuvs_objs PRIVATE NCCL::NCCL)
-      target_link_libraries(cuvs PUBLIC NCCL::NCCL)
+      target_link_libraries(cuvs PRIVATE NCCL::NCCL)
     endif()
 
     # Keep cuVS as lightweight as possible. Only CUDA libs and rmm should be used in global target.
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -579,7 +579,14 @@ if(BUILD_SHARED_LIBS)
     )
 
     if(BUILD_MG_ALGOS)
-      set(CUVS_COMMS_DEPENDENCY nccl)
+      rapids_find_generate_module(
+        NCCL
+        HEADER_NAMES nccl.h
+        LIBRARY_NAMES nccl
+        )
+      find_package(NCCL REQUIRED)
+      target_link_libraries(cuvs_objs PRIVATE NCCL::NCCL)
+      target_link_libraries(cuvs PUBLIC NCCL::NCCL)
     endif()
 
     # Keep cuVS as lightweight as possible. Only CUDA libs and rmm should be used in global target.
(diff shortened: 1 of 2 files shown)


bdice:
@robertmaynard Does this need to be revived for 25.06? @divyegala may care about this as he's doing some NCCL packaging work.

robertmaynard:
> @robertmaynard Does this need to be revived for 25.06? @divyegala may care about this as he's doing some NCCL packaging work.

Yes it does

cpp/tests/CMakeLists.txt
```diff
@@ -199,14 +199,17 @@ if(BUILD_TESTS)
   )
 
   if(BUILD_CAGRA_HNSWLIB)
-    ConfigureTest(NAME NEIGHBORS_HNSW_TEST PATH neighbors/hnsw.cu GPUS 1 PERCENT 100)
+    ConfigureTest(
+      NAME NEIGHBORS_HNSW_TEST PATH neighbors/hnsw.cu GPUS 1 PERCENT 100 ADDITIONAL_DEP NCCL::NCCL
```
divyegala, line 203: Why was the link needed here? AFAIK this code path does not use NCCL.

divyegala: Do you have the trace? This seems like a leak to me. No problem if not, I can track it down - but shouldn't hold this PR on that, can fix as a follow up.

cpp/CMakeLists.txt
```diff
@@ -579,15 +579,22 @@ if(BUILD_SHARED_LIBS)
...
     if(BUILD_MG_ALGOS)
-      set(CUVS_COMMS_DEPENDENCY nccl)
+      rapids_find_generate_module(
+        NCCL
+        HEADER_NAMES nccl.h
+        LIBRARY_NAMES nccl
+      )
+      find_package(NCCL REQUIRED)
+      target_link_libraries(cuvs_objs PRIVATE NCCL::NCCL)
+      target_link_libraries(cuvs PUBLIC NCCL::NCCL)
```
divyegala, line 589: I believe this should be a `PRIVATE` link. @viclafargue correct me if I am wrong but does any cuvs header include NCCL header or a RAFT header that includes NCCL header?

viclafargue: Thanks for noticing this. The public cuVS MNMG API should only take in a `raft::resources`, everything related to NCCL is hidden inside the implementation.

divyegala: What is currently the status though? Are we exposing the NCCL header directly or indirectly to cuVS users from a header file?

cpp/tests/CMakeLists.txt
robertmaynard: re-testing locally and I can't reproduce the nccl dependency. So I am dropping this change

cpp/CMakeLists.txt
robertmaynard: re-testing localled shows that I can have this as PRIVATE.

divyegala (approved):

divyegala:
@robertmaynard is this PR ready for a merge once CI passes?

robertmaynard:
Yes it isOn Apr 16, 2025, at 2:23 PM, Divye Gala ***@***.***> wrote:﻿
  @robertmaynard is this PR ready for a merge once CI passes?—Reply to this email directly, view it on GitHub, or unsubscribe.You are receiving this because you were mentioned.Message ID: ***@***.***>



divyegala left a comment (rapidsai/cuvs#720)
@robertmaynard is this PR ready for a merge once CI passes?

PR #724, opened 2025-02-25
Unify use of OpenMP for HNSW threading models

divyegala:
`cuvs_bench` hnswlib wrapper was using a custom threading pool while `cuvs` hnsw wrapper was using OpenMP for parallelism. This was causing unexpected deviations in measured timings.

### DIFF
--- a/cpp/bench/ann/src/hnswlib/hnswlib_wrapper.h
+++ b/cpp/bench/ann/src/hnswlib/hnswlib_wrapper.h
@@ -66,13 +66,13 @@ class hnsw_lib : public algo<T> {
   struct build_param {
     int m;
     int ef_construction;
-    int num_threads = omp_get_num_procs();
+    int num_threads = omp_get_max_threads();
   };
 
   using search_param_base = typename algo<T>::search_param;
   struct search_param : public search_param_base {
     int ef;
-    int num_threads = 1;
+    int num_threads = omp_get_max_threads();
   };
 
   hnsw_lib(Metric metric, int dim, const build_param& param);
@@ -175,12 +175,7 @@ void hnsw_lib<T>::set_search_param(const search_param_base& param_, const void* 
   auto param     = dynamic_cast<const search_param&>(param_);
   appr_alg_->ef_ = param.ef;
   num_threads_   = param.num_threads;
-  // bench_mode_ = param.metric_objective;
   bench_mode_ = Mode::kLatency;  // TODO(achirkin): pass the benchmark mode in the algo parameters
-
-  // Create a pool if multiple query threads have been set and the pool hasn't been created already
-  bool create_pool = (bench_mode_ == Mode::kLatency && num_threads_ > 1 && !thread_pool_);
-  if (create_pool) { thread_pool_ = std::make_shared<fixed_thread_pool>(num_threads_); }
 }
 
 template <typename T>
@@ -192,7 +187,10 @@ void hnsw_lib<T>::search(
     get_search_knn_results(query + i * dim_, k, indices + i * k, distances + i * k);
   };
   if (bench_mode_ == Mode::kLatency && num_threads_ > 1) {
-    thread_pool_->submit(f, batch_size);
+#pragma omp parallel for num_threads(num_threads_)
+    for (int i = 0; i < batch_size; i++) {
+      f(i);
+    }
   } else {
     for (int i = 0; i < batch_size; i++) {
       f(i);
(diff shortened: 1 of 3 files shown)


cpp/bench/ann/src/hnswlib/hnswlib_wrapper.h
```diff
@@ -66,13 +66,13 @@ class hnsw_lib : public algo<T> {
   struct build_param {
     int m;
     int ef_construction;
-    int num_threads = omp_get_num_procs();
+    int num_threads = omp_get_max_threads();
```
achirkin, line 69: What are the consequences of switching from `omp_get_num_procs()` to `omp_get_max_threads()` with respect to the SMT/hyperthreading?
Does hyperthreading make HNSW slower or faster?
Also, does OpenMP take into account whether the number of cores is limited by `numactl`?

achirkin (commented):
Thanks, very welcome change, but I have a few questions to make sure it works as intended.

 - I don't see any changes to CMakeLists. Is the benchmark executable already linked with openmp?
 - Can we now remove `cpp/bench/ann/src/common/thread_pool.hpp` completely?

cpp/bench/ann/src/hnswlib/hnswlib_wrapper.h
divyegala: Answering your questions one by one:
> I don't see any changes to CMakeLists. Is the benchmark executable already linked with openmp?
All benchmark executables link to OpenMP already.

> What are the consequences of switching from omp_get_num_procs() to omp_get_max_threads() with respect to the SMT/hyperthreading?
`omp_get_num_procs()` only returns number of physical cores. The difference in search times are very visibly apparent when accounting for thread usage to be all available hyperthreads vs just using physical cores. It makes HNSW faster.

> Also, does OpenMP take into account whether the number of cores is limited by numactl?
No, it does not. We would either have to write a custom implementation or use some thread-pool library that can account for this. I think it would fall out of the scope of this PR and would be a general design philosophy discussion that we would have to undertake with the team, as we use OpenMP in quite a lot of places outside of HNSW.

achirkin: Thanks for the answers!

> The difference in search times are very visibly apparent when accounting for thread usage to be all available hyperthreads vs just using physical cores. It makes HNSW faster.

Could you please add this to the comment above where we set this default?

> > Can we now remove cpp/bench/ann/src/common/thread_pool.hpp completely?
>
> No, unfortunately the FAISS wrappers still use it

Would it be worth the effort to adjust FAISS wrappers to use OpenMP in this PR as well?

tfeher (approved):
Thanks Divye, the PR looks good to me!

cpp/bench/ann/src/hnswlib/hnswlib_wrapper.h
```diff
@@ -66,13 +67,13 @@ class hnsw_lib : public algo<T> {
...
```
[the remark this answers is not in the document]
achirkin: Hi @tarang-jain the problem you're talking about is in the python post-processing, right? This sounds like a fragile setup, as nothing prevents anyone from adding more algorithms with the same build/search parameters in future. Perhaps, it would be better to solve this problem for good where it occurs? You could either join the columns, or prepend build_/search_ to their names, or just not use names as keys (allow duplicates).
I also think it would be a bit confusing for the user to having differentiate between the two.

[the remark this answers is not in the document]
tarang-jain: That's right! It occurs in the Python post-processing. Prepending build_/search_ to the names seems like the easiest thing to do.

cpp/bench/ann/src/hnswlib/hnswlib_wrapper.h
tarang-jain: Taking care of Faiss as part of this PR: https://github.com/rapidsai/cuvs/pull/911

cpp/bench/ann/src/hnswlib/hnswlib_wrapper.h
```diff
@@ -192,7 +175,10 @@ void hnsw_lib<T>::search(
     get_search_knn_results(query + i * dim_, k, indices + i * k, distances + i * k);
   };
   if (bench_mode_ == Mode::kLatency && num_threads_ > 1) {
-    thread_pool_->submit(f, batch_size);
+#pragma omp parallel for num_threads(num_threads_)
```
tarang-jain, line 178: Can we introduce a conditional here so that the thread creation is done only when batch_size is greater than 1?
`#pragma omp parallel if(batch_size > 1) for num_threads(num_threads_)`

tarang-jain (changes requested):

PR #733, opened 2025-02-27
ANN_BENCH: Fix segfault in CAGRA wrapper when moving its graph

achirkin:
When the graph in CAGRA wrapper needs to be relocated to a different memory resource, the old one is destructed before the data is copied. This leads to a segfault during copy. The fix is to retain the old graph pointer until the data is copied.

This error is likely to happen when one runs CAGRA search benchmark sweeping over multiple values of `graph_memory_type`.

### DIFF
--- a/cpp/bench/ann/src/cuvs/cuvs_cagra_wrapper.h
+++ b/cpp/bench/ann/src/cuvs/cuvs_cagra_wrapper.h
@@ -252,11 +252,12 @@ void cuvs_cagra<T, IdxT>::set_search_param(const search_param_base& param,
     auto mr = get_mr(graph_mem_);
 
     // Create a new graph, then copy, and __only then__ replace the shared pointer.
-    // We shouldn't use index_->graph() here, because it may be just a view of graph_.
-    auto new_graph = raft::make_device_mdarray<IdxT, int64_t>(handle_, mr, graph_->extents());
+    auto old_graph =
+      index_->graph();  // view of graph_ if it exists, of an internal index member otherwise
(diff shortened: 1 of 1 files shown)


cpp/bench/ann/src/cuvs/cuvs_cagra_wrapper.h
```diff
@@ -250,13 +250,16 @@ void cuvs_cagra<T, IdxT>::set_search_param(const search_param_base& param,
...
-      handle_, mr, raft::make_extents<int64_t>(index_->graph().extent(0), index_->graph_degree()));
 
-    raft::copy(graph_->data_handle(),
-               index_->graph().data_handle(),
-               index_->graph().size(),
+    // Create a new graph, then copy, and __only then__ replace the shared pointer.
+    // We shouldn't use index_->graph() here, because it may be just a view of graph_.
+    auto new_graph = raft::make_device_mdarray<IdxT, int64_t>(handle_, mr, graph_->extents());
+    raft::copy(new_graph.data_handle(),
+               graph_->data_handle(),
```
tfeher, line 258: Shouldn't we use `index_->graph()->data_handle()` here? Whether or not it is a view of `graph_`, it shall always point to a graph once the index is loaded.

When the index is just `load()`-ed, we do not populate `graph_`, therefore we should not use that as as a source.

achirkin: You're right, I've updated these lines to use the index_->graph().

Issue #718, opened 2025-02-24
[BUG] cuVS Scaler Quantization memory issue


singhmanas1:
**Describe the bug**
Running into `std::bad_alloc: out_of_memory: CUDA error` while training scalar embeddings with cuVS. Runs below 800,000 embeddings but throws error after. 

Embedding model - https://build.nvidia.com/nvidia/llama-3_2-nv-embedqa-1b-v2/modelcard
Dimensions - 2048
Precision - FP16

**Steps/Code to reproduce bug**
```
def scalar_quantize_cuvs(embeddings):
    embeddings_gpu = cp.asarray(embeddings)     
    params = scalar.QuantizerParams(quantile=0.99)
    quantizer = scalar.train(params, embeddings_gpu)
    transformed = scalar.transform(quantizer, embeddings_gpu)
    return transformed, quantization_time
```

**Error Statement**
```
Error during quantization: std::bad_alloc: out_of_memory: CUDA error at: [/pyenv/versions/3.12.9/lib/python3.12/site-packages/librmm/include/rmm/mr/device/cuda_memory_resource.hpp:62](https://vscode-remote+ssh-002dremote-002b10-002e176-002e1-002e125.vscode-resource.vscode-cdn.net/pyenv/versions/3.12.9/lib/python3.12/site-packages/librmm/include/rmm/mr/device/cuda_memory_resource.hpp:62): cudaErrorMemoryAllocation out of memory
Traceback:
Traceback (most recent call last):
  File "/tmp/ipykernel_765663/1026619002.py", line 35, in scalar_quantize_cuvs
    quantizer = scalar.train(params, embeddings_gpu)
                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "resources.pyx", line 110, in cuvs.common.resources.auto_sync_resources.wrapper
  File "scalar.pyx", line 126, in cuvs.preprocessing.quantize.scalar.scalar.train
  File "exceptions.pyx", line 37, in cuvs.common.exceptions.check_cuvs
[...]

PR #736, opened 2025-02-28
Reduce memory consumption of scalar quantizer

mfoerste4:
This is an optimization raised by #718 . 

The quantile selection process currently relies on sampling. The memory requirements for the originally used 
`raft::random::sample_without_replacement`
are roughly `48*input_len bytes` (`4*len*8byte + ~2*len*8byte for sort`). 
This has now been replaced by `raft::matrix::sample_rows` which essentially has the same requirement but based on the sample size instead of the original input.

### DIFF
--- a/cpp/src/preprocessing/quantize/detail/scalar.cuh
+++ b/cpp/src/preprocessing/quantize/detail/scalar.cuh
@@ -72,10 +72,11 @@ struct quantize_op {
   }
 };
 
-template <typename T>
-std::tuple<T, T> quantile_min_max(raft::resources const& res,
-                                  raft::device_matrix_view<const T, int64_t> dataset,
-                                  double quantile)
+template <typename T, typename IdxT = int64_t, typename accessor>
+std::tuple<T, T> quantile_min_max(
+  raft::resources const& res,
+  raft::mdspan<const T, raft::matrix_extent<IdxT>, raft::row_major, accessor> dataset,
+  double quantile)
 {
   // settings for quantile approximation
   constexpr size_t max_num_samples = 1000000;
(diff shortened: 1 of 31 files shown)


cpp/src/preprocessing/quantize/detail/scalar.cuh
```diff
@@ -87,10 +87,11 @@ std::tuple<T, T> quantile_min_max(raft::resources const& res,
...
   size_t subset_size = std::min(max_num_samples, n_elements);
-  auto subset        = raft::make_device_vector<T>(res, subset_size);
-  auto dataset_view  = raft::make_device_vector_view<const T>(dataset.data_handle(), n_elements);
-  raft::random::sample_without_replacement(
-    res, rng, dataset_view, std::nullopt, subset.view(), std::nullopt);
+
+  // select subsample element-wise
+  auto dataset_view =
+    raft::make_device_matrix_view<const T>(dataset.data_handle(), n_elements, 1UL);
+  auto subset = raft::matrix::sample_rows(res, rng, dataset_view, subset_size);
```
tfeher, line 94: I wonder whether we should check the available memory for before we try to allocate, or just make error reporting more informative, like below.
```suggestion
  auto subset = make_device_matrix<T, int64_t>(res, 0, 0);
  try {
    subset = raft::matrix::sample_rows(res, rng, dataset_view, subset_size);
   } catch (raft::logic_error& e) {
     // Informative error message, e.g. user can provide scaling factor to avoid extra memory needed for subsampling.
     throw;
   }
```

tfeher (commented):
Thanks Malte, it looks good, just one comment below.

tfeher (approved):
Thanks Malte for the update, LGTM.

PR #741, opened 2025-03-01
Remove unused raft cagra header in add_nodes.cuh

jiangyinzuo:
Vector search headers in raft will be removed in the future, so we should not include the raft cagra header.


tarang-jain (commented):
Can you rebase to 25.06?

cjnolet (approved):

PR #751, opened 2025-03-06
feat(libcuvs): port libcuvs to rattler-build

gforsyth:
This PR ports cuVS over to `rattler-build` for building conda packages.  

Some changes to the build process:

* `cuvs-bench` was recompiling the entire library separately -- caching was hopefully helping with this but it was an unnecessary double-compile.  Instead, I've added the `bench-ann` to the pre-build stage in `libcuvs` and create a `libcuvs-bench` output there, that `cuvs-bench` then depends on to provide the necessary shared objects.

* Edited `build.sh` so that if `-n` is passed, the `cuvs-bench` Python package is not installed.

* `libcuvs-static` seems to be a leftover from a previous incarnation -- the existing `libcuvs-static` packages on `rapidsai-nightly` consist only of build metadata files and are otherwise empty.  I've removed the unused `--compile-static-lib` option from `build.sh`


cjnolet:
@gforsyth libcuvs-static broke at some point and actually needs to be fixed.

gforsyth:
> @gforsyth libcuvs-static broke at some point and actually needs to be fixed.

Can you define for me what `libcuvs-static` is supposed to be?  Because the `static` build flag has been inoperable for >2 years

conda/recipes/cuvs-bench/recipe.yaml
```diff
@@ -0,0 +1,83 @@
...
+      - libaio
+      - libboost
+      - libcublas
+      - mkl
+      - if: cuda_major == "11"
+        then: cudatoolkit
+
+tests:
+  - script:
+      - test lib/python${{ py_version }}/site-packages/cuvs_bench/config/datasets/datasets.yaml
```
bdice, line 78: Does this work? It gets rid of `${{ py_version }}`, which may not be enough information if we support free-threaded builds in the future.
```suggestion
      - test "$(python -m site --user-site)"/cuvs_bench/config/datasets/datasets.yaml
```

conda/recipes/cuvs/recipe.yaml
```diff
@@ -0,0 +1,97 @@
...
+schema_version: 1
+
+context:
+  version: ${{ env.get("RAPIDS_PACKAGE_VERSION") }}
+  minor_version: ${{ (version | split("."))[:2] | join(".") }}
+  cuda_version: ${{ (env.get("RAPIDS_CUDA_VERSION") | split("."))[:2] | join(".") }}
+  cuda_major: '${{ (env.get("RAPIDS_CUDA_VERSION") | split("."))[0] }}'
+  date_string: '${{ env.get("RAPIDS_DATE_STRING") }}'
+  py_version: ${{ env.get("RAPIDS_PY_VERSION") }}
+  py_buildstring: ${{ py_version | version_to_buildstring }}
```
bdice, line 11: Is `py_buildstring` going to be treated as an integer instead of a string? Should it be single-quoted? (Probably applies to many recipes.)

conda/recipes/cuvs/recipe.yaml
```diff
@@ -0,0 +1,97 @@
...
+        - cuda-python >=11.8.5,<12.0a0
+        - cudatoolkit
+      else:
+        - cuda-python >=12.6.2,<13.0a0
+        - cuda-cudart-dev
+  run:
+    - ${{ pin_compatible("cuda-version", upper_bound="x", lower_bound="x") }}
+    - libcuvs =${{ version }}
+    - pylibraft =${{ minor_version }}
+    - python
```
bdice, line 72: I can't remember if this needs to be constrained to `=${{ py_version }}` or if that comes automatically from the `host` pinning of `python`. Please check.

conda/recipes/libcuvs/recipe.yaml
```diff
@@ -0,0 +1,408 @@
...
+        - ${{ pin_compatible("cuda-version", upper_bound="x", lower_bound="x") }}
+        - libraft-headers =${{ minor_version }}
+        - nccl
+        - if: cuda_major == "11"
+          then:
+            - cuda-cudart
+            - libcublas
+            - libcurand
+            - libcusolver
+            - libcusparse
```
bdice, line 141: I am guessing we need `cudatoolkit` on CUDA 11 and this list of math libraries on CUDA 12+?
```suggestion
        - if: cuda_major == "11"
          then:
            - cudatoolkit
          else:
            - cuda-cudart
            - libcublas
            - libcurand
            - libcusolver
            - libcusparse
```

conda/recipes/libcuvs/recipe.yaml
```diff
@@ -0,0 +1,408 @@
...
+          CMAKE_CXX_COMPILER_LAUNCHER: ${{ env.get("CMAKE_CXX_COMPILER_LAUNCHER") }}
+          CMAKE_GENERATOR: ${{ env.get("CMAKE_GENERATOR") }}
+          PARALLEL_LEVEL: ${{ env.get("PARALLEL_LEVEL") }}
+          RAPIDS_ARTIFACTS_DIR: ${{ env.get("RAPIDS_ARTIFACTS_DIR") }}
+          SCCACHE_BUCKET: ${{ env.get("SCCACHE_BUCKET") }}
+          SCCACHE_IDLE_TIMEOUT: ${{ env.get("SCCACHE_IDLE_TIMEOUT") }}
+          SCCACHE_REGION: ${{ env.get("SCCACHE_REGION") }}
+          SCCACHE_S3_USE_SSL: ${{ env.get("SCCACHE_S3_USE_SSL") }}
+          SCCACHE_S3_NO_CREDENTIALS: ${{ env.get("SCCACHE_S3_NO_CREDENTIALS") }}
+          SCCACHE_S3_KEY_PREFIX: libcuvs-${{ env.get("RAPIDS_CONDA_ARCH") }}
```
bdice, line 266: Probably a discussion for later -- we key our sccache on `${library}-${arch}`. In that cache, we have all kinds of artifacts that are incompatible because they are produced for different CUDA versions. Should CUDA version be a part of the sccache key? Or should we remove the arch prefix? The status quo isn't making a lot of sense as I consider it further.

conda/recipes/libcuvs/recipe.yaml
```diff
@@ -0,0 +1,408 @@
...
+    build:
+      script:
+        content: |
+          cmake --install cpp/build --component ann_bench
+
+          if [ "$RAPIDS_CONDA_ARCH" != "aarch64" ]; then
+            # Need this so from the diskann compile, but there isn't a
+            # component associated with it. Probably a better way to do this.
+            cp cpp/build/_deps/diskann-build/src/libdiskann.so $PREFIX/lib/
+          fi
```
bdice, line 344: Let's file an issue for follow-up here.

conda/recipes/libcuvs/recipe.yaml
```diff
@@ -0,0 +1,408 @@
...
+      build:
+        - cmake ${{ cmake_version }}
+        - ${{ stdlib("c") }}
+      host:
+        - librmm =${{ minor_version }}
+        - libraft-headers =${{ minor_version }}
+        - nccl ${{ nccl_version }}
+        - ${{ pin_subpackage("libcuvs", exact=True) }}
+        - cuda-version =${{ cuda_version }}
+        - openblas # required by some CPU algos in benchmarks
```
bdice, line 358: Separate comments in YAML by two spaces.
```suggestion
        - openblas  # required by some CPU algos in benchmarks
```

conda/recipes/libcuvs/recipe.yaml
```diff
@@ -0,0 +1,408 @@
...
+        - libraft-headers =${{ minor_version }}
+        - nccl
+        - if: cuda_major == "11"
+          then:
+            - cuda-cudart
+            - libcublas
+            - libcurand
+            - libcusolver
+            - libcusparse
+      ignore_run_exports:
```
bdice, line 395: `mkl` isn't being ignored for this package, as expected. Is the run-exported pinning for `mkl` tight enough / loose enough for what we want here?

bdice (commented):
A few comments but overall this is great. I'm really glad this rattler conversion is getting easier as we progress through more RAPIDS libraries.

conda/recipes/cuvs/recipe.yaml
gforsyth: No, `version_to_buildstring` makes sure it is a string. I'm waiting for upstream changes to make that easier for other context variables

conda/recipes/cuvs/recipe.yaml
gforsyth: This doesn't need to be constrained because the `python_abi` is inherited from `host`, e.g.

`python_abi       ┆ 3.11.* *_cp311 (RE of [host: python])     `

conda/recipes/libcuvs/recipe.yaml
gforsyth: You are right.  I am not sure I like that this passed tests like this

conda/recipes/libcuvs/recipe.yaml
gforsyth: hm, that's a good point.  I would defer to some of the more experienced `sccache` folk but it does seem like we would benefit from a more specific key

conda/recipes/libcuvs/recipe.yaml
```diff
@@ -0,0 +1,408 @@
...
+        - ${{ pin_compatible("cuda-version", upper_bound="x", lower_bound="x") }}
+        - libraft-headers =${{ minor_version }}
+        - nccl
+        - if: cuda_major == "11"
+          then:
+            - cuda-cudart
+            - libcublas
+            - libcurand
+            - libcusolver
+            - libcusparse
```
gforsyth, line 394: I need to fix this to also be for cuda 12 and `cudatoolkit` for cuda 11

conda/recipes/libcuvs/recipe.yaml
gforsyth: nice catch.  I've added an ignore here for the host version and explicitly added the pin to `run`

conda/recipes/libcuvs/recipe.yaml
gforsyth: #760

conda/recipes/cuvs-bench/recipe.yaml
gforsyth: Ok, there are file existence checks that `rattler` supports which can accept globs, which I think will be a bit more robust

gforsyth:
This is ready for another look @bdice

conda/recipes/cuvs-bench-cpu/recipe.yaml
```diff
@@ -0,0 +1,110 @@
...
+  script:
+    content: |
+
+      # Remove `-fdebug-prefix-map` line from CFLAGS and CXXFLAGS so the
+      # incrementing version number in the compile line doesn't break the
+      # cache
+      set -x
+      export CFLAGS=$(echo $CFLAGS | sed -E 's@\-fdebug\-prefix\-map[^ ]*@@g')
+      export CXXFLAGS=$(echo $CXXFLAGS | sed -E 's@\-fdebug\-prefix\-map[^ ]*@@g')
+      set +x
```
vyasr, line 30: Suggestion: https://github.com/rapidsai/rapids-cmake/issues/798#issuecomment-2737455259

conda/recipes/cuvs-bench/recipe.yaml
vyasr, line 1: Future suggestion: rather than maintaining separate cuvs-bench and cuvs-bench-cpu recipes, we should parametrize the recipe. The two should be almost identical.

vyasr (approved):
Couple notes for future work, but LGTM.

gforsyth:
Followups recorded in https://github.com/rapidsai/build-planning/issues/47#issuecomment-2696125711

PR #763, opened 2025-03-14
[BUG] Fix graph index sorting in CAGRA graph build by NN Descent 

enp1s0:
Graph neighbor sorting after the initial kNN graph construction by NN descent always uses the L2 norm, while NN descent also supports the inner product and cosine distance. This PR adds support for the inner product and cosine distance in graph neighbor sorting.

## TODO
- [x] Normalize vector when metric == Cosine


tfeher (approved):
Thanks @enp1s0 for the PR, it looks good to me!

PR #766, opened 2025-03-15
add function to build specific gpu-arches [REVIEW] 

nvrohanv:
This PR adds in functions to the root level build.sh so that you can specify which GPU architectures you want to build for. Currently the only options were to build for the current system or for all systems. It also adds this option into `examples/build.sh` in addition to fixing bugs with the clean command.

A `--gpu-arch` option is added that is meant to be used to provide which architectures to build for e.g. `./build.sh bench-ann --gpu-arch="80-real;90-real"`


build.sh
```diff
@@ -195,6 +195,37 @@ function buildMetrics {
     fi
 }
 
+function gpuArch {
+    # Check if both --gpu-arch and --allgpuarch are specified
+    if hasArg --allgpuarch && [[ -n $(echo $ARGS | { grep -E "\-\-gpu\-arch" || true; } ) ]]; then
+        echo "Error: Cannot specify both --gpu-arch and --allgpuarch"
+        echo "Use either:"
+        echo "  --gpu-arch=\"80-real;90-real\"    (for specific architectures)"
```
bkarsin, line 203: Can we add a link or reference about different architecture numbers and `real` vs. `virtual`? Some users may not know where to find this info.

nvrohanv: Added in instructions and a link in the help message for the main command. the example build.sh seems more barebone and doesn't have a help message so didn't add anything there, but presumably someone would've used the main build.sh first.

bkarsin (approved):

cjnolet (approved):
Very useful. Great to see users can't specify both allgpuarch and gpuarchs. Thanks!

lowener (approved):
LGTM

gforsyth (approved):

PR #768, opened 2025-03-15
Expose IVF PQ params in CAGRA C API

divyegala:


cpp/include/cuvs/neighbors/cagra.h
```diff
@@ -378,6 +379,12 @@ cuvsError_t cuvsCagraBuild(cuvsResources_t res,
                            DLManagedTensor* dataset,
                            cuvsCagraIndex_t index);
 
+cuvsError_t cuvsCagraBuildWithIVFPQ(cuvsResources_t res,
```
benfred, line 382: Rather than a separate function - could we just add the ivf-pq params to the cuvsCagraIndexParams object and then just expose the single cuvsCagraBuild function ?

We already have the `cuvsCagraGraphBuildAlgo` enum to specify building with ivf-pq in `cuvsCagraBuild` - we just need to add the ivf-pq params object so that we don't use the default.

divyegala: How do you want me to express the optional nature here? By defaulting to a NULL pointer?

divyegala: @benfred I reworked the PR

cpp/src/neighbors/cagra_c.cpp
```diff
@@ -349,17 +375,21 @@ extern "C" cuvsError_t cuvsCagraSearch(cuvsResources_t res,
...
-                                       .intermediate_graph_degree = 128,
-                                       .graph_degree              = 64,
-                                       .build_algo                = IVF_PQ,
-                                       .nn_descent_niter          = 20};
+    *params                       = new cuvsCagraIndexParams{.metric                    = L2Expanded,
+                                                             .intermediate_graph_degree = 128,
+                                                             .graph_degree              = 64,
+                                                             .build_algo                = IVF_PQ,
+                                                             .nn_descent_niter          = 20};
+    (*params)->graph_build_params = new cuvsIvfPqParams{nullptr, nullptr};
```
tfeher, line 383:
```suggestion
    (*params)->graph_build_params = new cuvsIvfPqParams{nullptr, nullptr, 1};
```

tfeher (changes requested):
Thanks Divye for the PR! I suggest to add the `refinement_rate` parameter as well, otherwise the PR looks good.

tfeher (approved):
Thanks Divye for the updates, LGTM!

PR #769, opened 2025-03-17
CAGRA search: int64_t indices in the API

achirkin:
Add `int64_t` as the type of returned neighbors in cagra::search while keeping the index type unchanged.

Down the implementation, the PR introduces the `OutputIndexT` template parameter and handles it differently for each implementation:
  - multi-kernel implementation already does the copy after search in a separate kernel, we change the index type during this copy.
  - multi-cta kernel runs as before, followed by top-k; conversion of the index type happens in a separate map kernel only if a simple type cast wouldn't suffice (i.e. data sizes do not match).
  - single-cta kernel must always run as a single kernel to maintain 'persistent' mode performance. I erase the index type in the kernel signature and pass the `sizeof(OutputIndexT)` as a tag in the pointer. Based on this tag, a proper output-writing function is dispatched in the epilogue of the single-cta kernel. By using this technique I avoid instantiating multiple kernels at the negligible runtime cost. 

Note: we cannot simply instantiate cagra with index type `int64_t`, because this will make the graph twice larger and will reduce the search throughput.

To test the new search overloads, I turn `AnnCagraTest::testCagra` into a template and add just two instances to the `*.cu files` to avoid bloating the number of tests.

The new search overloads are used in the ann-bench wrappers, which slightly improves the reported QPS in some cases due to reducing the index conversion overhead.


cpp/src/neighbors/detail/cagra/search_single_cta_kernel-inl.cuh
```diff
@@ -516,7 +517,7 @@ template <unsigned MAX_ITOPK,
           class DATASET_DESCRIPTOR_T,
           class SAMPLE_FILTER_T>
 __device__ void search_core(
-  typename DATASET_DESCRIPTOR_T::INDEX_T* const result_indices_ptr,       // [num_queries, top_k]
+  uintptr_t result_indices_ptr,                                           // [num_queries, top_k]
```
tfeher, line 520: Please you add a docstring, to describe the arguments.

cpp/src/neighbors/detail/cagra/search_single_cta_kernel-inl.cuh
```diff
@@ -863,15 +864,37 @@ __device__ void search_core(
     __syncthreads();
   }
 
+  // NB: tag the indices pointer with its element size.
+  //     This allows us to avoid multiplying kernel instantiations
+  //     and any costs for extra registers in the kernel signature.
```
tfeher, line 869:
```suggestion
  // NB: The indices pointer is tagged with its element size.
  //     Here we select the correct conversion operator at runtime.
  //     This allows us to avoid multiplying kernel instantiations
  //     and any costs for extra registers in the kernel signature.
```

tfeher (changes requested):
Thanks Artem, it looks good, just a few documentation issues. It is great that you could implement this without additional kernel instantiations.

tfeher (approved):
Thanks Artem for the update LGTM!

PR #781, opened 2025-03-25
Fix use of `rapids_cpm_generate_patch_command`

miscco:
Adopt to recent changes in rapids-cmake


miscco:
@KyleFromNVIDIA beat me by 30 min

PR #785, opened 2025-03-25
Multi-GPU batched knn graph building

jinsolp:
## Description
This PR adds a feature for multi-GPU batched all-neighbors knn graph building. Enables building of all-neighbors graph for datasets that do not fit on device memory.

**[Supported ANN algorithms]**

1. IVFPQ: higher recall, takes longer to run
2. NN Descent: lower recall in general (and has unacceptable low recall issues for some data distributions), but faster to run. Also uses less memory (since it uses half precision)

**[Supported data types]**
Currently only supports fp32 data types, but plan to explore and add fp16 data types in the future. 

**[Number of nearest clusters and total number of clusters]**
`batch_ann::index` has two important configurable parameters.
- `n_nearest_clusters`: number of clusters each data is assigned to
- `n_clusters`: total number of clusters

Some important notes about these two parameters;
1. the ratio of `n_nearest_clusters` / `n_clusters` determines device memory usage.
    - Approximately (`n_nearest_clusters` / `n_clusters` ) * `num_rows_in_entire_data` number of rows will be put on device memory at once.
    - E.g. between (`n_nearest_clusters` / `n_clusters` ) = 2/10 and 2/20, the latter will use less device memory.
2. larger `n_nearest_clusters` results in better accuracy of the final all-neighbors knn graph. 
    - E.g. With the similar device memory usages, (`n_nearest_clusters` / `n_clusters`) = 4/20 will have better accuracy than 2/10 at the cost of performance.

---
## How to Use
Can be used like the following
```
// in test.cpp file
batch_ann::index_params params;
params.n_clusters         = 10; // total number of clusters
params.n_nearest_clusters = 2; // number of clusters each data is assigned to

// Option 1: For usage with IVFPQ
params.graph_build_params = batch_ann::graph_build_params::ivf_pq_params{};
// Option 2: For usage with NN Descent
params.graph_build_params = batch_ann::graph_build_params::nn_descent_params{};

auto index = batch_ann::build(handle_, host_data_view, k, params);
```

the `batch_ann::build` internally runs on multi gpu or single gpu depending on visible cuda devices.

```
CUDA_VISIBLE_DEVICES=0 ./test.         // runs single GPU batching
CUDA_VISIBLE_DEVICES=0,1,2,3 ./test    // uses 4 GPUS for batching
```

---
## Numbers
All experiments run on L40 GPUs
**[Large dataset to check performance and memory usage]**
Below is a result of running on a 269GB dataset (wiki all).
- IVFPQ takes longer (because it goes through build - search -refine steps), and uses more memory. The reason for using more memory is because NN Descent puts data on device in half precision floating points, regardless of the given data type. Plan to test the impact of using fp16 IVFPQ on recall in the future.
- Gains are better for IVFPQ because it does more work on the GPU compared to NN Descent.
- With longer end to end runtime and more memory usage (unless it uses fp16), IVFPQ is still worth using because of recall issues with NN Descent.
Screenshot 2025-03-25 at 1 57 53 PM

**[Smaller dataset to verify recall]**
It is impossible to generate ground truth all-neighbors graph for a large dataset, so below is the result on a 3.8GB dataset (Deep images) to check how `n_nearest_clusters` and `n_clusters` affects recall.
- Recalls consistent regardless of how many GPUs are used
- Note the recall differences between (`n_nearest_clusters` / `n_clusters` ) = 2/20 and 4/40. These two configurations use similar amount of device memory, but the latter has much better recall
- Note that NN Descent recalls are generally lower than IVFPQ recalls (NN Descent results would have been much worse for other data distributions at this data size)
- At this data size, adding more GPUs doesn't always help
Screenshot 2025-03-25 at 2 09 01 PM


cpp/include/cuvs/neighbors/batch_ann.hpp
```diff
@@ -0,0 +1,194 @@
+/*
+ * Copyright (c) 2025, NVIDIA CORPORATION.
```
cjnolet, line 2: I'm trying to keep these specialized "multi-GPU" apis out of the public APIs as much as possible so users don't need to think much about. You can imagine if we have different includes and completley different APIs for single node multi-gpu, multi-node multi-gpu, batched vs non-batched, things will get complicated very quickly. Instead, we've opted for having APIs that can be configured easily (and implicitly where at all possible) so that the single-gpu APIs can function multi-gpu under the proper circumstances. This also means users can change one line of code to use the multi-gpu apis and not have to invoke specialized functions (we are going this way w/ UMAP also).

cpp/src/neighbors/batch_ann/batch_ann.cu
```diff
@@ -0,0 +1,44 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include "batch_ann.cuh"
+#include "cuvs/neighbors/common.hpp"
+
+namespace cuvs::neighbors::batch_ann {
```
cjnolet, line 20: Since you've redefined a new interface, it means these all need to be explicitly compiled. If you instead just have the single-gpu apis dispatch to the batched code when the settings meet the proper criteria, you won't need to explicitly instantiate the APIs.

cpp/src/neighbors/batch_ann/batch_ann.cuh
```diff
@@ -0,0 +1,476 @@
...
+#include <cstddef>
+#include <cuvs/neighbors/batch_ann.hpp>
+#include <cuvs/neighbors/brute_force.hpp>
+#include <cuvs/neighbors/refine.hpp>
+#include <raft/core/device_mdarray.hpp>
+#include <raft/core/host_mdarray.hpp>
+#include <raft/core/host_mdspan.hpp>
+#include <raft/core/managed_mdarray.hpp>
+#include <raft/core/resource/nccl_clique.hpp>
+#include <rmm/mr/device/statistics_resource_adaptor.hpp>
```
cjnolet, line 30: This is for debugging so we probably don't want this in the final code. We probably don't want `nccl_clique` in there either. Please cleanu p the includes in this file to only include what's actually being used.

cpp/include/cuvs/neighbors/batch_ann.hpp
cjnolet: Nevermind this comment- I misunderstood the contents of this PR.

cpp/include/cuvs/neighbors/batch_ann.hpp
```diff
@@ -0,0 +1,194 @@
...
+ */
+
+#pragma once
+
+#include <cuvs/neighbors/ivf_pq.hpp>
+#include <cuvs/neighbors/nn_descent.hpp>
+
+#include <variant>
+
+namespace cuvs::neighbors::batch_ann {
```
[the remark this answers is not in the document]
cjnolet: I think we should look for an alternative name here- I think we can find a better name than "batch_ann". Maybe we could find a name that implies more about the fact that we're building an all-neighbors graph.

cpp/include/cuvs/neighbors/batch_ann.hpp
```diff
@@ -0,0 +1,194 @@
...
+  auto operator=(index&&) -> index&      = default;
+  ~index()                               = default;
+
+ private:
+  raft::resources const& res_;
+  int64_t k_;
+  bool return_distances_;
+  raft::host_matrix<IdxT, IdxT, raft::row_major> graph_;
+  std::optional<raft::device_matrix<DistT, IdxT, row_major>> distances_;
+  raft::host_matrix_view<IdxT, IdxT, raft::row_major> graph_view_;
```
cjnolet, line 118: Views are super cheap to create. If we are storing the owning structure here, do we still need the view?

cpp/include/cuvs/neighbors/batch_ann.hpp
```diff
@@ -0,0 +1,194 @@
...
+  std::optional<raft::device_matrix_view<DistT, IdxT, row_major>> distances_view_;
```
cjnolet, line 119: Same here, if we store the owning structure, we can create a view on the fly which is super cheap (it's literally just creating an object wrapped around the array in memory.

cpp/include/cuvs/neighbors/batch_ann.hpp
```diff
@@ -0,0 +1,194 @@
...
+/**
+ * @brief ANN parameters used by the batching algorithm to build an all-neighbors knn graph
+ */
+namespace graph_build_params {
+
+/** Specialized parameters utilizing IVF-PQ to build knn graph */
+struct ivf_pq_params {
+  cuvs::neighbors::ivf_pq::index_params build_params;
+  cuvs::neighbors::ivf_pq::search_params search_params;
+  float refinement_rate = 2.0;
```
cjnolet, line 131: We might want to turn off refinement initially (set it to 1.0).

cpp/src/neighbors/batch_ann/batch_ann.cuh
```diff
@@ -0,0 +1,476 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+#include "../detail/ann_utils.cuh"
```
cjnolet, line 18: Everytime we include (and use) detail code, it means any kernels or device code has to be recompiled, which increases compile time and binary size. We should be very careful to make sure we're only using public APIs as much as possible, especially when we're building things on top like this API is doing. There's no actual public API for ann_utils, so my hope is that you aren't actually using it and we can just remove this include.

cpp/src/neighbors/batch_ann/batch_ann.cuh
```diff
@@ -0,0 +1,476 @@
...
+                          batch_params,
+                          cluster_data_view,
+                          inverted_indices_view,
+                          global_neighbors,
+                          global_distances);
+  }
+}
+
+template <typename T, typename IdxT>
+void multi_gpu_batch_build(const raft::resources& handle,
```
cjnolet, line 301: Very nice- I like that you're dispatching instead of providing two different public APIs.

cpp/src/neighbors/batch_ann/batch_ann.cuh
```diff
@@ -0,0 +1,476 @@
...
+  }
+}
+
+template <typename T, typename IdxT>
+void build(const raft::resources& handle,
+           raft::host_matrix_view<const T, IdxT, row_major> dataset,
+           const index_params& batch_params,
+           batch_ann::index<IdxT, T>& index)
+{
+  auto start      = raft::curTimeMillis();
```
cjnolet, line 377: We should probably remove this before the PR is merged.

cpp/src/neighbors/batch_ann/batch_ann_builder.cuh
```diff
@@ -0,0 +1,553 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+#pragma once
+
+#include "../detail/nn_descent.cuh"
```
cjnolet, line 18: Please use public APIs only. If something might require a specialized API that can only be found in "detail" namespace then that's a sign it should be getting exposed publicly (usually we do this by adding a "helpers" namespace in the public APIs.

cpp/src/neighbors/batch_ann/batch_ann_builder.cuh
```diff
@@ -0,0 +1,553 @@
...
+#include <raft/core/resource/cuda_stream.hpp>
+#include <raft/core/resources.hpp>
+#include <raft/util/cudart_utils.hpp>
+
+namespace cuvs::neighbors::batch_ann::detail {
+using namespace cuvs::neighbors;
+using align32 = raft::Pow2<32>;
+
+template <typename KeyType, typename ValueType>
+struct KeyValuePair {
```
cjnolet, line 39: Why are we redefining this object? There's a keyvalue pair type in RAFT

cpp/include/cuvs/neighbors/batch_ann.hpp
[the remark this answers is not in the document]
jinsolp: Yes the name is a bit misleading 😅
From a big picture, all this PR does is "building an all neighbors graph (that is somewhat approximate) using batching", exploiting all visible CUDA devices

cpp/src/neighbors/batch_ann/batch_ann.cuh
jinsolp: the `nccl_clique` is for figuring out how many visible CUDA devices there are when running the program. Is there a better way to figure that out? (to decide whether we want to go multi-gpu or single gpu)?

cpp/include/cuvs/neighbors/batch_ann.hpp
jinsolp: The views are here for a case where the memory is already allocated outside (like with UMAP). In such cases, the index is initialized like this;

```
cuvs::neighbors::batch_ann::index<int64_t, float> idx{
        handle,
        raft::make_host_matrix_view<float, int64_t>(out.knn_indices, inputsA.n, n_neighbors),
        raft::make_device_matrix_view<float, int64_t>(out.knn_dists, inputsA.n, n_neighbors)};
```

We can avoid allocating another chunk of memory and then having to copy the results back to what's allocated.

cpp/src/neighbors/batch_ann/batch_ann.cuh
jinsolp: Yes, I'm not sure why it slipped in...

cpp/src/neighbors/batch_ann/batch_ann_builder.cuh
jinsolp: Thank you for the comment Corey, but with NN Descent, this complicates things.

The actual NN descent `build` that does all the work is a function of the `GNND` class in `detail/nn_descent.cuh`, which allocates all necessary device and host matrices at initialization.
The NN Descent's `build` exposed as the public API makes a new GNND, then calls`gnnd.build()`.
So it's like this
```
public api build() {
     gnnd = GNND{};   // allocates a bunch of device/host memory
     gnnd.build(data);    // where all the work is done
}
```

However, now that we have to build iteratively, calling the public `build` API means we make GNND for every new cluster, thereby allocating (and deallocating) a large chunk of memory for every cluster. This takes a lot of time.
 
So what happens is right now I am making a GNND once, and reusing it over multiple clusters to save time.

```
// what's happening in this PR
gnnd = GNND{}    // this needs to be made outside the for loop
for cluster in clusters:
    cluster_data = get_cluster_data(data)
    gnnd.build(cluster_data)
```

Thus, if we didn't want to include the `detail/nn_descent.cuh` file here, we would have to expose the `GNND` class publicly, but that doesn't seem right.
Any suggestions would be greatly appreciated!

cpp/src/neighbors/batch_ann/batch_ann.cu
jinsolp: I'm guessing this comment is from _before_ we decided to expose this as a public API?

cpp/src/neighbors/batch_ann/batch_ann_builder.cuh
jinsolp: Handling this with [this new PR](https://github.com/rapidsai/cuvs/pull/803) to move declarations in a .hpp file instead (so that here we can include `nn_descent_gnnd.hpp` instead of the entire `nn_descent.cuh` file.

divyegala (commented):
This is a non-blocking review, and it's mostly just bookkeeping questions.

cpp/src/neighbors/all_neighbors/all_neighbors_batched.cuh
```diff
@@ -0,0 +1,499 @@
...
+                           const index_params& params,
+                           raft::host_vector_view<IdxT, IdxT, row_major> cluster_sizes,
+                           raft::host_vector_view<IdxT, IdxT, row_major> cluster_offsets,
+                           raft::host_vector_view<IdxT, IdxT, row_major> inverted_indices)
+{
+  size_t num_rows = dataset.extent(0);
+  size_t num_cols = dataset.extent(1);
+
+  int num_ranks = 0;
+  cudaGetDeviceCount(&num_ranks);
```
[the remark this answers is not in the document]
jinsolp: (comment below)

cpp/src/neighbors/all_neighbors/all_neighbors_batched.cuh
```diff
@@ -0,0 +1,499 @@
...
+
+template <typename T, typename IdxT>
+void batch_build(const raft::resources& handle,
+                 raft::host_matrix_view<const T, IdxT, row_major> dataset,
+                 const index_params& batch_params,
+                 all_neighbors::index<IdxT, T>& index)
+{
+  int num_ranks = 0;
+  cudaGetDeviceCount(&num_ranks);
+  if (num_ranks > 1) {
```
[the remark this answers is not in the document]
divyegala: But a user can have multiple visible CUDA devices for many reasons. Just because they have it visible, we should not assume they want us to use them all. This should be opt-in behavior ideally, but otherwise at least a log to let them know we're using all GPUs.

cpp/include/cuvs/neighbors/all_neighbors.hpp
```diff
@@ -0,0 +1,312 @@
...
+ * @param[in] params an instance of all_neighbors::index_params that are parameters
+ *               to build all-neighbors knn graph
+ * @param[in] return_distances boolean for whether to return the distances matrix as part of the
+ * index
+ * @return index<IdxT> index containing all-neighbors knn graph in host memory (and the
+ * corresponding distances in device memory if return_distances = true)
+ */
+auto build(const raft::resources& handle,
+           raft::host_matrix_view<const float, int64_t, row_major> dataset,
+           int64_t k,
```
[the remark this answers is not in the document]
jinsolp: I'm not sure if I understood your comment properly. Are you asking if it should be an index param because you think these parameters are not related to the index and therefore the name seems wrong?

cpp/include/cuvs/neighbors/all_neighbors.hpp
```diff
@@ -0,0 +1,312 @@
...
+           const index_params& params,
+           bool return_distances = false) -> index<int64_t, float>;
+
+/**
+ * @brief Build a new all-neighbors index
```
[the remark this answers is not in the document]
cjnolet: @divyegala it's an approximate knn, not an approximate pairwise distance matrix. All-neighbors just means we're computing the nearest neighbors for all the training vectors. I do agree we could define this more clearly in the docs.

cpp/src/neighbors/all_neighbors/all_neighbors_batched.cuh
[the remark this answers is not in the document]
cjnolet: > I don't know if this is the intention but I'm not a fan of the fact that we do multi-GPU build without explicit request from the user. 

100% agreed w/ divye here. In fact, @viclafargue built the single-node multi-GPU indexes for most of the ANN algorithms fairly quickly and the most time was spent on the API experience. We need to be very intentional and careful that we design the user experience and APIs to be consistent, flexible, and easy to use. Ideally we would be defining, configuring, and otherwise specifying multi-gpu configuration information in the underlying `raft::resources` as much as possible so that we can separate the concerns here. This is where I think you should take a look at the `raft::resources` APIs more deeply, @jinsolp (in `raft/include/core/resource`)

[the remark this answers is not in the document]
jinsolp: Thanks @divyegala @cjnolet . Looking through, I realized that we had `raft::device_resources_snmg` (seems like a relatively new merge). Would we want to use this to handle resources without explicitly making calls like `cudaSetDevice`? 
But if we want the user to specify how many / which GPUs they want to use, should we have the device ids as some form of vector as part of the `index_params`?

cpp/include/cuvs/neighbors/batch_ann.hpp
jinsolp: (comment above)

cpp/src/neighbors/all_neighbors/all_neighbors_batched.cuh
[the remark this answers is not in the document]
cjnolet: > But if we want the user to specify how many / which GPUs they want to use, should we have the device ids as some form of vector as part of the index_params?

Absolutely not. That's my point- we want the resources/environment to be completely independent of the machine learning hyperparameters. The index_params should be the same no matter if single or multi-gpu. This is now making me realize that we need more than just `device_resources_snmg` which always assumes nccl. I think what we need is for the algorithm to tell us whether we need a nccl clique and for the device_resources to respond appropriately, lazily, at runtime, based on the needs of the algorithm. This is actually why we use the factory pattern that we use inside the various `raft::resource` instances. Ideally the `device_resources_snmg` should carry with it ONLY the gpus that the user wants to use (we can default to all available when unspecified) and for the algorithm to ask for a nccl_clique when it needs it, or not. Does that make sense? We should be able to implement this with a moderately small change tot he current `device_resources_snmg`.

cpp/src/neighbors/all_neighbors/all_neighbors.cuh
```diff
@@ -0,0 +1,127 @@
...
+#include <cuvs/neighbors/all_neighbors.hpp>
+#include <raft/util/cudart_utils.hpp>
+
+namespace cuvs::neighbors::all_neighbors::detail {
+using namespace cuvs::neighbors;
+
+void check_metric(const index_params& params)
+{
+  if (std::holds_alternative<graph_build_params::nn_descent_params>(params.graph_build_params)) {
+    auto allowed_metrics = params.metric == cuvs::distance::DistanceType::L2Expanded ||
```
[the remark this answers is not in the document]
jinsolp: Now depending on [this PR](https://github.com/rapidsai/cuvs/pull/859)

cpp/include/cuvs/neighbors/all_neighbors.hpp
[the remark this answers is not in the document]
divyegala: It is an index parameter in the sense that the size of the index built depends on the value of k, making it a hyperparameter. It's not the same function as that in a query API.

cpp/include/cuvs/neighbors/all_neighbors.hpp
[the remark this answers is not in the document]
divyegala: I couldn't think of a better term, but wanted it expressed in the docs that this API builds graphs of shape (nrow, nrow)

cpp/include/cuvs/neighbors/all_neighbors.hpp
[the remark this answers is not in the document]
jinsolp: Oh sorry I think I was looking at the wrong line. 
We have `k` inside the `index` struct. If we had to choose between having `k` inside `index` vs `index_params`, I think the former makes more sense..??

cpp/include/cuvs/neighbors/all_neighbors.hpp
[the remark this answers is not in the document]
jinsolp: This API builds graphs of shape (nrow x k) which has the top k nearest neighbors for each data. So it's a nearest-neighbor finder on the entire dataset (no separated build and search functions)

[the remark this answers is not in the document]
divyegala: Where does it get k from?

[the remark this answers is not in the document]
jinsolp: in this case the k is part of the index struct (`index<int64_t, float>& idx`).

So either pass the "index params AND k" -> create and return an index which holds nrow x k matrices

OR "index params AND index" -> return a filled index

cpp/include/cuvs/neighbors/all_neighbors.hpp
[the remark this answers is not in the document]
cjnolet: @jinsolp I think i see where the disconnect is here. Imagine that at some point (hopefully in the not too distant future), `cuvs::neighbors::index` becomes an abstract class which provides unified method definitions for `build()` and `search()`. Does this make more sense? What @divyegala is (rightfully) pointing out is that k is usually a search argument, not a build argument, so this implementation of the `index` class would not follow this unification, and that would actually be worse than unification + a little inconsistency in the meaning of k. 

I do think in the case of all-neighbors, `k` is a hyperparameter in the same way `intermediate_graph_degree` is a hyperparameter in `CAGRA`. In fact, since `all-neighbors` is, in fact, a knn-graph, I would go as far as to say that perhaps we should change the name `k` here to `graph_degree` and add it to the `index_params` struct. What do you think?

cpp/include/cuvs/neighbors/all_neighbors.hpp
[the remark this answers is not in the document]
divyegala: I think `k` should either be part of index params or index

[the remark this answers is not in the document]
cjnolet: This API is different than the other APIs in cuVS. `k` (which I prefer to call `graph_degree`) needs to be specified by the user. The user can't specify this in the index struct because these are factory methods that exist to abstract the building of the index struct from the user. I'm less concerned about specifying `graph_degree` in the params struct than I am about `return_distances`. However, for consistency, I do think either 
1. we should add `return_distances` to the params struct.
2. we should define an additional "options" struct that can hold options for the user (such as `return_distances` and `graph_degree`) which are not actually hyperparams.  

The challenge is that we'd need to update all of the APIs to accept this new options struct, whether they need it or not, so that they are consistent. We can easily do this w/ the factory functions by adding a new overload and deprecating the existing functions. I'm starting to lean on that option.

cpp/include/cuvs/neighbors/all_neighbors.hpp
```diff
@@ -0,0 +1,321 @@
...
+ * @param[in] return_distances boolean for whether to return the distances matrix as part of the
+ * index
+ * @return index<IdxT> index containing all-neighbors knn graph in host memory (and the
+ * corresponding distances in device memory if return_distances = true)
+ */
+auto build(const raft::resources& handle,
+           raft::device_matrix_view<const float, int64_t, row_major> dataset,
+           int64_t k,
+           const index_params& params,
+           bool return_distances = false) -> index<int64_t, float>;
```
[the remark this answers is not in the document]
jinsolp: Following up on this comment and everything discussed above, I have a few questions about`k`. @cjnolet @divyegala 
1. `index` naturally holds k because that's just the number of columns in the distances and graph matrix. So when user provides `index`, we don't need a specified `k`. But when the user only provides `index_params`, we need a `k` that allows us to make the `index` that needs to be returned. If we have `k` in `index_params`, we would have 2 `k`s - one in `index_params` and one that is in `index`. Is this okay as long as we check if they are the same?
2. I'm having trouble understanding why we would want to call `k` a `graph_degree`. Both make sense, but since the users that make a call to this API would be expecting a knn-graph output, wouldn't it be more intuitive to call it `k` instead of `graph_degree`?
3. Agreeing that `return_distances` won't make sense for most other cases and that this may be different from all other index types, maybe we shouldn't call it an index at all? But just call it a knn graph? And not even inherit from cuvs index struct?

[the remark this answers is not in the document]
cjnolet: 1. The `build` functions are factory functions. I'd honestly prefer it if the user didn't supply the index. The index should be a (mostly) opaque object from the standpoint of the user in the way that they should be specifying state in the factory function and then the factory function should be able to build a valid (and populated) instance of it. I don't know why or how build() functions that accept an already instantiated index object ended up making their way through our public APIs, but as we're tryingt o clean up instiations anyways, I'd propose we remove those (and not even include in the all-neighbors APIs). 

2. Because we're builkding a knn graph and a knn graph is a fixed-degree graph, with a degree which is == `graph_degree`. CAGRA uses the term graph-degree to specify this value and so I'd like to ikeep this nomenclature consistent across the library. Please note that we use `k` (search argument) to denote the number of neighbors at query time, and while functionally this is similar, fundamentally it's actually a tad different. 

3. Yes, that's kind of what I was thinking as well. This isn't an index- we create an index to search it but we're not intending to search this. Instead, the user should provide the output matrices (and distances should be std::optional instead of using a `bool return_distances`).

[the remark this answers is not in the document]
jinsolp: >  This isn't an index- we create an index to search it but we're not intending to search this. Instead, the user should provide the output matrices (and distances should be std::optional instead of using a bool return_distances

Yes I think this will make more sense! Let me fix it so that we get rid of the `index` altogether

> If we're not going to use a proper index for this- I think we can relax the constraint that we use the same nomenclature as CAGRA here- so it can still accept a k instead of graph-degree.

will keep the `k` : )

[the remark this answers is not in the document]
divyegala: +1 on everything @cjnolet said, it just flows more naturally that way and conveys the intended purpose very clearly when we don't call this an index.

cpp/src/neighbors/all_neighbors/all_neighbors_merge.cuh
```diff
@@ -0,0 +1,266 @@
+/*
```
jinsolp, line 1: this file is mostly a copy of functions in the existing `cuvs/cpp/src/neighbors/detail/nn_descent_batch.cuh` file ([link](https://github.com/rapidsai/cuvs/blob/branch-25.06/cpp/src/neighbors/detail/nn_descent_batch.cuh))

cpp/include/cuvs/neighbors/all_neighbors.hpp
```diff
@@ -0,0 +1,149 @@
...
+   * graph. E.g. With the similar device memory usages, (n_nearest_clusters / n_clusters) = 4/20
+   * will have better accuracy than 2/10 at the cost of performance.
+   */
+  size_t n_nearest_clusters           = 2;
+  size_t n_clusters                   = 1;  // defaults to not batching
+  cuvs::distance::DistanceType metric = cuvs::distance::DistanceType::L2Expanded;
+};
+
+/**
+ * @brief Builds an approximate all-neighbors knn graph.
```
cjnolet, line 93: Can we loosely define what exact is an all-neighbors graph? It's a common term in vector search, but not one I generally expect a new user would know.

cpp/bench/ann/src/cuvs/cuvs_mg_cagra_wrapper.h
```diff
@@ -18,7 +18,7 @@
 #include "cuvs_ann_bench_utils.h"
```
jinsolp, line 18: these files in bench will go away after merged of [this PR](https://github.com/rapidsai/cuvs/pull/869)

cpp/src/neighbors/mg/snmg.cuh
```diff
@@ -17,7 +17,8 @@
 #pragma once
 
 #include "../detail/knn_merge_parts.cuh"
-#include <raft/core/resource/nccl_clique.hpp>
+#include <raft/core/resource/multi_gpu.hpp>
```
jinsolp, line 20: this file will go away after merged of [this PR](https://github.com/rapidsai/cuvs/pull/869)

cpp/tests/neighbors/mg.cuh
```diff
@@ -22,7 +22,7 @@
 #include <cuvs/neighbors/cagra.hpp>
```
jinsolp, line 22: this file will go away after merged of [this PR](https://github.com/rapidsai/cuvs/pull/869)

cpp/include/cuvs/neighbors/all_neighbors.hpp
jinsolp: Done!

cpp/cmake/thirdparty/get_raft.cmake
```diff
@@ -45,8 +45,8 @@ function(find_and_configure_raft)
             COMPONENTS          ${RAFT_COMPONENTS}
             CPM_ARGS
               EXCLUDE_FROM_ALL TRUE
-              GIT_REPOSITORY        https://github.com/${PKG_FORK}/raft.git
-              GIT_TAG               ${PKG_PINNED_TAG}
+              GIT_REPOSITORY        https://github.com/jinsolp/raft.git
```
jinsolp, line 48: need to revert this

python/libcuvs/pyproject.toml
```diff
@@ -106,4 +106,4 @@ select = [
 ]
 
 # detect when package size grows significantly
-max_allowed_size_compressed = '1.0G'
+max_allowed_size_compressed = '1.3G'
```
gforsyth, line 109: This the spot for general configuration of the size limit, but because of `nccl` vendoring differences between cuda 11 and cuda 12, the limits are changed (depending on cuda version) in `ci/validate_wheel/sh` -- and please do not bump the limit by any more than is strictly necessary

jinsolp: Got it, thank you

cpp/include/cuvs/neighbors/all_neighbors.hpp
```diff
@@ -0,0 +1,152 @@
...
+   * params.graph_build_params = all_neighbors::graph_build_params::nn_descent_params{};
+   *
+   * @endcode
+   */
+  GraphBuildParams graph_build_params;
+
+  /**
+   * Usage of n_nearest_clusters and n_clusters
+   *
+   * Hint1: the ratio of n_nearest_clusters / n_clusters determines device memory usage.
```
cjnolet, line 79: Instead of `hint1` and `hint2`, can we just put this in a paragraph form? the `hint` make it sound like a video game.

cpp/tests/neighbors/all_neighbors.cuh
```diff
@@ -0,0 +1,237 @@
...
+#pragma once
+
+#include "../test_utils.cuh"
+#include "ann_utils.cuh"
+#include "naive_knn.cuh"
+#include <cstddef>
+#include <cuvs/distance/distance.hpp>
+#include <cuvs/neighbors/all_neighbors.hpp>
+#include <cuvs/neighbors/ivf_pq.hpp>
+#include <cuvs/neighbors/nn_descent.hpp>
```
cjnolet, line 25: Very nice! We should be reusing `hpp` (aka public) APIs as much as possible so that we aren't implicitly compiling any new device code here. It's great to see we're doing that.

cpp/include/cuvs/neighbors/all_neighbors.hpp
jinsolp: hahaha 😂 sure, will do

cpp/src/neighbors/all_neighbors/all_neighbors_merge.cuh
```diff
@@ -0,0 +1,266 @@
...
+    idxBase = threadIdx.x * ITEMS_PER_THREAD;
+    for (int i = 0; i < ITEMS_PER_THREAD; i++) {
+      size_t colId = idxBase + i;
+      if (colId < 2 * graph_degree) {
+        blockKeys[colId]   = threadKeyValuePair[i].key;
+        blockValues[colId] = threadKeyValuePair[i].value;
+      }
+    }
+
+    __syncthreads();
```
tarang-jain, line 113: Correct my understanding if I am wrong, but shouldn't this `__syncthreads` be after the `Sort` operation (before storing the sorted stuff in `blockKeys` and `blockValues`)? Also, for computing the `uniqueMask`, it seems each thread is handling the values written by itself, so we might be able to avoid this sync after the storing.

tarang-jain: Ok the `uniqueMask` makes sense. But the reason I ask about the `__syncthreads` after the `Sort` is that we do it that way in Vamana (the `Sort` is preceded and succeeded by `__syncthreads`). If you are sure that the `BlockMergeSort` synchronizes before returning, this is fine. I don't want to go too deep into the implementation details here, just making sure we don't introduce a bug.

jinsolp: Thanks for pointing out. After looking up a bit it looks like there is no _explicitly_ guarantee of a sync. I'll add syncthreads to the code, shouldn't hurt to be safe.

tarang-jain: Yes, I looked up some other examples too, which do a `__syncthreads` both before and after `BlockMergeSort` / `BlockRadixSort`. We should be consistent.

cpp/include/cuvs/neighbors/all_neighbors.hpp
```diff
@@ -0,0 +1,152 @@
...
+
+#include <variant>
+
+namespace cuvs::neighbors::all_neighbors {
+
+/**
+ * @brief Parameters used to build an all-neighbors knn graph (find nearest neighbors for all the
+ * training vectors)
+ */
+namespace graph_build_params {
```
cjnolet, line 30: Are these copied over from CAGRA? Are they much different than CAGRA today? Any opportunity to reuse here would be great (by maybe putting this in `common.h`.

cpp/src/neighbors/all_neighbors/all_neighbors_batched.cuh
```diff
@@ -0,0 +1,466 @@
...
+
+template <typename T, typename IdxT>
+void batch_build(
+  const raft::resources& handle,
+  const all_neighbors_params& params,
+  raft::host_matrix_view<const T, IdxT, row_major> dataset,
+  raft::device_matrix_view<IdxT, IdxT, row_major> indices,
+  std::optional<raft::device_matrix_view<T, IdxT, row_major>> distances = std::nullopt)
+{
+  if (raft::resource::is_multi_gpu(handle)) {
```
cjnolet, line 383: This looks great!

cpp/include/cuvs/neighbors/all_neighbors.hpp
```diff
@@ -0,0 +1,152 @@
...
+   *
+   * // 2. Choose NN Descent algorithm for kNN graph construction
+   * params.graph_build_params = all_neighbors::graph_build_params::nn_descent_params{};
+   *
+   * @endcode
+   */
+  GraphBuildParams graph_build_params;
+
+  /**
+   * Usage of n_nearest_clusters and n_clusters
```
cjnolet, line 77: We can do this in a follow-up, but let's document parameters individually instead of grouping them. The reason being that we want to make sure our docs are presenting descriptions next to each parameter, otherwise it gets really confusing to the user to see one parameter with multiple docs and another param that doesn't look like it has any docs.

PR #790, opened 2025-03-31
`L2SqrtExpanded` metric support for NN Descent

jinsolp:
Adding `L2SqrtExpanded` support for NN Descent.
This is for better usages with cuML UMAP (since UMAP's default metric is `L2SqrtExpanded`)

### DIFF
--- a/cpp/src/neighbors/detail/nn_descent.cuh
+++ b/cpp/src/neighbors/detail/nn_descent.cuh
@@ -476,6 +476,7 @@ RAFT_KERNEL preprocess_data_kernel(
   __syncthreads();
 
   if (metric == cuvs::distance::DistanceType::L2Expanded ||
+      metric == cuvs::distance::DistanceType::L2SqrtExpanded ||
       metric == cuvs::distance::DistanceType::CosineExpanded) {
     int lane_id = threadIdx.x % raft::warp_size();
     for (int step = 0; step < raft::ceildiv(dim, raft::warp_size()); step++) {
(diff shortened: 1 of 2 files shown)


cpp/src/neighbors/detail/nn_descent.cuh
```diff
@@ -926,10 +930,13 @@ __launch_bounds__(BLOCK_SIZE, 4)
         s_distances[i] = -s_distances[i];
       } else if (metric == cuvs::distance::DistanceType::CosineExpanded) {
         s_distances[i] = 1.0 - s_distances[i];
-      } else {
+      } else {  // L2Expanded and L2SqrtExpanded
         s_distances[i] = l2_norms[old_neighbors[i % SKEWED_MAX_NUM_BI_SAMPLES]] +
                          l2_norms[new_neighbors[i / SKEWED_MAX_NUM_BI_SAMPLES]] -
                          2.0 * s_distances[i];
+        if (metric == cuvs::distance::DistanceType::L2SqrtExpanded) {
+          s_distances[i] = sqrtf(s_distances[i]);
```
tfeher, line 938: I recommend to move the sqrt computation to a post processing step.

tfeher (changes requested):
Thanks @jinsolp for the PR! This would work, but I would recommend a change. The sqrt function is more costly to compute than the regular multiply-add operations that we otherwise do. Therefore, instead of doing it in every NN descent iteration, I recommend to apply sqrt in a post processing step. You could simply add final mapping op [here](https://github.com/rapidsai/cuvs/blob/61a714b00915e833b4f62c36e2152a8eea48e064/cpp/src/neighbors/detail/nn_descent.cuh#L1414)

```C++
raft::linalg::map(res, output_dist_view, raft::sqrt_op(), output_dist_view);
```

jinsolp:
Thank you @tfeher I totally agree. Changes reflected in the new commit : )

tfeher (approved):
Thanks @jinsolp for the updates, LGTM!

PR #803, opened 2025-04-03
Moving NN Descent class and struct declarations to `nn_descent_gnnd.hpp`

jinsolp:
## Description
Moves declaration of the `GNND` class and a few other classes and structs to `nn_descent_gnnd.hpp` file.
Also adds a helper function for getting the build configurations based on a few checks.

## Why this is needed
Actual work for NN Descent build is done in the `GNND::build()` function, which is not exposed as a public API.
This complicates usage of NN Descent, especially when we want to create the GNND separately, and call gnnd.build() directly.

## Use Cases
Specifically, this is needed for the batched all-neighbors knn graph building implementation. 
GNND allocates all necessary device and host matrices at initialization.
The NN Descent's `build` exposed as the public API makes a new GNND, then calls gnnd.build().
So originally it's like this
```
public api build() {
     gnnd = GNND{};   // allocates a bunch of device/host memory
     gnnd.build(data);    // where all the work is done
}
```
However, now that we have to build iteratively in the batching algorithm, calling the public build API means we make GNND for every new cluster, thereby allocating (and deallocating) a large chunk of memory for every cluster. This takes a lot of time.

In such cases, it is better to make a GNND once, and reuse it over multiple clusters to save time. For example;
```
gnnd = GNND{}    // this needs to be made outside the for loop
for cluster in clusters:
    cluster_data = get_cluster_data(data)
    gnnd.build(cluster_data)
   ```
This PR will allow including `nn_descent_gnnd.hpp` instead of the entire `nn_descent.cuh` file.

### CODE REVIEW

File: cpp/src/neighbors/detail/nn_descent_gnnd.hpp [hunk 1]
```diff
@@ -0,0 +1,220 @@
+#include <cuvs/distance/distance.hpp>
+#include <cuvs/neighbors/nn_descent.hpp>
+
+#include <raft/core/device_mdarray.hpp>
+#include <raft/core/host_mdarray.hpp>
+#include <raft/core/operators.hpp>
+#include <raft/core/pinned_mdarray.hpp>
+#include <raft/core/resource/cuda_stream.hpp>
+#include <raft/core/resources.hpp>
+#include <raft/util/pow2_utils.cuh>
```
tarang-jain, line 10: (nit pick) Including a .cuh header in a .hpp header might work, but I dont think it is ideal practice. It means that this script has to be compiled with nvcc. And if that is the case, it might make sense to make it a `.cuh` beforehand.

tarang-jain (commented):
Thanks! A few small suggestions from my side.

cpp/src/neighbors/detail/nn_descent_gnnd.hpp
jinsolp: You are right! and all we're doing with with that header is use it to round up to the next multiple of 32.
Added a small inline function `roundUp32` to do this instead.

tarang-jain: Is there a reason for the preference of `.hpp` over `.cuh`? Yes the roundUp is a very small function, but I believe we should be reusing what we have already in raft rather than redefining a function.

divyegala (approved):

cpp/src/neighbors/detail/nn_descent_gnnd.hpp
divyegala: I agree with @tarang-jain here. This should be a `.cuh` file. But whether nvcc compiles the file or gcc does does not depend on the header extension being `.hpp` or `.cuh`. Header extensions are effectively completely ignored by the compiler.

Compiler choice entirely depends on the source file extension `.cpp` or `.cu`. This header, as it defines host-device functions, needs to be included in a `.cu` file and it is so there are no errors.

tarang-jain: Ah right! Thanks for pointing this out @divyegala . Looks like ultimately the source file where this is being included is a `.cu`, so it works!

PR #805, opened 2025-04-03
[CI] Enable Java test in CI workflow

rhdong:
This PR adds changes for Java CI.

Some scripts modified here also appear in [PR #831](https://github.com/rapidsai/cuvs/pull/831). Once 831 is merged, I’ll rebase and make sure everything stays consistent.


rhdong:
> @rhdong could you please put this PR into draft until you're ready for reviews? That'd reduce the notifications reviewers are getting, and help them understand when it's time to come review.

ci/build_java.sh
```diff
@@ -0,0 +1,44 @@
...
+rapids-print-env
+
+rapids-logger "Check GPU usage"
+nvidia-smi
+
+EXITCODE=0
+trap "EXITCODE=1" ERR
+set +e
+
+rapids-logger "Run Java build and tests"
```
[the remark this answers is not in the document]
rhdong: It will run tests.

chatman:
@narangvivek10 @rhdong @cjnolet I've committed a fix to download jextract automatically if not already installed. Reason for doing this is that jextract doesn't have a .deb or apt package for Ubuntu, and hence it the download of jextract needs to be scripted anyway.

chatman:
@narangvivek10 The jextract process failed due to:

```jextract downloaded to /__w/cuvs/cuvs/java/jextract-22
c_api.h:19:10: error: 'cuda_runtime.h' file not found
fatal: Unexpected exception org.openjdk.jextract.clang.TypeLayoutError: Invalid. segment: org.openjdk.jextract.clang.Type@cc813a2e, fieldName: n_probes occurred
Jextract encountered issues (returned value 5)
Bindings generation did not complete normally (returned value 5)
Forcing this build process to abort
```

Any ideas where the cuda_runtime.h will be found?

chatman:
@rhdong We have attempted to find the CUDA_HOME diras:
```CUDA_HOME=$(which nvcc | cut -d/ -f-4)```

And then tried to add the $CUDA_HOME/include dir to the include paths. Any ideas if this was the problem and is there a better way?

**Comment by chatman** [comment]
Also, I see the following:

```2025-04-24T16:03:41.9412876Z Bindings generation did not complete normally (returned value 5)
2025-04-24T16:03:41.9414077Z Forcing this build process to abort
2025-04-24T16:03:41.9513171Z 
2025-04-24T16:03:41.9516434Z [32mRAPIDS logger[0m » [04/24/25 16:03:41]
2025-04-24T16:03:41.9517878Z [32m┌─────────────────────────────────────────────────────────────────────────────┐[0m
2025-04-24T16:03:41.9519658Z [32m|    Initial Java build & test failed. Retrying with 'mvn clean verify -X'    |[0m
2025-04-24T16:03:41.9521203Z [32m└─────────────────────────────────────────────────────────────────────────────┘[0m
2025-04-24T16:03:41.9522013Z 
```

I think this retrying is not necessary, and not correct either, since here the failure was in a step even before Maven is invoked (failure is in the `generate-bindings.sh` file). Due to this retry, the logs are polluted with a lot of symbol not found issues via Maven, and it masks the original problem that the Panama bindings were not properly generated.

chatman:
@rhdong I've made the following changes:

* Debug printing of the CUDA_HOME variable, and the contents of the $CUDA_HOME/include
* If there's no include inside CUDA_HOME, try CUDA_HOME to be /usr/local/cuda

https://github.com/rapidsai/cuvs/pull/831/files/570fa2a7a792b39cb70c4ff1232661481ba8ecaa..306229d29b0123bc7f6e72adca6e7d155047f528

I'm hoping it will make things work. Can you please merge that and retest here?

rhdong:
> @rhdong We have attempted to find the CUDA_HOME diras: `CUDA_HOME=$(which nvcc | cut -d/ -f-4)`
> 
> And then tried to add the $CUDA_HOME/include dir to the include paths. Any ideas if this was the problem and is there a better way?

Hi @chatman @narangvivek10 , The docker image is `rapidsai/ci-conda:latest`, the cuda includes will be installed when creating the conda env `test`, as my local experiment, the `cuda_runtime.h` is in `/opt/conda/envs/test/targets/x86_64-linux/include/cuda_runtime.h` , the test env name is `test`.  So I fixed it by the top commit, and the new error comes up:

rhdong:
> Also, I see the following:
> 
> ```
> 2025-04-24T16:03:41.9414077Z Forcing this build process to abort
> 2025-04-24T16:03:41.9513171Z 
> 2025-04-24T16:03:41.9516434Z �[32mRAPIDS logger�[0m » [04/24/25 16:03:41]
> 2025-04-24T16:03:41.9517878Z �[32m┌─────────────────────────────────────────────────────────────────────────────┐�[0m
> 2025-04-24T16:03:41.9519658Z �[32m|    Initial Java build & test failed. Retrying with 'mvn clean verify -X'    |�[0m
> 2025-04-24T16:03:41.9521203Z �[32m└─────────────────────────────────────────────────────────────────────────────┘�[0m
> 2025-04-24T16:03:41.9522013Z 
> ```
> 
> I think this retrying is not necessary, and not correct either, since here the failure was in a step even before Maven is invoked (failure is in the `generate-bindings.sh` file). Due to this retry, the logs are polluted with a lot of symbol not found issues via Maven, and it masks the original problem that the Panama bindings were not properly generated.

Yeah, I agree, the main goal of the retry is to debug the issue of HNSW(has been resolved). The retry only happens when test fails. We can remove it at last.

cjnolet:
Closes #845

ci/build_java.sh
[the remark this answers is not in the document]
raydouglass: Can you split out the tests? Or even use `-DskipTests=true`? It would be great not to require a GPU for these builds.

cjnolet (approved):
LGTM!

ci/build_java.sh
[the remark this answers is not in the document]
cjnolet: @rhdong the way we do all of our other CI builds is the build gets done on hardware that may not have a GPU (in this case there's no actual kernel code in the java API so this should be straightforward) and then another Github Action job runs the tests on a GPU node.

chatman:
The C++ tests failed here with a segfault.

https://github.com/rapidsai/cuvs/actions/runs/14720279064/job/41313393923?pr=805#step:9:1585
```
 CMake Error at run_gpu_test.cmake:35 (execute_process):
  execute_process failed command indexes:

    1: "Abnormal exit with child return code: Segmentation fault"
```

@cjnolet @rhdong Any ideas, please?

ci/test_java.sh
```diff
@@ -0,0 +1,58 @@
...
+rapids-logger "Check GPU usage"
+nvidia-smi
+
+EXITCODE=0
+trap "EXITCODE=1" ERR
+set +e
+
+rapids-logger "Run Java build and tests"
+
+bash ./build.sh java
```
jameslamb, line 41: Will you eventually want to publish packages for the Java bindings, instead of doing a full rebuild from source here on a GPU runner? I hope so, as distributing packages would make it easier for users to integrate this code into their projects.

If yes, could you please link to a tracking issue for that work in a comment here, to make it clear that this `build.sh` should be reverted (in favor of installing packages) in the future?

dependencies.yaml
```diff
@@ -563,6 +579,14 @@ dependencies:
...
         packages:
           - scikit-learn
+  java:
+    common:
+      - output_types: conda
+        packages:
+          - *cmake_ver
+          - maven
+          - openjdk=22.*
+          - boost
```
jameslamb, line 589:
```suggestion
          - libboost-devel
```

`boost` is a Python package: 

https://github.com/conda-forge/boost-feedstock/blob/dc1f2d89e060eba365d67be99761e78eb4392d07/recipe/meta.yaml#L206-L225

I'm guessing that's not what you want here. If you need `libboost.so` and its headers, use `libboost-devel` instead.

build.sh
```diff
@@ -18,7 +18,7 @@ ARGS=$*
 # scripts, and that this script resides in the repo dir!
 REPODIR=$(cd $(dirname $0); pwd)
 
-VALIDARGS="clean libcuvs python rust go java docs tests bench-ann examples --uninstall  -v -g -n --allgpuarch --no-mg --no-cpu --cpu-only --no-shared-libs --no-nvtx --show_depr_warn --incl-cache-stats --time -h"
+VALIDARGS="clean libcuvs python rust go java docs tests bench-ann examples --uninstall  -v -g -n --allgpuarch --no-mg --no-cpu --cpu-only --no-shared-libs --no-nvtx --show_depr_warn --incl-cache-stats --time -h --skip-java-tests"
```
jameslamb, line 21: From my perspective, this should be flipped... the default should be to not run tests, and you should include a flag like `--run-java-tests` or something to opt into it. I think it's surprising to have a `build.sh` script that runs tests by default.

build.sh
```diff
@@ -494,12 +495,20 @@ if (( ${NUMARGS} == 0 )) || hasArg go; then
...
+  fi
+done
+
 if (( ${NUMARGS} == 0 )) || hasArg java; then
     if ! hasArg libcuvs; then
         echo "Please add 'libcuvs' to this script's arguments (ex. './build.sh libcuvs java') if libcuvs libraries are not already built"
     fi
     cd ${REPODIR}/java
-    ./build.sh
+    ./build.sh $SKIP_JAVA_TESTS
```
jameslamb, line 511:
```suggestion
    SKIP_JAVA_TESTS="${SKIP_JAVA_TESTS}" ./build.sh
```

As currently written, this is not going to work. `java/build.sh` on this branch is using `$SKIP_JAVA_TESTS` and not looking for any arguments passed positionally.

jameslamb (changes requested):
Thanks for merging in #831. I left some initial comments for your consideration.

Please also implement the `update-version.sh` changes I asked for in https://github.com/rapidsai/cuvs/pull/831#discussion_r2069550546

chatman:
```
-- Build files have been written to: /__w/cuvs/cuvs/java/internal/build
[1/2] Building C object CMakeFiles/cuvs_java.dir/src/cuvs_java.c.o
[2/2] Linking C shared library libcuvs_java.so
Starting Panama FFM API bindings generation ...
/opt/conda/envs/include does not exist.
Couldn't find a suitable CUDA include directory.

RAPIDS logger » [05/01/25 15:32:22]```

Trying to find includes in `/opt/conda/envs/include` is surprising.
The relevant parts from the script are:

```
CUDA_HOME=$(which nvcc | cut -d/ -f-4)

..

jextract \
 --include-dir ${REPODIR}/java/internal/build/_deps/dlpack-src/include/ \
 --include-dir ${CUDA_HOME}/targets/x86_64-linux/include \
 --include-dir ${REPODIR}/cpp/include \
 --output "${REPODIR}/java/cuvs-java/src/main/java22/" \
 --target-package ${TARGET_PACKAGE} \
 --header-class-name PanamaFFMAPI \
 ${CURDIR}/headers.h
```

Out of the three `--include-dir` params, I can't understand why any of those would point to `/opt/conda/envs/include`.

I have a few ideas:

1. Can we revert the change from ` --include-dir ${CUDA_HOME}/targets/x86_64-linux/include` to ` --include-dir ${CUDA_HOME}/include` and see if that works? In my local system, both these works (even though the `cuda_runtime.h` is only available in the targets/x86_64-linux dir.

2. Can we add more debug printing to understand how these variables like CUDA_HOME, REPODIR etc. are resolving? Consequent to that, shall we see if CUDA_HOME should be computed differently (instead of trying to find where `nvcc` lies)?

**Comment by chatman** [comment]
[fix-cudainclude.txt](https://github.com/user-attachments/files/20014543/fix-cudainclude.txt)

Oh, I misunderstood where the error is coming from. It is erroring out much before jextract command.

Can we try the patch attached here?

File: build.sh (hunk 5 above)
Comment (rhdong) [reply build.sh:511]: Yh Good eyesight! I found it was removed by last update, and it will be added back.

**Comment by chatman** [comment]
@jameslamb do you have a suggestion, please, for how to find the CUDA's include directory in the java/panama-bindings/generate-bindings.sh script?

File: ci/test_java.sh (hunk 2 above)
Comment (rhdong) [reply ci/test_java.sh:41]: Thanks for the suggestion! Yes, we plan to separate Java build and test in CI. I’ve filed [#868](https://github.com/rapidsai/cuvs/issues/868) to track that.

**Comment by rhdong** [comment]
> [fix-cudainclude.txt](https://github.com/user-attachments/files/20014543/fix-cudainclude.txt)
> 
> Oh, I misunderstood where the error is coming from. It is erroring out much before jextract command.
> 
> Can we try the patch attached here?

Thanks, and the last status is [here](https://github.com/rapidsai/cuvs/pull/805/commits/a5818980f03a433b5c7419d8b5a2bb243016553a)

File: java/panama-bindings/generate-bindings.sh [hunk 6]
```diff
@@ -9,10 +9,10 @@ CUDA_HOME=$(which nvcc | cut -d/ -f-4)
 TARGET_PACKAGE="com.nvidia.cuvs.internal.panama"
 
 # Try to verify that a include directory exists inside CUDA_HOME
-if [ ! -d "$CUDA_HOME/include" ];
+if [ ! -d "$CUDA_HOME/targets/x86_64-linux/include" ];
 then
-  echo "$CUDA_HOME/include does not exist."
-  if [ -d "/usr/local/cuda/include" ];
+  echo "$CUDA_HOME/targets/x86_64-linux/include does not exist."
+  if [ -d "/usr/local/cuda/targets/x86_64-linux/include" ];
```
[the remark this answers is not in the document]
rhdong: Done

chatman:
I just tested it locally and it works!
Here's a patch for fixing the javadocs plugin issue. We can fix it in a subsequent PR as well, if you think that's better.
[javadocs-plugin-fix.txt](https://github.com/user-attachments/files/20026013/javadocs-plugin-fix.txt)

build.sh
rhdong: Done

build.sh
rhdong: Done

rhdong:
> I just tested it locally and it works! Here's a patch for fixing the javadocs plugin issue. We can fix it in a subsequent PR as well, if you think that's better. [javadocs-plugin-fix.txt](https://github.com/user-attachments/files/20026013/javadocs-plugin-fix.txt)
> 
> It fixes these warnings in the build for me:
> 
> ```
> [WARNING] 
> [WARNING] Some problems were encountered while building the effective model for com.nvidia.cuvs:cuvs-java:jar:25.06.0
> [WARNING] 'build.plugins.plugin.(groupId:artifactId)' must be unique but found duplicate declaration of plugin org.apache.maven.plugins:maven-javadoc-plugin @ line 257, column 17
> [WARNING] 
> [WARNING] It is highly recommended to fix these problems because they threaten the stability of your build.
> [WARNING] 
> [WARNING] For this reason, future Maven versions might no longer support building such malformed projects.
> [WARNING] 
> ```

@chatman thank you! Done the patch!

ci/release/update-version.sh
```diff
@@ -106,3 +106,6 @@ sed_runner "s/VERSION=\".*\"/VERSION=\"${NEXT_FULL_JAVA_TAG}\"/g" java/build.sh
 for FILE in java/*/pom.xml; do
   sed_runner "/<!--CUVS_JAVA#VERSION_UPDATE_MARKER_START-->.*<!--CUVS_JAVA#VERSION_UPDATE_MARKER_END-->/s//<!--CUVS_JAVA#VERSION_UPDATE_MARKER_START--><version>${NEXT_FULL_JAVA_TAG}<\/version><!--CUVS_JAVA#VERSION_UPDATE_MARKER_END-->/g" "${FILE}"
 done
+
+sed_runner "s| [0-9][0-9].[0-9][0-9]| ${NEXT_SHORT_TAG}|g" java/README.md
+sed_runner "s|-[0-9][0-9].[0-9][0-9].[0-9]|-${NEXT_FULL_JAVA_TAG}|g" java/examples/README.md
```
jameslamb, line 111:
```suggestion
sed_runner "s| ${CURRENT_SHORT_TAG} | ${NEXT_SHORT_TAG} |g" java/README.md
sed_runner "s|-${CURRENT_SHORT_TAG}\.[0-9]\.jar|-${NEXT_FULL_JAVA_TAG}\.jar|g" java/examples/README.md
sed_runner "s|\/${CURRENT_SHORT_TAG}\.[0-9]\/|/${NEXT_FULL_JAVA_TAG}/|g" java/examples/README.md
```

In the future, please test changes to these `update-version.sh` scripts in the way I recommended in https://github.com/rapidsai/cuvs/pull/831#discussion_r2069550546

```shell
./ci/release/update-version.sh '25.08.00'

# this should return 0 results (CHANGELOG can be ignored)
git grep -E '25\.4|25\.04|25\.6|25\.06' \
```

That shows that this is missing the RAPIDS version embedded in some URLs in in `java/examples/README.md`.

image

I'm also suggesting here re-using `CURRENT_SHORT_TAG` in the script, to reduce the risk of this script changing the versions of other things that you didn't intend to change.

ci/test_java.sh
```diff
@@ -0,0 +1,44 @@
...
+
+rapids-dependency-file-generator \
+  --output conda \
+  --file-key test_java \
+  --prepend-channel "${CPP_CHANNEL}" \
+  --matrix "cuda=${RAPIDS_CUDA_VERSION%.*};arch=$(arch)" | tee "${ENV_YAML_DIR}/env.yaml"
+
+rapids-mamba-retry env create --yes -f "${ENV_YAML_DIR}/env.yaml" -n test
+
+export CMAKE_GENERATOR=Ninja
```
jameslamb, line 23:
```suggestion
```

Is `CMake` used at all in the Java builds? My understanding was that only `maven` is being invoked and that is never invokes `CMake`.

If I'm right about that, please remove this.

java/build.sh
```diff
@@ -17,8 +17,16 @@ cmake --build ./internal/build
 # Generate Panama FFM API bindings and update (if any of them changed)
 ./panama-bindings/generate-bindings.sh
 
+RUN_JAVA_TESTS="-DskipTests"
+for arg in "$@"
+do
+  if [ "$arg" == "--run-java-tests" ]; then
+    RUN_JAVA_TESTS=""
+  fi
+done
```
jameslamb, line 26:
```suggestion
MAVEN_VERIFY_ARGS=()
if ! hasArg --run-java-tests; then
  MAVEN_VERIFY_ARGS=("-DskipTests")
fi
```

I found the naming of `RUN_JAVA_TESTS` a bit confusing. Would you consider this alternative naming, that makes it clearer that it holds a set of arguments to be passed to `mvn verify`?

To make this suggestion work, copy the `hasArg` function from `build.sh` and put it near the top of this script

```shell
function hasArg {
    (( ${NUMARGS} != 0 )) && (echo " ${ARGS} " | grep -q " $1 ")
}
```

Then apply the suggestion in this comment.

And change the `mvn verify` call to the following:

```shell
mvn verify "${MAVEN_VERIFY_ARGS[@]}"
```

build.sh
```diff
@@ -491,12 +492,20 @@ if (( ${NUMARGS} == 0 )) || hasArg go; then
...
+  fi
+done
+
 if (( ${NUMARGS} == 0 )) || hasArg java; then
     if ! hasArg libcuvs; then
         echo "Please add 'libcuvs' to this script's arguments (ex. './build.sh libcuvs java') if libcuvs libraries are not already built"
     fi
     cd ${REPODIR}/java
-    ./build.sh
+    ./build.sh $RUN_JAVA_TESTS
```
jameslamb, line 508:
```suggestion
    if hasArg --run-java-tests; then
        ./build.sh --run-java-tests
    else
        ./build.sh
     fi
```

I think this is simpler than introducing a new variable (`RUN_JAVA_TESTS`) and a ` for` loop over all arguments. Would you please consider it?

build.sh
```diff
@@ -491,12 +492,20 @@ if (( ${NUMARGS} == 0 )) || hasArg go; then
...
 
 # Build the cuvs Java bindings
+RUN_JAVA_TESTS=""
+for arg in "$@"
+do
+  if [ "$arg" == "--run-java-tests" ]; then
+    RUN_JAVA_TESTS="--run-java-tests"
+  fi
+done
+
```
jameslamb, line 502:
```suggestion
```

This corresponds to my suggestion a few lines below to reuse `hasArg` that's already defined in this script, instead of introducing a new variable and a `for` loop over all arguments.

dependencies.yaml
```diff
@@ -111,6 +119,14 @@ files:
...
       - depends_on_libraft
       - depends_on_nccl
+  java:
+    output: none
+    includes:
+      - cuda
+      - cuda_version
+      - rapids_build
+      - depends_on_libcuvs
+      - java
```
jameslamb, line 129:
```suggestion
      - cuda
      - cuda_version
      - depends_on_libcuvs
      - depends_on_libraft
      - depends_on_nccl
      - java
      - rapids_build
```

For my own understanding... why is it that the Go and Rust bindings need `libraft` and NCCL but the Java bindings do not? If the answer is "the Java bindings do need `libraft` and NCCL and must just be getting those transitively via `libcuvs`", then please add explicit dependencies here. That'll ensure we get the correct versions of those things.

Also, please keep these lists alphabetized. That makes them a bit easier for people to read, reducing the risk of mistakes.

dependencies.yaml
```diff
@@ -563,6 +579,14 @@ dependencies:
...
         packages:
           - scikit-learn
+  java:
+    common:
+      - output_types: conda
+        packages:
+          - *cmake_ver
+          - maven
+          - openjdk=22.*
+          - libboost-devel
```
jameslamb, line 589:
```suggestion
          - *cmake_ver
          - libboost-devel
          - maven
          - openjdk=22.*
```

Please keep these lists alphabetized.

java/panama-bindings/generate-bindings.sh
```diff
@@ -5,21 +5,15 @@ set -e -u -o pipefail
...
-    exit 1
-  fi
+if [ -n "${CONDA_PREFIX:-}" ] && [ -d "${CONDA_PREFIX}/targets/x86_64-linux/include" ]; then
+  CUDA_INCLUDE_DIR="${CONDA_PREFIX}/targets/x86_64-linux/include"
+elif [ -d "/usr/local/cuda/targets/x86_64-linux/include" ]; then
+  CUDA_INCLUDE_DIR="/usr/local/cuda/targets/x86_64-linux/include"
+else
+  echo "Couldn't find a suitable CUDA include directory."
+  exit 1
 fi
```
jameslamb, line 17:
```suggestion
TARGET_DIR="targets/x86_64-linux/include"
if [ -n "${CONDA_PREFIX:-}" ] && [ -d "${CONDA_PREFIX}/${TARGET_DIR}" ]; then
  CUDA_INCLUDE_DIR="${CONDA_PREFIX}/${TARGET_DIR}"
elif [ -d "/usr/local/cuda/${TARGET_DIR}" ]; then
  CUDA_INCLUDE_DIR="/usr/local/cuda/${TARGET_DIR}"
else
  echo "Couldn't find a suitable CUDA include directory."
  exit 1
fi
```

Would you consider centralizing the repeat part of the path (the way the CUDA headers are layed out) into a variable, as suggested here? That'd make this a bit easier to read, in my opinion.

It'd also reduce the effort required if/when we add `aarch64` support for the Java bindings.

jameslamb (changes requested):
I left some suggestions for your consideration. There are enough of them that I'm leaving a blocking review... please `@` me when this is ready for another review.

If you'd prefer that I just push these changes directly to the PR instead, please let me know and I'd be happy to.

Also (not blocking this PR), I think it would be very helpful to reduce the duplication of the Java version across these builds. The assumption of Java 22 is hard-coded in many places right now, and it'd be easy to accidentally miss mentions like this during updates:

https://github.com/rapidsai/cuvs/blob/4a4d86363fd700f487593e65dd9e9ceb321a9c12/java/cuvs-java/pom.xml#L61-L622

If you agree with me about that, that could be a followup PR.

dependencies.yaml
cjnolet: Java bindings do need libraft (and NCCL, becaue cuVS requires NCCL).

rhdong:
> I left some suggestions for your consideration. There are enough of them that I'm leaving a blocking review... please `@` me when this is ready for another review.
> 
> If you'd prefer that I just push these changes directly to the PR instead, please let me know and I'd be happy to.
> 
> Also (not blocking this PR), I think it would be very helpful to reduce the duplication of the Java version across these builds. The assumption of Java 22 is hard-coded in many places right now, and it'd be easy to accidentally miss mentions like this during updates:
> 
> https://github.com/rapidsai/cuvs/blob/4a4d86363fd700f487593e65dd9e9ceb321a9c12/java/cuvs-java/pom.xml#L61-L622
> 
> If you agree with me about that, that could be a followup PR.

Sure, thank you so much for your suggestions, which are super helpful!  For this time, I will submit by myself to avoid duplicate commits, and pls feel free to push directly in next round.

build.sh
```diff
@@ -53,6 +53,7 @@ HELP="$0 [<target> ...] [<flag> ...] [--cmake-args=\"<args>\"] [--cache-tool=<to
    --no-nvtx                   - disable nvtx (profiling markers), but allow enabling it in downstream projects
    --no-shared-libs            - build without shared libraries
    --show_depr_warn            - show cmake deprecation warnings
+   --skip-java-test            - skip java tests
```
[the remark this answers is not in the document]
rhdong: Oh, basic mistake

ci/test_java.sh
jameslamb: Oh good, glad this worked! Thanks for removing it in https://github.com/rapidsai/cuvs/pull/805/commits/88165ad82833f3dabdfb90c0d8ca59196d2f4c7a#diff-76d7916ab418397eef76faf7d4eb5f000cff2b417def935c97f0d2c727afb6af

In the future, it'd be helpful for me as a reviewer if you'd comment on these suggestions saying that you fixed them, and then click "resolve conversation" to collapse the thread.

rhdong: Fixed

build.sh
rhdong: fixed

dependencies.yaml
rhdong: fixed

java/panama-bindings/generate-bindings.sh
rhdong: fixed

build.sh
rhdong: fixed

rhdong:
> I left some suggestions for your consideration. There are enough of them that I'm leaving a blocking review... please `@` me when this is ready for another review.
> 
> If you'd prefer that I just push these changes directly to the PR instead, please let me know and I'd be happy to.
> 
> Also (not blocking this PR), I think it would be very helpful to reduce the duplication of the Java version across these builds. The assumption of Java 22 is hard-coded in many places right now, and it'd be easy to accidentally miss mentions like this during updates:
> 
> https://github.com/rapidsai/cuvs/blob/4a4d86363fd700f487593e65dd9e9ceb321a9c12/java/cuvs-java/pom.xml#L61-L622
> 
> If you agree with me about that, that could be a followup PR.

Hi @chatman , per James’s comment regarding the Java version, do you have any suggestions or guidance on what the best practice typically is in this case? Thanks!

dependencies.yaml
jameslamb: Great, thank you. I'm glad NCCL and libraft were added here in https://github.com/rapidsai/cuvs/pull/805/commits/88165ad82833f3dabdfb90c0d8ca59196d2f4c7a

ci/build_java.sh
```diff
@@ -0,0 +1,44 @@
...
+rapids-dependency-file-generator \
+  --output conda \
+  --file-key java \
+  --prepend-channel "${CPP_CHANNEL}" \
+  --matrix "cuda=${RAPIDS_CUDA_VERSION%.*};arch=$(arch)" | tee "${ENV_YAML_DIR}/env.yaml"
+
+rapids-mamba-retry env create --yes -f "${ENV_YAML_DIR}/env.yaml" -n java
+
+export CMAKE_GENERATOR=Ninja
+
```
jameslamb, line 24:
```suggestion
```

This can probably be removed here as well, as it was in `ci/test_java.sh` as part of https://github.com/rapidsai/cuvs/pull/805#discussion_r2075676194.

And I see another issue... the change to add `libraft` and NCCL was not also applied to the `test_java` group in `dependency-file-generator`: https://github.com/rapidsai/cuvs/pull/805#discussion_r2075728734.

These types of issues are direct result of the high amount of duplication here... I think we should consolidate things.

`test_java.sh` only differs from `build_java.sh` in 2 ways:

* passes `--run-java-tests` to `java/build.sh`
* uses the `test_java` dependency group for `rapids-dependency-file-generator` (which would be identical to the `java` group there if `libraft` and NCCL were correctly added)

I think we should do the following:

* remove `test_java` group in `dependencies.yaml`
* consolidate all of the duplicated stuff in `build_java.sh` and `test_java.sh`
  - *for example, by having `build_java.sh` accept a `--run-java-tests` argument*
  - *that would also ensure `build_java.sh` is getting tested on PRs (currently it'd only run on nightly builds)*

Could I push changes like that to this branch?

jameslamb (changes requested):
Thanks for all the changes! I have one more important comment for your consideration. Sorry for not mentioning it earlier.

ci/build_java.sh
jameslamb: I've pushed these changes in https://github.com/rapidsai/cuvs/pull/805/commits/13034226bcf53c3b14e47c2983278372e6a8d595 

Let's see how CI goes.

ci/release/update-version.sh
```diff
@@ -106,3 +106,7 @@ sed_runner "s/VERSION=\".*\"/VERSION=\"${NEXT_FULL_JAVA_TAG}\"/g" java/build.sh
 for FILE in java/*/pom.xml; do
   sed_runner "/<!--CUVS_JAVA#VERSION_UPDATE_MARKER_START-->.*<!--CUVS_JAVA#VERSION_UPDATE_MARKER_END-->/s//<!--CUVS_JAVA#VERSION_UPDATE_MARKER_START--><version>${NEXT_FULL_JAVA_TAG}<\/version><!--CUVS_JAVA#VERSION_UPDATE_MARKER_END-->/g" "${FILE}"
 done
+
+sed_runner "s| ${CURRENT_SHORT_TAG} | ${NEXT_SHORT_TAG} |g" java/README.md
+sed_runner "s|-${CURRENT_SHORT_TAG}\.[0-9]\.jar|-${NEXT_FULL_JAVA_TAG}\.jar|g" java/examples/README.md
+sed_runner "s|\/${CURRENT_SHORT_TAG}\.[0-9]\/|/${NEXT_FULL_JAVA_TAG}/|g" java/examples/README.md
```
raydouglass, line 112: Using the current version can be fragile. And `[0-9]` has varying support per platform.

Tested these and seems to work:
```suggestion
sed_runner "s| CuVS [[:digit:]]\{2\}\.[[:digit:]]\{2\} | CuVS ${NEXT_SHORT_TAG} |g" java/README.md
sed_runner "s|-[[:digit:]]\{2\}\.[[:digit:]]\{2\}\.[[:digit:]]\{1,2\}\.jar|-${NEXT_FULL_JAVA_TAG}\.jar|g" java/examples/README.md
sed_runner "s|/[[:digit:]]\{2\}\.[[:digit:]]\{2\}\.[[:digit:]]\{1,2\}/|/${NEXT_FULL_JAVA_TAG}/|g" java/examples/README.md
```

jameslamb (approved):
From my perspective, this is ready to merge! The stuff I mentioned about reducing hard-coding of the Java version could come in a later PR (or not at all, if you choose not to pursue it).

Just check the `conda-java-tests` logs to be sure the tests are running after the most recent fix, otherwise I am ok with merging any time.

ci/release/update-version.sh
rhdong: fixed

raydouglass (approved):

cjnolet (approved):

PR #807, opened 2025-04-04
use `DISTANCE_T` instead of float in search_multi_cta.cuh

jiangyinzuo:


achirkin (changes requested):
Thank you for suggesting the change!
Could you please provide more details on the purpose of this? I think in this state it won't compile because `intermediate_distances` is passed to an internal function `_cuann_find_topk` which only has the `float` overload for the corresponding argument.

PR #813, opened 2025-04-08
Add support for half in CAGRA+HNSW

lowener:
This PR add support for half dtype for HNSW in C++, C and python, as well as some tests with it. I had to modify a bit the HNSW patch in order to enable computation on half data types.
I also added the support of inner-product distance for int8/uint8 data type.


divyegala (commented):
Thanks for your improvement to the docs! Looks like you caught a lot of typos.

cpp/src/neighbors/detail/hnsw.hpp
```diff
@@ -68,11 +73,12 @@ struct index_impl : index<T> {
...
+    if constexpr (std::is_same_v<T, float> || std::is_same_v<T, half>) {
       if (metric == cuvs::distance::DistanceType::L2Expanded) {
-        space_ = std::make_unique<hnswlib::L2Space>(dim);
+        space_ = std::make_unique<hnswlib::L2Space<T, typename hnsw_dist_t<T>::type>>(dim);
       } else if (metric == cuvs::distance::DistanceType::InnerProduct) {
-        space_ = std::make_unique<hnswlib::InnerProductSpace>(dim);
+        space_ =
+          std::make_unique<hnswlib::InnerProductSpace<T, typename hnsw_dist_t<T>::type>>(dim);
       }
     } else if constexpr (std::is_same_v<T, std::int8_t> or std::is_same_v<T, std::uint8_t>) {
```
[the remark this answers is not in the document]
lowener: True, I added support for that in the PR

cpp/bench/ann/src/hnswlib/hnswlib_wrapper.h
```diff
@@ -211,12 +210,16 @@ void hnsw_lib<T>::load(const std::string& path_to_index)
...
-      space_ = std::make_shared<hnswlib::L2Space>(dim_);
+      space_ = std::make_shared<hnswlib::L2Space<T, float>>(dim_);
     }
   } else if constexpr (std::is_same_v<T, uint8_t>) {
-    space_ = std::make_shared<hnswlib::L2SpaceI<T>>(dim_);
+    if (metric_ == Metric::kInnerProduct) {
+      space_ = std::make_shared<hnswlib::InnerProductSpace<T, typename hnsw_dist_t<T>::type>>(dim_);
+    } else {
+      space_ = std::make_shared<hnswlib::L2SpaceI<T>>(dim_);
+    }
```
divyegala, line 222: Can this entire if-else logic for type and metric combo now be simplified? Also, if I understand it correctly the accumulator type for `int8` and `uint8` will be `int` after your template additions in the hnswlib patch? Do we also want to do accumulations for `half` in `float`?
```suggestion
    static_assert(std::is_same_v<T, float> || std::is_same_v<T, half> || std::is_same_v<T, int8> || std::is_same_v<T, uint8>, "Only float, uint8, and int8 are supported");
    if (metric_ == Metric::kInnerProduct) {
      space_ = std::make_shared<hnswlib::InnerProductSpace<T, typename hnsw_dist_t<T>::type>>(dim_);
    } else {
      space_ = std::make_shared<hnswlib::L2SpaceI<T, typename hnsw_dist_t<T>::type>>>(dim_);
    }
```

cpp/bench/ann/src/hnswlib/hnswlib_wrapper.h
```diff
@@ -139,12 +134,16 @@ void hnsw_lib<T>::build(const T* dataset, size_t nrow)
...
```
divyegala, line 146: I don't see the additions for `half` and `int8`

cpp/src/neighbors/detail/hnsw.hpp
```diff
@@ -68,15 +73,19 @@ struct index_impl : index<T> {
...
       if (metric == cuvs::distance::DistanceType::L2Expanded) {
         space_ = std::make_unique<hnswlib::L2SpaceI<T>>(dim);
+      } else if (metric == cuvs::distance::DistanceType::InnerProduct) {
+        space_ =
+          std::make_unique<hnswlib::InnerProductSpace<T, typename hnsw_dist_t<T>::type>>(dim);
```
divyegala, line 88: Same comments from above for simplification of logic

python/cuvs/cuvs/tests/test_hnsw.py
```diff
@@ -76,12 +76,12 @@ def run_hnsw_build_search_test(
     skl_dist, skl_idx = nn_skl.kneighbors(queries, return_distance=True)
 
     recall = calc_recall(out_idx, skl_idx)
-    assert recall > 0.95
+    assert recall >= 0.9
```
divyegala, line 79: What combinations of params does the recall drop for?

cpp/bench/ann/src/hnswlib/hnswlib_wrapper.h
lowener: I will simplify it a bit by factoring the call to InnerProductSpace.
I tested the accumulations for half in float but I saw some recall drop (from over 98% to 82-92% on some cases). It would not pass the C++ tests.

python/cuvs/cuvs/tests/test_hnsw.py
lowener: uint8 with inner-product has a lower recall than the others

divyegala: Let's not drop the recall for all param combinations. Please keep the constraint as high as possible and pass the expected recall as a param

cpp/bench/ann/src/hnswlib/hnswlib_wrapper.h
divyegala: So there's no recall drop if you accumulate half in half but there is if you accumulate half in float?

lowener: Yes. The recall drop is noticeable especially on the Inner Product distance. On the L2 distance there is no recall drop on the current tests (that go up to 2000 dim), but I added a test with 5000 dimension just to see and the recall drops to 62%, while it is still above 98% with accumulation in float.

python/cuvs/cuvs/tests/test_hnsw.py
```diff
@@ -76,26 +77,30 @@ def run_hnsw_build_search_test(
...
 @pytest.mark.parametrize("k", [10, 20])
-@pytest.mark.parametrize("ef", [30, 40])
+@pytest.mark.parametrize("ef", [50, 150])
 @pytest.mark.parametrize("num_threads", [2, 4])
 @pytest.mark.parametrize("metric", ["sqeuclidean", "inner_product"])
 @pytest.mark.parametrize("build_algo", ["ivf_pq", "nn_descent"])
 @pytest.mark.parametrize("hierarchy", ["none", "cpu", "gpu"])
 def test_hnsw(dtype, k, ef, num_threads, metric, build_algo, hierarchy):
     # Note that inner_product tests use normalized input which we cannot
     # represent in int8, therefore we test only sqeuclidean metric here.
```
divyegala, line 92: Can we remove this comment now?

cpp/src/neighbors/hnsw_c.cpp
```diff
@@ -148,9 +148,12 @@ extern "C" cuvsError_t cuvsHnswIndexDestroy(cuvsHnswIndex_t index_c_ptr)
...
     auto index = *index_c_ptr;
 
-    if (index.dtype.code == kDLFloat) {
+    if (index.dtype.code == kDLFloat && index.dtype.bits == 32) {
       auto index_ptr = reinterpret_cast<cuvs::neighbors::hnsw::index<float>*>(index.addr);
       delete index_ptr;
+    } else if (index.dtype.code == kDLFloat && index.dtype.bits == 16) {
+      auto index_ptr = reinterpret_cast<cuvs::neighbors::hnsw::index<half>*>(index.addr);
+      delete index_ptr;
     } else if (index.dtype.code == kDLInt) {
```
divyegala, line 157: Should this and the other conditionals below check for number of bits?

divyegala (approved):
Pre-approving pending 2 comments. Everything else LGTM!

python/cuvs/cuvs/tests/test_hnsw.py
lowener: Yes I removed that comment now :+1:

cpp/src/neighbors/hnsw_c.cpp
lowener: Yes I added a check to enforce it

PR #816, opened 2025-04-10
Vendor RAPIDS.cmake

bdice:
This works around a problem where GitHub is rejecting requests to the `githubusercontent.com` CDN. We vendor the contents of `RAPIDS.cmake` from `rapidsai/rapids-cmake` to avoid a network call.

xref: https://github.com/rapidsai/rapids-cmake/pull/809
xref: https://github.com/rapidsai/build-infra/issues/206

### DIFF
--- a/java/internal/CMakeLists.txt
+++ b/java/internal/CMakeLists.txt
@@ -14,7 +14,7 @@
 
 cmake_minimum_required(VERSION 3.30.4 FATAL_ERROR)
 
-include(rapids_config.cmake)
+include(cmake/rapids_config.cmake)
 include(rapids-cmake)
 include(rapids-cpm)
 include(rapids-export)
--- a/java/internal/CMakeLists.txt
+++ b/java/internal/CMakeLists.txt
@@ -14,7 +14,7 @@
 
 cmake_minimum_required(VERSION 3.30.4 FATAL_ERROR)
 
-include(cmake/rapids_config.cmake)
+include(../../cmake/rapids_config.cmake)
 include(rapids-cmake)
 include(rapids-cpm)
 include(rapids-export)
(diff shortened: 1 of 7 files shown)


java/internal/CMakeLists.txt
```diff
@@ -14,7 +14,7 @@
 
 cmake_minimum_required(VERSION 3.30.4 FATAL_ERROR)
 
-include(rapids_config.cmake)
+include(cmake/rapids_config.cmake)
```
pentschev, line 17: I'm not sure if the relative path is elsewhere, will this still work?

rust/cuvs-sys/CMakeLists.txt
```diff
@@ -14,7 +14,7 @@
 
 cmake_minimum_required(VERSION 3.30.4 FATAL_ERROR)
 
-include(rapids_config.cmake)
+include(cmake/rapids_config.cmake)
```
pentschev, line 17: Similarly here, no relative path, might be right, but I can't tell for sure.

java/internal/CMakeLists.txt
KyleFromNVIDIA: Indeed, we can see [here](https://github.com/rapidsai/cuvs/actions/runs/14385459687/job/40340887589?pr=816#step:10:814) and [here](https://github.com/rapidsai/cuvs/actions/runs/14385459687/job/40341313531?pr=816#step:10:1973) that it's failing.

KyleFromNVIDIA:
```suggestion
include(../../cmake/rapids_config.cmake)
```

rust/cuvs-sys/CMakeLists.txt
KyleFromNVIDIA:
```suggestion
include(../../cmake/rapids_config.cmake)
```

KyleFromNVIDIA (approved):

PR #818, opened 2025-04-12
[cuvs_bench] distinguish search label from build label in data_export.py

jiangyinzuo:
When `configs/algos/*.yaml` has string configurations, they will be converted into "label" column in exported CSV files.

But the exported CSV files only have "label" column for build configuration. search configuration is missing.

Before this patch, the search label is overriden by build label in exported CSV files.

After this patch, CSV files have "label" column as build label and "search_label" column as search label.


jiangyinzuo:
Hello @cjnolet  , I have fixed the python code style issue. Could you please trigger the CI again?

PR #820, opened 2025-04-12
[FEA] Mutual Reachability Linkage

tarang-jain:
- Expose functions for building the single linkage on mutual reachability distances
- Migrate mutual reachability connectivities functor from cuML


tarang-jain:
Notes:
These functions have been migrated from HDBSCAN runner in cuML and exposed as a new public API, that is not tested directly here. Instead the full usage is tested end-to-end in HDBSCAN inside cuML. To veify correctness, the cuML PR (https://github.com/rapidsai/cuml/pull/6560), that uses this, currently fetches cuVS from the branch of this cuVS PR and the CI status on that PR can be checked. Once this PR is merged, I will revert the CMake changes to find the correct cuvs version in https://github.com/rapidsai/cuml/pull/6560

cpp/include/cuvs/cluster/agglomerative.hpp
```diff
@@ -119,6 +120,42 @@ void single_linkage(
...
+  cuvs::distance::DistanceType metric,
+  raft::device_vector_view<value_t, value_idx> core_dists,
+  raft::device_vector_view<value_idx, value_idx> indptr,
+  raft::sparse::COO<value_t, value_idx> mutual_reachability_coo,
+  raft::device_vector_view<value_idx, value_idx> out_mst_src,
+  raft::device_vector_view<value_idx, value_idx> out_mst_dst,
+  raft::device_vector_view<value_t, value_idx> out_mst_weights,
+  raft::device_vector_view<value_idx, value_idx> out_children,
+  raft::device_vector_view<value_t, value_idx> out_deltas,
+  raft::device_vector_view<value_idx, value_idx> out_sizes);
```
jinsolp, line 158: It seems like we are only instantiating this function for `value_t=float` and `value_idx=int`. How about replacing the `value_t` and `value_idx` here with these actual types instead of templates so we don't have to dig down to check which types are supported?

cpp/src/cluster/detail/single_linkage.cuh
```diff
@@ -122,4 +122,164 @@ void single_linkage(raft::resources const& handle,
...
+ * @param[in] core_dists core distances (size m)
+ * @param[in] indptr CSR indices of mutual reachability knn graph (size m + 1)
+ * @param[out] mutual_reachability_coo (symmetrized) maximum reachability distance for the k nearest
+ *             neighbors
+ * @param[out] out_mst_src src vertex of MST edges (size m - 1)
+ * @param[out] out_mst_dst dst vertex of MST eges (size m - 1)
+ * @param[out] out_mst_weights weights of MST edges (size m - 1)
+ * @param[out] out_dst children of output
+ * @param[out] out_delta distances of output
+ * @param[out] out_size cluster sizes of output
```
jinsolp, line 239:
```suggestion
 * @param[out] out_children children of output
 * @param[out] out_deltas distances of output
 * @param[out] out_sizes cluster sizes of output
```

cpp/src/cluster/single_linkage.cuh
```diff
@@ -101,4 +101,60 @@ void single_linkage(raft::resources const& handle,
...
+ * @param[in] core_dists core distances (size m)
+ * @param[in] indptr CSR indices of mutual reachability knn graph (size m + 1)
+ * @param[out] mutual_reachability_coo (symmetrized) maximum reachability distance for the k nearest
+ *             neighbors
+ * @param[out] out_mst_src src vertex of MST edges (size m - 1)
+ * @param[out] out_mst_dst dst vertex of MST eges (size m - 1)
+ * @param[out] out_mst_weights weights of MST edges (size m - 1)
+ * @param[out] out_dst children of output
+ * @param[out] out_delta distances of output
+ * @param[out] out_size cluster sizes of output
```
jinsolp, line 125:
```suggestion
 * @param[out] out_children children of output
 * @param[out] out_deltas distances of output
 * @param[out] out_sizes cluster sizes of output
```

jinsolp (commented):
Thank you for this PR Tarang! LGTM, just a few small suggestions : )

cpp/src/cluster/single_linkage_float.cu
```diff
@@ -39,4 +39,34 @@ void single_linkage(raft::resources const& handle,
...
+  raft::device_vector_view<int, int> indptr,
+  raft::sparse::COO<float, int>& mutual_reachability_coo,
+  raft::device_vector_view<int, int> out_mst_src,
+  raft::device_vector_view<int, int> out_mst_dst,
+  raft::device_vector_view<float, int> out_mst_weights,
+  raft::device_matrix_view<int, int> out_children,
+  raft::device_vector_view<float, int> out_deltas,
+  raft::device_vector_view<int, int> out_sizes)
+{
+  build_mutual_reachability_linkage<float, int>(handle,
```
divyegala, line 57: Just call into `detail` from here? `cpp/src/cluster/single_linkage.cuh` is adding an extra function layer that doesn't seem needed to me.

jinsolp (approved):

divyegala (approved):

tarang-jain:
/ok to test https://github.com/rapidsai/cuvs/pull/820/commits/e0c2171dae0004acf9b3006b0b23561badb502e8

cpp/include/cuvs/cluster/agglomerative.hpp
```diff
@@ -119,6 +120,41 @@ void single_linkage(
...
+ * @param[out] out_delta distances of output
+ * @param[out] out_size cluster sizes of output
+ */
+void build_mutual_reachability_linkage(
+  raft::resources const& handle,
+  raft::device_matrix_view<const float, int, raft::row_major> X,
+  cuvs::distance::DistanceType metric,
+  raft::device_vector_view<float, int> core_dists,
+  raft::device_vector_view<int, int> indptr,
+  raft::sparse::COO<float, int>& mutual_reachability_coo,
```
cjnolet, line 151: For the public API, can we please use `raft::sparse::device_coo_matrix_view` for this? I'd prefer not to use deprecated RAFT APIs in cuVS public APIs

cpp/include/cuvs/cluster/agglomerative.hpp
```diff
@@ -119,6 +120,41 @@ void single_linkage(
...
+ * @param[out] mutual_reachability_coo (symmetrized) maximum reachability distance for the k nearest
+ *             neighbors
+ * @param[out] out_mst_src src vertex of MST edges (size m - 1)
+ * @param[out] out_mst_dst dst vertex of MST eges (size m - 1)
+ * @param[out] out_mst_weights weights of MST edges (size m - 1)
+ * @param[out] out_dst children of output
+ * @param[out] out_delta distances of output
+ * @param[out] out_size cluster sizes of output
+ */
+void build_mutual_reachability_linkage(
```
cjnolet, line 145: Please put this in the `cuvs::cluster::agglomerative::helpers` namespace to designate it as a public API that is more of a helper function.

cpp/include/cuvs/cluster/agglomerative.hpp
```diff
@@ -21,6 +21,7 @@
 
 #include <raft/core/device_mdspan.hpp>
 #include <raft/core/resources.hpp>
+#include <raft/sparse/coo.hpp>
```
cjnolet, line 24: The more I'm thinking about this- why is this even in the `agglomerative` APIs? If this is just the mutual reachaiblity piece, why isn't it in `cuvs/neighbors/reachability` [here](https://github.com/rapidsai/cuvs/blob/d2adc6fcd56d5d4111bb82dfbae07a8f18ba6f1e/cpp/include/cuvs/neighbors/reachability.hpp#L67)? And why can't it be consolidated with the existing mutual_reachaiblity API?

cjnolet: Actually, I think we might want to separate the connectivities from the linkage here- we have a 'mutual_reachability' API already. Can't we separate the mutual reachability (aka all-neighbors graph construction) from the linkage? We can still expose the linkage functions in cuVS, but putting them all together in cuvs seems too specific to hdbscan.

cpp/src/cluster/single_linkage_float.cu
```diff
@@ -39,4 +39,34 @@ void single_linkage(raft::resources const& handle,
       handle, X, dendrogram, labels, metric, n_clusters, c);
   }
 }
+
+void build_mutual_reachability_linkage(
```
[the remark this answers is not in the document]
tarang-jain: Yeah we are indeed only doing the linkage here, but the connectivities functor used to connect components (to build the mst) is exclusively for Mutual Reachability. I wanted to capture the term "mutual_reachability" in the name to avoid confusion.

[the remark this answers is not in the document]
cjnolet: We don't name things based on where they are used, we name based on what they are doing. If this isn't doing a mutual reachability then we don't include that in the name. Please name this accordingly or break apart as needed.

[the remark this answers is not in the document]
cjnolet: These are primitives that are used to construct a linkage (aka knn graph + mst) and dendrogram. The reason we have primitives is so they can be composed in different ways for different algorithms. The more we can break them apart the more reusable they become. I asked you to bring these piece into cuVS because knn graph, mst, and thus single linkage cluster (aka nearest neighbors clustering) are all neighborhood methods.

cpp/include/cuvs/cluster/agglomerative.hpp
```diff
@@ -119,6 +120,41 @@ void single_linkage(
...
+ */
+void build_mutual_reachability_linkage(
+  raft::resources const& handle,
+  raft::device_matrix_view<const float, int, raft::row_major> X,
+  cuvs::distance::DistanceType metric,
+  raft::device_vector_view<float, int> core_dists,
+  raft::device_vector_view<int, int> indptr,
+  raft::sparse::COO<float, int>& mutual_reachability_coo,
+  raft::device_vector_view<int, int> out_mst_src,
+  raft::device_vector_view<int, int> out_mst_dst,
```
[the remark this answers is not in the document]
tarang-jain: The mst is sorted by weights (so the src and dst on the edges wont be in any ascending order). On the other hand, the rows and columns will be sorted if we have a COO, right?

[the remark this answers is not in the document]
tarang-jain: Ok upon reading up a bit -- it looks like the COO representation does not have row and column index sorting as a hard requirement. I can definitely use a COO in this case.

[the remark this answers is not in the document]
cjnolet: A coo isn’t required to be sorted. It’s what we call an edge list. The new sparse matrix types are just convenience containers so we don’t have to pass around and manage 3 separate arrays everywhere.

cpp/include/cuvs/cluster/agglomerative.hpp
cjnolet: I think `raft/sparse/coo.hpp` is unused in this file, now, right? It should be removed.

cpp/include/cuvs/cluster/agglomerative.hpp
```diff
@@ -119,6 +121,34 @@ void single_linkage(
...
+void build_linkage(raft::resources const& handle,
+                   raft::device_matrix_view<const float, int, raft::row_major> X,
+                   cuvs::distance::DistanceType metric,
+                   raft::device_vector_view<int, int> graph_indptr,
+                   raft::device_coo_matrix_view<float, int, int, size_t> graph,
+                   raft::device_coo_matrix_view<float, int, int, int> out_mst,
+                   raft::device_matrix_view<int, int> dendrogram,
+                   raft::device_vector_view<float, int> out_distances,
+                   raft::device_vector_view<int, int> out_sizes,
+                   red_op reduction_op);
```
cjnolet, line 151: We can't compile the reduction op into the binary. That's the whole challenge here.  We can't expose templare functions through the public APIs in cuVS.

tarang-jain: Ok what I could do is have an enum - such as "LINKAGE_TYPE" or "DISTANCE_TYPE", where one option can be L2 and the other can be mutual reachability. If it is mutual reachability I can keep the reduction op inside cuvs and call that.

tarang-jain: Moved everything to the reachability header in the helpers namespace. Removed the template reduction op.

cpp/include/cuvs/neighbors/reachability.hpp
```diff
@@ -73,6 +74,35 @@ void mutual_reachability_graph(
   raft::sparse::COO<float, int>& out,
   cuvs::distance::DistanceType metric = cuvs::distance::DistanceType::L2SqrtExpanded,
   float alpha                         = 1.0);
+
+namespace helpers {
+/**
+ * Given a mutual reachability graph, connects graph components and build dendrogram.
+ * Returns mst edges sorted by weight and the linkage.
+ * @tparam value_idx
```
cjnolet, line 82: Don't need tparams anymore w/ cuVS public APIs

cpp/src/cluster/detail/single_linkage.cuh
```diff
@@ -16,51 +16,132 @@
...
 
 /**
- * Single-linkage clustering, capable of constructing a KNN graph to
- * scale the algorithm beyond the n^2 memory consumption of implementations
- * that use the fully-connected graph of pairwise distances by connecting
- * a knn graph when k is not large enough to connect it.
-
- * @tparam value_idx
+ * Constructs a linkage by computing the minimum spanning tree and dendrogram in the Mutual
+ * Reachability space. Returns mst edges sorted by weight and the dendrogram.
```
jinsolp, line 124: Is this also or mutual reachability?

cpp/src/cluster/single_linkage_float.cu
```diff
@@ -39,4 +39,55 @@ void single_linkage(raft::resources const& handle,
...
+  raft::device_vector_view<int, int> out_sizes,
+  std::optional<raft::device_vector_view<float, int>> core_dists)
+{
+  /**
+   * Construct MST sorted by weights
+   */
+  if (std::holds_alternative<
+        cuvs::cluster::agglomerative::helpers::linkage_graph_params::mutual_reachability_params>(
+        linkage_graph_params)) {
+    RAFT_EXPECTS(core_dists.has_value(), "core_dists not passed");
```
jinsolp:
```suggestion
    RAFT_EXPECTS(core_dists.has_value(), "core_dists is required to build linkage with mutual reachability distances");
```

cpp/src/cluster/detail/single_linkage.cuh
tarang-jain: Yes thats correct. All the other steps remain the same as pairwise. The only difference is that the distance between two points is their mutual reachability distance (a trick hdbscan uses to isolate high density islands of points from the sparse sea of noise)

tarang-jain:
/ok to test [3b0737d](https://github.com/rapidsai/cuvs/pull/820/commits/3b0737d948807933beaccf32edd8dd366ab641f8)

cpp/src/cluster/detail/single_linkage.cuh
```diff
@@ -16,51 +16,136 @@
...
+    handle,
+    X.data_handle(),
+    m,
+    n,
+    metric,
+    min_samples,
+    alpha,
+    mutual_reachability_indptr.data_handle(),
+    core_dists.data_handle(),
+    mutual_reachability_coo);
```
divyegala, line 80: Why do we need to call the detail API instead of the public one? https://github.com/rapidsai/cuvs/blob/58735dbb718f7f8628cdef71fa54ed236fc69dfb/cpp/include/cuvs/neighbors/reachability.hpp#L67

cpp/src/cluster/detail/single_linkage.cuh
```diff
@@ -16,51 +16,136 @@
...
+
+  // auto color = raft::make_device_vector<value_idx, value_idx>(handle, static_cast<value_idx>(m));
+  rmm::device_uvector<value_idx> color(m, raft::resource::get_cuda_stream(handle));
+  cuvs::sparse::neighbors::MutualReachabilityFixConnectivitiesRedOp<value_idx, value_t>
+    reduction_op(core_dists.data_handle(), m);
+
+  size_t nnz = m * min_samples;
+
+  detail::build_sorted_mst<value_idx, value_t>(handle,
```
divyegala, line 89: Is it possible to combine the steps here into a single function for reuse by the next function as well?

cpp/src/cluster/detail/single_linkage.cuh
tarang-jain: The publc API will be deprecated (see https://github.com/rapidsai/cuvs/issues/982). Using the detail one here will make it easy to deprecate. @jinsolp is working on bringing brute force and mutual reachability into the all neighbors wrapper. Once that is done, we'll replace this function call with the one from the all neighbors wrapper.

cpp/src/cluster/detail/single_linkage.cuh
tarang-jain: The `build_dendrogram_host` call is the exact same in both the functions. But the `build_sorted_mst` call uses different reduction ops. I could do things like create a template argument for the reduction op (similar to what we do in `cross_component_nn`), but in my opinion that will unnecessarily complicate it.

cpp/src/cluster/detail/single_linkage.cuh
tarang-jain: I've ensured that the template args are the same as what is being instantiated in the `mutual_reachability_graph` src file.

cpp/src/cluster/detail/single_linkage.cuh
divyegala: I disagree, pulling out the reusable bits into a single function reads cleaner and passing a different reduction op is inline with how we write code in other places.

cpp/src/cluster/detail/single_linkage.cuh
divyegala: Can you add a comment to indicate deprecation and future work? A link to the issue for this work will be even better.

tarang-jain: Done.

cpp/src/cluster/detail/single_linkage.cuh
tarang-jain: Marking this as resolved based on offline discussion.

divyegala (approved):

PR #822, opened 2025-04-15
Reduce device memory usage for CAGRA's graph optimization process

anaruse:
CAGRA takes the initial knn graph as input and optimizes it to create a search graph. Several types of processing are performed in the graph optimization, the most memory-intensive of which is the counting of 2-hop detours.

Currently, the counting of 2-hop detours is performed on the GPU to speed up processing, and this requires that the entire initial knn graph be placed in device memory. In general, the size of the initial knn graph is 2x the size of the search graph. In other words, in the current implementation, roughly half the device memory size is the upper limit of the search graph that can be created. As it is, creating search graphs for huge datasets requires a GPU with a large amount of device memory, which is not practical.

To address this issue, this PR adds a CPU implementation of 2-hop detour counting and uses this CPU implementation to count 2-hop detours when device memory is insufficient.

The CPU implementation supports thread parallelism and is optimized to reduce conditional branches and is sufficiently fast. Of course, it is slower than the GPU implementation, but it can count 2-hop detours in about 3 to 4 times the time of the GPU implementation. Since the time for counting 2-hop detours on GPU is approximately 10% of the total indexing time, the overall time will increase by 20-30% when using the CPU implementation, but this is well within the practical range.

### DIFF
--- a/cpp/src/neighbors/detail/cagra/cagra_build.cuh
+++ b/cpp/src/neighbors/detail/cagra/cagra_build.cuh
@@ -44,6 +44,8 @@
 #include <cstdio>
 #include <vector>
 
+#include <sys/mman.h>
+
 namespace cuvs::neighbors::cagra::detail {
 
 template <typename IdxT>
@@ -485,6 +487,19 @@ auto iterative_build_graph(
     }
   }
 
+  // Allocate memory for neighbors list using Transparent HugePage
+  constexpr size_t thp_size = 2 * 1024 * 1024;
+  size_t byte_size          = sizeof(IdxT) * final_graph_size * topk;
+  if (byte_size % thp_size) { byte_size += thp_size - (byte_size % thp_size); }
+  IdxT* neighbors_ptr =
+    (IdxT*)mmap(NULL, byte_size, PROT_READ | PROT_WRITE, MAP_PRIVATE | MAP_ANONYMOUS, -1, 0);
+  if (neighbors_ptr == MAP_FAILED) {
+    perror("mmap");
+    exit(-1);
+  }
+  if (madvise(neighbors_ptr, byte_size, MADV_HUGEPAGE) != 0) { perror("madvise"); }
+  memset(neighbors_ptr, 0, byte_size);
+
   auto curr_graph_size = initial_graph_size;
   while (true) {
     RAFT_LOG_DEBUG("# graph_size = %lu (%.3lf)",
@@ -516,7 +531,9 @@ auto iterative_build_graph(
 
     auto dev_query_view = raft::make_device_matrix_view<const T, int64_t>(
       dev_dataset.data_handle(), (int64_t)curr_query_size, dev_dataset.extent(1));
-    auto neighbors = raft::make_host_matrix<IdxT, int64_t>(curr_query_size, curr_topk);
+
+    auto neighbors_view =
+      raft::make_host_matrix_view<IdxT, int64_t>(neighbors_ptr, curr_query_size, curr_topk);
 
     // Search.
     // Since there are many queries, divide them into batches and search them.
@@ -543,7 +560,7 @@ auto iterative_build_graph(
                                      batch_dev_distances_view);
 
       auto batch_neighbors_view = raft::make_host_matrix_view<IdxT, int64_t>(
-        neighbors.data_handle() + batch.offset() * curr_topk, batch.size(), curr_topk);
+        neighbors_view.data_handle() + batch.offset() * curr_topk, batch.size(), curr_topk);
       raft::copy(batch_neighbors_view.data_handle(),
                  batch_dev_neighbors_view.data_handle(),
                  batch_neighbors_view.size(),
@@ -556,8 +573,13 @@ auto iterative_build_graph(
     cagra_graph     = raft::make_host_matrix<IdxT, int64_t>(0, 0);  // delete existing grahp
     cagra_graph     = raft::make_host_matrix<IdxT, int64_t>(curr_graph_size, curr_graph_degree);
     optimize<IdxT>(
-      res, neighbors.view(), cagra_graph.view(), flag_last ? params.guarantee_connectivity : 0);
+      res, neighbors_view, cagra_graph.view(), flag_last ? params.guarantee_connectivity : 0);
     if (flag_last) { break; }
+  }
+
+  if (munmap(neighbors_ptr, byte_size) != 0) {
+    perror("munmap");
+    exit(-1);
   }
 
   return cagra_graph;
--- a/cpp/src/neighbors/detail/cagra/graph_core.cuh
+++ b/cpp/src/neighbors/detail/cagra/graph_core.cuh
@@ -1157,26 +1157,6 @@ void optimize(
   {
     raft::common::nvtx::range<cuvs::common::nvtx::domain::cuvs> block_scope(
       "cagra::graph::optimize/prune");
-    //
-    // Prune kNN graph
-    //
-    auto d_detour_count = raft::make_device_mdarray<uint8_t>(
-      res, large_tmp_mr, raft::make_extents<int64_t>(graph_size, input_graph_degree));
-
-    RAFT_CUDA_TRY(cudaMemsetAsync(d_detour_count.data_handle(),
-                                  0xff,
-                                  graph_size * input_graph_degree * sizeof(uint8_t),
-                                  raft::resource::get_cuda_stream(res)));
-
-    auto d_num_no_detour_edges = raft::make_device_mdarray<uint32_t>(
-      res, large_tmp_mr, raft::make_extents<int64_t>(graph_size));
-    RAFT_CUDA_TRY(cudaMemsetAsync(d_num_no_detour_edges.data_handle(),
-                                  0x00,
-                                  graph_size * sizeof(uint32_t),
-                                  raft::resource::get_cuda_stream(res)));
-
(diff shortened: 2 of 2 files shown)


cpp/src/neighbors/detail/cagra/graph_core.cuh
```diff
@@ -1191,57 +1171,155 @@ void optimize(
...
+    } else {
+      // Count 2-hop detours on CPU
+      raft::common::nvtx::range<cuvs::common::nvtx::domain::cuvs> block_scope(
+        "cagra::graph::optimize/prune/2-hop-counting-by-CPU");
+
+#pragma omp parallel for reduction(+ : num_keep, num_full)
+      for (IdxT iA = 0; iA < graph_size; iA++) {
+        uint32_t num_edges_no_detour = 0;
+        // Create a list of nodes, iB_candidates, that can be reached in 2-hops from node A.
+        auto iB_candidates = raft::make_host_vector<IdxT, int64_t>((input_graph_degree - 1) *
```
tfeher, line 1283: Could you describe the layout of this vector? If I understand correctly, the elements are sorted so that index range `0..k^2-1` store the elements that can be accessed by considering only the first `k` neighbors of the nodes in the graph.

tfeher (approved):
Thanks Akira, it is great to have this feature! I have a few comments and suggestions below.

tfeher (changes requested):
Clicked on the wrong button, I meant to request changes.

cpp/src/neighbors/detail/cagra/graph_core.cuh
```diff
@@ -1191,57 +1171,155 @@ void optimize(
...
+                                                                   (input_graph_degree - 1));
+        for (uint64_t kAC = 0; kAC < input_graph_degree - 1; kAC++) {
+          IdxT iC = input_graph_ptr[kAC + (input_graph_degree * iA)];
```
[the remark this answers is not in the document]
anaruse: I will change in that way. Also, the variable `input_graph_degree` looks like it should be renamed `knn_graph_degree`, etc.

cpp/src/neighbors/detail/cagra/graph_core.cuh
anaruse: `iB_candaites` is a temporary array that holds the node IDs reached from node-A by 2-hop routes. If degree of the knn graph is D, the number of 2-hop routes is `D^2`, but the size of `iB_candidates` is `(D-1)^2`. Why? Because if the rank of an edge from node-A is R `[0:D-1]`, then the only 2-hop routes that qualify as 2-hop detours for this edge are those that use two edges, both of which have ranks less than R. Therefore, the maximum number of possible 2-hop detours is `(D-1)^2`, which is why the size of `iB_candidates` is `(D-1)^2`.

cpp/src/neighbors/detail/cagra/graph_core.cuh
```diff
@@ -1191,57 +1171,155 @@ void optimize(
...
-    raft::copy(
-      host_stats.data_handle(), dev_stats.data_handle(), 2, raft::resource::get_cuda_stream(res));
-    const auto num_keep = host_stats.data_handle()[0];
-    const auto num_full = host_stats.data_handle()[1];
+      raft::copy(
+        host_stats.data_handle(), dev_stats.data_handle(), 2, raft::resource::get_cuda_stream(res));
+      num_keep = host_stats.data_handle()[0];
+      num_full = host_stats.data_handle()[1];
+    } else {
+      // Count 2-hop detours on CPU
```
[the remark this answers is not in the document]
anaruse: I've made the CPU version of 2-hop detour counting to a separate function.

cpp/src/neighbors/detail/cagra/cagra_build.cuh
```diff
@@ -485,6 +487,19 @@ auto iterative_build_graph(
...
+  if (byte_size % thp_size) { byte_size += thp_size - (byte_size % thp_size); }
+  IdxT* neighbors_ptr =
+    (IdxT*)mmap(NULL, byte_size, PROT_READ | PROT_WRITE, MAP_PRIVATE | MAP_ANONYMOUS, -1, 0);
+  if (neighbors_ptr == MAP_FAILED) {
+    perror("mmap");
+    exit(-1);
+  }
+  if (madvise(neighbors_ptr, byte_size, MADV_HUGEPAGE) != 0) { perror("madvise"); }
+  memset(neighbors_ptr, 0, byte_size);
+
```
tfeher, line 502: Does this need to be part of this PR? I would prefer if we have RAII wrapper for `mmap` allocations (e.g. create an `mdarray` with a custom `mmap` based allocator). I have started a discussion in #877.

tfeher (changes requested):
Thanks @anaruse for the updates! I have a few more request below.

cpp/src/neighbors/detail/cagra/graph_core.cuh
```diff
@@ -1191,57 +1225,132 @@ void optimize(
...
-        input_graph_degree,
-        1024);
+    auto detour_count = raft::make_host_matrix<uint8_t, int64_t>(graph_size, knn_graph_degree);
+
+    //
+    // If the available device memory is insufficient, do not use the GPU to count
+    // the number of 2-hop detours, but use the CPU.
+    //
+    // bool use_gpu = true;
+    bool use_gpu = false;
```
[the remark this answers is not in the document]
anaruse: That is correct. I forgot I had set it to `false` for testing.

cpp/src/neighbors/detail/cagra/graph_core.cuh
```diff
@@ -1191,57 +1225,132 @@ void optimize(
...
-          d_input_graph.data_handle(),
-          graph_size,
-          input_graph_degree,
-          output_graph_degree,
-          batch_size,
-          i_batch,
-          d_detour_count.data_handle(),
-          d_num_no_detour_edges.data_handle(),
-          dev_stats.data_handle());
+    if (use_gpu) {
```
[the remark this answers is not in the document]
anaruse: I would prefer the latter method, adding a `use_gpu` argument to `optimize()`. I'll fix it that way for now.

cpp/src/neighbors/detail/cagra/cagra_build.cuh
anaruse: This is necessary to obtain high performance on CPU, because access to the graph (neighbor list) is a kind of random access when counting the number of 2-hop detours.

Thank you for starting a discussion on how to create `mdarray` with `mmap`. I think that would make the source code easier to read and maintain. How about we leave it for now and switch to that when it is implemented?

tfeher: I will try to create a solution for #877. If we can get that in for the current release, then we can update the current PR accordingly. If the new helper gets delayed, then we will go ahead with this PR in the current form.

tfeher: Sorry for the delay. I could not solve #877 in time for this release. Still we shall use a RAII wrapper for handling `mmap` allocations. This is needed to guarantee the memory is deallocated in case an error is thrown later during index building. I would adapt from one of the helper class from Artem created for the benchmarks:

```c++
/** RAII wrapper for a mmap/munmap. */
struct mmap_owner {
  /** Allocate a new memory (not backed by a file). */
  mmap_owner(size_t size) : size_{size} {
    int flags = MAP_ANONYMOUS | MAP_PRIVATE | MAP_HUGETLB | MAP_HUGE_2MB;
    ptr_ = mmap(nullptr, size, PROT_READ | PROT_WRITE, flags, -1, 0);
    if (ptr_ == MAP_FAILED) {
      // Retry without huge pages
      flags = MAP_ANONYMOUS | MAP_PRIVATE;
      ptr_ = mmap(nullptr, size, PROT_READ | PROT_WRITE, flags, -1, 0);
      if (ptr_ == MAP_FAILED) {
        ptr_ = nullptr;
        throw std::runtime_error("cuvs::mmap_owner error");
      }
    }
  }

  ~mmap_owner() noexcept {
    if (ptr_ != nullptr) {
      munmap(ptr_, size_);
    }
  }

  // No copies for owning struct
  mmap_owner(const mmap_owner &res) = delete;
  auto operator=(const mmap_owner &other) -> mmap_owner & = delete;
  // Moving is fine
  mmap_owner(mmap_owner &&other)
      : ptr_{std::exchange(other.ptr_, nullptr)}, size_{std::exchange(
                                                      other.size_, 0)} {}
  auto operator=(mmap_owner &&other) -> mmap_owner & {
    std::swap(this->ptr_, other.ptr_);
    std::swap(this->size_, other.size_);
    return *this;
  }

  [[nodiscard]] auto data() const -> void * { return ptr_; }
  [[nodiscard]] auto size() const -> size_t { return size_; }

private:
  void *ptr_;
  size_t size_;
};
```

anaruse: Thanks Tamas, I rewrote the process of allocating transparent huge-page memory with mmap to use the RAII wrapper.

tfeher (approved):
Thanks Akira for the updates, the PR looks good to me.

cpp/src/neighbors/detail/cagra/graph_core.cuh
tfeher: I think we can skip this note, the way the data is assigned to the array describes the layout.

PR #828, opened 2025-04-17
Ann-bench: fix unsafe lazy blobs

achirkin:
The ann-bench dataset uses lazy-loading blobs to move data between storage and host and device memory.
The data may be moved between memory spaces at the moment some properties/pointers are requested.

In the search throughput mode, this sometimes causes a problem: two concurrent benchmark threads access the same property and concurrently modify the state of the blobs, which leads to various segfaults.

The fix is to guard the critical sections with a mutex lock. There shouldn't be any impact on benchmark QPS results.
Only one method, `dataset->dim()` is accessed within the benchmark loop. To avoid locking the mutex in this method, this PR changes the way `dim()` is evaluated; it's cached in `dim_` variable while still maintaining the behavior of loading it either from the query set or the base set depending on what is available/accessed first.


tarang-jain (approved):

tarang-jain:
Thanks for the update. It is a smart idea that we can put locks on the dataset layer (`dataset.hpp`) and still ensure thread safety in helpers in the blob layer (`blob.hpp`). IIRC currently the blob header is not used anywhere directly apart from `dataset.hpp`, so I do not have any concerns with this change.

cpp/bench/ann/src/common/dataset.hpp
```diff
@@ -98,6 +102,7 @@ struct dataset {
   [[nodiscard]] auto distance() const -> std::string { return distance_; }
   [[nodiscard]] auto dim() const -> int
```
tfeher, line 103: This is [called in the benchmark loop](https://github.com/rapidsai/cuvs/blob/4e0fb09c84c3cd7f8fbd9981bfd78e7013610d10/cpp/bench/ann/src/common/benchmark.hpp#L326). Please adjust the benchmark loop to store this value in a variable, so that we do not need to lock inside the benchmark loop.

tfeher (changes requested):
Hi Artem, thank you for this fix! Overall it is fine, but there is one issue.

cpp/bench/ann/src/common/dataset.hpp
achirkin: Thanks for catching this! I updated the logic to cache `dim()` and load it atomically.

tfeher (commented):
Thanks Artem for the updates. This works, but do we need all the added complexity?

When I asked to cache `dim`, I was only thinking of to store it in a variable outside the benchmark loop (like it is done for [query_set](https://github.com/rapidsai/cuvs/blob/4e0fb09c84c3cd7f8fbd9981bfd78e7013610d10/cpp/bench/ann/src/common/benchmark.hpp#L290)).

achirkin:
I actually thought this is a little bit of a simplification :) two variables `base_set_accessed_`, `query_set_accessed_` are replaced with a single `dim_`, and a helper function in place of plain assignments clearly states what is it doing for the program logic.

cpp/bench/ann/src/common/dataset.hpp
```diff
@@ -56,8 +58,21 @@ struct dataset {
...
-  mutable bool query_set_accessed_ = false;
+  // Protects the lazy mutations of the blobs accessed by multiple threads
+  mutable std::mutex mutex_;
+  // The dim can be read either from the training set or from the query set.
+  // This cache variable is filled from either of the two sets loaded first.
+  mutable std::atomic<int> dim_ = -1;
+
+  // Before returning the passed result, cache the dim value from the passed blob.
+  template <typename R>
+  auto with_cached_dim_from(const blob<DataT>& blob, R&& r) const -> R
```
tfeher, line 69: In order to cache dim from blob, we do not need `r`. I would recommend hat this function only caches the dimension. The call site can return `r`.

tfeher (changes requested):
> I actually thought this is a little bit of a simplification

Ok, I accept your point above. Still I believe the helper function is doing more than what it should.

cpp/bench/ann/src/common/dataset.hpp
achirkin: The point of forwarding `r` here is to shorten the code at the call site. Consider what we have now:
```
return with_cached_dim_from(query_set_, query_set_.data());
```
against what would be without forwarding:
```
auto* r = query_set_.data());
cache_dim_using(query_set_);
return r;
```
(I compute `r` first here, so that if the exception arises due to a missing file or something else, the reason of failure is more easily traced)
The helper is currently used eight times, so the result is 8 lines against 24.

I don't insist though :) and surely ok to change this if it would be easier to read for others.

tfeher: Thanks Artem for explanation. It was not clear to me that we need to evaluate the expression beforehand. Still I when I look at the code
```c++
return with_cached_dim_from(query_set_, query_set_.data());
```
I have no idea what this is returning. In contrast, the more verbose option is clear. Therefore I strongly prefer without forwarding. (Sorry for splitting hairs.)

tfeher (approved):
Thanks Artem for the updates, LGTM!

PR #829, opened 2025-04-21
Add compilation option to enable "sve" optimization level for CPU faiss

LizYou:


LizYou:
Some out of sync changes;

PR #832, opened 2025-04-21
Reduce device memory usage for CAGRA's graph optimization process (reverse graph creation)

anaruse:
Since CAGRA improves search accuracy by merging forward and reverse graphs, reverse graph is created as part of the graph optimization process. Currently, GPU is used to create the reverse graph at high speed, but when creating a graph for huge dataset, the amount of device memory may be insufficient to create the graph.

In this PR, as a countermeasure for the lack of device memory, we will add an implementation in which reverse graphs are created on the CPU.


tfeher (changes requested):
Thanks Akira for the PR, looks good overall! Please find my comments below.

anaruse:
Thanks for the review Tamas, I think I've addressed everything you pointed out, could you double check?

cpp/src/neighbors/detail/cagra/graph_core.cuh
```diff
@@ -1413,7 +1636,15 @@ void optimize(
...
   }
 
-  auto rev_graph       = raft::make_host_matrix<IdxT, int64_t>(graph_size, output_graph_degree);
+  // Allocate memory for rev_graph using Transparent HugePage
+  constexpr size_t thp_size = 2 * 1024 * 1024;
+  size_t byte_size          = sizeof(IdxT) * graph_size * output_graph_degree;
+  if (byte_size % thp_size) { byte_size += thp_size - (byte_size % thp_size); }
+  mmap_owner rev_graph(byte_size);
+  IdxT* rev_graph_ptr = (IdxT*)rev_graph.data();
+  memset(rev_graph_ptr, 0, byte_size);
```
achirkin, line 1645: I think we can remove the memset here. We use the `MAP_ANONYMOUS`, which seems to zero-initialize the memory on all supported architectures https://stackoverflow.com/a/65084762 . I did a few benchmarks using this branch over the last week and found that the `memset` here can sometimes take excessive amounts of time when the graph is larger. Not sure why though. Perhaps, memory fragmentation?

cpp/src/neighbors/detail/cagra/graph_core.cuh
```diff
@@ -1424,60 +1655,89 @@ void optimize(
     //
     const double time_make_start = cur_time();
 
-    device_matrix_view_from_host<IdxT, int64_t> d_rev_graph(res, rev_graph.view());
-    RAFT_CUDA_TRY(cudaMemsetAsync(d_rev_graph.data_handle(),
-                                  0xff,
-                                  graph_size * output_graph_degree * sizeof(IdxT),
-                                  raft::resource::get_cuda_stream(res)));
+    bool _use_gpu = use_gpu;
+    if (_use_gpu) {
```
achirkin, line 1659: Also while benchmarking, I encountered a problem that we setup the large workspace resource to the managed memory. Because of that, the `device_matrix_view_from_host d_input_graph` above allocates successfully yet leads to severe oversubscribing of UVM memory if the GPU RAM is tight. This in turn leads to an extremely slow `kern_make_rev_graph` kernel run.
Do you have any suggestions how can we be smarter about the use_gpu logic to fallback to CPU in such cases? Or maybe even avoid copying the input graph to the device when use_gpu = false?

cpp/src/neighbors/detail/cagra/graph_core.cuh
```diff
@@ -488,6 +490,53 @@ void shift_array(T* array, uint64_t num)
     array[i] = array[i - 1];
   }
 }
+
+// RAII wrapper for allocating memory with Transparent HugePage
+struct mmap_owner {
```
tfeher, line 493: We have already one version here https://github.com/rapidsai/cuvs/blob/2ca4d3a9ff1972199a0463bf963cd66ef398cdfe/cpp/src/neighbors/detail/cagra/cagra_build.cuh#L443

To avoid duplication, could you move this to a separate header and include that here as well as in `cagra_build.cuh`?

tfeher (changes requested):
Sorry for the delay @anaruse. I just have one more request, otherwise the PR looks good from my side.

cpp/src/neighbors/detail/cagra/graph_core.cuh
```diff
@@ -1424,60 +1655,89 @@ void optimize(
...
+      try {
+        // throw std::bad_alloc(); // **** DEBUG: force to use CPU implementation ****
+        auto d_rev_graph =
+          raft::make_device_matrix<IdxT, int64_t>(res, graph_size, output_graph_degree);
+      } catch (std::bad_alloc& e) {
+        RAFT_LOG_DEBUG("Insufficient memory for making reverse graph on GPU");
+        _use_gpu = false;
+      } catch (raft::logic_error& e) {
+        RAFT_LOG_DEBUG("Insufficient memory for making reverse graph on GPU (logic error)");
+        _use_gpu = false;
```
achirkin, line 1669:
```suggestion
      size_t free  = 0;
      size_t total = 0;
      RAFT_CUDA_TRY_NO_THROW(cudaMemGetInfo(&free, &total));
      _use_gpu = free >= sizeof(IdxT) * graph_size * output_graph_degree  // d_rev_graph
                           + sizeof(uint32_t) * graph_size                // d_rev_graph_count
                           + sizeof(IdxT) * graph_size                    // d_dest_nodes
        ;
```

cpp/src/neighbors/detail/cagra/graph_core.cuh
achirkin: Another problem is that the allocation call in the try catch uses the default rmm resource, whereas below we may end up using other memory resource set up in cuvs.
I'd suggest to explicitly calculate the free device memory and the required memory and act upon that. The suggestion for the reverse graph is in a separate comment, while I'd also suggest to change the similar chunk of code in the pruning step to the following:
```C++
      size_t free  = 0;
      size_t total = 0;
      RAFT_CUDA_TRY_NO_THROW(cudaMemGetInfo(&free, &total));
      // We often use managed memory for these allocations, so we may oversubscribe a little.
      _use_gpu = free * 2 >= sizeof(IdxT) * graph_size * knn_graph_degree         // d_input_graph
                               + sizeof(uint8_t) * graph_size * knn_graph_degree  // d_detour_count
                               + sizeof(uint32_t) * graph_size  // d_num_no_detour_edges
        ;
```

cpp/src/neighbors/detail/cagra/graph_core.cuh
```diff
@@ -488,6 +490,53 @@ void shift_array(T* array, uint64_t num)
...
+    ptr_      = mmap(nullptr, size, PROT_READ | PROT_WRITE, flags, -1, 0);
+    if (ptr_ == MAP_FAILED) {
+      ptr_ = nullptr;
+      throw std::runtime_error("cuvs::mmap_owner error");
+    }
+    if (madvise(ptr_, size, MADV_HUGEPAGE) != 0) {
+      munmap(ptr_, size);
+      ptr_ = nullptr;
+      throw std::runtime_error("cuvs::mmap_owner error");
+    }
```
achirkin, line 507: I don't think the call to `madvise` will ever fail on us, but if it does - maybe we can just ignore that? As long as mmap succeeded, we have the memory (perhaps just a bit slower).

cpp/src/neighbors/detail/cagra/graph_core.cuh
```diff
@@ -1156,6 +1205,180 @@ void count_2hop_detours(raft::host_matrix_view<IdxT, int64_t, raft::row_major> k
...
+  const uint64_t graph_size       = knn_graph.extent(0);
+  const uint64_t knn_graph_degree = knn_graph.extent(1);
+  const uint64_t rev_graph_degree = rev_graph.extent(1);
+
+  raft::common::nvtx::range<cuvs::common::nvtx::domain::cuvs> block_scope(
+    "cagra::graph::optimize/reverse_CPU-2scan");
+  RAFT_LOG_DEBUG("# reverse_CPU-2scan");
+
+  auto loc = raft::make_host_matrix<uint8_t, int64_t>(graph_size, knn_graph_degree);
+  constexpr uint8_t loc_max = 255;
```
achirkin, line 1240: Does that mean we don't support graph degrees larger than 255?

achirkin (changes requested):
Here's my couple suggestions on the cpu/gpu decision logic based on benchmarking this branch on deep-100M dataset.

PR #833, opened 2025-04-21
Add NN-Descent return_distances functionality to python/C

benfred:


cpp/src/neighbors/nn_descent_c.cpp
```diff
@@ -71,6 +71,57 @@ void* _build(cuvsResources_t res,
...
+    auto dst       = cuvs::core::from_dlpack<output_mdspan_type>(graph);
+    auto src       = index_ptr->graph();
+    auto res_ptr   = reinterpret_cast<raft::resources*>(res);
+
+    RAFT_EXPECTS(src.extent(0) == dst.extent(0), "Output graph has incorrect number of rows");
+    RAFT_EXPECTS(src.extent(1) == dst.extent(1), "Output graph has incorrect number of cols");
+
+    cudaMemcpyAsync(dst.data_handle(),
+                    src.data_handle(),
+                    dst.extent(0) * dst.extent(1) * sizeof(float),
```
jinsolp, line 90: I think this should be the type of indices (`sizeof(uint32_t)`) instead of `sizeof(float)`?

jinsolp (commented):
Thank you for this PR Ben! One minor suggestion : )

cpp/src/neighbors/nn_descent_c.cpp
benfred: good catch! I fixed in the latest commit

jinsolp (approved):

PR #835, opened 2025-04-23
Add missing nccl_clique header

benfred:
After a recent change in raft (https://github.com/rapidsai/raft/pull/2549) cuvs builds have been failing with a `error: namespace "raft::comms" has no member "nccl_clique"` error message. This is because the raft/comms/nccl_clique.hpp include was removed from raft/resource/nccl_clique.hpp - and we were relying on that here.

Fix by adding the required header

### DIFF
--- a/cpp/src/neighbors/mg/mg.cuh
+++ b/cpp/src/neighbors/mg/mg.cuh
@@ -17,6 +17,7 @@
 #pragma once
 
 #include "../detail/knn_merge_parts.cuh"
+#include <raft/comms/nccl_clique.hpp>
 #include <raft/core/resource/nccl_clique.hpp>
 #include <raft/core/serialize.hpp>
 #include <raft/linalg/add.cuh>


divyegala (approved):

cpp/src/neighbors/mg/mg.cuh
```diff
@@ -17,6 +17,7 @@
 #pragma once
 
 #include "../detail/knn_merge_parts.cuh"
+#include <raft/comms/nccl_clique.hpp>
```
jakirkham, line 20: Seeing the following build failure [in the ARM CUDA 11.8 wheel CI job]( https://github.com/rapidsai/cuvs/actions/runs/14606972256/job/40977947859?pr=835#step:11:551 ):

```cpp
  In file included from /__w/cuvs/cuvs/cpp/src/neighbors/mg/mg_pq_half_int64_t.cu:26:
  /__w/cuvs/cuvs/cpp/src/neighbors/mg/mg.cuh:20:10: fatal error: raft/comms/nccl_clique.hpp: No such file or directory
     20 | #include <raft/comms/nccl_clique.hpp>
        |          ^~~~~~~~~~~~~~~~~~~~~~~~~~~~
```

Does this header get packaged in our wheels?

jakirkham (commented):
Thanks Ben! 🙏

Seeing a wheel CI failure. Commented with details below

cpp/src/neighbors/mg/mg.cuh
benfred: sorry - I misread the raft change, and that file has been removed entirely in https://github.com/rapidsai/raft/pull/2549 (I thought we just were relying on a transitive include inside of raft and that the include had been removed, not that the entire file had been removed).

Changing cuvs to match raft is a bigger change than this unfortunately =(

PR #838, opened 2025-04-23
IVF-PQ tests: fix segfault when accessing empty lists

achirkin:
Check if the `index.lists()[label]` is not an empty shared pointer before accessing its content in tests.

If the list size is zero, the list shared pointer can be `.reset()` during deserialization/resizing. Hence IVF-Flat and IVF-PQ can sometimes have empty shared pointers in their `.lists()` members. In case of IVF-PQ, we've had a piece of tests where this wasn't checked before accessing. As a result, we have been observing occasional CI failures with segfaults - when the IVF-PQ index was very unbalanced, i.e. had empty lists.

So far, the affected test is the only place I could identify where the access to the list pointer is unprotected (either via explicit check or by the invariant list_size > 0).


bdice (approved):

tfeher (approved):
LGTM.

PR #703, opened 2025-02-14
[WIP] Serialize Vamana index with SSD sector alignment per MSFT DiskANN format, generate quantized dataset for integration with DiskANN

jamxia155:
Added an optional input flag to `cuvs::neighbors::vamana::serialize` to dump an input cuvs Vamana index to file with SSD sector alignment. File format follows [MSFT DiskANN](https://github.com/microsoft/DiskANN/blob/main/src/disk_utils.cpp).

Using the sector-aligned option also writes out the quantized dataset computed using user-supplied PQ codebooks file and rotation matrix file.

PR #846, opened 2025-04-25
Serialize Vamana index with SSD sector alignment per MSFT DiskANN format, generate quantized dataset for integration with DiskANN

jamxia155:
(This supersedes [PR!703](https://github.com/rapidsai/cuvs/pull/703))

Added an optional input flag to `cuvs::neighbors::vamana::serialize` to dump an input cuvs Vamana index to file with SSD sector alignment. File format follows [MSFT DiskANN](https://github.com/microsoft/DiskANN/blob/main/src/disk_utils.cpp).

Using the sector-aligned option also writes out the quantized dataset computed using user-supplied PQ codebooks file and rotation matrix file.


cpp/src/neighbors/detail/vpq_dataset.cuh
```diff
@@ -408,6 +408,112 @@ auto vpq_convert_math_type(const raft::resources& res, vpq_dataset<OldMathT, Idx
...
+  const uint32_t lane_id = raft::Pow2<kSubWarpSize>::mod(threadIdx.x);
+  const LabelT vq_label  = vq_labels(row_ix);
+
+  // write label
+  auto* out_label_ptr = reinterpret_cast<LabelT*>(&out_codes(row_ix, 0));
+  if (lane_id == 0) { *out_label_ptr = vq_label; }
+
+  auto* out_codes_ptr = reinterpret_cast<uint8_t*>(out_label_ptr + 1);
+  cuvs::neighbors::ivf_pq::detail::bitfield_view_t<PqBits> code_view{out_codes_ptr};
+  for (uint32_t j = 0; j < pq_dim; j++) {
```
tarang-jain, line 440: Is there a faster way to do this? We should be able to parallelize this loop, right? Looks like a single thread is working on one vector. Can we increase that?

tarang-jain (changes requested):

cpp/src/neighbors/detail/vamana/vamana_build.cuh
```diff
@@ -408,9 +534,119 @@ index<T, IdxT> build(
...
+
+    pq_params.pq_bits = std::lround(std::log2(pq_codebook_size));
+    pq_params.pq_dim  = pq_dim;
+
+    // transform PQEncodingTable (dimensions: pq_codebook_size x dim_per_subspace * pq_dim ) to
+    // pq_codebook (dimensions: pq_codebook_size * pq_dim, dim_per_subspace)
+    auto PQEncodingTable_device_vec = raft::make_device_vector<float, uint32_t>(
+      res, PQEncodingTable.size());  // logically a 2D matrix with dimensions pq_codebook_size x
+                                     // dim_per_subspace * pq_dim
+    RAFT_CUDA_TRY(cudaMemcpyAsync(PQEncodingTable_device_vec.data_handle(),
```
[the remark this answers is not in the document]
jamxia155: Updated. Thank you.

cpp/src/neighbors/detail/vamana/vamana_build.cuh
```diff
@@ -408,9 +534,119 @@ index<T, IdxT> build(
...
+        int codebook_id    = row_idx % pq_codebook_size;  // idx_pq_codebook_size
+        int id_in_subspace = i % dim_per_subspace;        // idx_dim_per_subspace
+
+        return PQEncodingTable_device_vec_view[codebook_id * pq_dim * dim_per_subspace +
+                                               subspace_id * dim_per_subspace + id_in_subspace];
+      });
+
+    // prepare rotation matrix OPQMatrix
+    auto OPQMatrix_device = raft::make_device_matrix<float, int64_t>(res, dim, dim);
+    RAFT_CUDA_TRY(cudaMemcpyAsync(OPQMatrix_device.data_handle(),
```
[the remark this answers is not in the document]
jamxia155: Updated.

cpp/src/neighbors/detail/vamana/vamana_serialize.cuh
```diff
@@ -34,8 +34,251 @@
 #include <fstream>
 #include <type_traits>
 
+#define DIV_ROUND_UP(X, Y) (((uint64_t)(X) + (Y)-1) / (Y))
+#define ROUND_UP(X, Y)     (((uint64_t)(X) + (Y)-1) / (Y) * (Y))
```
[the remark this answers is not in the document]
jamxia155: These are perfect. Thanks for the tip.

cpp/include/cuvs/neighbors/vamana.hpp
```diff
@@ -127,6 +131,13 @@ struct index : cuvs::neighbors::index {
     return *dataset_;
   }
 
+  /** Quantized dataset [size, codes_rowlen] */
+  [[nodiscard]] inline auto quantized_data() const -> const cuvs::neighbors::dataset<int64_t>&
+  {
```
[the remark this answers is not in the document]
jamxia155: I wanted to model this function after `vamana::index::data()` so that we can reuse the code in the serializer. Is there a reason we shouldn't return `dataset` from `quantized_data()` while it's ok to do so from `data()`?

cpp/src/neighbors/detail/vamana/vamana_build.cuh
```diff
@@ -376,6 +380,128 @@ void batched_insert_vamana(
...
+template <typename T>
+auto quantize_all_vectors(raft::resources const& res,
+                          raft::device_matrix_view<const T, int64_t> residuals,
+                          raft::device_matrix_view<float, uint32_t, raft::row_major> pq_codebook,
+                          cuvs::neighbors::vpq_params ps)
+  -> raft::device_matrix<uint8_t, int64_t, raft::row_major>
+{
+  auto dim         = residuals.extent(1);
+  auto vq_codebook = raft::make_device_matrix<float, uint32_t, raft::row_major>(res, 1, dim);
+  raft::linalg::map_offset(res, vq_codebook.view(), [] __device__(size_t i) { return 0; });
```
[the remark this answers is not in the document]
jamxia155: I removed this line and ran compute-sanitizer with initcheck. Looks like you're right that it's already initialized.

cpp/src/neighbors/detail/vamana/vamana_build.cuh
```diff
@@ -408,9 +534,119 @@ index<T, IdxT> build(
...
+    // Every row is (VEC_DIM/PQ_DIM) floats representing a group of cluster centroids.
+    // Every consecutive [PQ_DIM] rows is a set.
+    int32_t pq_codebook_size, pq_dim;
+
+    std::vector<float> PQEncodingTable = parse_pq_pivots_file(
+      params.codebook_prefix + "_pq_pivots.bin", dim, pq_codebook_size, pq_dim);
+    std::vector<float> OPQMatrix = parse_rotation_matrix_file(
+      params.codebook_prefix + "_pq_pivots.bin_rotation_matrix.bin", dim);
+
+    pq_params.pq_bits = std::lround(std::log2(pq_codebook_size));
```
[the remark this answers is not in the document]
jamxia155: Updated to `raft::log2` but could not find a float-to-integer rounding function in raft.

cpp/src/neighbors/detail/vamana/vamana_build.cuh
```diff
@@ -408,9 +534,119 @@ index<T, IdxT> build(
...
+    const uint32_t max_batch_size              = std::min(n_rows, kReasonableMaxBatchSize);
+    for (const auto& batch : cuvs::spatial::knn::detail::utils::batch_load_iterator<T>(
+           dataset.data_handle(),
+           n_rows,
+           dim,
+           max_batch_size,
+           raft::resource::get_cuda_stream(res),
+           rmm::mr::get_current_device_resource())) {
+      // convert dataset to float
+      auto dataset_float = raft::make_device_matrix<float, int64_t>(res, batch.size(), dim);
```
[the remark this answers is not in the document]
jamxia155: I gave it a try, but it turned out that the input parameter `dataset` has data type of `const T` while `raft::linalg::gemm` requires all input and output matrices to have the same template parameters so we have to make a copy even if we are only ever using float data.

cpp/src/neighbors/detail/vpq_dataset.cuh
jamxia155: Optimization is currently planned for after we release the functionality. Thanks for raising this point though! I'll make a note of this for later.

cpp/include/cuvs/neighbors/common.hpp
```diff
@@ -206,10 +206,14 @@ inline constexpr bool is_strided_dataset_v = is_strided_dataset<DatasetT>::value
...
  * @param[in] required_stride the leading dimension (in elements)
+ * @param[in] force_ownership force an owning_dataset to be returned (default: false)
  * @return maybe owning current-device-accessible strided matrix
  */
 template <typename SrcT>
-auto make_strided_dataset(const raft::resources& res, const SrcT& src, uint32_t required_stride)
+auto make_strided_dataset(const raft::resources& res,
+                          const SrcT& src,
+                          uint32_t required_stride,
+                          bool force_ownership = false)
```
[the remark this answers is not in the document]
jamxia155: As you might've suspected, setting this flag is only needed in specific situations like this: during `vamana::index::index()`, the function optionally computes a matrix for the quantized vectors using additional input files pointed to by `index_params::codebook_prefix`. Since this matrix is created within `index()`, its lifetime is limited to the index build (as opposed to the full dataset matrix which is provided to `index()` as an input) so we need to force ownership of the matrix. It may be possible to improve this by somehow moving the data instead of copying but that is not the focus for now.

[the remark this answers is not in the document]
cjnolet: I agree with @tarang-jain here. I think there's a cleaner way to do this. Please don't treat public APIs like this as a "just need to get this done". These APIs tend to stick around for a long time and we need to make sure they are as clean and flexible as they can be. In fact, we often spend more time discussing our public APIs than we do the impl details because once these are exposed they are very hard to change.

cpp/src/neighbors/detail/vamana/vamana_build.cuh
```diff
@@ -408,9 +533,117 @@ index<T, IdxT> build(
...
+      res, PQEncodingTable.size());  // logically a 2D matrix with dimensions pq_codebook_size x
+                                     // dim_per_subspace * pq_dim
+    raft::copy(PQEncodingTable_device_vec.data_handle(),
+               PQEncodingTable.data(),
+               PQEncodingTable.size(),
+               raft::resource::get_cuda_stream(res));
+    int dim_per_subspace = dim / pq_dim;
+    auto pq_codebook =
+      raft::make_device_matrix<float, uint32_t>(res, pq_codebook_size * pq_dim, dim_per_subspace);
+    raft::resource::sync_stream(res);
```
tarang-jain, line 564: Can we avoid this sync_stream? It is a H2D copy, so the stream will govern the ordering of device ops.

cpp/src/neighbors/detail/vamana/vamana_build.cuh
```diff
@@ -408,9 +533,117 @@ index<T, IdxT> build(
...
+    // TODO: with scaling workspace we could choose the batch size dynamically
+    constexpr uint32_t kReasonableMaxBatchSize = 65536;
+    const uint32_t max_batch_size              = std::min(n_rows, kReasonableMaxBatchSize);
+    for (const auto& batch : cuvs::spatial::knn::detail::utils::batch_load_iterator<T>(
+           dataset.data_handle(),
+           n_rows,
+           dim,
+           max_batch_size,
+           raft::resource::get_cuda_stream(res),
+           rmm::mr::get_current_device_resource())) {
```
tarang-jain, line 606: the `current_device_resource` would be the same as `res`, isn't it? In other places we use the workspace resource here.

cpp/include/cuvs/neighbors/vamana.hpp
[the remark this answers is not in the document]
tarang-jain: The return type of `vamana::index::data()` might have been an oversight. Here's what I think. `quantized_data()` is an important public API on the index. As a new cuvs user, one should not have to familiarize themselves with obscure data types such as `dataset`. My understanding is that those are for internal use-cases or to be used by other rapids repos. mdspan return types are a lot more user-friendly / readable on the other hand.

cpp/src/neighbors/detail/vamana/vamana_build.cuh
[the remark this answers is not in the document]
tarang-jain: You should be able to const_cast and convert to a non const mdspan right? I dont see any risk with that because even the non  const ptr won't be modified.

cpp/src/neighbors/detail/vpq_dataset.cuh
tarang-jain: Ok I see that `kSubWarpSize` threads are working on a single vector. This kernel is just supplying the subspace's codebook instead of a fixed codebook for all subspaces. At some point, we should be able to combine this with `process_and_fill_codes_kernel` with an option to travel to the indexed codebook if need be, rather than have a duplicate kernel. But that can be something to think about later. Thanks, this kernel LGTM!

cpp/include/cuvs/neighbors/vamana.hpp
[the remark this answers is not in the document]
bkarsin: While I agree that it would be nice to let users avoid the `dataset` type, the `data()` API call by other algorithms also returns a `dataset` object, not a matrix. So, if we want to change how these API calls work, we would need to be consistent across CAGRA and other algorithms as well.

cpp/include/cuvs/neighbors/common.hpp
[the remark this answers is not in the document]
bkarsin: We have talked a bit about how to handle this, and without the ability to force ownership here, we would need to have the `quantized_data` device matrix as a member of create/destroy it during the `vamana::index` constructor/destructor. Does this seem like a reasonable solution to everyone?

cpp/src/neighbors/detail/vamana/vamana_build.cuh
```diff
@@ -408,9 +533,117 @@ index<T, IdxT> build(
   batched_insert_vamana<T, float, IdxT, Accessor>(
     res, params, dataset, vamana_graph.view(), &medoid_id, metric);
 
+  std::optional<raft::device_matrix<uint8_t, int64_t>> quantized_vectors;
+  if (params.codebook_prefix.size()) {
+    cuvs::neighbors::vpq_params pq_params;
```
bkarsin, line 538: Question for @cjnolet - is the `vpq` API changing with 25.06 or in the future with the release of the new quantization API? If so, is CAGRA-Q having to change (which means this would have to change as well)?

cpp/src/neighbors/detail/vamana/vamana_serialize.cuh
```diff
@@ -36,6 +36,247 @@
...
+ *
+ * @param[in] res the raft resource handle
+ * @param[in] h_graph the host matrix representation of the VAMANA index
+ * @param[in] dataset the raw dataset from which the VAMANA index is built
+ * @param[in] medoid the medoid
+ * @param[out] output_writer the stream object for writing the output file
+ *
+ */
+template <typename T, typename IdxT, typename HostMatT>
+void serialize_sector_aligned(raft::resources const& res,
```
bkarsin, line 102: The files created, their names, and how to use them with CPU DiskANN search is quite complex. Added an issue to add documentation for this as a future item: https://github.com/rapidsai/cuvs/issues/906
Not a blocker for this PR, though.

cpp/include/cuvs/neighbors/vamana.hpp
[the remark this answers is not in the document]
tarang-jain: Ah I did not realize that data() returns a `dataset` object in CAGRA too. In that case `vamana::index::data()` should be fine. But maybe the quantized data can be returned as a matrix. That is just my opinion, but @cjnolet can help design the API.

cpp/include/cuvs/neighbors/common.hpp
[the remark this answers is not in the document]
tarang-jain: That seems like a reasonable solution to me.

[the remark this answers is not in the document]
jamxia155: Updated as discussed on [slack](https://nvidia.slack.com/archives/C06FHK4NVL7/p1747429041713989).

cpp/include/cuvs/neighbors/vamana.hpp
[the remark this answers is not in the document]
jamxia155: Updated to const `device_matrix_view`.

cpp/src/neighbors/detail/vamana/vamana_build.cuh
jamxia155: Makes sense. Removed.

cpp/src/neighbors/detail/vamana/vamana_build.cuh
jamxia155: Updated to follow existing convention.

cpp/include/cuvs/neighbors/vamana.hpp
```diff
@@ -514,12 +552,48 @@ void serialize(raft::resources const& handle,
...
+ *   int dim = 64;
+ *   // construct codebook parameters from input codebook files
+ *   auto codebooks = cuvs::neighbors::vamana::get_codebooks(codebook_prefix, dim);
+ * @endcode
+ *
+ * @param[in] codebook_prefix path prefix to pq pivots and rotation matrix files
+ * @param[in] dim dimension of vectors in dataset
+ *
+ */
+auto get_codebooks(const std::string& codebook_prefix, const int dim)
```
cjnolet, line 595: I think we can improve the verbiage here. Can we call this `deserialize_codebooks` so that it matches the format of the other functinos that load data from a file? `get_codebooks` sounds too much like an accessor that we would have for accessing local state on an object.

cpp/include/cuvs/neighbors/vamana.hpp
```diff
@@ -212,11 +233,24 @@ struct index : cuvs::neighbors::index {
     graph_view_ = graph_.view();
   }
 
+  /**
+   * Replace the quantized dataset.
```
cjnolet, line 237: Can you provide more description here as to what's expected? Is this expecting that the dataset is already quantized or is a new dataset of raw vectors that's going to be quantized once updated? It's unclear the expectations here. 

Also, for public API functions and methods, please always document the individual arguments (and mark whether they are input or output arguments).

cpp/include/cuvs/neighbors/vamana.hpp
jamxia155: Updated. Thanks for the suggestion!

cpp/tests/neighbors/ann_vamana.cuh
```diff
@@ -157,6 +157,8 @@ class AnnVamanaTest : public ::testing::TestWithParam<AnnVamanaInputs> {
 
     tmp_index_file index_file;
     vamana::serialize(handle_, index_file.filename, index);
+    if (index_params.codebooks)
+      vamana::serialize(handle_, index_file.filename + "_sector_aligned", index, true, true);
```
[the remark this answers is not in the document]
jamxia155: Added testing of serialization & quantization code path using codebooks of dim = 384. Also checks for creation of output files.

cpp/include/cuvs/neighbors/vamana.hpp
jamxia155: Currently this expects an already quantized dataset. Substantiated documentation.

cpp/tests/neighbors/ann_vamana.cuh
```diff
@@ -137,6 +138,10 @@ class AnnVamanaTest : public ::testing::TestWithParam<AnnVamanaInputs> {
     index_params.visited_size      = ps.visited_size;
     index_params.max_fraction      = ps.max_fraction;
     index_params.reverse_batchsize = ps.reverse_batchsize;
+    // use codebooks with dim=384 to test serialization & quantization code path
+    if (ps.dim == 384)
+      index_params.codebooks = vamana::deserialize_codebooks(
+        std::string(TEST_DATA_DIR) + "neighbors/ann_vamana/quantizer_fbv6_384_64", ps.dim);
```
tarang-jain, line 144: Hardcoding the path to the codebooks like this seems brittle. There should be a way of passing it in the index params or inputs.

cpp/tests/neighbors/ann_vamana.cuh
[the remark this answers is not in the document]
tarang-jain: I am not fully on board with the sufficiency of tests here. We are testing the existence of the output files, but looks like the recall is still measured against a CAGRA index similar to the Vamana in-memory index tests? The accuracy of the quantized vectors or whether the resultant DiskANN SSD index works as expected is not measured. I know this is not possible currently because DiskANN is not a dependency + we cannot hardcode entire expected quantized vectors, so this is not a request for changes, but rather just opening up a discussion thread to brainstorm.

jamxia155:
> Should the `ann_vamana` data files be handled in the same way that the `cuvs_bench` data are handled

We got the `ann_vamana` test data files deployed to s3 so may I please ask for some pointers to how the `cuvs_bench` data are handled? Thanks!

cjnolet:
@jamxia155 I sincerely apologize for the delayed response here. We had some critical issues for release 25.06 that took several folks from the team over a week to resolve. Good news is that the release is now complete and we're available to help with timely reviews and guidance. 

I notice the last several commits are related to the cmake download of the test files. These will need to work both locally for users, as well as in CI. Are they working locally for you?

jamxia155:
> I notice the last several commits are related to the cmake download of the test files. These will need to work both locally for users, as well as in CI. Are they working locally for you?

Hi @cjnolet, those changes worked locally for me but are not working in the CI yet. However, I have since received helpful guidance from @bdice on the pattern that's currently used in cuGraph and I think I have a good handle on what needs to be done now. Thanks.

ci/test_cpp.sh
```diff
@@ -30,6 +30,14 @@ rapids-print-env
 rapids-logger "Check GPU usage"
 nvidia-smi
 
+# RAPIDS_DATASET_ROOT_DIR is used by test scripts
+RAPIDS_DATASET_ROOT_DIR=${RAPIDS_TESTS_DIR}/dataset
+mkdir -p "${RAPIDS_DATASET_ROOT_DIR}"
+export RAPIDS_DATASET_ROOT_DIR
+pushd "${RAPIDS_DATASET_ROOT_DIR}"
+${GITHUB_WORKSPACE}/cpp/tests/get_test_data.sh --NEIGHBORS_ANN_VAMANA_TEST
```
bdice: This will only work in GitHub Actions and will fail when reproducing CI locally. We need a solution that doesn't involve `${GITHUB_WORKSPACE}`.

I would make the `get_test_data.sh` script aware of `${RAPIDS_DATASET_ROOT_DIR}`. That way it has a default location in which it downloads, and you can continue to call that script from the repository root with `./cpp/tests/get_test_data.sh`. Avoid changing directories if you can, just to keep this script cleaner.

Also I would consider moving `get_test_data.sh` to a different folder. cuGraph uses a `datasets/` directory to manage these scripts, which is good. https://github.com/rapidsai/cugraph/tree/branch-25.08/datasets Otherwise you could put it in `ci/`.

jamxia155: Avoided `${GITHUB_WORKSPACE}`, moved `get_test_data.sh` to `ci/` (there doesn't seem to be an existing location that's more fitting), let `get_test_data.sh` get the download path from `${RAPIDS_DATASET_ROOT_DIR}` if defined.

divyegala:
Approving binary size addition to `libcuvs.so` of ~8.3 MB given the feature importance. @bkarsin @jamxia155 to file a follow-on issue to work on identifying any potential for size reduction.

cpp/include/cuvs/neighbors/vamana.hpp
```diff
@@ -212,11 +234,28 @@ struct index : cuvs::neighbors::index {
...
+   *
+   * @param[in] res
+   * @param[in] new_quantized_dataset the new quantized dataset for the index
+   *
+   */
+  void update_quantized_dataset(
+    raft::resources const& res,
+    raft::device_matrix<uint8_t, int64_t, raft::row_major>&& new_quantized_dataset)
+  {
+    quantized_dataset_ = new_quantized_dataset;
```
tarang-jain: Is the index supposed to own the quantized data? Can you remind me what was finally decided? Furthermore, even if it is supposed to own it, I think the argument of this function should be an mdspan (device_matrix_view). I am not sure if we should assume ownership by moving the data that the user supplies, but instead make a copy of it.

cpp/src/neighbors/detail/vamana/vamana_build.cuh
```diff
@@ -376,6 +379,21 @@ void batched_insert_vamana(
...
 
+template <typename T>
+auto quantize_all_vectors(raft::resources const& res,
+                          raft::device_matrix_view<const T, int64_t> residuals,
+                          raft::device_matrix_view<float, uint32_t, raft::row_major> pq_codebook,
+                          cuvs::neighbors::vpq_params ps)
+  -> raft::device_matrix<uint8_t, int64_t, raft::row_major>
+{
+  auto dim         = residuals.extent(1);
+  auto vq_codebook = raft::make_device_matrix<float, uint32_t, raft::row_major>(res, 1, dim);
```
tarang-jain, line 390: Is this supposed to be only a single center?

tarang-jain (changes requested):

cpp/include/cuvs/neighbors/vamana.hpp
jamxia155: The index needs to own its quantized data, as the quantized data is created during index build and will not live beyond it. What is the incentive for avoiding ownership transfer and paying the cost of making a copy?

BTW I'm curious how we can make the argument an mdspan while assuming ownership (as you seemed to allude to as a possibility)? The RAFT [documentation](https://docs.rapids.ai/api/raft/stable/cpp_api/mdspan_mdspan/#mdspan-multi-dimensional-non-owning-view) seems to be explicit about mdspan being non-owning.

cpp/src/neighbors/detail/vamana/vamana_build.cuh
jamxia155: I believe so. Ben worked on `process_and_fill_codes_subspaces` to more efficiently train multiple codebooks for each subspace as compared to `process_and_fill_codes`. Ideally (and eventually), we'd want to unify these functions and associated kernels so that the CAGRA-Q kernel can just be a special case with only 1 subspace. However, Ben did not want to convolve that change with this PR as it is non-trivial due to its effects on CAGRA-Q code paths.

cpp/include/cuvs/neighbors/vamana.hpp
tarang-jain: Yeah I was just saying that we can create a copy of the quantized data. The incentive that I had in mind is to not assume ownership of objects created by the user who calls this function, so that if a user wants to use that object later on, they don't face problems. In my opinion, it should either be the user's responsibility to ensure that the quantized data is alive for the lifetime of the index, or the index creates a copy of it. But assuming ownership of data created by the user might not be right.

jamxia155: Since we use this function in a situation where we can't ensure the lifetime of the input quantized data, I've updated it to make a copy.

cjnolet (approved):

Issue #849, opened 2025-04-28
[BUG] NN Descent returns wrong distances for `metric=InnerProduct`


jinsolp:
NN Descent calculates `InnerProduct` metric by negating the dot product ([link](https://github.com/rapidsai/cuvs/blob/0778a959ad482ea7da80198c36a375c658efee6c/cpp/src/neighbors/detail/nn_descent.cuh#L627-L628)) and sorting the distances from smallest to largest.

The current implementation does not negate the distances back after sorting, which results in NN Descent returning proper indices, but wrong distances.

PR #859, opened 2025-04-29
[FIX] Revert negated InnerProduct distance for NN Descent

jinsolp:
Closing https://github.com/rapidsai/cuvs/issues/849

Many parts of NN Descent is written on top of the assumption that smaller distance is closer.
Therefore, using the current trick (using -(dot product) values) to make it consistent with other metrics to be sorted in ascending order, then negating the value at the end should be the neatest solution.


cpp/src/neighbors/detail/nn_descent.cuh
```diff
@@ -1196,9 +1196,16 @@ void GNND<Data_t, Index_t>::build(Data_t* data,
     raft::matrix::slice<DistData_t, int64_t, raft::row_major>(
       res, raft::make_const_mdspan(graph_d_dists.view()), output_dist_view, coords);
 
+    // distance post-processing
     if (build_config_.metric == cuvs::distance::DistanceType::L2SqrtExpanded) {
       raft::linalg::map(
         res, output_dist_view, raft::sqrt_op{}, raft::make_const_mdspan(output_dist_view));
+    } else if (build_config_.metric == cuvs::distance::DistanceType::InnerProduct) {
```
benfred, line 1203: nitpick: Can we use the `is_min_close` function here instead of checking for IP distance itself?
```suggestion
    } else if (!cuvs::distance::is_min_close(build_config_.metric)) {
```

cjnolet: +1

benfred (approved):

PR #860, opened 2025-04-30
[Feat] Expose C API for CAGRA `merge`

rhdong:
- Expose C API for CAGRA `merge`

### DIFF
--- a/cpp/include/cuvs/neighbors/cagra.h
+++ b/cpp/include/cuvs/neighbors/cagra.h
@@ -339,6 +339,38 @@ cuvsError_t cuvsCagraIndexGetDims(cuvsCagraIndex_t index, int* dim);
  */
 
 /**
+ * @defgroup cagra_c_merge_params C API for CUDA ANN Graph-based nearest neighbor search
+ * @{
+ */
+
+/**
+ * @brief Supplemental parameters to merge CAGRA index
+ *
+ */
+/** Strategy for merging CAGRA indices. */
+typedef enum {
+  PHYSICAL = 0,  ///< Merge indices physically
+  LOGICAL  = 1   ///< Merge indices logically (if supported)
+} cuvsCagraMergeStrategy;
+
+struct cuvsCagraMergeParams {
+  cuvsCagraIndexParams_t output_index_params;
+  cuvsCagraMergeStrategy strategy;
+};
+
+typedef struct cuvsCagraMergeParams* cuvsCagraMergeParams_t;
+
+/** Allocate CAGRA merge params with default values */
+cuvsError_t cuvsCagraMergeParamsCreate(cuvsCagraMergeParams_t* params);
+
+/** De-allocate CAGRA merge params */
+cuvsError_t cuvsCagraMergeParamsDestroy(cuvsCagraMergeParams_t params);
+
+/**
+ * @}
+ */
+
+/**
  * @defgroup cagra_c_index_build C API for CUDA ANN Graph-based nearest neighbor search
  * @{
  */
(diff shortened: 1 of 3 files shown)


cpp/include/cuvs/neighbors/cagra.h
```diff
@@ -334,6 +334,38 @@ cuvsError_t cuvsCagraIndexDestroy(cuvsCagraIndex_t index);
...
+ * @defgroup cagra_c_merge_params C API for CUDA ANN Graph-based nearest neighbor search
+ * @{
+ */
+
+/**
+ * @brief Supplemental parameters to merge CAGRA index
+ *
+ */
+/** Strategy for merging CAGRA indices. */
+typedef enum {
```
cjnolet, line 351: Can we reuse this in the C++ version so we don't have to duplicate?

rhdong: No, I'm afraid we can't because C doesn't support namespaces, `::` , and `enum class`. It's not first we redefine in C which should work well, like, [`hash_mode`](https://github.com/rapidsai/cuvs/blob/branch-25.06/cpp/include/cuvs/neighbors/cagra.hpp#L173) vs [`cuvsCagraHashMode`](https://github.com/rapidsai/cuvs/blob/branch-25.06/cpp/include/cuvs/neighbors/cagra.h#L224)

cpp/src/neighbors/cagra_c.cpp
```diff
@@ -263,6 +263,101 @@ void* _deserialize(cuvsResources_t res, const char* filename)
...
+          auto sp                                          = ivf_params->ivf_pq_search_params;
+          pq_params.search_params.n_probes                 = sp->n_probes;
+          pq_params.search_params.lut_dtype                = sp->lut_dtype;
+          pq_params.search_params.internal_distance_dtype  = sp->internal_distance_dtype;
+          pq_params.search_params.preferred_shmem_carveout = sp->preferred_shmem_carveout;
+        }
+        if (ivf_params->refinement_rate > 1.0f) {
+          pq_params.refinement_rate = ivf_params->refinement_rate;
+        }
+      }
```
benfred, line 327: A bunch of this code seems cut and paste from the _build function - can we move into a function and share it so that we don't need to keep up to date in two spots?

cpp/include/cuvs/neighbors/cagra.h
rhdong: > We do this with the the "DistanceType` in pairwise distances. Maybe we can do the same thing here?

Sorry, I found we defined the separate [DistanceType](https://github.com/rapidsai/cuvs/blob/branch-25.06/cpp/include/cuvs/distance/distance.h#L69C3-L69C19) and converted it to cpp type via [`static_cast`](https://github.com/rapidsai/cuvs/blob/branch-25.06/cpp/src/neighbors/cagra_c.cpp#L45). Maybe I found a wrong position. Could you hint me at the code position by any chance, many thanks?

cpp/src/neighbors/cagra_c.cpp
rhdong: Done

cpp/include/cuvs/neighbors/cagra.h
cjnolet: @rhdong, [this](https://github.com/rapidsai/cuvs/blob/7affbc034e36cb66288391302ba6f910b54cd517/cpp/include/cuvs/distance/distance.hpp#L29) is where we map the C++ type to the C type. That static cast shouldn't be needed and I suspect it's an artifact from before we have formally combined them. You are free to fix it if you have time.

rhdong: Oh, I understood! Let C++ be compatible with C :smiley:

benfred (approved):
lgtm!

rhdong:
> lgtm!

I'm processing the CI fail.. sorry

rhdong:
Sorry, I accidentally did a force push.

cpp/include/cuvs/neighbors/common_defs.h
```diff
@@ -0,0 +1,56 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * @file common_defs.h
```
rhdong, line 18: Hi @benfred @cjnolet, I introduced a new C header to hold the declaration shared between C & C++ domains, while minimizing the introduction of extra headers and C objects. Could you take a moment to help review? Thanks!

cjnolet: Any reason these can't be added to `common.h`?

rhdong: When we added these defines to common.h, when including it in cagra.hpp, many unnecessary symbols will dirty the C++ global space, since there are many headers in common.h and two enums with not-unique member naming.
The more difficult thing is that it will cause compilation error for not found '#include <dlpack/dlpack.h>'

rhdong: Hello @benfred @cjnolet , I did some changes on the headers arrangements, would you please take a moment to review? Thanks!

cjnolet (approved):

PR #870, opened 2025-05-03
Puneet/cagraprefiltering

punAhuja:
Added prefiltering logic for CAGRA search in the java layer.

Added tests to test prefiltering in CAGRA search

### DIFF
--- a/java/cuvs-java/src/main/java/com/nvidia/cuvs/CagraQuery.java
+++ b/java/cuvs-java/src/main/java/com/nvidia/cuvs/CagraQuery.java
@@ -18,6 +18,7 @@ package com.nvidia.cuvs;
 
 import java.util.Arrays;
 import java.util.List;
+import java.util.BitSet;
 
 /**
  * CagraQuery holds the CagraSearchParams and the query vectors to be used while
@@ -31,6 +32,8 @@ public class CagraQuery {
   private List<Integer> mapping;
   private float[][] queryVectors;
   private int topK;
+  private BitSet[] prefilters;
+  private int numDocs;
 
   /**
    * Constructs an instance of {@link CagraQuery} using cagraSearchParameters,
@@ -42,12 +45,14 @@ public class CagraQuery {
    * @param mapping               an instance of ID mapping
    * @param topK                  the top k results to return
    */
-  public CagraQuery(CagraSearchParams cagraSearchParameters, float[][] queryVectors, List<Integer> mapping, int topK) {
+  public CagraQuery(CagraSearchParams cagraSearchParameters, float[][] queryVectors, List<Integer> mapping, int topK, BitSet[] prefilters, int numDocs) {
     super();
     this.cagraSearchParameters = cagraSearchParameters;
     this.queryVectors = queryVectors;
     this.mapping = mapping;
     this.topK = topK;
+    this.prefilters = prefilters;
+    this.numDocs = numDocs;
   }
 
   /**
@@ -85,6 +90,14 @@ public class CagraQuery {
   public int getTopK() {
     return topK;
   }
+  
+  public BitSet[] getPrefilters() {
+  return prefilters;
+  }
+
+  public int getNumDocs() {
+    return numDocs;
+  }
 
   @Override
   public String toString() {
@@ -101,6 +114,8 @@ public class CagraQuery {
     private float[][] queryVectors;
     private List<Integer> mapping;
     private int topK = 2;
+    private BitSet[] prefilters;
+    private int numDocs;
 
     /**
      * Default constructor.
@@ -152,6 +167,12 @@ public class CagraQuery {
       this.topK = topK;
       return this;
     }
+    
+    public Builder withPrefilter(BitSet[] prefilters, int numDocs) {
+    this.prefilters = prefilters;
+    this.numDocs = numDocs;
+    return this;
+  }
 
     /**
      * Builds an instance of CuVSQuery.
@@ -159,7 +180,7 @@ public class CagraQuery {
      * @return an instance of CuVSQuery
      */
     public CagraQuery build() {
-      return new CagraQuery(cagraSearchParams, queryVectors, mapping, topK);
+      return new CagraQuery(cagraSearchParams, queryVectors, mapping, topK, prefilters, numDocs);
     }
   }
 }
--- a/java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/CagraIndexImpl.java
+++ b/java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/CagraIndexImpl.java
@@ -54,6 +54,7 @@ import com.nvidia.cuvs.internal.panama.cuvsCagraSearchParams;
 import com.nvidia.cuvs.internal.panama.cuvsIvfPqIndexParams;
 import com.nvidia.cuvs.internal.panama.cuvsIvfPqParams;
 import com.nvidia.cuvs.internal.panama.cuvsIvfPqSearchParams;
+import java.util.BitSet;
 
 /**
  * {@link CagraIndex} encapsulates a CAGRA index, along with methods to interact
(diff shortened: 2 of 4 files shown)


chatman:
Thanks @punAhuja.
For the sake of consistency, can we test this similarly to how we tested the BruteForceIndex's pre-filtering?
https://github.com/rapidsai/cuvs/blob/branch-25.06/java/cuvs-java/src/test/java/com/nvidia/cuvs/BruteForceRandomizedIT.java#L63

Maybe there's scope to refactor the `generateData` and `generateExpectedResults` in the `BruteForceRandomizedIT` to a common place and use it here.

java/cuvs-java/src/main/java/com/nvidia/cuvs/CagraQuery.java
```diff
@@ -153,13 +181,26 @@ public Builder withTopK(int topK) {
...
 
+    /**
+     * Sets the prefilters data for building the {@link CagraQuery}.
+     *
+     * @param prefilters array of bitsets, as many as queries, each containing as
+     *        many bits as there are vectors in the index
+     * @return an instance of this Builder
+     */  
+    public Builder withPrefilter(BitSet[] prefilters, int numDocs) {
+    this.prefilters = prefilters;
```
chatman, line 192: Indentation is improper, please fix.

cjnolet:
@chatman @punAhuja I've marked this PR as a breaking change for the Java API. Please let me know if that's not the case and I'll change it.

chatman:
I requested for a change to rename `withPrefilter` method to `withPrefilters`, even in previously implemented brute force query. If that change is incorporated, this PR will represent a breaking change for Java API.

java/cuvs-java/src/main/java/com/nvidia/cuvs/CagraQuery.java
punAhuja: Fixed

java/cuvs-java/src/main/java/com/nvidia/cuvs/CagraQuery.java
```diff
@@ -153,13 +181,26 @@ public Builder withTopK(int topK) {
...
     }
 
+    /**
+     * Sets the prefilters data for building the {@link CagraQuery}.
+     *
+     * @param prefilters array of bitsets, as many as queries, each containing as
+     *        many bits as there are vectors in the index
+     * @return an instance of this Builder
+     */  
+    public Builder withPrefilter(BitSet[] prefilters, int numDocs) {
```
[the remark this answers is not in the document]
punAhuja: Changed to withPrefilters for both CagraQuery and BruteForceQuery.

cjnolet:
@punAhuja thanks for this PR. Now that we have Java enabled in CI, can you update your branch to the current branch-25.06?

java/cuvs-java/src/main/java/com/nvidia/cuvs/BruteForceQuery.java
```diff
@@ -155,7 +155,7 @@ public Builder withTopK(int topK) {
      *        many bits as there are vectors in the index
      * @return an instance of this Builder
      */
-    public Builder withPrefilter(BitSet[] prefilters, int numDocs) {
+    public Builder withPrefilters(BitSet[] prefilters, int numDocs) {
```
cjnolet, line 158: I see now what you mean- this is an API compatabiliity thing because it depends on whether we can specify an array of bitset at the java layer or a single bitset. Can we specify a single bitset for now?

java/cuvs-java/src/main/java/com/nvidia/cuvs/CagraQuery.java
```diff
@@ -31,6 +32,8 @@ public class CagraQuery {
   private List<Integer> mapping;
   private float[][] queryVectors;
   private int topK;
+  private BitSet[] prefilters;
+  private int numDocs;
```
cjnolet, line 36: Are we assuming len(prefilters) == numDocs? Any reason we can't just use the length of prefilters?

java/cuvs-java/src/main/java/com/nvidia/cuvs/CagraQuery.java
```diff
@@ -86,6 +94,24 @@ public int getTopK() {
...
+   * Gets the prefilter BitSet array.
+   *
+   * @return an array of BitSet objects representing the prefilters
+   */
+  public BitSet[] getPrefilters() {
+    return prefilters;
+  }
+
+  /**
+   * Gets the number of documents supposed to be in this index, as used for prefilters
```
cjnolet, line 107:
```suggestion
   * Gets the number of documents in this index, as used for prefilters
```

java/cuvs-java/src/main/java/com/nvidia/cuvs/BruteForceQuery.java
punAhuja: In the C layer, there is support for single BitSet for cuvs, but for BruteForce, there is support for BitMap (which is like an array of BitSets)
So for cuvs we pass just one filter which is global and gets applied to all queries.

java/cuvs-java/src/main/java/com/nvidia/cuvs/CagraQuery.java
punAhuja: prefilters is a BitSet, and prefilters.length will give us the logical length (highest bit+1) .
It is safer to use numDocs to ensure proper flattening of the BitSet.

chatman: Just for context from the brute force case: 
It is possible for a user to specify a bitset with very few bits set. In that case, prefilters.length will be much less than the number of documents in the index. But, the underlying C api expects numDocs*numqueries elements in the bitset for the case of bruteforce (where per query filters are supported), and hence every filter should contain padded bits to be equal to the size of numDocs.

java/cuvs-java/src/main/java/com/nvidia/cuvs/CagraQuery.java
```diff
@@ -31,6 +32,8 @@ public class CagraQuery {
   private List<Integer> mapping;
   private float[][] queryVectors;
   private int topK;
+  private BitSet[] prefilters;
```
chatman, line 35: I thought Cagra doesn't support per query prefilters. Why do we have an array of prefilters here? Did I misunderstand the first part?

java/cuvs-java/src/main/java/com/nvidia/cuvs/BruteForceQuery.java
chatman: Regarding the user having to pass in a `numDocs`, I think this one should be deduced automatically. We can tackle this in a separate PR. I remember @narangvivek10 mentioning that there's some function/API in cuVS where meta information for a CAGRA index can be obtained (e.g. size of the index, num docs etc.). If we can use that function/API during the loading of the CAGRA index, then the `numDocs` info should be available in the CagraIndex class itself (instead of having the user pass it into the CagraQuery object). Lets explore this as a follow up to this PR, though.

chatman:
I've added more repetitions to the CagraRandomizedTest, relaxed the matching criteria even more (look at additional 10 neighbors). Still there are failures.

Changing:
```
    boolean usePrefilter = random.nextBoolean();
```
to
```
    boolean usePrefilter = false; //random.nextBoolean();
```

makes it always pass. Sounds like some bug in prefiltering (either at Java or C layers).
FYI @narangvivek10.

java/cuvs-java/src/test/java/com/nvidia/cuvs/CagraRandomizedIT.java
```diff
@@ -55,14 +55,23 @@ private void tmpResultsTopKWithRandomValues() throws Throwable {
...
     int numQueries = random.nextInt(NUM_QUERIES_LIMIT) + 1;
     int topK = Math.min(random.nextInt(TOP_K_LIMIT) + 1, datasetSize);
+    boolean usePrefilter = random.nextBoolean();
 
     if (datasetSize < topK)
       datasetSize = topK;
 
+    BitSet sharedPrefilter = null;
+    if (usePrefilter) {
+      sharedPrefilter = new BitSet(datasetSize);
```
narangvivek10, line 69: The prefilter is not passed to the `generateExpectedResults` below, and must be causing the test failures. Please fix this @punAhuja. FYI @chatman.

punAhuja:
@cjnolet Have added fix for the tests. please the ok to test.

chatman:
LGTM.

> -Discovered another issue: when results returned are less than topK (because of filter), in cuvs padding gets added with INT_MAX values. Added fix to ignore padding values as well.

It would be better to exclude these from the results, instead of ignoring these in the test.

chatman (approved):
Lgtm.

cjnolet (approved):

java/cuvs-java/src/main/java/com/nvidia/cuvs/BruteForceQuery.java
cjnolet: @chatman can you create a Github issue for this please, so that we don't lose sight of it?

cjnolet:
> @cjnolet the Java tests passed here. Can we merge this, please?

@chatman yes, the plan is to merge this. Unfortunately, we can't merge until all of CI passes and we had some flaky tests fail.

PR #871, opened 2025-05-04
Spectral Embedding

### LINKED ISSUE #1025: [FEA] Spectral Embedding
Tracking issue for Spectral Embedding cpp

aamijar:
Depends on https://github.com/rapidsai/raft/pull/2662 and Resolves #1025


cpp/include/cuvs/preprocessing/spectral/spectral_embedding.hpp
```diff
@@ -0,0 +1,30 @@
...
+#include <cuvs/preprocessing/spectral/spectral_embedding_types.hpp>
+
+#include <raft/core/device_mdspan.hpp>
+#include <raft/core/resources.hpp>
+
+namespace cuvs::preprocessing::spectral {
+
+// template <typename IndexTypeT, typename ValueTypeT>
+auto spectral_embedding(raft::resources const& handle,
+                        raft::device_matrix_view<float, int, raft::row_major> nums,
```
cjnolet, line 26: We always put params in the order of:

1. Resources
2. Params struct
3. Input mdspan
4. Output mdspan
5. Free params (and defaults) 

Please make this consistent.

cpp/include/cuvs/preprocessing/spectral/spectral_embedding_types.hpp
```diff
@@ -0,0 +1,36 @@
...
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include <cstdint>
+
+namespace cuvs::preprocessing::spectral {
+
+struct spectral_embedding_config {
```
cjnolet, line 23: cuVS is not header-only, so we no longer need an additional "types" header. Put the API definition and the params struct into the same header file please. We do this for RAFT because the params are lightweight but the header itself pulls in all the cuh headers- that way we have a lightweight header for being able to specify API vocabilary that's independent of the code that needs to be compiled w/ CUDA. Since cuVS is a formal C++ API (with source files) like cuML, we don't need this restriction. 

Please also rename this to `cuvs::preprocessing::spectral::params` so that it is consistent with the other APIs (no need to duplicate the "spectral_embedding_" since that's in the namespace).

cpp/src/preprocessing/spectral/spectral_embedding.cu
```diff
@@ -0,0 +1,241 @@
...
+                    thrust::counting_iterator<int>(nnz),
+                    knn_rows.data_handle(),
+                    [=] __device__(int64_t i) { return static_cast<int>(i / k_search); });
+
+  // Copy COO data to device
+  raft::copy(knn_coo.rows(), knn_rows.data_handle(), nnz, stream);
+  raft::copy(knn_coo.cols(), knn_cols.data_handle(), nnz, stream);
+  raft::copy(knn_coo.vals(), d_distances.data_handle(), nnz, stream);
+
+  raft::sparse::COO<float> coo_no_zeros(stream);  // Don't pre-allocate dimensions
```
cjnolet, line 110: It's okay to use this API in internal code, but we need to make sure we're using the new raft sparse `coo_matrix_view` in any public APIs. `raft::sparse::COO` is deprecated and we'll eventually have to change this call. Please create an issue in RAFT for updating this call to use the new raft sparse API types and reference that issue here for completeness. That way we can do a simple grep to find this and fix it.

cpp/src/preprocessing/spectral/spectral_embedding.cu
```diff
@@ -0,0 +1,241 @@
...
+  raft::sparse::op::coo_remove_zeros<float>(&knn_coo, &coo_no_zeros, stream);
+
+  // binarize to 1s
+  thrust::fill(thrust::device, coo_no_zeros.vals(), coo_no_zeros.vals() + coo_no_zeros.nnz, 1.0f);
```
cjnolet, line 114: We have `fill` functions in RAFT. Please use them.

cpp/src/preprocessing/spectral/spectral_embedding.cu
```diff
@@ -0,0 +1,241 @@
...
+  // Symmetrize the matrix
+  raft::sparse::linalg::coo_symmetrize(&coo_no_zeros, &sym_coo1, reduction_op, stream);
+
+  raft::sparse::op::coo_sort<float>(&sym_coo1, stream);
+
+  raft::sparse::COO<float> sym_coo(stream);  // Don't pre-allocate dimensions
+  raft::sparse::op::coo_remove_zeros<float>(&sym_coo1, &sym_coo, stream);
+
+  nnz = sym_coo.nnz;
+  printf("\nSymmetrized COO Matrix (nnz=%ld):\n", nnz);
```
cjnolet, line 134: Please remove prints. We shouldbe using the debug logger macros.

cpp/src/preprocessing/spectral/spectral_embedding.cu
```diff
@@ -0,0 +1,241 @@
...
+  using size_type = size_t;
+
+  raft::sparse::op::coo_sort<float>(&sym_coo, stream);
+  auto row_ind = raft::make_device_vector<int>(handle, sym_coo.n_rows + 1);
+  raft::sparse::convert::sorted_coo_to_csr(&sym_coo, row_ind.data_handle(), stream);
+
+  const int one = sym_coo.nnz;
+  raft::copy(row_ind.data_handle() + row_ind.size() - 1, &one, 1, stream);
+
+  auto csr_structure = raft::make_device_compressed_structure_view<int, int, int>(
```
cjnolet, line 147: You are ure using this here- why not also use the coo version above? That'll save us a lot of refactoring time in the future.

cpp/src/preprocessing/spectral/spectral_embedding.cu
```diff
@@ -0,0 +1,241 @@
...
+
+  auto diagonal            = raft::make_device_vector<float>(handle, csr_structure.get_n_rows());
+  auto laplacian           = spectral_embedding_config.norm_laplacian
+                               ? raft::sparse::linalg::compute_graph_laplacian_normalized(
+                         handle, csr_matrix_view, diagonal.view())
+                               : raft::sparse::linalg::compute_graph_laplacian(handle, csr_matrix_view);
+  auto laplacian_structure = laplacian.structure_view();
+
+  // L *= -1
+  thrust::transform(thrust::device,
```
cjnolet, line 165: Use unary_op everywhere you are using `thrust::transform` please.

cpp/src/preprocessing/spectral/spectral_embedding.cu
```diff
@@ -0,0 +1,241 @@
...
+  raft::sparse::solver::lanczos_compute_smallest_eigenvectors<int, float>(
+    handle,
+    config,
+    raft::make_device_csr_matrix_view<float, int, int, int>(laplacian.get_elements().data(),
+                                                            laplacian_structure),
+    std::nullopt,
+    eigenvalues.view(),
+    eigenvectors.view());
+
+  raft::print_device_vector(
```
cjnolet, line 194: This should be a debug log

cpp/include/cuvs/preprocessing/spectral/spectral_embedding.hpp
```diff
@@ -0,0 +1,43 @@
...
+  /** Whether to normalize the Laplacian matrix. */
+  bool norm_laplacian;
+  /** Whether to drop the first eigenvector. */
+  bool drop_first;
+  /** random seed */
+  uint64_t seed;
+};
+
+// template <typename IndexTypeT, typename ValueTypeT>
+auto spectral_embedding(raft::resources const& handle,
```
cjnolet, line 38: Please rename to `transform` to match other preprocessing APIs.

cpp/src/preprocessing/spectral/detail/spectral_embedding.cuh
cjnolet, line 1: Please don't check empty files into the repo

cpp/tests/preprocessing/spectral_embedding.cu
```diff
@@ -0,0 +1,243 @@
...
+    ASSERT_LE(trust_score, 1.0) << "Trustworthiness score should be <= 1.0";
+
+    // For good embeddings, we expect a reasonably high trustworthiness score
+    // This threshold can be adjusted based on empirical results
+    ASSERT_GT(trust_score, 0.7)
+      << "Trustworthiness score is too low, indicating poor embedding quality";
+
+    // Optional: Save the embedding for visualization
+    /*
+    std::ofstream out_file("embedding_output.csv");
```
cjnolet, line 186: If you intend to keep this, I would suggest making this a function and accepting an argument to determine whether this should be done. We shouldn't keep commented out code in the codebase because it becomes very confusing to maintain as this code passes through different hands.

cpp/include/cuvs/preprocessing/spectral/spectral_embedding.hpp
cjnolet, line 1: Now that you've been able to wrap this through cuML, can you provide a sense of the speedup for different datasets? It would be super helpful to know what we are working with here.

cpp/include/cuvs/preprocessing/spectral/spectral_embedding.hpp
```diff
@@ -0,0 +1,43 @@
...
+
+#pragma once
+
+#include <raft/core/device_mdspan.hpp>
+#include <raft/core/resources.hpp>
+
+namespace cuvs::preprocessing::spectral_embedding {
+
+struct params {
+  /** The number of components to reduce the data to. */
```
aamijar, line 25: change to doxygen format

cpp/include/cuvs/preprocessing/spectral/spectral_embedding.hpp
```diff
@@ -0,0 +1,43 @@
...
+  int n_neighbors;
+  /** Whether to normalize the Laplacian matrix. */
+  bool norm_laplacian;
+  /** Whether to drop the first eigenvector. */
+  bool drop_first;
+  /** random seed */
+  uint64_t seed;
+};
+
+// template <typename IndexTypeT, typename ValueTypeT>
```
aamijar, line 37: enable support for `<uint32_t/uint64_t, float/double>`

cpp/src/preprocessing/spectral/spectral_embedding.cu
```diff
@@ -0,0 +1,242 @@
...
+#include <raft/sparse/solver/lanczos_types.hpp>
+#include <raft/util/cudart_utils.hpp>
+
+#include <driver_types.h>
+#include <thrust/copy.h>
+#include <thrust/iterator/counting_iterator.h>
+#include <thrust/iterator/zip_iterator.h>
+#include <thrust/remove.h>
+#include <thrust/transform_scan.h>
+#include <thrust/tuple.h>
```
aamijar, line 40: clean up and remove unused includes

cpp/src/preprocessing/spectral/spectral_embedding.cu
aamijar: I'm currently working on using the new coo_matrix types and adding support for the relevant functions in raft. Tracking here: https://github.com/rapidsai/raft/issues/2659 https://github.com/rapidsai/raft/issues/2656
Although I'm running into some cuda invalid memory accesses and I can't seem to debug why its happening. It was passing with my gtests, but when I connected it with cuML it didn't pass the pytests. I've narrowed it down to fail in the gtests when I call transform twice in a row.

cpp/src/preprocessing/spectral/spectral_embedding.cu
```diff
@@ -0,0 +1,241 @@
...
+      raft::linalg::Apply::ALONG_COLUMNS,  // divide each row by corresponding diagonal element
+      [] __device__(float elem, float diag) { return elem / diag; });
+  }
+
+  // Replace the direct copy with a gather operation that reverses columns
+
+  // Create a sequence of reversed column indices
+  config.n_components = drop_first ? config.n_components - 1 : config.n_components;
+  auto col_indices    = raft::make_device_vector<int>(handle, config.n_components);
+  thrust::sequence(thrust::device,
```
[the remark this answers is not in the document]
aamijar: Tracking here https://github.com/rapidsai/raft/issues/2661

cpp/src/preprocessing/spectral/spectral_embedding.cu
aamijar: Okay, I think I found the issue which is that I needed to initialize some vectors with zeros.

cpp/include/cuvs/preprocessing/spectral/spectral_embedding.hpp
aamijar: brute_force only supports float, int64_t so doesn't make sense to support others

cpp/include/cuvs/preprocessing/spectral/spectral_embedding.hpp
aamijar: Sorry about the delay! I added the benchmarks in cuML PR description https://github.com/rapidsai/cuml/pull/6581

cpp/include/cuvs/preprocessing/spectral/spectral_embedding.hpp
```diff
@@ -0,0 +1,50 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include <raft/core/device_mdspan.hpp>
+#include <raft/core/resources.hpp>
+
+namespace cuvs::preprocessing::spectral_embedding {
```
cjnolet: Let's have the directory structure align w/ the namespace. I'm okay having `preprocessing::spectral_embedding` (since `spectral_clustering` would be `cluster::spectral_clustering`), but let's make the directory structure `preprocessing/spectral_embedding.hpp` and remove the intermediate `spectral` directory.

cpp/src/preprocessing/spectral/spectral_embedding.cu
```diff
@@ -0,0 +1,270 @@
+/*
```
cjnolet, line 1: I'm okay keeping the intermediate `spectral` in the impl code to separate from the other preprocessing impls.

cpp/include/cuvs/preprocessing/spectral_embedding.hpp
```diff
@@ -0,0 +1,49 @@
...
+  bool drop_first;
+
+  /** @brief Random seed for reproducibility */
+  uint64_t seed;
+};
+
+auto transform(raft::resources const& handle,
+               params config,
+               raft::device_matrix_view<float, int, raft::row_major> dataset,
+               raft::device_matrix_view<float, int, raft::col_major> embedding) -> int;
```
cjnolet: Sorry just noticing this- why is transform() returning an int in the public API? This isn't a pattern we've been using in the C++ API. In C++, users should be catching exceptions.

cpp/src/preprocessing/spectral/spectral_embedding.cu
```diff
@@ -0,0 +1,265 @@
...
+    embedding.extent(0));
+
+  raft::matrix::gather<float, int, int>(
+    handle,
+    raft::make_const_mdspan(eigenvectors_row_view),  // Source matrix (as row-major view)
+    raft::make_const_mdspan(col_indices.view()),     // Column indices to gather
+    embedding_row_view                               // Destination matrix (as row-major view)
+  );
+
+  return 0;
```
cjnolet: Please remove the return. We shouldn't be relying on this in the C++ layer. This is a C-layer thing.

cpp/src/preprocessing/spectral/spectral_embedding.cu
```diff
@@ -0,0 +1,265 @@
...
+#include <cstdio>
+
+namespace cuvs::preprocessing::spectral_embedding {
+
+auto transform(raft::resources const& handle,
+               params spectral_embedding_config,
+               raft::device_matrix_view<float, int, raft::row_major> dataset,
+               raft::device_matrix_view<float, int, raft::col_major> embedding) -> int
+{
+  const int n_samples     = dataset.extent(0);
```
cjnolet: This is quite a behemoth of a function. Can you please break this down into a few more manageable (and potentially reusable) functions?

cpp/src/preprocessing/spectral/spectral_embedding.cu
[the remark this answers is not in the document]
cjnolet: Great! Please reference the issue in a TODO in the code here (and above, where thrust::sequence is used). This is how we keep Github issues linked to the code so they don't get lost/forgotten.

cpp/src/preprocessing/spectral/spectral_embedding.cu
```diff
@@ -0,0 +1,253 @@
...
+#include <raft/util/cudart_utils.hpp>
+
+#include <rmm/device_uvector.hpp>
+
+#include <thrust/sequence.h>
+#include <thrust/tabulate.h>
+
+#include <cuvs/neighbors/brute_force.hpp>
+
+#include <cstdio>
```
lowener: Unnecessary include

cpp/src/preprocessing/spectral/spectral_embedding.cu
```diff
@@ -0,0 +1,253 @@
...
+    policy, knn_rows.data_handle(), knn_rows.data_handle() + nnz, [k_search] __device__(int idx) {
+      return idx / k_search;
+    });
+
+  // binarize to 1s
+  raft::matrix::fill(handle, raft::make_device_vector_view(d_distances.data_handle(), nnz), 1.0f);
+
+  auto reduction_op = [] __device__(int row, int col, float a, float b) { return 0.5f * (a + b); };
+
+  raft::resource::sync_stream(handle, stream);
```
lowener: Why so many `sync_stream` here and below?

cpp/src/preprocessing/spectral/spectral_embedding.cu
```diff
@@ -0,0 +1,253 @@
...
+  auto zero_scalar = raft::make_host_scalar<float>(0.0f);
+
+  auto sym_coo_matrix =
+    raft::make_device_coo_matrix<float, int, int, int>(handle, sym_coo1_n_rows, sym_coo1_n_cols);
+
+  // Create a const view of the input matrix
+  auto sym_coo1_matrix_const_view = raft::make_device_coo_matrix_view<const float, int, int, int>(
+    sym_coo1_matrix.get_elements().data(), sym_coo1_structure);
+
+  raft::sparse::op::coo_remove_scalar<1, float, int, int>(
```
lowener: Why TPB=1? I think you can increase that.

cpp/src/preprocessing/spectral/spectral_embedding.cu
aamijar: Ah yeah, I'll remove them. I think they were leftover from something I was debugging before. By the way ,do you have any guidelines on when to use sync_stream?

cpp/src/preprocessing/spectral/spectral_embedding.cu
aamijar: I will increase the value, any guidelines on how I should determine the value?

cpp/src/preprocessing/spectral/spectral_embedding.cu
lowener: In general:
- Before accessing data on host that originates from device (so after a data copy D2H)
- If using multiple streams, when you need to make sure that one stream is done with it's execution
- I also use it when switching from streams of different lib, such as thrust or cublas but I am not sure on that one

cpp/src/preprocessing/spectral/spectral_embedding.cu
lowener: I am not at all an expert in this but in general function that heavily use memory or register are better with a low TPB. If the function is not memory intensive then a higher TPB is recommended to maximize occupancy.
It should always be a multiple of 32 and TPB in the range 128-512 are generally safe choices.

cpp/src/preprocessing/spectral/spectral_embedding.cu
aamijar: Hi @cjnolet, I've cleaned up the function and made the code more concise. Can you see if it looks okay. I don't know if we should break it up further.

cpp/src/preprocessing/spectral/spectral_embedding.cu
cjnolet: Unfortunately, this is a bad (and deprecated) api exposed through RAFT today, which we should eventually clean up and fix (though we don't have time atm). I'd set this value to 128 in the meantime.

cpp/tests/preprocessing/spectral_embedding.cu
```diff
@@ -0,0 +1,227 @@
...
+ protected:
+  void SetUp() override
+  {
+    n_samples_    = params_.n_samples;
+    n_features_   = params_.n_features;
+    n_clusters_   = params_.n_clusters;
+    n_components_ = params_.n_components;
+
+    // Generate blob data with the specified parameters
+    // input_ = raft::make_device_matrix<T, int, raft::row_major>(handle, n_samples_, n_features_);
```
cjnolet: Please remove commented out code. This applies to tests as well. It's unclear to any future developers whether this was done intentionally, but we want to maintain good hhousekeeping habits. If this is meant to be something that's used for debugging, make it part of the test and supply a flag to turn it on or off.

cpp/src/preprocessing/spectral/spectral_embedding.cu
```diff
@@ -0,0 +1,217 @@
...
+
+  cuvs::neighbors::brute_force::search_params search_params;
+  cuvs::neighbors::brute_force::index_params index_params;
+  index_params.metric = cuvs::distance::DistanceType::L2SqrtExpanded;
+
+  auto d_indices   = raft::make_device_matrix<int64_t>(handle, n_samples, k_search);
+  auto d_distances = raft::make_device_matrix<float>(handle, n_samples, k_search);
+
+  auto index =
+    cuvs::neighbors::brute_force::build(handle, index_params, raft::make_const_mdspan(dataset));
```
cjnolet, line 62: Please pull this out into a "compute_nearest_neighbors" function. This will allow us to eventually swap the brute-force version out for an ANN with minimal changes. 

Even better, separate the  connectivities graph creation from the rest of the algorithm here- to clean it up a bit and make it easier to read and more maintainable. Maybe create a "create_connectivities_graph" function to pull all of that logic out of this behemoth function here.

cpp/src/preprocessing/spectral/spectral_embedding.cu
```diff
@@ -0,0 +1,217 @@
...
+  auto csr_matrix_view = raft::make_device_csr_matrix_view<float, int, int, int>(
+    const_cast<float*>(sym_coo_matrix.get_elements().data()),
+    raft::make_device_compressed_structure_view<int, int, int>(
+      const_cast<int*>(sym_coo_row_ind.data_handle()),
+      const_cast<int*>(sym_coo_matrix.structure_view().get_cols().data()),
+      n_samples,
+      n_samples,
+      sym_coo_matrix.structure_view().get_nnz()));
+
+  auto diagonal = raft::make_device_vector<float>(handle, n_samples);
```
cjnolet: After `create_connectivities`, please pull the laplacian logic into a separate ` construct_laplacian` function.

cpp/src/preprocessing/spectral/spectral_embedding.cu
```diff
@@ -0,0 +1,217 @@
...
+
+  auto config           = raft::sparse::solver::lanczos_solver_config<float>();
+  config.n_components   = spectral_embedding_config.n_components;
+  config.max_iterations = 1000;
+  config.ncv            = std::min(n_samples, std::max(2 * config.n_components + 1, 20));
+  config.tolerance      = 1e-5;
+  config.which          = raft::sparse::solver::LANCZOS_WHICH::LA;
+  config.seed           = spectral_embedding_config.seed;
+
+  auto eigenvalues =
```
cjnolet, line 183: Please pull all the eigensolver logic into a new function called something like "compute_eigenpairs". You can really simplify this function, making it easier to read and more maintainable by breaking the spectral embedding down into 3 steps:

Step 1: Construct connectivities graph
Step 2: Compute Laplacian
Step 3: Compute eigenpairs

This function should be really simple and just gluing together the wiring of the 3 steps above.

cpp/src/preprocessing/spectral/spectral_embedding.cu
aamijar: This is using new API which we worked on here https://github.com/rapidsai/raft/pull/2662

cpp/src/preprocessing/spectral/spectral_embedding.cu
aamijar: Thanks for the review! I've broken up the code so that the transform function does the three steps above.

cpp/include/cuvs/preprocessing/spectral_embedding.hpp
```diff
@@ -0,0 +1,76 @@
...
+  bool norm_laplacian;
+
+  /** @brief Whether to drop the first eigenvector. */
+  bool drop_first;
+
+  /** @brief Random seed for reproducibility */
+  uint64_t seed;
+};
+
+void create_connectivity_graph(raft::resources const& handle,
```
cjnolet: Is there a reason this needs to be exposed through the public APIs? This is all implementation details and should only be used within `src`/

cjnolet: To be clear- the user should only have to call `build()` and `tranform()` functions from cuVS to build a spectral embedding and then transform a series of vectors into that space. If we don't have those separated yet, we can just do a `transform` for now, but that should really be the only function we need to expose.

aamijar: I removed them from the public api now

PR #872, opened 2025-05-05
chore: lower wheel size threshold

gforsyth:


AyodeAwe (approved):

ci/build_wheel_libcuvs.sh
```diff
@@ -27,4 +27,28 @@ rapids-pip-retry install \
...
+pydistcheck \
+    --inspect \
+    --max-allowed-size-compressed $WHEEL_SIZE_LIMIT \
+    "$(echo "${RAPIDS_WHEEL_BLD_OUTPUT_DIR}"/*.whl)"
+
+rapids-logger "validate packages with 'twine'"
+
+twine check \
+    --strict \
+    "$(echo "${RAPIDS_WHEEL_BLD_OUTPUT_DIR}"/*.whl)"
```
jameslamb, line 54: Would you consider instead putting this logic down in `validate_wheel.sh`, as we used to do for other projects?

See the diff at https://github.com/rapidsai/cugraph/pull/4804/files#diff-0565e15203407439a89a2d7c3da1ca9569b26bb583b039009ad85394f1ff70f9L9 for how we previously did this for `cugraph`, as an example.

I think that's preferable because doing that avoids duplicating the calls to `pydistcheck` and `twine check` across multiple places, keeping the "what after-building validation of wheels do we do?" logic in one place.

jameslamb (changes requested):
I'm glad we're able to tighten this a bit! I left a suggestion on an alternative way to handle the sizes differing by CUDA version, though.

ci/build_wheel_libcuvs.sh
gforsyth: yep, happy to make these changes!  I started along this path and got waylaid by overcomplicating it by attempting to handle passing bash arrays as arguments between files (tip, don't do this)

gforsyth:
Ready for another look @jameslamb !

ci/validate_wheel.sh
```diff
@@ -7,6 +7,25 @@ package_dir=$1
...
+            --max-allowed-size-compressed '1.0G'
+        )
+    else
+        PYDISTCHECK_ARGS=(
+            --max-allowed-size-compressed '875M'
+        )
+    fi
+else
+    PYDISTCHECK_ARGS=()
+fi
```
jameslamb, line 25: You'll need to also add this to the `pydistcheck` call a few lines down, like:

```shell
pydistcheck \
    --inspect \
    "${PYDISTCHECK_ARGS[@]}" \
    "$(echo "${wheel_dir_relative_path}"/*.whl)"
``

gforsyth: yep, flying fast and loose here

jameslamb: to quote School of Rock: "you're not hardcore unless you live hardcore" 🤘🏻

jameslamb (approved):
looks great, thanks!

PR #888, opened 2025-05-13
Expose ivf-flat centers to python/c

benfred:
Similar to #881 - also expose centers for ivf-flat as well as ivf-pq


tarang-jain (changes requested):

cpp/src/neighbors/ivf_flat_c.cpp
```diff
@@ -139,6 +139,36 @@ void _extend(cuvsResources_t res,
...
+{
+  auto res_ptr   = reinterpret_cast<raft::resources*>(res);
+  auto index_ptr = reinterpret_cast<cuvs::neighbors::ivf_flat::index<T, IdxT>*>(index.addr);
+  auto dst       = cuvs::core::from_dlpack<output_mdspan_type>(centers);
+  auto src       = index_ptr->centers();
+
+  RAFT_EXPECTS(src.extent(0) == dst.extent(0), "Output centers has incorrect number of rows");
+  RAFT_EXPECTS(src.extent(1) == dst.extent(1), "Output centers has incorrect number of cols");
+
+  cudaMemcpyAsync(dst.data_handle(),
```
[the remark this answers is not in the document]
benfred: sure! I updated in the last commit to use the `raft::copy` from `raft/util/cudart_utils.hpp` .

(Fwiw, I originally tried to use the `copy` functions from `raft/core/copy.hpp` and `raft/core/copy.cuh` but couldn't - since we can't use the '.cuh' version inside the C-api, and the `.hpp` version was complaining about needing a cuda kernel for the D2H copy iirc, which is why I was using cudaMemCpyAsync directly here).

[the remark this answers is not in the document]
tarang-jain: I am seeing that `nn_descent_c.cpp` also uses `cudaMemcpyAsync`. If there is a reason we have avoided the cudart header, we can stick to `cudaMemcpyAsync`. My understanding was that having the `cudaMemcpyAsync` call means that the file would have to be compiled with nvcc anyway so we should be reusing raft functions instead.
(cc @cjnolet)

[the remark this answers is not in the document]
cjnolet: >  My understanding was that having the cudaMemcpyAsync call means that the file would have to be compiled with nvcc anyway 

@tarang-jain `cudaMemCpyXX()` is part of the CUDA runtime API, so it should not require `nvcc` to compile. Only the lower-level device function routines will need `nvcc`. Otherwise, it's just linking against the pre-compiled CUDA routines in the runtime API (kind of similar to what end-users do when they use cuVS C/C++ APIs).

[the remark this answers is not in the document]
benfred: I updated nn_descent_c.cpp in the last commit to use raft::copy.

the `cudaMemcpyAsync` code is fine (as is copy functions from `raft/util/cudart_utils.hpp`) - I just couldn't use this code https://github.com/rapidsai/raft/blob/c2dc3124ce3fbcb5ff2ccabd88d7f57570b6aea9/cpp/include/raft/core/copy.cuh#L57-L61  from raft . 

fwiw, one nice thing about using raft::copy is that it fixes one issue that this code used to have (wasn't checking the return value from cudaMemCpyAsync , which was a stupid oversight on my part =) ).

cpp/src/neighbors/ivf_pq/ivf_pq_build_common.cu
```diff
@@ -311,6 +311,27 @@ void extract_centers(raft::resources const& res,
                                   stream));
 }
 
+void extract_centers(raft::resources const& res,
```
tarang-jain, line 314: Since we need both -- the host and device mdspan functions for `extract_centers`, we can have a pointer based function in the detail namespace, rather than duplicating the function for host mdspan. Then the host and device mdspan functions in the src file can call that pointer based function.

cjnolet: I don't think we are going to be able to avoid this distinction here, and it's actually very low impact here to have the branched logic (in general, we want to move away from pointer-based functions across the codebase, not just in the public APIs... but we started w/ the public APIs). Not all device functions are super expensive.

tarang-jain: I am concerned that the content of the function is an exact copy except the function signature. In some other places we have had public APIs that are agnostic of what mdspan is being used (so the function argument is`raft::mdspan<const T, raft::matrix_extent<int64_t>, raft::row_major, data_accessor>`)

benfred: I changed in the last commit to reduce the cut and paste code here (not using a pointer based api - instead just using a template version on the output mdspan type).  lmk what you all think

tarang-jain: On the other hand, if we are okay with the two functions because it is not too expensive, that is also fine :))

cjnolet: > I am concerned that the content of the function is an exact copy except the function signature. In some other places we have had public APIs that are agnostic of what mdspan is being used (so the function argument israft::mdspan<const T, raft::matrix_extent<int64_t>, raft::row_major, data_accessor>)

I'm in favor of accepting general `mdspan`. Not in favor of accepting raw pointers. Raw pointers are prone to too much error- it would be a giant step backwards for our code.

benfred: this commit here is the one I meant https://github.com/rapidsai/cuvs/pull/888/commits/7080e2628b61d8c6fe6e2e4621b00ef2cad4f292 (takes a general mdspan )

tarang-jain: Looks good. Just one nit pick (feel free to ignore since it is not that important). We can move detail namespace stuff to `ivf_pq_build.cuh`

benfred: > Looks good. Just one nit pick (feel free to ignore since it is not that important). We can move detail namespace stuff to `ivf_pq_build.cuh`

moved in ed474d1

tarang-jain (approved):

mythrocks (commented):
Some trivial nitpicks.  Still coming to terms with the code.

The C++ side looks good to my eye.  +1 non-binding.

PR #890, opened 2025-05-13
CUVS_ANN_BENCH_SINGLE_EXE: find the mg library

achirkin:
Let the benchmark in CUVS_ANN_BENCH_SINGLE_EXE mode find the library for multi-gpu implementations.

Originally, all library names coincided with the generated target (library) names, which was used by the loader to find a proper implementation for an algorithm in the benchmark config.
A group of cuVS mg algorithms broke this pattern by using a single target for all cuVS mg algorithms. This PR adds this exception for the library discovery logic.


tfeher (approved):
Thanks for the fix Artem, LGTM.

PR #899, opened 2025-05-14
CAGRA Build + DiskANN Search cuvs-bench Wrapper

tarang-jain:
cuvs-bench wrapper for CAGRA build + Vamana style serialize. Search the index as a Diskann in-memory index.

This PR also brings a bunch of fixes for already existing algos:
- `cuvs_vamana` wrapper cleanup via default destructor gives SIGFPE (fixed by declaring raft handle first, so that it is destroyed last)
- Low throughput for diskann in-memory and ssd indexes (fixed via removing OpenMP reliance entirely during search and using the default benchmarking methods for multi-threading)
- `cuvs_vamana` was using the incorrect template params for device mdspan for the dataset
- DiskANN indexes to load the index using the correct number of threads using the `benchmark_n_threads` variable
- Faiss CAGRA + HNSW wrapper supports throughput mode search by resetting the CAGRA index `shared_ptr`, which is not used during search (this allows using different CPU threads)


python/cuvs_bench/cuvs_bench/config/algos/diskann_memory.yaml
```diff
@@ -4,10 +4,10 @@ constraints:
...
   base:
     build:
-      R: [64, 96]
+      R: [32, 64, 96]
       L_build: [128, 256, 384]
       alpha: [1.2]
       num_threads: [32]
     search:
-      num_threads: [32]
+      num_threads: [1]
```
cjnolet: @tarang-jain I'm not sure how I missed this one before, but this is different than all of the other algorithms we support. How does this work today? Are you using openmp for this? We should not be doing this differently than how we're doing this for all the other index types.

tarang-jain: We are using openmp for this yes. DiskANN does not expose any API to search a batch of queries. Rather they provide an API to search a single query. I am using that search API with OpenMP (DiskANN also does the same thing when supplied with a batch of n queries https://github.com/microsoft/DiskANN/blob/1f08328e25f2476286352bb93de19c158aef1fea/apps/search_memory_index.cpp#L162). I believe we are also doing the same thing with hnswlib benchmarks.
I see your point that it is different from other algos in that the other algos execute a batch of queries using one thread in latency mode. I can think of a way to do the same thing here.

tarang-jain: Got rid of the openmp stuff. Marking this as resolved!

cjnolet (approved):

Issue #698, opened 2025-02-14
cuvs-java: Support providing indexing data off-heap


ChrisHegarty:
Index builders allow to set the input vector dataset (to index) as a `float[][]`. This requires that all vectors be reified in the Java heap as a single contiguous block of memory. This can be expensive and may exhaust the java heap. Usage example:

```
float[][] dataset = ...

CagraIndex.newBuilder(resources)
  .withDataset(dataset)
```

For the Lucene use case, we write the raw vectors to disk and mmap them. To index them, then we need to copy them back into that Java heap as a `float[][]`. This is not ideal.  It would be better to allow the index builder to set the dataset as an opaque pointer + length.

mythrocks:
Should this be closed, now that https://github.com/rapidsai/cuvs/pull/902 has been merged?

ldematte:
> Should this be closed, now that https://github.com/rapidsai/cuvs/pull/902 has been merged?

I think https://github.com/rapidsai/cuvs/pull/902 is a step in that direction, but I think we'll also need a Dataset (implementation) that can be constructed from one (or N?) MemorySegments, and/or from just a tuple of longs (address + size). @ChrisHegarty wdyt?

I can pick this up next if you like!

ChrisHegarty:
> > Should this be closed, now that [#902](https://github.com/rapidsai/cuvs/pull/902) has been merged?
> 
> I think [#902](https://github.com/rapidsai/cuvs/pull/902) is a step in that direction, but I think we'll also need a Dataset (implementation) that can be constructed from one (or N?) MemorySegments, and/or from just a tuple of longs (address + size). [@ChrisHegarty](https://github.com/ChrisHegarty) wdyt?
> 
> I can pick this up next if you like!

@ldematte that is exactly what I was thinking also. Maybe hack together something around MemorySegments, to determine potential perf benefit.  For now, the API uses Java 21, where MemorySegment is preview, but we can address that separately later if the benefits prove significant. So just, use Java 22 for now, to get perf numbers.

ldematte:
👍 

Do we already have something like benchmarks to asses if there is a performance benefit? Or should I add something like a simple JMH with realistic vector sizes from float[][] and from MemorySegment directly?

ldematte:
RE: Java 21 vs 22 

>  from just a tuple of longs (address + size)

If I keep this in the interface, this is probably OK? We pay for a new MemorySegment, but no additional memcpy or allocation

[...]

PR #902, opened 2025-05-15
[Java] New off-heap Dataset support for CAGRA and Bruteforce

chatman:
As reported in #698, current `withDataset(float[][] arr)` requires the entire dataset to be copied in heap first, before writing out the MemorySegment for it.

Introducing a new `Dataset` (interface and impl) support with a `addVector(float[] vector)` support for adding the vectors into the MemorySegment one by one, without needing to load them all at once.

### DIFF
--- a/java/cuvs-java/src/main/java/com/nvidia/cuvs/CagraIndex.java
+++ b/java/cuvs-java/src/main/java/com/nvidia/cuvs/CagraIndex.java
@@ -21,6 +21,7 @@ import java.io.OutputStream;
 import java.nio.file.Path;
 import java.util.Objects;
 
+import com.nvidia.cuvs.BruteForceIndex.Builder;
 import com.nvidia.cuvs.spi.CuVSProvider;
 
 /**
@@ -184,12 +185,20 @@ public interface CagraIndex {
         Builder from(InputStream inputStream);
 
         /**
-         * Sets the dataset for building the {@link CagraIndex}.
+         * Sets the dataset vectors for building the {@link CagraIndex}.
          *
-         * @param dataset a two-dimensional float array
+         * @param vectors a two-dimensional float array
          * @return an instance of this Builder
          */
-        Builder withDataset(float[][] dataset);
+        Builder withDataset(float[][] vectors);
+
+        /**
+         * Sets the dataset for building the {@link CagraIndex}.
+         *
+         * @param dataset a {@link Dataset} object containing the vectors
+         * @return an instance of this Builder
+         */
+        Builder withDataset(Dataset dataset);
 
         /**
          * Registers an instance of configured {@link CagraIndexParams} with this
--- a/java/cuvs-java/src/main/java/com/nvidia/cuvs/spi/CuVSProvider.java
+++ b/java/cuvs-java/src/main/java/com/nvidia/cuvs/spi/CuVSProvider.java
@@ -52,7 +52,7 @@ public interface CuVSProvider {
       throws Throwable;
 
   Dataset newDataset(int size, int dimensions) throws UnsupportedOperationException;
-  
+
   /** Creates a new BruteForceIndex Builder. */
   BruteForceIndex.Builder newBruteForceIndexBuilder(CuVSResources cuVSResources)
       throws UnsupportedOperationException;
--- a/java/cuvs-java/src/main/java/com/nvidia/cuvs/spi/CuVSProvider.java
+++ b/java/cuvs-java/src/main/java/com/nvidia/cuvs/spi/CuVSProvider.java
@@ -51,6 +51,7 @@ public interface CuVSProvider {
   CuVSResources newCuVSResources(Path tempDirectory)
       throws Throwable;
 
+  /** Create a {@link Dataset} instance **/
   Dataset newDataset(int size, int dimensions) throws UnsupportedOperationException;
 
   /** Creates a new BruteForceIndex Builder. */
--- a/java/cuvs-java/src/main/java/com/nvidia/cuvs/spi/CuVSProvider.java
+++ b/java/cuvs-java/src/main/java/com/nvidia/cuvs/spi/CuVSProvider.java
@@ -19,6 +19,7 @@ package com.nvidia.cuvs.spi;
 import com.nvidia.cuvs.BruteForceIndex;
 import com.nvidia.cuvs.CagraIndex;
 import com.nvidia.cuvs.CuVSResources;
+import com.nvidia.cuvs.Dataset;
 import com.nvidia.cuvs.HnswIndex;
 
 import java.nio.file.Path;
@@ -50,6 +51,8 @@ public interface CuVSProvider {
   CuVSResources newCuVSResources(Path tempDirectory)
       throws Throwable;
 
+  Dataset newDataset(int size, int dimensions) throws UnsupportedOperationException;
+  
   /** Creates a new BruteForceIndex Builder. */
   BruteForceIndex.Builder newBruteForceIndexBuilder(CuVSResources cuVSResources)
       throws UnsupportedOperationException;
--- a/java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/BruteForceIndexImpl.java
+++ b/java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/BruteForceIndexImpl.java
@@ -43,6 +43,7 @@ import com.nvidia.cuvs.BruteForceIndex;
 import com.nvidia.cuvs.BruteForceIndexParams;
 import com.nvidia.cuvs.BruteForceQuery;
 import com.nvidia.cuvs.CuVSResources;
+import com.nvidia.cuvs.Dataset;
 import com.nvidia.cuvs.SearchResults;
 import com.nvidia.cuvs.internal.common.Util;
 import com.nvidia.cuvs.internal.panama.cuvsBruteForceIndex;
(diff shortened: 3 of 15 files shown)


chatman:
FYI, @ChrisHegarty @narangvivek10.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/DatasetImpl.java
```diff
@@ -0,0 +1,50 @@
+package com.nvidia.cuvs.internal;
```
narangvivek10, line 1: Please add the license header.

java/cuvs-java/src/main/java/com/nvidia/cuvs/CagraIndex.java
```diff
@@ -183,13 +184,21 @@ interface Builder {
...
+         * @param vectors a two-dimensional float array
+         * @return an instance of this Builder
+         */
+        Builder withDataset(float[][] vectors);
+
         /**
          * Sets the dataset for building the {@link CagraIndex}.
          *
-         * @param dataset a two-dimensional float array
+         * @param dataset a {@link Dataset} object containing the vectors
```
narangvivek10, line 198: Maybe also add a note here that this takes precedence (and the other is ignored internally) when both - the dataset and vectors are set.

narangvivek10:
Also, maybe update the title to `[Java] New off-heap Dataset support for CAGRA and Bruteforce`. Thanks!

java/cuvs-java/src/test/java/com/nvidia/cuvs/BruteForceRandomizedIT.java
```diff
@@ -77,7 +76,7 @@ private void tmpResultsTopKWithRandomValues() throws Throwable {
     }
 
     // Generate a random dataset
-    float[][] dataset = generateData(random, datasetSize, dimensions);
+    float[][] vectors = generateData(random, datasetSize, dimensions);
```
punAhuja, line 79: Similar variable change from "dataset" to "vectors" should also be done in the example code in these files:
examples/src/main/java/com/nvidia/cuvs/examples/CagraExample.java
examples/src/main/java/com/nvidia/cuvs/examples/BruteForceExample.java

chatman:
@narangvivek10 @punAhuja Incorporated your feedback, thanks! Can you approve the PR?
@cjnolet I think this is ready for CI testing/merge now.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/DatasetImpl.java
```diff
@@ -0,0 +1,68 @@
...
+  protected final MemorySegment seg;
+  private final int size;
+  private final int dimensions;
+  private int current = 0;
+
+  public DatasetImpl(int size, int dimensions) {
+    this.size = size;
+    this.dimensions = dimensions;
+
+    MemoryLayout dataMemoryLayout = MemoryLayout.sequenceLayout(size * dimensions, C_FLOAT);
```
mythrocks, line 38: Pardon my ignorance, but I should ask for my own understanding: 
Is there a material difference between a `SequenceLayout(size * dimensions)` and a `SequenceLayout(size, SequenceLayout(dimensions))`?

java/cuvs-java/src/main/java22/com/nvidia/cuvs/spi/JDKProvider.java
```diff
@@ -57,4 +59,9 @@ public CagraIndex.Builder newCagraIndexBuilder(CuVSResources cuVSResources) {
   public HnswIndex.Builder newHnswIndexBuilder(CuVSResources cuVSResources) {
     return HnswIndexImpl.newBuilder(Objects.requireNonNull(cuVSResources));
   }
+
+  @Override
+  public Dataset newDataset(int size, int dimensions) throws UnsupportedOperationException {
+	  return new DatasetImpl(size, dimensions);
```
mythrocks, line 65: Nit:  The indentation seems to be off a little here.  More like 6 spaces than 2.
```suggestion
      return new DatasetImpl(size, dimensions);
```

java/cuvs-java/src/test/java/com/nvidia/cuvs/BruteForceRandomizedIT.java
```diff
@@ -98,31 +97,44 @@ private void tmpResultsTopKWithRandomValues() throws Throwable {
...
       BruteForceIndexParams indexParams = new BruteForceIndexParams.Builder()
           .withNumWriterThreads(32)
           .build();
 
-      BruteForceIndex index = BruteForceIndex.newBuilder(resources)
-          .withDataset(dataset)
-          .withIndexParams(indexParams)
-          .build();
+      BruteForceIndex index;
+      boolean useNativeMemoryDataset = random.nextBoolean();
```
mythrocks, line 122: I'm not sure where I stand on conditionally testing with `Dataset`.  Shouldn't we test both ways?

java/cuvs-java/src/test/java/com/nvidia/cuvs/CagraRandomizedIT.java
```diff
@@ -85,23 +85,37 @@ private void tmpResultsTopKWithRandomValues() throws Throwable {
...
       CagraIndexParams indexParams = new CagraIndexParams.Builder()
           .withCagraGraphBuildAlgo(CagraGraphBuildAlgo.NN_DESCENT)
           .build();
-      CagraIndex index = CagraIndex.newBuilder(resources)
-          .withDataset(dataset)
-          .withIndexParams(indexParams)
-          .build();
+
+      CagraIndex index;
+      boolean useNativeMemoryDataset = random.nextBoolean();
```
mythrocks, line 103: Likewise here.  I wonder if we should test with both Vectors and with Datasets, one after the other.

Are we doing this to reduce test time?

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/BruteForceIndexImpl.java
```diff
@@ -284,7 +291,8 @@ public static BruteForceIndex.Builder newBuilder(CuVSResources cuvsResources) {
    */
   public static class Builder implements BruteForceIndex.Builder {
 
-    private float[][] dataset;
+    private float[][] vectors;
+    private Dataset dataset;
```
mythrocks, line 295: Do we foresee deprecating the `vectors` interface in the future, in favour of `Dataset`?

chatman:
@cjnolet Sincere apologies, I misclicked the upgrade branch button on this PR (instead of another PR), which stopped the tests that were running here.

Can you please trigger the CI again?

mythrocks (commented):
A couple of nits and clarifying questions only.  This looks mostly good to my eye.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/DatasetImpl.java
chatman: Not sure myself, I just went by what @narangvivek10 did in some other class.

java/cuvs-java/src/test/java/com/nvidia/cuvs/BruteForceRandomizedIT.java
chatman: The idea with randomized testing was to avoid loops to test for many variables, and run the test many times over and over to allow it to catch bugs. For this particular feature (Dataset vs float[][]), I think it makes sense to test for both. Fixing, thanks!

java/cuvs-java/src/main/java/com/nvidia/cuvs/spi/CuVSProvider.java
```diff
@@ -50,6 +51,8 @@ default Path nativeLibraryPath() {
   CuVSResources newCuVSResources(Path tempDirectory)
       throws Throwable;
 
+  Dataset newDataset(int size, int dimensions) throws UnsupportedOperationException;
```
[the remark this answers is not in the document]
chatman: Fixed, thanks.

chatman:
Thanks @mythrocks , incorporated your suggestions.

I think we need to add a style check for the Java project and standardize tabs vs spaces in a separate PR. This codebase is a mix right now :-(

java/cuvs-java/src/test/java/com/nvidia/cuvs/CagraRandomizedIT.java
chatman: Yes, wanted to reduce the test time by making it randomized. But, it feels important enough to test twice.

chatman:
@mythrocks can you please trigger a CI run?

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/BruteForceIndexImpl.java
ldematte: Very late here, but did you consider writing a second DatasetImpl to hold the float[][]? This way you do not need to keep around two fields, of which one is always null, and if between them every time in the various Index implementations. It would be easier to maintain.

I can draft a quick PR to show what I mean if you are interested.

ldematte: So, externally you can provide data either as a float[][] or a Dataset, but internally it will just be a Dataset.

chatman:
That's a good idea. Please feel free to rig up a PR. Thanks!

PR #908, opened 2025-05-19
ANN_BENCH: Avoid repeated calls to raft::get_device_for_address in CAGRA search

achirkin:
Cache the information about the IO buffer location between the benchmark search iterations.
This helps to avoid one extra CUDA context call per benchmark iteration, which otherwise adds a little bit of overhead under many-threads-throughput mode.


tfeher (approved):
Thanks Artem, LGTM.

PR #911, opened 2025-05-20
[ANN_BENCH] Fix OpenMP Sections in Wrappers

tarang-jain:
Fixes OpenMP num thread setting + dispatching search workloads in DiskANN and Faiss CPU cuvs-bench wrappers.
Hnswlib wrapper taken care of in https://github.com/rapidsai/cuvs/pull/724


tarang-jain:
@achirkin @tfeher for viz. Removed number of search threads entirely because user can control it with OpenMP environment variables, which is actually the most correct way of using OpenMP rather than passing in process-specific threads through cuvs-bench. We simply call Faiss' search APIs with the batch of queries and Faiss does all its parallelization internally.

divyegala:
@tarang-jain can you pull #724 into this PR?

divyegala (changes requested):
@tarang-jain can you pull https://github.com/rapidsai/cuvs/pull/724 into this PR?

tarang-jain:
@divyegala #724 has build failures in CI at the moment.

Issue #809, opened 2025-04-07
[FEA] CAGRA C and Python API Missing Options / Features


tarang-jain:
CAGRA python and C API needs to be updated to include new features. There were new features added on the C++ side:
- `guarantee_connectivity`
- attach_dataset_on_build (pointed out by #806) 
- `graph_build_params` struct to pass custom IVF-PQ and NN Descent params

tarang-jain:
The third one is done (`graph_build_params`) but the first two are still pending.

PR #918, opened 2025-05-21
Add IVF-PQ support in CAGRA index params in Python

### LINKED ISSUE #468: [FEA] Allow customization of IVF-PQ build/search params in CAGRA params in Python
Allow users to change the n_probe / n_list of the IVF-PQ params when building a CAGRA index through the Python API.

lowener:
- Updated CAGRA Index Params to include support for IVF-PQ parameters.
- Added `coarse_search_dtype` in IVFPQ params
- Added tests for IVF-PQ integration in `test_cagra.py`.

Closes #468
Related PR: #809


python/cuvs/cuvs/neighbors/cagra/cagra.pxd
```diff
@@ -32,6 +32,47 @@ from cuvs.distance_type cimport cuvsDistanceType
...
+
+    ctypedef struct cuvsIvfPqSearchParams:
+        uint32_t n_probes
+        cudaDataType_t lut_dtype
+        cudaDataType_t internal_distance_dtype
+        double preferred_shmem_carveout
+        cudaDataType_t coarse_search_dtype
+        uint32_t max_internal_batch_size
+
+    ctypedef cuvsIvfPqSearchParams* cuvsIvfPqSearchParams_t
```
benfred, line 74: Rather than redefine these here can we cimport them from the ivf_pq definitions? Something like 
```suggestion
from cuvs.neighbors.ivf_pq cimport codebook_gen, cuvsIvfPqIndexParams, cuvsIvfPqIndexParams_t, cuvsIvfPqSearchParams, cuvsIvfPqSearchParams_t
```

benfred (commented):
looks good! one minor comment:

python/cuvs/cuvs/neighbors/cagra/cagra.pxd
```diff
@@ -100,6 +120,10 @@ cdef extern from "cuvs/neighbors/cagra.h" nogil:
     cuvsError_t cuvsCagraCompressionParamsDestroy(
         cuvsCagraCompressionParams_t index)
 
+    cuvsError_t cuvsIvfPqParamsCreate(cuvsIvfPqParams_t* params)
+
+    cuvsError_t cuvsIvfPqParamsDestroy(cuvsIvfPqParams_t index)
+
```
benfred, line 126: I don't think these functions are used (or exist in the c-api) 
```suggestion
```

python/cuvs/cuvs/neighbors/cagra/cagra.pyx
```diff
@@ -177,7 +189,10 @@ cdef class IndexParams:
                  graph_degree=64,
                  build_algo="ivf_pq",
                  nn_descent_niter=20,
-                 compression=None):
+                 compression=None,
+                 ivf_pq_build_params: ivf_pq.IndexParams = None,
```
benfred, line 193: I like the type annotations! We should do this everywhere =)  (but not for the 25.06 release)

python/cuvs/cuvs/neighbors/cagra/cagra.pyx
```diff
@@ -219,6 +247,10 @@ cdef class IndexParams:
     def nn_descent_niter(self):
         return self.params.nn_descent_niter
 
+    @property
+    def refinement_rate(self):
+        return self.params.graph_build_params.refinement_rate
+
```
benfred, line 253: maybe we should add getters for the ivf_pq_search_params/ivf_pq_build_params too ? 
```python
    @property
    def ivf_pq_search_params(self):
        return self.ivf_pq_search_params

    @property
    def ivf_pq_build_params(self):
        return self.ivf_pq_build_params
```

python/cuvs/cuvs/neighbors/cagra/cagra.pxd
```diff
@@ -49,13 +62,20 @@ cdef extern from "cuvs/neighbors/cagra.h" nogil:
 
     ctypedef cuvsCagraCompressionParams* cuvsCagraCompressionParams_t
 
+    ctypedef struct cuvsIvfPqParams:
```
benfred, line 65: I know that this PR didn't introduce this struct - but I think that we should rename this to `cuvsCagraIvfPqParams` since this is for cagra only (and could be confusing for users of ivf-pq). @lowener can you create an issue to change this for 25.08? I think its a bit late to try to get this in for 25.06

benfred (commented):
this is looking good! some minor comments below:

cjnolet (approved):

python/cuvs/cuvs/neighbors/cagra/cagra.pxd
lowener: The all neighbor API recently introduced it's own version of this struct: https://github.com/jinsolp/cuvs/blob/branch-25.06/cpp/include/cuvs/neighbors/all_neighbors.hpp#L37.
I think we can keep the name and factorize both struct into `ivf_pq.hpp`

Issue #883, opened 2025-05-07
[FEA] PW KNN Graph Function to Migrate from cuVS


tarang-jain:
Distance computation calls `raft::sparse::neighbors::knn_graph` which should be moved to cuVS.

PR #927, opened 2025-05-22
[FEA] Use Native Brute Force for Sparse Pairwise KNN

tarang-jain:
Bring in `knn_graph` sparse APIs from raft and replace calls to it with cuVS functions.
Closes https://github.com/rapidsai/cuvs/issues/883

Binary size reduction:
branch-25.06 libcuvs.so size = 1403.169 MB
This PR: libcuvs.so size = 1375.059 MB


cpp/src/neighbors/detail/knn_graph.cuh
```diff
@@ -0,0 +1,106 @@
...
+  return std::min(n_samples, std::max((value_idx)2, (value_idx)floor(raft::log2(n_samples)) + c));
+}
+/**
+ * Constructs a (symmetrized) knn graph edge list from
+ * dense input vectors.
+ *
+ * Note: The resulting KNN graph is not guaranteed to be connected.
+ *
+ * @tparam value_idx
+ * @tparam value_t
```
jinsolp, line 47:
```suggestion
 * @tparam value_t
 * @tparam nnz_t
```

examples/cpp/src/vamana_example.cu
```diff
@@ -14,12 +14,13 @@
  * limitations under the License.
  */
 
+#include "../../../cpp/src/sparse/neighbors/cross_component_nn.cuh"
```
jinsolp, line 17: I'm not sure if we end up using anything in here?

jinsolp (commented):
Thank you for this PR Tarang, LGTM! Small suggestions;

cpp/src/neighbors/detail/knn_graph.cuh
tarang-jain: Done

jinsolp (approved):

Issue #904, opened 2025-05-16
[BUG] mg_index doesn't write output values when used on single GPU with default parameters


achirkin:
When built for a single GPU, mg_index doesn't write the outputs during search with default parameters.

The culprit is the sharded-tree-merge search implementation that is designed for even number of GPUs.
With `rank == 1` (single GPU) the code for copying the data to output is never reached:
https://github.com/rapidsai/cuvs/blob/2db85161fb59ce196c1cd03634622d41d4d60650/cpp/src/neighbors/mg/snmg.cuh#L427-L438
...because it's nested under the following loop:
https://github.com/rapidsai/cuvs/blob/2db85161fb59ce196c1cd03634622d41d4d60650/cpp/src/neighbors/mg/snmg.cuh#L367-L370

That is, if one builds and searches an SNMG index using ANN_BENCH tool with default parameters and a single GPU, the recall is always zero.

viclafargue:
Thanks for catching this, @achirkin. It appears that tree merge was chosen as the default behavior, leaving it up to the user to opt for direct merge when deploying with an odd number of GPUs. As a workaround, we could automatically detect when tree merge is used with an odd GPU count and fall back to direct merge, accompanied by an informative message.

viclafargue:
Upon closer examination, I found that the odd‐GPU scenario is already handled correctly: the lone GPU without a merge partner simply skips the first merge step (see [condition in the code](https://github.com/rapidsai/cuvs/blob/8cc25c3f45324e5eb20291d72647faa2b112be8d/cpp/src/neighbors/mg/snmg.cuh#L377)). Further testing revealed that the only remaining issue occurs when running a sharded search on a single GPU. I’ve opened a PR to address this specific edge case : https://github.com/rapidsai/cuvs/pull/1094.

PR #934, opened 2025-05-26
Adding check to ANN MG tree merge mode

viclafargue:
Answers https://github.com/rapidsai/cuvs/issues/904


cjnolet (approved):

cpp/src/neighbors/mg/snmg.cuh
```diff
@@ -605,6 +605,9 @@ void search(const raft::resources& clique,
                                        n_neighbors,
                                        n_batches);
     } else if (merge_mode == TREE_MERGE) {
+      ASSERT(index.num_ranks_ % 2 == 0,
```
cjnolet, line 621: We discussed this briefly in our meeting, but I think we should find a better alternative here in the short term. I don't mind merging this change so that users at least have a warning / reason instead of segfaulting, but I don't think having an even number of GPUs is a fair requirement to levvy on our users. At the very least, we should be able to pair off the reductions using P2P, where there may be 1 GPU that only contributes in the final merge stage.

cjnolet: Can you create an issue to follow-up with this (and link the issue in a comment in the code here so that any developers or users browsing the code know the plan)?

viclafargue: Sure, created an issue over here : https://github.com/rapidsai/cuvs/issues/979. Just to be clear, SNMG ANN sharded indexes can still be searched with an odd GPU count with the `MERGE_ON_ROOT_RANK` mode.

viclafargue:
This PR is actually not necessary. See my comment https://github.com/rapidsai/cuvs/issues/904#issuecomment-3052504730. Closing it.

PR #936, opened 2025-05-27
Add conversion functions for`faiss::IDSelector`

lowener:
This PR add utility fonctions to convert a `faiss::IDSelector` into a `cuvs::core::bitset_view`.


cpp/src/util/filters.cu
```diff
@@ -0,0 +1,181 @@
...
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#include <cuvs/core/bitmap.hpp>
+#include <cuvs/core/bitset.hpp>
+#include <faiss/impl/IDSelector.h>
```
cjnolet, line 19: I don't think we want to be using faiss types in cuVS. That would introduce a circular dependency. Instead, we should implement the conversion logic in FAISS itself.

PR #944, opened 2025-05-27
All-neighbors API docs

jinsolp:
docs for all-neighbors API

### DIFF
--- a/cpp/include/cuvs/cluster/kmeans.hpp
+++ b/cpp/include/cuvs/cluster/kmeans.hpp
@@ -519,9 +519,25 @@ void predict(raft::resources const& handle,
              raft::device_matrix_view<const float, int> X,
              std::optional<raft::device_vector_view<const float, int>> sample_weight,
              raft::device_matrix_view<const float, int> centroids,
-             raft::device_vector_view<int, int> labels,
              bool normalize_weight,
+             raft::device_vector_view<int, int> labels,
              raft::host_scalar_view<float> inertia);
+
+// This overload is retained for backward compatibility.
+[[deprecated(
+  "The argument order of kmeans::predict has been corrected. Please use the new function "
+  "instead.")]]
+inline void predict(raft::resources const& handle,
+                    const kmeans::params& params,
+                    raft::device_matrix_view<const float, int> X,
+                    std::optional<raft::device_vector_view<const float, int>> sample_weight,
+                    raft::device_matrix_view<const float, int> centroids,
+                    raft::device_vector_view<int, int> labels,
+                    bool normalize_weight,
+                    raft::host_scalar_view<float> inertia)
+{
+  predict(handle, params, X, sample_weight, centroids, normalize_weight, labels, inertia);
+}
 
 /**
  * @brief Predict the closest cluster each sample in X belongs to.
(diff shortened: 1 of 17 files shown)


cpp/include/cuvs/neighbors/all_neighbors.hpp
```diff
@@ -74,35 +81,37 @@ struct all_neighbors_params {
...
-   *
-   * Larger n_nearest_clusters results in better accuracy of the final all-neighbors knn
-   * graph. E.g. With the similar device memory usages, (n_nearest_clusters / n_clusters) = 4/20
-   * will have better accuracy than 2/10 at the cost of performance.
+   * number of nearest clusters each data point will be assigned to in the batching algorithm
+   */
+  size_t n_nearest_clusters = 2;
+
+  /**
+   * number of total clusters (aka batches) to split the data into. If set to 1, algorithm creates
```
cjnolet, line 89: Can we provide some light guidance on how one would select this value?

cpp/include/cuvs/neighbors/all_neighbors.hpp
```diff
@@ -43,17 +47,20 @@ using GraphBuildParams =
...
  * training vectors)
  *
- * graph_build_params: graph building parameters for the given graph building algorithm. defaults
- * to ivfpq.
- * n_nearest_clusters: number of nearest clusters each data point will be assigned to in
- * the batching algorithm
- * n_clusters: number of total clusters (aka batches) to split the data into. If set to 1, algorithm
- * creates an all-neighbors graph without batching
- * metric: metric type
+ * Usage of n_nearest_clusters and n_clusters
```
cjnolet, line 53: Let's keep the individual parameters documented in their own spots below (note that can you still reference a formula that might use multiple parameters without having to couple the docs for them tightly together. 

The  descrtion/doc for the `all_neighbors_params` struct should give a very brief (high-level) introduction of what the struct is for. Highlighting individual parameters is too low-level. You could give a very brief overview to the effect of: "For scalability, the all-neighbors graph construction method partitions a set of training vectors into overlapping clusters, computes a local knn graph on each cluster, and merges the local graphs into a single global graph. The algorithm used to build each local graph is configurable."

Then you can lead into the `graph_build_algo` with something to the effect of:
"Approximate nearest neighbors methods are (or can be) used to build the knn graph, which means they'll each have their own trade-offs between the speed and the quality of the final graph. The current options are 'IVF-PQ' and 'NN-Descent'....."

I think n_clusters and n_nearest_clusters is very confusing. I know this changes the API, but I think we should get this right up front before we have too many users of the API- could we maybe prefer the names `total_clusters` and `overlap_factor`? Can we somehow map the `n_nearest_clusters` into a degree of overlap? So for example, 1 would be no overlap (completely disjoint clusters), 2 would be 2x overlap (and thus 2x duplication). The reason I think this matters is because it gets users thinking about their parameterization more in terms of how they'll practically use the algorithm, rather than having to understand how the algorithm itself works. Even though they fundamentall mean the same thing, I personally can reason more about the global effects of "I want to partition my data into 10 roughly even-sized clusters, and I'm okay with a 3x duplication rate" than having to think more locally about each point, "I want to partition my data into 10 roughly even-sized clusters and each data point will be placed in its 3 closest clusters".

jinsolp: I think that's a great idea. `n_nearest_clusters` is in essence an overlap factor. I think all we will have to do is just change the name of it and how we explain it.

jinsolp: But for `n_clusters` I prefer using it as-is instead of `total_clusters` since the term `n_clusters` is already being used in cuML (as `nnd_n_clusters`)

PR #949, opened 2025-05-28
Common knn graph build params

jinsolp:
Reusing graph build params for cagra and all neighbors.


cpp/include/cuvs/neighbors/graph_build_types.hpp
```diff
@@ -0,0 +1,78 @@
+/*
```
cjnolet, line 1: cuVS isn't header-only, so the header files are very lightweight. Instead of introducing new headers for structs and enums, can we just reuse common.hpp?

jinsolp: `common.hpp` has the base `index_params` ([link](https://github.com/rapidsai/cuvs/blob/e8dbb887cf3a9646ed20dcf4b749528ecd3e15e6/cpp/include/cuvs/neighbors/common.hpp#L90-L95)), so declaring `graph_build_params` in that file results in a circular dependency (because we make use of `ivf_pq::index_params` and `nn_descent::index_params` inside `graph_build_params)

cpp/include/cuvs/neighbors/graph_build_types.hpp
```diff
@@ -0,0 +1,78 @@
...
+
+enum GRAPH_BUILD_ALGO { BRUTE_FORCE = 0, IVF_PQ = 1, NN_DESCENT = 1 };
+
+namespace graph_build_params {
+
+/** Specialized parameters utilizing IVF-PQ to build knn graph */
+struct ivf_pq_params {
+  cuvs::neighbors::ivf_pq::index_params build_params;
+  cuvs::neighbors::ivf_pq::search_params search_params;
+  float refinement_rate = 2.0;
```
lowener: Is the default value 1 or 2?

jinsolp: That's a bit tricky here... cagra uses the `ivf_pq_params()` constructor and expects a default refinement rate of 1.0, whereas all neighbors expects a default refinement rate of 2.0.

I agree that this is confusing. Maybe we should set the default to 1.0 and make changes from the all neighbors usage.

jinsolp:
@lowener Changed defaults. Ready for another round of reviews : )

jinsolp:
This is ready to merge if there are no further suggestions!

Issue #895, opened 2025-05-14
[BUG] ANN_BENCH: the data/query memory types not checked in multi-gpu implementations


achirkin:
All multi-gpu index wrappers do not check the pointer types and wrap them into host mdarray views:
https://github.com/rapidsai/cuvs/blob/73e4fbcb981a67048a7f63b39416a87cc878c1ba/cpp/bench/ann/src/cuvs/cuvs_mg_ivf_pq_wrapper.h#L125-L138

The pointer locations are configurable in the benchmarks at runtime, hence we should check them and, at the very least, throw an error if the memory location is not supported.

Issue #894, opened 2025-05-14
[BUG] ANN_BENCH: incorrect output index type in multi-gpu CAGRA


achirkin:
All multi-gpu implementations in ANN_BENCH ignore the benchmark index type `algo_base::index_type`, which is `int64_t`:
  - https://github.com/rapidsai/cuvs/blob/73e4fbcb981a67048a7f63b39416a87cc878c1ba/cpp/bench/ann/src/cuvs/cuvs_mg_cagra_wrapper.h#L161-L162
  - https://github.com/rapidsai/cuvs/blob/73e4fbcb981a67048a7f63b39416a87cc878c1ba/cpp/bench/ann/src/cuvs/cuvs_mg_ivf_flat_wrapper.h#L134-L135
  - https://github.com/rapidsai/cuvs/blob/73e4fbcb981a67048a7f63b39416a87cc878c1ba/cpp/bench/ann/src/cuvs/cuvs_mg_ivf_pq_wrapper.h#L131-L132

IVF-Flat and IVF-PQ index types coincide with the benchmark index type and thus show correct results.
CAGRA index type is always instantiated as `uint32_t`. As a result, `int64_t` array is casted to `int32_t`, and the benchmark interprets the algorithm output incorrectly.

We either need to map the outputs correctly as it's done in [IVF-PQ benchmark wrapper](https://github.com/rapidsai/cuvs/blob/73e4fbcb981a67048a7f63b39416a87cc878c1ba/cpp/bench/ann/src/cuvs/cuvs_ivf_pq_wrapper.h#L181-L222) (which adds extra overhead if the types do not match and puts the index at disadvantage in the benchmarks) or adjust the implementation of multi-gpu indexes to [decouple index-index type from the search output index type](https://github.com/rapidsai/cuvs/blob/73e4fbcb981a67048a7f63b39416a87cc878c1ba/cpp/bench/ann/src/cuvs/cuvs_cagra_wrapper.h#L368-L369) as it was done for CAGRA in https://github.com/rapidsai/cuvs/pull/769 .

achirkin:
CC @viclafargue

viclafargue:
[...]

PR #975, opened 2025-06-04
Adding int64 search for MG CAGRA

viclafargue:
Answers https://github.com/rapidsai/cuvs/issues/894 and https://github.com/rapidsai/cuvs/issues/895 by exposing new MG CAGRA search functions with `int64_t` neighbors output type.
The PR also updates the tests and benchs to use this new search function.
Additionally, the tree merge feature used to perform the KNN merge operation in-place which was an issue, this PR fixes this.


divyegala:
@viclafargue what is the effect on binary size of this PR?

cc @cjnolet

viclafargue:
> @viclafargue what is the effect on binary size of this PR?

The binary size remained at 199M since this is just wrapper code, the `int64_t` CAGRA search is already compiled as part of the library.

cjnolet (approved):

PR #981, opened 2025-06-05
[java] Copy `pom.xml` to `target/` on build

### LINKED ISSUE #978: [BUG] Include `pom.xml` in java artifact upload
Please include the `pom.xml` file used to build the cuvs-java jars in the artifact uploaded here https://github.com/rapidsai/cuvs/blob/branch-25.08/.github/workflows/build.yaml#L79

This is required to fully publish the jars to maven central.

For `25.06`, Ops will manually handle this, but future releases should be include the pom so it is automatically picked up.

mythrocks:
Fixes #978.

This commit modifies the `java/build.sh` script to also copy the `pom.xml` to the `target/` directory.  This will allow all artifacts for publication to Maven (i.e. all the jars and the pom) to be fetched from a single location.


mythrocks:
(Fixing style, again.)

jameslamb (approved):

PR #983, opened 2025-06-05
[java] Utility function for `cudaMemcpy`

mythrocks:
This commit does a minor rephrasing of calls to `cudaMemCpy`.

Prior to this change:
  1. Calls to `cudaMemcpy` were using hardcoded / undocumented magic numbers for the direction of the copy. These have been replaced an enum that mirrors what is used in the CUDA Runtime call.
  2. `cudaMemcpy` would return an error value that needed to be checked / converted into a RuntimeError.  This has been moved into the utility function, to reduce the noise at the call site.

This change also includes some minor cleanup of unused imports in the modified files.


cjnolet (approved):

mythrocks:
(Another case of gtest failures.)

PR #990, opened 2025-06-06
Update rust/cuvs-sys

benfred:
We can't have access files from parent directories in the rust publish stage - so remove the `../../cmake/rapids_config.cmake` reference in the CMakelists, and replace with a symlink to the `cmake` directory.

Also add a check to do a dry-run of the cuvs-sys publish in rust CI that should catch these type of problems in the future


ci/build_rust.sh
```diff
@@ -33,3 +33,8 @@ export LIBCLANG_PATH
 echo "LIBCLANG_PATH=$LIBCLANG_PATH"
 
 bash ./build.sh rust
+
+# Also test out that we can publish cuvs-sys via a dry-run
+pushd ./rust/cuvs-sys
```
raydouglass, line 38: Should it also do a dry run in `./rust/cuvs`?

KyleFromNVIDIA (approved):

ci/build_rust.sh
benfred: I'd like to - but the issue is that `rust/cuvs` depends on `rust/cuvs-sys` - and the dry-run is going to complain that it can't find the appropriate version cuvs-sys on crates.io (since it hasn't been published for real yet) =(.

Like running the dry-run on cuvs will look something like:
```
$ cargo publish --dry-run -p cuvs
    Updating crates.io index
   Packaging cuvs v25.6.0 (/home/ben/code/cuvs/rust/cuvs)
    Updating crates.io index
error: failed to prepare local package for uploading

Caused by:
  failed to select a version for the requirement `cuvs-sys = "^25.6.0"`
  candidate versions found which didn't match: 25.4.0, 24.6.0
  location searched: crates.io index
  required by package `cuvs v25.6.0 (/home/ben/code/cuvs/rust/cuvs)`
```

cuvs-sys is the biggest problem for building here - its the only one to directly link to libcuvs.so and use cmake-rs etc - so running the dry-run only on it should catch these cargo publish issues we've been seeing

raydouglass (approved):

PR #996, opened 2025-06-10
Fix C compile error in `tiered_index.h`

mythrocks:
This commit rephrases the definition of the `cuvsTieredIndexANNAlgo` enum in `tiered_index.h` to use a `typedef`.

With the old definition `enum cuvsTieredIndexANNAlgo{}`, compiling a C translation unit yields errors where `cuvsTieredIndexANNAlgo` is used without `enum`. (For instance at line#55 and line#88 of `tiered_index.h`.)
```
tiered_index.h:53:3: error: must use 'enum' tag to refer to type 'cuvsTieredIndexANNAlgo'
tiered_index.h:88:3: error: must use 'enum' tag to refer to type 'cuvsTieredIndexANNAlgo'
```

The typedef sorts this error out.


mythrocks:
Note:  I ran into this when working on JNI bindings for `tiered_index.h`.  `jextract` chokes on the usage of `cuvsTieredIndexANNAlgo`.

benfred (commented):
good catch - thanks for fixing!

We've had this kind of error happen previously - and one thing we've added for this is a some basic tests to make sure that we're not including C++ features in our C header files. 

Maybe we should add a couple of lines to https://github.com/rapidsai/cuvs/blob/6d071162169b67d0c5bd528bf33e0daa35505d2a/cpp/tests/neighbors/c_api.c#L24-L26 to verify that this is still valid C code so that this doesn't happen again in the future?

mythrocks:
> Maybe we should add a couple of lines to...

ACK.  I've added C-API compile tests for `tiered_index.h`.  Please let me know if this doesn't look agreeable.

cpp/tests/neighbors/c_api.c
```diff
@@ -16,16 +16,53 @@
 
 #include <cuvs/core/c_api.h>
 #include <cuvs/neighbors/cagra.h>
+#include <cuvs/neighbors/tiered_index.h>
+
```
mythrocks, line 20: `pre-commit` doesn't seem to apply to this C file or header.  I've grouped the headers per the code conventions.  I'm happy to roll this change back if that's preferable.

mythrocks:
As an aside, what's the project convention on copyright dates on the code?  I just realized that I neglected to update them on the files I modified.  It appears that several source files are out of date.

cjnolet:
> As an aside, what's the project convention on copyright dates on the code? I just realized that I neglected to update them on the files I modified. It appears that several source files are out of date.

We have a copyright checker script that runs as part of the style checker (and the pre-commit should be updating the copyrights) but it looks like both failed in this case.

mythrocks:
I have updated the copyright dates.  I haven't gotten to the bottom of why the copyright date checks from pre-commit don't apply to these files.  With your permission, we'll sort that out later.

@benfred: Please let me know if this looks alright now.

benfred:
> I have updated the copyright dates. I haven't gotten to the bottom of why the copyright date checks from pre-commit don't apply to these files. With your permission, we'll sort that out later.

The copyright checker wasn't working  - I have a fix here https://github.com/rapidsai/cuvs/pull/1017

benfred (approved):
lgtm!

mythrocks:
(Re-kicking CI.  Some of the pytests seem to be failing.)

PR #997, opened 2025-06-10
Enable `build_sorted_mst` with data on host memory

jinsolp:
Connecting knn graph requires the entire data to be on device memory to use `cuvs::sparse::neighbors::cross_component_nn`.

This PR enables connecting knn graph with data on host by randomly choosing edges between components. Thus, it ends up selecting `n_components-1` number of edges and adds that to the msf to make an mst.


cpp/src/cluster/detail/mst.cuh
```diff
@@ -117,6 +118,149 @@ void connect_knn_graph(
...
+  std::random_device rd;
+  std::mt19937 gen(rd());
+  std::uniform_int_distribution<> dis;
+
+  auto pairwise_dist = raft::make_device_matrix<value_t, int64_t>(handle, 1, 1);
+  auto data_u        = raft::make_device_matrix<value_t, int64_t>(handle, 1, n);
+  auto data_v        = raft::make_device_matrix<value_t, int64_t>(handle, 1, n);
+
+  // connect i-1 component and i component
+  for (int i = 1; i < n_components; ++i) {
```
divyegala, line 174: How many components are there usually? If the number is on the larger side, this could benefit with some omp parallelism

cpp/src/cluster/detail/mst.cuh
```diff
@@ -117,6 +118,149 @@ void connect_knn_graph(
...
+    value_idx v = nodes_b[dis(gen)];
+
+    // calculate the distance between the two data vectors
+    raft::copy(data_u.data_handle(), X + u * n, n, stream);
+    raft::copy(data_v.data_handle(), X + v * n, n, stream);
+    cuvs::distance::pairwise_distance(handle,
+                                      raft::make_const_mdspan(data_u.view()),
+                                      raft::make_const_mdspan(data_v.view()),
+                                      pairwise_dist.view(),
+                                      metric);
```
divyegala: Based on your answer to my previous question, if using omp, I would also create new handles with streams picked from a pool for this call to saturate the GPU with concurrency (pairwise distance of 1 row at a time on the GPU will likely be slower than doing this on the CPU)

cpp/src/cluster/detail/mst.cuh
jinsolp: very small. Based on heuristic observations, it's usually somewhere between 1 ~50.

divyegala: Got it, if approaching 50, I would try to run this in parallel because we'll be launching pairwise distance kernels once per iteration.

jinsolp: Okay, so will this be a if-else omp launch?
Also, I just wanted to confirm that we would want to use a separate stream per omp thread so that we can launch kernels concurrently?

divyegala: Not an if-else, we can assign the number of threads based on some condition before the for loop (n_components / 5)?

And yes, 1 stream per thread.

jinsolp: @divyegala ,thinking of it, if make this an omp region, we have to have `data_u`, `data_v`, and `pairwise_dist` matrix allocation inside the for loop. do you think this will be too much of an overhead?

maybe just gathering all vectors into a matrix and calling the pariwise dist kernel once will be a better idea?

divyegala: Yes, gather first is 100% better. For some reason I thought doing this with a for loop was a requirement.

Memory overhead is only (n_components, n_cols) and (n_components, n_components) where n_components ~ 50.

cpp/src/cluster/detail/mst.cuh
```diff
@@ -117,6 +120,143 @@ void connect_knn_graph(
...
+  // Build remapping table so that colors are compact integers (i.e. consecutive colors)
+  for (size_t i = 0; i < m; ++i) {
+    if (color_remap.find(h_color[i]) == color_remap.end()) {
+      color_remap[h_color[i]] = new_label++;
+    }
+  }
+
+  for (size_t i = 0; i < m; ++i) {
+    h_color[i] = color_remap[h_color[i]];
+  }
```
divyegala: If my understanding of this operation is correct, could you use this primitive? https://github.com/rapidsai/raft/blob/9abfca9a3fc2fe7628775bc3c1b1b2cb247302c2/cpp/include/raft/label/classlabels.cuh#L113-L117

jinsolp: Thanks! didn't know we had this

divyegala (commented):
How are we testing this feature?

jinsolp:
I've checked that it works locally (stitched together with other things that I'm working on right now). Let me add a test for this specifically to make sure.

PR #999, opened 2025-06-10
IVF-PQ coarse search: fix integer overflow and avoid excessive batch sizes

achirkin:
When the IVF-PQ batch size and the number of cluster are big enough, the GEMM output matrix size may exceed the `uint32_t` limits. This manifests as an illegal memory access popping up in further stages of IVF-PQ (due to everything being as async as possible).

This PR fixes the bug by converting the input arguments to 64-bits before multiplying them.

On top of the overflow fix, this PR also limits the batch size when it may hurt performance: (1) within IVF-PQ to avoid out-of-memory error (2) in CAGRA graph build to avoid switching to the large memory resource (which is typically backed by managed memory).

### DIFF
--- a/cpp/src/neighbors/detail/cagra/cagra_build.cuh
+++ b/cpp/src/neighbors/detail/cagra/cagra_build.cuh
@@ -183,15 +183,21 @@ void build_knn_graph(
   auto free_space_ratio    = raft::resource::get_workspace_free_bytes(res) / desired_workspace_size;
   bool use_large_workspace = false;
   if (free_space_ratio < kMinWorkspaceRatio) {
-    if (max_queries * free_space_ratio >= kMinLargeBatchSize * kMinWorkspaceRatio) {
-      max_queries = static_cast<uint32_t>(max_queries * free_space_ratio / kMinWorkspaceRatio);
+    auto adjusted_max_queries =
+      static_cast<uint32_t>(max_queries * free_space_ratio / kMinWorkspaceRatio);
+    if (adjusted_max_queries >= kMinLargeBatchSize) {
+      // adjust max_queries, so that the ratio free_space_ratio gets not larger than
+      // kMinWorkspaceRatio.
       RAFT_LOG_INFO(
         "CAGRA graph build: reducing IVF-PQ search max_internal_batch_size from %u -> %u to fit "
         "the workspace",
-        pq.search_params.max_internal_batch_size,
-        max_queries);
-      pq.search_params.max_internal_batch_size = max_queries;
+        max_queries,
+        adjusted_max_queries);
+      max_queries                              = adjusted_max_queries;
+      pq.search_params.max_internal_batch_size = adjusted_max_queries;
     } else {
+      // adjusting max_queries to a very small value isn't practical, so we use the large workspace
+      // instead.
       use_large_workspace = true;
       RAFT_LOG_WARN(
         "Using large workspace memory for IVF-PQ search during CAGRA graph build. Desired "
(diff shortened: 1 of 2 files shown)


cpp/src/neighbors/detail/cagra/cagra_build.cuh
```diff
@@ -166,25 +166,45 @@ void build_knn_graph(
...
+                   + sizeof(float) * gpu_top_k        // distances
+                   + sizeof(int64_t) * gpu_top_k      // neighbors
+                   + sizeof(float) * top_k            // refined_distances
+                   + sizeof(int64_t) * top_k          // refined_neighbors
+                  );
+  auto free_space_ratio    = raft::resource::get_workspace_free_bytes(res) / desired_workspace_size;
+  bool use_large_workspace = false;
+  if (free_space_ratio < kMinWorkspaceRatio) {
+    if (max_queries * free_space_ratio >= kMinLargeBatchSize * kMinWorkspaceRatio) {
+      max_queries = static_cast<uint32_t>(max_queries * free_space_ratio / kMinWorkspaceRatio);
```
tfeher: Do I understand correctly, that here the new `max_queries` will be set so that the corresponding `free_space_ratio` is equivalent to `kMinWorkspaceRatio`? Could you add this as a comment so that the reader don't need to the arithmetic?

tfeher: Or maybe
```suggestion
    uint32_t new_max_queries = static_cast<uint32_t>(max_queries * free_space_ratio / kMinWorkspaceRatio);
    if (new_max_queries >= kMinLargeBatchSize ) {
      max_queries = new_max_queries;
```

cpp/src/neighbors/ivf_pq/ivf_pq_search.cuh
```diff
@@ -778,6 +778,44 @@ inline auto get_max_batch_size(raft::resources const& res,
...
+{
+  size_t data_size = 4;
+  switch (params.coarse_search_dtype) {
+    case CUDA_R_32F: data_size = 4; break;
+    case CUDA_R_16F: data_size = 2; break;
+    case CUDA_R_8I: data_size = 1; break;
+    default: RAFT_FAIL("Unexpected coarse_search_dtype (%d)", int(params.coarse_search_dtype));
+  }
+  // How much data we allocate for coarse GEMM.
+  // This is NOT all memory we need, as a rule of thums max it out to half of the workspace.
```
tfeher:
```suggestion
  // This is NOT all memory we need, as a rule of thumb max it out to half of the workspace.
```

cpp/src/neighbors/ivf_pq/ivf_pq_search.cuh
```diff
@@ -867,9 +905,8 @@ inline void search(raft::resources const& handle,
   auto mr = raft::resource::get_workspace_resource(handle);
 
   // Maximum number of query vectors to search at the same time.
-  const auto max_queries =
-    std::min<uint32_t>(std::max<uint32_t>(n_queries, 1), params.max_internal_batch_size);
-  auto max_batch_size = get_max_batch_size(handle, k, n_probes, max_queries, max_samples);
+  const auto max_queries = get_max_queries(handle, params, n_probes, index.n_lists(), n_queries);
+  auto max_batch_size    = get_max_batch_size(handle, k, n_probes, max_queries, max_samples);
```
tfeher: Now these are batch sizes for coarse and fine search, right? We shall name them as such.

tfeher (changes requested):
Thanks Artem for the fix, I have left a few suggestions below.

cpp/src/neighbors/detail/cagra/cagra_build.cuh
achirkin: Thanks for the comments, all addressed

PR #1003, opened 2025-06-10
Reuse `<version_config.h>` instead of separate version constants in `c_api.h`

mythrocks:
This commit started off trying to fix an error in how version strings are replaced in the C API header. (Bumping the version to 25.08 causes the minor version to be written as "08" instead of "8", producing an invalid octal number.)

The octal problem was already solved as part of RAPIDS CMake, which is used in cuVS to generate a `version_configi.h` with the corrected version values.  This commit removes the old version constants, and switches to using what's provided by infrastructure.


mythrocks:
That the `yaml` files needed updating on `branch-25.08` came as a surprise.  I'd appreciate if someone familiar with them reviewed that change closely.

cjnolet (approved):

cpp/include/cuvs/core/c_api.h
```diff
@@ -21,8 +21,8 @@
 #include <stdint.h>
 
 static const uint16_t CUVS_VERSION_MAJOR = 25;
-static const uint16_t CUVS_VERSION_MINOR = 06;
-static const uint16_t CUVS_VERSION_PATCH = 00;
+static const uint16_t CUVS_VERSION_MINOR = 8;
+static const uint16_t CUVS_VERSION_PATCH = 0;
```
KyleFromNVIDIA: Rather than hard-coding these numbers in a C header, please consider generating them at CMake configure time with [`configure_file()`](https://cmake.org/cmake/help/latest/command/configure_file.html).

KyleFromNVIDIA (changes requested):

cpp/include/cuvs/core/c_api.h
mythrocks: I think I haven't grokked your suggestion properly.  

When `update-version.sh` is run, we would like for version-strings across all artifacts to be changed.  For instance, `update-version.sh 25.10` should cause the C API to return `25`, `8`, `0` as appropriate.

Should `update-version.sh` now change our CMake files, instead of directly changing `c_api.h`?

cjnolet:
@mythrocks 

I think the `update_version.sh` does change the cmake files, doesn't it? I think the suggest was just to have cmake update the c header, rather than relying on update_versions to do it. Honestly, i could go either way with this, so it doesn't matter a ton to me- so long as we aren't forced to update anything manually.

mythrocks:
(I'll fix the copyright dates shortly.)

mythrocks:
> I think the update_version.sh does change the cmake files, doesn't it?

Not directly.  It seems to update the `VERSION` file, and then CMake does its bit based on the `VERSION`.

> so long as we aren't forced to update anything manually.

I've gotten an idea of how @KyleFromNVIDIA would prefer this were fixed.  I'll try incorporate that change ASAP.

mythrocks:
It took me a little while to realize that what I'm trying to build here might be available already: https://github.com/rapidsai/cuvs/blob/a0804deebe093f5f1e737529ce1dad690bcfa61c/cpp/CMakeLists.txt#L40

mythrocks:
I have verified that `RAPIDS CMake` line mentioned above creates `build/include/cuvs/version_config.hpp` with constants similar to the ones I introduced in `c_api.h`:
```c++
#define CUVS_VERSION_MAJOR 25
#define CUVS_VERSION_MINOR 8
#define CUVS_VERSION_PATCH 0
```
This file lands up also being CMake-installed into the final install directory.

But I can't seem to find a way to include the header in `c_api.h`:
```
/home/mithunr/workspace/dev/cuvs/3/cpp/include/cuvs/core/c_api.h:19:10: fatal error: cuvs/version_config.hpp: No such file or directory
   19 | #include <cuvs/version_config.hpp>
      |          ^~~~~~~~~~~~~~~~~~~~~~~~~
```
I'll seek @KyleFromNVIDIA's guidance on how this needs to be included, for the first build.

mythrocks:
> But I can't seem to find a way to include the header in `c_api.h`:

I've modified the `CMakeLists` for `libcuvs` and its tests to find the header correctly.  I think this should address what @KyleFromNVIDIA had in mind.

mythrocks:
@KyleFromNVIDIA, I had neglected to remove the `sed` bit to update `c_api.h` version numbers.  None of that is required, now that we depend on the headers from `RAPIDS CMake`.  I trust this is closer now to final form.

mythrocks:
@KyleFromNVIDIA, please let me know if this looks acceptable.

cpp/tests/CMakeLists.txt
```diff
@@ -292,6 +292,9 @@ if(TARGET cuvs::c_api)
   endif()
 
   add_executable(cuvs_c_test core/c_api.c)
+  target_include_directories(
+    cuvs_c_test PUBLIC "$<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}/../include>"
+  )
```
KyleFromNVIDIA: This shouldn't be needed, since we're already linking against `cuvs::c_api`, which provides this include directory as `PUBLIC`.

```suggestion
```

cpp/tests/core/c_api.c
```diff
@@ -15,6 +15,8 @@
  */
 
 #include <cuvs/core/c_api.h>
+#include <cuvs/version_config.hpp>
```
KyleFromNVIDIA: I'm tempted to change this to `<cuvs/version_config.h>` at https://github.com/rapidsai/cuvs/blob/35d91f996b281d0923dcf2d5619cba58a564e89b/cpp/CMakeLists.txt#L40, since it's pure C and is being included from a C file. We should also change the template name at https://github.com/rapidsai/rapids-cmake/blob/d183fca37264c240c0fe69c28b92c2a4d51cfedc/rapids-cmake/cmake/write_version_file.cmake#L95, though that one's just an internal implementation detail that doesn't have any effect on the final filename. cc @robertmaynard

cpp/include/cuvs/core/c_api.h
```diff
@@ -20,10 +20,6 @@
 #include <stdbool.h>
 #include <stdint.h>
 
-static const uint16_t CUVS_VERSION_MAJOR = 25;
-static const uint16_t CUVS_VERSION_MINOR = 06;
-static const uint16_t CUVS_VERSION_PATCH = 00;
```
KyleFromNVIDIA, line 25: Since these are no longer here, I think we should also do:

```suggestion
#include <cuvs/version_config.hpp>
```

to maintain compatibility.

cpp/src/core/c_api.cpp
```diff
@@ -14,10 +14,10 @@
  * limitations under the License.
  */
 
-#include <cstdint>
 #include <cuvs/core/c_api.h>
 #include <cuvs/core/exceptions.hpp>
-#include <memory>
+#include <cuvs/version_config.hpp>
+
```
KyleFromNVIDIA: With this include in `c_api.h` from my suggestion above, this is no longer needed.

cpp/tests/core/c_api.c
KyleFromNVIDIA: I've opened https://github.com/rapidsai/rapids-cmake/pull/867 to address this.

cpp/tests/CMakeLists.txt
mythrocks: This is surprising.  The last I tried building the tests, they could not find the C-API include headers.  My hand was forced on changing the include paths.

Let me check this again.

cpp/tests/core/c_api.c
mythrocks: I did notice that too.  Thanks for picking that up for RAPIDS CMake.

cpp/include/cuvs/core/c_api.h
mythrocks: I'm happy to make that change.

cpp/src/core/c_api.cpp
mythrocks: Agreed.

cpp/tests/CMakeLists.txt
mythrocks: You're right.  This change isn't needed.  I've removed it.

cpp/include/cuvs/core/c_api.h
mythrocks: Fixed.

cpp/src/core/c_api.cpp
mythrocks: Fixed.

cpp/tests/core/c_api.c
KyleFromNVIDIA: Thanks for making all the other changes I requested! Could you please do the rename to `.h` as well?

mythrocks: I see that https://github.com/rapidsai/rapids-cmake/pull/867 has been merged.

I'm not sure how that change might get pulled in, in the build.  Does a version pin have to be moved or something?

KyleFromNVIDIA: It does not, you will simply get the latest change automatically. However, it doesn't matter - rapidsai/rapids-cmake#867 is an internal implementation detail that this project won't see. I'm asking for `version_config.hpp` to be renamed to `version_config.h`.

mythrocks: > I'm asking for version_config.hpp to be renamed to version_config.h.

I misunderstood how that was tied together.  This is now fixed.

mythrocks:
@KyleFromNVIDIA: After including the `cuvs/version_config.h` header in `c_api.h`, I'm seeing a [Java build failure](https://github.com/rapidsai/cuvs/actions/runs/15913296190/job/44886858359?pr=1003):
```
c_api.h:19:10: error: 'cuvs/version_config.h' file not found
```
(This wasn't happening before, because the version header was included only in `c_api.cpp`, and not subject to `jextract`.)

I fixed this in my local builds by [including `cpp/build/include` in the include paths](
[ed14bb7](https://github.com/rapidsai/cuvs/pull/1003/commits/ed14bb7c4947b98fc7ae32a9e83a2011d71680e5)).  But it appears that the path used in CI builds is a different one.

By my reading, even if we were to use the right one for CI, it would still cause `jextract` to fail in local dev builds.  Note that `jextract` is being run for the Java builds, and not from running `cmake`.

At this point, I'm inclined to change things back so that:
1. `<version_config.h>` is included only in the `c_api.cpp`.
2. Change the `jextract` code back to not depending on finding `<version_config.h>`.

Please let me know if that is agreeable.

mythrocks:
@jameslamb, @rhdong:  I was wondering if either of you might be able to help here.

The problem is as follows:
1. There is a header (`<cuvs/version_config.h>`) that is *generated* during the C++ build of `libcuvs`.  This is now required in the Java build.
2. I do know that on my workstation, `cuvs/version_config.h` is generated in `$PROJECT_ROOT/cpp/build/include`.  I don't know what the location of this file is, in the C++ CI build.  It doesn't seem to be in `cpp/build/include`.
3. I have tried including `"${REPODIR}"/cpp/build/include` as an include-directory for the Java build.  But the build still [fails to find the header](https://github.com/rapidsai/cuvs/pull/1003#issuecomment-3010499375).
4. Debugging the build, I ran a `find` for the header in `$REPODIR`.  It isn't found.

May I please know in which directory the C++ build artifacts go?  Also, what might be the best way to communicate that to the Java build?

(@KyleFromNVIDIA:  This is a last-ditch attempt.  If there isn't an easy/good way to resolve this, I'd very much like to include the version-header in the `c_api.c` file, and be done with it for now.  We can do a follow-on PR, in case someone misses the version strings being in the `c_api.h`.  That change was too recent for this to be a problem, I think.)

KyleFromNVIDIA:
> If there isn't an easy/good way to resolve this, I'd very much like to include the version-header in the `c_api.c` file, and be done with it for now. We can do a follow-on PR, in case someone misses the version strings being in the `c_api.h`. That change was too recent for this to be a problem, I think.

That's fine with me, if you can't figure out another solution then go ahead and do it.

PR #1005, opened 2025-06-11
[REVIEW][Java] Add jextract artifacts to gitignore

ldematte:
The build process for Java uses `jextract` to generate panama bindings. Jextract is downloaded from the JDK website.
This PR excludes the downloaded Jextract archive and the Jextract files from the repository by adding them to `.gitignore`

### DIFF
--- a/java/.gitignore
+++ b/java/.gitignore
@@ -3,5 +3,7 @@
 /cuvs-java/bin/
 /cuvs-java/src/main/java22/com/nvidia/cuvs/internal/panama/
 /cuvs-java/*.cag
+jextract-22/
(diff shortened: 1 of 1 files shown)


cjnolet:
@mythrocks @ldematte CI should be fixed shortly. It's held up because we recently changed the base OS and it's exposed a couple minor bugs in our tests.

mythrocks:
> @mythrocks @ldematte CI should be fixed shortly. It's held up because we recently changed the base OS and it's exposed a couple minor bugs in our tests.

@cjnolet: Thank you for https://github.com/rapidsai/cuvs/pull/1008.

Issue #764, opened 2025-03-14
cuvs-java: CagraBuildAndSearchIT crashes with pure virtual method called terminate called without an active exception


ChrisHegarty:
The `CagraBuildAndSearchIT` test randomly decides to " run concurrently". If you force this to true - so run concurrently, then the test exhibits failures and crashes intermittently, which still appearing to pass successfully.   This is, in part, because the test executes the concurrent tasks without checking their result.  It would be better to collect the set of task futures and assert the expected result ( minimally 0 or 1 to indicate that no exception was thrown ).

Additionally, frequently see an error from native relating to abstract function invocation after deconstruction.

```
pure virtual method called
terminate called without an active exception
Aborted (core dumped)
```

ldematte:
I was not able to reproduce this anymore on the 25.08 branch, but I was hitting this on the 25.06
My guess is that the recent removal of the C shim in https://github.com/rapidsai/cuvs/pull/916 fixed this; the error appears to be in native (C++), possibly a wrong cast (it is what you see when you try to call a function that has NULL in its v-table slot IIRC).

In any case, I changed the test in https://github.com/rapidsai/cuvs/pull/1006 so that a) it always runs both normally and concurrently and b) waits for the runners to complete and propagate the result (success or exceptional) to the caller.

PR #1006, opened 2025-06-11
[REVIEW][Java] Refactor CagraBuildAndSearchIT to explicitly express different execution modes

ldematte:
This PR refactors `CagraBuildAndSearchIT`, introducing separate test for different execution modes (same thread, different thread, concurrent).
This way different modes are tested every time, without the need to rely on randomization to hit a particular test path.
The execution is handled by utility functions (e.g. `runConcurrently`); right now these are private to the test suite, but if we think they could be useful across other tests I can move them to a utility class.

It also fixes a issue where we did not wait and/or report failures in the concurrent case (now exceptions are propagated back to the calling JUnit thread).

### DIFF
--- a/java/cuvs-java/src/test/java/com/nvidia/cuvs/CagraBuildAndSearchIT.java
+++ b/java/cuvs-java/src/test/java/com/nvidia/cuvs/CagraBuildAndSearchIT.java
@@ -17,15 +17,26 @@
 package com.nvidia.cuvs;
 
 import static com.carrotsearch.randomizedtesting.RandomizedTest.assumeTrue;
-import static org.junit.Assert.*;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.fail;
 
 import java.io.File;
 import java.io.FileInputStream;
 import java.io.FileOutputStream;
 import java.io.InputStream;
 import java.lang.invoke.MethodHandles;
-import java.util.*;
-import java.util.concurrent.*;
+import java.util.Arrays;
+import java.util.BitSet;
+import java.util.List;
+import java.util.Map;
+import java.util.UUID;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ExecutorService;
+import java.util.concurrent.Executors;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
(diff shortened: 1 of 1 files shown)


ChrisHegarty:
>It also fixes a issue where we did not wait and/or report failures in the concurrent case (now exceptions are propagated back to the calling JUnit thread).

This is an important and significant point, which I want to emphasise - it's the reason why the bug reported in https://github.com/rapidsai/cuvs/issues/764 went unnoticed.

java/cuvs-java/src/test/java/com/nvidia/cuvs/CagraBuildAndSearchIT.java
```diff
@@ -200,272 +236,264 @@ public void testPrefilteringReducesResults() throws Throwable {
...
 
+  // Commented out test for Logical merge strategy as it is not yet implemented in C yet
+  @Test
+  public void testMergeStrategies() throws Throwable {
+    float[][] vector1 = {
+      {0.0f, 0.0f},
+      {1.0f, 1.0f}
+    };
+
+    float[][] vector2 = {
```
mythrocks, line 405: Regarding `testMergeStrategies` and `testMergingIndexes`:  Am I right in understanding that these are only whitespace changes?

ldematte: Yeah, sorry about that, I tried to use the same convention but the file already had mixed 2 and 4 spaces indents. BTW, is there a convention?

mythrocks: No, we don't appear to have a convention yet.  I've been toying with getting `google-java-format` going for this project, but we don't have anything set up yet. 

For what it's worth, I prefer the formatting that you introduced here.

mythrocks: I'll file a separate issue to introduce a code formatter for `cuvs-java`.

mythrocks:
It looks like the PR merged with the `[REVIEW]` tag in the title. :/ We'll catch that the next time around.

cjnolet:
> It looks like the PR merged with the [REVIEW] tag in the title. :/ We'll catch that the next time around.

@mythrocks once a PR is set to review, it's ready for merge if it's approved. The only time we won't review/merge is if it has WIP In the title.

Issue #699, opened 2025-02-14
cuvs-java: Evaluate adding a query mapping function (avoids reifying a mapping list)


ChrisHegarty:
This issue has been filed to evaluate modifying the `Builder withMapping(List<Integer> mapping)` API, so that it does not require to reify a concrete list of map ids.

For example:
```
List<Integer> map =...

var query = new CagraQuery.Builder()
        .withMapping(map)
```

From my understanding a function of `int -> int` is sufficient. Maybe an additional `size()`. For Lucene this would work better than reifying the ordinal to doc mapping as a list.

e.g.   

`Builder withMapping(MappingFunction ordToDocFunc)`

```
@FunctionalInterface
public interface MappingFunction {
  /** The given a vector ordinal returns a docId */
  int apply(int value);

  int size(); // if needed
}
```

PR #1007, opened 2025-06-11
[Java] Using functions for mapping

ldematte:
This PR replaces "array" based ID mappings with a more free-form mapping based on functions.
This should be more efficient both when the number of IDs is huge (you don't need to actually allocate big list for all IDs to map) and in general (as it won't need to perform boxing/unboxing of integers).
It should also be more flexible; as an example, I have introduced a simple helper to wrap a list, so that the previous approach can still be used. The same can be done for maps etc.

Relates to https://github.com/rapidsai/cuvs/issues/699


java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/CagraIndexImpl.java
```diff
@@ -242,7 +242,7 @@ private IndexReference build() throws Throwable {
   public SearchResults search(CagraQuery query) throws Throwable {
     try (var localArena = Arena.ofConfined()) {
       checkNotDestroyed();
-      int topK = query.getMapping() != null ? Math.min(query.getMapping().size(), query.getTopK()) : query.getTopK();
+      int topK = query.getTopK();
```
ldematte, line 242: @ChrisHegarty do you think we need size as a mechanism to limit to N entries (like here), as you suggested in https://github.com/rapidsai/cuvs/issues/699? Or can we just use `java.util.function.LongToIntFunction` as I did in this PR?

java/cuvs-java/src/main/java/com/nvidia/cuvs/HnswQuery.java
```diff
@@ -128,7 +128,7 @@ public Builder withQueryVectors(float[][] queryVectors) {
      * @param mapping the ID mapping instance
      * @return an instance of this Builder
      */
-    public Builder withMapping(List<Integer> mapping) {
+    public Builder withMapping(LongToIntFunction mapping) {
```
ChrisHegarty, line 138: Would u mind updating and fleshing out the javadoc a bit more here? Thank you.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/CagraIndexImpl.java
ChrisHegarty: Yeah, I don't think that we need the size parameter at all - at least I cannot see any usage for it.  Is IntToInt sufficient? Or are the ordinals already longs?

ldematte: From the layout, the ordinals are longs, that's why I went for LongToIntFunction (the "identity" embeds the cast that used to be in `mapping.get((int) id)`).
But I'm fine either way.

ChrisHegarty: Ok, longs is fine then.

chatman:
@ldematte One of my wishlist items was to compute this mapping on the GPU (when possible) to cut down on time to return the results from the various algorithms. Maybe something like a transform operation (https://nvidia.github.io/cccl/thrust/api/function_group__transformations_1ga62a6227031c9c811c5b497b347398a8a.html), or something similar, if available in Raft/RAPIDS.

Towards that, maybe we could bring back (*in addition to* this function) a `int[] mapping` (not the currently available `List<Integer>` though) at a later point in time, and keep this function based mapping for the usecases where CPU is being used?

Alternatively, the same function can be used to construct an int[] for the mapping on the GPU. However, constructing it on every request could be as bad as doing the entire mapping application on the CPU itself, so maybe some strategy involving caching the constructed mapping int[] could be explored.

mythrocks:
This PR seems to be labeled as `Non-breaking`.  But the public interfaces for `BruteForceQuery`, `CagraQuery`, etc. are changing.

Any objection to labeling this as `Breaking`?

java/cuvs-java/src/main/java/com/nvidia/cuvs/CagraQuery.java
```diff
@@ -19,6 +19,7 @@
 import java.util.Arrays;
 import java.util.List;
```
mythrocks: This can be removed now, I think.  You've removed it from `BruteForceQuery.java`.
```suggestion
```

java/cuvs-java/src/main/java/com/nvidia/cuvs/SearchResults.java
```diff
@@ -18,9 +18,24 @@
...
 import java.util.List;
 import java.util.Map;
+import java.util.function.LongToIntFunction;
 
 public interface SearchResults {
 
+    /**
+     * The default identity function mapping neighbours IDs to user-defined IDs
+     */
+    LongToIntFunction IDENTITY_MAPPING = l -> (int) l;
```
mythrocks: Should this be a `static final`?

java/cuvs-java/src/main/java/com/nvidia/cuvs/SearchResults.java
```diff
@@ -18,9 +18,24 @@
...
+
+    /**
+     * Creates a mapping function from a list lookup of custom user IDs
+     * @param mappingAsList a positional list of custom user IDs
+     * @return a function that maps the input ordinal to a custom user IDs, using the input as an index in the list
+     */
+    static LongToIntFunction mappingsFromList(List<Integer> mappingAsList) {
+        return l -> mappingAsList.get((int) l);
```
mythrocks: 👏

mythrocks (approved):
A couple of nitpicks, but this looks good to go.  Thank you for this change, @ldematte!

ChrisHegarty:
> This PR seems to be labeled as `Non-breaking`. But the public interfaces for `BruteForceQuery`, `CagraQuery`, etc. are changing.
> 
> Any objection to labeling this as `Breaking`?

The change in the java API signature is clearly an incompatible change, no question.  However, the API is experimental, and as such is still evolving before being made final. So it't not a breaking change. We're likely going to have several of these kinds of change in the next release.

java/cuvs-java/src/main/java/com/nvidia/cuvs/CagraQuery.java
ldematte: Yes, hopefully after https://github.com/rapidsai/cuvs/pull/1049 all of this will be automatic!

java/cuvs-java/src/main/java/com/nvidia/cuvs/SearchResults.java
ldematte: SearchResults is an interface, so `static final` is redundant. But if you think it makes it clearer, I can add it

java/cuvs-java/src/test/java/com/nvidia/cuvs/HnswBuildAndSearchIT.java
```diff
@@ -97,43 +81,109 @@ public void testIndexingAndSearchingFlow() throws Throwable {
...
+            // Check results
+            log.info(results.getResults().toString());
+            checkResults(expectedResults, results.getResults());
+
+            // Cleanup
+            index.destroyIndex();
+            hnswIndex.destroyIndex();
+            if (hnswIndexFile.exists()) {
+                hnswIndexFile.delete();
+            }
```
mythrocks: This feels like it should be in a `finally {}` block.

mythrocks: (Non-binding.  We can revisit in a follow-up.)

mythrocks:
I've merged #1049.  I'm afraid this PR will need a rebase.

ldematte:
> I've merged https://github.com/rapidsai/cuvs/pull/1049. I'm afraid this PR will need a rebase.

All of the pending PRs will need that :) 
Done!

mythrocks:
Thank you for this change, @ldematte.  I've merged this change.

PR #1012, opened 2025-06-11
Add support for filtered search to the tiered_index code

benfred:
The tiered_index code wasn't working with filtered search - since the filter passed to the brute force portion wasn't translating for the number of rows seen already. Fix and add a python level unittest that would have caught this.

Also:
* expose 'tiered_index.compact' to python
* add java dependencies to 'all' conda environments


conda/environments/all_cuda-129_arch-x86_64.yaml
```diff
@@ -27,18 +27,21 @@ dependencies:
...
 - libcusparse-dev
 - librmm==25.8.*,>=0.0.0a0
 - make
+- maven
 - nccl>=2.19
 - ninja
 - numpy>=1.23,<3.0a0
 - numpydoc
 - openblas
+- openjdk=22.*
```
cjnolet, line 43: Don't need to do it in this PR, but at some point, I think we should separate the different languages into different "environment" files so that a user working predominently on Java, for example, doesn't need to have Go and Rust things installed.

PR #1016, opened 2025-06-11
Mutual Reachability using NN Descent in all-neighbors API

jinsolp:
Making mutual reachability part of all-neighbors. **Currently only for mutual reachability using NN Descent.** 
If `core_distances` optional matrix is given, automatically calculates distances and indices in mutual reachability space.

Brute force mutual reachability as part of all-neighbors will be in a different PR (didn't want to make the PR too big).


jinsolp:
Waiting on https://github.com/rapidsai/cuvs/pull/1073 and https://github.com/rapidsai/cuvs/pull/1062

jinsolp:
waiting on https://github.com/rapidsai/cuvs/pull/1108

cpp/include/cuvs/neighbors/graph_build_types.hpp
```diff
@@ -27,7 +27,7 @@ namespace cuvs::neighbors {
  * @{
  */
 
-enum GRAPH_BUILD_ALGO { BRUTE_FORCE = 0, IVF_PQ = 1, NN_DESCENT = 1 };
+enum GRAPH_BUILD_ALGO { BRUTE_FORCE = 0, IVF_PQ = 1, NN_DESCENT = 2 };
```
cjnolet, line 30: We used this pattern of suffixing public API files in RAFT w/ a `_types.hpp` file specifically because it's header only and so including more header files than needed can be costly. That being said, we no longer have that constraint here, since the public API headers are actually very lightweight. For that reason, I would highly suggest we move this directly into `all_neighbors.hpp` or, if it will need to be shared by multiple all_neighbors headers, moving it into `all_neighbors/common.hpp`. I would suggest the former at the present time, though.

cpp/src/neighbors/all_neighbors/all_neighbors_batched.cuh
```diff
@@ -34,6 +36,29 @@
...
+  size_t num_rows = static_cast<size_t>(global_neighbors.extent(0));
+  size_t k        = static_cast<size_t>(global_neighbors.extent(1));
+
+  bool select_min = cuvs::distance::is_min_close(metric);
+  IdxT global_neighbors_fill_value =
+    select_min ? std::numeric_limits<IdxT>::max() : std::numeric_limits<IdxT>::min();
+  T global_distances_fill_value =
+    select_min ? std::numeric_limits<T>::max() : std::numeric_limits<T>::min();
+
+  std::fill(global_neighbors.data_handle(),
```
cjnolet: We have a `fill` function in RAFT we should use.

cpp/include/cuvs/neighbors/graph_build_types.hpp
jinsolp: cagra uses this as well, and that's why it's not under the `all_neighbors` directory. We can't put this into `include/cuvs/neighbors/common.hpp` because that introduces a circular dependency. 

what would be a good way to handle this in this case?

jinsolp:
waiting for an edge case fix: https://github.com/rapidsai/cuvs/pull/1185

cpp/include/cuvs/neighbors/graph_build_types.hpp
cjnolet: If this is being used by more than just `all_neighbors` then I'm fine promiting it to a more common place. My only point is that these `_types` files should probbaly be consolidated into files called `common.hpp` since in cuVS the definitions are very lightweight.

jinsolp: Let me put this in a separate PR. Issue added [here](https://github.com/rapidsai/cuvs/issues/1233)

jinsolp:
@cjnolet Letting you know that all blocking PRs are merged and this is ready for a final round of reviews now 🙂

PR #1019, opened 2025-06-12
refactor(shellcheck): fix all remaining warnings/errors

gforsyth:
Finishes up the shellcheck work tracked in rapidsai/build-planning#135


gforsyth:
@jameslamb one more for you here

.shellcheckrc
```diff
@@ -0,0 +1 @@
+disable=SC1091
```
jameslamb, line 2:
```suggestion
# Disable file checks (otherwise every use of `gha-tools` will get flagged)
disable=SC1091
```

Let's bring over that same informative comment you've added in other repos like https://github.com/rapidsai/raft/pull/2703

build.sh
```diff
@@ -367,15 +372,13 @@ fi
 # Configure for building all C++ targets
 if (( NUMARGS == 0 )) || hasArg libcuvs || hasArg docs || hasArg tests || hasArg bench-prims || hasArg bench-ann; then
     COMPILE_LIBRARY=ON
-    if (( ${BUILD_SHARED_LIBS} == "OFF" )); then
-        CMAKE_TARGET="${CMAKE_TARGET};"
-    else
-        CMAKE_TARGET="${CMAKE_TARGET};cuvs"
+    if ! (( BUILD_SHARED_LIBS == "OFF" )); then
```
jameslamb:
```suggestion
    if [[ "${BUILD_SHARED_LIBS}" != "OFF" ]]; then
```

Very very minor personal preference thing, but I think using `[[` and `!=` for this would more closely match the style of the rest of the script. Across RAPIDS, I think I've only seen `(( ))` used for math.

build.sh
```diff
@@ -390,33 +393,33 @@ if (( NUMARGS == 0 )) || hasArg libcuvs || hasArg docs || hasArg tests || hasArg
...
           -DBUILD_SHARED_LIBS=${BUILD_SHARED_LIBS} \
-          ${CACHE_ARGS} \
-          ${EXTRA_CMAKE_ARGS}
+          "${CACHE_ARGS[@]}" \
+          "${EXTRA_CMAKE_ARGS[@]}"
 
   compile_start=$(date +%s)
-  if [[ ${CMAKE_TARGET} != "" ]]; then
-      echo "-- Compiling targets: ${CMAKE_TARGET}, verbose=${VERBOSE_FLAG}"
+  if [[ ${CMAKE_TARGET[*]} != "" ]]; then
```
jameslamb:
```suggestion
  if [[ ${#CMAKE_TARGET} -eq 0 ]]; then
```

I don't understand this one. It doesn't look to me like `CMAKE_TARGET` is an associative array... lines like this suggest it's just a regular un-keyed array of strings:

```shell
CMAKE_TARGET+=("cuvs")
```

Had to look it up in the `bash` manual to check, but `[*]` lists the *keys* of an associative (key-value) array. From https://www.gnu.org/software/bash/manual/bash.html

> `${!name[*]}`
> _If name is an array variable, expands to the list of array indices (keys) assigned in name. If name is not an array, expands to 0 if name is set and null otherwise. When `‘@’` is used and the expansion appears within double quotes, each key expands to a separate word._

I think here, it'd be better to replace this comparison to the empty strong with a check of whether the length is 0.

build.sh
```diff
@@ -390,33 +393,33 @@ if (( NUMARGS == 0 )) || hasArg libcuvs || hasArg docs || hasArg tests || hasArg
...
+      echo "-- Compiling targets: ${CMAKE_TARGET[*]}, verbose=${VERBOSE_FLAG}"
```
jameslamb:
```suggestion
      echo "-- Compiling targets: ${CMAKE_TARGET[@]}, verbose=${VERBOSE_FLAG}"
```

Similar to my comment above... I think here you want the *values*, not the *keys* (because there are no keys).

.shellcheckrc
jameslamb: Also similar to my comments in previous PRs, could you please add this file to the `ci-codeowners` block in `CODEOWNERS` and to the list of files in the `changed-files` config in `pr.yaml`?

jameslamb (changes requested):
I left a couple suggestions... leaving a blocking review because I think one is a correctness problem. The others are just small suggestions for improvement.

`@` me whenever you want another review.

build.sh
gforsyth: It's not an associative array, it's an indexed array.  The `'${CMAKE_TARGET[*]}'` is a concatenation of all elements into a space-delimited string

```bash
 > CMAKE_TARGET=()
 > echo ${CMAKE_TARGET[*]}

 > if [[ ${CMAKE_TARGET[*]} == "" ]]; then echo hi; fi
hi
```

```bash
 > CMAKE_TARGET=("one" "two" "three")
 > echo ${CMAKE_TARGET[*]}
one two three
 > if [[ ${CMAKE_TARGET[*]} == "" ]]; then echo hi; fi
 >
```

build.sh
gforsyth: Yeah, I think this is another weird implicit vs. explicit array concatenation thing -- `shellcheck` doesn't like it when strings and arrays are mixed: https://github.com/koalaman/shellcheck/wiki/SC2145

gforsyth: huh, well, I'm definitely wrong here because that didn't output any values

gforsyth: Hmm, or maybe `CMAKE_TARGET` was just empty?  I've pushed up a different way of spelling this, in any case, but I think this should work.

```bash
 > CMAKE_TARGET=('one' 'two')
 > echo "-- Compiling targets: ${CMAKE_TARGET[*]}"
-- Compiling targets: one two
```

build.sh
jameslamb: Ok, maybe I am misunderstanding what "key" means in those docs.

I still think "is the array length-0" is a bit simpler than "is the concatenation of all the elements equal to an empty string", but if `[*]` works than this suggestion moves from "please change" to "do what you want".

gforsyth: I think your solution is cleaner -- I'll switch to that

cpp/scripts/run-cmake-format.sh
divyegala, line 1: Would it make sense (if it's possible) to exclude this file from cpp-ownership?

jameslamb: I'd support that change.

jameslamb (approved):
Approving, I think all of my comments have been addressed.

Thanks for finishing this one up! This project has a bit more shell code than RAPIDS repos, because it has bindings for so many languages and scripts to build those bindings. I think we'll be really glad to have `shellcheck` helping us with development here 😁

benfred (approved):

PR #1024, opened 2025-06-16
[Java] Encapsulate on-heap float arrays into `Dataset`

ldematte:
This PR is a follow-up from https://github.com/rapidsai/cuvs/pull/902.
Still WIP (see self-comments on the changes) but I'd like some early feedback.


java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/ArrayDatasetImpl.java
```diff
@@ -0,0 +1,42 @@
...
+
+    public ArrayDatasetImpl(float[][] vectors) {
+        this.vectors = Objects.requireNonNull(vectors);
+        if (vectors.length == 0) {
+            throw new IllegalArgumentException("vectors should not be empty");
+        }
+    }
+
+    @Override
+    public void addVector(float[] vector) {
```
ldematte: I think addVector should be moved out of Dataset, and into some sort of builder if we want to retain the same capabilities.
E.g. something like
```
var builder = CuVSProvider.provider().newDatasetBuilder(size, dimensions);
builder.addVector(...).addVector(...) // move addVector from DatasetImpl to the new builder
var dataset = builder.build(); // this creates the Dataset

File: java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/BruteForceIndexImpl.java [hunk 2]
```diff
@@ -145,14 +142,13 @@ public void destroyIndex() throws Throwable {
...
     try (var localArena = Arena.ofConfined()) {
-      long rows = dataset != null? dataset.size(): vectors.length;
-      long cols = dataset != null? dataset.dimensions(): (rows > 0 ? vectors[0].length : 0);
+      long rows = dataset.size();
+      long cols = dataset.dimensions();
 
       Arena arena = resources.getArena();
-      MemorySegment datasetMemSegment = dataset != null? ((DatasetImpl) dataset).seg:
-          Util.buildMemorySegment(resources.getArena(), vectors);
+      MemorySegment datasetMemSegment = ((MemorySegmentProvider) dataset).asMemorySegment(arena);
```
ldematte: I'm not particularly fond of this, but we are the ones creating all the dataset instances so we are pretty much in control here, and if we don't want to bump req to Java 22 and/or enable preview APIs we have not much choice than to keep this here.

I've added an assert here too: `assert dataset instanceof MemorySegmentProvider;`

ldematte:
@chatman @mythrocks wdyt?

ldematte:
I added another commit https://github.com/rapidsai/cuvs/pull/1024/commits/3f348c6be56b5327161211960cb8639898f41930 in which I reduce the scope for the Dataset/MemorySegment: I think there is no need to keep it around once the index is built, it is not referenced again (but keep me honest here and point out if I'm wrong!)

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/ArrayDatasetImpl.java
ldematte: I have implemented this in this commit: https://github.com/rapidsai/cuvs/pull/1024/commits/97c01edb95b1347c3613cb096df0266a2052181c

mythrocks:
> @chatman @mythrocks wdyt?

Looking forward to reading/reviewing this next week.

chatman:
Thanks @ldematte. Looks good. I'll take a closer look at the scope of the
memory segment. I remember extending to scope of the memory segment to
closing it at index close due to some issues otherwise. I'll take another
look today to make sure that's not happening.

ldematte:
> I'll take a closer look at the scope of the
memory segment. I remember extending to scope of the memory segment to
closing it at index close due to some issues otherwise. I'll take another
look today to make sure that's not happening.

Yes please; it might be that the C++ side of things holds a reference to the input memory (wrapped and managed by the  `MemorySegment`), and therefore that that memory needs to be available for the lifetime of the Index.
If we even suspect this is the case, I'll revert that change - better hold onto some memory a bit longer than shorter than we need!

ldematte:
BTW, no hurry -- I wasn't aware you guys were off this week. This can wait till next week, no prob.

ldematte:
> I'll take a closer look at the scope of the
memory segment. I remember extending to scope of the memory segment to
closing it at index close due to some issues otherwise.

Hey @chatman, I dug into the C and C++ implementations of `cuvsCagraBuild`/`cuvs::neighbors::cagra::build` and a pointer to the original dataset seems to be passed down and retained/referenced by Index. _Seems_ because the DLTensor gets transformed to a mdspan, but I think it's just wrapped and no copy happens there, so we need to retain the dataset memorysegment "alive" for the lifetime of the index. I'll change this PR to do that.

chatman:
Thanks a lot for looking deeper. This was precisely my concern, and thanks
for looking into it. @cjnolet might know for sure, but it would be pretty
safe to keep the scope tied up with the lifecycle of the index.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/BruteForceIndexImpl.java
```diff
@@ -92,14 +88,14 @@ public class BruteForceIndexImpl implements BruteForceIndex {
...
-      throws Throwable {
-    this.vectors = vectors;
-    this.dataset = dataset;
-    this.resources = resources;
-    this.bruteForceIndexParams = bruteForceIndexParams;
-    this.bruteForceIndexReference = build();
+  private BruteForceIndexImpl(Dataset dataset, CuVSResourcesImpl resources,
+      BruteForceIndexParams bruteForceIndexParams) throws Exception {
+    Objects.requireNonNull(dataset);
+      try (dataset) {
```
ldematte: this is OK as `build` will make a copy of the data, and not hold any reference to the original dataset pointer/segment.
The copy still has issues, but I have handled that in https://github.com/rapidsai/cuvs/pull/1045

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/CagraIndexImpl.java
```diff
@@ -161,15 +149,17 @@ private void checkNotDestroyed() {
    * Invokes the native destroy_cagra_index to de-allocate the CAGRA index
    */
   @Override
-  public void destroyIndex() throws Throwable {
+  public void destroyIndex() throws Throwable{
     checkNotDestroyed();
     try (var arena = Arena.ofConfined()) {
       int returnValue = cuvsCagraIndexDestroy(cagraIndexReference.getMemorySegment());
       checkCuVSError(returnValue, "cuvsCagraIndexDestroy");
+      if (cagraIndexReference.dataset != null) {
```
ldematte, line 151: This will go away in https://github.com/rapidsai/cuvs/pull/1045

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/CagraIndexImpl.java
```diff
@@ -740,12 +715,14 @@ public CagraIndexImpl build() throws Throwable {
...
 
     private final MemorySegment memorySegment;
+    private final Dataset dataset;
 
     /**
      * Constructs CagraIndexReference and allocate the MemorySegment.
      */
     protected IndexReference(CuVSResourcesImpl resources) {
       memorySegment = cuvsCagraIndex.allocate(resources.getArena());
+      dataset = null;
```
ldematte: This will go away in https://github.com/rapidsai/cuvs/pull/1045

ldematte:
> but it would be pretty safe to keep the scope tied up with the lifecycle of the index.

👍 
I've done exactly that in the last commit

java/README.md
```diff
@@ -25,10 +25,14 @@ do `./build.sh java` in the top level directory or just do `./build.sh` in this
...
-To run a single test:
+To run a single test suite:
 ```sh
 cd cuvs-java/
-mvn verify -Dintegration-test=com.nvidia.cuvs.CagraBuildAndSearchIT
+mvn clean integration-test -Dit.test=com.nvidia.cuvs.CagraBuildAndSearchIT
+```
+or, for a single test:
+```sh
+mvn clean integration-test -Dit.test=com.nvidia.cuvs.CagraBuildAndSearchIT#testMergeStrategies
```
mythrocks, line 35: 👍

java/cuvs-java/src/main/java/com/nvidia/cuvs/CagraIndex.java
```diff
@@ -143,13 +143,6 @@ default void serializeToHNSW(OutputStream outputStream, Path tempFile) throws Th
    */
   void serializeToHNSW(OutputStream outputStream, Path tempFile, int bufferLength) throws Throwable;
 
-  /**
-   * Gets an instance of {@link CagraIndexParams}
-   *
-   * @return an instance of {@link CagraIndexParams}
-   */
-  CagraIndexParams getCagraIndexParameters();
```
mythrocks, line 151: [Question] Still looking to understand why this was removed.

mythrocks (approved):
LGTM!  Will schedule to merge this once #1045 is merged.

java/cuvs-java/src/main/java/com/nvidia/cuvs/CagraIndex.java
ldematte: Index parameters are only using at index build time; there is no reason to retain them in the index once this is build (which, in Java, we do directly in the ctor). They are simply inputs.

ldematte:
@mythrocks rebased on top of #1045 and ready to go! 👍

mythrocks:
I just resolved the conflicts from https://github.com/rapidsai/cuvs/pull/1081.  This should go in after CI passes.

mythrocks:
Thank you for this change, @ldematte.  This has been merged.

Issue #1014, opened 2025-06-11
[FEA] [Java] Add support for Tiered Indexes in `cuvs-java`


mythrocks:
#879 added support for Tiered Indexes in `25.06`.  [`cuvs/neighbors/tiered_index.h`](https://github.com/rapidsai/cuvs/blob/branch-25.08/cpp/include/cuvs/neighbors/tiered_index.h) has the C API for it.

The `cuvs-java` layer needs to expose these for Java users.

mythrocks:
Including `tiered_index.h` in `headers.h` revealed the compile problems listed in #996.  That should sort it out.  

I'm wrestling a little with Panama for this header now.  I'll post a WIP branch as I progress.

mythrocks:
This should be resolved by #1028.  The PR is under review now.

PR #1028, opened 2025-06-17
[Java] Support for tiered index

punAhuja:
TieredIndex exposed in the Java layer.

 Code is written and tested for CAGRA algorithm only.
 
 Serialization, Deserialization and merge is still not supported in the C layer.

Fixes: #1014


mythrocks:
Thank you for working on this.  I'm taking a look at this now.

java/cuvs-java/src/main/java/com/nvidia/cuvs/TieredIndex.java
```diff
@@ -0,0 +1,195 @@
...
+
+import java.io.InputStream;
+import java.util.Objects;
+
+import com.nvidia.cuvs.spi.CuVSProvider;
+
+/**
+ * {@link TieredIndex} encapsulates a Tiered index, along with methods to
+ * interact with it.
+ *
```
mythrocks:
```suggestion
```

mythrocks:
@punAhuja:  I'm afraid I've merged #1049, so this PR will need its conflicts resolved.

punAhuja:
@mythrocks Resolved the conflicts.

java/cuvs-java/src/main/java/com/nvidia/cuvs/TieredIndex.java
```diff
@@ -0,0 +1,193 @@
...
+   */
+  SearchResults search(TieredIndexQuery query) throws Throwable;
+
+  /**
+   * Returns the algorithm type backing this TieredIndex.
+   *
+   * @return The {@link TieredIndexType} indicating the underlying algorithm
+   *         (e.g., CAGRA)
+   */
+  TieredIndexType getIndexType();
```
mythrocks, line 53: I haven't understood where this is used, or called from.  

Is this meant to be used when `merge()` is implemented?  Like calling the appropriate `merge()` depending on the underlying index algo?

java/cuvs-java/src/main/java/com/nvidia/cuvs/TieredIndex.java
```diff
@@ -0,0 +1,193 @@
...
+
+  /**
+   * Returns the configuration parameters used to build this TieredIndex.
+   *
+   * @return An instance of {@link TieredIndexParams} containing the index
+   *         configuration
+   */
+  TieredIndexParams getIndexParameters();
```
mythrocks: Where is this called from?  Is this meant to be used when `merge()` is implemented?

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/TieredIndexImpl.java
```diff
@@ -0,0 +1,665 @@
...
+              : Util.buildMemorySegment(resources.getArena(), vectors);
+
+      Arena arena = resources.getArena();
+      long cuvsRes = resources.getMemorySegment().get(cuvsResources_t, 0);
+      MemorySegment stream = arena.allocate(cudaStream_t);
+      var returnValue = cuvsStreamGet(cuvsRes, stream);
+      checkCuVSError(returnValue, "cuvsStreamGet");
+
+      // TieredIndex REQUIRES device memory - allocate it
+      MemorySegment datasetD = arena.allocate(C_POINTER);
```
mythrocks:
```suggestion
      MemorySegment datasetD = localArena.allocate(C_POINTER);
```

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/TieredIndexImpl.java
```diff
@@ -0,0 +1,665 @@
...
+      long datasetSize = C_FLOAT_BYTE_SIZE * rows * cols;
+      returnValue = cuvsRMMAlloc(cuvsRes, datasetD, datasetSize);
+      checkCuVSError(returnValue, "cuvsRMMAlloc");
+
+      MemorySegment datasetDP = datasetD.get(C_POINTER, 0);
+
+      // Copy host to device
+      returnValue = cudaMemcpy(datasetDP, hostDataSeg, datasetSize, 1); // cudaMemcpyHostToDevice
```
mythrocks: We can use `Util.cudaMemcpy` here instead.  It's a little less opaque:
```suggestion
      returnValue = Util.cudaMemcpy(datasetDP, hostDataSeg, datasetSize, CudaMemcpyKind.HOST_TO_DEVICE); 
```

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/TieredIndexImpl.java
```diff
@@ -0,0 +1,665 @@
...
+      checkCudaError(returnValue, "cudaMemcpy");
+
+      // Create tensor from device memory
+      long datasetShape[] = {rows, cols};
```
mythrocks:
```suggestion
      long[] datasetShape = {rows, cols};
```

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/TieredIndexImpl.java
```diff
@@ -0,0 +1,665 @@
...
+      // Get host data
+      MemorySegment hostDataSeg =
+          dataset != null
+              ? ((DatasetImpl) dataset).seg
+              : Util.buildMemorySegment(resources.getArena(), vectors);
+
+      Arena arena = resources.getArena();
+      long cuvsRes = resources.getMemorySegment().get(cuvsResources_t, 0);
+      MemorySegment stream = arena.allocate(cudaStream_t);
+      var returnValue = cuvsStreamGet(cuvsRes, stream);
```
mythrocks: This isn't required.  We seem to be fetching the CUDA stream here.  But we don't use it downstream.

The stream-sync farther down in this function uses the resources-handle for sync.  The stream need not be fetched explicitly.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/TieredIndexImpl.java
```diff
@@ -0,0 +1,665 @@
...
+
+      // Create tensor from device memory
+      long datasetShape[] = {rows, cols};
+      MemorySegment datasetTensor = prepareTensor(arena, datasetDP, datasetShape, 2, 32, 2, 2, 1);
+
+      MemorySegment index = arena.allocate(cuvsTieredIndex_t);
+      returnValue = cuvsTieredIndexCreate(index);
+      checkCuVSError(returnValue, "cuvsTieredIndexCreate");
+
+      returnValue = cuvsStreamSync(cuvsRes);
```
mythrocks, line 183: I'm still reviewing this part, in light of the findings in #1045.  

Now that I think about it, this version of the code might be right, and the precautions in #1045 might not be required. 

I've run this through `compute-sanitizer`.  I'm unable to find any mem-check violations.

I'll update here, after thinking through it again.

java/cuvs-java/src/main/java/com/nvidia/cuvs/TieredIndex.java
punAhuja: So tiered index supports other algoriths like IVF-PQ and IVF-FLAT but we have only implemented it for CAGRA for now. We can add support for more algorithms in the future.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/TieredIndexImpl.java
mythrocks: Yeah, I'm now thinking this part is actually right.  We can revisit this if required.

java/cuvs-java/src/main/java/com/nvidia/cuvs/TieredIndex.java
punAhuja: This is just a getter method for the index parameters passed during tiered index creation, can be used by the user.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/TieredIndexImpl.java
punAhuja: Resolved.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/TieredIndexImpl.java
punAhuja: Resolved.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/TieredIndexImpl.java
punAhuja: Resolved.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/TieredIndexImpl.java
punAhuja: Resolved.

java/cuvs-java/src/main/java/com/nvidia/cuvs/TieredIndex.java
punAhuja: Resolved.

java/cuvs-java/src/main/java/com/nvidia/cuvs/TieredIndex.java
punAhuja: On further discussion,  I think we don't need to keep this. Since We deserialize from disk, these parameters will be null.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/TieredIndexImpl.java
mythrocks: So this part of the code should be alright.  The problem is in the next section.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/TieredIndexImpl.java
```diff
@@ -0,0 +1,649 @@
...
+          prepareTensor(localArena, datasetDP, datasetShape, 2, 32, 2, 2, 1);
+
+      MemorySegment index = arena.allocate(cuvsTieredIndex_t);
+      returnValue = cuvsTieredIndexCreate(index);
+      checkCuVSError(returnValue, "cuvsTieredIndexCreate");
+
+      returnValue = cuvsStreamSync(cuvsRes);
+      checkCuVSError(returnValue, "cuvsStreamSync");
+
+      returnValue = cuvsTieredIndexBuild(cuvsRes, indexParamsMemorySegment, datasetTensor, index);
```
mythrocks: It is incorrect to pass the `index` in the call to `cuvsTieredIndexBuild()`.

`index` at this point is the `MemorySegment` equivalent of a `cuvsTieredIndex_t*` (or a `cuvsTieredIndex**`). 

`cuvsTieredIndexBuild` takes a `cuvsTieredIndex_t` (i.e. `cuvsTieredIndex*`).  Passing the index directly to this function is likely to cause `__device__` memory corruptions.  We should instead pull out the `C_POINTER` that was written to `index`, and pass that as a `MemorySegment` here instead.

Something like:
```java 
var actualIndexPtr = index.get(cuvsTieredIndex_t, 0);
// ...
cuvsTieredIndexBuild(cuvsRes, indexParamsMemorySegment, datasetTensor, actualIndexPtr);
```

Please refer to how it's now done for [`BruteForceIndexImpl.java`](https://github.com/rapidsai/cuvs/blob/26ca5d2ec8f1a968fe1be2cd67050d034ef92fec/java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/BruteForceIndexImpl.java#L372).

mythrocks (changes requested):
Need to fix how the tiered `index` `MemorySegment` is used to build the tiered index.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/TieredIndexImpl.java
punAhuja: Have pushed a change fixing this.

mythrocks:
I'm doing another pass on this PR.  It looks good so far.  I hope to approve this shortly.

java/cuvs-java/src/test/java/com/nvidia/cuvs/TieredIndexIT.java
```diff
@@ -0,0 +1,254 @@
...
+      TieredIndexQuery query1 =
+          new TieredIndexQuery.Builder()
+              .withTopK(1)
+              .withQueryVectors(queries)
+              .withSearchParams(
+                  new CagraSearchParams.Builder(resources).withMaxIterations(20).build())
+              .build();
+
+      SearchResults results1 = index.search(query1);
+      assertEquals(1, results1.getResults().get(0).size());
```
mythrocks:
```suggestion
      assertEquals(1, results1.getResults().getFirst().size());
```
Is there a reason we aren't checking the result values, instead of the right number of results?

java/cuvs-java/src/test/java/com/nvidia/cuvs/TieredIndexIT.java
```diff
@@ -0,0 +1,254 @@
...
+                  new CagraSearchParams.Builder(resources).withMaxIterations(20).build())
+              .build();
+
+      SearchResults results3 = index.search(query3);
+      assertEquals(3, results3.getResults().get(0).size());
+    }
+  }
+
+  /**
+   * Test prefilter functionality with debug logging
```
mythrocks: I'm not sure I follow why "with debug logging" is mentioned.  We seem to be `info()` logging in the test.  What have I missed?

Also, could we please drop the test-logging severity (to `debug`), so that the test screen isn't clobbered with results?  Ideally, the console would be left uncluttered, and the tests run silently.

We should crank up the severity when debugging a problem, just not by default.

java/cuvs-java/src/test/java/com/nvidia/cuvs/TieredIndexIT.java
```diff
@@ -0,0 +1,254 @@
...
+
+      TieredIndexQuery queryWithoutFilter =
+          new TieredIndexQuery.Builder()
+              .withTopK(3)
+              .withQueryVectors(queryVectors)
+              .withSearchParams(searchParams)
+              .build();
+
+      SearchResults resultsWithoutFilter = index.search(queryWithoutFilter);
+      log.info("Results WITHOUT prefilter: {}", resultsWithoutFilter.getResults());
```
mythrocks:
```suggestion
      log.debug("Results WITHOUT prefilter: {}", resultsWithoutFilter.getResults());
```

mythrocks: Likewise, elsewhere too.  The test screen is quite noisy at the moment.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/TieredIndexImpl.java
```diff
@@ -0,0 +1,651 @@
...
+   * Gets the index type
+   *
+   * @return the index type
+   */
+  @Override
+  public TieredIndexType getIndexType() {
+    TieredIndexType indexType =
+        tieredIndexParameters != null && tieredIndexParameters.getCagraParams() != null
+            ? TieredIndexType.CAGRA
+            : TieredIndexType.CAGRA; // Default to CAGRA for now
```
mythrocks:
```suggestion
    TieredIndexType indexType = TieredIndexType.CAGRA; // Default to CAGRA for now
```

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/TieredIndexImpl.java
mythrocks: Thanks for fixing this bit.

java/cuvs-java/src/test/java/com/nvidia/cuvs/TieredIndexIT.java
punAhuja: Added more asserts to check the values.

java/cuvs-java/src/test/java/com/nvidia/cuvs/TieredIndexIT.java
punAhuja: Cleaned up the test, made info logs to debug logs.

java/cuvs-java/src/test/java/com/nvidia/cuvs/TieredIndexIT.java
punAhuja: Cleaned up the test and removed unnecessary comments, and made the info logs to debug logs.

cjnolet:
Linking https://github.com/cjnolet/nv_elastic/issues/11

mythrocks:
Apologies, I haven't gotten a moment to review the latest change.  I'll go over it shortly.

mythrocks (approved):
Alright, this looks a whole lot better.  Thank you for your patience, and for incorporating the review changes.

mythrocks:
I've merged this change.  Thank you for this one, @punAhuja.

PR #1032, opened 2025-06-19
Improve Vamana index build performance and recall

bkarsin:
Series of optimizations for the Vamana index build implementation including:
- Reduce shared memory usage by the main kernels: GreedySearch and RobustPrune
- Replace cub::blockSort with manual merge that reduces shared memory and improves perf
- Remove other cub Sort calls to separate kernel to improve resource utilization
- Other changes to resource allocation and management to improve performance

Other changes that significantly improve recall (closing most of the gap with CPU build and giving the option to further improve recall with parameters):
- Re-worked the RobustPrune kernel to use a multi-pass "occlusion" approach. This closes 60-80% of the recall gap seen between previous GPU Vamana and CPU DiskANN.
- Changed the "vamana_iters" parameter to a float to allow fractional iterations. This lets us precisely control recall vs. performance to close the gap (or surpass CPU recall). 

Further optimizations are underway, but this first round address many needs of users to both improve performance and recall.


cpp/src/neighbors/detail/vamana/greedy_search.cuh
```diff
@@ -64,6 +65,27 @@ __forceinline__ __device__ void sort_visited(
 
 namespace {
 
+template <typename T,
+          typename accT,
+          typename IdxT     = uint32_t,
+          typename Accessor = raft::host_device_accessor<std::experimental::default_accessor<T>,
```
achirkin: The Accessor template parameter doesn't seem to be used within the kernel. Can we remove it?

cpp/include/cuvs/neighbors/vamana.hpp
```diff
@@ -60,7 +60,7 @@ struct index_params : cuvs::neighbors::index_params {
    * literature **/
   uint32_t visited_size = 64;
   /** Number of Vamana vector insertion iterations (each iteration inserts all vectors). */
-  uint32_t vamana_iters = 1;
+  float vamana_iters = 1.0;
```
cjnolet, line 74: How does a fraction of an iteration work here?

cpp/src/neighbors/detail/vamana/greedy_search.cuh
cjnolet: +1 to removing unnecessary template params.

cpp/src/neighbors/detail/vamana/vamana_structs.cuh
```diff
@@ -51,19 +51,21 @@ struct __align__(16) DistPair {
...
 
-  __device__ __host__ DistPair& operator=(const DistPair& other)
-  {
-    dist = other.dist;
-    idx  = other.idx;
-    return *this;
-  }
+  /*
+    __device__ __host__ DistPair& operator=(const DistPair& other)
+    {
```
cjnolet: Reminder to remove commented out code.

cpp/tests/neighbors/ann_vamana.cuh
```diff
@@ -63,6 +63,7 @@ struct AnnVamanaInputs {
   int visited_size;
   double max_fraction;
   cuvs::distance::DistanceType metric;
+//  float insert_iters;
```
cjnolet: Reminder to remove commented out code

examples/cpp/CMakeLists.txt
```diff
@@ -36,24 +36,24 @@ set(BUILD_CUVS_C_LIBRARY OFF)
 include(../cmake/thirdparty/get_cuvs.cmake)
 
 # -------------- compile tasks ----------------- #
-add_executable(BRUTE_FORCE_EXAMPLE src/brute_force_bitmap.cu)
-add_executable(CAGRA_EXAMPLE src/cagra_example.cu)
-add_executable(CAGRA_PERSISTENT_EXAMPLE src/cagra_persistent_example.cu)
-add_executable(DYNAMIC_BATCHING_EXAMPLE src/dynamic_batching_example.cu)
-add_executable(IVF_FLAT_EXAMPLE src/ivf_flat_example.cu)
-add_executable(IVF_PQ_EXAMPLE src/ivf_pq_example.cu)
+#add_executable(BRUTE_FORCE_EXAMPLE src/brute_force_bitmap.cu)
```
cjnolet: Reminder to uncomment this. Perhaps we can add a `--build-examples` flag to `build.sh` that allows the user to specify which examples they want to build (we already have this for benchmarks, for example).

cpp/src/neighbors/detail/vamana/greedy_search.cuh
```diff
@@ -117,13 +137,14 @@ __global__ void GreedySearchKernel(
...
+  size_t smem_offset = 0;
 
   T* s_coords = reinterpret_cast<T*>(&smem[smem_offset]);
   smem_offset += (dim + align_padding) * sizeof(T);
 
-  Node<accT>* topk_pq = reinterpret_cast<Node<accT>*>(&smem[smem_offset]);
-  smem_offset += topk * sizeof(Node<accT>);
+  Node<accT>* topk_pq = &topk_pq_mem[blockIdx.x * topk];
+  //  Node<accT>* topk_pq = reinterpret_cast<Node<accT>*>(&smem[smem_offset]); // Used to test
+  //  scenarios using more shared memory smem_offset += topk * sizeof(Node<accT>);
```
jamxia155: Remove commented code.

cpp/src/neighbors/detail/vamana/greedy_search.cuh
```diff
@@ -245,6 +266,7 @@ __global__ void GreedySearchKernel(
         if (neighbor_array[j] == raft::upper_bound<IdxT>())
           atomicMin(&num_neighbors, (int)j);  // warp-wide min to find the number of neighbors
       }
+      //      __syncthreads();
```
jamxia155: Remove commented code.

cpp/src/neighbors/detail/vamana/robust_prune.cuh
```diff
@@ -149,101 +151,163 @@ __global__ void RobustPruneKernel(
...
+          occlusion_list[pass_start] = raft::lower_bound<float>();  // Mark as "accepted"
+          accept_count++;
 
-        int tot_size = degree + query_list[i].size;
-        for (int k = j + 1; k < tot_size; k++) {
-          T* mem_ptr = const_cast<T*>(&dataset((size_t)new_nbh_list[k].idx, 0));
-          if (new_nbh_list[k].idx != raft::upper_bound<IdxT>()) {
-            accT dist_starprime = dist<T, accT>(s_query.coords, mem_ptr, dim, metric);
-            // TODO - create cosine and selector fcn
+          // Update rets of the occlusion list
```
jamxia155: *rest?

cpp/src/neighbors/detail/vamana/robust_prune.cuh
```diff
@@ -149,101 +151,163 @@ __global__ void RobustPruneKernel(
...
+              accT djk     = dist<T, accT>(cand_ptr, k_ptr, dim, metric);
+              accT new_occ = (float)(new_nbh_list[occId].dist / djk);
 
-            if (threadIdx.x == 0 && alpha * dist_starprime <= new_nbh_list[k].dist) {
-              new_nbh_list[k].idx = raft::upper_bound<IdxT>();
+              // USED TO DEBUG
+              //	      if(threadIdx.x==0 && i==0) {
+              //                printf("pass:%d, occId:%d, dist:%0.3f, djk:%0.3f, new_occ:%0.3f\n",
+              //                pass_start, occId, new_nbh_list[occId].dist, djk, new_occ);
+              //	      }
```
jamxia155: Remove debug code.

cpp/src/neighbors/detail/vamana/robust_prune.cuh
```diff
@@ -149,101 +151,163 @@ __global__ void RobustPruneKernel(
...
-      if (threadIdx.x == 0) { query_list[i].size = writeId; }
 
-      __syncthreads();
-      for (int j = writeId + threadIdx.x; j < degree;
-           j += blockDim.x) {  // Zero out any unfilled neighbors
-        query_list[i].ids[j]   = raft::upper_bound<IdxT>();
-        query_list[i].dists[j] = raft::upper_bound<accT>();
+      // Move all "accepted" candidates to front of list and zero out the rest
+      int out_idx = 1;
+      for (int read_idx = 1; out_idx < accept_count; read_idx++) {
```
jamxia155: Can we put this loop in an `if (threadIdx.x == 0)` branch to make it clear that it's a sequential operation that only needs to run on one thread?

cpp/src/neighbors/detail/vamana/robust_prune.cuh
```diff
@@ -149,101 +151,163 @@ __global__ void RobustPruneKernel(
...
+        if (occlusion_list[read_idx] == raft::lower_bound<float>()) {  // If it is "accepted"
+          new_nbh_list[out_idx].idx  = new_nbh_list[read_idx].idx;
+          new_nbh_list[out_idx].dist = new_nbh_list[read_idx].dist;
+          out_idx++;
+        }
       }
+      for (int out_idx = accept_count; out_idx < degree; out_idx++) {
```
jamxia155: Seems to me that this for loop doesn't have to be run sequentially, as long as it is not started until the previous loop has completed. Why not let the threads work on this cooperatively i.e. each thread updates a successive element in `new_nbh_list`?

cpp/src/neighbors/detail/vamana/vamana_build.cuh
```diff
@@ -157,116 +174,185 @@ void batched_insert_vamana(
...
+  double rev_time          = 0.0;
+  double batch_prune       = 0.0;
+#endif
 
   // Random medoid has minor impact on recall
   // TODO: use heuristic for better medoid selection, issue:
   // https://github.com/rapidsai/cuvs/issues/355
-  *medoid_id = rand() % N;
+  //  *medoid_id = rand() % N;
+  *medoid_id = 123742;
```
jamxia155: Should this be changed back to random?

cpp/src/neighbors/detail/vamana/vamana_build.cuh
```diff
@@ -157,116 +174,185 @@ void batched_insert_vamana(
...
-      create_reverse_edge_list<accT, IdxT>
-        <<<num_blocks, blockD, 0, stream>>>(query_list_ptr.data_handle(),
-                                            step_size,
-                                            degree,
-                                            edge_src.data_handle(),
-                                            edge_dest.data_handle());
-
+  // Loop through batches and call the insert and prune kernels - can insert > N times based on
+  // iters parameter
+  for (int start = 0; start < (int)(insert_iters * (float)N);) {
```
jamxia155, line 240: For the termination condition, if `insert_iters * (float)N` happens to be, say, 0.5, do we want the iteration to continue? Currently in such a case, casting 0.5 to `int` will truncate it to 0 so the loop will terminate, but I'm not sure if that is the intention here.

cpp/src/neighbors/detail/vamana/vamana_build.cuh
```diff
@@ -157,116 +174,185 @@ void batched_insert_vamana(
...
+#if KERNEL_TIMING
+    start_t = std::chrono::system_clock::now();
+#endif
+
+    if (start + step_size > (int)(insert_iters * (float)N)) {
+      step_size = (int)(insert_iters * (float)N) - start;
+    }
+    if (start + step_size > N) {
+      int new_size = N - start;
+      step_size    = new_size;
```
jamxia155: How about just `step_size = N - start`?

cpp/src/neighbors/detail/vamana/vamana_build.cuh
```diff
@@ -72,6 +72,8 @@ void create_insert_permutation(std::vector<IdxT>& insert_order, uint32_t N)
   }
 }
 
+#define KERNEL_TIMING 0
```
jamxia155: Should we keep any of the `KERNEL_TIMING` code?

cpp/src/neighbors/detail/vamana/vamana_build.cuh
```diff
@@ -157,116 +174,185 @@ void batched_insert_vamana(
...
+
+    // compute prefix sums of query_list sizes - TODO parallelize prefix sums
+    auto d_total_edges = raft::make_device_mdarray<int>(
+      res, raft::resource::get_workspace_resource(res), raft::make_extents<int64_t>(1));
+    prefix_sums_sizes<accT, IdxT>
+      <<<1, 1, 0, stream>>>(query_list, step_size, d_total_edges.data_handle());
+
+    int total_edges;
+    raft::copy(&total_edges, d_total_edges.data_handle(), 1, stream);
+    RAFT_CUDA_TRY(cudaStreamSynchronize(stream));
```
jamxia155: This stream sync may be unnecessary as [copy from device memory to host is synchronous](https://docs.nvidia.com/cuda/cuda-runtime-api/api-sync-behavior.html#api-sync-behavior).

cpp/src/neighbors/detail/vamana/vamana_build.cuh
```diff
@@ -276,100 +362,167 @@ void batched_insert_vamana(
...
+#endif
+
+    start += step_size;
+    if (start >= N) {
+      start = 0;
+      insert_iters -= 1.0;
+      step_size = max_batchsize;
+    }
+    step_size *= base;
+    if (step_size > max_batchsize) { step_size = max_batchsize; }
```
jamxia155: nit: `step_size = min(step_size, max_batchsize)` is more concise.

cpp/src/neighbors/detail/vamana/vamana_structs.cuh
```diff
@@ -80,13 +82,24 @@ __device__ __host__ void swap(DistPair<IdxT, accT>* a, DistPair<IdxT, accT>* b)
...
 struct CmpDist {
+  __device__ bool operator()(const DistPair<IdxT, accT>& lhs, const DistPair<IdxT, accT>& rhs)
+  {
+    return lhs.dist < rhs.dist;
+  }
+};
+
+// Structure to sort by distance
+/*
+struct CmpDist2 {
```
jamxia155: Commented-out code.

cpp/src/neighbors/detail/vamana/vamana_structs.cuh
```diff
@@ -393,7 +407,12 @@ __global__ void write_graph_edges_kernel(raft::device_matrix_view<IdxT, int64_t>
 // Create src and dest edge lists used to sort and create reverse edges
 template <typename accT, typename IdxT = uint32_t>
 __global__ void create_reverse_edge_list(
-  void* query_list_ptr, int num_queries, int degree, IdxT* edge_src, IdxT* edge_dest)
+  //  void* query_list_ptr, int num_queries, int degree, IdxT* edge_src, IdxT* edge_dest)
```
jamxia155: Commented-out code.

cpp/tests/neighbors/ann_vamana.cuh
```diff
@@ -137,6 +138,7 @@ class AnnVamanaTest : public ::testing::TestWithParam<AnnVamanaInputs> {
     index_params.visited_size      = ps.visited_size;
     index_params.max_fraction      = ps.max_fraction;
     index_params.reverse_batchsize = ps.reverse_batchsize;
+//    index_params.vamana_iters      = ps.insert_iters;
```
jamxia155: Commented-out code.

cpp/tests/neighbors/ann_vamana.cuh
```diff
@@ -272,6 +274,7 @@ inline std::vector<AnnVamanaInputs> generate_inputs()
     {64, 256},  // visited_size
     {0.06, 0.1},
     {cuvs::distance::DistanceType::L2Expanded},
+//    {1, 1.5},
```
jamxia155: Commented-out code.

cpp/tests/neighbors/ann_vamana.cuh
```diff
@@ -289,6 +292,7 @@ inline std::vector<AnnVamanaInputs> generate_inputs()
     {128, 512},  // visited_size
     {0.06},
     {cuvs::distance::DistanceType::L2Expanded},
+//    {1, 1.5},
```
jamxia155: Commented-out code.

cpp/tests/neighbors/ann_vamana.cuh
```diff
@@ -307,6 +311,7 @@ inline std::vector<AnnVamanaInputs> generate_inputs()
     {256},  // visited_size
     {0.06},
     {cuvs::distance::DistanceType::L2Expanded},
+//    {1, 1.5},
```
jamxia155: Commented-out code.

cpp/tests/neighbors/ann_vamana.cuh
```diff
@@ -325,6 +330,7 @@ inline std::vector<AnnVamanaInputs> generate_inputs()
     {512, 1024},  // visited_size
     {0.06},
     {cuvs::distance::DistanceType::L2Expanded},
+//   {1, 1.5},
```
jamxia155: Commented-out code.

cpp/src/neighbors/detail/vamana/vamana_build.cuh
```diff
@@ -276,100 +362,167 @@ void batched_insert_vamana(
...
-                                                                  prune_smem_sort_size);
-
-        // Write new edge lists to graph
-        write_graph_edges_kernel<accT, IdxT><<<num_blocks, blockD, 0, stream>>>(
-          d_graph.view(), reverse_list_ptr.data_handle(), degree, reverse_batch);
+    }
+
+    DistPair<IdxT, accT>* temp_ptr = edge_dist_pair.data_handle();
+    raft::linalg::map_offset(
+      res, edge_dest.view(), [temp_ptr] __device__(size_t i) { return temp_ptr[i].idx; });
```
achirkin, line 397: Would something like this work? I think it gives a clearer intent
```suggestion
    raft::linalg::map(
      res, edge_dest.view(), [] __device__(auto x) { return x.idx; }, raft::make_const_mdspan(edge_dist_pair.view()));
```

cpp/src/neighbors/detail/vamana/robust_prune.cuh
```diff
@@ -24,6 +24,8 @@
 #include "macros.cuh"
 #include "vamana_structs.cuh"
 
+#define GRAPH_SLACK_FACTOR 1.05
```
tfeher: Is this used anywhere? If this is needed, please prefer const variables.

cpp/src/neighbors/detail/vamana/vamana_build.cuh
tfeher: `raft::copy` uses asynchronous memory copies, please keep the sync.

https://github.com/rapidsai/raft/blob/913f481544c15fd1685170a2f8f1e24f0bc14963/cpp/include/raft/util/cudart_utils.hpp#L146-L149

tfeher: `rmm::device_scalar` can help to simplify this
```suggestion
        rmm::device_scalar<int> d_total_edges(stream);
        prefix_sums_sizes<accT, IdxT>
      <<<1, 1, 0, stream>>>(query_list, step_size, d_total_edges.data());
        int total_edges = d_total_edges.value(stream);
```

tfeher (changes requested):
Thanks Ben for the PR. I am still walking through the code, but I am sharing below a few small issues.

cpp/src/neighbors/detail/vamana/vamana_build.cuh
jamxia155: According to the link I shared earlier, whether the `memcpy` function name has the "-async" suffix does not dictate the synchronization behavior:

> The API provides memcpy/memset functions in both synchronous and asynchronous forms, the latter having an "Async" suffix. This is a misnomer as each function may exhibit synchronous or asynchronous behavior depending on the arguments passed to the function.

Further down it also states that device-to-host copies are synchronous:

> For transfers from device to either pageable or pinned host memory, the function returns only once the copy has completed.

I have had past experiences where the behavior didn't make sense if we took the "-async" suffix at face value, but after coming across the above reference, I have yet to encounter any behavior to the contrary.

cpp/include/cuvs/neighbors/vamana.hpp
bkarsin: Good question - this is the main workaround for any minor recall differences from CPU to GPU. It re-iterates over the dataset and stops after the fraction is reached. For example, with at 1.1, it inserts the entire dataset, then re-inserts the first 10% of the data to improve the graph quality. Since the early vectors were inserted into an empty graph, re-inserting them has the biggest impact on recall.

cpp/src/neighbors/detail/vamana/vamana_build.cuh
bkarsin: Well, N is very large here and `insert_iters` needs to be >=1.0 (I'll add a check to make sure it's at least 1). So, it will always be >= N (dataset size).

cpp/src/neighbors/detail/vamana/vamana_build.cuh
bkarsin: I'm open to remove it or change it to `RAFT_DEBUG` statements or something. Just nice to be able to enable this during optimization and benchmarking to see which pieces of the code are bottlenecks.

cpp/include/cuvs/neighbors/vamana.hpp
tfeher: What is the acceptable range of values? I assume 1.0 is the minimum, but would be good to state it explicitly. If there are such constraint, then please enforce it with `RAFT_EXPECTS` in `vamana_build()`.

cpp/src/neighbors/detail/vamana/vamana_build.cuh
tfeher: It is a good idea to control it based on the log level.  [Here is some info](https://github.com/rapidsai/rapids-logger/blob/d4fb163306407b7e0ad67c5bfe94f6fe99447c73/README.md?plain=1#L17-L27) about compile time and runtime options.

This means you could replace the `#if KERNEL_TIMING` blocs with something like
```
#if RAFT_LOG_ACTIVE_LEVEL <= RAPIDS_LOGGER_LOG_LEVEL_DEBUG
```

Note that the default log level is [INFO](https://github.com/rapidsai/rapids-logger/blob/d4fb163306407b7e0ad67c5bfe94f6fe99447c73/include/rapids_logger/log_levels.h#L24), which means these debug messages would not be compiled.

cpp/src/neighbors/detail/vamana/vamana_build.cuh
```diff
@@ -119,6 +121,10 @@ void batched_insert_vamana(
     visited_size = power;
   }
 
+#if KERNEL_TIMING
+  auto start_t = std::chrono::system_clock::now();
+#endif
+
   // create gpu graph and set to all -1s
```
tfeher: I know it is from a previous PR, but maybe we can update this?
```suggestion
  // Initialize graph with invalid neighbor indices (raft::upper_bound<IdxT>()).
```

cpp/src/neighbors/detail/vamana/vamana_build.cuh
```diff
@@ -157,116 +174,185 @@ void batched_insert_vamana(
   int search_smem_sort_size = 0;
   int prune_smem_sort_size  = 0;
   SELECT_SMEM_SIZES(degree, visited_size);  // Sets above 2 variables to appropriate sizes
+  prune_smem_sort_size = 0;
```
tfeher: Just to make sure this is not a remnant from debugging: do we intentionally set it to zero?

cpp/src/neighbors/detail/vamana/robust_prune.cuh
tfeher, line 1: `sort_edges_and_cands()` is not used anywhere. Can we remove that, or is it planned to be used for further optimizations?

cpp/src/neighbors/detail/vamana/vamana_build.cuh
tfeher: It looks to me `prune_smem_sort_size` is not really used. Can we just remove it and simplify the `SELECT_SMEM_SIZES` macro?

cpp/src/neighbors/detail/vamana/vamana_build.cuh
```diff
@@ -157,116 +174,185 @@ void batched_insert_vamana(
...
-  if (prune_smem_sort_size == 0) {  // If sizes not supported, smem sizes will be 0
-    RAFT_FAIL("Vamana graph parameters not supported: graph_degree=%d, visited_size:%d\n",
-              degree,
-              visited_size);
-  }
+  int prune_smem_total_size =
+    prune_smem_sort_size +  //(dim + align_padding) * sizeof(T) +
+                            //                              (degree) * sizeof(DistPair<IdxT, accT>);
+    (degree + visited_size) * sizeof(float) +  // Occlusion list
+    (degree + visited_size) * sizeof(DistPair<IdxT, accT>);
```
tfeher: In case `prune_smem_sort_size` can be removed then this becomes
```suggestion
    (degree + visited_size) * sizeof(float) +  // Occlusion list
    (degree + visited_size) * sizeof(DistPair<IdxT, accT>);
```

cpp/src/neighbors/detail/vamana/greedy_search.cuh
```diff
@@ -272,8 +294,6 @@ __global__ void GreedySearchKernel(
 
     __syncthreads();
     if (self_found) query_list[i].size--;
-
-    SEARCH_SELECT_SORT(topk);
```
tfeher, line 276: Does this mean that `sort_visited` function and the `SEARCH_SELECT_SORT` macro is not needed anymore?

cpp/src/neighbors/detail/vamana/robust_prune.cuh
```diff
@@ -129,7 +131,8 @@ __global__ void RobustPruneKernel(
   int visited_size,
   cuvs::distance::DistanceType metric,
   float alpha,
-  int sort_smem_size)
+  int sort_smem_size,
```
tfeher: Do we still need the `sort_smem_size` parameter? It looks to me that we are not sorting anymore. (Or do we plan to reenable that?).

cpp/src/neighbors/detail/vamana/vamana_build.cuh
```diff
@@ -157,116 +174,185 @@ void batched_insert_vamana(
...
+#if KERNEL_TIMING
+    cudaDeviceSynchronize();
+    end_t           = std::chrono::system_clock::now();
+    elapsed_seconds = end_t - start_t;
+    search_time += elapsed_seconds.count();
+    start_t = std::chrono::system_clock::now();
+#endif
+
+    // SEGMENTED SORT ON QUERY_LIST
+    SortPairsKernel<T, accT, IdxT, Accessor><<<num_blocks, blockD, search_smem_sort_size, stream>>>(
```
tfeher: GitHub does a terrible job for rendering the diff here, but it looks like the main diff in the loop is to have a separate `SortPairsKernel`.

cpp/src/neighbors/detail/vamana/vamana_build.cuh
```diff
@@ -157,116 +174,185 @@ void batched_insert_vamana(
...
+      <<<num_blocks, blockD, search_smem_total_size, stream>>>(d_graph.view(),
+                                                               dataset,
+                                                               query_list_ptr.data_handle(),
+                                                               step_size,
+                                                               *medoid_id,
+                                                               visited_size,
+                                                               metric,
+                                                               queue_size,
+                                                               topk_pq_mem.data_handle());
+    // search_smem_sort_size);
```
tfeher:
```suggestion
    RAFT_CUDA_TRY(cudaPeekAtLastError());
    // search_smem_sort_size);
```

cpp/src/neighbors/detail/vamana/vamana_build.cuh
```diff
@@ -157,116 +174,185 @@ void batched_insert_vamana(
...
+      <<<num_blocks, blockD, prune_smem_total_size, stream>>>(d_graph.view(),
+                                                              dataset,
+                                                              query_list_ptr.data_handle(),
+                                                              step_size,
+                                                              visited_size,
+                                                              metric,
+                                                              alpha,
+                                                              prune_smem_sort_size,
+                                                              s_coords_mem.data_handle());
+    // SEGMENTED SORT ON QUERY_LIST
```
tfeher:
```suggestion
    RAFT_CUDA_TRY(cudaPeekAtLastError());
    // SEGMENTED SORT ON QUERY_LIST
```

cpp/src/neighbors/detail/vamana/vamana_build.cuh
```diff
@@ -157,116 +174,185 @@ void batched_insert_vamana(
...
+    SortPairsKernel<T, accT, IdxT, Accessor><<<num_blocks, blockD, search_smem_sort_size, stream>>>(
+      query_list_ptr.data_handle(), step_size, degree);
+
```
tfeher, line 308:
```suggestion
  RAFT_CUDA_TRY(cudaPeekAtLastError());

```

cpp/src/neighbors/detail/vamana/vamana_build.cuh
```diff
@@ -157,116 +174,185 @@ void batched_insert_vamana(
...
+    end_t           = std::chrono::system_clock::now();
+    elapsed_seconds = end_t - start_t;
+    prune1_time += elapsed_seconds.count();
+    start_t = std::chrono::system_clock::now();
+#endif
+
+    // Write results from first prune to graph edge list
+    write_graph_edges_kernel<accT, IdxT><<<num_blocks, blockD, 0, stream>>>(
+      d_graph.view(), query_list_ptr.data_handle(), degree, step_size);
+
```
tfeher, line 321:
```suggestion
  RAFT_CUDA_TRY(cudaPeekAtLastError());

```

cpp/src/neighbors/detail/vamana/vamana_build.cuh
```diff
@@ -157,116 +174,185 @@ void batched_insert_vamana(
...
+                                      raft::make_extents<int64_t>(total_edges));
+
+    // Create reverse edge list
+    create_reverse_edge_list<accT, IdxT>
+      <<<num_blocks, blockD, 0, stream>>>(query_list_ptr.data_handle(),
+                                          step_size,
+                                          degree,
+                                          edge_src.data_handle(),
+                                          edge_dist_pair.data_handle());
+
```
tfeher, line 363:
```suggestion
  RAFT_CUDA_TRY(cudaPeekAtLastError());

```

cpp/src/neighbors/detail/vamana/vamana_build.cuh
```diff
@@ -276,100 +362,167 @@ void batched_insert_vamana(
...
+        <<<num_blocks, blockD, 0, stream>>>(reverse_list,
+                                            edge_src.data_handle(),
+                                            edge_dest.data_handle(),
+                                            unique_indices.data_handle(),
+                                            unique_dests,
+                                            total_edges,
+                                            dataset.extent(0),
+                                            rev_start,
+                                            reverse_batch);
+
```
tfeher, line 501: RAFT_CUDA_TRY(cudaPeekAtLastError());

cpp/src/neighbors/detail/vamana/vamana_build.cuh
```diff
@@ -276,100 +362,167 @@ void batched_insert_vamana(
...
+      // Recompute distances (avoided keeping it during sorting)
+      recompute_reverse_dists<T, accT, IdxT>
+        <<<num_blocks, blockD, 0, stream>>>(reverse_list, dataset, reverse_batch, metric);
+
```
tfeher, line 506: RAFT_CUDA_TRY(cudaPeekAtLastError());

cpp/src/neighbors/detail/vamana/vamana_build.cuh
```diff
@@ -276,100 +362,167 @@ void batched_insert_vamana(
...
+      RobustPruneKernel<T, accT, IdxT>
+        <<<num_blocks, blockD, prune_smem_total_size, stream>>>(d_graph.view(),
+                                                                raft::make_const_mdspan(dataset),
+                                                                reverse_list_ptr.data_handle(),
+                                                                reverse_batch,
+                                                                visited_size,
+                                                                metric,
+                                                                alpha,
+                                                                prune_smem_sort_size,
+                                                                s_coords_mem.data_handle());
```
tfeher, line 516:
```suggestion
                                                                s_coords_mem.data_handle());
    RAFT_CUDA_TRY(cudaPeekAtLastError());

```

cpp/src/neighbors/detail/vamana/vamana_build.cuh
```diff
@@ -276,100 +362,167 @@ void batched_insert_vamana(
...
 
-    }  // Batch of inserts
+      // SEGMENTED SORT ON QUERY_LIST
+      SortPairsKernel<T, accT, IdxT, Accessor>
+        <<<num_blocks, blockD, search_smem_sort_size, stream>>>(
+          reverse_list_ptr.data_handle(), reverse_batch, degree);
```
tfeher: RAFT_CUDA_TRY(cudaPeekAtLastError());

cpp/src/neighbors/detail/vamana/vamana_build.cuh
```diff
@@ -276,100 +362,167 @@ void batched_insert_vamana(
...
 
-  }  // insert iterations
+      // Write new edge lists to graph
+      write_graph_edges_kernel<accT, IdxT><<<num_blocks, blockD, 0, stream>>>(
+        d_graph.view(), reverse_list_ptr.data_handle(), degree, reverse_batch);
+    }
+
```
tfeher, line 529: RAFT_CUDA_TRY(cudaPeekAtLastError());

cpp/src/neighbors/detail/vamana/vamana_build.cuh
```diff
@@ -157,116 +174,185 @@ void batched_insert_vamana(
...
+                                                               query_list_ptr.data_handle(),
+                                                               step_size,
+                                                               *medoid_id,
+                                                               visited_size,
+                                                               metric,
+                                                               queue_size,
+                                                               topk_pq_mem.data_handle());
+    // search_smem_sort_size);
+#if KERNEL_TIMING
+    cudaDeviceSynchronize();
```
tfeher: Please add error handling for these
```suggestion
    RAFT_CUDA_TRY(cudaDeviceSynchronize());
```

cpp/src/neighbors/detail/vamana/greedy_search.cuh
bkarsin: Removed

cpp/include/cuvs/neighbors/vamana.hpp
bkarsin: Added a check in `vamana_build()` that it's >1.0.

examples/cpp/CMakeLists.txt
bkarsin: Un-commented all of the tests, can add `--build-examples` or more convenient build script with a future PR

cpp/src/neighbors/detail/vamana/robust_prune.cuh
bkarsin: Added check around sequential loop

cpp/src/neighbors/detail/vamana/vamana_build.cuh
bkarsin: Changed it to:
```KERNEL_TIMING (RAFT_LOG_ACTIVE_LEVEL <= RAPIDS_LOGGER_LOG_LEVEL_DEBUG)```

Is this OK, or better to have the `#if` check at each place instead of defining `KERNEL_TIMING`?

File: cpp/src/neighbors/detail/vamana/vamana_build.cuh (hunk 17 above)
Comment (bkarsin) [reply cpp/src/neighbors/detail/vamana/vamana_build.cuh]: Done

File: cpp/tests/neighbors/ann_vamana.cuh (hunk 21 above)
Comment (bkarsin) [reply cpp/tests/neighbors/ann_vamana.cuh]: Fixed tests so it now tests `vamana_iterations=1 and 1.5`

File: cpp/src/neighbors/detail/vamana/vamana_build.cuh (hunk 25 above)
Comment (bkarsin) [reply cpp/src/neighbors/detail/vamana/vamana_build.cuh:397]: Changed to this, thanks for the more concise suggestion.

File: cpp/src/neighbors/detail/vamana/robust_prune.cuh (hunk 26 above)
Comment (bkarsin) [reply cpp/src/neighbors/detail/vamana/robust_prune.cuh]: Was used during some dev/testing, removed.

File: cpp/src/neighbors/detail/vamana/vamana_build.cuh (hunk 16 above)
Comment (bkarsin) [reply cpp/src/neighbors/detail/vamana/vamana_build.cuh]: Much simpler, changed to use `device_scalar`, thanks!

File: cpp/src/neighbors/detail/vamana/vamana_build.cuh (hunk 28 above)
Comment (bkarsin) [reply cpp/src/neighbors/detail/vamana/vamana_build.cuh]: Indeed this variable is no longer needed. Removed.

File: cpp/src/neighbors/detail/vamana/robust_prune.cuh
Comment (bkarsin) [reply cpp/src/neighbors/detail/vamana/robust_prune.cuh:1]: Removed `sort_edges_and_cands()`

File: cpp/src/neighbors/detail/vamana/greedy_search.cuh (hunk 30 above)
Comment (bkarsin) [reply cpp/src/neighbors/detail/vamana/greedy_search.cuh:276]: Still used by the new `SortPairsKernel()`, so need to keep this macro and function.

File: cpp/src/neighbors/detail/vamana/robust_prune.cuh (hunk 31 above)
Comment (bkarsin) [reply cpp/src/neighbors/detail/vamana/robust_prune.cuh]: Removed `sort_smem_size` from the `RobustPrune` kernel.

File: cpp/src/neighbors/detail/vamana/vamana_build.cuh (hunk 32 above)
Comment (bkarsin) [reply cpp/src/neighbors/detail/vamana/vamana_build.cuh]: Yes, I separated out this call to `CUB::BlockSort` into this new kernel to avoid the main `GreedySearch` and `RobustPrune` kernels from needing so much shared memory. Now just this small `SortPairsKernel` does this segmented sort" of the candidate list.

File: cpp/src/neighbors/detail/vamana/vamana_build.cuh (hunk 33 above)
Comment (bkarsin) [reply cpp/src/neighbors/detail/vamana/vamana_build.cuh]: Thanks for the suggestion, added this error checking after each kernel call.

File: cpp/src/neighbors/detail/vamana/robust_prune.cuh (hunk 11 above)
Comment (bkarsin) [reply cpp/src/neighbors/detail/vamana/robust_prune.cuh]: As suggested, parallelized this loop across the warp.

File: cpp/src/neighbors/detail/vamana/vamana_build.cuh [hunk 44]
```diff
@@ -122,7 +145,11 @@ void batched_insert_vamana(
     visited_size = power;
   }
 
-  // create gpu graph and set to all -1s
+#if KERNEL_TIMING
+  auto start_t = std::chrono::system_clock::now();
+#endif
+
+  // Initialize graph with valid neighbor indices (raft::upper_bound<IdxT>()).
```
tfeher: Isn't that supposed to be invalid indices?
```suggestion
  // Initialize graph with invalid neighbor indices (raft::upper_bound<IdxT>()).
```

tfeher (approved):
Thanks Ben for addressing the issues. The PR looks good to me, there is just a small issue left with one of the code comments.

achirkin (approved):

cjnolet:
@bkarsin looks like we have some style checker errors. Can you run precommit on your changes?

PR #1033, opened 2025-06-19
[Java] Add Java API benchmarks

ldematte:
This PR introduces a maven project configured with JMH, so we can write (micro-)benchmarks to test out different implementation choices and alternatives and their relative performances.

As an example, the PR introduces 2 benchmarks for the Cagra index API: one testing the performances of building and serializing, the other the performances of building an index from a dataset in heap memory. I plan to follow up with PRs that introduce different ways to serialize an index, and different way of specify an input dataset (off-heap memory), adding corresponding tests to see what are the performance gains.


ChrisHegarty:
Thanks @ldematte - jmh is the standard and accepted way to do micro-benchmarking for java progs, it will be invaluable to have the infrastructure and set of micro-benches here.

java/.gitignore
```diff
@@ -1,9 +1,8 @@
+*.iml
+target/
+jextract-22/
+openjdk-22-jextract*
 # cuvs-java
-/cuvs-java/target/
```
mythrocks, line 2: Was it intentional to remove this?

java/benchmarks/pom.xml
```diff
@@ -0,0 +1,154 @@
...
+ * limitations under the License.
+ */
+-->
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+
+    <groupId>com.nvidia.cuvs</groupId>
+    <artifactId>benchmarks</artifactId>
+    <!--CUVS_JAVA#VERSION_UPDATE_MARKER_START--><version>25.08.0</version><!--CUVS_JAVA#VERSION_UPDATE_MARKER_END-->
```
mythrocks, line 24: How interesting!  I hadn't realized that `ci/release/update-version.sh` is written to work on sub-directories.

java/benchmarks/src/main/java/com/nvidia/cuvs/CagraIndexBenchmarks.java
```diff
@@ -0,0 +1,114 @@
...
+
+            // Saving the index on to the disk.
+            var indexFilePath = Path.of(UUID.randomUUID() + ".cag");
+            try (var outputStream = Files.newOutputStream(indexFilePath)) {
+                index.serialize(outputStream);
+            }
+
+            // Cleanup
+            Files.deleteIfExists(indexFilePath);
+            index.destroyIndex();
```
mythrocks, line 89: Nit:  I'm really leaning towards making indexes `AutoCloseable`.  These tasks feel like they should be `finally {}`.

mythrocks (approved):
LGTM.

mythrocks:
This change is now merged.  Thank you, @ldematte.

PR #1034, opened 2025-06-19
[Java] Add Dataset based on `MemorySegment`

ldematte:
This PR adds the ability to define a Dataset directly over a MemorySegment, "wrapping" it instead of allocating a new one.

- Depends on https://github.com/rapidsai/cuvs/pull/1033 and https://github.com/rapidsai/cuvs/pull/1024
- ~~The new API has a `Object memorySegment` parameter, as we target Java 21 for the API (but 22 for the implementation); it works but it's definitely a hack and we need to sort this out~~
   - As discussed, we want to keep targeting Java 21 for the API. This means the API will return a `MethodHandle`, and the Java 22 implementation will use it to return a factory method to build a Dataset from a MemorySegment.
   - This factory method can then be used as shown in the tests (see the `DatasetHelper` convenience class/method).
- Benchmarks show a sizeable speedup -- it is still tiny related to the "big picture" (index build time), but there is an improvement and above all we avoid a whole new copy of the input data (halving the memory requirements).

Fixes https://github.com/rapidsai/cuvs/issues/698


ldematte:
I've refactored the code a bit to better handle MemorySegment allocation/deallocation; now I am able to benchmark Dataset creation separately:
```

Benchmark                                          (dims)  (size)  Mode  Cnt      Score       Error  Units
CagraIndexBenchmarks.testDatasetFromHeap             1024     100  avgt    5  45798.418 ± 12296.887  ns/op
CagraIndexBenchmarks.testDatasetFromMemorySegment    1024     100  avgt    5      2.736 ±     0.016  ns/op
```
Of course this is a tiny part of building an index; even when I use a small index (size 100, dims 1024) the time to create the index still dominates
```
Benchmark                                           (dims)  (size)   Mode  Cnt    Score   Error  Units
CagraIndexBenchmarks.testIndexingFromHeap             1024     100  thrpt    5  123.619 ± 1.854  ops/s
CagraIndexBenchmarks.testIndexingFromMemorySegment    1024     100  thrpt    5  184.753 ± 2.229  ops/s
```
But there still is a benefit, and I have the poorest of the GPUs on my dev machine (it's a 2070 mobile).

ldematte:
When I bump up data size (`-p size=N`), of course the margin becomes smaller and smaller as the index build time dominates; still, if the data comes from off-heap memory (e.g. memory mapped file), we potentially use half the memory for the input data (or 1/3 in some cases), which I think is the key point.

ChrisHegarty:
Thanks for doing this @ldematte, I think that this is case for revisiting the Java 21/22 decision that we previously made. I'll take a closer look and give it some thought.

ldematte:
@ChrisHegarty indirection via a MH costs a little, but it is still a big improvement (and we do avoid yet another memory copy)
```
Benchmark                                          (dims)  (size)  Mode  Cnt      Score      Error  Units
CagraIndexBenchmarks.testDatasetFromHeap             1024     100  avgt    5  47742.967 ± 1377.968  ns/op
CagraIndexBenchmarks.testDatasetFromMemorySegment    1024     100  avgt    5     15.025 ±    0.521  ns/op
```

java/cuvs-java/src/test/java/com/nvidia/cuvs/DatasetHelper.java
```diff
@@ -0,0 +1,39 @@
...
+ * limitations under the License.
+ */
+package com.nvidia.cuvs;
+
+import com.nvidia.cuvs.spi.CuVSProvider;
+import java.lang.foreign.MemorySegment;
+
+public class DatasetHelper {
+
+  public static Dataset fromMemorySegment(MemorySegment memorySegment, int size, int dimensions) {
```
ChrisHegarty, line 27: Should be make size a `long` here?

java/cuvs-java/src/main/java/com/nvidia/cuvs/spi/CuVSProvider.java
```diff
@@ -52,6 +53,8 @@ default Path nativeLibraryPath() {
   /** Create a {@link Dataset.Builder} instance **/
   Dataset.Builder newDatasetBuilder(int size, int dimensions);
 
+  MethodHandle newNativeDatasetBuilder();
```
ChrisHegarty, line 71: Can we please add a minimal javadoc, describing the returned method handles arguments.

java/cuvs-java/src/test/java/com/nvidia/cuvs/DatasetHelper.java
```diff
@@ -0,0 +1,39 @@
...
+    try {
+      return (Dataset)
+          CuVSProvider.provider()
+              .newNativeDatasetBuilder()
```
ChrisHegarty: I wonder if we should store the MH directly here in a static final field? (rather than access through the provider each invocation). I'm not sure if it'll help performance or not, but we should try to replicate similar to what we expect the usage to be, no?

ChrisHegarty (commented):
I think this looks very good. The use of MethodHandle is nice and allows access to MemorySegment without compiling it statically into the library code. 👍

java/cuvs-java/src/test/java/com/nvidia/cuvs/DatasetHelper.java
ldematte: All the other methods use a `int` here; I would keep this consistent and update everything to use `long` in a follow-up. Wdyt?

ChrisHegarty: Agreed. Let's consider this separately.

ChrisHegarty (approved):
LGTM

java/cuvs-java/src/main/java22/com/nvidia/cuvs/spi/JDKProvider.java
```diff
@@ -32,12 +32,32 @@
...
+    } catch (NoSuchMethodException | IllegalAccessException e) {
+      throw new RuntimeException(e);
+    }
+  }
+
+  private static Dataset createNativeDataset(
+      MemorySegment memorySegment, int size, int dimensions) {
+    return new DatasetImpl(null, memorySegment, size, dimensions);
+  }
+
```
mythrocks, line 60: I have to say, this looks quite ingenious.

We'd like to be able to construct a `DataSet` from a `MemorySegment`.  But we'd like not to have to mention `MemorySegment` in the public API.

So instead, the method that constructs the `DataSet` is made private, and we return a `MethodHandle` to it.  And that's safe to do for Java 21.

To invoke the `MethodHandle`, we'd need an instance of the `MemorySegment`.  But by then, we're committed to Java 22.

Do I understand this correctly?

ChrisHegarty: That is correct @mythrocks

ldematte:
Thank you @mythrocks, rebased!

mythrocks:
Hmm.  There seem to be `java` [test failures here](https://github.com/rapidsai/cuvs/actions/runs/16205221544/job/45757572559?pr=1034):
```
Error:  Failures: 
Error:    CagraBuildAndSearchIT.testIndexingAndSearchingFlow:145->queryAndCompare:464->CuVSTestCase.checkResults:148 expected:<[{0=0.83774555, 2=0.3590463, 3=0.038782578}, {0=0.12472608, 1=0.31918612, 2=0.21700792}, {0=0.48305473, 2=0.20332818, 3=0.047766715}, {0=0.59063464, 1=0.15224178, 3=0.5986642}]> but was:<[{0=0.23431994, 2=0.23431946, 3=0.23431946}, {0=0.68692017, 2=0.6869197, 3=0.6869197}, {0=0.33261782, 2=0.3326173, 3=0.3326173}, {0=0.3379389, 2=0.33793885, 3=0.33793885}]>
Error:  Errors: 
Error:    CagraBuildAndSearchIT.testIndexing:233->runConcurrently:85 ? Execution java.lang.AssertionError: Exception while executing runnable: java.lang.RuntimeException: cuvsCagraBuild returned 0[RAFT failure at file=/tmp/conda-bld-output/bld/rattler-build_libcuvs/work/cpp/src/neighbors/detail/cagra/graph_core.cuh line=1406: Could not generate an intermediate CAGRA graph because the initial kNN graph contains too many invalid or duplicated neighbor nodes. This error can occur, for example, if too many overflows occur during the norm computation between the dataset vectors.
Obtained 6 stack frames
#1 in /opt/conda/envs/java/lib/jvm/lib/server/../../../libcuvs.so(+0x49117d) [0x7c03a02ec17d]
#2 in /opt/conda/envs/java/lib/jvm/lib/server/../../../libcuvs.so: void cuvs::neighbors::cagra::detail::graph::optimize<unsigned int, raft::host_device_accessor<std::experimental::default_accessor<unsigned int>, (raft::memory_type)0> >(raft::resources const&, std::experimental::mdspan<unsigned int, std::experimental::extents<long, 18446744073709551615ul, 18446744073709551615ul>, std::experimental::layout_right, raft::host_device_accessor<std::experimental::default_accessor<unsigned int>, (raft::memory_type)0> >, std::experimental::mdspan<unsigned int, std::experimental::extents<long, 18446744073709551615ul, 18446744073709551615ul>, std::experimental::layout_right, raft::host_device_accessor<std::experimental::default_accessor<unsigned int>, (raft::memory_type)0> >, bool, bool) +0x1aa3 [0x7c03a0d3b4d3]
#3 in /opt/conda/envs/java/lib/jvm/lib/server/../../../libcuvs.so: cuvs::neighbors::cagra::index<float, unsigned int> cuvs::neighbors::cagra::detail::build<float, unsigned int, raft::host_device_accessor<std::experimental::default_accessor<float const>, (raft::memory_type)2> >(raft::resources const&, cuvs::neighbors::cagra::index_params const&, std::experimental::mdspan<float const, std::experimental::extents<long, 18446744073709551615ul, 18446744073709551615ul>, std::experimental::layout_right, raft::host_device_accessor<std::experimental::default_accessor<float const>, (raft::memory_type)2> >) +0xd81 [0x7c03a0d629b1]
#4 in /opt/conda/envs/java/lib/jvm/lib/server/../../../libcuvs.so: cuvs::neighbors::cagra::build(raft::resources const&, cuvs::neighbors::cagra::index_params const&, std::experimental::mdspan<float const, std::experimental::extents<long, 18446744073709551615ul, 18446744073709551615ul>, std::experimental::layout_right, raft::host_device_accessor<std::experimental::default_accessor<float const>, (raft::memory_type)2> >) +0x21 [0x7c03a0d1d761]
#5 in /opt/conda/envs/java/lib/jvm/lib/server/../../../libcuvs_c.so: cuvsCagraBuild +0x806 [0x7c04352dbb96]
#6 in [0x7c045c0a6ec7]
```

ldematte:
> Hmm. There seem to be java [test failures here](https://github.com/rapidsai/cuvs/actions/runs/16205221544/job/45757572559?pr=1034):

Yeah, it's a chicken and egg problem :) I fixed those in another PR (https://github.com/rapidsai/cuvs/pull/1089) but that PR depends on this to be merged first. These tests randomly fail because they are using resources and threading wrong.
I will cherry pick the test fixes from https://github.com/rapidsai/cuvs/pull/1089 so we can unblock the situation.

ldematte:
(BTW, the issue is already there, it's just random that it happened to fail in this PR -- better fix it sooner than later)
@mythrocks, I cherry picked the test fix, it is good to go again.

ldematte:
@mythrocks is there anything else I can do to unblock this PR?

mythrocks:
The CI builds should be fixed once #1114 is resolved.

PR #1035, opened 2025-06-20
[Java] Use MemorySegment for index serialization

ldematte:
NOTE: this WIP/draft uses code from https://github.com/rapidsai/cuvs/pull/1033 and https://github.com/rapidsai/cuvs/pull/1024  -- don't worry, I'll sort out the merge order of what we want/can merge and update accordingly.

This PR adds a new method to serialize indices (for now, only Cagra -- for demonstration purposes) to memory, instead of relying on a separate temp file. This is cleaner, saves 50% of disk space, and it's up to 10x faster.

The PR includes both the necessary Java changes and the related C API changes -- again, this draft is for discussion and to showcase how this could work; if we like the approach, my plan is to raise separate C and Java PRs (first get the C API change merged, then consume/use it in the Java API).
If we like the approach, a similar change would be needed for deserialization (`CagraIndex#build`/ `cuvsCagraDeserialize`).

Benchmarks:
```
Benchmark                                           (dims)  (size)   Mode  Cnt     Score    Error  Units
CagraSerializationBenchmarks.testSerializeToFile      1024     100  thrpt    5   561.699 ±  3.401  ops/s
CagraSerializationBenchmarks.testSerializeToMemory    1024     100  thrpt    5  6055.371 ± 33.982  ops/s
```


ldematte:
On the C side things are pretty simple: the C++ API already has an overload accepting a `std::ostream`; the new C API builds a very simple memory-based `std::ostream` using the provided pointer + size and calls down to the existing C++ API.

I'd like to refine this further; ATM the C API user (including our Java bindings) needs to "guess" a "large enough" buffer to pass to `cuvsCagraSerializeToMemory`. 
From what I see from https://github.com/rapidsai/cuvs/blob/0dd7bde8c16d3eb8da9a2fd546f540fbf7e40c81/cpp/src/neighbors/detail/cagra/cagra_serialize.cuh#L51 and 
https://github.com/rapidsai/cuvs/blob/a57227310a54b42481e20aaece72d0879f4c5b96/cpp/src/neighbors/detail/dataset_serialize.hpp#L45-L67
it is possible to know the total size of the data before even serializing it (and with no need to grab data from the GPU too).

So I was thinking we may add functions to the C and C++ API to get the serialization size. Something like
```
cuvsCagraIndexSerializationSize(res, index, include_dataset, &length)
```
Wdyt?

chatman:
> On the C side things are pretty simple: the C++ API already has an overload accepting a std::ostream; the new C API builds a very simple memory-based std::ostream using the provided pointer + size and calls down to the existing C++ API.

FYI, I was trying to chase this exact optimization, by doing a direct serialization. I've just opened a draft PR #1071, where I was passing an existing file, and have cuVS append the index to it. After the file has been written to, checking the file size would give us the offset.

ldematte:
>  I was passing an existing file, and have cuVS append the index to it

I thought about that, but it looked a bit of a "hack" to me and I preferred this approach. Let me elaborate.
On the C++ side there is a nice API that uses a `std::ostream`; on the C side it would be difficult to have something so flexible. 
To achieve our goals this can be:

1. a file name + offset + length
2. a fd + offset + length
3. a pointer + length (the caller can use a file, via mmap -- we can do that the Java side too) 
4. a function pointer `void (*f)(void*, int)`, that the C API will use any time it needs to write 1 or more bytes. 

I think 4. is the closest to a ostream but it's more complex; 1 and 2 are "weird" in a API IMO; 1 would also require re-opening on the C side something that might also already (exclusively) open for write on the Java side (by ES or Lucene); 2 could avoid that but dealing with file descriptors is difficult/impossible to do in Java. Also, 1 and 2 limit the functionality of the API to files (3/4 are more generic).

That's why I landed with 3 for this draft - it was generic enough and easy to implement. 
@benfred, wdyt? I'd really like your input here. 
I'm open to explore alternatives, including the function pointer one if you think it could make a better C API.

ldematte:
Closed in favour of https://github.com/rapidsai/cuvs/pull/1085

Issue #1037, opened 2025-06-23
[Improvement] [Java] Audit memory usage in the Java API


narangvivek10:
To check memory usage and fix any memory leaks (if existing) in the Java API.

Assigning it to self.

ldematte:
@narangvivek10 there are several places were we use `resources.getArena()`; I think these should be all examined. I started doing that for index build (especially Cagra) moving things that can be clearly short lived to use a local (confined) arena. I think there is still work to be done here, and also that some of the things should be allocated with an arena which has a more limited lifetime (e.g. the same lifetime of an index).

chatman:
Here are some thoughts:
* We should perhaps use an arena per index lifecycle, instead of global arena (arena per cuvsResource)
* Should we use some refcounting of the arena instances to track leaks during tests?
* If yes to previous, I think we can use `forbidden-apis` to ban Arena.ofConfined() etc., and instead let developers only use refcounted instances of arenas.

ldematte:
> We should perhaps use an arena per index lifecycle, instead of global arena (arena per cuvsResource)

++
Or even a local arena whenever we can -- e.g. when we are allocating pointers for C functions to fill in values.
Or in the case of functions like `search`, we probably can have a function lifecycle.
I think the goal should be to remove the cuvsResource arena entirely (or at least do not make it public).
We may not be able to do that, but that should be our north star IMO.

> Should we use some refcounting of the arena instances to track leaks during tests?
If yes to previous, I think we can use forbidden-apis to ban Arena.ofConfined() etc., and instead let developers only use refcounted instances of arenas.

I was thinking along the same lines -- we want something to trace bytes allocated and freed.
If we don't find any ready-made tool (there might be tools already doing that, e.g. using the instrumentation API profilers/debuggers use) I was thinking to do the same you say, wrap those functions and when in debug/test mode (or via some flag) enable tracing of allocations.
I think we want to do the same for function pairs like `cuvsRMMAlloc`/`cuvsRMMFree` (possibly in a AutoClosable wrapper).

ldematte:
I went on and finished looking at the usages of the cuvsResources arena, and they can all be removed. I have a draft PR that does this, reducing the scope of `Arena`s either to local (lifetime of a single function) or to a single Index (tying that data to the index lifetime).

[...]

PR #1045, opened 2025-06-24
[Review][Java] Fix random segabort/segfault/double free problems

ldematte:
The cuVS C API we are leveraging uses a common C pattern for "object creation": you pass down a pointer, and that gets populated. Often, this is a **, and memory gets allocated on the other side:
```
create(voi**ptrPtr) {
   *ptrPtr = new Something;
}
void** ptrPtr = // allocate space for a pointer
create(ptrPtr);
void* ptr = *ptrPtr;
```
In Panama, this means 
- allocating a small pointer size MemorySegment, very short lived
- call `create`
- extract a MemorySegment from the MemorySegment

Existing code sometimes skipped the extraction bit (the dereference), and/or created the structure directly instead of using the C `create` functions; this lead to some segfaults (addressing the wrong memory/wrong size) and to some double frees (the JVM freed MemorySegments we allocated via an arena, but we also called the `destroy` functions, so the C side tried to free them too).

Plus, this PR tries to reduce the lifetime of some allocations, avoiding to use the `CuVSResources` arena when it is clearly not needed -- but I did not want to expand the scope too much, so I just did the more obvious ones. Optimizing memory allocation is still a WIP, especially on the search side, but this I think will deserve its own PR.

These changes makes the test suite pass reliably - I have not seen a segmentation fault over hundreds of executions locally, and lowers the native memory occupied a little.


mythrocks:
> ```c++
> create(voi**ptrPtr) {
>    *ptrPtr = new Something;
> }
> void** ptrPtr = // allocate space for a pointer
> create(ptrPtr);
> void* ptr = *ptrPtr;
> ```

Correct me if I'm wrong, but I thought the code flow was more like:
```c++
create(void** ptrPtr) {
   *ptrPtr = new Something;
}
void* ptr; // (Automatically) allocate space for a pointer.
create(&ptr);
```

Edit:  I think I might be nitpicking.  I'll examine the crux of your fix shortly.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/BruteForceIndexImpl.java
```diff
@@ -130,8 +126,13 @@ private void checkNotDestroyed() {
   public void destroyIndex() throws Throwable {
```
mythrocks, line 127: Possibly an aside, but given that `BruteForceIndexImpl` holds a `bruteForceIndexReference`, should we consider making `BruteForceIndex` an `AutoClosable`?

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/BruteForceIndexImpl.java
```diff
@@ -130,8 +126,13 @@ private void checkNotDestroyed() {
...
     checkNotDestroyed();
     try {
-      int returnValue = cuvsBruteForceIndexDestroy(bruteForceIndexReference.getMemorySegment());
+      int returnValue = cuvsBruteForceIndexDestroy(bruteForceIndexReference.indexPtr);
       checkCuVSError(returnValue, "cuvsBruteForceIndexDestroy");
+
+      if (bruteForceIndexReference.datasetBytes > 0) {
+        returnValue = cuvsRMMFree(resources.getHandle(), bruteForceIndexReference.datasetPtr, bruteForceIndexReference.datasetBytes);
+        checkCuVSError(returnValue, "cuvsRMMFree");
```
mythrocks: Yikes.  Am I reading this right?  Was this a memory leak that you just fixed?

mythrocks: Ah, I might be mistaken.  The `IndexReference` now has two `MemorySegment`s:  One for the reference pointer, and another for the actual data.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/CuVSResourcesImpl.java
```diff
@@ -46,19 +45,22 @@ public class CuVSResourcesImpl implements CuVSResources {
...
-    arena = Arena.ofShared();
-    resourcesMemorySegment = arena.allocate(cuvsResources_t);
-    int returnValue = cuvsResourcesCreate(resourcesMemorySegment);
-    checkCuVSError(returnValue, "cuvsResourcesCreate");
+    this.arena = Arena.ofShared();
+    try (var localArena = Arena.ofConfined()) {
+      var resourcesMemorySegment = localArena.allocate(cuvsResources_t);
+      int returnValue = cuvsResourcesCreate(resourcesMemorySegment);
+      checkCuVSError(returnValue, "cuvsResourcesCreate");
+      resourceHandle = resourcesMemorySegment.get(cuvsResources_t, 0);
```
mythrocks, line 51: Ok, I think I see where you're going with this.  It makes more sense to keep the resource handle handy than the `MemorySegment` that contains it.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/CuVSResourcesImpl.java
```diff
@@ -79,9 +81,9 @@ private void checkNotDestroyed() {
    *
    * @return cuvsResources MemorySegment
```
mythrocks: Need to fix the doc here.  We would have retrieved the resource-handle here.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/BruteForceIndexImpl.java
```diff
@@ -154,44 +155,38 @@ private IndexReference build() throws Throwable {
...
       cudaMemcpy(datasetMemorySegmentP, datasetMemSegment, datasetBytes, INFER_DIRECTION);
 
-      long datasetShape[] = { rows, cols };
+      long[] datasetShape = { rows, cols };
       MemorySegment datasetTensor = prepareTensor(arena, datasetMemorySegmentP, datasetShape, 2, 32, 2, 2, 1);
 
-      MemorySegment index = arena.allocate(cuvsBruteForceIndex_t);
-
-      returnValue = cuvsBruteForceIndexCreate(index);
-      checkCuVSError(returnValue, "cuvsBruteForceIndexCreate");
```
mythrocks, line 186: This might be the crux of the problem.  Good find, @ldematte. 

C-wise, `cuvsBruteForceIndexCreate()` takes a `cuvsBruteForceIndex** p`, allocates the `cuvsBruteForceIndex`, and copies the pointer to the allocation into `*p`.

It appears that the erstwhile code was allocating a pointer (uninitialized), and passes that to `cuvsBruteForceIndexCreate()`.  I'm surprised this doesn't segfault more often.  

I *think* the new code fixes this problem.  I'll keep reviewing.

ldematte:
> Correct me if I'm wrong, but I thought the code flow was more like:
```
create(void** ptrPtr) {
   *ptrPtr = new Something;
}
void* ptr; // (Automatically) allocate space for a pointer.
create(&ptr);
```
You are absolutely correct, I was writing "C the way we can use in Java": if we cannot generate a "pointer" (a MemorySegment) referencing a stack variable, `void* ptr; create(&ptr);` becomes 
```
void** ptrPtr = (void*)malloc(sizeof(void*)); 
create(ptrPtr);
void* ptr = *ptrPtr;
free(ptrPtr);
```
or, equivalent in Java
```
try (var localArena = Arena.ofConfined) {
   MemorySegment ptrPtr = localArena.allocate(C_POINTER); // malloc
   create(ptrPtr)
   MemorySegment ptr = ptrPtr.get(...); // dereference
} // free
```
My point is: the original code was missing the indirection and dereference.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/BruteForceIndexImpl.java
ldematte: Yeah, I think there was no `cuvsRMMFree` for the data; the `IndexReference` now has two `MemorySegment`s because we need to track the "dataset pointer" too in order to free it when we are done.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/CuVSResourcesImpl.java
ldematte: Correct; there were a lot of `resources.getMemorySegment().get(cuvsResources_t, 0)` in the code, which I think it was error prone. Plus, we were keeping alive a 8 bytes pointer -- not much, I know, but why :) if we can avoid it.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/CuVSResourcesImpl.java
ldematte: Good point, I'll do it

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/BruteForceIndexImpl.java
ldematte: Yes, that was what was happening
> I'm surprised this doesn't segfault more often

Me too, but I think part of the reason is because we were allocating most of the stuff from `resources.arena`, which has a long lifetime. And I think the C heap checks if there is corruption/something that is not right at free time (that is where I was getting the segfaults).

chatman:
Excellent catch, thanks a lot @ldematte.

> Optimizing memory allocation is still a WIP, especially on the search side, but this I think will deserve its own PR.

We can use this issue, if needed: https://github.com/rapidsai/cuvs/issues/1037
Some thoughts I had around this was to perform some tests/checks for memory leaks (esp. non-heap) in our tests. Not sure how easy it would be without diving deeper into it.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/BruteForceIndexImpl.java
```diff
@@ -154,44 +155,38 @@ private IndexReference build() throws Throwable {
       MemorySegment datasetMemSegment = dataset != null? ((DatasetImpl) dataset).seg:
           Util.buildMemorySegment(resources.getArena(), vectors);
 
-      long cuvsResources = resources.getMemorySegment().get(cuvsResources_t, 0);
-      MemorySegment stream = arena.allocate(cudaStream_t);
-      var returnValue = cuvsStreamGet(cuvsResources, stream);
```
mythrocks, line 163: I don't think I understood what the old code is doing here.

I see that we call `cuvsStreamGet()`, which presumably populates the `stream` `MemorySegment` with the appropriate CUDA-stream.  I don't see how/where said `stream` was subsequently used.

It looks like @ldematte's new code just does away with the stream entirely.  Was this never required in the first place?

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/BruteForceIndexImpl.java
```diff
@@ -223,30 +218,27 @@ public SearchResults search(BruteForceQuery cuvsQuery) throws Throwable {
...
       }
 
       MemorySegment querySeg = buildMemorySegment(arena, cuvsQuery.getQueryVectors());
 
       int topk = cuvsQuery.getTopK();
-      long cuvsResources = resources.getMemorySegment().get(cuvsResources_t, 0);
+      long cuvsResources = resources.getHandle();
       MemorySegment stream = arena.allocate(cudaStream_t);
       var returnValue = cuvsStreamGet(cuvsResources, stream);
       checkCuVSError(returnValue, "cuvsStreamGet");
```
mythrocks, line 242: This part is a little confusing.  @ldematte , @narangvivek10:  After fetching the cuvs stream, I don't see where it's being used in actually executing the query. 

What am I missing here?

mythrocks (commented):
I'm generally positive on this change. 

There are a couple of minor quibbles, and some clarifying questions.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/BruteForceIndexImpl.java
chatman: I think the stream should be synchronized after the build completes. AFAIR, @cjnolet mentioned that some operations in cuVS require the stream to be synchronized. However, it isn't clear to me whether this is needed here or not. 

I think it would be good practice anyway to call a cudaStreamSync after these build calls. I'm just guessing @narangvivek10 obtained the reference to the stream to do that, but might have decided against doing it, but forgot to remove the stream reference. In any case, I think we can remove the stream reference here, and add proper stream synchronization in a separate PR.

ldematte: ++ I thought that could have been a "copy-paste" error, or some leftover. 
I see we do `cuvsStreamSync` at the end, which should be automatically sync the streams associated with the resource.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/BruteForceIndexImpl.java
ldematte: I think that could have been a "copy-paste" error, or some leftover.
We do a `cuvsStreamSync` at the end of the build, which should automatically sync the streams associated with the resource.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/BruteForceIndexImpl.java
ldematte: I was wondering exactly the same: is there a reason why this method called `destroyIndex`?  If there is no strong reason for that, I'd rename it to `close` and implement `AutoClosable`

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/BruteForceIndexImpl.java
mythrocks: I see your point about this being a copy-paste error.

I had initially thought the stream was being fetched for a `cudaStreamSynchronize()` after a `cudaMemcpy`.  My understanding is that that synchronize isn't required, at least after `cudaMemcpy`.

I'll look to @benfred and @cjnolet for guidance on `cudaStreamSynchronize()` after index builds.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/BruteForceIndexImpl.java
ldematte: I have checked the C/C++ code: the `cuvsStreamSync` we do before and after calling the _build_ functions (e.g. `cuvsBruteForceImpl`) do basically a `cuvsStreamGet` + `cudaStreamSynchronize`, so there is no need to do that manually too (it would be redundant).

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/BruteForceIndexImpl.java
ldematte: Same as above: `cuvsStreamSync` (which we call before and after the build) do a `raft::resource::sync_stream` on the main cuvs resource stream (so we effectively do a cudaStreamSynchronize(cuvsStreamGet(cuvsResources)) after the cudaMemcpy)

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/BruteForceIndexImpl.java
ldematte: I've taken a note about this, but I prefer to address it in a follow-up PR so the change can be discussed (and the PR is more focused)

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/BruteForceIndexImpl.java
mythrocks: I've been giving this some more thought.  I'm no longer convinced that this is a problem.  :/  (I wonder if I'll change my mind back again, tomorrow.)

Here is the reasoning.  Consider `cuvsRMMAlloc(void**)` at line#165 above. 
1. We allocate a `C_POINTER` (equivalent to `void*`) from the arena.  The `dataMemorySegment` we get from the allocation is the equivalent of a `void**`.
2. We pass `dataMemorySegment` to `cuvsRMMAlloc()`, which writes the address (`void*`)  into the allocated memory in `dataMemorySegment`. (i.e. ` *(dataMemorySegment) = addr;`)

The situation is similar with `cuvsBruteForceIndexCreate(cuvsCagraIndex_t*)` (i.e. `(cuvsCagraIndex**)`).
1. We allocate a `C_POINTER` (equivalent to `cuvsCagraIndex*`) from the arena.  The memorySegment functions as the `cuvsCagraIndex**`.
2. We pass the segment to `cuvsCagraIndexCreate()`.  That function allocates memory and constructs a `cuvsCagraIndex`.  It then copies the address of the index into our memorySegment.  i.e. `*(myMemorySegment) = &cuvsConstructedIndex`.

The code is correct to my eye.

mythrocks: In other words, I now think the changes in `createBruteForceIndex` are not required. 

Does my reasoning above make sense?

ldematte: Partially.

The problem is not in these 2 lines, but in the lack of getting the "filled in" pointer.
We can keep these lines here if you want, but we need to get the pointer out. Right now, we are passing `index` around as this is the "index pointer" (or better, "index handle"), but we are passing around a _pointer_ to it.
So we could just "fix" this by adding:
```
index = index.get(cuvsBruteForceIndex_t, 0);
```
If you see below, we do that with `cuvsRMMAlloc`, where we have both `queriesD` (a pointer to a pointer-sized location that `cuvsRMMAlloc` will fill) and `queriesDP` (where we read and store the pointer filled in by `cuvsRMMAlloc`). We need the same here.

Plus, I find extracting the code to a separate function (createBruteForceIndex()) has 2 advantages:
- It's not leaking the pointer-sized memory location, by using a local arena (that is the closes thing to a stack allocated variable in C). The current code retains it for the lifetime of CuVSResource. It's not much a matter of "loosing" 8 bytes, but of the code telling what it does -- otherwise one may thing we _need_ to retain this for the lifetime of the resource.
- it encapsulates this complex logic in one point. I have done the same for `cuvsRMMAlloc` in another PR, and the code is (IMO) much easier to ready, with this nitty-gritty (and admittedly confusing) details tucked away

ldematte: In fact without this change I can easily get our test suite to fail, and fail during "free" -- this happens at the end of the test in our case, as "free" is when we destroy the index or the CuVSResource. IIRC, the C/C++ runtime checks the heap during free to see if it is consistent. My suspicion here is that by passing the pointer to the index pointer (instead of the pointer), functions operating on `index` will write beyond the memory they should touch, corrupting the heap and causing free to fail.

ChrisHegarty (approved):
LGTM. Thanks for fixing this intermittent failure.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/BruteForceIndexImpl.java
mythrocks: Ah, right.  That's what I was missing.

It's not that the allocation was bad.  It is that after the allocation, the `cuvsBruteForceIndex_t*` is being treated as a `cuvsBruteForceIndex_t`.  The type-erasure on `MemorySegment` makes this hard to catch.

Excellent!  Thank you for your patience in explaining this, @ldematte.  I'll have to keep my eyes open for this sort of problem in the future.

mythrocks: As to your other refactors in this PR, I agree:  It reads much better now.

mythrocks: I should mention that for some reason, I have been unable to repro this failure locally. :/ It seems to have been intermittent in CI as well.

I've run CUDA mem-checks in the hope of catching this.  No dice.  Hopefully, this resolves the problem.

ldematte: > I should mention that for some reason, I have been unable to repro this failure locally. :/ It seems to have been intermittent in CI as well.

I think I know why: I have compiled libcuvs locally, and I think I have linked it with the _debug_ C runtime, which have additional checks on malloc and free (new/delete). That's probably why it was failing all the time for me.

PR #1049, opened 2025-06-25
[Java] Enforce Java code format standard

### LINKED ISSUE #1042: [FEA] [Java] Enforce Java code style format
To update the cuVS pre-commit configuration to enforce a Java code style format.

narangvivek10:
Configuration changes to enforce Java code format standard using [google-java-format](https://github.com/google/google-java-format) via [spotless's](https://github.com/diffplug/spotless) [maven plugin](https://github.com/diffplug/spotless/tree/main/plugin-maven#google-java-format) to comply with the [Google Java Style Guide](https://google.github.io/styleguide/javaguide.html). Spotless is a popular choice and is also used in [Apache Lucene](https://github.com/apache/lucene/blob/branch_10_1/gradle/validation/spotless.gradle)

With this change, the Java code format is checked when invoking `mvn verify` and precisely during its `compile` phase. Any violations are reported with the build failure. These reported violations can be fixed by doing `mvn spotless:apply`. Additionally, this plugin also reports missing license headers and adds them with `mvn spotless:apply`. 

@chatman FYI

Fixes #1042

### DIFF
--- a/java/cuvs-java/license-header.txt
+++ b/java/cuvs-java/license-header.txt
@@ -1,0 +1,15 @@
+/*
+ * Copyright (c) 2025, NVIDIA CORPORATION.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
--- a/java/cuvs-java/pom.xml
+++ b/java/cuvs-java/pom.xml
@@ -102,10 +102,42 @@
             <scope>test</scope>
         </dependency>
 
+        <dependency>
+            <groupId>com.diffplug.spotless</groupId>
+            <artifactId>spotless-maven-plugin</artifactId>
+            <version>2.44.5</version>
+        </dependency>
+
     </dependencies>
 
     <build>
         <plugins>
+            <plugin>
+                <groupId>com.diffplug.spotless</groupId>
+                <artifactId>spotless-maven-plugin</artifactId>
+                <version>2.44.5</version>
+                <executions>
+                    <execution>
+                        <goals>
+                            <goal>check</goal>
+                        </goals>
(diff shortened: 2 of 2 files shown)


java/examples/pom.xml
```diff
@@ -20,9 +20,47 @@
       <!-- NOTE: The version automatically gets updated when ci/release/update-version.sh is invoked. -->
       <!--CUVS_JAVA#VERSION_UPDATE_MARKER_START--><version>25.08.0</version><!--CUVS_JAVA#VERSION_UPDATE_MARKER_END-->
     </dependency>
+
+    <dependency>
+        <groupId>com.diffplug.spotless</groupId>
```
chatman, line 25: Should we leave this out of the example project? Goal of this project is to have a very simple, barebones example that a user can copy and use as a starting point, if needed. Towards that, having the spotless plugin pre-configured is nice, but not necessary. I defer to your judgement on this, though.

chatman (approved):
LGTM

java/examples/pom.xml
narangvivek10: Yes, the Spotless plugin may seem optional in examples, but the reasoning behind adding this is for helping us, especially when we are adding or modifying existing examples. As far as simplicity goes, this is just a maven dependency with a plugin configuration, so keeping this should be okay IMHO.

ldematte (commented):
Thanks for this, I really appreciate having not to guess a code format to use.
One nitpick: can we add some mention to it to the README? Perhaps adding a note on how this can be used to format your code to automatically fix all the violations (`mvn spotless::apply` or similar?)

narangvivek10:
> One nitpick: can we add some mention to it to the README? Perhaps adding a note on how this can be used to format your code to automatically fix all the violations (`mvn spotless::apply` or similar?)

Yes great idea, in fact, I was thinking that we should add a `Contributing` section in the readme file with helpful notes (including this one), and any guidelines for developers in general.

chatman:
Good idea about contributing section. I would prefer a section that
explains setting up IDE, how to debug (across the different language
layers), how to use tools like jextract, nsight, compute-sanitize etc. This
could go into the top level project with a section in Java. FYI, I still
haven't been able to setup the VSCode properly for the C/C++ codebase
(devcontainers didn't work, neither did a regular cmake based project).

mythrocks:
I have merged this PR.  Thank you for this one, @narangvivek10.

narangvivek10:
> Can we add some mention to it to the README? Perhaps adding a note on how this can be used to format your code to automatically fix all the violations (mvn spotless::apply or similar?)

This (and other documentation suggestions above) will be tracked under #1070 and will go in as a separate PR. Thanks!

Issue #1051, opened 2025-06-25
Deprecate and remove batch NN Descent code


jinsolp:
Since everything is migrated to `all_neighbors`, we can deprecate and remove batch NN Descent code (batching code specific to NN Descent) in `nn_descent_batch.cuh`

jinsolp:
closed by #1249

PR #1052, opened 2025-06-25
Deprecation notice for batching code specific to NN Descent

jinsolp:
Planning to deprecate batching code specific to NN Descent, now that we have the more general `all_neighbors` API.
Related issue: https://github.com/rapidsai/cuvs/issues/1051


cpp/src/neighbors/detail/nn_descent_batch.cuh
```diff
@@ -581,11 +581,17 @@ template <typename T,
           typename IdxT = uint32_t,
           typename Accessor =
             host_device_accessor<std::experimental::default_accessor<float>, memory_type::host>>
+[[deprecated(
+  "NN Descent batch build is deprecated and will be removed in a future release. Please use "
+  "all_neighbors::build instead.")]]
```
divyegala:
```suggestion
  "cuvs::all_neighbors::build(...) instead.")]]
```

cpp/src/neighbors/detail/nn_descent_batch.cuh
```diff
@@ -581,11 +581,17 @@ template <typename T,
...
 void batch_build(raft::resources const& res,
                  const index_params& params,
                  mdspan<const T, matrix_extent<int64_t>, row_major, Accessor> dataset,
                  index<IdxT>& global_idx)
 {
+  RAFT_LOG_WARN(
+    "NN Descent batch build is deprecated and will be removed in a future release. Please use "
+    "all_neighbors::build instead.");
```
divyegala:
```suggestion
    "cuvs::all_neighbors::build(...) instead.");
```

cjnolet:
This seems to be deprecating internal code, but it's really the external code we want to deprecate. The reasoning for deprecating is so that users of a deprecated public API know to update their code before we remove it.

cpp/src/neighbors/nn_descent.cuh
```diff
@@ -67,6 +67,9 @@ auto build(raft::resources const& res,
            raft::device_matrix_view<const T, int64_t, raft::row_major> dataset) -> index<IdxT>
 {
   if (params.n_clusters > 1) {
+    RAFT_LOG_WARN(
```
cjnolet, line 71: Please create / reference an issue in comments here in the code so that this change doesn't get lost or forgotten.

jinsolp: Done!

PR #1053, opened 2025-06-25
Improve memory efficiency for returning NN Descent distances

jinsolp:
NN descent has its distances calculated and stored in a host matrix, while the return is expected to be on device.

For the slicing operation, right now we allocate a device matrix of size `nrow x node_degree`. This PR improves memory usage by doing the slicing operation on host, and copying the results into the pre-allocated device memory output. 

This operation is done once at the end of NN Descent, so moving the slicing device computation to be computed on host should not bring noticeable perf issues.


divyegala (approved):

jinsolp:
@divyegala ready to be merged if there are no further suggestions

PR #1057, opened 2025-06-27
hnsw::from_cagra: avoid allocating the graph twice

achirkin:
When CAGRA graph is not host accessible, HNSW::from_cagra routine copies the data from it to the base layer of HNSW in two stages: first to a host matrix, then to the HNSW index object. As a result, the graph, which can be huge, is allocated in the host memory twice.
For big enough graphs this manifests itself in the program being killed by the linux kernel at the end of the index building process due to not enough system memory.

This PR avoids the staging graph buffer by using cudaMemcpy2DAsync to copy the data in an appropriate format in one go.


lowener (approved):
LGTM

PR #1064, opened 2025-06-27
Dynamically read from VERSION file for Java builds

KyleFromNVIDIA:
Noticed this while reviewing https://github.com/rapidsai/cuvs/pull/1003.


jameslamb (approved):
Seems like an improvement to me!

But I'd get another review from someone actively working on the Java bindings, like @mythrocks  ... I don't know if having `${version}` require interpolation like this would break any workflows they have (for example, if they have re-building workflows driven by an IDE instead of running a shell script).

java/cuvs-java/pom.xml
```diff
@@ -21,8 +21,7 @@
     <modelVersion>4.0.0</modelVersion>
     <groupId>com.nvidia.cuvs</groupId>
     <artifactId>cuvs-java</artifactId>
-    <!-- NOTE: The version automatically gets updated when ci/release/update-version.sh is invoked. -->
-    <!--CUVS_JAVA#VERSION_UPDATE_MARKER_START--><version>25.08.0</version><!--CUVS_JAVA#VERSION_UPDATE_MARKER_END-->
+    <version>${version}</version>
```
mythrocks, line 24: I thought I was supportive of this change until I got to this part of the change.  I don't think it's a good idea not to have a concrete version number in the project pom.

For one, every `mvn` command run by developers would now have to be suffixed with a `-Dversion=../get-version.sh`.

For another, this `pom.xml` accompanies the project artifact at the time of publication to Maven Central.  That would need the explicit version in the `pom.xml` and in the artifact jar.

mythrocks:
For the moment, I am inclined to keep the `pom.xml` static, and not dynamically generated.  The dynamism will make dev builds slightly more cumbersome.

We can revisit this if required, at a later date.

cjnolet:
@mythrocks @KyleFromNVIDIA should we close this PR in the meantime?

KyleFromNVIDIA:
Yes, closing now.

PR #1067, opened 2025-06-30
[Java] Refactor SearchResults implementation classes

ldematte:
This PR refactors the `SearchResults` implementation classes, by removing inheritance and replacing it with factory methods to create results from native data.
The current implementation classes use inheritance and polymorphism just during creation (in the ctors); this is confusing as it forces to navigate across derived/base class to understand the mechanism, and forces the base class to hold onto references to MemorySegments -- something we don't need to do.

This is preliminary work towards https://github.com/rapidsai/cuvs/issues/1037, as it makes the lifetime of those MemorySegment clearer.

### DIFF
--- a/cpp/src/cluster/detail/mst.cuh
+++ b/cpp/src/cluster/detail/mst.cuh
@@ -18,6 +18,7 @@
 
 #include "../../sparse/neighbors/cross_component_nn.cuh"
 #include <cuvs/distance/distance.hpp>
+#include <raft/core/memory_type.hpp>
 #include <raft/core/resource/cuda_stream.hpp>
 #include <raft/label/classlabels.cuh>
 #include <raft/matrix/detail/gather.cuh>
@@ -299,9 +300,7 @@ void build_sorted_mst(
   int iters        = 1;
   int n_components = cuvs::sparse::neighbors::get_n_components(color, m, stream);
 
-  cudaPointerAttributes attr;
-  RAFT_CUDA_TRY(cudaPointerGetAttributes(&attr, X));
-  bool data_on_device = attr.type == cudaMemoryTypeDevice;
+  bool data_on_device = raft::memory_type_from_pointer(X) != raft::memory_type::host;
 
   while (n_components > 1 && iters < max_iter) {
     if (data_on_device) {
--- a/python/cuvs_bench/cuvs_bench/run/__main__.py
+++ b/python/cuvs_bench/cuvs_bench/run/__main__.py
@@ -1,5 +1,5 @@
 #
-# Copyright (c) 2024, NVIDIA CORPORATION.
+# Copyright (c) 2024-2025, NVIDIA CORPORATION.
 #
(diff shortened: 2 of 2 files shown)


java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/SearchResultsImpl.java
```diff
@@ -0,0 +1,58 @@
...
+
+import java.lang.foreign.MemoryLayout;
+import java.lang.foreign.MemorySegment;
+import java.lang.foreign.SequenceLayout;
+import java.util.LinkedHashMap;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
+
+class SearchResultsImpl implements SearchResults {
```
ldematte, line 28: This has been moved, but git does not recognize it as it also changed a bit.
I moved it to this package so I was able to restrict visibility to be package-private.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/CagraSearchResults.java
```diff
@@ -27,46 +30,44 @@
...
+    List<Map<Integer, Float>> results = new LinkedList<>();
+    Map<Integer, Float> intermediateResultMap = new LinkedHashMap<>();
+    var neighboursVarHandle =
+        neighboursSequenceLayout.varHandle(MemoryLayout.PathElement.sequenceElement());
+    var distancesVarHandle =
+        distancesSequenceLayout.varHandle(MemoryLayout.PathElement.sequenceElement());
+
     int count = 0;
     for (long i = 0; i < topK * numberOfQueries; i++) {
-      long id = (long) neighboursVarHandle.get(neighboursMemorySegment, 0L, i);
```
mythrocks, line 59: Nit:  Couldn't help but notice this change.  Should this be `0` or `0L`?  `SearchResultsImpl` uses a `long` here.

mythrocks:
This change has been merged.  Thank you, @ldematte.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/CagraSearchResults.java
ldematte: True, for cleanliness. In practice, since we are declaring the type (long) instead of letting the compiler infer it (var), the compiler will initialize this correctly. But it's cleaner to use 0L to state the intent, I'll fix it in a next PR

ldematte: Oh, I see now you are referring to `neighboursVarHandle` @mythrocks 
According to `CagraIndexImpl#search`, that's a sequencelayouyt of C_INT (differently from bruteforce and hnsw, which have C_LONG). 
Is that a mistake?

PR #1068, opened 2025-06-30
Fix compilation

hhy3:

### DIFF
--- a/cpp/tests/CMakeLists.txt
+++ b/cpp/tests/CMakeLists.txt
@@ -292,10 +292,10 @@ if(TARGET cuvs::c_api)
   endif()
 
   add_executable(cuvs_c_test core/c_api.c)
-  target_link_libraries(cuvs_c_test PUBLIC cuvs::c_api)
+  target_link_libraries(cuvs_c_test PUBLIC cuvs::c_api PRIVATE stdc++ OpenMP::OpenMP_CXX NCCL::NCCL)
 
   add_executable(cuvs_c_neighbors_test neighbors/c_api.c)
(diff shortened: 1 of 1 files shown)


cjnolet:
Thanks for the contribution @hhy3. Can you please add a description that contains the conditions by which this compilation is failing (for eg. newer gcc or cmake version)? Some information about your environment would be helpful as well.

hhy3:
@cjnolet hi I just build with the instruction
```
conda env create --name cuvs -f conda/environments/all_cuda-128_arch-x86_64.yaml
conda activate cuvs
./build.sh libcuvs
```
and it reports compilation errors
gcc version: 13.3.0
cmake versoin: 4.0.3

robertmaynard (changes requested):
None of this is needed. The `OpenMP::OpenMP_CXX NCCL::NCCL` are transitive picked up from `cuvs::cuvs`.

The direct linking to `libstdc++` indicates that your compiler toolchain is broken.
Which means that `all_cuda-129_arch-x86_64.yaml` needs updates and not the CMake side.

robertmaynard:
No issues with the latest 25.08 ( `3fcaf8d6e55fb1a516d3ed5a6e49dca86273b5e3` ) with the updated `all_cuda-129_arch-x86_64.yaml`.

Most likely cause is a non bash shell which don't support the conda compiler activation logic. Therefore you build with a mixture of two toolchains ( conda, and host ) which isn't compatible.

hhy3:
@robertmaynard thanks for your reply! I'm using fish now, maybe I'll check bash

cjnolet:
@hhy3 I'm going to close this PR because it sounds like this doesn't require changes to cuVS. Please feel free to reopen if you disagree.

PR #1071, opened 2025-06-30
Adding file mode option to serialize() methods

chatman:
In serialize() methods of CAGRA, Brute Force and HNSW, it would be preferable to add the ability to append to a file (ios::app). This will be useful for Lucene, where the file written to the segment contains a header, post which the serialized index could be written, by opening the file in append mode.

Also, useful for composite files (file containing multiple indexes). For this usecase, though, a deserialize() method where indexes from the same file can be loaded using offsets+length. This is not in scope for this PR, though.

In this case, the additional file_mode parameter can have 'a' for ios::app or 'w' for ios::out.

```
cuvsError_t cuvsBruteForceSerializeWithMode(cuvsResources_t res,
                                             const char* filename,
                                             cuvsBruteForceIndex_t index,
                                             char file_mode);
```

### DIFF
--- a/cpp/cmake/modules/ConfigureCUDA.cmake
+++ b/cpp/cmake/modules/ConfigureCUDA.cmake
@@ -45,6 +45,7 @@ if(CUDA_LOG_COMPILE_TIME)
 endif()
 
 list(APPEND CUVS_CUDA_FLAGS --expt-extended-lambda --expt-relaxed-constexpr)
+list(APPEND CUVS_CUDA_FLAGS -allow-unsupported-compiler)
 list(APPEND CUVS_CXX_FLAGS "-DCUDA_API_PER_THREAD_DEFAULT_STREAM")
 list(APPEND CUVS_CUDA_FLAGS "-DCUDA_API_PER_THREAD_DEFAULT_STREAM")
 # make sure we produce smallest binary size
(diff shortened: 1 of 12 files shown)


cpp/cmake/modules/ConfigureCUDA.cmake
```diff
@@ -45,6 +45,7 @@ if(CUDA_LOG_COMPILE_TIME)
 endif()
 
 list(APPEND CUVS_CUDA_FLAGS --expt-extended-lambda --expt-relaxed-constexpr)
+list(APPEND CUVS_CUDA_FLAGS -allow-unsupported-compiler)
```
mythrocks, line 48: Question:  Why was this option required for `nvcc`?

cpp/include/cuvs/neighbors/brute_force.h
```diff
@@ -184,6 +184,34 @@ cuvsError_t cuvsBruteForceSearch(cuvsResources_t res,
...
+ * @param[in] res cuvsResources_t opaque C handle
+ * @param[in] filename the file name for saving the index
+ * @param[in] index BRUTEFORCE index
+ * @param[in] file_mode File mode: 'w' for write (ios::out), 'a' for append (ios::app)
+ *
+ */
+cuvsError_t cuvsBruteForceSerializeWithMode(cuvsResources_t res,
+                                             const char* filename,
+                                             cuvsBruteForceIndex_t index,
+                                             char file_mode);
```
mythrocks, line 199: Is the indent off by 1 here?

cjnolet:
@chatman is this still relevant given the more recent move to serializing cagra out to Lucene HNSW format directly?

cjnolet:
@chatman ping ^^

chatman:
> @chatman is this still relevant given the more recent move to serializing cagra out to Lucene HNSW format directly?

For CAGRA or Tiered index serialization, either this or #1085 is needed. My preference would be #1085, and in that case we can close this.

chatman:
Closing this in favour of #1085.

PR #1074, opened 2025-07-01
Initial deprecation of CAGRA uint8 API

lowener:
WIP
This PR will deprecate uint8 API of CAGRA.
So far the build and search operations are replaced by mapping the data to int8 to still offer some support for this data type.


cpp/src/neighbors/cagra_build_uint8.cu
```diff
@@ -14,30 +14,78 @@
...
+           raft::device_matrix_view<const uint8_t, int64_t, raft::row_major> dataset)
+  -> cuvs::neighbors::cagra::index<uint8_t, uint32_t>
+{
+  uint8_t* dataset_ptr_uint8 = const_cast<uint8_t*>(dataset.data_handle());
+  auto dataset_int8          = raft::make_device_matrix_view<int8_t, int64_t, raft::row_major>(
+    reinterpret_cast<int8_t*>(dataset_ptr_uint8), dataset.extent(0), dataset.extent(1));
+  cuvs::spatial::knn::detail::utils::mapping<int8_t> map_i8;
+  cuvs::spatial::knn::detail::utils::mapping<uint8_t> map_u8;
+  if (params.metric != cuvs::distance::DistanceType::BitwiseHamming) {
+    raft::linalg::map(handle, dataset_int8, map_i8, dataset);
```
tarang-jain, line 35: A direct conversion from uint8 to int8 like this might not be accurate. The norm and direction of a vector (x, y, z) is not the same as (and can be drastically different from) that of (x-a, y-a, z-a). So this casting could change the fundamental properties of the dataset.

tarang-jain: Oh I realized the mapping operation is actually a right shift by 1 bit rather than subtraction of 256. In that case, this should work with some minor loss of data (2*n and 2*n + 1 would be cast to the same value).

lowener: Even if a translation was used it would not impact the nearest neighbor search as long as it is applied to the whole dataset. The norm and direction would change but the translation would preserve the nearest neighbors.
Exemple: Point A [0, 3, 33] -> [-128, -125, -95]
Point B: [1, 2, 31] -> [-127, -126, -97]
Euclidean distance between A and B: sqrt(6) before and after translation

cjnolet:
Closing for now as I think this is no longer needed after the LTO changes. Please re-open if I'm wrong.

PR #1083, opened 2025-07-03
[BUG] Fix device access check in build_sorted_mst

tarang-jain:
X may be in managed memory.

### DIFF
--- a/cpp/src/cluster/detail/mst.cuh
+++ b/cpp/src/cluster/detail/mst.cuh
@@ -299,12 +299,12 @@ void build_sorted_mst(
   int iters        = 1;
   int n_components = cuvs::sparse::neighbors::get_n_components(color, m, stream);
 
-  cudaPointerAttributes attr;
-  RAFT_CUDA_TRY(cudaPointerGetAttributes(&attr, X));
-  bool data_on_device = attr.type == cudaMemoryTypeDevice;
+  cudaPointerAttributes ptr_attrs;
+  RAFT_CUDA_TRY(cudaPointerGetAttributes(&ptr_attrs, X));
+  const bool device_accessible = ptr_attrs.devicePointer != nullptr;
 
   while (n_components > 1 && iters < max_iter) {
-    if (data_on_device) {
+    if (device_accessible) {
       connect_knn_graph<value_idx, value_t>(handle, X, mst_coo, m, n, color, reduction_op);
     } else {
       connect_knn_graph<value_idx, value_t>(handle, X, mst_coo, m, n, n_components, color, metric);


jinsolp:
Did you run into an issue with the current check? Because we previously discussed this issue when we were adding a hotfix for UMAP for the 25.06 release. ([link](https://github.com/rapidsai/cuml/pull/6826)).

tarang-jain:
@jinsolp I just saw the hotfix. Yes I did run into an issue. In the cuml accel tests for hdbscan, the host side `connect_knn_graph` will be called, but the Mutual Reachabilty distance across components is actually embedded inside the reduction op, which is not handled by the host side code path at the moment (seems to be hard coded to pairwise).

Secondly, there is an edge case arising because of this code path, which is not handled by raft's mst solver. That edge case is that the mst solver is called on a graph that has only a single edge. I am opening a PR in raft to handle this.

The second one is causing the cuML optional check (https://github.com/rapidsai/cuml/issues/6948) to fail.

tarang-jain:
How about a boolean to force one code path over the other (when the X matrix is device accessible)?

jinsolp:
> How about a boolean to force one code path over the other (when the X matrix is device accessible)?

I see, and I agree. I guess this is what this PR is doing?

tarang-jain:
@jinsolp I'll update this PR. At this time, it is forcing the device route if X is device accessible, whereas for UMAP you prefer to use the host path despite it being device accessible. A boolean override variable is the only way to satisfy both these scenarios.

cpp/src/cluster/detail/mst.cuh
```diff
@@ -299,12 +299,14 @@ void build_sorted_mst(
   int iters        = 1;
   int n_components = cuvs::sparse::neighbors::get_n_components(color, m, stream);
 
-  cudaPointerAttributes attr;
-  RAFT_CUDA_TRY(cudaPointerGetAttributes(&attr, X));
-  bool data_on_device = attr.type == cudaMemoryTypeDevice;
+  // TODO: Think about how to handle this for G+H. C-style malloc'd arrays can still lead to the
+  // device route being taken.
+  cudaPointerAttributes ptr_attrs;
```
cjnolet: We have a function for this which was introduced more recently in cuda utils. We should also strive to use reusable logic for these types of things as much as possible so we can centralize these types of decisions and avoid having to make changes everywhere when logic and APIs need to be updated.

tarang-jain: Using raft::memory_type_from_pointer.

cjnolet (approved):

Issue #914, opened 2025-05-20
[BUG] IVF-FLAT distance computation in FP16 can lead to infinite distances


lowener:
IVF-Flat in FP16 accumulate distances in FP16, which can lead to infinite distances.
To ensure accurate distance computation, FP32 should be used for accumulation.
This can be reproduced by changing the uniform distribution we're using in test to a uniform between 0.0 and 100.0

## Python example
```python
from cuvs.neighbors import ivf_flat
import cupy as cp

a=cp.random.uniform(0,100,(1000,128)).astype(cp.float16)
b=cp.random.uniform(0,100,(20,128)).astype(cp.float16)

index = ivf_flat.build(ivf_flat.IndexParams(n_lists=20), a)
distances, neighbors = index.search(ivf_flat.SearchParams(n_probes=20), index, b, 1)

print(distances.copy_to_host())
# [inf, inf, inf, ...]
```

PR #1084, opened 2025-07-04
Fix fp16 overflow

hhy3:
This PR fixes issue #914 that accumulation using fp16 causes overflow

### DIFF
--- a/cpp/src/neighbors/detail/ann_utils.cuh
+++ b/cpp/src/neighbors/detail/ann_utils.cuh
@@ -136,7 +136,7 @@ struct config<float> {
 };
 template <>
 struct config<half> {
-  using value_t                    = half;
+  using value_t                    = float;
   static constexpr double kDivisor = 1.0;
 };
 template <>
--- a/cpp/bench/ann/src/cuvs/cuvs_benchmark.cu
+++ b/cpp/bench/ann/src/cuvs/cuvs_benchmark.cu
@@ -75,8 +75,8 @@ auto create_algo(const std::string& algo_name,
   if constexpr (std::is_same_v<T, uint8_t>) {}
 
 #ifdef CUVS_ANN_BENCH_USE_CUVS_IVF_FLAT
-  if constexpr (std::is_same_v<T, float> || std::is_same_v<T, uint8_t> ||
-                std::is_same_v<T, int8_t>) {
+  if constexpr (std::is_same_v<T, float> || std::is_same_v<T, half> ||
+                std::is_same_v<T, uint8_t> || std::is_same_v<T, int8_t>) {
     if (algo_name == "raft_ivf_flat" || algo_name == "cuvs_ivf_flat") {
       typename cuvs::bench::cuvs_ivf_flat<T, int64_t>::build_param param;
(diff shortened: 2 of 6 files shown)


cpp/src/neighbors/detail/ann_utils.cuh
```diff
@@ -148,7 +148,7 @@ struct config<float> {
 };
 template <>
 struct config<half> {
-  using value_t                    = half;
+  using value_t                    = float;
```
achirkin, line 139: Please check whether this is used outside the IVF-Flat. Changing the accumulation type like this can have a drastic impact on performance.

achirkin (changes requested):
Thank you for the contribution. I see the PR changes the accumulation type from fp16 to fp32 to avoid fp overflow. However I'm not sure if this is desirable in general and whether the speed drop is acceptable for the cases when the overflow doesn't happen.
Maybe we'd better just advise the user to switch to fp32 variant of the algorithm?
Please support the PR with the benchmark results (using cuvs ann-bench) before and after the PR for fp16 and fp32 if you decide to proceed with this approach.

cpp/src/neighbors/detail/ann_utils.cuh
hhy3: I did simple benchmark and it showed no significant differences. I'll use cuvs ann-bench to get a more detailed benchmark results later

cjnolet: @hhy3 any updates here? We're about to begin burndown for 25.08 release. Should we consider this for 25.08 or push to 25.10 (October)?

hhy3: @cjnolet hi, push it to 25.10, thx

hhy3: @cjnolet sorry for my late reply. I just benchmark the performance, and it shows some regression on performance, especially when nprobe is small, but significantly improves recall: 

### IVF-Flat FP16 AccT=float vs AccT=half Benchmark Results

- Dataset: cohere-768-angular-fp16 (1M vectors, 768 dims, inner_product)
- GPU: NVIDIA A100-PCIE-40GB
- nlist=4096, ratio=10, niter=20, k=100

#### Latency (bs=64)

| nprobe | half latency (ms) | float latency (ms) | change | half recall | float recall |
|--------|-------------------|---------------------|--------|-------------|--------------|
| 32     | 4.02              | 5.04                | +25%   | 0.815       | 0.849        |
| 64     | 6.22              | 6.89                | +11%   | 0.867       | 0.913        |
| 128    | 9.87              | 10.2                | +3%    | 0.897       | 0.955        |

#### Throughput (bs=1000, threads:1)

| nprobe | half QPS | float QPS | change | half recall | float recall |
|--------|----------|-----------|--------|-------------|--------------|
| 32     | 24.7k   | 22.8k     | -8%    | 0.816       | 0.850        |
| 64     | 13.8k   | 13.2k     | -4%    | 0.868       | 0.913        |
| 128    | 7.87k   | 7.84k     | -0.4%  | 0.898       | 0.956        |

PR #1086, opened 2025-07-04
Expose graph and dataset accessors for CAGRA to C/Python

benfred:
Add the ability to get the graph and dataset from a CAGRA index to the c-api and python apis, as well as being able to reconstruct the cagra index from a graph and dataset.

The eventual goal here is to be more flexible in terms of allowing other serialization formats. Rather than supporting every format inside of cuvs, by exposing the raw data needed to recreate a cagra index - we can let consumers of cuvs decide how they want to serialize an index.


.pre-commit-config.yaml
```diff
@@ -108,7 +108,7 @@ repos:
           - id: verify-copyright
             files: |
               (?x)
-                  [.](cmake|cpp|cu|cuh|h|hpp|sh|pxd|py|pyx|rs)$|
+                  [.](cmake|cpp|cu|cuh|h|hpp|sh|pxd|py|pyx|rs|java)$|
```
mythrocks, line 111: 👍

cpp/include/cuvs/cluster/kmeans.h
```diff
@@ -30,7 +30,7 @@ extern "C" {
  * @{
  */
 
-enum cuvsKMeansInitMethod {
+typedef enum {
```
mythrocks, line 33: 👍

ldematte:
@benfred I tried to check this out and build locally, ~~but it does not compile, claiming template errors around `matrix_vector_op`.
However, I have the same problem on the current "main" `branch-25.08` -- so I suppose it's unrelated to your changes, and something else broke it (I'm 100% sure it was compiling a week ago)~~

Scratch that, I needed to update my conda environment as the raft include files changed.

python/cuvs/cuvs/neighbors/cagra/cagra.pyx
```diff
@@ -734,3 +784,49 @@ def load(filename, resources=None):
     ))
     idx.trained = True
     return idx
+
+
+@auto_sync_resources
+def from_graph(graph, dataset, metric="sqeuclidean", resources=None):
+    """
+    Construct a cagra index from an existing graph nad dataset
```
ldematte: Nit: nad -> and

cpp/include/cuvs/neighbors/cagra.h
```diff
@@ -613,6 +655,21 @@ cuvsError_t cuvsCagraSerializeToHnswlib(cuvsResources_t res,
...
+ * @param[in] metric cuvsDistanceType to use in the index
+ * @params[in] graph the knn graph to use , shape (size, graph_degree)
+ * @params[in] dataset the dataset to use , shape (size, dim)
+ * @param[out] index CAGRA index constructed from the graph and dataset
+ */
+cuvsError_t cuvsCagraIndexFromGraph(cuvsResources_t res,
+                                    cuvsDistanceType metric,
+                                    DLManagedTensor* graph,
+                                    DLManagedTensor* dataset,
+                                    cuvsCagraIndex_t index);
```
ldematte: Shouldn't this be marked as `in/out`? And maybe have a better description?
From the description, it seems the function will create the index. 
From the code and the type, it seems to me you will need to create it first with `cuvsCagraIndexCreate`, and pass it down for this function to "fill" it.

benfred: good point! I've added a better description (and a code example) in the last commit

ldematte: Thanks!
BTW, I have written some preliminary Java bindings for this, and it seems to work well (it does correctly round-trip)
https://github.com/ldematte/cuvs/blob/java/cagra-read-graph-dataset/java/cuvs-java/src/test/java/com/nvidia/cuvs/CagraBuildAndSearchIT.java#L283

Still WIP, we are trying to figure out a good API, but it works.

cpp/src/neighbors/cagra_c.cpp
```diff
@@ -120,6 +120,46 @@ void* _build(cuvsResources_t res, cuvsCagraIndexParams params, DLManagedTensor*
...
+                  DLManagedTensor* graph_tensor,
+                  DLManagedTensor* dataset_tensor)
+{
+  auto metric  = static_cast<cuvs::distance::DistanceType>((int)_metric);
+  auto dataset = dataset_tensor->dl_tensor;
+  auto graph   = graph_tensor->dl_tensor;
+  auto res_ptr = reinterpret_cast<raft::resources*>(res);
+
+  void* index = NULL;
+  if (cuvs::core::is_dlpack_device_compatible(dataset)) {
```
cjnolet, line 135: This looks good. Do we also need to update the accessors for the PQ codebooks to follow this design? (in a separate PR, of course).

cpp/include/cuvs/neighbors/cagra.h
mythrocks: I see now that this isn't a `cuvsCreateCagraIndexFromGraph`, but a `cuvsPopulateCagraIndexFromGraph`.

cpp/src/neighbors/cagra_c.cpp
```diff
@@ -120,6 +120,46 @@ void* _build(cuvsResources_t res, cuvsCagraIndexParams params, DLManagedTensor*
...
+      using graph_mdspan_type = raft::device_matrix_view<uint32_t const, int64_t, raft::row_major>;
+      auto graph_mds          = cuvs::core::from_dlpack<graph_mdspan_type>(graph_tensor);
+      index = new cuvs::neighbors::cagra::index<T, uint32_t>(*res_ptr, metric, mds, graph_mds);
+    } else {
+      using graph_mdspan_type = raft::host_matrix_view<uint32_t const, int64_t, raft::row_major>;
+      auto graph_mds          = cuvs::core::from_dlpack<graph_mdspan_type>(graph_tensor);
+      index = new cuvs::neighbors::cagra::index<T, uint32_t>(*res_ptr, metric, mds, graph_mds);
+    }
+  } else if (cuvs::core::is_dlpack_host_compatible(dataset)) {
+    using mdspan_type = raft::host_matrix_view<T const, int64_t, raft::row_major>;
```
mythrocks, line 148: Given that this code is similar to the `device_compatible` branch, I've been wondering if there was a way to avoid the repetition.  Any sort of double-dispatch I can think of seems as verbose as this, though. :/

cpp/src/neighbors/cagra_c.cpp
```diff
@@ -297,6 +337,68 @@ void* _merge(cuvsResources_t res,
...
+  auto res_ptr   = reinterpret_cast<raft::resources*>(res);
+  auto index_ptr = reinterpret_cast<cuvs::neighbors::cagra::index<T, IdxT>*>(index->addr);
+  auto dst       = cuvs::core::from_dlpack<output_mdspan_type>(dataset);
+  auto src       = index_ptr->dataset();
+
+  RAFT_EXPECTS(src.extent(0) == dst.extent(0), "Output dataset has incorrect number of rows");
+  RAFT_EXPECTS(src.extent(1) == dst.extent(1), "Output dataset has incorrect number of cols");
+
+  // dataset from cagra could be strided, account for that when copying over
+  raft::copy_matrix(dst.data_handle(),
```
mythrocks: Does this mean that the `dataset` is copied unconditionally to `__host__` memory?  Does the recent discussion regarding returning a `__device__` memory reference have bearing on this bit?

cjnolet: @benfred and I spoke offline about this. The current plan is to have the C API populate the given dltensor and just reuse the existing device pointer (from the index).

benfred: I've added this in  https://github.com/rapidsai/cuvs/pull/1086/commits/2f22d5356657dabd810ec8a74d36980b02f286ee - basically returns a DLTensor that references the device memory from the cagra index itself without copying.

cjnolet:

mythrocks:
Asking here as opposed to Slack, so as to keep (part of) the conversation documented.  I'd be happy to discuss offline if preferred. 

@ldematte, @ChrisHegarty:  I was wondering under what scenario the user might need to copy the "dataset" portion of the index back to `__host__` memory? 

Could it be that this index might have been freshly deserialized from disk/stream, and the dataset used to create the index is no longer at hand?

Also, would there be value in a mechanism to copy small portions of the dataset to `__host__` memory for incremental processing, rather than the whole hog?  

Speaking with @cjnolet, one wonders if that incremental processing involves say, updating a(nother) CAGRA index or performing some other GPU operation.  If yes, would that use be better served by keeping the dataset on the GPU instead of copying it off?

Edit: I'm likely overthinking this. Is the focus currently to enable conversion to a CPU-queryable format?  Incremental stuff to be considered later on?

ldematte:
I think ATM we are focusing on the graph portion (but that does not mean we can/will apply the same to the dataset portion too)

> Also, would there be value in a mechanism to copy small portions of the dataset to __host__ memory for incremental processing, rather than the whole hog?

That's what we plan to do on the Java side, and one of the reasons why I structured the API in #1111 to have row access + column access instead of just have a `get(r, c)`, so I can easily implement a "paging" mechanism (and the API users are "encouraged" to perform incremental, row-by-row processing OR copy out a portion of the data).
I can do that completely in Java, but if you think there is value for C users too, and want to add a specific C API, I can use that. My 2c: it's not worth the effort: C users can/will just use cudaMemcpy/cudaMemcpy2D (but I'll let you decide on this point), and I think that from a Java bindings perspective it won't be easier or more difficult (but please challenge me on this point if you think otherwise!)

> updating a(nother) CAGRA index or performing some other GPU operation. If yes, would that use be better served by keeping the dataset on the GPU instead of copying it off?

I think this is a reasonable scenario, and definitely one thing I was considering myself when I started #1111 : if we can pass around an object that just represents "some memory" (e.g. `CuVSDeviceMatrix`), we can keep the dataset on the GPU and perform further operations. Like quantize + index or something similar.

To the point: I think it's good for completeness/orthogonality to have these functions "return" (via an out parameter) either device or host memory like we have now; however, if I have to pick one, I would say "device memory", and then the API user (Java or directly C code) can decide how, when and if they want to copy data back.

ldematte:
TL;DR: I liked the API as it was in https://github.com/rapidsai/cuvs/commit/2f22d5356657dabd810ec8a74d36980b02f286ee (`cuvsCagraIndexGetGraphView` + ` cuvsCagraIndexCopyGraph`), but I like the last implementation ("view" + `cuvsCopyMatrix`) even more. 

If anything, in a follow-up PR, it would be great to have partial copy (e.g. `cuvsCopyMatrix` with `fromRow`, `toRow` or something similar) so in the Java side I don't have to replicate what is done there and/or in `raft::copy_matrix<T>`.

benfred:
> If anything, in a follow-up PR, it would be great to have partial copy (e.g. cuvsCopyMatrix with fromRow, toRow or something similar) so in the Java side I don't have to replicate what is done there and/or in raft::copy_matrix<T>.

How about having something like a `cuvsSliceMatrix` or `cuvsSliceRows` functions to do this ?   (where the cuvsSliceRows function doesn't copy data, just adjusts the shape / strides / data pointer to slice the matrix without copying data). We could then pass the slice to the copy matrix function , like:

```c++
// get the cagra graph
DLManagedTensor graph;
cuvsCagraIndexGetGraph(index, &graph);

// get the first 1K rows from the graph
DLManagedTensor subgraph;
cuvsSliceMatrix(&graph, 0, 1024,  &subgraph);

// copy the subgraph to host  memory
cuvsCopyMatrix(&subgraph,  &subgraphHost);
```

(this actually makes me think that I should rename cuvsCopyMatrix to cuvsMatrixCopy , so that we can have cuvsMatrixSlice etc - and to be consistent with other cuvs API's)

mythrocks:
> if you think there is value for C users too, and want to add a specific C API, I can use that. My 2c: it's not worth the effort: C users can/will just use `cudaMemcpy`/`cudaMemcpy2D` (but I'll let you decide on this point)...

@benfred / @cjnolet will keep me honest here. 

I think the C-lang user should not use `cudaMemcpy` directly.  The dataset is likely to be padded to an 8-byte boundary when stored in `__device__` memory.  A naive `cudaMemcpy` will copy/interpret padding bytes that aren't actual data.

Using a `cuVS`-specific `copy()` API would be preferable to insulate the user from padding.  If a future CAGRA implementation does away with the padding, the `cuVSMatrixCopy()` user would be insulated from the change.  The `cudaMemcpy()` user might not.

ldematte:
> How about having something like a cuvsSliceMatrix or cuvsSliceRows functions to do this

That's even better!

> this actually makes me think that I should rename cuvsCopyMatrix to cuvsMatrixCopy

++

cpp/src/neighbors/cagra_c.cpp
benfred: yeah - I'm open to suggestions here, but can't think of a better way to do this either =(

cpp/include/cuvs/core/detail/interop.hpp
```diff
@@ -143,9 +143,10 @@ inline bool is_c_contiguous(DLManagedTensor* managed_tensor)
 
 #pragma GCC diagnostic push
 #pragma GCC diagnostic ignored "-Wunused-function"
-static void free_dlmanaged_tensor_shape(DLManagedTensor* tensor)
+static void free_dlmanaged_tensor_metadata(DLManagedTensor* tensor)
 {
   delete[] tensor->dl_tensor.shape;
+  if (tensor->dl_tensor.strides != NULL) { delete[] tensor->dl_tensor.strides; }
```
mythrocks: [Nit]:  FWIW, the `nullptr` check might be superfluous.  It should be safe to `delete []` the null pointer.

```suggestion
  delete[] tensor->dl_tensor.strides;
```

cpp/include/cuvs/core/detail/interop.hpp
```diff
@@ -157,14 +158,21 @@ static void to_dlpack(MdspanType src, DLManagedTensor* dst)
   tensor->dtype  = data_type_to_DLDataType<typename MdspanType::value_type>();
   tensor->device = accessor_type_to_DLDevice<typename MdspanType::accessor_type>();
   tensor->ndim   = MdspanType::extents_type::rank();
-  tensor->data   = src.data_handle();
-
-  tensor->shape = new int64_t[tensor->ndim];
-  dst->deleter  = free_dlmanaged_tensor_shape;
-
+  tensor->data   = const_cast<typename MdspanType::value_type*>(src.data_handle());
```
mythrocks, line 161: [Nit]:  This `const_cast` is unfortunate.  We shouldn't have to grant write-access to the source side of the copy.

Not critical for this PR, but I'd like to wrestle with this a little, in a follow-up.

cpp/include/cuvs/neighbors/cagra.h
```diff
@@ -358,6 +358,66 @@ cuvsError_t cuvsCagraIndexDestroy(cuvsCagraIndex_t index);
...
+ * This function returns a non-owning view of the CAGRA graph.
+ * The output will be referencing device memory that is directly used
+ * in CAGRA, without copying the graph at all. This means that the
+ * output is only valid as long as the CAGRA index is alive, and once
+ * cuvsCagraIndexDestroy is called on the cagra index - the returned
+ * graph view will be invalid.
+ *
+ * Note that the DLManagedTensor graph returned will have an associated
+ * 'deleter' function that must be called when the graph is no longer
+ * needed. This will free up host memory that stores the shape of the
```
mythrocks:
```suggestion
 * needed. This will free up host memory that stores the metadata for the
```

cpp/include/cuvs/core/detail/interop.hpp
mythrocks: I'm not sure this is really a problem.  We do put the `const` back when we call `raft::copy_matrix<>()`.  It would be good if we could iron out that wrinkle, later on.

cpp/include/cuvs/core/detail/interop.hpp
```diff
@@ -157,14 +158,21 @@ static void to_dlpack(MdspanType src, DLManagedTensor* dst)
...
+  tensor->shape  = new int64_t[tensor->ndim];
   for (int64_t i = 0; i < tensor->ndim; ++i) {
     tensor->shape[i] = src.extent(i);
   }
-}
 
+  if constexpr (std::is_same_v<typename MdspanType::layout_type, raft::row_major>) {
+    tensor->strides = NULL;
```
mythrocks:
```suggestion
    tensor->strides = nullptr;
```

cpp/src/core/c_api.cpp
```diff
@@ -165,3 +166,117 @@ extern "C" cuvsError_t cuvsVersionGet(uint16_t* major, uint16_t* minor, uint16_t
...
+  DLTensor& src = src_managed->dl_tensor;
+  DLTensor& dst = dst_managed->dl_tensor;
+
+  int64_t src_row_stride = src.strides == nullptr ? src.shape[1] : src.strides[0];
+  int64_t dst_row_stride = dst.strides == nullptr ? dst.shape[1] : dst.strides[0];
+  auto res_ptr           = reinterpret_cast<raft::resources*>(res);
+
+  raft::copy_matrix<T>(static_cast<T*>(dst.data),
+                       dst_row_stride,
+                       static_cast<const T*>(src.data),
```
mythrocks, line 183: [Nit]:  Is this cast required?  Shouldn't there be a trivial conversion from `T*` to `const T*`?

mythrocks: Oh, wait.  Nix that.  `src.data` is a `void*` at this point.

cpp/include/cuvs/core/detail/interop.hpp
benfred: thats a good point - 

for future work, I'd like to also move to the latest version of the dlpack spec, and use the `DLManagedTensorVersioned` object instead. One of the things thats been added is a `flags` object that we can indicate whether or not the tensor is readonly (https://dmlc.github.io/dlpack/latest/c_api.html#c.DLManagedTensorVersioned.flags) . 

However, thats a much bigger change that would impact the rest of the cuvs c-api - and I think we should leave this as is for the time being

python/cuvs/cuvs/common/device_tensor_view.pyx
```diff
@@ -0,0 +1,145 @@
...
+from cuvs.common.cydlpack cimport DLManagedTensor, dlpack_c
+
+from pylibraft.common.cai_wrapper import wrap_array
+
+from cuvs.common.cydlpack import dl_data_type_to_numpy
+from cuvs.common.exceptions import check_cuvs
+from cuvs.common.resources import auto_sync_resources
+
+
+cdef class DeviceTensorView:
```
lowener, line 34: Can't we return `device_matrix_view` without intermediate copies? I am not 100% sure I see the benefits of adding this class compared to device views?

benfred: The python API is using the c-api (instead of the c++ api) - meaning we can't use `device_matrix_view` directly inside python. Instead we're using `dlpack.DLManagedTensor` inside our C-API.

Previous to this PR, our c-api only accepted dlpack arrays as inputs (and would either use the contents in functions like `cagra::build` - or fill pre-allocated arrays with the outputs in functions like `cagra::search`).  We hadn't yet exposed memory that was allocated in our C++ codebase to python. 

This PR changes that - and we are now returning dlpack DLManagedTensor objects that return memory that is owned by and allocated inside our C++ codebase.   This code does that without copying the data , with the flow going : device_matrix_view (c++) -> DLManagedTensor (C) -> DeviceTensorView (python) . At each step we aren't copying the data, there isn't an intermediate copy - so much as intermediate objects that have a pointer to the original data, and also extra information such as the shape/dtype/strides etc.  

This DeviceTensorView code is necessary because we didn't have anything that would take a DLManagedTensor and return something that could be easily consumed in python. the closest object we had was the `pylibraft.device_ndarray` object - but that object wouldn't have worked here.

cpp/include/cuvs/neighbors/cagra.h
```diff
@@ -609,10 +670,51 @@ cuvsError_t cuvsCagraSerializeToHnswlib(cuvsResources_t res,
...
+ * DLManagedTensor graph;
+ *
+ * cuvsDistanceType metric = L2Expanded;
+ *
+ * // Build the CAGRA Index from the graph/dataset
+ * cuvsError_t status = cuvsCagraIndexFromGraph(res, metric, &graph, &dataset, index);
+ *
+ * @endcode
+ */
+cuvsError_t cuvsCagraIndexFromGraph(cuvsResources_t res,
```
cjnolet: Nitpick- can we please rename this to `cuvsCagraIndexFromParams` or `cuvsCagraIndexFromArgs`? I'd like to keep tthe API design consistent and having to specify specific args in the name will get unwieldy quickly.

benfred: I renamed to `cuvsCagraIndexFromArgs` in the last commit -  (went with FromArgs instead of FromParams - since I think the FromParams could be confused with the Index Params we use to build the index)

bdice (approved):
Approving packaging changes.

PR #1089, opened 2025-07-08
[Java] Introduce scoped resource access

ldematte:
This PR wraps access to the cuVS resources handle into a autocloseable accessor. The reason is to provide a clear scope for resource access, with the possibility to add application logic to the scope begin/end.

An example of this logic is the `CheckedCuVSResources` decorator introduced in the test suite, which checks that the `CuVSResources` it decorates has not been destroyed yet, and that access to the underlying cuVS resources handle is not concurrent (is not done by more than 1 thread at every time).

I preferred to write it as a decorator, as this gives the API consumers more freedom: if they want the check, they can use `CheckedCuVSResources` (or a similar class); if they want direct, straight, fast access, they can just use the accessor; if they need more logic (e.g. logging or anything else), that is possible too.

Tests are updated to use  `CheckedCuVSResources`, so we can spot issues quickly in tests without burdening production code with additional logic.


ChrisHegarty:
I do very much like this approach and API. It closely aligns with the intent of the C API, while allowing to enforce higher-level constraints through typical java patterns, e.g. the decorator pattern. The consume is free to decide if, and how to, enforce thread semantics.

ldematte:
@mythrocks I've rebased this PR to include the latest merges (TY!)
I think this should be the next one to go in, as it is touching several files - changes should be minor, but somehow git merge does not like them (too much whitespace change).

cjnolet:

mythrocks:
Reviewing now.  This interface looks very elegant, on first glance.  I haven't gotten to the thread-safety checks yet.

mythrocks:
> I haven't gotten to the thread-safety checks yet.
 
I think I might have misunderstood the intent behind this PR.  The title (`Thread check on resource access`) led me to understand that this PR addresses concerns regarding the multi-threaded access of `cuvsResources_t`.

Unless I've missed something crucial, this PR introduces _infrastructure_ for such checks to be added in the future?  A `ScopedAccess` interface could potentially mark off the critical section where the `CuVSResource` is accessed through a single thread.

As interfaces go, this solution is quite elegant.  We should be able add thread-sync support nicely.

But I think we should change the PR title to indicate that this provides a scoped-accessor for the resources, and mention thread-safety as a possible add-on in the future.

Please correct me if I've missed the thread-safety angle in the PR.

mythrocks:
(#1028 has been merged, so it's borked this change. :/ A similar change will be needed in `TieredIndexImpl`.) I can try get to this tomorrow.

I'm largely 👍 on this change, barring the pending fix for `TieredIndexImpl`.

mythrocks:
I've tried to get a handle (aha!) on the thread access constraints for `cuvsResource_t` from @benfred and @cjnolet.  This PR looks a reasonable place to document some thoughts around it.  We can carry this forward to the future `ThreadSyncScopedAccess` PR.

Ideally, the `cuvsResource_t` would be created and used in the same thread.  In the `cuvs-java` APIs, one sees that the `cuvsResource_t` native handle is created in the constructor of `CuVSResources` class.  This is then used in the index `build()` and the `search()` operations.

### Creation thread vs build/search thread
One contrived, potential misuse pattern could be that a thread creates a `CuVSResources`, sticks it in a `Runnable` that's added to a job-queue, serviced by a threadpool.  The thread that creates the `CuVSResources` might be different from the pool-thread that runs `build()`/`search()`.  Per @benfred, this should be ok, so long as only one thread uses the resource at any given time.

### Sync on `CuVSResources` or the handle?
Would it be enough to `synchronize()` on the `CuVSResources` object?  I'm thinking yes.  I don't see how two different `CuVSResources` objects could have the same handle, via copies.  So finding a way to sync on the handle seems unnecessary.

### Per-thread default stream (PTDS)
Operations like `build()`/`search()` should `cudaStreamSync` on the same CUDA stream, to ensure results copied to/from `__host__`/`__device__` are ready for consumption.
But with PTDS-enabled, each CPU thread would have a separate default CUDA stream associated with it.  Could the use of a threadpool (with different default streams) affect things?  I don't think it does, because we'd ensure that the resources are accessed serially from threads.

ldematte:
> Unless I've missed something crucial, this PR introduces infrastructure for such checks to be added in the future? 

Yes, this PR introduces infrastructure to do that, _and_ it introduces an implementation that does such checks in tests (see `CheckedCuVSResources`). It is currently not my intention to add these check in production  code: I do not want to force the burden of access check or of sync upon users which code respects the threading constraints by design.
This PR does allow devs that need that for whatever reason (e.g. introduce CuVS inside legacy/complicated code) to add them if they need or want to.

ldematte:
> But I think we should change the PR title to indicate that this provides a scoped-accessor for the resources, and mention thread-safety as a possible add-on in the future.

Agreed; I changed the PR title, and reviewed and slightly adjust the description.

> I've tried to get a handle (aha!) on the thread access constraints for cuvsResource_t

Ah! Nice pun! :D 

>  We can carry this forward to the future ThreadSyncScopedAccess PR.

Again, I don't think we need it; whoever needs this can write its own decorator. 
BUT I'm OK to add it as a optional utility in the production code and/or move `CheckedCuVSResources` to production code if we feel this might be a recurrent use case and many users can benefit from it.

@mythrocks WDYT? (In any case, I would leave that for a follow-up PR)

> Per @benfred, this should be ok, so long as only one thread uses the resource at any given time.

We are actually testing this scenario in our IT tests, and it is OK, so long as only one thread uses the resource at any given time (this is what's enforced by `CheckedCuVSResources` btw).

> Would it be enough to synchronize() on the CuVSResources object? I'm thinking yes. I don't see how two different CuVSResources objects could have the same handle, via copies. So finding a way to sync on the handle seems unnecessary.

Yes, I think this would be how a hypothetical `SyncCuVSResources` (which would return a `ThreadSyncScopedAccess` or similar) would be implemented (sync on the `CuVSResources` it decorates).

ldematte:
> (https://github.com/rapidsai/cuvs/pull/1028 has been merged, so it's borked this change. :/ A similar change will be needed in TieredIndexImpl.) I can try get to this tomorrow.

I don't see merge conflicts but I bet it fails to compile. Let me fix that.

ldematte:
@mythrocks I merged https://github.com/rapidsai/cuvs/pull/1028 and fixed/cleaned up a few things besides adding resource access:
- used `Dataset.ofArray` to avoid passing around float[][] and reduce duplication and complexity
- used named constants where possible to increase readability (e.g. cudaMemcpyHostToDevice)
- used `allocateRMMSegment` to reduce duplication
- fixed dataset memory leak

Some of the changes (e.g. `Dataset.ofArray` and avoid passing float[][] down) were necessary for this to change (so I could use CuVSResources without the need to cast up to a specific implementation). Others could go to another cleanup PR, but I wanted to get this right ASAP.

ldematte:
Still missing: the new code is allocating params directly (e.g. `cuvsTieredIndexParams.allocate(arena);`). It should use `cuvsTieredIndexParamsCreate` + Destroy like I did in https://github.com/rapidsai/cuvs/pull/1109 and https://github.com/rapidsai/cuvs/pull/1110. It's OK, I can fix it later, but for the future can we point this in the review of the PRs, so that old code/old patterns does not keep sneaking in?

punAhuja:
> Still missing: the new code is allocating params directly (e.g. `cuvsTieredIndexParams.allocate(arena);`). It should use `cuvsTieredIndexParamsCreate` + Destroy like I did in #1109 and #1110. It's OK, I can fix it later, but for the future can we point this in the review of the PRs, so that old code/old patterns does not keep sneaking in?

Will raise a PR and change this. @ldematte

ldematte:
> Will raise a PR and change this. @ldematte

That would be awesome @punAhuja, thanks! To keep things consistent, would you mind using `CloseableHandle` from my PRs, so we keep a single pattern to do that?

punAhuja:
> > Will raise a PR and change this. @ldematte
> 
> That would be awesome @punAhuja, thanks! To keep things consistent, would you mind using `CloseableHandle` from my PRs, so we keep a single pattern to do that?

Yes, have looked at your PRs, will follow a similar pattern for consistency. @ldematte

mythrocks:
> It's OK, I can fix it later, but for the future can we point this in the review of the PRs, so that old code/old patterns does not keep sneaking in?

Agreed. I had held up #1028 long enough;  I was hoping that could be handled in a follow-up.  Thank you for pointing it out, @ldematte.  Also, thank you for picking that up, @punAhuja.

mythrocks:
P.S. Apologies for not having caught some of this in the review of #1028. 

Thanks for making the fixes in `TieredIndexImpl`, @ldematte.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/TieredIndexImpl.java
```diff
@@ -129,6 +102,10 @@ public void destroyIndex() throws Throwable {
     try {
       int returnValue = cuvsTieredIndexDestroy(tieredIndexReference.getMemorySegment());
       checkCuVSError(returnValue, "cuvsTieredIndexDestroy");
+      if (dataset != null) {
+        dataset.close();
```
mythrocks, line 106: Ah, this is resource leak you mentioned.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/TieredIndexImpl.java
```diff
@@ -142,59 +119,54 @@ public void destroyIndex() throws Throwable {
...
-      MemorySegment datasetDP = datasetD.get(C_POINTER, 0);
+        // Copy host to device
+        Util.cudaMemcpy(datasetDP, hostDataSeg, datasetSize, HOST_TO_DEVICE);
 
-      // Copy host to device
-      Util.cudaMemcpy(datasetDP, hostDataSeg, datasetSize, HOST_TO_DEVICE);
+        // Create tensor from device memory
+        long[] datasetShape = {rows, cols};
+        MemorySegment datasetTensor =
+            prepareTensor(localArena, datasetDP, datasetShape, kDLFloat(), 32, 2, kDLCUDA(), 1);
```
mythrocks, line 148: Ok, there's the magic numbers you mentioned.  Yep, your way is better.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/TieredIndexImpl.java
```diff
@@ -142,59 +119,54 @@ public void destroyIndex() throws Throwable {
...
+        long cuvsRes = resourceAccess.handle();
 
-      // TieredIndex REQUIRES device memory - allocate it
-      MemorySegment datasetD = localArena.allocate(C_POINTER);
-      long datasetSize = C_FLOAT_BYTE_SIZE * rows * cols;
-      int returnValue = cuvsRMMAlloc(cuvsRes, datasetD, datasetSize);
-      checkCuVSError(returnValue, "cuvsRMMAlloc");
+        // TieredIndex REQUIRES device memory - allocate it
+        long datasetSize = C_FLOAT_BYTE_SIZE * rows * cols;
+        MemorySegment datasetDP = allocateRMMSegment(cuvsRes, datasetSize);
```
mythrocks, line 140: Nitpick: Not to be handled in the current PR.

Do we stand the risk of leaking the allocation from RMM if there is an exception before the cleanup?  

Might need to look at this more closely, separately.

PR #1093, opened 2025-07-08
Processing mutual reachability dist in connect_knn_graph for host data

jinsolp:
Closes https://github.com/rapidsai/cuvs/issues/1096

Existing `connect_knn_graph` function for host data does not consider the reduction op.

This PR adds a fix to calculate mutual reachability distances for selected random cross-component neighbors instead of simple pairwise distances when `reduction_op == cuvs::sparse::neighbors::MutualReachabilityFixConnectivitiesRedOp`.


cpp/src/cluster/detail/mst.cuh
```diff
@@ -190,34 +192,50 @@ void connect_knn_graph(
...
+                             cuvs::sparse::neighbors::MutualReachabilityFixConnectivitiesRedOp<
+                               value_idx,
+                               value_t>>::value) {
+    thrust::device_ptr<value_t> pairwise_ptr(pairwise_dist_vec.data_handle());
+    thrust::device_ptr<value_t> core_dist_ptr(reduction_op.core_dists);
+    thrust::device_ptr<value_idx> device_u_ptr(device_u_indices.data_handle());
+    thrust::device_ptr<value_idx> device_v_ptr(device_v_indices.data_handle());
+    thrust::counting_iterator<value_idx> idx(0);
+
+    thrust::transform(idx, idx + pairwise_dist_vec.size(), pairwise_ptr, [=] __device__(int i) {
```
tarang-jain: Can we use `raft::linalg::map_offset` instead of thrust?

cpp/src/cluster/detail/mst.cuh
```diff
@@ -190,34 +192,50 @@ void connect_knn_graph(
...
                                     pairwise_dist.view(),
                                     metric);
 
-  auto pairwise_dist_vec = raft::make_device_vector<value_t, int64_t>(handle, n_components - 1);
+  auto pairwise_dist_vec = raft::make_device_vector<value_t, value_idx>(handle, n_components - 1);
   raft::matrix::get_diagonal(
     handle, raft::make_const_mdspan(pairwise_dist.view()), pairwise_dist_vec.view());
 
-  size_t new_nnz = n_components - 1;
+  if constexpr (std::is_same<red_op,
```
tarang-jain, line 231: Rather than passing the reduction op struct and then checking the type, can we use a boolean or an enum instead in the function argument to indicate if we want MR linkage or pairwise? While currently, there are only two types of reduction ops, this might give incorrect results if in the future some other reduction op is used. In the docs of this function, also can you add a line saying that only pairwise or mutual reachability connectivities op are supported?

tarang-jain: Actually it looks like the reduction op is already created before calling connect_knn_graph. In that case, doing a boolean check and again creating a new reduction op based on the type chosen is inefficient. So it makes sense to pass the reduction op here, which was already created earlier. What you could do however is add a check if it is not one of FixConnectivitesRedOp and FixConnectivitiesMutualReachabilityRedOp, so that it throws an error in that case.

cpp/src/cluster/detail/mst.cuh
```diff
@@ -132,22 +135,30 @@ void connect_knn_graph(
...
   size_t m,
   size_t n,
-  int n_components,
   value_idx* color,
+  red_op reduction_op                 = red_op{},
   cuvs::distance::DistanceType metric = cuvs::distance::DistanceType::L2SqrtExpanded)
 {
-  auto stream = raft::resource::get_cuda_stream(handle);
+  using namespace cuvs::sparse::neighbors;
+  static_assert(
```
tarang-jain, line 153: I think a compile-time check here is awesome! This PR LGTM!

PR #1094, opened 2025-07-09
Fix single GPU sharded search merge

viclafargue:
https://github.com/rapidsai/cuvs/issues/904


cjnolet (approved):

cpp/src/neighbors/mg/snmg.cuh
```diff
@@ -210,6 +210,29 @@ void sharded_search_with_direct_merge(
     auto query_partition            = raft::make_host_matrix_view<const T, int64_t, row_major>(
       queries.data_handle() + query_offset, n_rows_of_current_batch, n_cols);
 
+    if (index.num_ranks_ == 1) {
+      const raft::resources& dev_res = raft::resource::set_current_device_to_root_rank(clique);
```
cjnolet: It looks like we have the same logic here copied twice. Can we please pull this out into a function that can be reused? Centralized this will help us maintain it better in the future.

cjnolet (changes requested):

cpp/src/neighbors/mg/snmg.cuh
viclafargue: I realized I could use the already existing `run_search_batch` function. Updated the PR with it.

cjnolet (approved):

cjnolet (approved):

PR #1095, opened 2025-07-09
Reduce binary size of refine functions

tfeher:
The refine functions that work with GPU data use IVF-Flat under the hood to perform the refinement operation. This PR adds extern template declarations for `ivfflat_interleaved_scan` and uses these in the refine functions. This way we avoid recompiling the IVF-Flat search kernels, and save binary size.

### DIFF
--- a/cpp/src/neighbors/ivf_flat/ivf_flat_interleaved_scan_ext.cuh
+++ b/cpp/src/neighbors/ivf_flat/ivf_flat_interleaved_scan_ext.cuh
@@ -1,0 +1,81 @@
+/*
+ * Copyright (c) 2025, NVIDIA CORPORATION.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include <cuvs/neighbors/common.hpp>
+#include <raft/core/resource/cuda_stream.hpp>
+#include <raft/core/resources.hpp>
+
+namespace cuvs::neighbors::ivf_flat::detail {
+template <typename T, typename AccT, typename IdxT, typename IvfSampleFilterT>
+void ivfflat_interleaved_scan(const index<T, IdxT>& index,
+                              const T* queries,
+                              const uint32_t* coarse_query_results,
+                              const uint32_t n_queries,
+                              const uint32_t queries_offset,
+                              const cuvs::distance::DistanceType metric,
+                              const uint32_t n_probes,
+                              const uint32_t k,
+                              const uint32_t max_samples,
+                              const uint32_t* chunk_indices,
+                              const bool select_min,
+                              IvfSampleFilterT sample_filter,
+                              uint32_t* neighbors,
+                              float* distances,
+                              uint32_t& grid_dim_x,
+                              rmm::cuda_stream_view stream);
+
+#define CUVS_INST_IVF_FLAT_INTERLEAVED_SCAN(T, IdxT, AccT, SampleFilterT)      \
+  extern template void ivfflat_interleaved_scan<T, AccT, IdxT, SampleFilterT>( \
+    const index<T, IdxT>& index,                                               \
+    const T* queries,                                                          \
+    const uint32_t* coarse_query_results,                                      \
+    const uint32_t n_queries,                                                  \
+    const uint32_t queries_offset,                                             \
+    const cuvs::distance::DistanceType metric,                                 \
+    const uint32_t n_probes,                                                   \
+    const uint32_t k,                                                          \
+    const uint32_t max_samples,                                                \
+    const uint32_t* chunk_indices,                                             \
+    const bool select_min,                                                     \
+    SampleFilterT sample_filter,                                               \
+    uint32_t* neighbors,                                                       \
+    float* distances,                                                          \
+    uint32_t& grid_dim_x,                                                      \
+    rmm::cuda_stream_view stream);
+
+CUVS_INST_IVF_FLAT_INTERLEAVED_SCAN(float,
+                                    int64_t,
+                                    float,
+                                    cuvs::neighbors::filtering::none_sample_filter);
+CUVS_INST_IVF_FLAT_INTERLEAVED_SCAN(half,
+                                    int64_t,
+                                    float,
+                                    cuvs::neighbors::filtering::none_sample_filter);
+
+CUVS_INST_IVF_FLAT_INTERLEAVED_SCAN(int8_t,
+                                    int64_t,
+                                    float,
+                                    cuvs::neighbors::filtering::none_sample_filter);
+CUVS_INST_IVF_FLAT_INTERLEAVED_SCAN(uint8_t,
+                                    int64_t,
+                                    float,
+                                    cuvs::neighbors::filtering::none_sample_filter);
+
+#undef CUVS_INST_IVF_FLAT_INTERLEAVED_SCAN
+
+}  // namespace cuvs::neighbors::ivf_flat::detail
--- a/cpp/src/neighbors/refine/refine_device.cuh
+++ b/cpp/src/neighbors/refine/refine_device.cuh
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2023-2024, NVIDIA CORPORATION.
+ * Copyright (c) 2023-2025, NVIDIA CORPORATION.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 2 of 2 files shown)


tfeher:
The `ivfflat_interleaved_scan` function is expected to be the largest contributor in binary size for the `refine_device`. We still have a few other kernel calls in `refine_device`, In a separate PR I will check if we can get rid of those.

tfeher:
The binary size is nicely reduced, but test fail due to undefined symbols. ~~It worked locally,~~ I will look into it.

| filname | compile time | binary size |
| -- | -- | -- |
| refine_device_half_float.cu.o | 112.149 s | 2.185 MB |
| refine_device_float_float.cu.o | 111.532 s | 2.263 MB |
| refine_device_uint8_t_float.cu.o | 111.155 s | 2.183 MB |
| refine_device_int8_t_float.cu.o | 110.015 s | 2.183 MB |

tfeher:
The error is related to the filter type used for instantiating the search kernels. I am looking into the details.

cpp/src/neighbors/ivf_flat/ivf_flat_interleaved_scan.cuh
```diff
@@ -40,11 +40,6 @@ using namespace cuvs::spatial::knn::detail;  // NOLINT
 
 constexpr int kThreadsPerBlock = 128;
 
-auto RAFT_WEAK_FUNCTION is_local_topk_feasible(uint32_t k) -> bool
-{
-  return k <= raft::matrix::detail::select::warpsort::kMaxCapacity;
-}
-
```
tfeher, line 47: This is moved to `ivf_flat_search.cuh`

cpp/src/neighbors/ivf_flat/ivf_flat_interleaved_scan_ext.cuh
```diff
@@ -0,0 +1,96 @@
...
+                              const uint32_t max_samples,
+                              const uint32_t* chunk_indices,
+                              const bool select_min,
+                              IvfSampleFilterT sample_filter,
+                              uint32_t* neighbors,
+                              float* distances,
+                              uint32_t& grid_dim_x,
+                              rmm::cuda_stream_view stream) RAFT_EXPLICIT;
+
+#define CUVS_INST_IVF_FLAT_INTERLEAVED_SCAN(T, IdxT, SampleFilterT)                        \
```
achirkin, line 47: As per our offline discussion, could you please check if it makes sense to unify and reuse this macro in all four source files below? (it looks like the only difference is the `extern` keyword, which can be prepended outside of the macro)

cpp/src/neighbors/ivf_common.cuh
```diff
@@ -148,11 +148,11 @@ __device__ inline auto find_chunk_ix(uint32_t& sample_ix,  // NOLINT
   return ix_min;
 }
 
-template <int BlockDim, typename IdxT>
+template <int BlockDim, typename IdxT, typename DbIdxT>
 __launch_bounds__(BlockDim) RAFT_KERNEL
   postprocess_neighbors_kernel(IdxT* neighbors_out,                // [n_queries, topk]
                                const uint32_t* neighbors_in,       // [n_queries, topk]
-                               const IdxT* const* db_indices,      // [n_clusters][..]
+                               const DbIdxT* const* db_indices,    // [n_clusters][..]
```
achirkin, line 155: Have a look at the assignment expression below outside of the PR changeset (lines 173-174):
```C++
neighbors_out[k] =
    valid ? db_indices[clusters_to_probe[chunk_ix]][data_ix] : kOutOfBoundsRecord<IdxT>;
```
`kOutOfBoundsRecord<IdxT>` and `db_indices` here may have different types and I'm not 100% sure how are they unified in the ternary expression. We probably should't change the type of the `kOutOfBoundsRecord`, so I'd suggest to wrap `db_indices` into a `static_cast<IdxT>(...)` to avoid any ambiguity.

cpp/src/neighbors/ivf_flat/ivf_flat_build.cuh
```diff
@@ -135,7 +136,10 @@ RAFT_KERNEL build_index_kernel(const LabelT* labels,
   auto source_ix = source_ixs == nullptr ? i + batch_offset : source_ixs[i];
   // In the context of refinement, some indices may be invalid (the generating NN algorithm does
   // not return enough valid items). Do not add the item to the index in this case.
-  if (source_ix == ivf::kInvalidRecord<IdxT> || source_ix == raft::upper_bound<IdxT>()) { return; }
+  if (source_ix == ivf::kInvalidRecord<SourceIdxT> ||
+      source_ix == raft::upper_bound<SourceIdxT>()) {
+    return;
+  }
```
achirkin, line 142: I'm a little bit confused whether we should `SourceIdxT` or `IdxT` in here when the types are different (so the `kInvalidRecord` is also different for the two types).
A quick search through the code tells that we only set `kInvalidRecord` in one place: [when we initialize empty IVF lists](https://github.com/rapidsai/cuvs/blob/1155a3a427cd1d1bfaf8fe74a937ed6dfa797ec7/cpp/src/neighbors/ivf_list.cuh#L70-L73). Could you confirm that we indeed initialize the lists in `SourceIdxT` and not in `IdxT`?

tfeher: We had a previous ANN search that gave us the candidates in `source_ixs` in `SourceIdxT` type (i.e we used an `ivf_pq<SourceIdxT>` index). I expect that that the invalid values are marked as `upper_bound<SourceIdxT>()`.

tfeher:
object file | size after [MB] | size before [MB]
-- | -- | --
ivf_flat/ivf_flat_interleaved_scan_uint8_t_int64_t.cu.o | 49.2 |  
ivf_flat/ivf_flat_interleaved_scan_int8_t_int64_t.cu.o | 48.7 |  
ivf_flat/ivf_flat_interleaved_scan_half_int64_t.cu.o | 45.1 |  
ivf_flat/ivf_flat_interleaved_scan_float_int64_t.cu.o | 43.6 |  
refine/detail/refine_device_uint8_t_float.cu.o | 2.2 | 25.5
refine/detail/refine_device_half_float.cu.o | 2.2 | 23.4
refine/detail/refine_device_float_float.cu.o | 2.2 | 43.4
refine/detail/refine_device_int8_t_float.cu.o | 2.2 | 25.1
ivf_flat/ivf_flat_search_half_int64_t.cu.o | 1.2 | 45.9
ivf_flat/ivf_flat_search_int8_t_int64_t.cu.o | 1.2 | 49.5
ivf_flat/ivf_flat_search_float_int64_t.cu.o | 1.2 | 44.4
ivf_flat/ivf_flat_search_uint8_t_int64_t.cu.o | 1.2 | 50.0
total |  200.0 | 307.2

cpp/src/neighbors/ivf_flat/ivf_flat_interleaved_scan_ext.cuh
tfeher: Yes, it makes sense to put the macro in a separate header and reuse it. I have modified the code so that the explicit instantiations reuse the same macro from `ivf_flat_interleaved_scan_explicit_inst.cuh`. I am not yet using the same macro for the `extern` template definitions, I will return to this in a follow up PR (that will unify instantiations for different filtering methods).

cpp/src/neighbors/ivf_common.cuh
tfeher: That is a good idea to make this explicit! I have also added a static check to confirm that it is not a narrowing conversion.

tfeher (commented):
Thank you Artem for the review, I have addressed the issues.

cpp/src/neighbors/ivf_flat/ivf_flat_build.cuh
```diff
@@ -118,10 +119,10 @@ auto clone(const raft::resources& res, const index<T, IdxT>& source) -> index<T,
  * @param veclen size of vectorized loads/stores; must satisfy `dim % veclen == 0`.
  *
  */
-template <typename T, typename IdxT, typename LabelT, bool gather_src = false>
+template <typename T, typename IdxT, typename LabelT, bool gather_src = false, typename SourceIdxT>
```
divyegala: As `SourceIdxT` is meant to be inferred, we should put it before template arguments with default params.

```suggestion
template <typename T, typename IdxT, typename LabelT, typename SourceIdxT, bool gather_src = false>
```

cpp/src/neighbors/ivf_flat/ivf_flat_interleaved_scan_float_int64_t.cu
```diff
@@ -0,0 +1,25 @@
...
+ */
+
+#include "ivf_flat_interleaved_scan_explicit_inst.cuh"
+
+namespace cuvs::neighbors::ivf_flat::detail {
+
+CUVS_INST_IVF_FLAT_INTERLEAVED_SCAN(float, int64_t, filtering::none_sample_filter);
+CUVS_INST_IVF_FLAT_INTERLEAVED_SCAN(float,
+                                    int64_t,
+                                    filtering::bitset_filter<uint32_t COMMA int64_t>);
```
divyegala: Can we split these out in different TUs? It will speed up the compilation as well as lay the groundwork to compile the filter type instantiations to LTO-IR instead of SASS.

cpp/src/neighbors/ivf_flat/ivf_flat_build.cuh
tfeher: Fixed.

cpp/src/neighbors/ivf_flat/ivf_flat_interleaved_scan_float_int64_t.cu
tfeher: Done.

tfeher (commented):
Thanks Divye for the review, I have addressed the issues.

PR #1099, opened 2025-07-09
[FEA] Binary IVF Flat Index

tarang-jain:
Depends on https://github.com/rapidsai/raft/pull/2770

Implementation of binary ivf flat index (bitwise hamming metric for the IVF Flat index)

### 1. **Binary Index Structure**
- Added `binary_centers_` field to store cluster centers as packed `uint8_t` arrays for binary data
- Index automatically detects BitwiseHamming metric and configures itself for binary operation
- Only support `uint8_t` inputs with BitwiseHamming and add only single instantiations of newly added kernels

### 2. **K-means Clustering for Binary Data**
The clustering approach for binary data required special handling:

- **Expanded Space Clustering**: Binary data (uint8_t) is expanded to signed representation (int8_t) where each bit becomes ±1
  - 0 → -1, 1 → +1 transformation enables meaningful centroid computation
  - Clustering performed using L2 distance in the expanded dimensional space
  
- **Centroid Quantization**: After computing float centroids in expanded space, they are converted back to binary format:
  - Centroids are stored as packed uint8_t arrays
  - KMeans (coarse) prediction is done on these quantized centroids with the BitwiseHamming distance.

#### Coarse Search (Cluster Selection)
- Implemented specialized `bitwise_hamming_distance_op` for query-to-centroid distances in order to compute `PairwiseDistances`

#### Fine-Grained Search (Within Clusters)  
Extended the interleaved scan kernel (`ivf_flat_interleaved_scan.cuh`) with specialized templates for BitwiseHamming:

- **Veclen-based optimization**: Different code paths based on vectorization width
  - Veclen=16,8,4: Load data as `uint32_t`, use `__popc(x ^ y)` for 4-byte Hamming distance
  - Veclen=1,2: Byte-wise XOR and population count

- **Efficient memory access patterns**: 
  - Maintains interleaved data layout for coalesced memory access
  - Specialized `loadAndComputeDist` templates for `uint8_t` that leverage vectorized loads

**as of 10/17/2025**
Binary size increase:
branch-25.12 (CUDA 12.9 + X86): 1232.414 MB
This PR (CUDA 12.9 + X86): 1251.051 MB


cpp/src/distance/detail/fused_distance_nn/simt_kernel.cuh
```diff
@@ -82,104 +83,115 @@ __launch_bounds__(P::Nthreads, 2) RAFT_KERNEL fusedDistanceNNkernel(OutT* min,
                                                                     OpT distance_op,
                                                                     FinalLambda fin_op)
 {
-// compile only if below non-ampere arch.
-#if __CUDA_ARCH__ < 800
```
tarang-jain, line 75: The diff here seems big, but the only real difference is the introduction of the if constexpr to compile for bitwise hamming distance if __CUDA_ARCH__ >= 800.

cpp/src/distance/detail/fused_distance_nn/simt_kernel.cuh
```diff
@@ -71,104 +83,115 @@ __launch_bounds__(P::Nthreads, 2) RAFT_KERNEL fusedDistanceNNkernel(OutT* min,
...
-                         DataT * regyn,
-                         IdxT gridStrideX,
-                         IdxT gridStrideY) {
-    KVPReduceOpT pairRed_op(pairRedOp);
-
-    // intra thread reduce
-    const auto acccolid = threadIdx.x % P::AccThCols;
-    const auto accrowid = threadIdx.x / P::AccThCols;
+    using AccT = std::conditional_t<std::is_same_v<DataT, uint8_t>, uint32_t, DataT>;
+    typedef raft::KeyValuePair<IdxT, AccT> KVPair;
```
tarang-jain, line 89: We use `uint32_t` as the distance type with `uint8_t` inputs so with this line, our KVP correctly mirrors that.

cpp/src/neighbors/ivf_flat_index.cpp
```diff
@@ -135,7 +164,11 @@ IdxT index<T, IdxT>::size() const noexcept
 template <typename T, typename IdxT>
 uint32_t index<T, IdxT>::dim() const noexcept
 {
-  return centers_.extent(1);
+  if (binary_index_) {
+    return binary_centers_.extent(1);
```
achirkin, line 168: Does that give the actual dataset dimensionality / 8? This feels a little bit unsettling from the user perspective. Maybe I'm a little bit out of the context, did we have the discussion about this already?

cpp/src/neighbors/ivf_flat/ivf_flat_search.cuh
```diff
@@ -94,76 +97,113 @@ void search_impl(raft::resources const& handle,
...
-                       (IdxT)n_queries,
-                       index.center_norms()->data_handle(),
-                       (IdxT)index.n_lists(),
-                       distance_buffer_dev.data(),
-                       stream);
-      RAFT_LOG_TRACE_VEC(index.center_norms()->data_handle(), std::min<uint32_t>(20, index.dim()));
-      RAFT_LOG_TRACE_VEC(distance_buffer_dev.data(), std::min<uint32_t>(20, index.n_lists()));
-      break;
+  if (index.metric() == cuvs::distance::DistanceType::BitwiseHamming) {
+    if constexpr (std::is_same_v<T, uint8_t>) {
```
achirkin, line 109: Maybe better invert the two conditions? So that the BitwiseHamming branch doesn't even appear in the code for other data types.

cpp/src/distance/detail/fused_distance_nn/fused_bitwise_hamming_nn.cuh
```diff
@@ -0,0 +1,73 @@
...
+                           ReduceOpT redOp,
+                           KVPReduceOpT pairRedOp,
+                           bool sqrt,
+                           cudaStream_t stream)
+{
+  typedef Policy P;
+
+  dim3 blk(P::Nthreads);
+  constexpr auto maxVal = std::numeric_limits<DataT>::max();
+  typedef ::raft::KeyValuePair<IdxT, OutT> KVPair;
```
achirkin: Somewhat opinionated nitpick: maybe better use `using` keyword as this is more common approach across the codebase
```C++
using kv_pair_type =  ::raft::KeyValuePair<IdxT, OutT>;
using distance_op_type = ops::bitwise_hamming_distance_op<DataT, uint32_t, IdxT>;
using fin_op_type = ::raft::identity_op;
...
kernel<<<...>>>(..., distance_op_type{k}, fin_op_type{});
```
But also.. is `KVPair` used at all?

cpp/src/cluster/detail/kmeans_balanced.cuh
```diff
@@ -207,6 +207,83 @@ inline std::enable_if_t<std::is_floating_point_v<MathT>> predict_core(
...
+    true,
+    params.metric,
+    0.0f,
+    stream);
+
+  thrust::transform(raft::resource::get_thrust_policy(handle),
+                    minClusterAndDistance.data_handle(),
+                    minClusterAndDistance.data_handle() + n_rows,
+                    labels,
+                    raft::compose_op<raft::cast_op<LabelT>, raft::key_op>());
```
achirkin: Can you reformulate the to calls to thrust in terms of `raft::linalg::map`?

cpp/src/cluster/detail/kmeans_balanced.cuh
```diff
@@ -207,6 +207,83 @@ inline std::enable_if_t<std::is_floating_point_v<MathT>> predict_core(
...
+                                    LabelT* labels,
+                                    rmm::device_async_resource_ref mr)
+{
+  RAFT_EXPECTS(params.metric == cuvs::distance::DistanceType::BitwiseHamming,
+               "uint8_t data only supports BitwiseHamming distance");
+
+  auto stream = raft::resource::get_cuda_stream(handle);
+
+  auto workspace = raft::make_device_mdarray<char, IdxT>(
+    handle, mr, raft::make_extents<IdxT>((sizeof(int)) * n_rows));
```
achirkin, line 203: Just checking: do you expect the two device allocations here surely fit the limited workspace? i.e. do we batch somewhere above along the call stack? If not, would it be better to use the large workspace resource?

achirkin (changes requested):
Thanks for the thoughtful work!
I feel a little bit unsettled by the way we treat the binary distance type.
What if we use some other type to designate the binary data, such as `std::byte`  or create our own on top of `uint8_t`? Then:
  1. We could parameterise the `centers_` type to be the same type - no need for an extra member `binary_centers_`.
  2. We would not introduce more kernel instance than we already do: `index<std::byte>` would support only the Hamming distance and the other types wouldn't support that distance
  3. It would be easier to show descriptive error messages for distance type not being supported.

cpp/src/neighbors/ivf_flat_index.cpp
tarang-jain: >Does that give the actual dataset dimensionality / 8?

This is correct (from a user's perspective). Essentially, number of bits / 8.

>Maybe I'm a little bit out of the context, did we have the discussion about this already?

We did not. So should I do this then:
`binary_centers_.extent(1) * 8;`

cpp/src/cluster/detail/kmeans_balanced.cuh
tarang-jain: raft::linalg::map does not seem to work with KVP type. We are using thrust in other places too in the same file.

cjnolet: I would also prefer if we can use raft where possible so we can shield ourself from thrust. What is the error you are getting when trying to use KVP w/ linalg::map? It should be device accessible.

cpp/cmake/thirdparty/get_raft.cmake
```diff
@@ -39,8 +39,8 @@ function(find_and_configure_raft)
             COMPONENTS          ${RAFT_COMPONENTS}
             CPM_ARGS
               EXCLUDE_FROM_ALL TRUE
-              GIT_REPOSITORY        https://github.com/${PKG_FORK}/raft.git
-              GIT_TAG               ${PKG_PINNED_TAG}
+              GIT_REPOSITORY        https://github.com/tarang-jain/raft.git
```
tfeher: Just a marking this so that we don't forget to revert once the RAFT PR is merged.

cpp/include/cuvs/neighbors/ivf_flat.hpp
```diff
@@ -204,6 +204,12 @@ struct index : cuvs::neighbors::index {
   raft::device_matrix_view<float, uint32_t, raft::row_major> centers() noexcept;
   raft::device_matrix_view<const float, uint32_t, raft::row_major> centers() const noexcept;
 
+  /** packed k-means cluster centers corresponding to the lists [n_lists, dim] when the
+   * BitwiseHamming metric is selected */
+  raft::device_matrix_view<uint8_t, uint32_t, raft::row_major> binary_centers() noexcept;
+  raft::device_matrix_view<const uint8_t, uint32_t, raft::row_major> binary_centers()
```
tfeher: Please use `int64_t` as matrix extent type (instead of `uint32_t`) if possible. I see that other internal arrays also use 32 bit index type, I guess you were following that pattern. We are currently discussing to refactor everything to use 64 bit extent types.

cpp/include/cuvs/neighbors/ivf_flat.hpp
```diff
@@ -267,7 +275,9 @@ struct index : cuvs::neighbors::index {
   std::vector<std::shared_ptr<list_data<T, IdxT>>> lists_;
   raft::device_vector<uint32_t, uint32_t> list_sizes_;
   raft::device_matrix<float, uint32_t, raft::row_major> centers_;
+  raft::device_matrix<uint8_t, uint32_t, raft::row_major> binary_centers_;
   std::optional<raft::device_vector<float, uint32_t>> center_norms_;
+  bool binary_index_ = metric_ == cuvs::distance::DistanceType::BitwiseHamming;
```
tfeher: Nitpick, but since all the other members are initialized in the [member initializer list of the constructor](https://github.com/rapidsai/cuvs/blob/2cf5fa7666d703dccbe655f8214656b0952bb69b/cpp/src/neighbors/ivf_flat_index.cpp#L34), I would prefer if we move this to the same place.

cpp/src/cluster/detail/kmeans_balanced.cuh
tfeher: We do batch before this is called.

cpp/src/cluster/detail/kmeans_balanced.cuh
```diff
@@ -207,6 +207,83 @@ inline std::enable_if_t<std::is_floating_point_v<MathT>> predict_core(
...
+
+  thrust::transform(raft::resource::get_thrust_policy(handle),
+                    minClusterAndDistance.data_handle(),
+                    minClusterAndDistance.data_handle() + n_rows,
+                    labels,
+                    raft::compose_op<raft::cast_op<LabelT>, raft::key_op>());
+}
+
+template <typename IdxT, typename LabelT>
+inline void predict_bitwise_hamming(const raft::resources& handle,
```
tfeher, line 234: Why do we need this overload? We have a variant above that takes `kmeans::balanced_params`, can't we just use that everywhere?

cpp/src/distance/detail/distance_ops/bitwise_hamming.cuh
```diff
@@ -0,0 +1,60 @@
...
+    return Policy::SmemSize;
+  }
+
+  __device__ __forceinline__ void core(AccT& acc, DataT& x, DataT& y) const
+  {
+    static_assert(std::is_same_v<DataT, uint8_t>, "BitwiseHamming only supports uint8_t");
+    // Ensure proper masking and casting to avoid undefined behavior
+    uint32_t xor_val    = static_cast<uint32_t>(static_cast<uint8_t>(x ^ y));
+    uint32_t masked_val = xor_val & 0xffu;
+    int popcount        = __popc(masked_val);
```
tfeher, line 46: This works, but it is inefficient to use `__popc` one byte at a time. Could you create an issue to note that there is a potential improvement if we use 4 byte storage type for x and y?

cpp/src/distance/detail/fused_distance_nn.cuh
```diff
@@ -68,16 +69,31 @@ void fusedDistanceNNImpl(OutT* min,
 
   switch (metric) {
     case cuvs::distance::DistanceType::CosineExpanded:
-      fusedCosineNN<DataT, OutT, IdxT, P, ReduceOpT, KVPReduceOpT>(
-        min, x, y, xn, yn, m, n, k, workspace, redOp, pairRedOp, sqrt, stream);
+      if constexpr (std::is_same_v<DataT, uint8_t> || std::is_same_v<DataT, int8_t>) {
+        assert(false && "Cosine distance is not supported for uint8_t/int8_t data types");
```
tfeher: These `asserts` will only trigger if the code is compiled in debug mode. I don't think this is what we want.

cpp/src/neighbors/detail/ann_utils.cuh
```diff
@@ -199,6 +199,27 @@ HDI constexpr auto mapping<int8_t>::operator()(const float& x) const -> int8_t
...
+
+  HDI constexpr auto operator()(const IdxT& i) -> OutT
+  {
+    IdxT row_id = i / uncompressed_dim;
+    IdxT col_id = i % uncompressed_dim;
+    return static_cast<OutT>(
+      -1 + 2 * static_cast<OutT>(
+                 (binary_vecs[row_id * compressed_dim + (col_id >> 3)] >> (col_id & 7)) & 1));
+  };
+};
```
tfeher, line 229: This becomes too complex and even the compiler will have difficulty simplifying it. In practice we just want to extract the i-th bit, so let's just do that.

Note: the compiler is clever enough to replace `/ 8` and `% 8` with the corresponding bit shift and logical ops, therefore I recommend to use the more readable form. In contrast div and mod with `uncompressed_dim` is costly. 

```suggestion
  bitwise_decode_op(const uint8_t* const binary_vecs)
    : binary_vecs(binary_vecs)
  {
  }
  const uint8_t* binary_vecs;

  /// Returns 1 if the i-th bit is 1, otherwise return -1.
  HDI constexpr auto operator()(const IdxT& i) -> OutT
  {
    return static_cast<OutT>((binary_vecs[i / 8] >> (i % 8)) & 1) ? 1 : -1);
  };
};
```

cpp/src/neighbors/ivf_flat/ivf_flat_interleaved_scan.cuh
```diff
@@ -1130,6 +1133,20 @@ struct inner_prod_dist {
   }
 };
 
+template <int Veclen, typename T, typename AccT>
```
tfeher: Why do we need template type `T`?

cpp/src/neighbors/ivf_flat/ivf_flat_search.cuh
```diff
@@ -94,76 +97,113 @@ void search_impl(raft::resources const& handle,
...
+                                                             raft::identity_op{},
+                                                             stream,
+                                                             true);
+
+      // Convert uint32_t distances to float for compatibility with rest of pipeline
+      raft::linalg::unaryOp(
+        distance_buffer_dev.data(),
+        uint32_distances.data(),
+        n_queries * index.n_lists(),
+        [] __device__(uint32_t val) { return static_cast<float>(val); },
```
tfeher:
```suggestion
        raft::cast_op<float>{},
```

cpp/src/neighbors/ivf_flat_index.cpp
tfeher: There is a certain logic to both variant. For `ivf_pq` dim returns the dimension of the original dataset, and I would prefer to do the same here. Please also improve the docstring in ivf_flat.hpp.

tfeher (changes requested):
Thank you @tarang-jain for this PR! It looks good overall, please find a few issues below.

One thing is not clear for me: why is `uint8_t `used internally during k-means training? Each vector is a bitstring, and `uint8_t` is just a storage type that we use. We could chose `uint32_t` as well to make the popcount ops more efficient.  Since this is an implementation detail, I do not think we need to fix that in the current PR.

cpp/src/neighbors/ivf_flat/ivf_flat_interleaved_scan.cuh
tarang-jain: We static_cast to uint32_t because __popc takes uint32_t. Can potentially combine the two.

cpp/src/distance/detail/fused_distance_nn/fused_bitwise_hamming_nn.cuh
tarang-jain: Fixed!

cpp/src/neighbors/ivf_flat/ivf_flat_search.cuh
tarang-jain: Fixed!

cpp/include/cuvs/neighbors/ivf_flat.hpp
tarang-jain: Done!

cpp/src/distance/detail/fused_distance_nn.cuh
tarang-jain: Changed to RAFT_FAIL

cpp/include/cuvs/neighbors/ivf_flat.hpp
tarang-jain: Done!

cpp/src/neighbors/ivf_flat_index.cpp
tarang-jain: I have kept this as is (returns the dimension of the original dataset) and added a note in ivf_flat.hpp. Marking it as resolved.

cpp/src/distance/detail/distance_ops/bitwise_hamming.cuh
tarang-jain: yeah the reason I did this was to avoid separate compilation paths for larger datatypes such as half and uint32. The idea was to dispatch to the corresponding type (uint32 or half) depending on the dimension (whether it is divisible by 4 or by 2 or not). But that would bloat the binary size because the distance op would have to be defined for these additional types.

cpp/src/cluster/detail/kmeans_balanced.cuh
tarang-jain: Because then we'd be compiling the non-BitwiseHamming paths for dtype = uint8 and distance dtype = uint32 for all the metrics. I did this for separate selective compilation.

tarang-jain:
@tfeher addressing your comment here:
> One thing is not clear for me: why is uint8_t used internally during k-means training? Each vector is a bitstring, and uint8_t is just a storage type that we use. We could chose uint32_t as well to make the popcount ops more efficient. Since this is an implementation detail, I do not think we need to fix that in the current PR.

The problem with that it wont allow number of bytes (dim) which is not divisible by 4 if we use uint32_t. uint8_t would be the most inclusive for dims.

cpp/cmake/thirdparty/get_raft.cmake
tarang-jain: The RAFT pr was merged. Marking this one as resolved.

cpp/src/neighbors/ivf_flat/ivf_flat_interleaved_scan.cuh
tfeher: We do not have any input or output with type T, therefore I do not see why we need this param. Or is the static assert participating SFINAE logic?

In any case, this is does not need to hold the PR.

cpp/src/distance/detail/distance_ops/bitwise_hamming.cuh
tfeher: Do we expect use cases where padding dims to be divisible by 32 would be a problem? But this discussion need not hold this PR, I  have created https://github.com/rapidsai/cuvs/issues/1613 to continue this discussion.

tfeher (approved):
Hi Tarang, thank you for your work on this PR, it looks good to me!

cpp/src/distance/detail/distance_ops/bitwise_hamming.cuh
tarang-jain: That actually makes a lot of sense. Padding dims to be divisible by 32 and using uint32_t seems like the more optimal approach.

cpp/src/cluster/detail/kmeans_balanced.cuh
tarang-jain: Created https://github.com/rapidsai/raft/pull/2913 to support KVP in `raft::linalg::map`

PR #1104, opened 2025-07-11
[Java]Initial changes for binary and scalar quantization

punAhuja:


java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/BinaryQuantizerImpl.java
```diff
@@ -0,0 +1,175 @@
...
+  private static byte[][] performTransform(
+      CuVSResourcesImpl resources,
+      Arena localArena,
+      long rows,
+      long cols,
+      MemorySegment datasetMemSegment,
+      Util.CudaMemcpyKind memcpyDirection)
+      throws Throwable {
+
+    long cuvsResourcesPtr = resources.getMemorySegment().get(cuvsResources_t, 0);
```
ldematte: You may want to rebase your PR, this is no longer needed (we have a direct getter for the native resource handle now)

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/BinaryQuantizerImpl.java
```diff
@@ -0,0 +1,175 @@
...
+
+    // Allocate device memory for input and output
+    MemorySegment datasetD = localArena.allocate(C_POINTER);
+    MemorySegment outputD = localArena.allocate(C_POINTER);
+
+    long datasetBytes = C_FLOAT_BYTE_SIZE * rows * cols;
+    long outputCols = (cols + 7) / 8; // Ceiling division for bit packing
+    long outputBytes = C_BYTE_SIZE * rows * outputCols;
+
+    int returnValue = cuvsRMMAlloc(cuvsResourcesPtr, datasetD, datasetBytes);
```
ldematte: Similarly, https://github.com/rapidsai/cuvs/pull/1069 introduces an `allocateRMMSegment` utility to replace the need to allocate + get (but that has not been merged yet).

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/BinaryQuantizerImpl.java
```diff
@@ -0,0 +1,175 @@
...
+
+    try {
+      // Copy input data to device
+      cudaMemcpy(datasetPtr, datasetMemSegment, datasetBytes, memcpyDirection);
+
+      // Prepare tensors
+      long datasetShape[] = {rows, cols};
+      long outputShape[] = {rows, outputCols};
+      MemorySegment datasetTensor =
+          prepareTensor(localArena, datasetPtr, datasetShape, 2, 32, 2, 2, 1);
```
ldematte: Nit: it would be better to use named constants for these, e.g. the device would be `kDLCUDA()`, the "code" would be `kDLFloat()`.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/BinaryQuantizerImpl.java
```diff
@@ -0,0 +1,175 @@
...
+
+      returnValue = cuvsStreamSync(cuvsResourcesPtr);
+      checkCuVSError(returnValue, "cuvsStreamSync");
+
+      // Copy result back to host
+      MemorySegment outputMemSegment = localArena.allocate(C_CHAR, outputBytes);
+      cudaMemcpy(outputMemSegment, outputPtr, outputBytes, DEVICE_TO_HOST);
+
+      // Unpack bits to individual byte array
+      return unpackBinaryResult(outputMemSegment, rows, cols, outputCols);
```
ldematte: This will force users of these functions to use a heap array; that could lead to an additional memory allocation that in some cases would be preferable to avoid (e.g. if the caller wants to keep the results in native memory, or save them to a file directly).
It's true that we don't want to expose a MemorySegment directly, as we want to keep the API Java21 compatible, but what about using something like
```
interface ByteMatrix implements AutoCloseable {
   byte get(int row, int col);
   byte[][] toArray();
}
...
      var resultArena = Arena.ofShared():
      MemorySegment outputMemSegment = resultArena.allocate(C_CHAR, outputBytes);
      cudaMemcpy(...);
      
      return new ByteMatrix() {
          byte get(int row, int col) {
             // implement via outputMemSegment access
          }
          void close() {
             resultArena.close();
           }
         byte[][] toArray() {
             return unpackBinaryResult(outputMemSegment, rows, cols, outputCols);
         }
     };
``` 
(just a sketch, probably is a good idea to make this a inner class and add checks on rows/cols)
This way the caller can decide to keep things in native memory and read bytes directly from there, or defer the transformation to an on-heap Java array.

(I'm working on something similar here: https://github.com/ldematte/cuvs/blob/java/cagra-read-graph-dataset/java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/CagraIndexImpl.java#L428)

ldematte: Or why not use (and extend) `Dataset` (so that it supports byte[][] too)? So it can be used as input for e.g. index build (in the future?)

punAhuja:
PR #716 introduced a bug in the binary.h file which caused jextract to fail and broke the Java bindings generation.

As part of this PR, I’ve resolved the issue and successfully generated the necessary bindings for binary.h.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/BinaryQuantizerImpl.java
punAhuja: Using the direct getter for resources now.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/BinaryQuantizerImpl.java
```diff
@@ -0,0 +1,212 @@
...
+
+      MemorySegment datasetMemSegment = ((CuVSMatrixBaseImpl) dataset).memorySegment();
+      long cuvsResourcesPtr = resourcesAccessor.handle();
+
+      // TODO: GPU Quantization Optimization Needed
+      // Current implementation is inefficient due to:
+      // 1. Host-to-device copy of input dataset
+      // 2. Device-to-host copy of quantized results
+      //
+      // Future improvement: Support GPU-native dataset creation in Java API
```
ldematte: 👍 
Thanks for the TODO, I'm actually working on this now so having places where I can try and plug it into it's valuable.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/CagraIndexImpl.java
```diff
@@ -225,19 +255,33 @@ public SearchResults search(CagraQuery query) throws Throwable {
...
+      if (query.hasQuantizedQueries()) {
+        CuVSMatrix quantizedQueries = query.getQuantizedQueries();
+        numQueries = quantizedQueries.size();
+        vectorDimension = (int) quantizedQueries.columns();
+        floatsSeg = ((CuVSMatrixBaseImpl) quantizedQueries).memorySegment();
+      } else {
+        float[][] queryVectors = query.getQueryVectors();
+        numQueries = queryVectors.length;
+        vectorDimension = numQueries > 0 ? queryVectors[0].length : 0;
+        floatsSeg = buildMemorySegment(localArena, queryVectors);
```
ldematte, line 294: Nit: wouldn't it be better to do this earlier, and avoid carrying on "mutually-null" fields? Like we did with datasets? (as inputs of `build()`)

java/cuvs-java/src/main/java/com/nvidia/cuvs/CagraQuery.java
```diff
@@ -206,12 +240,41 @@ public Builder withPrefilter(BitSet prefilter, int numDocs) {
...
+      }
+
+      CuVSMatrix quantized = null;
+      float[][] floatsForQuery = queryVectors;
+
+      if (quantizer != null) {
+        // wrap float[][] in a CuVSMatrix and quantize
+        CuVSMatrix tmp = CuVSMatrix.ofArray(queryVectors);
+        if (tmp.precision() != 32) {
+          tmp.close();
```
ldematte: Nit: this could be a try-with-resources

java/cuvs-java/src/main/java/com/nvidia/cuvs/CuVSMatrix.java
```diff
@@ -68,6 +68,12 @@ static CuVSMatrix ofArray(byte[][] vectors) {
     return CuVSProvider.provider().newMatrixFromArray(vectors);
   }
 
+  /**
+   * Returns the bit precision of this matrix data.
+   * @return The bit precision (1 for binary, 8 for 8-bit, 32 for float32)
+   */
+  int precision();
```
ldematte: I have mixed feelings about this -- I see the value of keeping the data type and the precision separate, but of course not every combination has sense or is possible. Should we at least add validation?

ldematte: I mean, if the implementation is the one in CuVSHostMatrixImpl, I don't see the point here.
If we allow BYTE to have precision 1, it's a different story. IDK, should we just add "BIT" and use the `DataType`?

punAhuja: I was using just precision earlier, but after I pulled the latest upstream changes, I noticed DataType. I think we can just keep DataType, looking into this.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/CagraIndexImpl.java
punAhuja: Have addressed this in latest commit.

java/cuvs-java/src/main/java/com/nvidia/cuvs/CagraQuery.java
punAhuja: Have addressed this in latest commit.

java/cuvs-java/src/main/java/com/nvidia/cuvs/CuVSMatrix.java
punAhuja: Have addressed this in latest commit. We actually dont need BIT, BinaryQuantizer returns packed bits in 8bit unsigned integers.

cpp/include/cuvs/preprocessing/quantize/binary.h
```diff
@@ -30,7 +30,7 @@ extern "C" {
  * and sampling_median thresholds are calculated separately for each dimension.
  *
  */
-enum cuvsBinaryQuantizerThreshold { ZERO, MEAN, SAMPLING_MEDIAN };
+typedef enum { ZERO, MEAN, SAMPLING_MEDIAN }cuvsBinaryQuantizerThreshold;
```
mythrocks:
```suggestion
typedef enum { ZERO, MEAN, SAMPLING_MEDIAN } cuvsBinaryQuantizerThreshold;
```

cpp/include/cuvs/preprocessing/quantize/binary.h
```diff
@@ -39,12 +39,12 @@ struct cuvsBinaryQuantizerParams {
   /*
    * specifies the threshold to set a bit in cuvsBinaryQuantizerTransform
    */
-  cuvsBinaryQuantizerThreshold threshold = MEAN;
```
mythrocks, line 42: I can see why this change is required from the Java/Panama side.  

I wouldn't mind confirmation from @benfred / @cjnolet that this is acceptable on the C++ side.  I don't see where this is being compensated for, anywhere.  The disappearance of the default initializers will hopefully not surprise non-Java users of this header.

java/cuvs-java/src/main/java/com/nvidia/cuvs/spi/CuVSProvider.java
```diff
@@ -81,6 +81,9 @@ default Path nativeLibraryPath() {
   /** Create a {@link CuVSMatrix} from an on-heap array **/
   CuVSMatrix newMatrixFromArray(byte[][] vectors);
 
+  /** Create a {@link CuVSMatrix} backed by a on-heap byte array */
+  CuVSMatrix newByteArrayDataset(byte[][] vectors);
```
mythrocks: I'm not sure I've understood why there is a new interface function for what looks near identical to `newMatrixFromArray()`.  Is there a semantic distinction?

The implementation seems to be forwarding to the previously available function.

java/cuvs-java/src/main/java/com/nvidia/cuvs/spi/CuVSProvider.java
```diff
@@ -120,6 +123,22 @@ default CagraIndex mergeCagraIndexes(CagraIndex[] indexes, CagraMergeParams merg
...
 
+  Object createScalar8BitQuantizerImpl(CuVSResources resources, CuVSMatrix trainingDataset)
+      throws Throwable;
+
+  CuVSMatrix inverseTransformScalar8Bit(Object impl, CuVSMatrix quantizedData) throws Throwable;
+
+  CuVSMatrix transformBinary(CuVSResources resources, CuVSMatrix input, int thresholdType)
+      throws Throwable;
+
+  CuVSMatrix transformBinary(CuVSResources resources, CuVSMatrix input) throws Throwable;
```
mythrocks: These look like they could use descriptive documentation.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/spi/JDKProvider.java
```diff
@@ -188,4 +188,42 @@ public CuVSMatrix newMatrixFromArray(byte[][] vectors) {
...
+
+  @Override
+  public Object createScalar8BitQuantizerImpl(CuVSResources resources, CuVSMatrix trainingDataset)
+      throws Throwable {
+    return Scalar8BitQuantizerImpl.create(resources, trainingDataset);
+  }
+
+  @Override
+  public CuVSMatrix transformScalar8Bit(Object impl, CuVSMatrix input) throws Throwable {
+    return ((Scalar8BitQuantizerImpl) impl).transform(input);
```
mythrocks, line 200: I'm not sure I've understood the asymmetry between how `BinaryQuantizer` can work without needing an `impl` `Object`, but `Scalar8bitQuantizer` needs one.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/BinaryQuantizerImpl.java
mythrocks: Please be aware that I'm considering moving `allocateRMMSegment` from `Util` to `CloseableRMMAllocation` in #1215.  There might be some minor changes in this bit.

java/cuvs-java/src/main/java/com/nvidia/cuvs/spi/CuVSProvider.java
mythrocks: It might also be good to group the 8-bit quantizer methods together, and apart form the Binary quantizer methods.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/spi/JDKProvider.java
mythrocks: The `Scalar8bitQuantizer` has state that needs to be kept alive, something that `BinaryQuantizer` does not seem to share?

java/cuvs-java/src/main/java22/com/nvidia/cuvs/spi/JDKProvider.java
```diff
@@ -188,4 +188,42 @@ public CuVSMatrix newMatrixFromArray(byte[][] vectors) {
...
+    return ((Scalar8BitQuantizerImpl) impl).inverseTransform(quantizedData);
+  }
+
+  @Override
+  public CuVSMatrix transformBinary(CuVSResources resources, CuVSMatrix input) throws Throwable {
+    return BinaryQuantizerImpl.transform(resources, input, 0);
+  }
+
+  @Override
+  public CuVSMatrix transformBinary(CuVSResources resources, CuVSMatrix input, int thresholdType)
```
mythrocks: Regarding `thresholdType`:  Shouldn't this be using the `ThresholdType` defined in`BinaryQuantizer`?

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/BinaryQuantizerImpl.java
```diff
@@ -0,0 +1,273 @@
...
+          cuvsBinaryQuantizerDestroy(quantizerPtr);
+          cuvsBinaryQuantizerParamsDestroy(paramsPtr);
+        }
+
+      } finally {
+        int returnValue = cuvsRMMFree(cuvsResourcesPtr, datasetPtr, datasetBytes);
+        checkCuVSError(returnValue, "cuvsRMMFree for dataset");
+
+        returnValue = cuvsRMMFree(cuvsResourcesPtr, outputPtr, outputBytes);
+        checkCuVSError(returnValue, "cuvsRMMFree for output");
```
mythrocks: Rather than allocating and freeing every time in the loop, it might be viable to allocate `(ROWS_PER_BUFFER * cols * element_size)` at the beginning, and free it at the very end.

I suspect only the last iteration would have more space than content.  It would still beat having to go back and forth, allocating/deallocating memory, in every loop.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/Scalar8BitQuantizerImpl.java
```diff
@@ -0,0 +1,460 @@
...
+
+        MemorySegment resultChunk =
+            ((CuVSMatrixBaseImpl) result).memorySegment().asSlice(startRow * cols, outputBytes);
+        cudaMemcpy(resultChunk, outputPtr, outputBytes, DEVICE_TO_HOST);
+
+      } finally {
+        int returnValue = cuvsRMMFree(cuvsResourcesPtr, datasetPtr, datasetBytes);
+        checkCuVSError(returnValue, "cuvsRMMFree");
+        returnValue = cuvsRMMFree(cuvsResourcesPtr, outputPtr, outputBytes);
+        checkCuVSError(returnValue, "cuvsRMMFree");
```
mythrocks: I left a similar note in the code for `BinaryQuantizerImpl`:  It might be better to allocate the work-slice for both the input dataset and the output, at the top, outside the loop.  We can then free it after the loop has exited.  

It would save the constant allocation/deallocation within the loop.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/Scalar8BitQuantizerImpl.java
```diff
@@ -0,0 +1,460 @@
...
+            cuvsScalarQuantizerTransform(
+                cuvsResourcesPtr, quantizerSeg, datasetTensor, outputTensor);
+        checkCuVSError(returnValue, "cuvsScalarQuantizerTransform");
+
+        returnValue = cuvsStreamSync(cuvsResourcesPtr);
+        checkCuVSError(returnValue, "cuvsStreamSync");
+
+        MemorySegment resultChunk =
+            ((CuVSMatrixBaseImpl) result).memorySegment().asSlice(startRow * cols, outputBytes);
+        cudaMemcpy(resultChunk, outputPtr, outputBytes, DEVICE_TO_HOST);
```
mythrocks: ~~I'm trying to ascertain whether we need a stream-sync here.  My guess is that we don't, because we're not using `cudaMemcpyAsync`.~~

Nope.  No need for stream-sync.

java/cuvs-java/src/main/java/com/nvidia/cuvs/spi/CuVSProvider.java
ldematte: ++ 
Probably a leftover, but it would be so much better to just use newMatrixFromArray and avoid introducing this

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/BinaryQuantizerImpl.java
ldematte: ++ to bring this out of the loop

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/Scalar8BitQuantizerImpl.java
ldematte: I agree, cudaMemcpy is sync (non-returning, i.e. CPU sync) and non-overlapping (stream-sync)

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/BinaryQuantizerImpl.java
punAhuja: Addressed this in latest commits.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/Scalar8BitQuantizerImpl.java
punAhuja: Addressed this in latest commits.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/spi/JDKProvider.java
punAhuja: Was using int only in the implementation part. Using enums everywhere for ThresholdType now.

java/cuvs-java/src/main/java/com/nvidia/cuvs/spi/CuVSProvider.java
punAhuja: Yes, realized after pulling latest changes. Addressed this in latest commits.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/BinaryQuantizerImpl.java
```diff
@@ -0,0 +1,273 @@
...
+              kDLCPU(),
+              1);
+
+      returnValue =
+          cuvsBinaryQuantizerTransformWithParams(
+              cuvsResourcesPtr, quantizerPtr, datasetTensor, outputTensor);
+      checkCuVSError(returnValue, "cuvsBinaryQuantizerTransformWithParams (host)");
+
+      return result;
+
```
mythrocks, line 257:
```suggestion
```

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/BinaryQuantizerImpl.java
mythrocks: > Addressed this in latest commits.

@punAhuja:  Maybe I'm missing something here.  I don't see how this concern was addressed.  The allocation and `cuvsRMMFree` still seem to be within the for-loop.  The suggestion was to move this out of the loop, and reuse the memory that's allocated once.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/BinaryQuantizerImpl.java
mythrocks: @punAhuja:  What of this suggestion from @ldematte?

punAhuja: This is now being handled in CuVSMatrix. CuVSMatrix was not introduced when this suggestion was made.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/BinaryQuantizerImpl.java
punAhuja: You are right, I had just incorporated CloseableRMMAllocation here. Moved allocation outside the loop now. Apologies for the misunderstanding.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/BinaryQuantizerImpl.java
ldematte: Yep, what I was suggesting eventually folded into CuVSMatrix, so using it is good! Thanks for addressing this.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/BinaryQuantizerImpl.java
mythrocks: Thanks for fixing this.

mythrocks (approved):
Thank you for your patience with this review.

mythrocks:
@benfred:  Would you mind having a look at the tiny change in `binary.h` here?  I can't approve on behalf of cpp-owners.

The header has been changed to un-C++-ify a `typedef struct`, so as to make it available to Java/JNI.

cjnolet (approved):

cjnolet:
Can we please halt on this feature and put all of our effort to finish getting milestone 1 all the way through the finish like for the POR? This feature is not even in the POR. We have customers and users expecting us to deliver the milestones by a specific timeline.

thank you for understanding.

mythrocks:
> Can we please halt on this feature and put all of our effort to finish getting milestone 1 all the way through...

ACK.  Agreed, @cjnolet.  This feature doesn't contribute to the goals on our current milestone.

If this PR lands up needing more work before merging, we should park it and refocus.

mythrocks:
~~/merge~~

mythrocks:
Argh, this was a mistake.  #1232 was to go in ahead of this PR.

@punAhuja, I apologize, but I'd better revert this change.  We can rebase this on top, when we revisit.

PR #1105, opened 2025-07-11
[WIP][Java] Java API for Cagra graph accessors

ldematte:
Based on https://github.com/rapidsai/cuvs/pull/1086

Exposes the off-heap (host/CPU) copy of a CAGRA graph to consumers, in a Java21 friendly API.
WIP (will need rebase after #1086 is merged).


java/cuvs-java/src/main/java/com/nvidia/cuvs/CagraGraph.java
```diff
@@ -0,0 +1,62 @@
...
+package com.nvidia.cuvs;
+
+import com.nvidia.cuvs.spi.CuVSProvider;
+
+/**
+ * Represent a CAGRA graph backed by off-heap memory.
+ *
+ * @since 25.08
+ */
+public interface CagraGraph extends AutoCloseable {
```
ldematte: The more I look at this, the more I think "what if we collapse all into `Dataset`"?

What if `Dataset` becomes an interface to construct, carry around and access off-heap matrices (of a set of primitive types we need, e.g. starting with 8 and 32 bit integers and 32 bit floats?)

We can then provide uniform utility functions to e.g. copy them to on-heap arrays, wrap them in DLTensors, etc.

For now just an idea I'm kicking around to see if it makes sense.

java/cuvs-java/src/main/java/com/nvidia/cuvs/spi/CuVSProvider.java
```diff
@@ -52,9 +53,12 @@ default Path nativeLibraryPath() {
   /** Create a {@link Dataset.Builder} instance **/
   Dataset.Builder newDatasetBuilder(int size, int dimensions);
 
-  /** Create a {@link Dataset} backed by a on-heap array **/
+  /** Create a {@link Dataset} from an on-heap array **/
   Dataset newArrayDataset(float[][] vectors);
 
+  /** Create a {@link CagraGraph} from an on-heap array **/
+  CagraGraph newArrayGraph(int[][] graph);
```
ldematte: This is here mainly for test purposes; I do not anticipate a scenario where a cagra graph is built in Java, but it's useful to have it as a utility.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/CagraGraphImpl.java
```diff
@@ -0,0 +1,104 @@
...
+  @Override
+  public IntList getNeighbours(long nodeIndex) {
+    return new SliceIntList(
+        memorySegment()
+            .asSlice(nodeIndex * graphDegree * C_INT_BYTE_SIZE, graphDegree * C_INT_BYTE_SIZE),
+        graphDegree);
+  }
+
+  @Override
+  public int get(int row, int col) {
```
ldematte: I'm unsure if we need/want to keep this; on the one hand, if we end collapsing CagraGraph into Dataset, it makes sense. On the other hand, if we don't need "random" access, maybe it's better to enforce row access via `getNeighbours`?

java/cuvs-java/src/main/java/com/nvidia/cuvs/CagraGraph.java
ldematte: Expanding on this: `Dataset` can mimic DLTensor and have a "device" type; this way we could even have a _device_ memory. Of course, reading it would be limited to copy to a whole array or row-by-row (not single element access), but it could be used the same as it is used internally by cuvs, copying only when needed.

I'm not 100% sure of this, as it will have to carefully manage memory scope and lifetime (even more critical if it is GPU memory!), but if we want to do it, better to do it in one place, where we reuse well tested and verified code, instead of scattering GPU memory allocations/free through the codebase.

ldematte: I've implemented this in the last commit, and I think it's a powerful interface. I'll extract the changes in a separate PR so it can be reviewed separately, but it looks to work very well in this case.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/CagraGraphImpl.java
ldematte: I'll remove this from the final "Dataset" PR; it limits our freedom of implementation and does not give us a real advantage over the "row access" API.

java/cuvs-java/src/main/java/com/nvidia/cuvs/CagraGraph.java
chatman: > The more I look at this, the more I think "what if we collapse all into Dataset"?

I feel we should keep the "dataset" and the "graph" as two distinct abstractions, whereby the "dataset" (as the name suggests) contains the raw vectors, and the "graph" contains the graph/connections. If at all we want to unify those interfaces, then we should consider something different from "dataset" (as the name of the unified entity) to be more precise.

ldematte: I second your idea for a better name. I think we could have a "Matrix" or similar data structure that represents "something that holds 2D data whenever they are and whatever they are". 
Then if we feel we need to make a distinction, we can have lightweight Dataset/Graph classes that use it.

cjnolet:

mythrocks:
@ChrisHegarty:  #1086 has been merged.  This PR will need a rebase and update.  As will #1111, actually.

I'll keep you posted in case I get to #1111 before you do.

ldematte:
Superseded by #1216

PR #1106, opened 2025-07-11
[Java] Fix POM

### LINKED ISSUE #1090: [BUG][Java] The JAR with dependencies (for maven repository installation) is broken
We generate a "jar-with-dependencies" jar to be uploaded to maven (local), but for some reason this jar contains a lot of extraneous files, including multiple `module-info.class` and a `MANIFEST.mf` file with wrong entries. As such, the jar is not consumable from a modularized Java application.

(my suspicion is that we added the `spotless` dependency in the wrong way and that is now being jammed into the jar, but it's completely unverified)

ldematte:
As reported in https://github.com/rapidsai/cuvs/issues/1090, we are producing a "broken" "jar with dependecies" today: the Jar contains extra files that make it impossible to consume from a modularized application, as it breaks the module definition.

This PR fixes this by:
- removing `spotless` as a dependency: adding it as a dependency is not needed, as spotless is used at build time
- ~~removing the `jar-with-dependency` build task: we don't need it, as we don't have dependencies today. If we have in the future, we can re-introduce it, with a better handling of the merge process~~
- change the build script to install in the local maven repository the "regular" jar produced by the build process

EDIT: after review, we decided to revert the removal of the task creating  `jar-with-dependency`, as we might need it in the future (will need fixing anyway, but we can defer that until we need it).

Closes #1090

### DIFF
--- a/java/build.sh
+++ b/java/build.sh
@@ -40,5 +40,5 @@ fi
 export LD_LIBRARY_PATH=${CURDIR}/../cpp/build:${LD_LIBRARY_PATH}
 cd cuvs-java
 mvn verify "${MAVEN_VERIFY_ARGS[@]}" \
-  && mvn install:install-file -Dfile=./target/cuvs-java-$VERSION-jar-with-dependencies.jar -DgroupId=$GROUP_ID -DartifactId=cuvs-java -Dversion=$VERSION -Dpackaging=jar \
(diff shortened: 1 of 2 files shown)


ldematte:
@mythrocks I have updated the PR to keep the  `jar-with-dependencies`

mythrocks:
The CI builds should be fixed once #1114 is resolved.

mythrocks:
Will prioritize #1034 first, to address the build failures.

mythrocks:
@KyleFromNVIDIA, could I please bother you for a packaging-owners review for this change?

bdice (approved):

PR #1102, opened 2025-07-10
[WIP][Java] Exposing CAGRA graph

chatman:
Initial attempt at exposing the CAGRA graph exposed via the Java API. This is currently based on @benfred's PR, I'll change the base branch to branch-25.08 once #1086 is merged there.

PR #1110, opened 2025-07-12
[Review][Java] Fix HNSW params allocation

ldematte:
Similar to https://github.com/rapidsai/cuvs/pull/1109, but for HNSW index type (a bit simpler).

PR #1111, opened 2025-07-14
[Java] Extend `Dataset` to work as an output data container

ldematte:
In #902 and https://github.com/rapidsai/cuvs/pull/1034 we introduced a `Dataset` interface to support on-heap and off-heap ("native") memory seamlessly as inputs for cagra and bruteforce index building.

As we expand the functionality of cuvs-java, we realized we have similar needs for outputs (see e.g. https://github.com/rapidsai/cuvs/pull/1105 / https://github.com/rapidsai/cuvs/pull/1102 or https://github.com/rapidsai/cuvs/pull/1104).

This PR extends `Dataset` to support being used as an output, wrapping native (off-heap) memory in a convenient and efficient way, and providing common utilities to transform to and from on-heap memory.
This work is inspired by the existing raft `mdspan` and `DLTensor` data structures, but tailored to our needs (2d only, just 3 data types, etc.). The PR keeps the current implementation simple and minimal on purpose, but structured in a way that is simple to extend.

By itself, the PR is just a refactoring to extend the `Dataset` implementation and reorganize the implementation classes; its real usefulness will be in using it in the PRs mentioned above (in fact, this PR has been extracted from https://github.com/rapidsai/cuvs/pull/1105).
The implementation class hierarchy is implemented with future extensions in mind: atm we have one `HostMemoryDatasetImpl`, but we are already thinking to have a corresponding `DeviceMemoryDatasetImpl` that will wrap and manage (views) on GPU memory to avoid (in some cases) extra copies of data from GPU memory to CPU memory only to process them or forward them to another algorithm (e.g quantization followed by indexing).

Future work will also include add support/refactoring to allocate and manage GPU memory and DLTensors (e.g. working better with/refactoring `prepareTensor`).

### DIFF
--- a/cpp/bench/ann/src/common/util.hpp
+++ b/cpp/bench/ann/src/common/util.hpp
@@ -449,10 +449,8 @@ inline auto cuda_info()
   props.emplace_back("gpu_pageableMemoryAccess", std::to_string(device_prop.pageableMemoryAccess));
   props.emplace_back("gpu_pageableMemoryAccessUsesHostPageTables",
                      std::to_string(device_prop.pageableMemoryAccessUsesHostPageTables));
-#if defined(CUDART_VERSION) && CUDART_VERSION >= 12000
   props.emplace_back("gpu_gpuDirectRDMASupported",
                      std::to_string(device_prop.gpuDirectRDMASupported));
-#endif
 #endif
   return props;
 }
--- a/cpp/cmake/modules/FindSVE.cmake
+++ b/cpp/cmake/modules/FindSVE.cmake
@@ -1,0 +1,45 @@
+# Copyright (c) 2025, NVIDIA CORPORATION.
+# =============================================================================
+# FindSVE.cmake This module finds the SVE (Scalable Vector Extension) support in the compiler.
+# =============================================================================
+
+INCLUDE(CheckCXXSourceRuns)
+
+set(SVE_CODE
+    "
+  #include <arm_sve.h>
+
+  int main()
+  {
+    svfloat32_t a = svdup_f32(0);
+    return 0;
+  }
+"
+)
+
+# Check for SVE support
+message("Checking for SVE support")
+set(CMAKE_REQUIRED_FLAGS "-march=armv9-a+sve")
+check_cxx_source_runs("${SVE_CODE}" CXX_HAS_SVE)
+
+if(CXX_HAS_SVE)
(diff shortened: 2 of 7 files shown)


ldematte:
As @chatman pointed out in https://github.com/rapidsai/cuvs/pull/1105, this needs a better name, as it would not be a "Dataset" neither a "Graph" in the lexicon of cuvs.
@mythrocks what about `CuVSMatrix` or `CuVSArray2d`?

cjnolet:

mythrocks:
> @mythrocks what about `CuVSMatrix` or `CuVSArray2d`?

Sorry for the delay.  (It has been a packed day, today.)  I like the idea of calling this the `CuVSMatrix`. 👍

There might be value in differentiating whether the matrix is in `__host__` or `__device__` memory.  Separate types:  `CuVSDeviceMatrix` vs `CuVSHostMatrix`?

cjnolet:
> There might be value in differentiating whether the matrix is in __host__ or __device__ memory. Separate types: CuVSDeviceMatrix vs CuVSHostMatrix?

+1 for this. We do this in all of our other APIs (python, c++) and it really helps us keep them clean and relatively self-documenting

ldematte:
I like the idea of separate types, I was already planning to do this at implementation level (e.g. `HostMemoryDatasetImpl`). I can have intermediate CPU/GPU interfaces, but I'd like to keep a common ancestor. So we'll have `CuVSMatrix`, which can be either, and sub-interfaces `CuVSDeviceMatrix` and `CuVSHostMatrix`.
Why the common type? So when we want, we can be specific. When we don't want to be specific, we can pass the more generic type to any cuvs function and choose the path to take "dynamically" like we do like in the C API: `if (isGPU(matrix)) then do() else copyToGPUMemory(); do();` (very pseudo).

WDYT?

mythrocks:
Rebasing this PR turned out to be a heavier lift than anticipated.  Still WIP.

mythrocks:
This rebase wasn't easy, but it's complete:  https://github.com/mythrocks/cuvs/tree/fork/ldematte/java/fix-cuvs-params-creation-2

This might need modification once #1110 goes through.  I'll take a look once #1110 is merged.

mythrocks (approved):
This LGTM.  There are a few nitpicks/clarifications that can wait till a follow up.

PR #1112, opened 2025-07-14
Vamana C / Python API

tarang-jain:
Fixes https://github.com/rapidsai/cuvs/issues/853


python/cuvs/cuvs/neighbors/vamana/__init__.py
```diff
@@ -0,0 +1,23 @@
+# Copyright (c) 2024-2025, NVIDIA CORPORATION.
```
cjnolet: Please make sure all new files only have the current year.

cjnolet:
Looks like this is nearly there, but needs some tests.

cpp/include/cuvs/neighbors/vamana.h
```diff
@@ -0,0 +1,227 @@
...
+#include <dlpack/dlpack.h>
+#include <stdbool.h>
+#include <stdint.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/**
+ * @defgroup vamana_c_index_params C API for Vamana index build
```
benfred, line 31: Can we add this to the RST files in `docs/source/c_api` ?

benfred (commented):
looks good! 

Can we make sure that the docs are added to both the C-API and Python sections in the RST files ? Aside from that I think this is ready to merge

cpp/include/cuvs/neighbors/vamana.h
```diff
@@ -0,0 +1,227 @@
...
+  /** Distance type. */
+  cuvsDistanceType metric;
+  /** Maximum degree of output graph corresponds to the R parameter in the original Vamana
+   * literature. */
+  uint32_t graph_degree;
+  /** Maximum number of visited nodes per search corresponds to the L parameter in the Vamana
+   * literature **/
+  uint32_t visited_size;
+  /** Number of Vamana vector insertion iterations (each iteration inserts all vectors). */
+  uint32_t vamana_iters;
```
bkarsin: `vamana_iters` is a float value now. Was changed a few PRs ago to better be able to control recall.

python/cuvs/cuvs/neighbors/vamana/vamana.pxd
```diff
@@ -0,0 +1,80 @@
...
+from cuvs.distance_type cimport cuvsDistanceType
+
+
+cdef extern from "cuvs/neighbors/vamana.h" nogil:
+
+    ctypedef struct cuvsVamanaIndexParams:
+        cuvsDistanceType metric
+        uint32_t graph_degree
+        uint32_t visited_size
+        uint32_t vamana_iters
```
bkarsin: Same as previous comment - is a `float` value now

bkarsin (changes requested):
Looks good to me - just the one issue with `vamana_iters` being uint32 instead of float.

python/cuvs/cuvs/neighbors/vamana/vamana.pxd
tarang-jain: Done!

bkarsin (approved):
Looks good, thanks for adding this support!

PR #1120, opened 2025-07-16
[WIP] SCaNN Index build

rmaschal:
This PR gives a proof-of-concept implementation of GPU-based index build for the ScaNN index. The ScaNN index defined here is similar to IVF-PQ index  in structure (a tree structure coming from kmeans, plus product quantization of vectors assigned to leaf nodes), together with “AVQ update” of the kmeans centroids and a spilled cluster assignment from the “SOAR” loss.

Other features, optimizations, and customizability options to appear in subsequent PRs.

* scann_build.cuh 

This file contains the implementation for build(..). The general pipeline looks like:
Train kmeans centers on sampled data
Assign all dataset vectors to kmeans clusters by minimizing L2 loss
Update kmeans centers with AVQ
Train PQ codebook on sampled residual vectors (here we use VPQ, slightly modified to perform product quantization on individual subspaces, e.g. each subspace has its own codebook)
Quantization loop (batched):
Compute spilled SOAR labels (performed here to minimize HtoD copies)
Compute and quantize residuals/soar residuals using trained pq codebook
If enabled, compute bf16 quantization of dataset vectors (performed here to minimize HtoD copies).

* scann_avq.cuh

This file contains apply_avq(..), which recomputes cluster centers using AVQ. The main technique is a single application of Theorem 4.2 in https://arxiv.org/pdf/1908.10396 to each cluster, using parameters:
h_i_parallel = eta * || x_i || ^ (eta - 1) 
h_i_orthogonal = ||x _i || ^ (eta -1)
The implementation of Theorem 4.2 is in compute_avq_centroid(..)

The overall pipeline for apply_avq(..) is:
Build clusters from kmeans cluster assignments 
For each cluster:
Gather cluster vectors into single matrix
Update kmeans center via compute_avq_centroid
Rescale updated centroids (I need to add more details about this step).

* scann_quantize.cuh

This file contains helpers for PQ. Codebooks are created from residual vectors using train_pq from vpq_dataset.cuh (using a single vq center which is set to zero). Unlike in VPQ, codebooks are generated separately for each subspace, rather than collapsing all subspaces into a single space and computing a global codebook. 

* scann_soar.cuh

The main function is compute_soar_labels(..), which computes a second, spilled cluster assignment by minimizing the SOAR loss function (Theorem 3.1 in https://arxiv.org/pdf/2404.00774) 

* scann_serialize.cuh

Contains the implementation of serialize(..). The goal is to serialize ScaNN artifacts in a way that is usable with open-source ScaNN search with minimal additional post-processing. The cluster assignments, quantized vectors (for both the primary and spilled SOAR assignments), and bf16 dataset are all stored in separate .npy files for direct consumption by open-source ScaNN. The codebook and cluster centers are also serialized separately, but require additional post-processing into the correct Protobuf structs (not included in this PR). 

Test Plan:
This code is mostly tested via CPU search with open-source ScaNN. Additional protobuf artifacts are created from the cuVS serialized index via an external tool. A pareto for OpenAI 5M is shown here:


cpp/src/neighbors/detail/vpq_dataset.cuh
achirkin, line 1: Looks like some automatic refactoring happened on this file. Could you please undo this change?
Fiy we don't enforce `.clang-tidy` rules, but we try to stick to them, i.e. we should use camel case for the template arguments https://github.com/rapidsai/cuvs/blob/b230ebe73f3fe7430a9e56199e8ffb54d2325c05/cpp/.clang-tidy#L211C42-L212

cpp/include/cuvs/neighbors/scann.hpp
```diff
@@ -0,0 +1,310 @@
...
+
+  // Residual quanitzation params
+  uint32_t pq_dim  = 8;
+  uint32_t pq_bits = 8;
+  int64_t pq_n_rows_train;
+  uint32_t pq_train_iters = 10;
+
+  bool bf16_enabled = false;
+  // TODO - add other scann build params
+};
```
achirkin: We try to make all parameter structures in cuVS default-constructible. Please add defaults to all parameters to satisfy this requirement. If it's not possible, try to reformulate: e.g. kmeans_n_rows_train -> kmeans_trainset_fraction

cpp/src/neighbors/scann/detail/scann_avq.cuh
```diff
@@ -0,0 +1,443 @@
...
+                                      temp_storage_bytes,
+                                      clusters.data_handle(),
+                                      cluster_sizes.data_handle(),
+                                      num_levels,
+                                      lower_level,
+                                      upper_level,
+                                      clusters.extent(0),
+                                      stream);
+
+  raft::resource::sync_stream(dev_resources, stream);
```
achirkin: Could you please add a comment why we need a stream sync here?  I couldn't figure at first glance.

cpp/src/neighbors/scann/detail/scann_avq.cuh
```diff
@@ -0,0 +1,443 @@
...
+
+  RAFT_CUSOLVER_TRY(cusolverDnSpotrf(
+    cusolverH, uplo, n, A.data_handle(), lda, d_work.data(), lwork, devInfo.data_handle()));
+
+  // solve Ax = b
+  int ldb  = b.extent(0);
+  int nrhs = 1;
+
+  RAFT_CUSOLVER_TRY(cusolverDnSpotrs(
+    cusolverH, uplo, n, nrhs, A.data_handle(), lda, b.data_handle(), ldb, devInfo.data_handle()));
```
achirkin: Please use the raft wrappers from `raft::linalg::detail` at https://github.com/rapidsai/raft/blob/branch-25.08/cpp/include/raft/linalg/detail/cusolver_wrappers.hpp . They're not in public namespace, but at least they take on responsibility of managing the cusolver state for us. 
Or even better maybe something from raft::linalg would work in place of this whole function? If not, would it make sense to open a PR in raft to add it there instead?

cpp/src/neighbors/scann/detail/scann_avq.cuh
```diff
@@ -0,0 +1,443 @@
...
+                                num_items);
+}
+
+// Sum elements of device vector into device scalar
+template <typename T>
+void sum_reduce_vector(raft::resources const& dev_resources,
+                       raft::device_vector_view<T, int64_t> v,
+                       raft::device_scalar_view<T> s)
+{
+  NVTX3_FUNC_RANGE();
```
achirkin: Where this macro is defined? Please use `core/nvtx.hpp` instead, i.e. something like this:
```suggestion
  raft::common::nvtx::range<cuvs::common::nvtx::domain::cuvs> fun_scope("scann::sum_reduce_vector(...)", ...);
```

cpp/src/neighbors/scann/detail/scann_avq.cuh
```diff
@@ -0,0 +1,443 @@
...
+                              x.extent(1),
+                              x.extent(0),
+                              eta_m_1.data_handle(),
+                              x_trans.view().data_handle(),
+                              x.extent(0),
+                              x.data_handle(),
+                              x.extent(1),
+                              sum_norms_eta_1.data_handle(),
+                              x_trans_x.view().data_handle(),
+                              x.extent(1)));
```
achirkin, line 296: Please use `raft::linalg::gemm`

cpp/src/neighbors/scann/detail/scann_build.cuh
```diff
@@ -0,0 +1,375 @@
...
+  // sample/project
+
+  cuvs::cluster::kmeans::balanced_params kmeans_params;
+  kmeans_params.n_iters = params.kmeans_n_iters;
+
+  {
+    auto trainset =
+      raft::make_device_matrix<T, int64_t>(res, params.kmeans_n_rows_train, dataset.extent(1));
+
+    printf("Sampling rows.\n");
```
achirkin: Please search/replace printfs with RAFT_LOG_DEBUG if appropriate or remove them completely

achirkin (changes requested):
Thanks for the PR! Overall the code looks rather clean, and I like the extensive usage of raft primitives. I'm not familiar with the algorithm, but a here are few cosmetic and technical suggestions.

cpp/include/cuvs/neighbors/scann.hpp
```diff
@@ -0,0 +1,310 @@
...
+ * @{
+ */
+
+/**
+ * @brief ANN parameters used by ScaNN to build index
+ *
+ */
+struct index_params : cuvs::neighbors::index_params {
+  // partitioning
+  uint32_t scann_clusters;
```
tfeher: Is this the number of clusters? Could you add docstring?

cpp/include/cuvs/neighbors/scann.hpp
```diff
@@ -0,0 +1,310 @@
...
+  int64_t kmeans_n_rows_train;
+  uint32_t kmeans_n_iters = 24;
+  float partitioning_eta  = 1;
+  float soar_lambda;
+
+  // Residual quanitzation params
+  uint32_t pq_dim  = 8;
```
tfeher: Does `dim` need to be divisible by `pq_dim`?

cpp/include/cuvs/neighbors/scann.hpp
```diff
@@ -0,0 +1,310 @@
...
+  index(raft::resources const& res, const index_params& params, IdxT n_rows, IdxT dim)
+    : index(res,
+            params.metric,
+            params.scann_clusters,
+            params.pq_bits,
+            params.pq_dim,
+            n_rows,
+            dim,
+            1 << params.pq_bits,
+            dim / params.pq_dim,
```
tfeher, line 171: If user sets `pq_dim > dim`, that would zero `pq_num_subspaces`. I assume that is not valid. Please add checks (see usage of `RAFT_EXPECTS`).

cpp/include/cuvs/neighbors/scann.hpp
```diff
@@ -0,0 +1,310 @@
...
+
+/**
+ * @defgroup scann_cpp_index_build ScaNN index build functions
+ * @{
+ */
+/**
+ * @brief Build the index from the dataset for efficient search.
+ *
+ */
+auto build(raft::resources const& handle,
```
tfeher, line 278: I see that the supported dataset types are `float`, `int8_t` and `uint8_t`. This nicely aligns with the rest of the vector search algorithms. 

But due to growing binary size, we are considering to keep only one of the int8 types. That brings up the questions:
- What data types does the original SCANN implementation supports?
- What is the effect on the binary size (@tfeher to answer this)

tfeher (commented):
Hi Robert, thank you for this work! I have started my review, here are my first batch of comments.

cpp/src/neighbors/scann/detail/scann_avq.cuh
rmaschal: Idk either, removed.

cpp/src/neighbors/scann/detail/scann_avq.cuh
rmaschal: I'll use the wrappers in raft::linalg::detail. Anything in raft/linalg/lstsq.cuh will work, but is slower/less accurate for this particular case (symmetric, pos-def matrix A). This whole function can move to lstsq.cuh. I can leave that as a follow-up, and remove this later.

cpp/src/neighbors/scann/detail/scann_avq.cuh
rmaschal: They're defined in  nvtx3/nvtx3.hpp. I'll just remove them for now.

cpp/src/neighbors/scann/detail/scann_avq.cuh
rmaschal: I suggest we make this a follow-up/todo. Both gemm apis in gemm.hpp and gemm.cuh result in a segfault for me. I don't think either api is handling cudaPointerMode correctly (both always using host mode). My alpha/beta are computed on-device, so it's preferred that I use device mode. 

I'll file an issue on raft and clean-up here once fixed.

achirkin: Are you sure you pass device_scalar_view and not the device_scalar to the gemm in gemm.hpp? Weirdly enough, looks like it accepts both, but will only catch up the residency correctly for the view https://github.com/rapidsai/raft/blob/5bc46992c571de11421d04e55f416bfb272de831/cpp/include/raft/linalg/gemm.hpp#L82-L83C36
(It's definitely a bug though, we should be checking the accessor of there rather than the specific view type)

rmaschal: Yes. I think the bug is deeper: https://github.com/rapidsai/raft/blob/branch-25.08/cpp/include/raft/linalg/detail/cublaslt_wrappers.hpp#L304-L320

Independent of how kDeviceMode is computed, it's not passed forward to legacy_matmul, in which device mode defaults to false: https://github.com/rapidsai/raft/blob/branch-25.08/cpp/include/raft/linalg/detail/cublaslt_wrappers.hpp#L234-L235

gemm(..) in gemm.cuh has the same issue

cpp/include/cuvs/neighbors/scann.hpp
rmaschal: Actually we only support float atm. The int8/uint8 apis here were added early, but only float is implemented via scann_build/serialize_float.cu. I can remove them from this PR, and add additional dtypes in subsequent PRs. 

At least we need float/uint8 (e.g. for BigAnn-1b) in the short term, but I'm not sure what's the full list. I can follow-up.

cpp/src/neighbors/scann/detail/scann_common.cuh
```diff
@@ -0,0 +1,87 @@
...
+      cluster_vecs.data_handle(), pinned_cluster.data_handle(), pinned_cluster.size(), stream);
+    raft::resource::sync_stream(res, stream);
+  }
+};
+
+template <typename T,
+          typename IdxT     = int64_t,
+          typename Accessor = raft::host_device_accessor<std::experimental::default_accessor<T>,
+                                                         raft::memory_type::host>>
+void sample_rows(
```
tfeher: This is very similar to [raft::matrix::sample_rows](https://github.com/rapidsai/raft/blob/5bc46992c571de11421d04e55f416bfb272de831/cpp/include/raft/matrix/sample_rows.cuh#L39), it would be good to reuse.

I see that here we have a simpler random sampling of indices. Is that a relevant change (e.g. saving memory by omitting sorting steps)? If yes, then we should add this variant as an option to raft. (Since raft is in code freeze, this would be a task for next release).

examples/cpp/src/scann_example.cu
```diff
@@ -0,0 +1,85 @@
...
+  auto index = scann::build(dev_resources, index_params, dataset);
+  auto end = std::chrono::system_clock::now();
+  std::chrono::duration<double> elapsed_seconds = end - start;
+
+  // TODO - output statistics about the index
+
+  std::cout << "Time to build index: " << elapsed_seconds.count() << "s\n";
+
+  // Output index to file
+//  serialize(dev_resources, "test_out", index); // Not yet implemented
```
tfeher: Is it still `Not yet implemented`? I see a `serialize()` function.

cpp/src/neighbors/scann/detail/scann_serialize.cuh
```diff
@@ -0,0 +1,150 @@
...
+
+  auto labels_of = open_file(labels_path);
+
+  raft::serialize_mdspan(res, labels_of, combined_labels.view());
+
+  labels_of.close();
+}
+
+/**
+ * Save the ScaNN index into multiple files. The format is meant
```
tfeher: Maybe move "The format is meant" from the brief description to the detailed description, and also finish the sentence

cpp/include/cuvs/neighbors/scann.hpp
```diff
@@ -0,0 +1,310 @@
...
+ */
+
+/**
+ * @brief ANN parameters used by ScaNN to build index
+ *
+ */
+struct index_params : cuvs::neighbors::index_params {
+  // partitioning
+  uint32_t scann_clusters;
+  int64_t kmeans_n_rows_train;
```
tfeher: Is this the absolute number of training rows? Is it what the reference implementation also does? Asking this because in other IVF methods we found that keeping the training rows proportional to number of lists has its benefits too.

cpp/src/neighbors/scann/detail/scann_build.cuh
```diff
@@ -0,0 +1,375 @@
...
+  index<T, IdxT> idx(res, params, dataset.extent(0), dataset.extent(1));
+  raft::device_matrix_view<T, int64_t> centroids_view = idx.centers();
+  raft::random::RngState random_state{137};
+  // sample/project
+
+  cuvs::cluster::kmeans::balanced_params kmeans_params;
+  kmeans_params.n_iters = params.kmeans_n_iters;
+
+  {
+    auto trainset =
```
tfeher: How large this matrix typically is? If there is a chance to go out of memory, then it would be great throw a descriptive error message, e.g.

https://github.com/rapidsai/cuvs/blob/ed32f0fb6689de6a346f7d551662d58e2965588e/cpp/src/neighbors/ivf_pq/ivf_pq_build.cuh#L1738-L1741

tfeher (changes requested):
@rmaschal could you please add some tests? I understand that without a search method it is tricky to test, but we should have at least some smoke test, like testing the API for different dataset sizes. These should also include some edge cases.

Below you will find my second batch of comments.

cpp/src/neighbors/scann/detail/scann_common.cuh
rmaschal: I don't recall raft::matrix::sample_rows having a sorting step last I saw, that could be useful. The important difference is I use my own gather instead of raft::matrix::gather. At least for the sample sizes I experimented with, the allocation of pinned buffers in raft::matrix::gather took most of the time, and ran slower overall. But I think this warrants some further study.

cpp/include/cuvs/neighbors/scann.hpp
tfeher: Do I see it correctly that `pq_dim` is the number of dimensions in a subspace? Our IVF-PQ and CAGRA implementations uses `pq_dim` as the number of PQ subspaces, and we use internally `pq_len` as the number of dimensions in a subspace. Having different meaning of `pq_dim` across algorithms is confusing, can we align with IVF-PQ? Or would that contradict SCANN's naming convention?

cpp/include/cuvs/neighbors/scann.hpp
rmaschal: Yes, absolute. This mirrors the configuration of OSS ScaNN. 

For IVF methods, did you find a particular ratio of samples/n_lists to work well? I've noticed larger ratios tends to give poorer recall in this algorithm, which I'm guessing is due to balanced kmeans (in a later PR i'll add support for unbalanced kmeans to further investigate)

cpp/include/cuvs/neighbors/scann.hpp
rmaschal: pq_dim is the dimension of each subspace. This is the convention of OSS ScaNN. I agree it's confusing, but I'm not sure who should bear the confusion :)

cpp/src/neighbors/scann/detail/scann_quantize.cuh
```diff
@@ -0,0 +1,262 @@
...
+    raft::make_device_matrix<T, uint32_t, raft::row_major>(res, 1, residuals.extent(1));
+  raft::linalg::map_offset(res, vq_code_book.view(), [] __device__(size_t i) { return 0; });
+
+  auto pq_code_book = cuvs::neighbors::detail::train_pq<T>(
+    res, ps, residuals, raft::make_const_mdspan(vq_code_book.view()));
+
+  return pq_code_book;
+}
+
+template <typename T, typename LabelT>
```
tfeher, line 160: Although the code is easy to read, it would still be useful to add a docstring that defines the expectations on input dimensions.

```suggestion
/**
 * @brief Subtract cluster center coordinates from each dataset vector.
 * 
 * residual[i, k] = dataset[i ,k] - centers[l, k], 
 * where l = labels[i], the cluster label corresponding to vector i.
 * 
 * @tparam T 
 * @tparam LabelT 
 * @param res raft resources
 * @param dataset dataset vectors, size [n_rows, dim]
 * @param centers cluster center coordinates, size [n_clusters, dim]
 * @param labels cluster labels, size [n_rows]
 * @return device matrix with the residuals, size [n_rows, dim]
 */
template <typename T, typename LabelT>
```

cpp/src/neighbors/scann/detail/scann_build.cuh
```diff
@@ -0,0 +1,375 @@
...
+                                                      dataset,
+                                                      raft::make_const_mdspan(centroids_view),
+                                                      raft::make_const_mdspan(labels_view),
+                                                      pq_n_rows_train);
+
+  int num_subspaces    = dataset.extent(1) / params.pq_dim;
+  int dim_per_subspace = params.pq_dim;
+  int num_clusters     = 1 << params.pq_bits;
+
+  auto temp_codebook = raft::make_device_matrix<T, uint32_t>(res, 1, num_clusters);
```
tfeher: This is not used, please remove.

cpp/include/cuvs/neighbors/scann.hpp
```diff
@@ -0,0 +1,303 @@
...
+  float partitioning_eta = 1;
+
+  /** the value of lambda for SOAR spilling **/
+  float soar_lambda = 1;
+
+  // Residual quanitzation params
+  /** the dimension of pq subspaces (must divide dataset dimension)**/
+  uint32_t pq_dim = 8;
+
+  /** the number of bits for pq codes (but be 4 or 8, for 16 and 256 codes respectively) **/
```
tfeher:
```suggestion
  /** the number of bits for pq codes (must be 4 or 8, for 16 and 256 codes respectively) **/
```

cpp/src/neighbors/scann/detail/scann_quantize.cuh
```diff
@@ -0,0 +1,262 @@
...
+template <typename T, typename IdxT, typename LabelT>
+auto quantize_residuals(raft::resources const& res,
+                        raft::device_matrix_view<const T, int64_t> residuals,
+                        raft::device_matrix_view<T, uint32_t, raft::row_major> pq_codebook,
+                        cuvs::neighbors::vpq_params ps)
+  -> raft::device_matrix<uint8_t, IdxT, raft::row_major>
+{
+  auto dim         = residuals.extent(1);
+  auto vq_codebook = raft::make_device_matrix<T, uint32_t, raft::row_major>(res, 1, dim);
+  raft::linalg::map_offset(res, vq_codebook.view(), [] __device__(size_t i) { return 0; });
```
tfeher: I think `memset` would be preferred, do we have any style check that stops us from using it?
```suggestion
RAFT_CUDA_TRY(cudaMemsetAsync(vq_codebook.data_handle(), 0, vq_codebook.size() * sizeof(T), resource::get_cuda_stream(res)));
```
If yes, then we could still simplify this:
```suggestion
  // #include <raft/core/operators.hpp>
  raft::linalg::map_offset(res, vq_codebook.view(), raft::const_op<T>{0});
```

cpp/src/neighbors/scann/detail/scann_build.cuh
```diff
@@ -0,0 +1,383 @@
...
+  //	auto full_codebook = raft::make_device_matrix<T,uint32_t>(res, num_clusters,
+  // num_subspaces*dim_per_subspace);
+  auto full_codebook =
+    raft::make_device_matrix<T, uint32_t>(res, num_clusters * num_subspaces, dim_per_subspace);
+
+  auto temp_avq_quant =
```
tfeher: This also does not seem to be used.

cpp/src/neighbors/scann/detail/scann_build.cuh
```diff
@@ -0,0 +1,383 @@
...
+                                                      pq_n_rows_train);
+
+  int num_subspaces    = dataset.extent(1) / params.pq_dim;
+  int dim_per_subspace = params.pq_dim;
+  int num_clusters     = 1 << params.pq_bits;
+
+  auto temp_codebook = raft::make_device_matrix<T, uint32_t>(res, 1, num_clusters);
+  //	auto full_codebook = raft::make_device_matrix<T,uint32_t>(res, num_clusters,
+  // num_subspaces*dim_per_subspace);
+  auto full_codebook =
```
tfeher, line 204: I find that having a having a conceptually 3D array allocated as a 2D matrix is unfortunate. It makes offset calculations more tedious than it should be. E,g if you have
```c++
auto full_codebook = raft::make_device_mdarray<T>(res, raft::make_extents<uint32_t>(num_clusters, num_subspaces, dim_per_subspace));
```
then we can rely  on mdspan addressing operator to calculate offsets, e.g `full_codebook(i, 0, 0)` (and `&full_codebook(i, 0, 0)` also works).

I understand that this would lead to a lot of changes in the code, and I do not insist doing this now, but something to keep in mind for the future.

cpp/src/neighbors/scann/detail/scann_build.cuh
```diff
@@ -0,0 +1,383 @@
...
+    // For VPQ, pq_dim is the number of subspaces, not the dimension of the subspaces
+    pq_params.pq_dim = 1;
+    // We handle sampling/training set construction above, so use the full set in VPQ
+    pq_params.pq_kmeans_trainset_fraction = 1.0;
+    pq_params.kmeans_n_iters              = params.pq_train_iters;
+
+    // Create pq codebook for this subspace
+    auto sub_pq_codebook =
+      create_pq_codebook<T>(res, raft::make_const_mdspan(sub_trainset.view()), pq_params);
+    raft::copy(full_codebook.data_handle() + (subspace * sub_pq_codebook.size()),
```
tfeher, line 232: The mdspan addressing operator is often useful for calculating offsets 
```suggestion
    raft::copy(&full_codebook(subspace * n_clusters, 0),
```

cpp/src/neighbors/scann/detail/scann_soar.cuh
```diff
@@ -0,0 +1,172 @@
...
+  auto dim = dataset.extent(1);
+
+  raft::linalg::map_offset(
+    dev_resources, residuals.view(), [dataset, centers, labels, dim] __device__(size_t i) {
+      int row_idx  = i / dim;
+      int el_idx   = i % dim;
+      LabelT label = labels(row_idx);
+
+      return dataset(row_idx, el_idx) - centers(label, el_idx);
+    });
```
tfeher: Could we use `compute_residuals` from `scann_quantize.cuh`?

Alternatively: aren't these residuals calculated at the call site in the same loop? Would it be possible to pass those as argument?

cpp/src/neighbors/scann/detail/scann_build.cuh
```diff
@@ -0,0 +1,383 @@
...
+
+    if (params.bf16_enabled) {
+      raft::linalg::map_offset(res, bf16_dataset.view(), [batch_view, dim] __device__(size_t i) {
+        int64_t row_idx = i / dim;
+        int64_t col_idx = i % dim;
+
+        nv_bfloat16 val = __float2bfloat16(batch_view(row_idx, col_idx));
+
+        return reinterpret_cast<int16_t&>(val);
+      });
```
tfeher: Since this does the same op for each element, can't we just run it in 1D?
```suggestion
      raft::linalg::map(
          res, 
          make_device_vector_view<int16_t, int64_t>(bf16_dataset.data_handle(), bf16.size()),
          make_device_vector_view<T, int64_t>(batch_view.data_handle(), batch_view.size(),
          [] __device__(T x) {
              nv_bfloat16 val = __float2bfloat16(x);
              return reinterpret_cast<int16_t&>(val);
          });
```
After writing this I feel that our effort to express everything as mdspan does lead to less readable code in this particular case, and I would propose to call an old version of `unaryOp` that still works with pointers (and calls map under the hood):

```suggestion
      raft::linalg::unaryOp(bf16_dataset.data_handle(), batch_view.data_handle(), batch_view.size(), 
      __device__(T x) {
        nv_bfloat16 val = __float2bfloat16(x);
        return reinterpret_cast<int16_t&>(val);
      }, resource::get_cuda_stream(res));
```

cpp/src/neighbors/scann/detail/scann_build.cuh
```diff
@@ -0,0 +1,383 @@
...
+      batch.size(),
+      num_subspaces);
+
+    raft::copy(h_quantized_residuals_view.data_handle(),
+               quantized_residuals.data_handle(),
+               quantized_residuals.size(),
+               stream);
+
+    auto h_quantized_soar_residuals_view = raft::make_host_matrix_view<uint8_t, uint32_t>(
+      idx.quantized_soar_residuals().data_handle() + batch.offset() * num_subspaces,
```
tfeher: We could simplify like this. But do we even need this view? We could just put the same line into `raft::copy`
```suggestion
      &(idx.quantized_soar_residuals())(batch.offset(), 0),
```

cpp/src/neighbors/scann/detail/scann_build.cuh
```diff
@@ -0,0 +1,383 @@
...
+               quantized_soar_residuals.data_handle(),
+               quantized_soar_residuals.size(),
+               stream);
+
+    if (params.bf16_enabled) {
+      auto h_bf16_dataset_view = raft::make_host_matrix_view<int16_t, int64_t>(
+        idx.bf16_dataset().data_handle() + batch.offset() * dim, batch.size(), dim);
+
+      raft::copy(
+        h_bf16_dataset_view.data_handle(), bf16_dataset.data_handle(), bf16_dataset.size(), stream);
```
tfeher: would this work?
```suggestion
      raft::copy(
        &(idx.bf16_dataset())(batch.offset(), 0), bf16_dataset.data_handle(), bf16_dataset.size(), stream);
```

tfeher (changes requested):
Hi Robert, I went through a large part of the implementation (some details in soar and avq left). Overall the code looks great: clean, well structured, easy to read, and I feel good about getting this in for the release.

One problem is the duplication of building blocks from IVF-PQ. Since it is outside the scope of the PR to simplify our tooling for  IVF-PQ, SCANN and CAGRA-Q, I have created an issue to track those things: https://github.com/rapidsai/cuvs/issues/1182.

Please find below my next batch of comments.

cpp/include/cuvs/neighbors/scann.hpp
tfeher: @cjnolet what is your opinion?

cpp/src/neighbors/scann/detail/scann_common.cuh
tfeher: I have added it to a tracker issue https://github.com/rapidsai/cuvs/issues/1182, we can return to this in the next release.

cpp/include/cuvs/neighbors/scann.hpp
tfeher: [FAISS has 256 as default](https://github.com/facebookresearch/faiss/wiki/How-to-make-Faiss-run-faster#k-means-clustering).  See also [this link](https://github.com/facebookresearch/faiss/wiki/FAQ#how-many-training-points-do-i-need-for-k-means). In cuVS we often use 10% of the dataset, but thinking about to change that: https://github.com/rapidsai/cuvs/pull/1125#discussion_r2223959821

cpp/src/neighbors/scann/detail/scann_soar.cuh
```diff
@@ -0,0 +1,172 @@
...
+  int64_t n_rows = dataset.extent(0);
+
+  auto cublas_handle = raft::resource::get_cublas_handle(dev_resources);
+
+  auto alpha = raft::make_host_scalar<float>(1.0);
+  auto beta  = raft::make_host_scalar<float>(0.0);
+
+  auto dataset_dot_residual = raft::make_device_vector<float, int64_t>(dev_resources, n_rows);
+
+  cublasSgemvStridedBatched(cublas_handle,
```
tfeher:
```suggestion
  RAFT_CUBLAS_TRY(cublasSgemvStridedBatched(cublas_handle,
```

cpp/src/neighbors/scann/detail/scann_avq.cuh
```diff
@@ -0,0 +1,445 @@
...
+    dev_resources, x.extent(1), x.extent(1));
+
+  raft::matrix::eye(dev_resources, x_trans_x.view());
+
+  auto eta_m_1       = raft::make_device_scalar<float>(dev_resources, eta - 1);
+  auto cublas_handle = raft::resource::get_cublas_handle(dev_resources);
+
+  cublasPointerMode_t pm;
+  cublasGetPointerMode(cublas_handle, &pm);
+  cublasSetPointerMode(cublas_handle, CUBLAS_POINTER_MODE_DEVICE);
```
tfeher: We have a [helper class](https://github.com/rapidsai/raft/blob/18ac77dfdf03957a72677568c06ce34df909cf76/cpp/include/raft/linalg/detail/cublas_wrappers.hpp#L35) to manage the lifetime of this. Unfortunately it is in the `detail` namespace, but I think we could still use it. I have this to tracker #1182.
```suggestion
 raft::linalg::detail::cublas_device_pointer_mode pm;
```

cpp/src/neighbors/scann/detail/scann_quantize.cuh
```diff
@@ -0,0 +1,262 @@
...
+      int row_idx = i / dim;
+      int el_idx  = i % dim;
+      return dataset(row_idx, el_idx) - centers(labels(row_idx), el_idx);
+    });
+
+  return residuals;
+}
+
+template <typename T, typename IdxT, typename LabelT>
+auto quantize_residuals(raft::resources const& res,
```
tfeher, line 198: Could you add a docstring that mention the sizes of the input and return matrices? I find documenting the shape of objects that we are operating with helps a lot understanding the code.

cpp/src/neighbors/scann/detail/scann_avq.cuh
```diff
@@ -0,0 +1,445 @@
...
+
+  dim3 block(32, 1, 1);
+  dim3 grid((dataset.extent(0) + block.x - 1) / block.x, 1, 1);
+
+  build_clusters<uint32_t, uint32_t><<<grid, block>>>(labels_view.data_handle(),
+                                                      clusters.view().data_handle(),
+                                                      cluster_ptrs.view().data_handle(),
+                                                      dataset.extent(0),
+                                                      labels_view.extent(0));
+
```
tfeher, line 422:
```suggestion
  RAFT_CUDA_TRY(cudaPeekAtLastError());
```

tfeher (changes requested):
I have finished reviewing the code. It looks good overall, here are my latest comments.

The biggest issue is the lack of test. At the minimum we could test that the output arrays have expected shape.

cpp/include/cuvs/neighbors/scann.hpp
```diff
@@ -0,0 +1,303 @@
...
+  -> cuvs::neighbors::experimental::scann::index<float, int64_t>;
+/**
+ * @defgroup scann_cpp_serialize ScaNN serialize functions
+ * @{
+ */
+/**
+ * Save the index to file.
+ */
+
+void serialize(raft::resources const& handle,
```
cjnolet, line 303: This implementation is a lot of code to check into a public library like this without any tests. We need to find a way to test this before the experimental designator is removed. We should also have a caveat in the doxygen for the scann index that automated tests are pending, so results may vary. I want to make sure we convey to users that this is early code and that it's possible accuracy and performance could diverge (without warning).

cpp/src/neighbors/scann/detail/scann_common.cuh
rmaschal: I tried out the raft solution on some large sample sizes, and found the raft solution to be faster starting around 1m rows. I'm going to remove my custom solution and use the raft version, and we can look for further improvements later as part of the tracker.

cpp/src/neighbors/scann/detail/scann_build.cuh
rmaschal: Is the suggestion that we should use a 3d array for the full codebook? We went with the 2d structure here given that's ultimately how OSS ScaNN represents the codebook (although it is stored in a protobuf struct). Using the 2d structure made it easier to align with the OSS for search, and made a few things easier to test/debug when evaluating recall w/ OSS search.

FWIW we did use a 3d array in early iterations. We were trying to use IVF-PQ for the PQ step (since per subspace PQ is naturally supported), which uses the 3d array for codebook storage. We were never able to successfully unpack codes for use in OSS search (though this was probably due to the interleaving and possibly other factors, rather than the 3d structure).

cpp/src/neighbors/scann/detail/scann_build.cuh
rmaschal: I'll think about this, I got a compile error when I tried (taking the address of temporary full_codebook(..) )

cpp/src/neighbors/scann/detail/scann_build.cuh
rmaschal: I don't think so. Taking the address of a temporary, as above. But I removed the additonal "view" per the previous comments for the quantized residuals.

cpp/include/cuvs/neighbors/scann.hpp
tfeher: Sounds good. The binary grows with the following objects:
```
scann_build_float.cu.o		7.917 MB
scann_serialize_float.cu.o	1.877 MB
```

cpp/src/neighbors/scann/detail/scann_build.cuh
tfeher: It makes sense to align with OSS ScaNN.

cpp/src/neighbors/scann/detail/scann_build.cuh
tfeher: Normally this should work (https://godbolt.org/z/cz83vxr1h), because the return value is a reference. ~~We do use it in some places in the code, but I have hard time finding it now.~~ Let's not hold up the PR for this.

tfeher: You are right, indeed this triggers a compiler error in rapids.

rmaschal:
@tfeher I did add some basic tests last night, I'm not sure if you caught that. They of course don't measure anything related to the quality of the index, but it at least checks dimensions and otherwise ensure build succeeds.

tfeher (approved):
Thanks @rmaschal for adding the tests, the PR looks good to me.

achirkin:
Hi @rmaschal , could you please revert the vpq_dataset file? https://github.com/rapidsai/cuvs/pull/1120#discussion_r2226096396

rmaschal:
@achirkin Oh I could have sworn I did that. New commit reverting those changes.

PR #1123, opened 2025-07-16
Add `build_from_args` for IVF-PQ

### LINKED ISSUE #1107: [FEA] Allow IVFPQ index object to be constructed with coarse centroids and codebooks
We have been getting many requests from users to be able to use IVFPQ's encoding of vectors, but to enable CPU search. Of course, the first recommendation we make for this is to use Faiss for CPU/GPU interoperability but these users want to rely only on cuVS and their own CPU implementations. 

We have been discussing a first-class PQ encoder API for some time, but it becomes challenging to be flexible. Many of these users also train the centroids and codebooks once and want to use them across many different shards of IVFPQ indexes. While we still plan to host such an API eventually, an intermediate solution would look something like this:

1. User trains coarse centroids and codebooks using cuVS
2. User extracts coarse centroids and codebooks from IVFPQ index
3. User construct new IVFPQ instance from existing coarse centroids and codebooks
4. User calls extend() on IVFPQ instance, extracts the vectors from the lists and copies them into their CPU indexes. 
5. User uses their CPU version for search

This ultimately comes down to our exposing a constructor for the IVFPQ index that accepts the coarse centroids and codebooks. But we don't want to expose these constructors directly, of course, so we'll need to also expose a new factory function overload for `build()` in the public APIs that accepts the centroids and codebooks and returns the built IVFPQ index.

lowener:
In this PR, a new way to build IVF-PQ is added: `build_from_args` in Python that use a new overload of `build` in C++.
This new build method takes views on device memory to construct an IVF-PQ index that is extendable and searchable with as few data transfer as possible.
It is made available in C++ and Python, and will be re-used in a future preprocessing function for product quantization.

Closes #1107.

### DIFF
--- a/cpp/include/cuvs/neighbors/ivf_pq.h
+++ b/cpp/include/cuvs/neighbors/ivf_pq.h
@@ -262,19 +262,19 @@ cuvsError_t cuvsIvfPqIndexCreate(cuvsIvfPqIndex_t* index);
 cuvsError_t cuvsIvfPqIndexDestroy(cuvsIvfPqIndex_t index);
 
 /** Get the number of clusters/inverted lists */
-cuvsError_t cuvsIvfPqIndexGetNLists(cuvsIvfPqIndex_t index, uint32_t* n_lists);
+cuvsError_t cuvsIvfPqIndexGetNLists(cuvsIvfPqIndex_t index, int64_t* n_lists);
 
 /** Get the dimensionality */
-cuvsError_t cuvsIvfPqIndexGetDim(cuvsIvfPqIndex_t index, uint32_t* dim);
+cuvsError_t cuvsIvfPqIndexGetDim(cuvsIvfPqIndex_t index, int64_t* dim);
 
 /** Get the pq_dim */
-cuvsError_t cuvsIvfPqIndexGetPqDim(cuvsIvfPqIndex_t index, uint32_t* pq_dim);
+cuvsError_t cuvsIvfPqIndexGetPqDim(cuvsIvfPqIndex_t index, int64_t* pq_dim);
 
 /** Get the pq_len */
-cuvsError_t cuvsIvfPqIndexGetPqLen(cuvsIvfPqIndex_t index, uint32_t* pq_len);
+cuvsError_t cuvsIvfPqIndexGetPqLen(cuvsIvfPqIndex_t index, int64_t* pq_len);
 
 /** Get the pq_book_size */
-cuvsError_t cuvsIvfPqIndexGetPqBookSize(cuvsIvfPqIndex_t index, uint32_t* pq_book_size);
+cuvsError_t cuvsIvfPqIndexGetPqBookSize(cuvsIvfPqIndex_t index, int64_t* pq_book_size);
 
 /**
  * @brief Get the cluster centers corresponding to the lists in the original space
--- a/python/cuvs/cuvs/neighbors/ivf_pq/ivf_pq.pyx
+++ b/python/cuvs/cuvs/neighbors/ivf_pq/ivf_pq.pyx
@@ -247,14 +247,14 @@ cdef class Index:
     @property
     def n_lists(self):
         """ The number of inverted lists (clusters) """
-        cdef uint32_t n_lists
+        cdef int64_t n_lists
         check_cuvs(cuvsIvfPqIndexGetNLists(self.index, &n_lists))
         return n_lists
 
     @property
     def dim(self):
         """ dimensionality of the cluster centers """
-        cdef uint32_t dim
+        cdef int64_t dim
         check_cuvs(cuvsIvfPqIndexGetDim(self.index, &dim))
         return dim
 
@@ -311,21 +311,21 @@ cdef class Index:
     @property
     def pq_dim(self):
         """ Get the pq_dim """
-        cdef uint32_t pq_dim
+        cdef int64_t pq_dim
         check_cuvs(cuvsIvfPqIndexGetPqDim(self.index, &pq_dim))
         return pq_dim
 
     @property
     def pq_len(self):
         """ Get the pq_len """
-        cdef uint32_t pq_len
+        cdef int64_t pq_len
         check_cuvs(cuvsIvfPqIndexGetPqLen(self.index, &pq_len))
         return pq_len
 
     @property
     def pq_book_size(self):
         """ Get the pq_book_size """
-        cdef uint32_t pq_book_size
+        cdef int64_t pq_book_size
         check_cuvs(cuvsIvfPqIndexGetPqBookSize(self.index, &pq_book_size))
         return pq_book_size
 
--- a/cpp/src/neighbors/ivf_pq_c.cpp
+++ b/cpp/src/neighbors/ivf_pq_c.cpp
@@ -393,7 +393,7 @@ extern "C" cuvsError_t cuvsIvfPqExtend(cuvsResources_t res,
   });
 }
 
-extern "C" cuvsError_t cuvsIvfPqIndexGetNLists(cuvsIvfPqIndex_t index, uint32_t* n_lists)
+extern "C" cuvsError_t cuvsIvfPqIndexGetNLists(cuvsIvfPqIndex_t index, int64_t* n_lists)
 {
   return cuvs::core::translate_exceptions([=] {
     auto index_ptr = reinterpret_cast<cuvs::neighbors::ivf_pq::index<int64_t>*>(index->addr);
@@ -401,7 +401,7 @@ extern "C" cuvsError_t cuvsIvfPqIndexGetNLists(cuvsIvfPqIndex_t index, uint32_t*
   });
 }
 
-extern "C" cuvsError_t cuvsIvfPqIndexGetDim(cuvsIvfPqIndex_t index, uint32_t* dim)
+extern "C" cuvsError_t cuvsIvfPqIndexGetDim(cuvsIvfPqIndex_t index, int64_t* dim)
 {
   return cuvs::core::translate_exceptions([=] {
     auto index_ptr = reinterpret_cast<cuvs::neighbors::ivf_pq::index<int64_t>*>(index->addr);
@@ -443,7 +443,7 @@ extern "C" cuvsError_t cuvsIvfPqIndexGetCentersRot(cuvsIvfPqIndex_t index,
   });
 }
 
-extern "C" cuvsError_t cuvsIvfPqIndexGetPqDim(cuvsIvfPqIndex_t index, uint32_t* pq_dim)
+extern "C" cuvsError_t cuvsIvfPqIndexGetPqDim(cuvsIvfPqIndex_t index, int64_t* pq_dim)
 {
   return cuvs::core::translate_exceptions([=] {
     auto index_ptr = reinterpret_cast<cuvs::neighbors::ivf_pq::index<int64_t>*>(index->addr);
@@ -451,7 +451,7 @@ extern "C" cuvsError_t cuvsIvfPqIndexGetPqDim(cuvsIvfPqIndex_t index, uint32_t* 
   });
 }
 
-extern "C" cuvsError_t cuvsIvfPqIndexGetPqLen(cuvsIvfPqIndex_t index, uint32_t* pq_len)
+extern "C" cuvsError_t cuvsIvfPqIndexGetPqLen(cuvsIvfPqIndex_t index, int64_t* pq_len)
 {
   return cuvs::core::translate_exceptions([=] {
     auto index_ptr = reinterpret_cast<cuvs::neighbors::ivf_pq::index<int64_t>*>(index->addr);
@@ -459,7 +459,7 @@ extern "C" cuvsError_t cuvsIvfPqIndexGetPqLen(cuvsIvfPqIndex_t index, uint32_t* 
   });
 }
 
-extern "C" cuvsError_t cuvsIvfPqIndexGetPqBookSize(cuvsIvfPqIndex_t index, uint32_t* pq_book_size)
+extern "C" cuvsError_t cuvsIvfPqIndexGetPqBookSize(cuvsIvfPqIndex_t index, int64_t* pq_book_size)
 {
   return cuvs::core::translate_exceptions([=] {
     auto index_ptr = reinterpret_cast<cuvs::neighbors::ivf_pq::index<int64_t>*>(index->addr);
--- a/python/cuvs/cuvs/neighbors/ivf_pq/ivf_pq.pxd
+++ b/python/cuvs/cuvs/neighbors/ivf_pq/ivf_pq.pxd
@@ -15,7 +15,7 @@
 #
 # cython: language_level=3
 
-from libc.stdint cimport uint32_t, uintptr_t
+from libc.stdint cimport int64_t, uint32_t, uintptr_t
 from libcpp cimport bool
 
 from cuvs.common.c_api cimport cuvsError_t, cuvsResources_t
@@ -85,19 +85,19 @@ cdef extern from "cuvs/neighbors/ivf_pq.h" nogil:
     cuvsError_t cuvsIvfPqIndexDestroy(cuvsIvfPqIndex_t index)
 
     cuvsError_t cuvsIvfPqIndexGetNLists(cuvsIvfPqIndex_t index,
-                                        uint32_t* n_lists)
+                                        int64_t* n_lists)
 
     cuvsError_t cuvsIvfPqIndexGetDim(cuvsIvfPqIndex_t index,
-                                     uint32_t* dim)
+                                     int64_t* dim)
 
     cuvsError_t cuvsIvfPqIndexGetPqDim(cuvsIvfPqIndex_t index,
-                                       uint32_t* pq_dim)
+                                       int64_t* pq_dim)
 
     cuvsError_t cuvsIvfPqIndexGetPqLen(cuvsIvfPqIndex_t index,
-                                       uint32_t* pq_len)
+                                       int64_t* pq_len)
 
     cuvsError_t cuvsIvfPqIndexGetPqBookSize(cuvsIvfPqIndex_t index,
-                                            uint32_t* pq_book_size)
+                                            int64_t* pq_book_size)
 
     cuvsError_t cuvsIvfPqIndexGetCenters(cuvsIvfPqIndex_t index,
                                          DLManagedTensor * centers)


cjnolet:
Nitpick- but can we please call this "build_from_codebooks" instead of "build_from_centroids"? The codebooks are really the important piece here. Another thing we might consider doing, which would probably be more general would be to just overload the standard build function to accept the codebooks, centroids, and rotation matrix, etc... That would be more precise and more clean IMO.

cjnolet:
> Nitpick- but can we please call this "build_from_codebooks" instead of "build_from_centroids"? The codebooks are really the important piece here. 

Actually, it might be more preferable to call this `build_from_args`. @benfred recently introduced a C API that allows building a CAGRA index from a dataset and graph and the nomenclature used was "build_from_args". I think we should be consistent.

cpp/include/cuvs/neighbors/ivf_pq.h
```diff
@@ -334,6 +334,61 @@ cuvsError_t cuvsIvfPqBuild(cuvsResources_t res,
...
+ * @param[in] params cuvsIvfPqIndexParams_t used to build IVF-PQ index
+ * @param[in] dim dimensionality of the input data
+ * @param[in] pq_centers DLManagedTensor* PQ cluster centers
+ * @param[in] rotation_matrix DLManagedTensor* rotation matrix
+ * @param[in] centers DLManagedTensor* cluster centers
+ * @param[in] centers_rot DLManagedTensor* cluster centers in rotated space
+ * @param[out] index cuvsIvfPqIndex_t Newly built IVF-PQ index
+ * @return cuvsError_t
+ */
+cuvsError_t cuvsIvfPqBuildFromCentroids(cuvsResources_t res,
```
cjnolet: Can we rename this to `cuvsIvfPqBuldFromArgs` please?

python/cuvs/cuvs/neighbors/ivf_pq/ivf_pq.pyx
```diff
@@ -339,6 +339,105 @@ def build(IndexParams index_params, dataset, resources=None):
     return idx
 
 
+@auto_sync_resources
+def build_from_centroids(IndexParams index_params, dim, pq_centers,
```
cjnolet: Please rename to `build_from_args`

cpp/include/cuvs/neighbors/ivf_pq.h
```diff
@@ -267,18 +267,45 @@ uint32_t cuvsIvfPqIndexGetNLists(cuvsIvfPqIndex_t index);
...
- * @param[in] res cuvsResources_t opaque C handle
  * @param[in] index cuvsIvfPqIndex_t Built Ivf-Pq index
  * @param[out] centers Preallocated array on host or device memory to store output,
  * dimensions [n_lists, dim]
  * @return cuvsError_t
  */
-cuvsError_t cuvsIvfPqIndexGetCenters(cuvsResources_t res,
-                                     cuvsIvfPqIndex_t index,
-                                     DLManagedTensor* centers);
+cuvsError_t cuvsIvfPqIndexGetCenters(cuvsIvfPqIndex_t index, DLManagedTensor* centers);
```
cjnolet, line 287: Was this just an oversight? Is `res` not actually used in this function?

cpp/include/cuvs/neighbors/ivf_pq.h
```diff
@@ -334,6 +361,58 @@ cuvsError_t cuvsIvfPqBuild(cuvsResources_t res,
                            cuvsIvfPqIndexParams_t params,
                            DLManagedTensor* dataset,
                            cuvsIvfPqIndex_t index);
+
+/**
+ * @brief Build a IVF-PQ index from existing centroids and codebook. The input tensors must be on
+ *        device memory.
```
cjnolet, line 377: Very nice to call this out.

cpp/include/cuvs/neighbors/ivf_pq.h
```diff
@@ -334,6 +361,58 @@ cuvsError_t cuvsIvfPqBuild(cuvsResources_t res,
...
+ *
+ * // de-allocate `index_params`, `index` and `res`
+ * cuvsError_t params_destroy_status = cuvsIvfPqIndexParamsDestroy(index_params);
+ * cuvsError_t index_destroy_status = cuvsIvfPqIndexDestroy(index);
+ * cuvsError_t res_destroy_status = cuvsResourcesDestroy(res);
+ * @endcode
+ *
+ * @param[in] res cuvsResources_t opaque C handle
+ * @param[in] params cuvsIvfPqIndexParams_t used to build IVF-PQ index
+ * @param[in] dim dimensionality of the input data
```
cjnolet, line 412: Seems like we should know this from the pqindex, right? At least, I can't see why this wouldn't be stored on the index, especially since it's responsible for storing the lists.

cjnolet: Oh, nevermind. We're building the index from the arguments.... now I understand.

cpp/include/cuvs/neighbors/ivf_pq.h
lowener: This was not an oversight, `GetCenters` previously did a copy so the `res` was used. With the new `DeviceTensorView` we can provide a view without needing a copy.
I am still on the verge with this change because even though it would unify the API to only have non-owning views, the view of centers is strided which might not be supported in many functions.

benfred: if the view is strided (which also happens with the dataset in cagra) - the user can always call `cuvsMatrixCopy` to get a non-strided view

cpp/include/cuvs/neighbors/ivf_pq.h
```diff
@@ -267,18 +267,55 @@ uint32_t cuvsIvfPqIndexGetNLists(cuvsIvfPqIndex_t index);
 /** Get the dimensionality */
 uint32_t cuvsIvfPqIndexGetDim(cuvsIvfPqIndex_t index);
 
+/** Get the pq_dim */
+uint32_t cuvsIvfPqIndexGetPqDim(cuvsIvfPqIndex_t index);
```
benfred: We should decide on the API convention for trivial getters like this in the C-API.

For ivf-pq we have the getters returning the value directly,  but in cagra we have the getters returning an error code with the output being passed by a pointer : https://github.com/rapidsai/cuvs/blob/699d32c8be9a344c83456232851430e4c3840484/cpp/include/cuvs/neighbors/cagra.h#L359

I think we should be consistent here across index types, and while I like this version since the corresponding method in the C++ api should be `noexcept` - I still think we maybe should be returning an error code here instead (just in case a different index can't guarantee `noexcept` etc).

@lowener @cjnolet thoughts here?

cpp/include/cuvs/neighbors/ivf_pq.hpp
```diff
@@ -2865,6 +2969,35 @@ void extract_centers(raft::resources const& res,
...
+ * @param[in] res raft resource
+ * @param[in] index IVF-PQ index (passed by reference)
+ * @param[out] codebook PQ codebook [index.pq_dim(), index.pq_len(), index.pq_book_size()]
+ */
+/*
+void extract_codebook(raft::resources const& res,
+                     const index<int64_t>& index,
+                     raft::device_matrix_view<float, uint32_t, raft::row_major> codebook);
+
+void extract_codebook(raft::resources const& res,
```
benfred:
```suggestion

/// @copydoc extract_codebook
void extract_codebook(raft::resources const& res,
```

cpp/include/cuvs/neighbors/ivf_pq.hpp
```diff
@@ -428,8 +448,9 @@ struct index : cuvs::neighbors::index {
    *   - codebook_gen::PER_SUBSPACE: [pq_dim , pq_len, pq_book_size]
    *   - codebook_gen::PER_CLUSTER:  [n_lists, pq_len, pq_book_size]
    */
-  raft::device_mdspan<float, pq_centers_extents, raft::row_major> pq_centers() noexcept;
   raft::device_mdspan<const float, pq_centers_extents, raft::row_major> pq_centers() const noexcept;
+  /** Owning view of the PQ centers */
+  raft::device_mdspan<float, pq_centers_extents, raft::row_major> pq_centers_owning_view();
```
benfred: Do we need to expose both `pq_centers` and `pq_centers_owning_view` ?  I'm worried this is complicating the public API, and leaking implementation details about whether the underlying memory is owned or not to the user.   

CAGRA handles this by returning the `graph_view_` directly if its non-owning, and setting the `graph_view_` to point at the owned graph if its owning the memory https://github.com/rapidsai/cuvs/blob/699d32c8be9a344c83456232851430e4c3840484/cpp/include/cuvs/neighbors/cagra.hpp#L512 . This means the `index.graph()` getter handles both cases transparently.

Would something similar work here? (and if not - why do we need to have two different views on potentially different pq_centers inside the index?).

(also the `owning_view` naming convention here is a bit confusing to me - since I don't see how you can have a mdspan view that also owns the data )

lowener: Yes it is adding a bit of complexity to the API but it allows us to be able to get non-const pointers to those arrays that can be used in the building step.
If the only way to write to these owned array was by making a full copy, then operations like [these](https://github.com/rapidsai/cuvs/blob/branch-25.10/cpp/src/neighbors/ivf_pq/ivf_pq_build.cuh#L290-L313) that transpose the pq centers from a workspace buffer to the index would require an extra buffer of the same size to make the transpose into, then copy it into the index-owned array.

The user would not get two views of different pq-centers:
- If IVF-PQ was built with a dataset the 2 views (const and non-const) point to the same memory location
- If IVF-PQ was built from args, only the const view should be used, the non-const throws an exception

I think there is some value in being able to get a non-const pointer when needed.
What would be a better naming convention for those that would be less confusing?

achirkin: Can't you have this as an overload with the same name? I think they should be distinguishable here at the compile time due to different const-ness of the index. This would also be more along the lines of a typical C++ code.

cpp/include/cuvs/neighbors/ivf_pq.h
lowener: I think it is a good proposition, I made the changes to support error codes for the getters

cpp/include/cuvs/neighbors/ivf_pq.hpp
lowener: It could be an overload with the same name, but it would be much more error-prone for the user.
`index.pq_centers().size()` would be ambiguous and could return the empty mdarray size (0) instead of the mdspan size.
I just tested locally the overload and for a call to `raft::linalg::gemm` ([here](https://github.com/rapidsai/cuvs/pull/1123/files#diff-c607879b080ca6fd0254f451060c19a8a6c7c85e05862730bf83b36a04650621R1355-R1369)) that requires a const pointer, `index.rotation_matrix().data_handle()` is ambiguous as well since the compiler can take non-const view and get a const pointer from it.

achirkin: Well, that gemm should probably be replaced with an mdspan-based version instead of using raw pointers :) but I see your point. I guess, one could use std::enable_if_t<...> to disable the inappropriate version, or even make a single function out of the two with an `if constexpr` inside?..

lowener: `std::enable_if_t<..>` would only work if we added a template specifically for this purpose, which would require the compilation of both template possibilities and make the library bigger.
Unless I'm misunderstanding what you meant there can't be a single function returning either a const and non-const view.
A way to achieve the goal of making it more C++ typical would be to use inheritance to have a "const" index  that stores only the mdspans and a "modifiable" index that stores the mdarrays and can give non-const views. But this would require more changes all over the codebase and I'm not sure it would be worth it

achirkin: I agree adding inheritance or dummy template parameters is not worth it.

> I just tested locally the overload and for a call to `raft::linalg::gemm` ([here](https://github.com/rapidsai/cuvs/pull/1123/files#diff-c607879b080ca6fd0254f451060c19a8a6c7c85e05862730bf83b36a04650621R1355-R1369)) that requires a const pointer, `index.rotation_matrix().data_handle()` is ambiguous as well since the compiler can take non-const view and get a const pointer from it.

I've just tried pull your changes and replace all occurrences of `pq_centers_owning_view` with `pq_centers`. Surprisingly, it did compile. Not sure what was the problem?

lowener: (We discussed about it in a call, the problem is that having both under the same function name can lead to ambiguity at runtime and it would run into runtime exception)

cjnolet: I think we should just keep the const or not keep it and not opt to support both. In practice, I think most users are going to be getting the centers and not modifying them, but we could probably remove the const designator altogether. Would that solve this problem?

cjnolet: But I agree w/ the others above- the user shouldn't be burdened with this- they should just be invoking an accessor method on the object, ideally named after the member they are accessing.

lowener: I modified that code to use kind of the same mechanism as CAGRA with `update_*` to only allow setting that memory through copy. The API is less confusing but users now don't have access to a non-const view. Let me know your opinions on it.

cjnolet: > I modified that code to use kind of the same mechanism as CAGRA with update_* to only allow setting that memory through copy. 

We need to find a way to allow zero copy- both for accessing and mutating. They should be able to do the copy themselves if they so need to. The general assumption should be "direct access, copies should be explicit"

achirkin (changes requested):
Above we've been discussing the peculiarities of accessing the parts of the index depending on how it was constructed. It seems there's no perfect solution: before the PR, IVF-PQ index can be const or non-const, and that implies a user can modify/write its members, such as PQ centers. After the PR, it seems there's no way to override the pq_centers content if the user has built the index using the old api.

Also I'm not a fan of the idea to have IVF-PQ sometimes own its members and sometimes not own. We've tried that with CAGRA and I think that is a sub-optimal user experience.

What if instead we'd pass the mdarrays by rvalue references to the new `build` overload and to the new constructor? Let the user pass the ownership to IVF-PQ. This would keep the index fully owning all its members and save the user from the burden of managing the lifetimes of multiple dependent objects.

And a small nitpick: could you please add a proper description to the PR? Just adding "closes #issue" doesn't look very nice in the main branch commit history and doesn't reflect how the PR addresses the issue. It is also funny that the `build_from_args` in the title of the PR only ever appears in the changeset in a docsting :)

cpp/include/cuvs/neighbors/ivf_pq.h
```diff
@@ -262,23 +262,60 @@ cuvsError_t cuvsIvfPqIndexCreate(cuvsIvfPqIndex_t* index);
 cuvsError_t cuvsIvfPqIndexDestroy(cuvsIvfPqIndex_t index);
 
 /** Get the number of clusters/inverted lists */
-uint32_t cuvsIvfPqIndexGetNLists(cuvsIvfPqIndex_t index);
+cuvsError_t cuvsIvfPqIndexGetNLists(cuvsIvfPqIndex_t index, uint32_t* n_lists);
 
 /** Get the dimensionality */
-uint32_t cuvsIvfPqIndexGetDim(cuvsIvfPqIndex_t index);
+cuvsError_t cuvsIvfPqIndexGetDim(cuvsIvfPqIndex_t index, uint32_t* dim);
```
tarang-jain: https://github.com/rapidsai/cuvs/pull/1272 is using int64_t for these getters for ivf-flat and CAGRA? Can we do the same?

cpp/include/cuvs/neighbors/ivf_pq.hpp
```diff
@@ -470,17 +498,22 @@ struct index : cuvs::neighbors::index {
...
   raft::device_matrix_view<const int8_t, uint32_t, raft::row_major> centers_int8(
     const raft::resources& res) const;
   raft::device_matrix_view<const half, uint32_t, raft::row_major> centers_half(
     const raft::resources& res) const;
 
   /** Cluster centers corresponding to the lists in the rotated space [n_lists, rot_dim] */
-  raft::device_matrix_view<float, uint32_t, raft::row_major> centers_rot() noexcept;
   raft::device_matrix_view<const float, uint32_t, raft::row_major> centers_rot() const noexcept;
+  /** Setter for the cluster centers in the rotated space [n_lists, rot_dim] */
+  void update_centers_rot(
```
tarang-jain, line 514: Can we have these update functions in the `cuvs::neighbors::ivf_pq::helpers` namespace? (consistent with `set_centers`)

cpp/include/cuvs/neighbors/ivf_pq.hpp
```diff
@@ -470,17 +498,22 @@ struct index : cuvs::neighbors::index {
   raft::device_vector_view<const uint32_t, uint32_t, raft::row_major> list_sizes() const noexcept;
 
   /** Cluster centers corresponding to the lists in the original space [n_lists, dim_ext] */
-  raft::device_matrix_view<float, uint32_t, raft::row_major> centers() noexcept;
   raft::device_matrix_view<const float, uint32_t, raft::row_major> centers() const noexcept;
+  /** Setter for the centers [n_lists, dim_ext] or [n_lists, dim] */
+  void update_centers(raft::resources const& handle,
```
tarang-jain, line 503: How exactly is this different from the `set_centers` helper that we currently have? Can we re-use `set_centers` in places where you are calling `update_centers`?

lowener: The new `update_*` functions are required because there is no possibility to obtain a non-const pointer to those data members. So any modification to those data members must go through an index member function.
So in the current implementation `set_centers` can't modify the centers directly anymore, it has to go through `update_centers`. There is still some value in keeping `set_centers` because it will compute the rotated centers for the users.

cpp/include/cuvs/neighbors/ivf_pq.hpp
lowener: I think it would introduce a lot of redonduncy for nothing. These update functions are public and accessible so users might as well directly call them.

cpp/include/cuvs/neighbors/ivf_pq.hpp
tarang-jain: Sounds good! Feel free to mark this one as resolved.

PR #1125, opened 2025-07-16
CAGRA build heuristics for HNSW

achirkin:
Add helper functions to construct a reasonable default CAGRA build parameters based on HNSW parameters.
The goal is to build a CAGRA graph that can be converted to an HNSW graph which has very similar search performance as the corresponding HNSW-built graph.

### DIFF
--- a/cpp/include/cuvs/neighbors/graph_build_types.hpp
+++ b/cpp/include/cuvs/neighbors/graph_build_types.hpp
@@ -56,10 +56,32 @@ struct ivf_pq_params {
   ivf_pq_params(raft::matrix_extent<int64_t> dataset_extents,
                 cuvs::distance::DistanceType metric = cuvs::distance::DistanceType::L2Expanded)
   {
-    build_params = cuvs::neighbors::ivf_pq::index_params::from_dataset(dataset_extents, metric);
+    build_params    = ivf_pq::index_params::from_dataset(dataset_extents, metric);
+    auto n_rows     = dataset_extents.extent(0);
+    auto n_features = dataset_extents.extent(1);
+    if (n_features <= 32) {
+      build_params.pq_dim  = 16;
+      build_params.pq_bits = 8;
+    } else {
+      build_params.pq_bits = 4;
+      if (n_features <= 64) {
+        build_params.pq_dim = 32;
+      } else if (n_features <= 128) {
+        build_params.pq_dim = 64;
+      } else if (n_features <= 192) {
+        build_params.pq_dim = 96;
+      } else {
+        build_params.pq_dim = raft::round_up_safe<uint32_t>(n_features / 2, 128);
+      }
+    }
+
+    build_params.n_lists                  = std::max<uint32_t>(1, n_rows / 2000);
+    build_params.kmeans_n_iters           = 10;
+    build_params.kmeans_trainset_fraction = std::min(1.0, 1.0 / std::sqrt(n_rows * 1e-5));
+    build_params.codebook_kind            = ivf_pq::codebook_gen::PER_SUBSPACE;
 
     search_params                         = cuvs::neighbors::ivf_pq::search_params{};
-    search_params.n_probes                = std::max<uint32_t>(10, build_params.n_lists * 0.01);
+    search_params.n_probes                = std::round(std::sqrt(build_params.n_lists) / 20 + 4);
     search_params.lut_dtype               = CUDA_R_16F;
     search_params.internal_distance_dtype = CUDA_R_16F;
     search_params.coarse_search_dtype     = CUDA_R_16F;
--- a/cpp/include/cuvs/neighbors/hnsw.hpp
+++ b/cpp/include/cuvs/neighbors/hnsw.hpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2024, NVIDIA CORPORATION.
+ * Copyright (c) 2024-2025, NVIDIA CORPORATION.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -63,6 +63,30 @@ struct index_params : cuvs::neighbors::index_params {
    */
   int num_threads = 0;
 };
+
+/**
+ * @brief Create a CAGRA index parameters compatible with HNSW index
+ *
+ * @param dataset The shape of the input dataset.
+ * @param M HNSW index parameter M.
+ * @param ef_construction HNSW index parameter ef_construction.
+ * @param metric The distance metric to search.
+ *
+ * Usage example:
+ * @code{.cpp}
+ *   using namespace cuvs::neighbors;
+ *   raft::resources res;
+ *   auto dataset = raft::make_device_matrix<float, int64_t>(res, N, D);
+ *   auto cagra_params = cagra_compat_params(dataset.extents(), M, efc);
+ *   auto cagra_index = cagra::build(res, cagra_params, dataset);
+ *   auto hnsw_index = hnsw::from_cagra(res, hnsw_params, cagra_index);
+ * @endcode
+ */
+cuvs::neighbors::cagra::index_params cagra_compat_params(
+  raft::matrix_extent<int64_t> dataset,
+  int M,
+  int ef_construction,
+  cuvs::distance::DistanceType metric = cuvs::distance::DistanceType::L2Expanded);
 
 /**@}*/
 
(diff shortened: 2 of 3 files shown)


achirkin:
An example of the produced index: build speedup and the QPS-Recall curves.
openai_5M_L40_comparison

cpp/include/cuvs/neighbors/graph_build_types.hpp
```diff
@@ -56,10 +56,32 @@ struct ivf_pq_params {
   ivf_pq_params(raft::matrix_extent<int64_t> dataset_extents,
                 cuvs::distance::DistanceType metric = cuvs::distance::DistanceType::L2Expanded)
   {
-    build_params = cuvs::neighbors::ivf_pq::index_params::from_dataset(dataset_extents, metric);
+    build_params    = ivf_pq::index_params::from_dataset(dataset_extents, metric);
```
cjnolet, line 59: I really like the `from_dataset` factory method. Could we just expose that here too? It's more explicit and follows our consistent paradigm of using factory functions/methods for object construction instead of relying on constructors (and it helps protect us against changes to the class hierarchy, should we make them in the future).

cpp/include/cuvs/neighbors/hnsw.hpp
```diff
@@ -64,6 +64,39 @@ struct index_params : cuvs::neighbors::index_params {
...
+ * @code{.cpp}
+ *   using namespace cuvs::neighbors;
+ *   raft::resources res;
+ *   auto dataset = raft::make_device_matrix<float, int64_t>(res, N, D);
+ *   auto cagra_params = cagra_compat_params(dataset.extents(), M, efc);
+ *   auto cagra_index = cagra::build(res, cagra_params, dataset);
+ *   auto hnsw_index = hnsw::from_cagra(res, hnsw_params, cagra_index);
+ * @endcode
+ */
+cuvs::neighbors::cagra::index_params cagra_compat_params(
```
cjnolet: Not in love with the name here. let's try to brainstorm something terse but descriptive (and obvious).

cpp/bench/ann/src/cuvs/cuvs_cagra_wrapper.h
```diff
@@ -78,37 +80,18 @@ class cuvs_cagra : public algo<T>, public algo_gpu {
...
-    cuvs::neighbors::cagra::index_params cagra_params;
-    CagraBuildAlgo algo;
-    std::optional<cuvs::neighbors::nn_descent::index_params> nn_descent_params = std::nullopt;
-    std::optional<float> ivf_pq_refine_rate                                    = std::nullopt;
-    std::optional<cuvs::neighbors::ivf_pq::index_params> ivf_pq_build_params   = std::nullopt;
-    std::optional<cuvs::neighbors::ivf_pq::search_params> ivf_pq_search_params = std::nullopt;
-    size_t num_dataset_splits                                                  = 1;
+    // The optimal defaults depend on the dataset shape and thus only available once the build
+    // function is called.
+    using dataset_dependent_params = std::function<cuvs::neighbors::cagra::index_params(
```
tfeher, line 85: Why do we need a function pointer here (as opposed to always calling the same function)? Do we have different init functions based on whether it is standalone CAGRA or CAGRA_HNSW?

cpp/bench/ann/src/cuvs/cuvs_cagra_hnswlib_wrapper.h
```diff
@@ -94,8 +95,16 @@ void cuvs_cagra_hnswlib<T, IdxT>::build(const T* dataset, size_t nrow)
   bool dataset_is_on_host = raft::get_device_for_address(dataset) == -1;
 
   // re-use the CAGRA wrapper to parse build params
-  auto bps                                 = build_param_.cagra_build_param;
-  bps.cagra_params.attach_dataset_on_build = !dataset_is_on_host;
+  auto bps = build_param_.cagra_build_params;
+  // Not very conveniently, the CAGRA wrapper resolves parameters after the dataset shape is known,
+  // so it takes a lambda to do it. Even though we now the shape, we want to use the wrapper as-is,
```
tfeher: typo?
```suggestion
  // so it takes a lambda to do it. Even though we know the shape, we want to use the wrapper as-is,
```

cpp/include/cuvs/neighbors/hnsw.hpp
tfeher: `get_cagra_params`?

cpp/include/cuvs/neighbors/graph_build_types.hpp
```diff
@@ -56,10 +56,32 @@ struct ivf_pq_params {
...
+      } else if (n_features <= 192) {
+        build_params.pq_dim = 96;
+      } else {
+        build_params.pq_dim = raft::round_up_safe<uint32_t>(n_features / 2, 128);
+      }
+    }
+
+    build_params.n_lists                  = std::max<uint32_t>(1, n_rows / 2000);
+    build_params.kmeans_n_iters           = 10;
+    build_params.kmeans_trainset_fraction = std::min(1.0, 1.0 / std::sqrt(n_rows * 1e-5));
```
tfeher: Since `n_lists` increases linearly with `n_rows`, and `kmeans_transet_fraction` ~ `1/sqrt(n_rows)`, this means that 
```
point_per_cluster = n_train / n_list = n_rows * kmeans_trainset_fraction / n_lists ~ 1 / sqrt(n_rows)
```

In practice for 1B points we get 20 `points_per_cluster`. In contrast, [FAISS has 256 as default](https://github.com/facebookresearch/faiss/wiki/How-to-make-Faiss-run-faster#k-means-clustering). But [they also mention](https://github.com/facebookresearch/faiss/wiki/FAQ#how-many-training-points-do-i-need-for-k-means) that the  required `points_per_cluster` diminish as `n_rows` increases.

Therefore I am  wondering whether we are decreasing `points_per_cluster` too aggressively. Was this benchmarked for 1B scale datasets?

cpp/include/cuvs/neighbors/graph_build_types.hpp
achirkin: I'm hesitant to make a breaking change; shall I add a thin wrapper `from_dataset` to this class?
Btw, unfortunately we haven't been very consistent with using a single approach for generating index parameters. Here's the current state overview: https://docs.google.com/document/d/1OGML7V4V2p65trpN9fSwr2e0eRusRohB0hW_1fykImw/edit?tab=t.0#heading=h.yydzz7kh8k3i

cpp/include/cuvs/neighbors/hnsw.hpp
achirkin: Or `make_cagra_params`? For me both are ok; we're in the `hnsw` namespace so hopefully there's no ambiguity if we add a similar function to `cagra` namespace.

cpp/bench/ann/src/cuvs/cuvs_cagra_wrapper.h
achirkin: Yes, we use different helpers to generate the defaults in the two cases.

cpp/include/cuvs/neighbors/graph_build_types.hpp
achirkin: Good point! I only did the benchmarks up to 100M points; this gives 63 points per cluster and I haven't noticed any impact on the quality of the final graph. Do you think we should cap it, just to be safe?
The maximum we can go with CAGRA is 4B rows, which gives 10 points per cluster.

tfeher: I suspect we would need to have a minimum points per cluster for the kmeans training. But I am open to setting that after more benchmarks in a follow up PR.

tfeher (approved):
Thanks Artem for the updates. The implementation details look good to me. Let's continue to discuss the naming issues on the open threads with Corey.

cpp/include/cuvs/neighbors/hnsw.hpp
cjnolet: Maybe `to_cagra_params()` just to denote that we're doing a conversion?

cpp/include/cuvs/neighbors/graph_build_types.hpp
achirkin: I've added the minimum of 32 points as a reasonable safety measure for now.

PR #1128, opened 2025-07-17
Fix for `hellinger` metric

viclafargue:
The implementation of the distance function for the `hellinger` metric had two issues :
1) It did not restore data integrity correctly : double `sqrt` operation
2) It did not check for data overlap (two data chunks/tiles in the pairwise operation have overlapping data).

Closes https://github.com/rapidsai/cuml/issues/7011

### DIFF
--- a/cpp/src/distance/detail/distance.cuh
+++ b/cpp/src/distance/detail/distance.cuh
@@ -298,27 +298,49 @@ void distance_impl(raft::resources const& handle,
 {
   cudaStream_t stream = raft::resource::get_cuda_stream(handle);
 
-  rmm::device_uvector<DataT> y_temp(n * k, stream);
-  raft::copy(y_temp.data(), y, n * k, stream);
-  const DataT* y_temp_ptr = y_temp.data();
+  // Check if arrays overlap
+  const DataT* x_end  = x + m * k;
+  const DataT* y_end  = y + n * k;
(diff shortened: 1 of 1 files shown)


cpp/src/distance/detail/distance.cuh
```diff
@@ -298,11 +298,14 @@ void distance_impl(raft::resources const& handle,
 {
   cudaStream_t stream = raft::resource::get_cuda_stream(handle);
 
+  rmm::device_uvector<DataT> y_temp(n * k, stream);
+  raft::copy(y_temp.data(), y, n * k, stream);
```
tarang-jain: Isn't this copying the whole input matrix? This can be huge! I've been trying to understand the tiling logic, but looking at the hellinger distance op, it seems like that is just a product of sqrt(x) and sqrt(y), so is it possible to do the sqrt and revert for y in place?
On the other hand, if you do require to not alter the y matrix, is there a way you can embed the sqrt in the hellinger op itself rather than copying the whole matrix?

viclafargue: Thanks for the review Tarang! I managed to do it in-place with some pointer arithmetic.

tarang-jain (approved):

cjnolet (approved):

PR #1133, opened 2025-07-18
Binary CAGRA with NN Descent

tarang-jain:
NN Descent can be used as the graph building algo for CAGRA with the BitwiseHamming metric.


jinsolp (approved):
Thanks Tarang, LGTM!

chatman:
Is there any example, test or documentation on how to use binary quantization with CAGRA (say in a C program), please? CC @punAhuja

PR #1134, opened 2025-07-18
[W.I.P][Feat] cuVS+SCADA: Add Dockerfile and minimal benchmark example

rhdong:
Please refer to the [README.md](https://github.com/rapidsai/cuvs/commit/a195a18d7e36cad437ae01ea7a13a2476c906bf3#diff-f684dffd049b84af6913e8374ef2d8bb48736f19d9d4af5f49ec61546964d814) in the PR.

### DIFF
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -1,0 +1,129 @@
+# syntax=docker/dockerfile:1.5
+
+# ==============================================================================
+# RAPIDS cuVS Development Docker Image
+#
+# Uses official RAPIDS devcontainer base image and conda environment configuration
+# for consistency and maintainability. Designed as a comprehensive development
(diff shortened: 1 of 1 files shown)


cjnolet:
@rhdong I think what we really want here is a standard docker image that contains cuvs (ideally the Python package, but that also requires/depends on libcuvs). 

We can (and should) build upon the base cuVS Dockerfile with SCADA bits, and I even think we should have it in a branch in cuvs. The first step, however, is a standard cuVS dockerfile.

cjnolet:
@rhdong the dockerfile is looking good. Can you open a new PR with just the Dockerfile in the root of the repository? This is a heavy ask from users and customers (and relatively low hanging fruit). This will be immediately valuable to users!

PR #1155, opened 2025-07-23
Merge support for tiered index

benfred:


cpp/include/cuvs/neighbors/cagra.hpp
```diff
@@ -1975,6 +1975,11 @@ void serialize_to_hnswlib(
  * @note: When device memory is sufficient, the dataset attached to the returned index is allocated
  * in device memory by default; otherwise, host memory is used automatically.
  *
+ * @note: The `merge_params.merge_strategy` parameter cannot be set to MERGE_STRATEGY_LOGICAL
+ * using this API. Please use the API in the `cuvs/neighbors/composite/merge.hpp` instead
```
cjnolet: i'm hesitant to point users towards thi API because IMO it shouldn't exist in the public APIs at all. The composite index should be fairly opaque to the user and we should be able to polymorpically return it instead of another index type (and have `merge()` coded to return "some" index, whether it be a cagra index or a composite index of cagra indexes. I think you are working towards this but want to make it concrete.

cpp/include/cuvs/neighbors/tiered_index.hpp
```diff
@@ -197,4 +197,30 @@ void search(raft::resources const& res,
...
+ * This function merges multiple tiered indices into one, combining both the datasets and graph
+ * structures.
+ *
+ * @param[in] res
+ * @param[in] indices A vector of pointers to the indices to merge. All indices should
+ *                    be of the same type, and have datasets with the same dimensionality
+ *
+ * @return A new tiered index containing the merged indices
+ */
+auto merge(raft::resources const& res,
```
cjnolet, line 213: We discussed this briefly offline, but I agree it can be kind of awkward to reason about what a merged tiered index would do, and even more awkward- what it would look like. I think instead we should consider the user having to merge the underlying indexes at each of the tiers and provide new fsctory functions that can take those new merged indexes (whether physical or logical) and construct a new tiered index with them. Ideally, the bfknn would only do a physical merge, but I suppose we could always support  a logical merge and if more points are added, we create a new bfknn index in the composite and just add to that?

cpp/src/neighbors/tiered_index.cu
```diff
@@ -175,6 +175,30 @@ void search(raft::resources const& res,
     res, search_params, ivf_pq::typed_search, queries, neighbors, distances, sample_filter);
 }
 
+auto merge(raft::resources const& res,
+           const std::vector<tiered_index::index<cagra::index<float, uint32_t>>*>& indices)
+  -> tiered_index::index<cagra::index<float, uint32_t>>
```
cjnolet, line 181: I think ideally we would have merge() return index<type of the input> and, as you have done, use the PIMPL idom to allow the return of either the same index type that was input, or a composite_index.

cpp/include/cuvs/neighbors/cagra.hpp
benfred: I've removed the reference in the last commit

cpp/include/cuvs/neighbors/cagra.hpp
```diff
@@ -1975,6 +1975,9 @@ void serialize_to_hnswlib(
  * @note: When device memory is sufficient, the dataset attached to the returned index is allocated
  * in device memory by default; otherwise, host memory is used automatically.
  *
+ * @note: This API only supports physical merge (`merge_strategy = MERGE_STRATEGY_PHYSICAL`), and
```
cjnolet, line 1978: Based on our discussion yesterday, we should just remove the "merge strategy" concept all together and support only physical in the "merge" apis. We can still provide a composite that will allow for multiple indexes to be searched concurrently, but don't need to make it part of the merge() apis.

cpp/include/cuvs/neighbors/tiered_index.hpp
cjnolet: Update: After discussion this more offiline, we've decided to remove the merge strategies altogether and only support physical merges in the "merge()" apis. The concept of logical merge might still be done in future APIs (likely through the use of something like a composite_index), but that won't be part of the official merge() API.

cpp/src/neighbors/detail/tiered_index.cuh
```diff
@@ -297,6 +297,79 @@ auto build(
   return std::shared_ptr<index_state<UpstreamT>>(ret);
 }
 
+/**
+ * @brief merge multiple indices together
+ */
+template <typename UpstreamT>
+auto merge(raft::resources const& res,
```
cjnolet, line 304: I wonder- if we are going to rebuild the cagra index anyways to consolidate tiered indices, should we just go ahead and add all the vectors in tier0 to that cagra index too (e.g. should we also do a compaction?). Disregard if this is already what we're doing.

cpp/include/cuvs/neighbors/tiered_index.h
```diff
@@ -246,6 +246,30 @@ cuvsError_t cuvsTieredIndexSearch(cuvsResources_t res,
...
+/**
+ * @}
+ */
+
+/**
+ * @defgroup tiered_c_index_merge Tiered index merge
+ * @{
+ */
+/**
+ * @brief Merge multiple indices together into a single index
```
cjnolet, line 258: Can we flesh out this description a little more to provide more details for the user on exaclty how this does the merge? We should do the same in the c++ description also. Maybe add details about what happens to the vectors in tier 0, what happens to the indexes in tier 1, what are some of the memory considerations (I know there's copies- what are some of the assumptions a user will need to make?)

cjnolet (approved):

PR #1156, opened 2025-07-23
Fix TieredIndexIT,  Scoped resource access change was missed

punAhuja:
#1089 Scoped Resource access changes were missed in the TieredIndex test causing them to fail.
 - Using CheckedCuVSResources.create() now
 - Skip TieredIndex tests if CuVS native support is unavailable

### DIFF
--- a/java/cuvs-java/src/test/java/com/nvidia/cuvs/TieredIndexIT.java
+++ b/java/cuvs-java/src/test/java/com/nvidia/cuvs/TieredIndexIT.java
@@ -18,6 +18,7 @@ package com.nvidia.cuvs;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertFalse;
 import static org.junit.Assert.assertTrue;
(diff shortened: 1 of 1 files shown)


java/cuvs-java/src/test/java/com/nvidia/cuvs/TieredIndexIT.java
```diff
@@ -39,6 +40,7 @@ public class TieredIndexIT extends CuVSTestCase {
 
   @Before
   public void setup() {
+    assumeTrue("not supported on " + System.getProperty("os.name"), isLinuxAmd64());
```
mayya-sharipova, line 43: I think it is good for now, as it conforms with other Tests. But may be later we can get a stronger assumption that CuVSResources themselves could be created, and if not skip the test.

mythrocks:
Good catch, here.  LGTM, pending CI.

mythrocks (approved):
LGTM, pending CI.

PR #1160, opened 2025-07-23
MG C API

viclafargue:


cpp/include/cuvs/core/c_api.h
```diff
@@ -75,6 +75,23 @@ cuvsError_t cuvsResourcesCreate(cuvsResources_t* res);
...
 cuvsError_t cuvsResourcesDestroy(cuvsResources_t res);
 
+/**
+ * @brief Create an Initialized opaque C handle for C++ type `raft::device_resources_snmg`
+ *        for multi-GPU operations
+ *
+ * @param[in] res cuvsResources_t opaque C handle
+ * @return cuvsError_t
+ */
+cuvsError_t cuvsMgResourcesCreate(cuvsResources_t* res);
```
cjnolet: Small nitpick- and maybe I'm the only one who feels this way (so anyone else reviewing this, please feel free to override my opinion). I had to do a double and a triple take to realize that this was not attempting to overload the C function above, but rather it's creating a multi-gpu resources instance. I'd be in favor of actually spelling out `cuvsMultiGpuResources` here just to make sure that's absolutely clear to users. I feat they might not notice this because it's so subtle and it'll just lead to confusion (and ultimately frustration) with our APIs.

cpp/include/cuvs/neighbors/ivf_pq.h
```diff
@@ -459,3 +459,6 @@ cuvsError_t cuvsIvfPqExtend(cuvsResources_t res,
 #ifdef __cplusplus
 }
 #endif
+
+// Include Multi-GPU support
+#include <cuvs/neighbors/mg_ivf_pq.h>
```
cjnolet: Can we please just put the functions right in this file so we don't need to have separate headers for these? Ideally, we'd try to minimize the number of headers because these are very very lightweight (just prototype definitions). Especially if we are always going to include these, I'd prefer just to put them right in `ivf_pq.h`, for example.

cpp/include/cuvs/neighbors/mg_common.h
```diff
@@ -0,0 +1,67 @@
...
+  /** Each search query is processed by a single GPU in a round-robin fashion */
+  CUVS_MG_ROUND_ROBIN
+} cuvsMgReplicatedSearchMode;
+
+/**
+ * @brief Merge mode when using a sharded index
+ */
+typedef enum {
+  /** Search batches are merged on the root rank */
+  CUVS_MG_MERGE_ON_ROOT_RANK,
```
cjnolet: We could have potentially many things in the future namespaced with "CUVS_MG" which might exist above and beyond just the indexes. Please further nest this namespace- `CUVS_NEIGHBORS_MG...`. Even better, let's mirror the function names: `CUVS_INDEX_MULTIGPU`, `CUVS_CAGRA_INDEX_MULTIGPU`, etc..

cpp/include/cuvs/neighbors/mg_common.h
```diff
@@ -0,0 +1,67 @@
...
+  /** Index is split on several devices, favors scaling */
+  CUVS_MG_SHARDED
+} cuvsMgDistributionMode;
+
+/**
+ * @brief Search mode when using a replicated index
+ */
+typedef enum {
+  /** Search queries are split to maintain equal load on GPUs */
+  CUVS_MG_LOAD_BALANCER,
```
cjnolet: See comment below on further nesting these.

cpp/include/cuvs/neighbors/mg_ivf_flat.h
```diff
@@ -0,0 +1,285 @@
...
+ */
+cuvsError_t cuvsMgIvfFlatIndexParamsCreate(cuvsMgIvfFlatIndexParams_t* index_params);
+
+/**
+ * @brief De-allocate Multi-GPU IVF-Flat Index params
+ *
+ * @param[in] index_params
+ * @return cuvsError_t
+ */
+cuvsError_t cuvsMgIvfFlatIndexParamsDestroy(cuvsMgIvfFlatIndexParams_t index_params);
```
cjnolet: As mentioned above, I know it's verbose, but `MultiGpu` would be much easier to spot

cpp/src/neighbors/mg_cagra_c.cpp
```diff
@@ -0,0 +1,373 @@
...
+        delete mg_index_ptr;
+      }
+      delete index;
+    }
+  });
+}
+
+namespace cuvs::neighbors::cagra {
+
+// Forward declarations for functions defined in cagra_c.cpp
```
benfred: we have header files for these in `cagra_c.hpp` in this directory - rather than redefine the declaration we should just `#include "cagra_c.hpp"` here instead

cpp/src/neighbors/mg_cagra_c.cpp
```diff
@@ -0,0 +1,373 @@
...
+      .n_rows_per_batch = 1LL << 20               // Default to 1M rows per batch
+    };
+  });
+}
+
+extern "C" cuvsError_t cuvsMgCagraSearchParamsDestroy(cuvsMgCagraSearchParams_t params)
+{
+  return cuvs::core::translate_exceptions([=] {
+    if (params) {
+      // Base parameters are destroyed automatically
```
benfred: I think we might need to call `cuvsSearchParamsDestroy` here to free up the memory - same is with the index params

cpp/src/neighbors/mg_cagra_c.cpp
```diff
@@ -0,0 +1,373 @@
...
+{
+  return cuvs::core::translate_exceptions([=] {
+    auto dataset      = dataset_tensor->dl_tensor;
+    index->dtype.code = dataset.dtype.code;
+    index->dtype.bits = dataset.dtype.bits;
+
+    if (dataset.dtype.code == kDLFloat && dataset.dtype.bits == 32) {
+      index->addr = reinterpret_cast<uintptr_t>(_mg_build<float>(res, *params, dataset_tensor));
+
+    } else if (dataset.dtype.code == kDLInt && dataset.dtype.bits == 8) {
```
benfred, line 278: Can we be support fp16 here? I see that we have a `mg_cagra_half_uint32.cu` - so it seems like we should have instantiations in the c++ layer for this.

Also, should we be adding both int8/uint8 support here? I know we've been talking about going down to just one 8bit int for saving binary sizes

cpp/src/neighbors/mg_cagra_c.cpp
```diff
@@ -0,0 +1,373 @@
...
+  });
+}
+
+extern "C" cuvsError_t cuvsMgCagraDeserialize(cuvsResources_t res,
+                                              const char* filename,
+                                              cuvsMgCagraIndex_t index)
+{
+  return cuvs::core::translate_exceptions([=] {
+    // We need to read dtype from file since we don't know it yet
+    // For now, we'll default to float32
```
benfred: We have support for reading the dtype in the single gpu cagra bindings - can do the same thing here too?

https://github.com/rapidsai/cuvs/blob/60e8af26cc72a5336f8d9c19dd95a25462913018/cpp/src/neighbors/cagra_c.cpp#L742-L747

cpp/src/neighbors/mg_cagra_c.cpp
```diff
@@ -0,0 +1,373 @@
...
+      .mode        = CUVS_MG_SHARDED  // Default to sharded mode
+    };
+  });
+}
+
+extern "C" cuvsError_t cuvsMgCagraIndexParamsDestroy(cuvsMgCagraIndexParams_t index_params)
+{
+  return cuvs::core::translate_exceptions([=] {
+    if (index_params) {
+      // Base parameters are destroyed automatically
```
benfred: how are base_params destroyed automatically? Right now they are creatted on line 34 with cuvsCagraIndexParamsCreate - and I would expect that we'd need to call cuvsIndexParamsDestroy here to free up the memory

cpp/src/neighbors/mg_ivf_flat_c.cpp
```diff
@@ -0,0 +1,364 @@
...
+      .mode        = CUVS_MG_SHARDED  // Default to sharded mode
+    };
+  });
+}
+
+extern "C" cuvsError_t cuvsMgIvfFlatIndexParamsDestroy(cuvsMgIvfFlatIndexParams_t index_params)
+{
+  return cuvs::core::translate_exceptions([=] {
+    if (index_params) {
+      // Base parameters are destroyed automatically
```
benfred: I think we need to explicitly destroy base parameters (?)

cpp/src/neighbors/mg_ivf_flat_c.cpp
```diff
@@ -0,0 +1,364 @@
...
+        delete mg_index_ptr;
+      }
+      delete index;
+    }
+  });
+}
+
+namespace cuvs::neighbors::ivf_flat {
+
+// Forward declarations for functions defined in ivf_flat_c.cpp
```
benfred: can we `#include "ivf_flat_c.hpp"` here instead of these forward declarations?

cpp/include/cuvs/neighbors/ivf_pq.h
viclafargue: I think that the `ivf_pq.h` file should indeed probably not include the the `mg_ivf_pq.h` header file here. But these two should be different files as we would like to be able to compile the library with and without the multi-GPU feature.

cpp/include/cuvs/neighbors/mg_common.h
viclafargue: This option is common to every index type. But, I could still scope it better with `CUVS_NEIGHBORS_MG`.

cpp/src/neighbors/mg_cagra_c.cpp
viclafargue: You're right, this is a pointer to a struct. I just fixed it.

cpp/src/neighbors/mg_cagra_c.cpp
viclafargue: Just added the half type. Adding both int8 and uint8 types for now as I would like to follow what is available in the C++ API, but I think that we can remove one of them in a follow-up PR when deemed necessary. I think that most of the binary size is due to kernel compilations, not the MG wrapper.

cpp/include/cuvs/neighbors/mg_ivf_flat.h
```diff
@@ -0,0 +1,285 @@
+/*
+ * Copyright (c) 2024-2025, NVIDIA CORPORATION.
```
cjnolet: Please make sure all new files only contain the current year.

cpp/src/neighbors/mg_ivf_flat_c.cpp
```diff
@@ -0,0 +1,397 @@
...
+                                                cuvsMultiGpuIvfFlatIndexParams_t params,
+                                                DLManagedTensor* dataset_tensor,
+                                                cuvsMultiGpuIvfFlatIndex_t index)
+{
+  return cuvs::core::translate_exceptions([=] {
+    auto dataset      = dataset_tensor->dl_tensor;
+    index->dtype.code = dataset.dtype.code;
+    index->dtype.bits = dataset.dtype.bits;
+
+    if (dataset.dtype.code == kDLFloat && dataset.dtype.bits == 32) {
```
benfred, line 271: Can we add half precision support for ivf-flat here too? I see that you've added for cagra and ivf-pq which is great -  but looks like ivf-flat doesn't have this (in either the c-api here, or in the C++ api too).

viclafargue: Oops, thanks for noticing this. It looks like the half type was missing from the MG IVF-Flat API.

viclafargue:
I wanted to note that I’ve made some adjustments to the serialization/deserialization process to account for how the C API handles index deserialization. In C++, the deserialization is templated, so the function already “knows” the type of index it’s about to read. In contrast, the C API must determine the index type by reading it from the file itself. To support this, MG index files will now include the index type in the first 4 bytes. Please note that this change may cause incompatibility with index files serialized using earlier versions.

cc @cjnolet

PR #1175, opened 2025-07-24
Adding unfused GEMM + reduction path to L2 NN

vinaydes:


cpp/src/distance/unfused_distance_nn.cuh
```diff
@@ -0,0 +1,333 @@
...
+      thread_min.key   = col;
+    }
+  }
+  typedef cub::BlockReduce<OutT, TPB> BlockReduceT;
+  __shared__ typename BlockReduceT::TempStorage temp_storage;
+  auto block_result = BlockReduceT(temp_storage).Reduce(thread_min, Reducer<AccT, IdxT>{});
+
+  if (threadIdx.x == 0) {
+    if (is_sqrt) {
+      out[row] = block_result;
```
vinaydes: Need to call sqrt function here.

cpp/src/cluster/detail/kmeans_balanced.cuh
```diff
@@ -62,6 +65,21 @@ namespace cuvs::cluster::kmeans::detail {
 
 constexpr static inline float kAdjustCentersWeight = 7.0f;
 
+template <typename MathT, typename IdxT, typename LabelT>
+bool use_fused(IdxT m, IdxT n, IdxT k)
+{
+#if __CUDA_ARCH__ > 800
```
tfeher: This will not work, the macro is not defined in host code. Try using the helpers from `raft/util/arch.cuh`

https://github.com/rapidsai/cuvs/blob/1155a3a427cd1d1bfaf8fe74a937ed6dfa797ec7/cpp/src/neighbors/detail/nn_descent.cuh#L1210-L1222

cpp/src/distance/unfused_distance_nn.cuh
```diff
@@ -0,0 +1,333 @@
...
+    } else {
+      out[row] = block_result;
+    }
+  }
+}
+
+template <typename DataT, typename AccT, typename OutT, typename IdxT>
+void reduce_min(OutT* out,
```
tfeher, line 167: It would be great if we do not need a separate reduction kernel here. But I see that this fuses two things:
- apply normalization
- min reduce

Fusing these ops is important, because the matrix that we are reducing is expected to be large. Therefore expressing these elementwise ops followed by a separate reduction is not an option. Alternative option would be to include the additional terms in the gemm, like in ivf_pq

https://github.com/rapidsai/cuvs/blob/1155a3a427cd1d1bfaf8fe74a937ed6dfa797ec7/cpp/src/neighbors/ivf_pq/ivf_pq_search.cuh#L95-L104

But for the moment we can probably keep this simple reduction utility.

cpp/src/cluster/detail/kmeans_balanced.cuh
vinaydes: Added a TODO item for this, thanks. Marking it resolved for now.

tfeher: This is not just a request to use existing helpers. We are in host code, the __CUDA_ARCH__ macro is not defined. Currently the function always returns true.

vinaydes:
I am closing this PR as it is slightly dated now. I'll open a fresh one for same work.

Issue #1056, opened 2025-06-27
[BUG] `pairwise_distances` returns different distances for `InnerProduct` metric


jinsolp:
**Describe the bug**
`cuvs::distance::pairwise_distance` functions returns different distances when they should be the same.

Given dataset `X` of size `n x dim` the following have different values for relatively small `N` and likely prime values of `dim`.
- distance between vector 0 and vector 1 by running `pairwise_dist(X, X)` -> resulting in distances matrix `n x n`
- distance between vector 0 and vector 1 by running `pairwise_dist(X[:512], X)` -> resulting in distances matrix `512 x n`

**Steps/Code to reproduce bug**
```
#include <cuvs/distance/distance.hpp>

#include <raft/random/make_blobs.cuh>
#include <raft/core/device_mdarray.hpp>
#include <raft/core/host_mdarray.hpp>
#include <raft/core/device_mdspan.hpp>

#include <iostream>

void compare(const raft::resources& handle, size_t n_rows, size_t dim, cuvs::distance::DistanceType metric) {
    auto dataset = raft::make_device_matrix<float, int64_t>(handle, n_rows, dim);
    auto labels = raft::make_device_vector<int64_t, int64_t>(handle, n_rows);
    raft::random::make_blobs(handle, dataset.view(), labels.view()); 

    // search on the entire dataset. resulting output distance is n_rows x n_rows
    auto out_pairwise_dists = raft::make_device_matrix<float, int64_t>(handle, n_rows, n_rows);
    cuvs::distance::pairwise_distance(
        handle,
        raft::make_const_mdspan(dataset.view()),
        raft::make_const_mdspan(dataset.view()),
        out_pairwise_dists.view(),
        metric,
        2.0);
    
    // this holds distance between vector 0 and vector 1
    float vec0_vs_vec1;
    raft::copy(&vec0_vs_vec1, out_pairwise_dists.data_handle()+1, 1, raft::resource::get_cuda_stream(handle));

    // make search matrix the first 512 rows. Resulting output distances is 512 x n_rows
    size_t tile_size = 512;
    auto out_pairwise_dists_tiled = raft::make_device_matrix<float, int64_t>(handle, tile_size, n_rows);
    auto tiled_search_matrix_view = raft::make_device_matrix_view<float, int64_t>(dataset.data_handle(), tile_size, dim);
    cuvs::distance::pairwise_distance(
        handle,
        raft::make_const_mdspan(tiled_search_matrix_view),
[...]

PR #1185, opened 2025-07-26
Make duplicate removal in all neighbors robust to distance drift across batches

jinsolp:
This is to fix an edge case that happens and the root cause is in issue: https://github.com/rapidsai/cuvs/issues/1056, which is about different distance results from `raft::linalg::gemm` based on the input sizes.

Right now, when merging two knn graphs from different batches, we sort by distances (i.e. keys), and if the distances are same we sort by indices (i.e. values). After doing so, we compare indices right next to each other to check for duplicates under assumption that same vectors end up with same distances.

However, due to the problem stated in issue 1056, distance for same index can be slightly different based on the size of the input matrix to gemm (or where the vector is in the entire matrix).

For example, say we are calculating nearest neighbors for vector 0.
we could end up with
```
indices = [1, 2, 3, 2, ....]
distances = [0.023, 0.02355981, 0.02355983, 0.02355987]
```
because distance between vector 0 and vector 2 is calculated as 0.02355981 in the first batch, and 0.02355987 in the second batch.

This PR fixes this issue by sweeping the entire row for duplicates, instead of checking the one right next to itself.

### DIFF
--- a/cpp/src/neighbors/all_neighbors/all_neighbors_merge.cuh
+++ b/cpp/src/neighbors/all_neighbors/all_neighbors_merge.cuh
@@ -119,7 +119,21 @@ RAFT_KERNEL merge_subgraphs_kernel(IdxT* cluster_data_indices,
     for (int i = 0; i < ITEMS_PER_THREAD; i++) {
       size_t colId = idxBase + i;
       if (colId > 0 && colId < 2 * graph_degree) {
-        uniqueMask[colId] = static_cast<int16_t>(blockValues[colId] != blockValues[colId - 1]);
+        // this assumes same distance between vector from two different batches, which should be
+        // true. however, currently there are subtle differences in the result based on the matrix
+        // size used to call gemm. This makes it difficult to remove duplicates, because they might
+        // no longer be right next to each other after sorting by distances. Thus, for now we sweep
+        // the whole row to check for duplicates, and keep the first occurrence only.
+        // related issue: https://github.com/rapidsai/cuvs/issues/1056
+        // uniqueMask[colId] = static_cast<int16_t>(blockValues[colId] != blockValues[colId - 1]);
+        int is_unique = 1;
+        for (int j = 0; j < 2 * graph_degree; ++j) {
+          if (j < colId && blockValues[j] == blockValues[colId]) {
+            is_unique = 0;
+            break;
+          }
+        }
+        uniqueMask[colId] = static_cast<int16_t>(is_unique);
       }
     }
 


cjnolet:
@jinsolp thanks for working to fix this issue. Have you done a profiling or benchmarking to gauge the impact of the fix? Just want to make sure this doesn't have a huge impact on the resulting perf (and quantify it in some way).

jinsolp:
On average about +3 seconds to run all_neighbors e2e using 10 clusters on a 10M x 128 dataset with k=32. I don't think it's too concerning.

cjnolet:
@jinsolp can you share more info about the benchmark? What percentage of the overall time was 3 seconds? Was it <1%? Was it 70% of the overall time?

jinsolp:
Oh yes sure the entire e2e time was about 67 seconds vs 70 seconds

cjnolet:
Thanks @jinsolp, so that amounts to about a 4.4% hit in perf, unfortunately. Still, I'd rather have a 5% hit for better correctness than to have a faster impl that is incorrect. 

Do you have a sense for how much this drift happens in practice? Does it pop up nearly all the time? Is it an edge case?

jinsolp:
This doesn't happen that often (especially if the distances are large in scale so are unlikely to end up being the same). It’s also not a major issue for NN Descent, only for brute-force search that directly uses `raft::linalg::gemm`. Still, I’m a bit concerned because I’ve seen this issue a few times adding mutual reachability for brute force. After knowing that `raft::linalg::gemm` is causing it, I ran (non-mr) brute force with a larger k which ends up using `raft::linalg::gemm` internally, and it has similar issue.

cjnolet:
> This doesn't happen that often (especially if the distances are large in scale so are unlikely to end up being the same). It’s also not a major issue for NN Descent, only for brute-force search that directly uses raft::linalg::gemm. Still, I’m a bit concerned because I’ve seen this issue a few times adding mutual reachability for brute force. After knowing that raft::linalg::gemm is causing it, I ran (non-mr) brute force with a larger k which ends up using raft::linalg::gemm internally, and it has similar issue.

Yeah, that's where I'm goin with things. I definitely don't think we should ignore this issue by any means, but I wonder if we could lessen the perf effects by characterizing under what conditions we see it happening and try to come up with some rules that we can apply so that this only need to be done under those conditions. For example, is there are specific dimentionality where it starts happening? Or a specific number of rows? Does it always happen below/above those sizes? Is it specific to small matrices only? Is it that it doesn't happen with the larger scales or that its effects are less noticeable? 

Trust me, I"m not trying to halt progress here, but imagine if we made a bunch of these fixes that each had a 4% or more impact on perf... eventually the perf gap would become unacceptable. Just trying to make sure we can "soften the blow" so to speak.

jinsolp:
Got it, I'll try to narrow this down a bit to see what we can do. Let me also try to run this many times before averaging it (right now I only ran it 5 times each to average) because I didn't expect this to add +3 seconds.

jinsolp:
@cjnolet since this is a rare edge case, and doesn't happen across many different vectors, it should be enough to sweep a small neighboring window instead of sweeping the entire row. Window size 4 is chosen heuristically.

jinsolp:
Hmmm I realized that the duplicate problem happens for mutual reachability unless we sweep the whole row.
For example

```
[After ordinary KNN pass]
neighbors of 25: [25,1840,2435,1185,3745,290,955,410,970,4410,3220,3435,1420,3640,490,320];
distances for 25: [-0,187.677,189.562,194.995,197.123,199.602,205.618,205.697,207.173,207.21,211.438,211.749,212.001,212.875,214.381,214.964];

Core dist for this is 214.964

[After MR]
neighbors of 25: [25,290,320,410,490,970,1185,1420,1840,2435,3220,3435,3640,3745,4410,320];
distances for 25: [214.964,214.964,214.964,214.964,214.964,214.964,214.964,214.964,214.964,214.964,214.964,214.964,214.964,214.964,214.964,214.965];
```

neighbor 320 is calculated to have distance 214.965 with vector 25 in the second run, so now duplicate 320 cannot be detected unless the entire row is sweeped.


I think what we can do is sweep the whole row when we are calculating MR, and just sweep a small window otherwise.

jinsolp:
Window sweep (checking 4 neighbors) doesn't add too much overhead. We only sweep all if we set the template param `SweepAll=true`.
`merge_subgraph_kernel` times below.
Screenshot 2025-08-06 at 2 59 17 PM

cjnolet:
@jinsolp are you comfortabl with the fix now or are you still trying to find a solution to the problem you listed above?

jinsolp:
I think I am comfortable with the new fix!

PR #1188, opened 2025-07-28
[FEA] Experimental: automatic generation of JSON for parameter types and more 

achirkin:
Enable automatic generation of `to_json` / `from_json` using `nlohmann_json` for simple types.

### Usage:
Declare an instance of `cuvs::core::generic<T>` for type T. For example:
```C++
struct index_params {
  friend class cuvs::core::generic<index_params>;  // a mark for code generation
  cuvs::distance::DistanceType metric = cuvs::distance::DistanceType::L2Expanded;
  float metric_arg = 2.0f;
};
```
The friend class declaration does nothing but declares an implicit instance of `cuvs::core::generic<index_params>`.
Then, at the call site, one can use helper functions `cuvs::generic::to_json<T>, cuvs::generic::from_json<T>, cuvs::generic::to_string<T>` declared in `cuvs/core/generic.hpp`.

### Implementation
During the CMake configure step, the `generate_generic_json` tool scans the compilation database to find all instances of `cuvs::core::generic` and generates to_json/from_json static members for them. The tool then generates a cmake target per header file and injects the generated targets into the existing real targets.

### Rationale

The original use case is to be able to print the actual build/search parameters of cuVS algorithms in the ann-bench tool: we have a more and more complicated logic to set up defaults and infer parameters from the problem size; it's not always clear anymore what are the actual parameters passed to the algorithm.

This JSON serialization, however, can be useful in other situations as well:
 1. The same printing during tests
 2. Use from_json in benchmarks to automatically expose all parameters for all algorithms.
 3. Use to_json/from_json in the main de/serialization API

### Alternatives
The main alternative is using the macros from nlohmann_json library: https://json.nlohmann.me/api/macros/nlohmann_define_derived_type/ .
 - The nlohmann variant uses macros and requires us to explicitly manually enumerate all fields including parents, which is tedious and may error prone (especially when modifying a parent of an existing class).
 - The proposed solution relies on clang tooling and cmake. Thus it adds up to build time and the complexity of the cmake setup. It also adds the build-time dependency on clang-dev (NB NOT the library dependency, i.e. the difference in the library size should be negligible).

### WIP - Todo

This is a prototype; if we decide to proceed with this approach, I'll need to:
  - replace the current parameter parsing in benchmarks where possible
  - replace the current parameter reporting in tests where possible
  - cleanup the generated code/helpers
  - review and cleanup cmake files (manual and generated)


cjnolet:
Hey @achirkin this looks like it could be a useful tool. However, don't the cuvs-bench algorithms already print the build and search params to the console and wrote them in the underlying csv files? In other words, where and how do you see this being used in a cuvs-bench workflow?

achirkin:
Hi @cjnolet sorry I missed your comment. Currently, the cuvs-bench algorithms do only print the parameters explicitly passed in the json. We don't have a way to find out exactly which parameters are passed to the build/search functions at the moment. This has been a rather big pain point, because of all sorts of substitutions and heuristics we have in the benchmarks and in the library (e.g. some parameters are computed based on the dataset size, others on the number of SMs in the GPU etc).

PR #1197, opened 2025-07-30
Spectral Embedding precomputed graph api

### LINKED ISSUE #1184: Spectral Embedding precomputed graph API
We need to add an option to pass in a precomputed connectivity graph to spectral embedding. This is because umap already computes the connectivity graph for us.

aamijar:
Resolves #1184

This API is needed so that we have the option to pass in a precomputed connectivity graph. This allows users to build a connectivity graph using their preferred method whether it is through brute_force or other means and then pass the input to the spectral embedding computation.

### DIFF
--- a/cpp/include/cuvs/preprocessing/spectral_embedding.hpp
+++ b/cpp/include/cuvs/preprocessing/spectral_embedding.hpp
@@ -48,4 +48,9 @@ void transform(raft::resources const& handle,
                raft::device_matrix_view<float, int, raft::row_major> dataset,
                raft::device_matrix_view<float, int, raft::col_major> embedding);
 
+void transform_precomputed(raft::resources const& handle,
+                           params config,
+                           raft::device_coo_matrix<float, int, int, int>& connectivity_graph,
+                           raft::device_matrix_view<float, int, raft::col_major> embedding);
+
 }  // namespace cuvs::preprocessing::spectral_embedding
--- a/cpp/include/cuvs/preprocessing/spectral_embedding.hpp
+++ b/cpp/include/cuvs/preprocessing/spectral_embedding.hpp
@@ -48,9 +48,9 @@ void transform(raft::resources const& handle,
                raft::device_matrix_view<float, int, raft::row_major> dataset,
                raft::device_matrix_view<float, int, raft::col_major> embedding);
 
-void transform_precomputed(raft::resources const& handle,
-                           params config,
-                           raft::device_coo_matrix<float, int, int, int>& connectivity_graph,
-                           raft::device_matrix_view<float, int, raft::col_major> embedding);
+void transform(raft::resources const& handle,
+               params config,
+               raft::device_coo_matrix<float, int, int, int>& connectivity_graph,
+               raft::device_matrix_view<float, int, raft::col_major> embedding);
 
 }  // namespace cuvs::preprocessing::spectral_embedding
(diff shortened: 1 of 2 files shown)


cpp/include/cuvs/preprocessing/spectral_embedding.hpp
```diff
@@ -48,4 +48,9 @@ void transform(raft::resources const& handle,
                raft::device_matrix_view<float, int, raft::row_major> dataset,
                raft::device_matrix_view<float, int, raft::col_major> embedding);
 
+void transform_precomputed(raft::resources const& handle,
```
cjnolet: Since the input here is different than in the sense case, let’s just call this “transform()” and overload it. It provides a nicer experience for users to not have to remember more function names.

cpp/src/preprocessing/spectral/spectral_embedding.cu
```diff
@@ -234,6 +234,9 @@ void compute_eigenpairs(raft::resources const& handle,
     raft::make_const_mdspan(col_indices.view()),     // Column indices to gather
     embedding_row_view                               // Destination matrix (as row-major view)
   );
+
+  // raft::print_device_vector("embedding", embedding.data_handle(), embedding.extent(0) *
```
cjnolet: Note: remember to remove this or put it in a debug log statement.

cpp/src/preprocessing/spectral/spectral_embedding.cu
```diff
@@ -257,4 +260,31 @@ void transform(raft::resources const& handle,
...
+  const int n_samples = connectivity_graph.structure_view().get_n_rows();
+
+  auto sym_coo_row_ind = raft::make_device_vector<int>(handle, n_samples + 1);
+  auto diagonal        = raft::make_device_vector<float, int>(handle, n_samples);
+
+  auto csr_matrix_view =
+    coo_to_csr_matrix(handle, n_samples, sym_coo_row_ind.view(), connectivity_graph);
+  auto laplacian =
+    create_laplacian(handle, spectral_embedding_config, csr_matrix_view, diagonal.view());
+
```
cjnolet: Same as above- remove or put in a debug log statement

aamijar:
Thanks for the review! I have addressed your comments here https://github.com/rapidsai/cuvs/pull/1197/commits/8ee53793b49c0bbcda377c7a58df0fd83d980bcb

cpp/src/preprocessing/spectral/spectral_embedding.cu
```diff
@@ -112,35 +112,35 @@ raft::device_csr_matrix_view<float, int, int, int> coo_to_csr_matrix(
...
                                            n_samples,
                                            stream);
 
-  auto sym_coo_nnz = sym_coo_matrix.structure_view().get_nnz();
+  auto sym_coo_nnz = sym_coo_matrix_view.structure_view().get_nnz();
   raft::copy(sym_coo_row_ind.data_handle() + sym_coo_row_ind.size() - 1, &sym_coo_nnz, 1, stream);
 
   auto csr_matrix_view = raft::make_device_csr_matrix_view<float, int, int, int>(
-    const_cast<float*>(sym_coo_matrix.get_elements().data()),
+    const_cast<float*>(sym_coo_matrix_view.get_elements().data()),
```
cjnolet, line 137: Good catch here- absolutely the old object used to be always owning and we'd pass it around via pointer. The new object has explicit owning / view vairants.

cjnolet (approved):

PR #1209, opened 2025-08-04
[Java] Add decorator for synchronized CuVSResource access

ldematte:
As a follow up to #1089 and #1082, this PR introduces a decorator to ensure that access to a shared CuVSResource is synchronized. This could be useful in cases where application logic is complex and it's not easy to efficiently and cleanly guarantee that access to a CuVSResources is done by at most one thread at each given time.

### DIFF
--- a/java/cuvs-java/src/test/java/com/nvidia/cuvs/CagraMultiThreadStabilityIT.java
+++ b/java/cuvs-java/src/test/java/com/nvidia/cuvs/CagraMultiThreadStabilityIT.java
@@ -16,9 +16,7 @@
 package com.nvidia.cuvs;
 
 import static com.carrotsearch.randomizedtesting.RandomizedTest.assumeTrue;
-import static org.junit.Assert.assertNotNull;
-import static org.junit.Assert.assertTrue;
-import static org.junit.Assert.fail;
+import static org.junit.Assert.*;
 
 import com.carrotsearch.randomizedtesting.RandomizedRunner;
 import com.nvidia.cuvs.CagraIndexParams.CagraGraphBuildAlgo;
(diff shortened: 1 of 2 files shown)


java/cuvs-java/src/test/java/com/nvidia/cuvs/CagraMultiThreadStabilityIT.java
```diff
@@ -86,105 +103,125 @@ public void testQueryingUsingMultipleThreads() throws Throwable {
...
-      for (Future<?> future : futures) {
-        try {
-          future.get(120, TimeUnit.SECONDS); // Longer timeout for stress test
-        } catch (Exception e) {
-          allCompleted = false;
-          log.error("Thread failed: {}", e.getMessage(), e);
-          if (firstError.get() == null) {
-            firstError.set(e);
+        // Start all threads simultaneously to maximize contention
+        log.info("Starting all {} threads simultaneously...", numThreads);
```
mythrocks, line 144: Not a problem for this PR here.  But I would like the tests to run without noise on the screen.  We've done this, to a degree, in the `TieredIndex` tests.

This test is far less noisy than some of the other (extant) ones.  We'll address that separately.

mythrocks (approved):
LGTM!

java/cuvs-java/src/test/java/com/nvidia/cuvs/CagraMultiThreadStabilityIT.java
ldematte: ++ to review logging levels. Probably a bit of "info" level feedback is OK, but most statements should probably be debug or trace, so we can turn them on when needed.

PR #1211, opened 2025-08-04
Removing filter templates in ivfpq for binary size reduction

jinsolp:
### Summary

This PR removed filters from templates in ivfpq. Only applied to `compute_similarity_kernel` in this PR, but can be applied elsewhere (e.g. `ivfflat_interleaved_scan`) which will be a follow-up PR to this one.

`ivf_to_sample_filter` is passed from the user as-is as a host side filter. `ivf_to_sample_dev_wrapper` handles the lifetime of its device counterpart.

### Perf check
Ran the default setting with bitset flter 20 times to check if this slows anything down.
All numbers are averaged across the 20 runs. Profiing done with nsys.

`compute_similarity` kernel runtime
Screenshot 2025-08-04 at 1 10 31 PM

Think the extra time is coming from the `switch` statements inside `ivf_to_sample_filter_dev`'s `operator()`. But not too much of an overhead.

### Binary size reduction
Reduces uncompressed size from 2910.70 MB -> 2670.04 MB


cpp/include/cuvs/neighbors/common.hpp
```diff
@@ -508,6 +514,9 @@ struct none_sample_filter : public base_filter {
   FilterType get_filter_type() const override { return FilterType::None; }
 };
 
+_RAFT_HOST_DEVICE inline none_sample_filter::none_sample_filter()  = default;
+_RAFT_HOST_DEVICE inline none_sample_filter::~none_sample_filter() = default;
```
lowener: Is there a reason to define them as default here and not in the struct?

cpp/src/neighbors/sample_filter.cuh
```diff
@@ -140,4 +146,89 @@ void bitmap_filter<bitmap_t, index_t>::to_csr(raft::resources const& handle, csr
...
+  using OuterFilterT = filtering::ivf_to_sample_filter<int64_t, InnerFilterT>;
+  OuterFilterT* p_outer_filter;
+  InnerFilterT* p_inner_filter;
+  filtering::ivf_to_sample_filter_dev h_placeholder;
+
+  ivf_to_sample_dev_wrapper(FilterType tag,
+                            const int64_t* const* inds_ptrs,
+                            InnerArgs... inner_args)
+  {
+    cudaMalloc(&p_inner_filter, sizeof(InnerFilterT));
```
lowener: Check cudaMalloc return value

cpp/src/neighbors/sample_filter.cuh
```diff
@@ -140,4 +146,89 @@ void bitmap_filter<bitmap_t, index_t>::to_csr(raft::resources const& handle, csr
...
+    init_inner_filter<<<1, 1>>>(p_inner_filter, inner_args...);
+    cudaDeviceSynchronize();
+
+    cudaMalloc(&p_outer_filter, sizeof(OuterFilterT));
+    init_outer_filter<<<1, 1>>>(p_outer_filter, inds_ptrs, p_inner_filter);
```
lowener: Can you combine both kernels into one?

cpp/src/neighbors/sample_filter.cuh
```diff
@@ -140,4 +146,89 @@ void bitmap_filter<bitmap_t, index_t>::to_csr(raft::resources const& handle, csr
...
+  OuterFilterT* p_outer_filter;
+  InnerFilterT* p_inner_filter;
+  filtering::ivf_to_sample_filter_dev h_placeholder;
+
+  ivf_to_sample_dev_wrapper(FilterType tag,
+                            const int64_t* const* inds_ptrs,
+                            InnerArgs... inner_args)
+  {
+    cudaMalloc(&p_inner_filter, sizeof(InnerFilterT));
+    init_inner_filter<<<1, 1>>>(p_inner_filter, inner_args...);
```
lowener: Launch the kernels on the current cudaStream

cpp/src/neighbors/sample_filter.cuh
```diff
@@ -140,4 +146,89 @@ void bitmap_filter<bitmap_t, index_t>::to_csr(raft::resources const& handle, csr
...
+    cudaMalloc(&p_outer_filter, sizeof(OuterFilterT));
+    init_outer_filter<<<1, 1>>>(p_outer_filter, inds_ptrs, p_inner_filter);
+    cudaDeviceSynchronize();
+
+    h_placeholder = {tag, p_outer_filter};
+  }
+
+  ~ivf_to_sample_dev_wrapper()
+  {
+    destruct_dev_filter<<<1, 1>>>(p_outer_filter);
```
lowener: Call twice the kernel (or combine it) then cudaFree both. the sync is unnecessary I think

lowener (changes requested):

cpp/include/cuvs/neighbors/common.hpp
jinsolp: From the [previous CI run](https://github.com/rapidsai/cuvs/actions/runs/16733267569/job/47366398051), I could see that the build was failing because it wouldn't let me declare + default with the `__device__` annotation. It seems like nvcc ignores `__host__ __device__` on default functions when they are defaulted at their first declaration within the struct.

cpp/src/neighbors/sample_filter.cuh
jinsolp: Shouldn't we keep the `cudaDeviceSynchronize();` because we might be calling `cudaFree` before the destructor is properly called inside the kernel?

cpp/src/neighbors/sample_filter.cuh
```diff
@@ -140,4 +146,86 @@ void bitmap_filter<bitmap_t, index_t>::to_csr(raft::resources const& handle, csr
...
+  }
+
+  filtering::ivf_to_sample_filter_dev get_dev_filter() const { return h_placeholder; }
+};
+
+using ivf_to_sample_dev_none = filtering::ivf_to_sample_dev_wrapper<filtering::none_sample_filter>;
+using ivf_to_sample_dev_bitset =
+  filtering::ivf_to_sample_dev_wrapper<filtering::bitset_filter<uint32_t, int64_t>,
+                                       cuvs::core::bitset_view<uint32_t, int64_t>>;
+
```
achirkin: The memory management here may have some performance issues and goes slightly against our design guidelines.
1) You synchronize the whole device multiple times (on cudaMalloc and then explicitly). This is going to be a bottleneck if a user doesn't cache the filter objects between the searches. I'd suggest using the pool (async) memory, packing both structures into one combined struct to do only one malloc call, and avoid the device sync at the end completely. The promise here is that the stream passed to the wrapper should be the same as during the search, so allocation/initialization is ordered by the stream.
2) If possible, try to refactor this to pass the raft::resources handle in place of cuda stream view, and use raft::device_scalar or rmm::uvector with appropriate (async device) memory resource.
3) It looks like the problem you're solving is very similar to what I do in CAGRA [dataset descriptor](https://github.com/rapidsai/cuvs/blob/6e734d5eef728c101762233c7b85a7ee8552ef4c/cpp/src/neighbors/detail/cagra/compute_distance.hpp#L221-L236). The host wrapper here manages a device-initialized structure. It does the initialization lazily to make sure the initialization kernel is only launched once, shortly before the main search kernel, and it allows caching/thread-safe reuse. Could you try to adapt this to your case?

cpp/src/neighbors/sample_filter.cuh
lowener: Yes that make sense. The sync must be done on the stream not on the device so `stream_.synchronize()` would be better because it is calling  `cudaStreamSynchronize()`.

jinsolp:
image

Thanks @achirkin @lowener  for your feedbacks! In the binary size sync today, @divyegala suggested using the enum + creating filter inside kernel approach. This is a much less complicated approach, and also shows similar perf to what we already have in this PR (launching separate kernels to track lifetime of kernels). Latest commit introduces this approach.

divyegala:
@jinsolp thanks for the changes! IMO the diff looks much cleaner now.

achirkin:
@jinsolp thanks, would you mind to benchmark this for somewhat larger values of `k`? let's say 64, 128, 256.
The kind of changes introduced in the kernel seem likely to increase the register pressure, which may lead to performance degradation when we already use a lot of registers - that is with a beefier fused top-k component.

cpp/src/neighbors/ivf_pq/ivf_pq_compute_similarity_impl.cuh
```diff
@@ -261,35 +259,44 @@ __device__ auto ivfpq_compute_score(uint32_t pq_dim,
...
+  const uint32_t* _chunk_indices,
+  const float* queries,
+  const uint32_t* index_list,
+  float* query_kths,
+  LutT* lut_scores,
+  OutT* _out_scores,
+  uint32_t* _out_indices,
+  filtering::FilterType filter_tag,
+  const int64_t* const* inds_ptrs,
+  cuda::std::optional<cuvs::core::bitset_view<uint32_t, int64_t>> filter_bitset_view)
```
achirkin: If we're going the closed-world approach, I think it would be nicer to define a sum type somewhere outside the kernel and put the selection logic in there. This would make it easier to support other filter types in future, slightly simplify the signature, and maybe even reduce the register overheads of passing all possible filter arguments like this.

cpp/src/neighbors/ivf_pq/ivf_pq_compute_similarity.cuh
```diff
@@ -166,82 +166,23 @@ auto compute_similarity_select(const cudaDeviceProp& dev_props,
     const float* queries,                                                                      \
     const uint32_t* index_list,                                                                \
     float* query_kths,                                                                         \
-    IvfSampleFilterT sample_filter,                                                            \
+    const filtering::base_filter& sample_filter,                                               \
```
cjnolet, line 166: @robertmaynard is this going to impact the update you just made for CUDA 13?

jinsolp:
Screenshot 2025-08-11 at 12 50 59 PM
@achirkin This is the result of using different k's. Some interesting observations

1. This approach is faster for a larger k (or at least I think we can say that there are no concerning drops in perf)
2. Kernel suddenly gets faster for `k=256` compared to `k=128`. I'm not sure why this is the case.

cpp/src/neighbors/ivf_pq/ivf_pq_compute_similarity_impl.cuh
jinsolp: Thanks for the feedback @achirkin : ) Is my understanding correct that you are suggesting that we use some form of struct as an argument to pass here?

achirkin: Yes, some combination of a tagged union and per-filter-type structs, e.g.:
```C++
using none_filter_type = std::monostate;
using bitset_filter_type = std::tuple<const int64_t* const*, cuda::std::optional<cuvs::core::bitset_view<uint32_t, int64_t>>>
using one_more_filter_type = <...>;

using ivf_pq_filter_type = std::variant<none_filter_type, bitset_filter_type, one_more_filter_type>;
```
I'm not sure though if you can use `std::variant` in the device code, because its destructor is not a `constexpr` in C++17 - otherwise we'd have to go with some thing like a this:
```C++
struct ivf_pq_filter_type {
  filtering::FilterType tag;
  union ivf_pq_filter_variant {
    none_filter_type none_filter;
    bitset_filter_type bitset_filter;
    one_more_filter_type one_more_filter;
  } value;

  constexpr auto filter(uint32_t query_ix, ...) const -> bool {
      switch (tag) {
         case ...:
           return <... encapsulated filtering logic ...>;
      }
  }
};
```
In any case the idea is to (1) keep the filtering logic in a single, separate place, (2) re-use the same registers for different arguments depending on the filter type.

jinsolp: Thanks for the detailed explanation! Latest commits include this change.

cpp/src/neighbors/sample_filter.cuh
```diff
@@ -140,4 +146,48 @@ void bitmap_filter<bitmap_t, index_t>::to_csr(raft::resources const& handle, csr
...
+struct ivf_filter_dev {
+  filtering::FilterType tag_;
+
+  union ivf_filter_dev_args_variant {
+    none_filter_args_t none_filter_args;
+    bitset_filter_args_t bitset_filter_args;
+
+    _RAFT_HOST_DEVICE ivf_filter_dev_args_variant() {}
+    _RAFT_HOST_DEVICE ~ivf_filter_dev_args_variant() {}
```
achirkin: Can we get away with no explicitly defined constructor/destructor? I have a feeling this looks more complicated than should :)

cpp/src/neighbors/sample_filter.cuh
```diff
@@ -140,4 +146,48 @@ void bitmap_filter<bitmap_t, index_t>::to_csr(raft::resources const& handle, csr
...
+  } args_;
+
+  _RAFT_HOST_DEVICE ivf_filter_dev(none_filter_args_t args = {}) : tag_(FilterType::None)
+  {
+    new (&args_.none_filter_args) none_filter_args_t(args);
```
achirkin: Can't we just initialize using the `args_(args)` same as you initialize the `tag_` above (after you remove explicit constructor/destructor)?

cpp/src/neighbors/sample_filter.cuh
```diff
@@ -140,4 +146,48 @@ void bitmap_filter<bitmap_t, index_t>::to_csr(raft::resources const& handle, csr
...
+  constexpr inline _RAFT_HOST_DEVICE bool operator()(const uint32_t query_ix,
+                                                     const uint32_t cluster_ix,
+                                                     const uint32_t sample_ix)
+  {
+    switch (tag_) {
+      case FilterType::None:
+        return filtering::none_sample_filter{}(query_ix, cluster_ix, sample_ix);
+      case FilterType::Bitset: {
+        auto& inds_ptrs_   = std::get<0>(args_.bitset_filter_args);
+        auto& bitset_view_ = std::get<1>(args_.bitset_filter_args);
```
achirkin: A nitpick:
```suggestion
        auto& [inds_ptrs_, bitset_view_] = args_.bitset_filter_args;
```

cpp/src/neighbors/sample_filter.cuh
divyegala: +1

cpp/src/neighbors/sample_filter.cuh
```diff
@@ -140,4 +146,48 @@ void bitmap_filter<bitmap_t, index_t>::to_csr(raft::resources const& handle, csr
...
+  {
+    new (&args_.none_filter_args) none_filter_args_t(args);
+  }
+
+  _RAFT_HOST_DEVICE ivf_filter_dev(const bitset_filter_args_t& args) : tag_(FilterType::Bitset)
+  {
+    new (&args_.bitset_filter_args) bitset_filter_args_t(args);
+  }
+
+  constexpr inline _RAFT_HOST_DEVICE bool operator()(const uint32_t query_ix,
```
divyegala: Just in case
```suggestion
  constexpr __forceinline__ _RAFT_HOST_DEVICE bool operator()(const uint32_t query_ix,
```

cpp/src/neighbors/sample_filter.cuh
jinsolp: The compiler complains if I remove these explicit constructor/destructor
```
error: the default constructor of "cuvs::neighbors::filtering::ivf_filter_dev::ivf_filter_dev_args_variant" cannot be referenced -- it is a deleted function
```

Looks like this is because the `bitset_filter_args_t` is **not** trivially constructible. Is there a way to workaround this?

divyegala: What line is this error occurring on?

jinsolp: Happens when I get rid of the explicit constructor/destructor for the union (these two lines)

```
_RAFT_HOST_DEVICE ivf_filter_dev_args_variant() {}
_RAFT_HOST_DEVICE ~ivf_filter_dev_args_variant() {}
```

divyegala: Can you report what line the compiler says the error occurs on? Must be somewhere during construction of this struct?

jinsolp: line 159 and 164 of `sample_filter.cuh`.  That would be the `ivf_filter_dev` constructors.

```
sample_filter.cuh(159): error: the default constructor of "cuvs::neighbors::filtering::ivf_filter_dev::ivf_filter_dev_args_variant" cannot be referenced -- it is a deleted function
    {
    ^

sample_filter.cuh(164): error: the default constructor of "cuvs::neighbors::filtering::ivf_filter_dev::ivf_filter_dev_args_variant" cannot be referenced -- it is a deleted function
    {
    ^
```

divyegala: Artem's comment below is relevant to this. If you simplify the construction it should be possible.

cpp/src/neighbors/sample_filter.cuh
jinsolp: @achirkin @divyegala 
Changed to use the initializer list. We still need constructors in the union because `bitset_filter_args_t` is not trivially constructible and therefore does not have a default constructor. But we don't need the destructor anymore.

cpp/src/neighbors/sample_filter.cuh
```diff
@@ -140,4 +140,49 @@ void bitmap_filter<bitmap_t, index_t>::to_csr(raft::resources const& handle, csr
   raft::sparse::convert::bitmap_to_csr(handle, bitmap_view_, csr);
 }
 
+struct none_filter_args_t {};
+using bitset_filter_args_t =
+  std::tuple<const int64_t* const*, cuvs::core::bitset_view<uint32_t, int64_t>>;
```
lowener, line 145: Can this be tuple<const int64_t* const*, filtering::bitset_filter<uint32_t, int64_t>> to construct the bitset_filter object just once instead of every call to operator()?

jinsolp: Unfortunately, we have problems in each of the following scenarios if we want to do so;
If we were to make the filter once...
- outside the `compute_similairty_kernel`: we have to launch separate kernels for the filter construction, which complicates things
- inside the `compute_similairty_kernel`: we don't want to dynamic dispatch inside the kernel, so it is difficult to share common code.

PR #1215, opened 2025-08-05
[Java] Exception-safe RMM Allocations

mythrocks:
This commit introduces exception safety for RMM allocations.

Previously, device memory allocated through `cuvsRmmAlloc()` was freed manually using `cuvsRmmFree()`, in all the index impl classes.  The problem there is that if an exception is thrown in the intervening time between alloc and free, it would lead to a leak of device memory.

This commit extends the `CloseableHandle` class to encapsulate the allocation of device memory.  This new class is used in try-with-resources blocks, to make device memory allocations exception-safe.


java/cuvs-java/src/test/java/com/nvidia/cuvs/CagraMultiThreadStabilityIT.java
```diff
@@ -86,7 +86,8 @@ public void testQueryingUsingMultipleThreads() throws Throwable {
 
       log.info("CAGRA index created, starting high-contention multi-threaded search...");
 
-      // Create high contention scenario that would fail without using separate resources in every thread
+      // Create high contention scenario that would fail without using separate resources in every
+      // thread
```
mythrocks: I think these changes just came in from running the formatter.  These are unrelated to the task at hand.

chatman:
+1, very important change.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/BruteForceIndexImpl.java
```diff
@@ -158,25 +159,28 @@ private IndexReference build(
...
+        checkCuVSError(returnValue, "cuvsBruteForceBuild");
 
-      returnValue = cuvsStreamSync(cuvsResources);
-      checkCuVSError(returnValue, "cuvsStreamSync");
+        returnValue = cuvsStreamSync(cuvsResources);
+        checkCuVSError(returnValue, "cuvsStreamSync");
 
-      return new IndexReference(datasetMemorySegmentP, datasetBytes, tensorDataArena, index);
+        closeableDataMemorySegmentP.release();
+        return new IndexReference(datasetMemorySegmentP, datasetBytes, tensorDataArena, index);
```
ldematte: I see a problem here: we pass the dataset "pointer" to IndexReference to hold it and clean it when we are done with the index (see destroyIndex()), so this will lead to a double free. 
Your change is good only if we are able to determine that we don't need the dataset device memory after we built the index; however, I was not able to determine if we need it or not; I think we might need it, as it might not be copied over again.

ldematte: Ah, I see now why you need "release": it's a way to work around that. 
I think it is better to avoid it, and simply avoid to use try with resources here.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/common/CloseableRMMAllocation.java
```diff
@@ -0,0 +1,63 @@
...
+    this.numBytes = numBytes;
+    this.pointer = pointer;
+  }
+
+  @Override
+  public MemorySegment handle() {
+    return pointer;
+  }
+
+  public MemorySegment release() {
```
ldematte: See above comment: I think it's better to not have "release", we can just avoid to use try (and avoid calling close) when not needed.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/common/Util.java
```diff
@@ -352,14 +352,15 @@ public static MemorySegment prepareTensor(
     return tensor;
   }
 
-  public static MemorySegment allocateRMMSegment(long resourceHandle, long datasetBytes) {
+  public static CloseableRMMAllocation allocateRMMSegment(long resourceHandle, long datasetBytes) {
```
ldematte: What about moving this to CloseableRMMAllocation? (and make the CloseableRMMAllocation ctor private?)

ldematte (commented):
It's a good idea! It might interfere with work I'm doing on the device CuVSMatrix, but I'll take the duty of merging them. 
Left a couple of comments.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/BruteForceIndexImpl.java
mythrocks: I think I haven't conveyed the utility of `CloseableRMMAllocation` properly. 

## Simple case (`CagraIndexImpl`, etc.)

Consider the following simplified (if contrived) case, representative of how `allocRMMSegment()` is used in `CagraIndexImpl`, `TieredIndexImpl`, etc.:
```java
// search()
{
  var queriesDP = allocateRMMSegment(...);
  // ...
  cudaMemcpy(...); // Can throw.
  checkCuVSError( cuvsCagraSearch(...) ); // Can also throw.
  // ...
  // And finally.
  cuvsRmmFree( queriesDP );
}
```
There are several throwable points between the `alloc()` and the `free()`.  If any of them fire, `queriesDP` is leaked in `__device__` memory.  This is the simple case that `CloseableRMMAllocation` addresses.

## The case for `.release()` (`BruteForceIndexImpl`)

Similar example as above, except that the `Index` *adopts* the allocation, and holds it until `destroyIndex()` is called.
```java
// build()
{
  var datasetMemorySegmentP = allocateRMMSegment(...);
  // ...
  cudaMemcpy(...); // Can throw.
  checkCuVSError( cuvsBruteForceBuild(...) ); // Can also throw.
  // ...
  // And finally, commit.  No free().
  return new IndexReference( datasetMemorySegmentP, ... );
}
```
All the perils of the first example appear here as well;  there are many throwable points between the allocation and the creation of the `IndexReference`.  We need a way to clean up the memory allocation if there's any throw before the final commit (to `IndexReference`).   
This is why we `release()` right before the return.

Not using throw-with-resources would mean that we're still open to `__device__` memory leaks.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/common/Util.java
mythrocks: Good idea.  I'll check the call-sites to make sure this is viable.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/common/CloseableRMMAllocation.java
mythrocks: [Counterpoint.](https://github.com/rapidsai/cuvs/pull/1215#discussion_r2255077242)

mythrocks:
Hmm... There seems to be a logic error in `CagraIndexImpl::search()`.  I'm debugging it right now.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/common/Util.java
mythrocks: (This change is still pending.)

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/BruteForceIndexImpl.java
ldematte: Ah, I see. Still, I feel like `release()` is not clear; a more explicit way would be to do an old, simple try/catch:
```
var closeableDataMemorySegment = allocateRMMSegment(cuvsResources, datasetBytes));
try {
   MemorySegment datasetMemorySegment = closeableDataMemorySegment.handle();
   // use handle
   // here ownership is "transferred"
   return new IndexReference(closeableDataMemorySegment, ... );
} catch (Throwable t) {
   closeableDataMemorySegment.close();
   throw;
}
```
let me see if I can think of a different pattern.

ldematte: I think what you are trying can be seen as similar to C++ `unique_ptr`; the difference is that we don't have move semantics in Java (as a simple way to transfer ownership).
An alternative could be to implement `release()` closer to what C++ has, e.g. returning the enclosed object:
```
try (var closeableDataMemorySegment = allocateRMMSegment(cuvsResources, datasetBytes)) {
   MemorySegment datasetMemorySegment = closeableDataMemorySegment.handle();
   // use handle
   // here ownership is "transferred" more explicitly
   return new IndexReference(
         new CloseableRMMAllocation(datasetMemorySegment.release()), // release it, return the "raw pointer", pass it immediately to another `CloseableRMMAllocation` that will handle its lifetime
         ... );
}
```
While I think this models more clearly the "I am transferring ownership" idea, there is one drawback: what if IndexReference or CloseableRMMAllocation ctors throw? This is not a danger here, they are both no-throw ctors (just simple assignements), but still it's a little bit less robust. 

I think I still like the explicit try/catch more.

ldematte: I want to highlight this though:
>  If any of them fire, queriesDP is leaked in `__device__` memory

Leaking device memory seems particularly bad; but how bad is it actually? Like, would be leaking it even after the process is gone, or the OS/device driver will be able to reclaim that memory (like for host memory)?

In any case, calling @ChrisHegarty in to see if we can protect further against this possibility (using cleaners maybe?)

mythrocks: > Similar to C++ `unique_ptr`...

Exactly right.  That's what this is modeled around.  This will turn into an RAII wrapper after I have moved the allocation into the constructor. 

That's also the pattern we use in https://github.com/rapidsai/cudf, when we transfer ownership of the underlying `__device__` memory in a  CUDF column. 

~~My initial version also had the pointer returned from release.~~ 

Edit:  Looks like it's not just my initial version;  `release()` does currently return the old pointer when relinquishing the memory.  I didn't use it in `BruteForceIndexImpl` because the original pointer was already at hand in the same scope.  This will get tighter once the RAII change is made.

mythrocks: > would be leaking it even after the process is gone, or the OS/device driver will be able to reclaim that memory (like for host memory)?

No, the leak does not persist beyond the lifetime of the process.  That should be reclaimed after the process exits, yes.

But I would like not to make an assumption that the users of `cuvs-java` are short-running processes, and require them to be tolerant of leaking memory.  Even if current users might be alright with a leak in exceptional events, a future user might be a long-running application.

java/cuvs-java/src/test/java/com/nvidia/cuvs/CagraMultiThreadStabilityIT.java
mythrocks: Nope, this was the result of a botched `git merge`.  This has been resolved now.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/common/Util.java
mythrocks: The latest commit addresses this.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/common/CloseableRMMAllocation.java
mythrocks: Closing this thread.  We can carry on this discussion on the [main thread](https://github.com/rapidsai/cuvs/pull/1215#discussion_r2253915198).

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/BruteForceIndexImpl.java
ldematte: > I would like not to make an assumption that the users of cuvs-java are short-running processes

Definitely not, I was just trying to understand what's the worst case scenario here, and if we need something extra, beyond taking care of exceptions.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/BruteForceIndexImpl.java
```diff
@@ -158,25 +159,28 @@ private IndexReference build(
...
+        checkCuVSError(returnValue, "cuvsBruteForceBuild");
 
-      returnValue = cuvsStreamSync(cuvsResources);
-      checkCuVSError(returnValue, "cuvsStreamSync");
+        returnValue = cuvsStreamSync(cuvsResources);
+        checkCuVSError(returnValue, "cuvsStreamSync");
 
-      return new IndexReference(datasetMemorySegmentP, datasetBytes, tensorDataArena, index);
+        return new IndexReference(
+            closeableDataMemorySegmentP.release(), datasetBytes, tensorDataArena, index);
```
ldematte: I still lean more towards an explicit try-catch --  IDK, this pattern looks a bit strange in Java. I understand it because I have a C/C++ background, but I bet it's not immediate for "pure Java" folks. Then again, people that are going to work on this library would need to have some C/C++ knowledge, so...

like this much better than the original though; I would even go an extra mile here (optional and/or good for a follow-up PR) and make `IndexReference` taking a `CloseableRMMAllocation` as the first param (, changing/moving the logic of cleaning up in `destroyIndex` so it calls `close` on the `CloseableRMMAllocation`.

ldematte: If you do this, you can even avoid exposing release:
```
IndexReference(
    new CloseableRMMAllocation(closeableDataMemorySegmentP), // note that we have a "copy-ctor" here, not calling release
    datasetBytes, 
    tensorDataArena, 
    index);

```
and in `CloseableRMMAllocation`
```
public CloseableRMMAllocation(CloseableRMMAllocation other) {
   var rawPointer = other.release(); // release can be private! or inlined here
   // copy other private members as needed too
}
```
So you cannot just release a raw pointer, you need to transfer it to another `CloseableRMMAllocation`, making it explicit you need to `close` it (and Intellij should help you with warnings too if you don't do that).

Wdyt?

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/CagraIndexImpl.java
```diff
@@ -235,100 +236,92 @@ public SearchResults search(CagraQuery query) throws Throwable {
...
       MemorySegment floatsSeg = buildMemorySegment(localArena, query.getQueryVectors());
 
-      long queriesBytes = C_FLOAT_BYTE_SIZE * numQueries * vectorDimension;
-      long neighborsBytes = C_INT_BYTE_SIZE * numQueries * topK;
-      long distancesBytes = C_FLOAT_BYTE_SIZE * numQueries * topK;
+      final long queriesBytes = C_FLOAT_BYTE_SIZE * numQueries * vectorDimension;
+      final long neighborsBytes = C_INT_BYTE_SIZE * numQueries * topK;
+      final long distancesBytes = C_FLOAT_BYTE_SIZE * numQueries * topK;
+      final boolean hasPreFilter = query.getPrefilter() != null;
+      final BitSet[] prefilters =
```
ldematte, line 245: I very much like how this change has forced us to restructure the code and make these final

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/CagraIndexImpl.java
```diff
@@ -235,100 +236,92 @@ public SearchResults search(CagraQuery query) throws Throwable {
...
+          hasPreFilter ? new BitSet[] {query.getPrefilter()} : new BitSet[0];
```
ldematte: Nit: since we are not going to use it, BitSet[0] here can be a const (private static final), so we avoid one allocation we don't need.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/TieredIndexImpl.java
```diff
@@ -134,35 +135,33 @@ private IndexReference build() {
...
+          long[] datasetShape = {rows, cols};
+          MemorySegment datasetTensor =
+              prepareTensor(
+                  localArena, datasetDP.handle(), datasetShape, kDLFloat(), 32, kDLCUDA(), 1);
 
-        // Create tensor from device memory
-        long[] datasetShape = {rows, cols};
-        MemorySegment datasetTensor =
-            prepareTensor(localArena, datasetDP, datasetShape, kDLFloat(), 32, kDLCUDA(), 1);
+          MemorySegment index = localArena.allocate(cuvsTieredIndex_t);
```
ldematte, line 148: Unrelated to this PR, but I'm conflicted about using cuvsTieredIndex_t, since we need a POINTER here (as we see below, when we are extracting a C_POINTER).
Right now, cuvsTieredIndex_t is a pointer, but if that changes, we would end up allocating e.g. a struct, and use it to store (and retrieve) a pointer...

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/TieredIndexImpl.java
```diff
@@ -199,113 +198,123 @@ public SearchResults search(TieredIndexQuery query) throws Throwable {
...
-        final MemorySegment prefilterDP;
+        final long queriesBytes = C_FLOAT_BYTE_SIZE * numQueries * vectorDimension;
+        final long neighborsBytes = C_LONG_BYTE_SIZE * numQueries * topK; // 64-bit for tiered index
+        final long distancesBytes = C_FLOAT_BYTE_SIZE * numQueries * topK;
+        final boolean hasPreFilters = query.getPrefilter() != null;
         final long prefilterBytes;
-
-        if (query.getPrefilter() != null) {
+        if (hasPreFilters) {
           BitSet[] prefilters = new BitSet[] {query.getPrefilter()};
```
ldematte: The creation of these arrays here looks redundant, all we need here is to find `prefilterBytes`, which can be 
```
long prefilterDataLength = query.getNumDocs(); // length is 1
long prefilterLen = (prefilterDataLength + 31) / 32;
prefilterBytes = C_INT_BYTE_SIZE * prefilterLen;
```

ldematte: If you want to avoid recomputing them below, you can extract `BitSet[] prefilters` and `long[] prefilterShape` to a final variable here like you did above

ldematte (approved):
Looks good, I left a couple of optional comments (for this PR or for a follow-up)

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/TieredIndexImpl.java
mythrocks: > The creation of these arrays here looks redundant, all we need here is to find `prefilterBytes`...

I think I worked on `TieredIndexImpl` before the rest.  `CagraIndexImpl` certainly evolved a fair bit further, since it came towards the end.

These are great suggestions.  I'll get to work on these, and update shortly.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/TieredIndexImpl.java
mythrocks: > Right now, `cuvsTieredIndex_t` is a pointer, but if that changes...

I think that could apply to any of the typedefs exposed in the C++ API.  Changing those structures would be a non-trivial event.  I would think @benfred and gang are likely to keep those stable. :]

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/BruteForceIndexImpl.java
mythrocks: You're right.  This does seem to clean up a fair bit of code, as well.

Working on the rest of your suggestions now.

mythrocks:
@ldematte:  I've addressed the remainder of your concerns.  That copy-constructor suggestion helped the code DRY a bit more.

Does this look more agreeable?

ldematte (approved):
Was already good to me, but now it looks even better I think. Thanks for sticking with me and implementing my suggestions!

mythrocks:
~~/merge~~

@cjnolet, any chance you might approve this change?  @ldematte has approved it already.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/CagraIndexImpl.java
```diff
@@ -235,100 +238,92 @@ public SearchResults search(CagraQuery query) throws Throwable {
...
-          checkCuVSError(returnValue, "cuvsRMMFree");
+        try (var queriesDP = allocateRMMSegment(cuvsRes, queriesBytes);
+            var neighborsDP = allocateRMMSegment(cuvsRes, neighborsBytes);
+            var distancesDP = allocateRMMSegment(cuvsRes, distancesBytes);
+            var prefilterDP =
+                hasPreFilter
+                    ? allocateRMMSegment(cuvsRes, prefilterBytes)
+                    : CloseableRMMAllocation.EMPTY) {
+
+          cudaMemcpy(queriesDP.handle(), floatsSeg, queriesBytes, INFER_DIRECTION);
```
cjnolet, line 262: Not something to address in this PR, but we should really be using `cudaMemcpyAsync` everywhere (and accepting a stream / cuvs resources. This is going to synchronize the whole device and thatll slow things down immensely. I suspect this could be why @chatman and team have seen that multi-threaded search reduces perf significantly.

cjnolet (approved):
The change intended in this PR LGTM. We should address the cudaMemcpy -> cudaMemCpyAsync ASAP, though, because it will affect concurrency / perf drastically. We reallys hould go through the whole C/Java API and repalce cudaMemcpy w/ cudaMemCpyAsync everywhere (and accept `cuvsresources` everywhere needed.

ldematte:
> The change intended in this PR LGTM. We should address the cudaMemcpy -> cudaMemCpyAsync ASAP, though, because it will affect concurrency / perf drastically. We reallys hould go through the whole C/Java API and repalce cudaMemcpy w/ cudaMemCpyAsync everywhere (and accept cuvsresources everywhere needed).

@cjnolet a couple of question about this to cement my understanding: `cudaMemcpyAsync` is "async" in 2 ways: it does not block the caller (the CPU thread calling it), and it _can_ allow for overlapping copy/computation (if it uses a non-0 stream parameter).

- If we want the overlapping copy/computation but we need to be sure the operation is finished before proceeding (e.g. copied out the data) with the CPU side of things, we need to pass a non-0 `stream` and then "block" the caller (CPU thread) with a `cudaStreamSynchronize(stream)`, correct? Operations on the device will still overlap.
- If we want overlapping copy/computation with cuvs operations we need to use a _different_ stream, right? We cannot use `cuvsGetStream()`/`raft::resources::get_cuda_stream()` for the copy too, or it will use the same stream it uses for e.g. indexing, and we will be unable to overlap. So, different resources (or different stream, manually managed), correct?

ldematte:
I still have to benchmark it, but this is where I am leaning with `CuVSDeviceMatrix`: create (and manage) a separate `stream` for copy operations via `cudaMemcpyAsync`. The additional reason is that we also won't need to keep a reference to resources, which might be problematic (scope, threading), but it is my understanding that having a different stream it's also for the best performance wise.

cjnolet:
> If we want the overlapping copy/computation but we need to be sure the operation is finished before proceeding (e.g. copied out the data) with the CPU side of things, we need to pass a non-0 stream and then "block" the caller (CPU thread) with a cudaStreamSynchronize(stream), correct? Operations on the device will still overlap.

Even worse- `cudaMemcpy` blocks the whole device! So imagine having multiple threads, each performing search() with different streams, they won't be able to overlap because `cudaMemCpy` will block the entire device each time it's called until the memory has been fully copied. This is why we NEVER use `cudaMemCpy` and we always use `cudaMemCpyAsync`. 

> If we want overlapping copy/computation with cuvs operations we need to use a different stream, right? We cannot use cuvsGetStream()/raft::resources::get_cuda_stream() for the copy too, or it will use the same stream it uses for e.g. indexing, and we will be unable to overlap. So, different resources (or different stream, manually managed), correct?

A stream is necesary to queue up the memory copy, but this is still also very important because the goal here is to queue up work (if even on the same stream) and not block the client so that the GPU can work off the queue asynchronously. Of course, if you want to overlap multiple memory copies and kernel launches, the cuvsresources can also be configured by the user to provide multiple "worker streams" which can be used to perform these copies asynchronously and independently. 

In order to take advanage of any of this, we need to use cudaMemCpyAsync- it's not a question of whether we should use it. We need to use it everywhere.

cjnolet:
> If we want overlapping copy/computation with cuvs operations we need to use a different stream, right?

I mentioned this above, but want to expand it just to make it fully clear- asynchronous computation even with the same stream is just as important to queue up and synchronize as little as possible. Queueing up work on a stream (such as a memory copy, memory allocation/free, or launching a kernel) has very little overhead, if any. But if we have to synchronize the cpu thread every time we queue up one of these operations, we're going to insert blocking gaps in the pipeline and these can be very costly in latency sensitive workloads (You can use nsight-systems application to profile this, you'll literally see the gaps in the pipeline). Insteas, we utilize the asynchronous nonblocking behavior as much as possible so the CPU can continue queuing up work for the gpu independent of the thread that's making the calls to the runtime API. THis is also why we only explicitly synchronize the stream in the cuVS C++/C layer when it's absolutely necessary (which in general is either because we just did a device to host copy and need to immediately read the memory on host, or if we are about to use multiple streams to overlap some operations, we synchronize the main stream before splitting off into a series of "worker streams" for things like overlapping computation / memory copies as you mentioned).

cjnolet:
> The additional reason is that we also won't need to keep a reference to resources, which might be problematic (scope, threading), but it is my understanding it's also for the best performance wise.

resources are expensive to create but the individual resources that get stored on the cuvsresources instance get created lazily as algorithms need them to try and amortize the cost there. I hope we're not saying we are recreating these resource objects often. Ideally we should create them once at the beginning and reuse them in subsequent calls. They don't have a high memory cost, but things like cublas handles can be very costly to create and that's used very often.

mythrocks:
I'll address the `cudaMemcpyAsync` change in a follow-on, shortly.

Thank you for the reviews, @ldematte, @cjnolet.

ldematte:
> In order to take advantage of any of this, we need to use cudaMemCpyAsync- it's not a question of whether we should use it. We need to use it everywhere.

Absolutely; like you mentioned, even with the same stream + synchronize on the CPU thread this is still better (not blocking the device -- other threads can still make progress)

PR #1216, opened 2025-08-05
[Java] Add CAGRA index graph accessor/build from graph (host memory)

ldematte:
This PR leverages the functions introduced by #1086 and the data structures introduced by #1111 to access, copy, and re-create an index to/from a CAGRA graph.

Supersedes #1105


ldematte:
@mythrocks did you have a chance to look into this? Any comments?

mythrocks:
> @mythrocks did you have a chance to look into this? Any comments?

Sorry, will do shortly.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/CuVSMatrixBaseImpl.java
```diff
@@ -45,4 +48,21 @@ public long columns() {
...
+      case BYTE -> 8;
+    };
+  }
+
+  protected int code() {
+    return switch (dataType) {
+      case FLOAT -> kDLFloat();
+      case INT -> kDLInt();
+      case UINT, BYTE -> kDLUInt();
+    };
```
mythrocks, line 64: 👍

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/CuVSHostMatrixImpl.java
```diff
@@ -131,6 +130,12 @@ public ValueLayout valueLayout() {
     return valueLayout;
   }
 
+  @Override
+  public MemorySegment toTensor(Arena arena) {
```
mythrocks, line 134: This is a clean way to do it:  Host matrices having a method to convert to the appropriate tensor.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/CagraIndexImpl.java
```diff
@@ -386,6 +414,81 @@ public void serialize(OutputStream outputStream, Path tempFile, int bufferLength
...
+          cuvsCagraIndexGetGraphDegree(cagraIndexReference.getMemorySegment(), outPtr),
+          "cuvsCagraIndexGetGraphDegree");
+      int graphDegree = outPtr.get(__uint32_t, 0);
+
+      checkCuVSError(
+          cuvsCagraIndexGetSize(cagraIndexReference.getMemorySegment(), outPtr),
+          "cuvsCagraIndexGetGraphDegree");
+      int size = outPtr.get(__uint32_t, 0);
+
+      // TODO: use a "device" graph + tensor, avoid (defer) copy
```
mythrocks, line 427: 👍

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/CagraIndexImpl.java
```diff
@@ -386,6 +414,81 @@ public void serialize(OutputStream outputStream, Path tempFile, int bufferLength
...
+    try (var localArena = Arena.ofConfined()) {
+      var outPtr = localArena.allocate(__uint32_t);
+      checkCuVSError(
+          cuvsCagraIndexGetGraphDegree(cagraIndexReference.getMemorySegment(), outPtr),
+          "cuvsCagraIndexGetGraphDegree");
+      int graphDegree = outPtr.get(__uint32_t, 0);
+
+      checkCuVSError(
+          cuvsCagraIndexGetSize(cagraIndexReference.getMemorySegment(), outPtr),
+          "cuvsCagraIndexGetGraphDegree");
```
mythrocks:
```suggestion
          "cuvsCagraIndexGetGraphGetSize");
```

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/CagraIndexImpl.java
```diff
@@ -386,6 +414,81 @@ public void serialize(OutputStream outputStream, Path tempFile, int bufferLength
...
+      int size = outPtr.get(__uint32_t, 0);
```
mythrocks: This part does have me worried.  I've been wondering about this for a bit (excuse the pun.)

The size type, per the C++ API, is an unsigned type.  We're casting to a signed type.  In these situations, shouldn't we at least check that the number of vectors does not exceed SIGNED_INT_MAX?

mythrocks: Other projects that I've worked on bump up to `long`, in these situations.  This might be challenging, given that the JNI interfaces are generated code.

mythrocks: Note:  I doubt the `degree` would strain the limits of `uint32_t`, but the size would.

mythrocks (commented):
A couple of minor nits, but this looks good to me, mostly.  We'll need to discuss how to handle the case where the number of rows approaches max-int.

This needs to be rebased. (Sorry! #1215.)

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/CagraIndexImpl.java
ldematte: Yeah, Java does not have unsigned and jextract is basically ignoring that, mapping everything to the signed types, as you can see here.
We can bump up to long, even if I think it will not be straightforward (we do need to get 4 bytes out, so it will be a int, and then treat the int bits logically ("verbatim") -- no direct cast to long, that would be arithmetical and extend the sign, which we don't want). Some math will be needed I think, let me think about it a bit.

ldematte:
@mythrocks rebased and addressed comments, thanks!

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/CagraIndexImpl.java
mythrocks: I see this is marked as resolved.  ~~We can revisit this afterwards.  I'll check how viable it would be to change this to a signed type in C++.~~
Thank you for addressing the up-cast to `long`.

mythrocks (approved):
LGTM.  We can address the graph-size problem as a follow-on.

PR #1217, opened 2025-08-05
Add NN.extend and kmeans to Python doc

lowener:
All the nearest neighbors `extend` functions were missing.


cjnolet (approved):
Thanks for finding and fixing this!

PR #1219, opened 2025-08-05
Updates needed when building with CUDA 13

robertmaynard:
cuvs also requires https://github.com/rapidsai/raft/pull/2774 to be merged


cjnolet (approved):

cpp/include/cuvs/neighbors/common.hpp
```diff
@@ -485,7 +485,7 @@ namespace filtering {
 enum class FilterType { None, Bitmap, Bitset };
 
 struct base_filter {
-  virtual ~base_filter()                     = default;
+  ~base_filter()                             = default;
```
mythrocks, line 488: Likely a silly question, but wouldn't this cause leaks or incorrect behaviour, if `base_filter` instances are deleted?

robertmaynard: I didn't see any derived types that had local storage. If we have any derived types that do have storage they would still be okay as long as they are deleted as the derived type. ( So no holding a vector of base_filter pointers and calling delete on that ).

mythrocks: Agreed.  I couldn't find that sort of thing either. 

Thank you.

PR #1220, opened 2025-08-05
modify ivf-flat example to print vectors

tfeher:


examples/cpp/src/ivf_flat_example.cu
```diff
@@ -39,7 +39,7 @@ void ivf_flat_build_search_simple(raft::device_resources const& dev_resources,
   using namespace cuvs::neighbors;
 
   ivf_flat::index_params index_params;
-  index_params.n_lists                  = 1024;
+  index_params.n_lists                  = 100;
```
cjnolet, line 38: Was this meant to be reduced or was it just for debugging / running locally?

examples/cpp/src/ivf_flat_example.cu
```diff
@@ -148,8 +173,8 @@ int main()
                                raft::make_const_mdspan(dataset.view()),
                                raft::make_const_mdspan(queries.view()));
 
-  // Build and extend example.
-  ivf_flat_build_extend_search(dev_resources,
-                               raft::make_const_mdspan(dataset.view()),
-                               raft::make_const_mdspan(queries.view()));
+  // // Build and extend example.
+  // ivf_flat_build_extend_search(dev_resources,
```
cjnolet, line 170: Should this be uncommented before the PR is merged?

PR #1224, opened 2025-08-06
Fix missing sync_stream in ScaNN build

rmaschal:
As title. For larger number of clusters (or faster PCI-E), the work required to find cluster assignments would take longer than the HtoD copy for each batch. This would lead to overwriting of one of the batch_load_iterator device buffers before kmeans predict completed, and thus lower recall.

This adds a raft::resource::sync_stream at the end of the kmeans predict loop to avoid this. Although not technically necessary (as the prior DtoH copy into pageable memory is synchronous), sync_stream is also added to the end of the quantization loop for consistency/correctness. 

Also removed one unused comment.

### DIFF
--- a/cpp/src/neighbors/scann/detail/scann_build.cuh
+++ b/cpp/src/neighbors/scann/detail/scann_build.cuh
@@ -173,6 +173,9 @@ index<T, IdxT> build(
       res, kmeans_params, batch_view, raft::make_const_mdspan(centroids_view), batch_labels_view);
 
     dataset_vec_batches.prefetch_next_batch();
+
+    // Make sure work on device is finished before swapping buffers
+    raft::resource::sync_stream(res);
   }
 
   // AVQ update of KMeans centroids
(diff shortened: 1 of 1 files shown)


cjnolet (approved):

cpp/src/neighbors/scann/detail/scann_build.cuh
```diff
@@ -173,6 +173,9 @@ index<T, IdxT> build(
       res, kmeans_params, batch_view, raft::make_const_mdspan(centroids_view), batch_labels_view);
 
     dataset_vec_batches.prefetch_next_batch();
+
+    // Make sure work on device is finished before swapping buffers
+    raft::resource::sync_stream(res);
```
tarang-jain, line 178: prefetch_next_batch seems to be [synchronizing at the very end](https://github.com/rapidsai/cuvs/blob/46760d444567347a5ab132f6dd03246b85a0aee3/cpp/src/neighbors/detail/ann_utils.cuh#L550). Would it make better sense to have the sync_stream before prefetching the batch (above line 175)?

rmaschal: prefetch_next_batch() only syncs on the stream used for copying, so if the copy happens on a separate stream from the work (e.g. if you want to prfetch/overlap) then work might not be done when buffers are swapped (which is what I am trying to fix). Moving sync stream before prefetch will fix the problem, but at the cost of no overlapping of work/copy, reducing performance.

tarang-jain: Good point. I had not realized that prefetching is done using a different stream

PR #1226, opened 2025-08-06
Reduce noise in Java tests

mythrocks:
When the Java tests run currently, they produce a large amount of logs to the console.  This includes contents of the test datasets, and various messages that ought to be trace/debug messages.

This commit changes a lot of these logs to trace/debug messages, to reduce the console log noise.  

Note that this doesn't completely eliminate unnecessary noise.  There will need to be follow-up changes to `libcuvs` itself.  We might also need to investigate some of the warnings that were previously obscured by the log volume.


chatman:
On similar lines, I think we should configure some log4j settings to show warn and errors in red, rest in black. It will aid readability.

cjnolet (approved):

PR #1230, opened 2025-08-08
[BUG] NEIGHBORS_ALL_NEIGHBORS_TEST build ignores --no-mg

enp1s0:
The following error occurs when building the unit tests on a machine where NCCL is not installed, even though the `--no-mg` option is specified.
```
CMake Error at tests/CMakeLists.txt:47 (target_link_libraries):
  Target "NEIGHBORS_ALL_NEIGHBORS_TEST" links to:

    NCCL::NCCL

  but the target was not found.  Possible reasons include:

    * There is a typo in the target name.
    * A find_package call is missing for an IMPORTED target.
    * An ALIAS target is missing.

Call Stack (most recent call first):
  tests/CMakeLists.txt:213 (ConfigureTest)
```

This is because `NCCL::NCCL` is specified in tests/CmakeLists.txt
https://github.com/rapidsai/cuvs/blob/afc24ee8302a57a5c7192a38fc2d74a1a47aa95f/cpp/tests/CMakeLists.txt#L213-L216

I could build the unit tests without the NCCL dependency. Is this dependency really necessary even when the `--no-mg` option is specified?
This PR modifies the CMakeLists.txt file so that the NCCL dependency is set only when the `--no-mg` option is not specified.

Rel: https://github.com/rapidsai/cuvs/pull/785 vis: @jinsolp


cpp/tests/CMakeLists.txt
```diff
@@ -210,9 +210,13 @@ if(BUILD_TESTS)
...
   ConfigureTest(NAME NEIGHBORS_BALL_COVER_TEST PATH neighbors/ball_cover.cu GPUS 1 PERCENT 100)
 
+  set(NCCL_DEPS "")
+  if(BUILD_MG_ALGOS)
+    list(APPEND NCCL_DEPS NCCL::NCCL)
+  endif()
   ConfigureTest(
     NAME NEIGHBORS_ALL_NEIGHBORS_TEST PATH neighbors/all_neighbors/test_float.cu GPUS 1 PERCENT 100
-    ADDITIONAL_DEP NCCL::NCCL
+    ADDITIONAL_DEP ${NCCL_DEPS}
```
jinsolp: Thanks for finding this! We can just get rid of this NCCL line altogether because this doesn't depend on nccl at all.

enp1s0: Thank you @jinsolp for letting me know. I removed the line.

cjnolet (approved):

cjnolet:
@robertmaynard mind re-reviewing when you have a moment?

PR #1232, opened 2025-08-08
[Java] CuVSMatrix for device memory

ldematte:
This PR introduces implementation classes for `CuVSDeviceMatrix` (a `CuVSMatrix` backed by device memory).
It reworks the base implementation classes a bit to increase reuse, and adds benchmarks and tests for the new classes.

Benchmarks were used to try out different implementations so the best one could be chosen:
- Row access is backed by a buffer of pinned memory
- Builders for device memory use `cudaMemcpyAsync` with the `critical` linker option to directly access heap-based memory
- `cuvsMatrixCopy` is used across the board, as it has the same performances of the various `cudaMemcpy*` functions.

There are some places in the codebase that will benefit from refactoring to use `CuVSDeviceMatrix` (or a generic  `CuVSMatrix` plus `toHost`/`toTensor`/`fromTensor` functions); replacing these multiple ad-hoc implementations with `CuVSDeviceMatrix` will be addressed in a follow-up PR.

Final numbers:
```
Benchmark                                            (dims)  (size)   Mode  Cnt   Score   Error  Units
CuVSDeviceMatrixBenchmarks.matrixCopyDeviceToHost      2048   16384  thrpt    5  70.531 ± 0.322  ops/s
CuVSDeviceMatrixBenchmarks.matrixDeviceBuilder         2048   16384  thrpt    5  35.493 ± 0.772  ops/s
CuVSDeviceMatrixBenchmarks.matrixReadRowsFromDevice    2048   16384  thrpt    5  83.616 ± 0.745  ops/s
```
With theoretical max for the PCI-E bus of 15.7 GB/s and a data size of 128MB, we get close to 2/3 of the maximum theoretical throughput (see comments for details).


java/cuvs-java/src/main/java/com/nvidia/cuvs/CagraIndex.java
```diff
@@ -50,6 +50,13 @@ public interface CagraIndex {
    */
   SearchResults search(CagraQuery query) throws Throwable;
 
+  /** Returns the CAGRA graph
+   *
+   * @return a {@link CuVSMatrix} encapsulating the native int (uint32_t) array used to represent
+   * the cagra graph
+   */
+  CuVSMatrix getGraph();
```
ldematte: Changes to CagraIndex/Impl are due to #1086, you can ignore them for now (they will go away after the rebase).

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/CuVSDeviceMatrixImpl.java
```diff
@@ -0,0 +1,359 @@
...
+      long size,
+      long columns,
+      DataType dataType,
+      ValueLayout valueLayout,
+      int copyType) {
+    super(deviceMemorySegment, dataType, valueLayout, size, columns);
+    this.resources = resources;
+
+    var bufferFillerType = copyType & 0xF;
+    switch (bufferFillerType) {
```
ldematte: Ignore this crude use of interger flags and the different "copy" implementations: these are temporarily here for testing and benchmarking. The best variant will be kept, the other removed (as well as the "magic" flags and `copyType`/`bufferFillerType`).

java/cuvs-java/src/main/java22/com/nvidia/cuvs/spi/JDKProvider.java
```diff
@@ -129,19 +149,32 @@ private void internalAddVector(Object vector) {
...
       }
     };
   }
 
+  @Override
+  public CuVSMatrix.Builder newDeviceMatrixBuilder(
+      CuVSResources resources, int size, int columns, CuVSMatrix.DataType dataType, int copyType)
+      throws UnsupportedOperationException {
+
+    var builderCopyType = copyType & 0xF0;
```
ldematte: If you wonder what this is, see https://github.com/rapidsai/cuvs/pull/1232/files#r2263476356

TL;DR: different implementations to benchmark, only one will make into the final version, and the flags/switch will be removed too.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/common/Native.java
```diff
@@ -0,0 +1,46 @@
...
+
+public final class Native {
+
+  private Native() {}
+
+  private static final Linker LINKER = Linker.nativeLinker();
+
+  private static final MethodHandle cudaMemcpy$mh =
+      LINKER.downcallHandle(
+          cudaMemcpy$address(), cudaMemcpy$descriptor(), Linker.Option.critical(true));
```
ldematte: If you are wondering why we are redefining this, since it is already present in the classes generated by jextract, it's because of this linker option. TL;DR: this allows us to pass a MemorySegment over a heap array to native functions, effectively "pinning" the data. Currently benchmarking this to see if avoiding a copy helps for our usages.

ldematte:
Preliminary micro benchmarks revealed that the CuVS C API functions (e.g. `cuvsMatrixCopy`) have comparable performances to calling the corresponding cuda functions directly; so I removed those and focused on different ways of using those.

By "different ways" I mean different memory: "normal" paged memory (native), direct access to Java heap memory, and pinned memory (allocated with cudaHostAlloc)

```
Benchmark                                                         (dims)  (size)   Mode  Cnt   Score   Error  Units
CuVSDeviceMatrixBenchmarks.matrixCopyToHost                         2048   16384  thrpt    5  34.889 ± 1.166  ops/s
CuVSDeviceMatrixBenchmarks.matrixDeviceBuilderCudaMemcpyCudaHost    2048   16384  thrpt    5  27.454 ± 0.209  ops/s
CuVSDeviceMatrixBenchmarks.matrixDeviceBuilderCudaMemcpyHeap        2048   16384  thrpt    5  33.176 ± 0.578  ops/s
CuVSDeviceMatrixBenchmarks.matrixDeviceBuilderCudaMemcpyNative      2048   16384  thrpt    5  31.574 ± 2.331  ops/s
CuVSDeviceMatrixBenchmarks.matrixReadRowsWithNativeBuffer           2048   16384  thrpt    4  68.148 ± 4.488  ops/s
CuVSDeviceMatrixBenchmarks.matrixReadRowsWithPinnedBuffer           2048   16384  thrpt    5  84.511 ± 1.174  ops/s
```

For the record: these are run on my laptop, which has a 2070 Max-Q on a PCI Express x16 Gen3 bus.
The theoretical max for the bus is 15.7 GB/s (I plan to run them on a different (more modern, enterprise-grade) GPU too, but I wanted to get some numbers to spot obvious problems first).

The dimensions used lead to a matrix with 16K * 2K -> 32M of floats (128MB data)

So 33 ops/s -> 4.1 GByte/s or ~1/4 of the theoretical bandwidth
While 85 ops/s -> 10.6  GByte/s or ~66% of the theoretical bandwidth

`matrixCopyToHost` and `matrixDeviceBuilderCudaMemcpyCudaHost` are a bit surprising -- especially the second, as it seems to indicate that using pinned memory in that case is comparable/a bit slower. While it should be ~2x faster.
Probably these scenarios don't matter too much: the way we are shaping the API, the preferred way would be to read a matrix by row. Still, I want to investigate why they are 2.5x slower. And again, try and see if I see the same results on other GPUs.

ldematte:
Further benchmarks/investigation: `matrixCopyToHost` difference is due to Java allocations/deallocations. If we take that out of the loop, we are back up to > 65ops/s, very much like `matrixReadRowsWithNativeBuffer`, which is what one should expect.

`matrixDeviceBuilder*` performance is due to the size of data copied: since data is copied row-by-row, we are in a situation where we copy in KBs (not MBs) for each call.
This can be optimized; it probably isn't a hot path (it's used to get data from Java heap to device, an already sub-optimal path), and it's not "unusable slow", but I'll give it a go.

ldematte:
Final numbers:
```
Benchmark                                            (dims)  (size)   Mode  Cnt   Score   Error  Units
CuVSDeviceMatrixBenchmarks.matrixCopyDeviceToHost      2048   16384  thrpt    5  70.531 ± 0.322  ops/s
CuVSDeviceMatrixBenchmarks.matrixDeviceBuilder         2048   16384  thrpt    5  35.493 ± 0.772  ops/s
CuVSDeviceMatrixBenchmarks.matrixReadRowsFromDevice    2048   16384  thrpt    5  83.616 ± 0.745  ops/s
```

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/common/Util.java
```diff
@@ -309,12 +323,14 @@ public static BitSet concatenate(BitSet[] arr, int maxSizeOfEachBitSet) {
...
-   * @param[in] data the data pointer points to the allocated data
-   * @param[in] shape the shape of the tensor
-   * @param[in] code the type code of base types
-   * @param[in] bits the shape of the tensor
+   * @param arena      the Arena to use to allocate native memory for this data structure
+   * @param data       the data pointer points to the allocated data
+   * @param shape      the shape of the tensor
+   * @param code       the type code of the matrix elements (e.g. kDLFloat())
+   * @param bits       the size in bits of the matrix elements
+   * @param deviceType the device where {@code data} is held (kDLCPU() or kDLCUDA())
```
ChrisHegarty, line 372: probably worth add the lanes param here too.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/common/Util.java
```diff
@@ -325,31 +341,59 @@ public static MemorySegment prepareTensor(
...
+   * Helper function for creating a DLManagedTensor instance
+   *
+   * @param arena      the Arena to use to allocate native memory for this data structure
+   * @param data       the data pointer points to the allocated data
+   * @param shape      the shape of the tensor
+   * @param strides    the strides for each dimension (can be null)
+   * @param code       the type code of the matrix elements (e.g. kDLFloat())
+   * @param bits       the size in bits of the matrix elements
+   * @param deviceType the device where {@code data} is held (kDLCPU() or kDLCUDA())
+   * @return DLManagedTensor
```
ChrisHegarty, line 390: probably worth add the lanes param here too.

java/cuvs-java/src/main/java/com/nvidia/cuvs/CagraIndex.java
```diff
@@ -55,7 +55,7 @@ public interface CagraIndex {
    * @return a {@link CuVSMatrix} encapsulating the native int (uint32_t) array used to represent
    * the cagra graph
    */
-  CuVSMatrix getGraph();
+  CuVSDeviceMatrix getGraph();
```
ChrisHegarty, line 59: prob update the link in the javadoc to CuVSDeviceMatrix

java/cuvs-java/src/main/java/com/nvidia/cuvs/CuVSDeviceMatrix.java
```diff
@@ -18,4 +18,25 @@
...
+   * Fills the provided, pre-allocated host matrix with data from this device matrix.
+   * The content of the provided host matrix will be overwritten; the 2 matrices must have the
+   * same element type and dimension.
+   *
+   * @param hostMatrix  the host-memory-backed matrix to fill.
+   */
+  void toHost(CuVSHostMatrix hostMatrix);
+
+  /**
+   * FReturns a new, host matrix with data from this device matrix.
```
ChrisHegarty: typo "FReturns"

java/benchmarks/src/main/java/com/nvidia/cuvs/CuVSDeviceMatrixBenchmarks.java
```diff
@@ -0,0 +1,108 @@
...
+    var builder0 = CuVSMatrix.deviceBuilder(resources, size, dims, CuVSMatrix.DataType.FLOAT);
+
+    for (int i = 0; i < size; ++i) {
+      var array = data[i];
+      builder0.addVector(array);
+    }
+
+
+    deviceMatrix = (CuVSDeviceMatrix) builder0.build();
+    hostMatrix = (CuVSHostMatrix) CuVSMatrix.hostBuilder(size, dims, CuVSMatrix.DataType.FLOAT).build();
```
ChrisHegarty: can we generify the builder interface to infer the built subtype? (in order to avoid the need to cast)

ldematte: That's a neat idea, done

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/common/Util.java
ldematte: I removed lanes entirely as a parameter, as we reasonably have and want only scalar types for matrices elements

ldematte:
@mythrocks this is ready for review; if you have time to review it, that would be great!

java/benchmarks/src/main/java/com/nvidia/cuvs/CuVSDeviceMatrixBenchmarks.java
```diff
@@ -0,0 +1,108 @@
...
+    resources = CuVSResources.create();
+
+    var builder0 = CuVSMatrix.deviceBuilder(resources, size, dims, CuVSMatrix.DataType.FLOAT);
+
+    for (int i = 0; i < size; ++i) {
+      var array = data[i];
+      builder0.addVector(array);
+    }
+
+
```
mythrocks:
```suggestion
```

java/cuvs-java/src/main/java/com/nvidia/cuvs/RowView.java
```diff
@@ -16,7 +16,7 @@
 package com.nvidia.cuvs;
 
 /**
- * Represent a contiguous list of integers (32-bit) backed by off-heap memory.
+ * Represent a contiguous list of elements backed by off-heap memory.
```
mythrocks, line 19: 👍 
Argh. Missed that in the previous review.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/common/Native.java
```diff
@@ -0,0 +1,49 @@
...
+
+public final class Native {
+
+  private Native() {}
+
+  private static final Linker LINKER = Linker.nativeLinker();
+
+  private static final MethodHandle cudaMemcpyAsync$mh =
+      LINKER.downcallHandle(
+          cudaMemcpyAsync$address(), cudaMemcpyAsync$descriptor(), Linker.Option.critical(true));
```
mythrocks: 👏

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/common/Util.java
```diff
@@ -120,6 +122,18 @@ public static void cudaMemcpy(MemorySegment dest, MemorySegment src, long numByt
     Util.cudaMemcpy(dest, src, numBytes, CudaMemcpyKind.INFER_DIRECTION);
   }
 
+  /**
+   * Helper to get the default CUDA stream associated with a {@link CuVSResources}
+   */
+  public static MemorySegment getDefaultStream(CuVSResources resources) {
```
mythrocks: I am hesitant to call this the "default" stream.  In CUDA parlance, the default-stream is one of the following:
1. the single stream on which all kernels are scheduled by default, OR
2. when per-thread default streams are enabled, the stream on which kernels are scheduled, for a specific CPU thread.

This variable name overloads the parlance.  Could we consider calling this function `getStream(CuVSResources)`?

Alternatively, could we please add this as a method called `getStream()` in `CuVSResources`?

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/CuVSMatrixBaseImpl.java
```diff
@@ -15,22 +15,36 @@
...
   protected final DataType dataType;
+  protected final ValueLayout valueLayout;
   protected final long size;
   protected final long columns;
 
   protected CuVSMatrixBaseImpl(
-      MemorySegment memorySegment, DataType dataType, long size, long columns) {
+      MemorySegment memorySegment,
+      DataType dataType,
+      ValueLayout valueLayout,
```
mythrocks, line 42: I would have thought `ValueLayout` would be the only thing that would need to be stored here, and that the `DataType` could be queried/derived from the `ValueLayout`.

That wouldn't work for two reasons:
1. We need the datatype to distinguish between the storage of INT and UINT.
2. I can't seem to find a way to query the datatype from the `ValueLayout` anyway. 🤷 

Looks like we would need both.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/HnswIndexImpl.java
```diff
@@ -98,18 +98,23 @@ public SearchResults search(HnswQuery query) throws Throwable {
...
       MemorySegment neighborsMemorySegment = localArena.allocate(neighborsSequenceLayout);
       MemorySegment distancesMemorySegment = localArena.allocate(distancesSequenceLayout);
       MemorySegment querySeg = buildMemorySegment(localArena, queryVectors);
 
       long[] queriesShape = {numQueries, vectorDimension};
-      MemorySegment queriesTensor = prepareTensor(localArena, querySeg, queriesShape, 2, 32, 1, 1);
+      MemorySegment queriesTensor =
+          prepareTensor(localArena, querySeg, queriesShape, kDLFloat(), 32, kDLCPU());
       long[] neighborsShape = {numQueries, topK};
+      // TODO: check type code and bits across all implementations -- they are inconsistent
```
mythrocks, line 110: 👍

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/CuVSMatrixBaseImpl.java
ldematte: Yes. I considered getting rid of the ValueLayout and re-extract it from the DataType everytime, but it was affecting performance as it was on a hot path, so I decided to keep them both. Not great, but a good compromise I think.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/common/Util.java
ldematte: > This variable name overloads the parlance. Could we consider calling this function getStream(CuVSResources)?

Agreed

java/cuvs-java/src/main/java/com/nvidia/cuvs/RowView.java
ldematte: NP, I missed that too!

java/cuvs-java/src/main/java22/com/nvidia/cuvs/spi/JDKProvider.java
```diff
@@ -188,4 +225,86 @@ public CuVSMatrix newMatrixFromArray(byte[][] vectors) {
     Util.copy(dataset.memorySegment(), vectors);
     return dataset;
   }
+
+  private static class HeapSegmentBuilder implements CuVSMatrix.Builder<CuVSDeviceMatrix> {
```
mythrocks, line 233: Could we please add an `@brief` here, indicating the function of this class?

java/cuvs-java/src/main/java22/com/nvidia/cuvs/spi/JDKProvider.java
```diff
@@ -129,19 +147,38 @@ private void internalAddVector(Object vector) {
...
+        return matrix;
       }
     };
   }
 
+  @Override
+  public CuVSMatrix.Builder<CuVSDeviceMatrix> newDeviceMatrixBuilder(
+      CuVSResources resources, long size, long columns, CuVSMatrix.DataType dataType)
+      throws UnsupportedOperationException {
+
```
mythrocks:
```suggestion
```

mythrocks:
Sorry for the delay.  It's taking a while for me to get my head around this.  Looks good so far.  I should be able to complete this review tomorrow.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/spi/JDKProvider.java
ldematte: I don't usually add docs on internal private implementation classes, but I surely can add a comment.
(note: @brief is not a valid javadoc tag, it's doxygen)

ldematte: I updated javadocs on the parent (public) interface, and added a comment on this class.

mythrocks: Thank you.  (The `@brief` is a holdover from C++.  Thanks for taking the suggestion as I meant it.)

java/cuvs-java/src/main/java22/com/nvidia/cuvs/spi/JDKProvider.java
```diff
@@ -188,4 +224,93 @@ public CuVSMatrix newMatrixFromArray(byte[][] vectors) {
...
+   * an intermediate allocation and copy to a native (off-heap) segment.
+   * It requires the copy function ({@code cudaMemcpyAsync}) to have the {@code Critical} linker option in order
+   * to allow the access to on-heap memory (see {@link Native#cudaMemcpyAsync}).
+   */
+  private static class HeapSegmentBuilder implements CuVSMatrix.Builder<CuVSDeviceMatrix> {
+    private final long columns;
+    private final long size;
+    private final CuVSDeviceMatrixImpl matrix;
+    private final MemorySegment stream;
+    int current;
```
mythrocks: Wait, shouldn't this be private?

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/common/Native.java
```diff
@@ -0,0 +1,49 @@
...
+
+  /**
+   * Bindings for {@code cudaMemcpyAsync}; differently from the {@code headers_h} bindings (which are
+   * automatically generated by {@code jextract}), these bindings specify the {@code critical} linker option,
+   * which allows to directly access heap-based {@link MemorySegment}s.
```
mythrocks: I hadn't read this closely enough on the first pass.  I had to return here after your update to `HeapSegmentBuilder`.  Thank you, this documentation is informative. 👏

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/common/Native.java
```diff
@@ -0,0 +1,49 @@
...
+   * {@snippet lang=c :
+   * extern cudaError_t cudaMemcpyAsync(void *dst, const void *src, size_t count, enum cudaMemcpyKind kind, cudaStream_t stream)
+   * }
+   */
+  public static int cudaMemcpyAsync(
+      MemorySegment dst, MemorySegment src, long count, int kind, MemorySegment stream) {
```
mythrocks: Nitpick: I'm not keen on the `int kind` parameter to this function.  `Util.cudaMemcpy` is a wrapper that takes a `CudaMemcpyKind`, an enum for the direction.

We could consider either moving that function to `Native`, or this function to `Util`, and have them both use `CudaMemcpyKind`.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/spi/JDKProvider.java
```diff
@@ -188,4 +224,93 @@ public CuVSMatrix newMatrixFromArray(byte[][] vectors) {
...
+        throw new ArrayIndexOutOfBoundsException();
+      }
+
+      long rowBytes = columns * matrix.valueLayout().byteSize();
+
+      var dstOffset = ((current++) * rowBytes);
+      var dst = matrix.memorySegment().asSlice(dstOffset);
+      checkCudaError(
+          Native.cudaMemcpyAsync(dst, vector, rowBytes, cudaMemcpyHostToDevice(), stream),
+          "cudaMemcpyAsync");
```
mythrocks: This might warrant discussion.  I might not have a good handle on the call stack or usage pattern for this function.

As I understand it, the `vector` has of the order of 1000s of elements?  Is this to be called occasionally, or is this the regular control flow?

You're likely aware:  Calling `cudaMemcpy`/`cudaMemcpyAsync` in a loop for kilobytes at a time is going to have much lower throughput than doing a block copy.

This isn't only meant for use from tests, right?

java/cuvs-java/src/test/java/com/nvidia/cuvs/CuVSMatrixIT.java
```diff
@@ -255,18 +332,32 @@ public void testIntDatasetBuilder() {
...
+    }
+  }
+
+  private void testByteDatasetBuilder(int rows, int cols, CuVSMatrix.Builder builder) {
 
     var data = new byte[rows][cols];
     for (int r = 0; r < rows; ++r) {
       for (int c = 0; c < cols; ++c) {
-        data[c][r] = randomByte();
+        data[r][c] = randomByte();
```
mythrocks, line 357: 👏

mythrocks (commented):
Mostly 👍.

I'd like clarification on the `addVector()` APIs.  Particularly in the case of the device-matrices, I'd like clarification on whether the copy is expected to be row-by-row.

(Edit:  That, and a minor nit regarding `cudaMemcpyAsync` using `CudaMemcpyKind`.)

java/cuvs-java/src/main/java22/com/nvidia/cuvs/spi/JDKProvider.java
ldematte: The class is a private inner class, so it really does not make any difference. But you are correct, I like uniformity, and it carries the intent that this is not meant to be accessed, not even by the enclosing class.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/spi/JDKProvider.java
ldematte: I'm aware of that, and yes (see my comment above on the benchmarks https://github.com/rapidsai/cuvs/pull/1232#issuecomment-3178789800).

> matrixDeviceBuilder* performance is due to the size of data copied: since data is copied row-by-row, we are in a situation where we copy in KBs (not MBs) for each call.
This can be optimized; it probably isn't a hot path (it's used to get data from Java heap to device, an already sub-optimal path), and it's not "unusable slow", but I'll give it a go.

I started implementing this, buffering calls to addVector to a pinned host memory buffer, then calling `cudaMemcpyAsync` to transfer a reasonable sized chunck -- like we already do the other way around (device to host via `getRow` -- we don't get one row at the time). But this PR was already complex enough, I want to keep this optimization separate.

Plus:

> This isn't only meant for use from tests, right?

Yes, that's the most likely use case. Still, it is possible to make it faster, so expect a follow-up PR (with lower priority).

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/common/Native.java
ldematte: I was debating this myself -- it makes little sense to have both this here and `Util`. I will move everything to `Util`.

ldematte:
> I'd like clarification on the addVector() APIs. Particularly in the case of the device-matrices, I'd like clarification on whether the copy is expected to be row-by-row.

Yep, it is expected, I measured it and it is slower, but it's not meant to be used on a hot path and I have ideas and half-backed code to improve it. But I think this PR is already quite big and complex, I'd like to address this separately (in a follow-up PR), if that's OK with you.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/common/Native.java
ldematte: (side note: the `int kind` parameter is kind of needed here, as it needs to match the underlying C binding data type. Moving this to Util will help disguise that)

mythrocks (approved):
Thank you for your patience with the review, @ldematte.  LGTM.

mythrocks:
Seems to be conda-related errors in CI at the moment.  I'll look into this tomorrow.

mythrocks:
Please ignore the merge conflicts here.  I intended to merge this ahead of #1104.

I have raised #1274 to revert #1104, so as to prioritize #1232.

The error is regretted.  I'll re-kick CI when the revert has gone through.

ldematte:
@mythrocks thanks for the heads up!

Issue #1233, opened 2025-08-08
Refactor `graph_build_types.hpp` to consolidate with existing headers


jinsolp:
graph build types in `graph_build_types.hpp` are used by `all_neighbors` and cagra. 
Need to consolidate this into other files. Best would be adding it to `common.hpp`. 

However, this results in a circular dependency (`common.hpp` -> `brute_force.hpp` -> `graph_build_types.hpp` -> `common.hpp`), so should think about how we can do this.

jinsolp:
Closed by this PR: https://github.com/NVIDIA/cuvs/pull/1235

PR #1235, opened 2025-08-08
Refactor knn graph build params to remove `graph_build_types.hpp`

jinsolp:
Closing https://github.com/rapidsai/cuvs/issues/1233

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -35,7 +35,7 @@ concurrency:
 jobs:
   cpp-build:
     secrets: inherit
-    uses: rapidsai/shared-workflows/.github/workflows/conda-cpp-build.yaml@main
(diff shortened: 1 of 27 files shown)


cpp/include/cuvs/neighbors/cagra.hpp
```diff
@@ -35,6 +34,10 @@
 #include <optional>
 #include <variant>
 
+namespace cuvs::neighbors::graph_build_params {
```
cjnolet, line 33: I wonder- could we put these in all_neighbors.hpp and include in cagra.hpp? I know it sounds awkward to do this today, but if we consider that CAGRA will be using all_neighbors soon to build the underlying all-neighbors graph then that'll feel more natural. We could even add a comment in cagra.hpp where we include all_neighbors.hpp, stating that CAGRA shoudl beusing that soon (and maybe link to a Github issue so we remember to remove the comment eventually).

Issue #1191, opened 2025-07-29
[FEA] Support multi-vector lookup


cjnolet:
We need to find a good strategy for multi-vector retrieval. Chamfer distance might be a good option. Vespa (and many other databases) use MaxSim. We should look info these approaches and see what might make the most sense.

PR #1251, opened 2025-08-13
CAGRA: decouple source idx type from graph idx type and add a mapping between them

achirkin:
Add a mapping from internal graph index type to the source (user) index type for CAGRA index.

This PR contains a minimal set of changes required to decouple the graph index type from the public API index type:
  - Add the mapping - a device vector of `n_rows` elements storing the original indices (which are provided by the user)
  - Add corresponding setters/getters and update the serialization code
  - Modify all search kernels accordingly: (1) using the map before coercing the output to `OutputIdxT` and (2) using the map every time the filter operator is used (so that the filtering happens on "user-space" indices rather than on internal indices).

In the detail namespace, I add a new template parameter `SourceIdxT` , which doesn't currently multiply the template instantiations. `SourceIdxT` is bound directly to IdxT in `index<T, IdxT>`, whereas the `IndexT` in the detail namespace becomes hard-coded to `graph_idx_t`, which is `uint32_t`.

***What's not here***
This PR does not modify the order of the graph or the dataset during construction. Hence to test this feature now, a user needs to reorder both the dataset and the graph manually using the `update_xxx` methods of the CAGRA index.

***IVF-XX Feature parity***
With the addition of the index mapping, it will be possible to add the indices argument to the `extend` function similar to the other algorithms.

***Multi-vector index***
(related issue https://github.com/rapidsai/cuvs/issues/1191)
This PR can be seen as a prerequisite for <single_id - multiple_vectors> mapping: since the ids in the source_indicies vector are not necessarily unique. To actually allow multi-vector usage in a meaningful way, we also need to:
  - De-duplicate the indices in the search results
  - Extend the public API - how do add new vectors and indices alongside them, and where/how to keep the dataset.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -34,7 +34,7 @@ concurrency:
 jobs:
   cpp-build:
     secrets: inherit
-    uses: rapidsai/shared-workflows/.github/workflows/conda-cpp-build.yaml@branch-25.12
+    uses: rapidsai/shared-workflows/.github/workflows/conda-cpp-build.yaml@main
     with:
       build_type: ${{ inputs.build_type || 'branch' }}
       branch: ${{ inputs.branch }}
@@ -44,7 +44,7 @@ jobs:
   rust-build:
     needs: cpp-build
     secrets: inherit
-    uses: rapidsai/shared-workflows/.github/workflows/custom-job.yaml@branch-25.12
+    uses: rapidsai/shared-workflows/.github/workflows/custom-job.yaml@main
     # Artifacts are not published from these jobs, so it's safe to run for multiple CUDA versions.
     # If these jobs start producing artifacts, the names will have to differentiate between CUDA versions.
     strategy:
@@ -65,7 +65,7 @@ jobs:
   go-build:
     needs: cpp-build
     secrets: inherit
-    uses: rapidsai/shared-workflows/.github/workflows/custom-job.yaml@branch-25.12
+    uses: rapidsai/shared-workflows/.github/workflows/custom-job.yaml@main
     # Artifacts are not published from these jobs, so it's safe to run for multiple CUDA versions.
     # If these jobs start producing artifacts, the names will have to differentiate between CUDA versions.
     strategy:
@@ -86,7 +86,7 @@ jobs:
   java-build:
     needs: cpp-build
     secrets: inherit
-    uses: rapidsai/shared-workflows/.github/workflows/custom-job.yaml@branch-25.12
+    uses: rapidsai/shared-workflows/.github/workflows/custom-job.yaml@main
     # Artifacts are not published from these jobs, so it's safe to run for multiple CUDA versions.
     # If these jobs start producing artifacts, the names will have to differentiate between CUDA versions.
     strategy:
@@ -108,7 +108,7 @@ jobs:
   python-build:
     needs: [cpp-build]
     secrets: inherit
-    uses: rapidsai/shared-workflows/.github/workflows/conda-python-build.yaml@branch-25.12
+    uses: rapidsai/shared-workflows/.github/workflows/conda-python-build.yaml@main
     with:
       build_type: ${{ inputs.build_type || 'branch' }}
       branch: ${{ inputs.branch }}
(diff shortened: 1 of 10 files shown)


achirkin:
## Discussion topic
The current implementation assumes the dataset and the graph are both using the internal indices, i.e. they are in the same order. The question is if we want to use the original indices for the dataset, or decide on this based on whether the index owns the dataset, or allow the user to choose between the two behaviors?

cpp/include/cuvs/neighbors/cagra.hpp
```diff
@@ -291,6 +291,7 @@ struct index : cuvs::neighbors::index {
   using index_type         = IdxT;
   using value_type         = T;
   using dataset_index_type = int64_t;
+  using graph_index_type   = uint32_t;
```
cjnolet, line 294: Makes total sense. Here we need to capture the `max graph node id` separately from the `number of nonzeros`.

achirkin:
Benchmarked on Wiki-10M dataset, search performance is identical (PR vs branch-25.10). Hence moving the PR out from the draft state.

Note, I don't set the mapping in the benchmarks, so the results don't tell anything about the performance when the mapping is actually used.

cpp/src/neighbors/detail/cagra/cagra_search.cuh
```diff
@@ -48,15 +48,18 @@ template <typename DataT,
...
-void search_main_core(raft::resources const& res,
-                      search_params params,
-                      const dataset_descriptor_host<DataT, IndexT, DistanceT>& dataset_desc,
-                      raft::device_matrix_view<const IndexT, int64_t, raft::row_major> graph,
-                      raft::device_matrix_view<const DataT, int64_t, raft::row_major> queries,
-                      raft::device_matrix_view<OutputIdxT, int64_t, raft::row_major> neighbors,
-                      raft::device_matrix_view<DistanceT, int64_t, raft::row_major> distances,
-                      CagraSampleFilterT sample_filter = CagraSampleFilterT())
+          typename SourceIdxT = IndexT,
+          typename OutputIdxT = SourceIdxT>
```
divyegala, line 54: Should we add this to ensure we always instantiate the graph as `uint32_t`?
```suggestion
          typename GraphIndexT,
          typename SourceIdxT = IndexT,
          typename OutputIdxT = SourceIdxT,
          std::enable_if_t<std::is_same_v<GraphIndexType, uint32_t>>* = nullptr>
```

cpp/src/neighbors/detail/cagra/cagra_search.cuh
```diff
@@ -145,25 +150,42 @@ void search_main(raft::resources const& res,
                  CagraSampleFilterT sample_filter = CagraSampleFilterT())
 {
   // n_rows has the same type as the dataset index (the array extents type)
-  using ds_idx_type = decltype(index.data().n_rows());
+  using ds_idx_type    = decltype(index.data().n_rows());
+  using graph_idx_type = uint32_t;
```
divyegala, line 158: Based on comment above, this will become unneeded then.

cpp/src/neighbors/detail/cagra/search_multi_cta_kernel-inl.cuh
divyegala, line 1: What is going on with the diff render for this file, do you have an idea?

cpp/src/neighbors/detail/cagra/cagra_search.cuh
achirkin: In this context, I find it useful to name types just for better readability: the meaning of the template parameters is clear when passed to template-heavy functions.

cpp/src/neighbors/detail/cagra/cagra_search.cuh
achirkin: We could as well just remove the template parameter altogether or have this check as static_assert in the body of the function. But it will error either way if instantiated with something else. I's prefer not to build up more things in the function signature :)

cpp/src/neighbors/detail/cagra/search_multi_cta_kernel-inl.cuh
achirkin: Good question! That's unfortunate, because there are couple meaningful changes in the kernel (using `to_source_index` lambda).

cpp/src/neighbors/detail/cagra/cagra_search.cuh
divyegala: Static assertion is fine with me!

benfred (approved):

PR #1252, opened 2025-08-13
[REVIEW][Java] Rename destroyIndex() to close(), extend AutoCloseable

ldematte:
The various `Index` classes in Java all have a `destroyIndex` method that must be called in order to free resources associated with that index, when done.
Other classes in the repository use `close()` instead, and extend or implement `AutoCloseable`; this have the advantage of support in language (try-with-resources) and tooling (e.g. IntelliJ warnings).
Additionally, having a single pattern is better for ease of use and consistency.

This PR renames all instances of `destroyIndex` to `close` and make the declaring interfaces extend `AutoCloseable`


mythrocks:
Thank you for taking this up.  I've been meaning to make this change myself.

100% with you here.

mythrocks:
Argh.  I should have removed the `[REVIEW]` tag.  I'll double check, next time.

PR #1255, opened 2025-08-13
Run `cargo fmt` in the pre-commit hooks

benfred:


cjnolet (approved):

jameslamb (approved):
Pretty nice! I just tried this out locally on my Mac, where I don't do any Rust development and so don't have `cargo` already installed.

```console
$ type -f cargo
-bash: type: cargo: not found
```

Found that `pre-commit` happily set it up for me, in its own dedicated environment. That took about 2 minutes which I think is tolerable.

```shell
pre-commit run --all-files
```

The second run of `pre-commit run` took less than 10 seconds, so most of that time was one-time setup of `cargo`.

I support this!

Issue #666, opened 2025-02-06
[BUG] cuvs-java: creating a carga index with a single vectors fails with cudaErrorIllegalAddress:an illegal memory access was encountered


ChrisHegarty:
**Describe the bug**
Creating a carga index with a single vectors fails with cudaErrorIllegalAddress:an illegal memory access was encountered

**Steps/Code to reproduce bug**
```
// This is a non-realistic trivial dataset, with just a single small
// vector. It demonstrates that trying to build a cagra index with a
// single vector fails. Such a scenario can happen when data is indexed
// and merged across shards, e.g. in Lucene or Elasticsearch.
@Test
public void testIndexAndSearchSingle() throws Throwable {
  float[][] dataset = {
      { 0.74021935f, 0.9209938f }
  };

  try (CuVSResources resources = CuVSResources.create()) {
    CagraIndexParams indexParams = new CagraIndexParams.Builder()
      .withCagraGraphBuildAlgo(CagraGraphBuildAlgo.NN_DESCENT)
      .withGraphDegree(1)
      .withIntermediateGraphDegree(2)
      .withNumWriterThreads(32)
      .withMetric(CuvsDistanceType.L2Expanded)
      .build();
    CagraIndex index = CagraIndex.newBuilder(resources)
      .withDataset(dataset)
      .withIndexParams(indexParams)
      .build();
    index.destroyIndex();
  }
}
```

**Expected behavior**

It will simply java clients if they do not have to special case when the dataset contain a single vector.

**Environment details (please complete the following information):**
 - Environment location: [Bare-metal, Docker, Cloud(specify cloud provider)]
   - AWS g4dn.2xlarge, Ubuntu 22.04 + cuda 12.6 

set the JAVA_HOME env variable to JDK 22
```
JAVA_HOME=/home/ubuntu/binaries/jdk-22.0.2/
```

**Additional context**

Fails on `branch-25.02`

narangvivek10:
Additional information:

I tried reproducing this issue [using a modified version](https://gist.github.com/narangvivek10/ad6f62107588f1406926f145d107e6ed) of the [C example in the repo](https://github.com/rapidsai/cuvs/blob/branch-25.02/examples/c/src/cagra_c_example.c) and also saw the index build failing with the similar reason. @cjnolet fyi.

```
[...]

PR #1256, opened 2025-08-14
Single vector CAGRA index to not throw cudaErrorIllegalAddress (#666)

chatman:
This is a fix for #666. It is useful for github.com/cuvs-lucene to avoid hacks that fallback to using Brute Force Index if dataset size is 1.

### DIFF
--- a/cpp/src/neighbors/detail/cagra/cagra_build.cuh
+++ b/cpp/src/neighbors/detail/cagra/cagra_build.cuh
@@ -674,6 +674,27 @@ index<T, IdxT> build(
                                      : "device",
     intermediate_degree,
     graph_degree);
+
+  // Handle edge case with single vector in dataset
+  if (dataset.extent(0) == 1) {
+    // Return an index with just the single vector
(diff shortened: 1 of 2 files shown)


achirkin:
Thanks, this is definitely something we should fix on the cuvs/C++ side. But also I think this deserves a deeper investigation: does this same error happen with just two vectors in the index? or three? Generally, how the graph is constructed for the number of vectors less than the graph degree?

cjnolet:
Ishan, we should be using the tiered index in lucene so that we can avoid
this very problem.

Cagra graph is simply not valid for such small numbers of vectors. It’s
meant to be a batched created index. It’s not the same we as HNSW.

I’d prefer to focus the engineering team’s efforts on legitimate bugs and
have cuVS-lucene use the proper index type.

divyegala:
@chatman do you have an update to Artem's question above? https://github.com/rapidsai/cuvs/pull/1256#issuecomment-3187414017

Issue #456, opened 2024-11-08
[FEA]  Expose `build_knn_graph()` and `optimize()` functions in `cuvs::neighbors::cagra::helpers`.


cjnolet:
Several users have made us aware that they are using these functions but we haven't yet ported them over to cuVS. We need to do this, and soon.

Issue #1146, opened 2025-07-21
[FEA] Establish streaming incremental micro-batched index build for CAGRA


cjnolet:
Oracle has asked that we overlap batches as much as possible to support the ability to stream batches into an API so that we can start building a large CAGRA graph on the batches prior to the last batch being received. The idea is that we immediately start building so that we don't have to wait for the last batch to arrive. 

We need to find a good solution for this, because it's going to keep coming up as we push more heavily on serverless and micro-service infrastructures. 

One idea is that we once again provide a separate `optimize()` function in `cuvs::neighbors::cagra::helpers` so that we can build the all-neighbors graph in a streaming fashion (using a variant of our new `cuvs::neighbors::all_neighbors` APIs) and then run that graph through the optimization process. 

This, of course, assumes the optimization process doesn't end up becoming the bottleneck. If it does, we might need to separate that our further.

cjnolet:
cc @abc99lr 

cc @jinsolp for awareness.

abc99lr:
Thanks @cjnolet for creating this issue.

> One idea is that we once again provide a separate optimize() function in cuvs::neighbors::cagra::helpers

It seems cuvs already have two versions `cuvs::neighbor::cagra::optimize` defined [here](https://github.com/rapidsai/cuvs/blob/branch-25.10/cpp/src/neighbors/cagra_optimize.cu). However, there are no declarations for these methods in any headers. The first step of this work would be adding a public headers for these two optimize methods, but do we want to keep using `cuvs::neighbor::cagra` namespace as showed in the code or declare them under `cuvs::neighbor::cagra::helpers` as you suggested here?

cjnolet:
Hey @abc99lr, welcome back! Glad you are picking this up. 

> there are no declarations for these methods in any headers.

There are no declarations because these functions aren't part of the public APIs (yet). They used to be at one point when they were in RAFT, but we had removed them when bringing them over to cuVS while we determined whether they should be public again or not. 

> namespace as showed in the code or declare them under cuvs::neighbor::cagra::helpers as you suggested here?

As mentioned, the public functions should be exposed in the `cagra::helpers` namespace. This to separate them from the other standard-use CAGRA APIs and designate them as "special functions that not everyone will use, but which provide additional useful capabilities".

jinsolp:
@cjnolet @abc99lr Is there anything I could help with here?

cjnolet:
[...]

PR #1260, opened 2025-08-14
[REVIEW] Add a public API for CAGRA graph optimize

abc99lr:
This change exposes CAGRA's graph optimization functionality through a new public API in the `cuvs::neighbors::cagra::helpers` namespace. The optimize function allows users to optimize a custom KNN graphs to create CAGRA search graphs.

Key Changes:

- Added `cpp/include/cuvs/neighbors/cagra_optimize.hpp` with a public API declaration. As cagra_optimize assumes both input and output are host matrices, this PR exposes the API with host inputs only. Given this, users need to explicitly manage device-to-host transfers if KNN graph is not a RAFT host matrix
- Modified previously unused `cpp/src/neighbors/cagra_optimize.cu` with implementation that forwards to internal CAGRA optimize. Removed the version using device matrix for `knn_graph`, which generates runtime segfaults
- Added unit tests for the new API

This PR partially resolves https://github.com/rapidsai/cuvs/issues/456 and completes one of the steps for https://github.com/rapidsai/cuvs/issues/1146


cpp/include/cuvs/neighbors/cagra_optimize.hpp
```diff
@@ -0,0 +1,57 @@
...
+ *   auto h_out = raft::make_host_matrix<uint32_t, int64_t>(N, K_out);
+ *   cuvs::neighbors::cagra::helpers::optimize(res, h_knn.view(), h_out.view());
+ * @endcode
+ *
+ * @param[in] handle RAFT resources
+ * @param[in] knn_graph Input KNN graph on host [n_rows, k_in]
+ * @param[out] new_graph Output CAGRA graph on host [n_rows, k_out]
+ */
+void optimize(raft::resources const& handle,
+              raft::host_matrix_view<uint32_t, int64_t, raft::row_major> knn_graph,
```
cjnolet, line 46: I think we want to expose device overloads for this too, right?

abc99lr: Currently `cagra::optimize` cannot directly work on device KNN graph (e.g., `count_2hop_detours ` and `mst_optimization` both require host KNNG). We could add a device-to-host copy in `cagra::optimize` to make the device KNNG input work though. But I think a device overload with an internal copy may confuse users with poorer performance and it's better to for users to handle the copy if needed. Also, stream CAGRA doesn't require the device overload. That's said, I can add a device version if people think there is value.

cpp/include/cuvs/neighbors/cagra_optimize.hpp
```diff
@@ -0,0 +1,57 @@
...
+ * This function prunes and optimizes a k-NN graph to create a more efficient
+ * CAGRA search graph. The input graph must be on host memory as the optimization
+ * algorithm uses CPU-based processing.
+ *
+ * Usage example:
+ * @code{.cpp}
+ *   raft::resources res;
+ *
+ *   // If starting with device data, copy to host first:
+ *   auto d_knn = raft::make_device_matrix<uint32_t, int64_t>(res, N, K_in);
```
cjnolet: I think we can exclude this from the example- we don't need to assume user has the knn on device.

cpp/include/cuvs/neighbors/cagra_optimize.hpp
cjnolet: Yeah, it's a tough call- because I think there are things we could be doing w/ UVM and pinned memory inside of optimize(), for example. Okay, let's just include the host functions for now to be explicit. I suspect we'll soon have device variants, but we can add those when the time comes.

abc99lr: Sounds good. Thanks

abc99lr:
Thanks for the comments @cjnolet. It seems I cannot add labels to this PR which is needed for CI. Could you help me add labels. Thanks!

cjnolet:
@abc99lr CI is broken across the board atm. We shall have it fixed soon (by Monday hopefully) and then I'll run the checks in this PR.

abc99lr:
@cjnolet, is there anything missing to get this merged? I plan to start porting the streamed CAGRA example after this is merged. Thanks!

PR #1268, opened 2025-08-21
Fix balanced kmeans API

enp1s0:
This PR updates cluster::balanced_kmeans to:
1. use `int64_t` as the `mdspan` index type to support a large dataset
2. fix the `fit` function for `data_t` = `int8_t`
3. test `fit` and `predict` in the unit test.

## 1. Use `int64_t` as the `mdspan` index type to support a large dataset

In the current implementation, the mdspan index type is `int32_t`. Even if `int64_t` is specified, it is cast to `int32_t`.

e.g.
```cpp
auto dataset = raft::make_device_matrix<float, int64_t>(handle, n, d);
cuvs::cluster::kmeans::fit(handle, params, mkconst(dataset.view()), centroids_view); // Here, the index type of the dataset view is cast to int32_t since there is only an API for int32_t.
```

This cast results in an unexpected error in this check. `IndexT` is `int32_t` even if the input index type is `int64_t`.
https://github.com/rapidsai/cuvs/blob/d0a83d43f7be934f0c4dcef585302a7fad287df0/cpp/src/cluster/kmeans_balanced.cuh#L86-L88

This PR changes the index type to `int64_t`. If `int32_t` is specified, it is cast to `int64_t`. Therefore, this change is not breaking.

## 2. Fix the `fit` function for `data_t` = `int8_t`
The centroid data type of the `predict` and `fit_predict` functions is `float`, but `int8_t` for the `fit` function (`fit_predict` is a function that sequentially calls the `fit` and `predict` functions internally).
This PR changes it to `float` since this data type must be the same as the one of `predict`.

## 3. Test `fit` and `predict` in the unit test.
In the current unit test, only the `fit_predict` function is tested. This PR adds tests for the `fit` and `predict` functions.

## Discussion
There is no consistency in the label data types for `predict` and `fit_predict`: `predict` is `int32_t`, while `fit_predict` is `uint32_t`. Would it be better to:
1. Support both types in both functions (e.g., by reinterpret_cast-ing int32_t* to uint32*),
2. Unify on one type, or
3. Keep the inconsistency?


cpp/include/cuvs/cluster/kmeans.hpp
```diff
@@ -781,9 +782,53 @@ void predict(const raft::resources& handle,
...
+ *                                 [dim = n_clusters x n_features]
+ * @param[out]    labels           Index of the cluster each sample in X
+ *                                 belongs to.
+ *                                 [len = n_samples]
+ */
+void predict(const raft::resources& handle,
+             cuvs::cluster::kmeans::balanced_params const& params,
+             raft::device_matrix_view<const float, int64_t> X,
+             raft::device_matrix_view<const float, int64_t> centroids,
+             raft::device_vector_view<int, int64_t> labels);
```
tarang-jain, line 831: Shouldn't the labels type and extent be of the same type? By allowing int64_t extents, labels can also go beyond int32 limits

enp1s0: Thank you for the comment. I don't think `int64_t` is needed for the label data type because the number of centroids is typically not that large. I believe `int32_t` is sufficient.

PR #1269, opened 2025-08-21
Added UINT to assert in toArray

punAhuja:
Added assert for UINT datatype.
As per issue encountered and debugged by @narangvivek10 while working on Lucene HNSW serialization

### DIFF
--- a/java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/CuVSHostMatrixImpl.java
+++ b/java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/CuVSHostMatrixImpl.java
@@ -84,7 +84,7 @@ public class CuVSHostMatrixImpl extends CuVSMatrixBaseImpl implements CuVSHostMa
 
   @Override
   public void toArray(int[][] array) {
-    assert dataType == DataType.INT;
+    assert (dataType == DataType.UINT || dataType == DataType.INT);
     assert (array.length >= size) : "Input array is not large enough";
     assert (array.length == 0 || array[0].length >= columns) : "Input array is not large enough";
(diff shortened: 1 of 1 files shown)


chatman:
@narangvivek10 please add more details on how you encountered this issue. Also, any ideas how we can test for that scenario?

cjnolet (approved):

mythrocks:
It would be good to raise a bug indicating the error that @narangvivek10 ran into on this front, and add a "fixes XXXX" in the description of this PR.  Maybe include a snippet of how the bug manifests in the PR description.

ldematte:
> Also, any ideas how we can test for that scenario?

We are missing tests where the input data is int and the datatype is UINT. Those can be added to `CuVSMatrixIT` (e.g. testIntDatasetCopy -> also have testUIntDatasetCopy, etc.)

Maybe related/more importantly, we are missing a lot of tests on index operations (build, search, etc.): those are only with a host matrix/heap input (no device matrix) and only for float (no int or byte).
I have a plan to add those in the next week or so, but if anyone wants to do it you'll be more than welcome! :)

ldematte:
But wait, I believe your branch is just out of date -- I have fixed this already as part of #1232 

EDIT: yep, very unfortunate timing: you raised this PR on August 21, and my PR got merged on August 22.
I think this can be closed.

mythrocks:
@punAhuja:  Could you please check if the latest commits already address what this PR attempts to solve?

punAhuja:
> @punAhuja: Could you please check if the latest commits already address what this PR attempts to solve?

Yes, the latest commits already address what this PR attempts to solve. Will close this PR.

PR #1272, opened 2025-08-21
Use int64_t for getters in the cagra/ivf_flat c-api

benfred:
Java doesn't support unsigned types well, and this is leading to us only supporting 2^31 items in the cagra index - instead of 2^32.

Fix by changing the getters in the c-api to return int64_t values for the size/dim etc, instead of uint32_t.

We were also not consistent on returning error codes in ivf_flat/ivf_pq for these getters - with cagra returning an error code, and the ivf* methods just returning the value directly. Change to always return an error code to be consistent. (Note that #1123 is also making this change for ivf-pq - and I'm skipping ivf-pq to avoid merge conflicts).

### DIFF
--- a/cpp/include/cuvs/neighbors/ivf_pq.h
+++ b/cpp/include/cuvs/neighbors/ivf_pq.h
@@ -262,10 +262,13 @@ cuvsError_t cuvsIvfPqIndexCreate(cuvsIvfPqIndex_t* index);
 cuvsError_t cuvsIvfPqIndexDestroy(cuvsIvfPqIndex_t index);
 
 /** Get the number of clusters/inverted lists */
-uint32_t cuvsIvfPqIndexGetNLists(cuvsIvfPqIndex_t index);
+cuvsError_t cuvsIvfPqIndexGetNLists(cuvsIvfPqIndex_t index, int64_t* n_lists);
 
 /** Get the dimensionality */
-uint32_t cuvsIvfPqIndexGetDim(cuvsIvfPqIndex_t index);
+cuvsError_t cuvsIvfPqIndexGetDim(cuvsIvfPqIndex_t index, int64_t* dim);
+
+/** Get the size of the index */
+cuvsError_t cuvsIvfPqIndexGetSize(cuvsIvfPqIndex_t index, int64_t* size);
 
 /**
  * @brief Get the cluster centers corresponding to the lists in the original space
--- a/cpp/src/neighbors/ivf_pq_c.cpp
+++ b/cpp/src/neighbors/ivf_pq_c.cpp
@@ -347,16 +347,28 @@ extern "C" cuvsError_t cuvsIvfPqExtend(cuvsResources_t res,
   });
 }
 
-extern "C" uint32_t cuvsIvfPqIndexGetNLists(cuvsIvfPqIndex_t index)
+extern "C" cuvsError_t cuvsIvfPqIndexGetNLists(cuvsIvfPqIndex_t index, int64_t* n_lists)
 {
-  auto index_ptr = reinterpret_cast<cuvs::neighbors::ivf_pq::index<int64_t>*>(index->addr);
-  return index_ptr->n_lists();
+  return cuvs::core::translate_exceptions([=] {
+    auto index_ptr = reinterpret_cast<cuvs::neighbors::ivf_pq::index<int64_t>*>(index->addr);
+    *n_lists       = index_ptr->n_lists();
+  });
 }
 
-extern "C" uint32_t cuvsIvfPqIndexGetDim(cuvsIvfPqIndex_t index)
+extern "C" cuvsError_t cuvsIvfPqIndexGetDim(cuvsIvfPqIndex_t index, int64_t* dim)
 {
-  auto index_ptr = reinterpret_cast<cuvs::neighbors::ivf_pq::index<int64_t>*>(index->addr);
-  return index_ptr->dim();
+  return cuvs::core::translate_exceptions([=] {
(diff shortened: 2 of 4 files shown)


cpp/src/neighbors/cagra_c.cpp
```diff
@@ -426,23 +426,23 @@ extern "C" cuvsError_t cuvsCagraIndexDestroy(cuvsCagraIndex_t index_c_ptr)
   });
 }
 
-extern "C" cuvsError_t cuvsCagraIndexGetDims(cuvsCagraIndex_t index, int* dim)
+extern "C" cuvsError_t cuvsCagraIndexGetDims(cuvsCagraIndex_t index, int64_t* dim)
 {
   return cuvs::core::translate_exceptions([=] {
     auto index_ptr = reinterpret_cast<cuvs::neighbors::cagra::index<float, uint32_t>*>(index->addr);
```
mythrocks, line 432: Tangential question:  How come `index->addr` is a `uintprt_t` and not a `void*`?

(Functionally, it's likely equivalent, except for the ability to use `static_cast` instead of `reinterpret_cast`.)

mythrocks (approved):
C++ and Java changes: 👍 

Python changes: Look good to my eye, but should probably be reviewed by an expert.

cpp/src/neighbors/ivf_flat_c.cpp
```diff
@@ -395,50 +395,54 @@ extern "C" cuvsError_t cuvsIvfFlatExtend(cuvsResources_t res,
...
-      reinterpret_cast<cuvs::neighbors::ivf_flat::index<uint8_t, int64_t>*>(index->addr);
-    return index_ptr->n_lists();
-  } else {
-    return 0;
-  }
+  return cuvs::core::translate_exceptions([=] {
+    if (index->dtype.code == kDLFloat && index->dtype.bits == 32) {
+      auto index_ptr =
+        reinterpret_cast<cuvs::neighbors::ivf_flat::index<float, int64_t>*>(index->addr);
+      *n_lists = index_ptr->n_lists();
```
tarang-jain, line 382: Since this is a non-narrowing conversion (uint32_t -> int64_t), an explicit `static_cast` is not required, but explicit casting can still help with readability and safety.
Same comment for all the other getters (such as cuvsIvfFlatIndexGetDim) that are doing an implicit casting.

cpp/src/neighbors/ivf_flat_c.cpp
```diff
@@ -395,50 +395,54 @@ extern "C" cuvsError_t cuvsIvfFlatExtend(cuvsResources_t res,
...
+        reinterpret_cast<cuvs::neighbors::ivf_flat::index<uint8_t, int64_t>*>(index->addr);
+      *n_lists = index_ptr->n_lists();
+    } else {
+      RAFT_FAIL("Unsupported index dtype: %d and bits: %d", index->dtype.code, index->dtype.bits);
+    }
+  });
 }
 
-extern "C" uint32_t cuvsIvfFlatIndexGetDim(cuvsIvfFlatIndex_t index)
+extern "C" cuvsError_t cuvsIvfFlatIndexGetDim(cuvsIvfFlatIndex_t index, int64_t* dim)
```
tarang-jain, line 401: Nit: Actually, it might be possible to return value rather than resetting the value inside a pointer via a void function, right? We can have int64_t as the return type in the function signature and static_cast the result to int64_t. That way we can avoid the slightly cumbersome route for C-API users to first allocate the scalar and then use the getter.

benfred: The goal here is to be consistent - we have some methods returning the value directly, and others returning an error code.   I want every method here to follow the same pattern, and not have the CAGRA getters return an error code and the ivf-* methods return the value directly  (see also https://github.com/rapidsai/cuvs/pull/1123#discussion_r2277748506 ).

We have to pick one convention to be consistent - and I think returning an error code here is slightly safer . This method actually can return an error right now (if the index dtype is messed up, previously we returned a `0` value and now we return an error code / text  - even though in practice this shouldn't happen).  

I agree that its slightly more cumbersome for users of the C-api - but if you look at the corresponding changes in the python / java code its not *that* bad - since we have to check an error code for all the other function calls in the c-api anyways

mythrocks: > The goal here is to be consistent...

+1.  It was a little surprising that the C-API wasn't already consistent on this front.  Thanks for sorting that out, @benfred.

tarang-jain: Sounds good @benfred ! So then we should change the getters of IVF-PQ also to this pointer based approach (if its not already being done that way)

benfred: yup, we should also update ivf-pq in the same way! 

I briefly mentioned mentioned this in the PR description - but the change for ivf-pq ishttps://github.com/rapidsai/cuvs/pull/1123 , and I didn't make the change here to avoid merge conflicts

mythrocks (approved):

cpp/src/neighbors/ivf_flat_c.cpp
cjnolet: 100% on being consistent and returning error codes everywhere (instead of returning values)

cjnolet (approved):

cpp/src/neighbors/ivf_flat_c.cpp
benfred: I've updated ivf-pq here https://github.com/rapidsai/cuvs/pull/1272/commits/ce8016a541c10323859e274338c5c72fa4696807 to be consistent since I'm not sure when 1123 will get merged

PR #1273, opened 2025-08-22
Build and test with CUDA 13.0.0

jameslamb:
Contributes to https://github.com/rapidsai/build-planning/issues/208

* uses CUDA 13.0.0 to build and test
* adds CUDA 13 devcontainers
* moves some dependency pins:
  - `cuda-python`: `>=12.9.2` (CUDA 12), `>=13.0.1` (CUDA 13)
  - `cupy`: `>=13.6.0`

Contributes to https://github.com/rapidsai/build-planning/issues/68

* updates to CUDA 13 dependencies in fallback entries in `dependencies.yaml` matrices (i.e., the ones that get written to `pyproject.toml` in source control)

## Notes for Reviewers

This switches GitHub Actions workflows to the `cuda13.0` branch from here: https://github.com/rapidsai/shared-workflows/pull/413

A future round of PRs will revert that back to `branch-25.10`, once all of RAPIDS supports CUDA 13.

### What about Go, Java, and Rust?

This PR expands building / testing for those bindings to cover CUDA 13, but more changes probably need to be made to support distributing packages for those.

Proposing deferring that to follow-ups:

* #1275
* #1276
* #1277


jameslamb:
The first runs all failed with some form of "need RAFT packages", so nothing else learned from this yet.

build link: https://github.com/rapidsai/cuvs/actions/runs/17145663792/job/48641578548

Will update this once https://github.com/rapidsai/raft/pull/2787 is merged and we have RAFT packages.

jameslamb:
Happy to say that wheels are looking happy, so `libcuvs.so` CUDA 13 support is probably close!

But some benchmarks are failing to compile, like this:

```text
 │ │ [680/773] Building CXX object bench/ann/CMakeFiles/HNSWLIB_ANN_BENCH.dir/src/hnswlib/hnswlib_benchmark.cpp.o
 │ │ FAILED: [code=1] bench/ann/CMakeFiles/HNSWLIB_ANN_BENCH.dir/src/hnswlib/hnswlib_benchmark.cpp.o 
 │ │ sccache $BUILD_PREFIX/bin/aarch64-conda-linux-gnu-c++ -DANN_BENCH_BUILD_MAIN> -DBENCHMARK_STATIC_DEFINE -DCUVS_ANN_BENCH_USE_HNSWLIB=CUVS_ANN_BENCH_USE_HNSWLIB -I$SRC_DIR/cpp/include -I$SRC_DIR/cpp/build/_deps/benchmark-src/include -I$SRC_DIR/cpp/build/_deps/hnswlib-src -isystem $BUILD_PREFIX/targets/sbsa-linux/include/cccl -isystem /opt/conda/include -isystem $BUILD_PREFIX/include -fvisibility-inlines-hidden -fmessage-length=0 -ftree-vectorize -fPIC -fstack-protector-strong -fno-plt -O3 -pipe -isystem $PREFIX/include   -I$PREFIX/targets/sbsa-linux/include -I$PREFIX/targets/sbsa-linux/include/cccl -I$BUILD_PREFIX/targets/sbsa-linux/include -I$BUILD_PREFIX/targets/sbsa-linux/include/cccl -O3 -DNDEBUG -std=gnu++17 -fPIE -pthread -fopenmp "-ffile-prefix-map=$PREFIX/=''" -MD -MT bench/ann/CMakeFiles/HNSWLIB_ANN_BENCH.dir/src/hnswlib/hnswlib_benchmark.cpp.o -MF bench/ann/CMakeFiles/HNSWLIB_ANN_BENCH.dir/src/hnswlib/hnswlib_benchmark.cpp.o.d -o bench/ann/CMakeFiles/HNSWLIB_ANN_BENCH.dir/src/hnswlib/hnswlib_benchmark.cpp.o -c $SRC_DIR/cpp/bench/ann/src/hnswlib/hnswlib_benchmark.cpp
 │ │ In file included from $SRC_DIR/cpp/bench/ann/src/hnswlib/hnswlib_wrapper.h:20,
 │ │                  from $SRC_DIR/cpp/bench/ann/src/hnswlib/hnswlib_benchmark.cpp:18:
 │ │ $SRC_DIR/cpp/bench/ann/src/hnswlib/../common/util.hpp: In function 'auto cuvs::bench::cuda_info()':
 │ │ $SRC_DIR/cpp/bench/ann/src/hnswlib/../common/util.hpp:451:64: error: 'struct cudaDeviceProp' has no member named 'clockRate'
 │ │   451 |   props.emplace_back("gpu_sm_freq", std::to_string(device_prop.clockRate * 1e3));
 │ │       |                                                                ^~~~~~~~~
 │ │ $SRC_DIR/cpp/bench/ann/src/hnswlib/../common/util.hpp:452:65: error: 'struct cudaDeviceProp' has no member named 'memoryClockRate'
 │ │   452 |   props.emplace_back("gpu_mem_freq", std::to_string(device_prop.memoryClockRate * 1e3));
 │ │       |                                                                 ^~~~~~~~~~~~~~~
```

([conda-cpp-build link](https://github.com/rapidsai/cuvs/actions/runs/17161321696/job/48691211150?pr=1273#step:10:1783))

jameslamb:
Updated to latest `branch-25.10` and pushed this patch from @robertmaynard : https://github.com/rapidsai/cuvs/pull/1273/commits/d1228561376a9624acd473bc1c49838c7dfe716f

Let's see how that goes. Should get faster feedback now that we've mostly filled up the `sccache` caches for CUDA 13 builds here 😁

cpp/bench/ann/src/common/util.hpp
```diff
@@ -446,10 +446,15 @@ inline auto cuda_info()
     throw std::runtime_error{"cuda_info: call to cudaGetDeviceProperties failed with code " +
                              std::to_string(err_code)};
   }
+  int clockRate       = 0;
+  int memoryClockRate = 0;
+  err_code            = cudaDeviceGetAttribute(&clockRate, cudaDevAttrClockRate, dev);
+  err_code            = cudaDeviceGetAttribute(&memoryClockRate, cudaDevAttrMemoryClockRate, dev);
```
jameslamb, line 452: The Java builds are failing like this:

```text
[INFO] -------------------------------------------------------------
Error:  /__w/cuvs/cuvs/java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/common/Util.java:[192,23] cannot find symbol
  symbol:   method cudaGetDeviceProperties_v2(java.lang.foreign.MemorySegment,int)
  location: class com.nvidia.cuvs.internal.common.Util
[INFO] 1 error
[INFO] -------------------------------------------------------------
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  13.340 s
[INFO] Finished at: 2025-08-22T21:20:16Z
[INFO] ------------------------------------------------------------------------
Error:  Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.13.0:compile (compile-java-22) on project cuvs-java: Compilation failure
Error:  /__w/cuvs/cuvs/java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/common/Util.java:[192,23] cannot find symbol
Error:    symbol:   method cudaGetDeviceProperties_v2(java.lang.foreign.MemorySegment,int)
Error:    location: class com.nvidia.cuvs.internal.common.Util
```

([build link](https://github.com/rapidsai/cuvs/actions/runs/17165578398/job/48706812601?pr=1273#step:11:5832))

@mythrocks do the Java bindings need to be updated? If so could you do that and push them to this PR?

I don't have Java set up in my development environment.

mythrocks: Hmm. This is exactly the sort of error that I'm running into in my work on the portable fat-jar PR.  Except, I'm building on 12.9. 

I think this is an artifact of some strange `jextract` behaviour. 

Thanks for the confirmation; so I'm not imagining it. 

I'll dig into this more.

ldematte: As a first step I would try to see if `cudaGetDeviceProperties_v2(java.lang.foreign.MemorySegment,int)` is defined and what signature it has.
If this passes on 12.9 but fails consistently in 13.0, I suspect this is a header issue.
We generate functions bindings for Java using `jextract`, but this may or may not be the issue. I suspect CUDA headers `#define` cudaGetDeviceProperties as "something", and that "something" as a different signature in 12 vs 13. Just a guess, let me see if I can validate it.

ldematte: (Java, as the other non-C langs, can bind to C functions but of course not via macros - that's why we need to call `cudaGetDeviceProperties_v2` directly -- again, modulo my guesswork being correct and cudaGetDeviceProperties is indeed a macro)

ldematte: Yep, I see `#define cudaGetDeviceProperties cudaGetDeviceProperties_v2` in my cuda_runtime_api.h. Now let me see how it looks like in 13.0

ldematte: And yes, it does seem that in CUDA 13 it is no more a macro, there is no v2:
```
extern __host__ __cudart_builtin__ cudaError_t CUDARTAPI cudaGetDeviceProperties(struct cudaDeviceProp *prop, int device);
```
So this needs to change in Java too. I think we do not keep separate branches for CUDA 12 vs 13 right? So we'll need something in Java to distinguish them too. We could use environment variables if CMake defines them --  I think we can use RAPIDS_CUDA_MAJOR?
And we would need to define the MH ourselves, so we bind to the correct function based on the version. 
Let me show you in a draft.

ldematte: Untested, but should work with your branch, if you can test it this draft/patch should address the issue (or something along these lines): https://patch-diff.githubusercontent.com/raw/rapidsai/cuvs/pull/1280.diff

jameslamb: Thanks for looking into this! I just pushed commits to test that, let's see how it goes.

I am concerned about this though:

```diff
+  private static final String cudaGetDevicePropertiesSymbolName =
+      "12".equals(System.getenv("RAPIDS_CUDA_MAJOR"))
+          ? "cudaGetDeviceProperties_v2"
+          : "cudaGetDeviceProperties";
```

Not a problem here in CI, but that `RAPIDS_CUDA_MAJOR` environment variable probably won't exist in the environments where developers are building from source, and you'll need to define it.

Hopefully @mythrocks can help suggest a different way to detect the target CUDA major version at build time in the Java builds here.

ldematte: Yep, I only grabbed the first env variable I saw :) 
I'm definitely not sure myself this is correct, or if we should use something else (possibly define a new env variable via the Java CMake?)

mythrocks: I haven't looked into it closely yet, but what surprises me is that in my repro, this was happening with CTK 12.9, not just 13.0.  One wonders if there's something more at play.

jameslamb: It does look like this patch allowed the build to succeed with both CTK 12.9 and 13.0: https://github.com/rapidsai/cuvs/actions/runs/17210987762/job/48824942399?pr=1273

*(the 12.9 job there failed with an error about CI artifact name conflicts which I can fix, but it successfully compiled)*

I want to try to move most CUDA 13 support in `cuvs` forward today, and not block conda packages, wheels, etc. on this.

Which of these would you prefer?

* merge this patch in for now and remove in a follow-up
* remove this patch and skip CUDA 13 Java PRs for now

mythrocks: Let's accept this for today. 

If we find a better solution, we'll follow up.  No sense in holding up progress.

jameslamb: Alright thanks! And at least merging this as-is will allow you to catch other changes breaking CUDA 13 support.

I'll fix the CI artifacts stuff and then put this up for review. Appreciate both of you jumping in to help so quickly!

jameslamb: Ok just pushed a change that I think will help. Now instead of being called `result`, the content of `java/cuvs-java/target` will be zipped up and uploaded as an artifact named `cuvs-java-cuda{cuda_version}`, e.g. `cuvs-java-cuda12.9.1`.

Pretty cool that a feature request for our CI tooling from `cuopt` (@rgsl888prabhu , https://github.com/rapidsai/shared-workflows/issues/392) ended up enabling this change here in `cuvs`! It's nice when things work out like that 😁

mythrocks:
This looks good to me right now.  However, I would like to address the issue that @cjnolet pointed out in #1280.

I'll try whip something up today.  Unless someone else beats me to it.

mythrocks:
I have something that works to reduce the number of times we `cudaGetDeviceProperties()`.  But I'd prefer not to raise a PR for this bit.  I'd like to reduce churn all around. So here's what I propose:
1. This current PR goes in, as is. Looks good to me, and is no worse than the current state w.r.t `cudaGetDeviceProperties()`.
2. We address caching as part of #1267.  It would be incorrect to cache `GpuInfo` objects so long as they report the current free-memory levels.  @ldematte is refactoring this so that the runtime-variable info is removed from `GpuInfo`.  We can cache things there.

mythrocks:
FWIW, a quick and dirty caching solution might replace the current implementation of `Util.availableGPUs()` with the following:
```java
  // Lazy initialization for list of available GPUs.
  private static class AvailableGpuInitializer {

    // Available GPUs are initialized only once when first accessed.
    // This is assumed to be invariant for the lifetime of the program.
    static final List<GPUInfo> AVAILABLE_GPUS = availableGPUs();

    private static List<GPUInfo> availableGPUs() {

        MemorySegment numGpus = localArena.allocate(C_INT);
        int returnValue = cudaGetDeviceCount(numGpus);
        checkCudaError(returnValue, "cudaGetDeviceCount");

        int numGpuCount = numGpus.get(C_INT, 0);
        List<GPUInfo> gpuInfoArr = new ArrayList<GPUInfo>();

        MemorySegment free = localArena.allocate(size_t);
        MemorySegment total = localArena.allocate(size_t);
        MemorySegment deviceProp = cudaDeviceProp.allocate(localArena);

        for (int i = 0; i < numGpuCount; i++) {

          returnValue = cudaSetDevice(i);
          checkCudaError(returnValue, "cudaSetDevice");

          returnValue = cudaGetDeviceProperties(deviceProp, i);
          checkCudaError(returnValue, "cudaGetDeviceProperties");

          returnValue = cudaMemGetInfo(free, total);
          checkCudaError(returnValue, "cudaMemGetInfo");

          float computeCapability =
              Float.parseFloat(
                  cudaDeviceProp.major(deviceProp) + "." + cudaDeviceProp.minor(deviceProp));

          GPUInfo gpuInfo =
              new GPUInfo(
                  i,
                  cudaDeviceProp.name(deviceProp).getString(0),
                  free.get(C_LONG, 0), // TODO:  #1267 will remove free memory from GPUInfo.  Truly invariant then.
                  total.get(C_LONG, 0),
                  computeCapability);

          gpuInfoArr.add(gpuInfo);
        }
        return gpuInfoArr;
      }
    }
  }

  /**
   * Gets all the available GPUs
   *
   * @return a list of {@link GPUInfo} objects with GPU details
   */
  private static List<GPUInfo> availableGPUs() {
    return AvailableGpuInitializer.AVAILABLE_GPUS;
  }
```

ldematte:
> This current PR goes in, as is. Looks good to me, and is no worse than the current state w.r.t cudaGetDeviceProperties().

Agreed. Plus, I think the functions that calls into `cudaGetDeviceProperties` seems to be unused ATM.

> We address caching as part of
https://github.com/rapidsai/cuvs/pull/1267. It would be incorrect to cache GpuInfo objects so long as they report the current free-memory levels. @ldematte is refactoring this so that the runtime-variable info is removed from GpuInfo. We can cache things there.

++, see comments on https://github.com/rapidsai/cuvs/pull/1267

docs/source/build.rst
```diff
@@ -61,7 +61,7 @@ For CUDA 12 packages:
 
 .. code-block:: bash
 
-    pip install cuvs-cu12 --extra-index-url=https://pypi.nvidia.com
+    pip install cuvs-cu13 --extra-index-url=https://pypi.nvidia.com
```
divyegala, line 71: Above this, it says "For CUDA 12 packages"

python/cuvs/pyproject.toml
```diff
@@ -117,7 +117,7 @@ regex = "(?P<value>.*)"
 [tool.rapids-build-backend]
 requires = [
     "cmake>=3.30.4",
-    "cuda-python",
+    "cuda-python>=13.0.1,<14.0a0",
```
divyegala, line 120: I am forgetting, can you please explain how or where this is overwritten for CUDA 12 build?

jameslamb: Sure.

See the use of `cuda: "12.*"` here:

https://github.com/jameslamb/cuvs/blob/9fe24a7a04b432f9acb15c771d0d0a4b627367ec/dependencies.yaml#L498-L509

When `cuvs` is built, `rapids-build-backend` will detect the CUDA major version by querying `nvcc --version`:

https://github.com/rapidsai/rapids-build-backend/blob/f3681f71e5170b48990a067fafc065bda6740763/rapids_build_backend/impls.py#L88

And then use that to set the filter `cuda={cuda_major}` in a call to `rapids_dependency_file_generator` (its Python API, not the CLI):

https://github.com/rapidsai/rapids-build-backend/blob/f3681f71e5170b48990a067fafc065bda6740763/rapids_build_backend/impls.py#L230-L240

That filtering will select the CUDA 12 dependencies for the wheel's metadata.

docs/source/build.rst
jameslamb: sorry about that, thanks for catching it.

Fixed in https://github.com/rapidsai/cuvs/pull/1273/commits/3b6b8d695c5c46f2d0a490b876e997f9328757cc

python/cuvs/pyproject.toml
divyegala: Thanks so much!

msarahan (approved):
approving for code-owners

jameslamb:
Merging in recent changes, I now see build failures for `conda-cpp-build` jobs on CUDA 13:

```text
 │ │ $SRC_DIR/cpp/include/cuvs/neighbors/common.hpp(465): error #20011-D: calling a __host__ function("cuvs::neighbors::filtering::base_filter::~base_filter()") from a __host__ __device__ function("cuvs::neighbors::filtering::base_filter::~base_filter [subobject]") is not allowed
 │ │ Remark: The warnings can be suppressed with "-diag-suppress <warning-number>"
 │ │ 1 error detected in the compilation of "$SRC_DIR/cpp/src/neighbors/cagra_search_uint8.cu".
 │ │ [4/766] Building CUDA object CMakeFiles/cuvs-cagra-search.dir/src/neighbors/cagra_search_int8.cu.o
 │ │ FAILED: [code=1] CMakeFiles/cuvs-cagra-search.dir/src/neighbors/cagra_search_int8.cu.o 
```

([build link](https://github.com/rapidsai/cuvs/actions/runs/17279483303/job/49044594134?pr=1273#step:10:1164))

I strongly suspect that's a result of #1211. @jinsolp could you please take a look? There are CUDA 13 devcontainers on this branch that you might find helpful for investigating.

mythrocks (approved):
Approving for Java-owners.  @jameslamb: Good work, mate.

docs/source/build.rst
```diff
@@ -43,24 +43,26 @@ C/C++ Package
...
-   conda install -c rapidsai -c conda-forge libcuvs cuda-version=12.9
+   conda install -c rapidsai -c conda-forge libcuvs cuda-version=13.0
 
 Python Package
 ~~~~~~~~~~~~~~
 
 .. code-block:: bash
 
-   conda install -c rapidsai -c conda-forge cuvs cuda-version=12.9
+   conda install -c rapidsai -c conda-forge cuvs cuda-version=13.0
```
divyegala: Can we add "For CUDA 12" and "For CUDA 13" commands here as well to have uniformity with the Pip commands section?

divyegala (approved):
Just a single comment, pre-approving. Thanks!

docs/source/build.rst
jameslamb: Sure, happy to. I'll push that in the next commit (assuming we'll need another one to fix https://github.com/rapidsai/cuvs/pull/1273#issuecomment-3229889041)

robertmaynard:
> Merging in recent changes, I now see build failures for `conda-cpp-build` jobs on CUDA 13:
> 
> ```
>  │ │ $SRC_DIR/cpp/include/cuvs/neighbors/common.hpp(465): error #20011-D: calling a __host__ function("cuvs::neighbors::filtering::base_filter::~base_filter()") from a __host__ __device__ function("cuvs::neighbors::filtering::base_filter::~base_filter [subobject]") is not allowed
>  │ │ Remark: The warnings can be suppressed with "-diag-suppress <warning-number>"
>  │ │ 1 error detected in the compilation of "$SRC_DIR/cpp/src/neighbors/cagra_search_uint8.cu".
>  │ │ [4/766] Building CUDA object CMakeFiles/cuvs-cagra-search.dir/src/neighbors/cagra_search_int8.cu.o
>  │ │ FAILED: [code=1] CMakeFiles/cuvs-cagra-search.dir/src/neighbors/cagra_search_int8.cu.o 
> ```
> 
> ([build link](https://github.com/rapidsai/cuvs/actions/runs/17279483303/job/49044594134?pr=1273#step:10:1164))
> 
> I strongly suspect that's a result of #1211. @jinsolp could you please take a look? There are CUDA 13 devcontainers on this branch that you might find helpful for investigating.

Yes, absolutely caused by #1211, since it removed the changes I made in https://github.com/rapidsai/cuvs/pull/1219 to allow for CUDA 13 compilation.

jameslamb:
Thanks for looking and confirming @robertmaynard ! If you know how to fix this, could you push a patch to my branch here?

docs/source/build.rst
jameslamb: fixed in https://github.com/rapidsai/cuvs/pull/1273/commits/d4e3154272cfb06b2b813608e714cb228eebab99

cpp/bench/ann/src/common/util.hpp
jameslamb: Looks like this was broken by #1267

```text
Error:  Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.13.0:compile (compile-java-22) on project cuvs-java: Compilation failure
Error:  /__w/cuvs/cuvs/java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/GPUInfoProviderImpl.java:[55,25] cannot find symbol
Error:    symbol:   method cudaGetDeviceProperties_v2(java.lang.foreign.MemorySegment,int)
Error:    location: class com.nvidia.cuvs.internal.GPUInfoProviderImpl.AvailableGpuInitializer
Error:  
Error:  -> [Help 1]
```

([conda-java-tests link](https://github.com/rapidsai/cuvs/actions/runs/17310232795/job/49146837737?pr=1273#step:11:5782))

I'll try to push a fix.

jameslamb: I tried something here: https://github.com/rapidsai/cuvs/pull/1273/commits/104508b31edede3e008880e20b295c02b20b8dda

Would appreciate your input on it @ldematte @mythrocks . If you look and that seems wrong, please feel free to push to my branch here. I'd really like to get this PR in as soon as possible, so we can move on to other projects in RAPIDS that depend on `cuvs`.

jameslamb: Ok one more commit (https://github.com/rapidsai/cuvs/pull/1273/commits/8530e22090568f7fe8ca2b1a5c90688493ec2681) because I missed that that was a static method (sorry, I haven't written anything for the JVM in a while 😅)

And looks like that work!

I see this as a minor extension of logic you all were already ok with putting in temporarily, so I'm planning to merge this once CI passes. Thanks again for all the help!

mythrocks: I'm not sure this was borked by #1267, exactly.  I thought your change [here](https://github.com/rapidsai/cuvs/pull/1273#discussion_r2298197046) would've dealt with it.

That said, your fix seems right.  +1 to checking this in on passing CI.

mythrocks:
CI seems to have borked on an intermittent test failure.  I'm going to re-kick it.

cpp/bench/ann/src/common/util.hpp
jameslamb: Thanks!

I think it was.... we found in this thread that `cudaGetDeviceProperties_v2()` could not be used unconditionally, and #1267 introduces code that uses it unconditionally:

https://github.com/rapidsai/cuvs/blob/e02b7d1280264cad19885b309590e5cdddbe0efe/java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/GPUInfoProviderImpl.java#L55-L56

jameslamb:
Thanks very much @mythrocks for re-running CI and to you and everyone else for your help on this!!!!

I'm going to merge this, `@` me any time if it seems like it caused issues.

PR #1278, opened 2025-08-23
Add support for PQ preprocessing API

### LINKED ISSUE #107: [FEA] Product quantization API
We need a separate product quantization API that is decoupled from IVF but can still be composed into IVF. 

Ideally this API would follow FAISS or Scikit-learn'a transformer estimators.

lowener:
Closes #107

This PR adds support for a PQ preprocessing API. It gives access to `train()`, `transform()` and `inverse_transform()` function that can be used to transform a dataset into PQ codes. It is re-using the VPQ functions from CAGRA-Q.


cpp/include/cuvs/preprocessing/quantize/product.hpp
```diff
@@ -0,0 +1,164 @@
+/*
+ * Copyright (c) 2024-2025, NVIDIA CORPORATION.
```
cjnolet: Please only include the current year in new files

cpp/include/cuvs/preprocessing/quantize/product.hpp
```diff
@@ -0,0 +1,164 @@
...
+ * auto quantizer = cuvs::preprocessing::quantize::product::train(handle, params, dataset);
+ * @endcode
+ *
+ * @param[in] res raft resource
+ * @param[in] params configure product quantizer, e.g. quantile
+ * @param[in] dataset a row-major matrix view on host
+ *
+ * @return quantizer
+ */
+/*quantizer train(raft::resources const& res,
```
cjnolet: Just making note of commented out code for cleanup before this is merged

cpp/include/cuvs/preprocessing/quantize/product.hpp
```diff
@@ -0,0 +1,164 @@
...
+ * @param[out] out a row-major matrix view on device
+ *
+ */
+void transform(raft::resources const& res,
+               const quantizer& quantizer,
+               raft::device_matrix_view<const float, int64_t> dataset,
+               raft::device_matrix_view<uint8_t, int64_t> out);
+
+/** @copydoc transform */
+/*void transform(raft::resources const& res,
```
cjnolet: Making note of commented out code for tracking

cpp/src/preprocessing/quantize/detail/product.cuh
```diff
@@ -0,0 +1,164 @@
...
+
+  // TODO: Call detail::extend to avoid clone()
+  auto extended_index =
+    cuvs::neighbors::ivf_pq::extend(res, dataset, indices_view_opt, quantizer.pq_index);
+
+  auto n_lists        = extended_index.n_lists();
+  auto lists          = extended_index.lists();
+  auto indices_uint32 = rmm::device_uvector<uint32_t>(0, raft::resource::get_cuda_stream(res));
+  for (size_t i = 0; i < n_lists; i++) {
+    /*auto current_list_size = lists[i]->size.load();
```
cjnolet: Commented out code

cpp/src/preprocessing/quantize/detail/product.cuh
```diff
@@ -0,0 +1,164 @@
...
+  }
+  // TODO: Resize the extended index lists to 0
+}
+
+/*
+template <typename T>
+quantizer train(
+  raft::resources const& res,
+  const params params,
+  raft::host_matrix_view<const T, int64_t> dataset)
```
cjnolet: Commented out code

cpp/src/preprocessing/quantize/product.cu
```diff
@@ -0,0 +1,83 @@
...
+
+#define CUVS_INST_QUANTIZATION(T, QuantI)                                         \
+  auto train(raft::resources const& res,                                          \
+             const params params,                                                 \
+             raft::device_matrix_view<const T, int64_t> dataset) -> quantizer     \
+  {                                                                               \
+    return detail::train(res, params, dataset);                                   \
+  }                                                                               \
+  /*auto train(raft::resources const& res,                                        \
+             const params params,                                                 \
```
cjnolet: Commented out code (lots in this file)

cpp/tests/preprocessing/product_quantization.cu
```diff
@@ -0,0 +1,307 @@
...
+  {500, 40, 5, 8, cuvs::neighbors::ivf_pq::codebook_gen::PER_CLUSTER, false, 10, 42ULL},
+  {500, 60, 7, 8, cuvs::neighbors::ivf_pq::codebook_gen::PER_SUBSPACE, false, 10, 42ULL},
+
+  // Larger dataset
+  {1000, 40, 4, 4, cuvs::neighbors::ivf_pq::codebook_gen::PER_SUBSPACE, true, 20, 42ULL},
+  {3000, 1024, 4, 20, cuvs::neighbors::ivf_pq::codebook_gen::PER_CLUSTER, false, 10, 42ULL},
+  {1000, 2048, 4, 20, cuvs::neighbors::ivf_pq::codebook_gen::PER_SUBSPACE, false, 10, 42ULL},
+
+  // Benchmark datasets
+  //{500000, 2048, 8, 20, cuvs::neighbors::ivf_pq::codebook_gen::PER_SUBSPACE, false, 1000, 42ULL},
```
cjnolet: Commented out code

cpp/include/cuvs/preprocessing/quantize/product.hpp
```diff
@@ -0,0 +1,164 @@
...
+ * @brief Product Quantizer parameters.
+ */
+struct params {
+  /*
+   * The bit length of the vector element after compression by PQ.
+   *
+   * Possible values: [4, 5, 6, 7, 8].
+   */
+  int64_t pq_bits = 8;
+  int64_t pq_dim  = 0;
```
cjnolet: Will need to document these parameters as well

cpp/src/preprocessing/quantize/detail/product.cuh
```diff
@@ -0,0 +1,98 @@
...
+  auto pq_params                         = cuvs::neighbors::ivf_pq::index_params();
+  pq_params.n_lists                      = params.n_lists;
+  pq_params.pq_bits                      = params.pq_bits;
+  pq_params.pq_dim                       = params.pq_dim;
+  pq_params.codebook_kind                = params.codebook_kind;
+  pq_params.force_random_rotation        = params.force_random_rotation;
+  pq_params.add_data_on_build            = false;
+  pq_params.max_train_points_per_pq_code = params.max_train_points_per_pq_code;
+
+  auto pq_index = cuvs::neighbors::ivf_pq::build(res, pq_params, dataset);
```
gland1: I've tried using ivf build + pq-centeres api to get the generated centroids as part of generating pq for diskann. It appears I get significant lower recall values. I think it because the pq strategy used by ivf has 2 quantizers and also ivf uses balanced kmeans which can lower the recall.

cpp/include/cuvs/preprocessing/quantize/product.hpp
```diff
@@ -0,0 +1,122 @@
...
+ * auto quantizer = cuvs::preprocessing::quantize::product::train(handle, params, dataset);
+ * @endcode
+ *
+ * @param[in] res raft resource
+ * @param[in] params configure product quantizer, e.g. quantile
+ * @param[in] dataset a row-major matrix view on device
+ *
+ * @return quantizer
+ */
+quantizer train(raft::resources const& res,
```
gland1: I think you should also add an API to get the centroids after training, in case assignment is done elsewhere

KyleFromNVIDIA (approved):

cpp/include/cuvs/preprocessing/quantize/product.hpp
lowener: The centroids are available through `quantizer.vpq_codebooks.pq_code_book`

cpp/tests/CMakeLists.txt
```diff
@@ -285,8 +285,17 @@ if(BUILD_TESTS)
...
+    PREPROCESSING_TEST
+    PATH
+    preprocessing/scalar_quantization.cu
+    preprocessing/binary_quantization.cu
+    preprocessing/product_quantization.cu
+    preprocessing/spectral_embedding.cu
+    GPUS
+    1
+    PERCENT
+    100
```
KyleFromNVIDIA: Not a show-stopper for this PR, but we should probably update `cpp/cmake/config.json` to take `ConfigureTest()` into account so we don't get ugly argument lists like this.

KyleFromNVIDIA: I've opened https://github.com/rapidsai/cuvs/pull/1385. We don't have to wait on it to be merged to merge this PR, but if it does get merged first, we need to merge it into this PR before merging this PR.

cpp/src/neighbors/detail/vpq_dataset.cuh
```diff
@@ -200,19 +201,21 @@ auto train_pq(const raft::resources& res,
...
     using index_type = typename DatasetT::index_type;
     raft::linalg::map_offset(
       res,
       pq_trainset.view(),
-      [labels = vq_labels.view(), centers = vq_centers, dim] __device__(index_type off, MathT x) {
+      [labels = vq_labels.view(), centers = vq_centers.value(), dim] __device__(index_type off,
+                                                                                MathT x) {
         index_type i = off / dim;
         index_type j = off % dim;
         return x - centers(labels(i), j);
```
tarang-jain, line 253: If this is a general PQ quantization API, coarse centroids would not be necessary, right? In a number of applications, raw vectors are directly product quantized, not the residuals.

cpp/src/preprocessing/quantize/detail/product.cuh
tarang-jain: Based on my understanding, the current VPQ also has two quantizers. @lowener correct me if I am missing something, but based on what I had seen in DiskANN, the raw vectors are PQ quantized directly, there is no coarse quantizer (VQ centroids).

cpp/src/neighbors/detail/vpq_dataset.cuh
lowener: Yes, this implementation focus only on direct product quantization.

tarang-jain: Looks like we are computing the residual here, right? I would assume we don't require `vq_centers` at all. Lmk if my understanding is correct!

cpp/tests/CMakeLists.txt
KyleFromNVIDIA: https://github.com/rapidsai/cuvs/pull/1385 has been merged. I've just merged it in.

cpp/src/preprocessing/quantize/detail/product.cuh
```diff
@@ -0,0 +1,177 @@
...
+  auto n_rows = dataset.extent(0);
+  auto dim    = dataset.extent(1);
+  raft::common::nvtx::range<cuvs::common::nvtx::domain::cuvs> fun_scope(
+    "preprocessing::quantize::product::train(%zu, %u)", size_t(n_rows), dim);
+
+  auto ps = cuvs::neighbors::detail::fill_missing_params_heuristics(params, dataset);
+  auto vq_code_book =
+    raft::make_device_matrix<T, uint32_t, raft::row_major>(res, ps.vq_n_centers, dim);
+  raft::matrix::fill(res, vq_code_book.view(), T{0.0f});
+  auto pq_code_book = cuvs::neighbors::detail::train_pq<T>(res, ps, dataset, std::nullopt);
```
gland1: from what I see in the current code this method uses balanced_kmeans. From experiment I've done,
balanced kmeans leads to inferior recall vs normal kmeans. is it possible to give an option to use normal kmeans?

lowener: Yes I added that to the parameters to let users choose 👍

c/include/cuvs/preprocessing/quantize/pq.h
```diff
@@ -0,0 +1,149 @@
...
+ */
+cuvsError_t cuvsProductQuantizerGetPqBits(cuvsProductQuantizer_t quantizer, uint32_t* pq_bits);
+
+/**
+ * @brief Get the dimensionality of the vector after compression by PQ.
+ *
+ * @param[in] quantizer product quantizer
+ * @param[out] pq_dim dimensionality of the vector after compression by PQ
+ */
+cuvsError_t cuvsProductQuantizerGetPqDim(cuvsProductQuantizer_t quantizer, uint32_t* pq_dim);
```
cjnolet, line 187: We're also going to need accessors for the underlying codebooks. I understand they are in a struct, but the goal w/ C apis is to make that state structs as opaque as possible.

cpp/src/preprocessing/quantize/detail/product.cuh
gland1: Great, Thanks!

c/include/cuvs/cluster/kmeans.h
```diff
@@ -132,6 +132,11 @@ cuvsError_t cuvsKMeansParamsCreate(cuvsKMeansParams_t* params);
  */
 cuvsError_t cuvsKMeansParamsDestroy(cuvsKMeansParams_t params);
 
+/**
+ * @brief Type of k-means algorithm.
+ */
+typedef enum { KMeans, KMeansBalanced } cuvsKMeansType;
```
robertmaynard: All enums in the C API need explicit values

c/include/cuvs/preprocessing/quantize/product.h
```diff
@@ -0,0 +1,163 @@
...
+struct cuvsProductQuantizerParams {
+  /**
+   * The bit length of the vector element after compression by PQ.
+   *
+   * Possible values: [4, 5, 6, 7, 8].
+   *
+   * Hint: the smaller the 'pq_bits', the smaller the index size and the better the search
+   * performance, but the lower the recall.
+   */
+  uint32_t pq_bits = 8;
```
robertmaynard: C doesn't allow default values for struct members. You need to move all of these into the `cuvsProductQuantizerParamsCreate` function

cpp/include/cuvs/cluster/kmeans.hpp
```diff
@@ -136,6 +136,11 @@ struct balanced_params : base_params {
   uint32_t n_iters = 20;
 };
 
+/**
+ * @brief Type of k-means algorithm.
+ */
+enum class kmeans_type { KMeans, KMeansBalanced };
```
robertmaynard: Please make these explicit values as well

cpp/src/preprocessing/quantize/detail/product.cuh
lowener: I added VQ as a first-level quantizer

cpp/include/cuvs/neighbors/common.hpp
```diff
@@ -65,8 +70,26 @@ struct vpq_params {
...
+   * Whether to use vector quantization (VQ) before product quantization (PQ).
+   * When true, VQ is used before PQ.
+   */
+  bool use_vq = false;
+  /**
+   * Whether to use subspaces for product quantization (PQ).
+   * When true, one PQ codebook is used for each subspace. Otherwise, a single
+   * PQ codebook is used.
+   */
+  bool use_subspaces = true;
```
tfeher: The default values here are different then what CAGRA-Q would select as default before this PR. Therefore this change would affect CAGRA-Q users. This might be ok, since we have more algorithms using the same parameter structure. But we should review CAGRA-Q parameter setting, and make sure we have good default setting for that too.

cpp/include/cuvs/preprocessing/quantize/pq.hpp
```diff
@@ -0,0 +1,118 @@
...
+  cuvs::neighbors::vpq_dataset<T, int64_t> vpq_codebooks;
+};
+
+/**
+ * @brief Initializes a product quantizer to be used later for quantizing the dataset.
+ *
+ * Usage example:
+ * @code{.cpp}
+ * raft::handle_t handle;
+ * cuvs::preprocessing::quantize::pq::params params;
```
tfeher, line 101: According to our offline discussion, here we might want to update the example to show how a pool allocator can be used to improve training time.

cpp/include/cuvs/preprocessing/quantize/pq.hpp
```diff
@@ -0,0 +1,118 @@
...
+ *
+ * @param[in] res raft resource
+ * @param[in] params configure product quantizer, e.g. quantile
+ * @param[in] dataset a row-major matrix view on device
+ *
+ * @return quantizer
+ */
+quantizer<float> train(raft::resources const& res,
+                       const params params,
+                       raft::device_matrix_view<const float, int64_t> dataset);
```
tfeher, line 113: Shouldn't we support host dataset as well?

cpp/src/neighbors/detail/vamana/vamana_build.cuh
```diff
@@ -560,9 +560,17 @@ auto quantize_all_vectors(raft::resources const& res,
   auto dim         = residuals.extent(1);
   auto vq_codebook = raft::make_device_matrix<float, uint32_t, raft::row_major>(res, 1, dim);
   raft::matrix::fill<float>(res, vq_codebook.view(), 0.0);
-
-  auto codes = cuvs::neighbors::detail::process_and_fill_codes_subspaces<float, int64_t>(
-    res, ps, residuals, raft::make_const_mdspan(vq_codebook.view()), pq_codebook);
+  auto codes_rowlen = sizeof(uint32_t) * (1 + raft::div_rounding_up_safe<int64_t>(
```
tfeher: Note that Vamana actually needs quantized vector without the vector quantization, therefore [here](https://github.com/rapidsai/cuvs/blob/91c51b1cc43d45cc3d949830e153725de5a2c972/cpp/src/neighbors/detail/vamana/vamana_build.cuh#L703-L714) we remove the cluster idx. The new PQ preprocessing API should be used to do this without the additional copy operation. Could you create an issue to track this?

cpp/tests/preprocessing/product_quantization.cu
```diff
@@ -0,0 +1,251 @@
...
+  raft::device_matrix<T, int64_t, raft::row_major> dataset_;
+  raft::device_vector<int64_t, int64_t, raft::row_major> labels_;
+  params config_;
+};
+
+// Define test cases with different parameters
+template <typename T>
+const std::vector<ProductQuantizationInputs<T>> inputs = {
+  // Small dataset
+  {100, 64, 4, 8, cuvs::cluster::kmeans::kmeans_type::KMeansBalanced, 0, false, false, 42ULL},
```
tfeher: Could you add a test extreme cases: 1 vector, 1 dim, etc.

cpp/src/preprocessing/quantize/detail/pq.cuh
```diff
@@ -0,0 +1,353 @@
...
+    vq_code_book_view_opt                                                           = std::nullopt;
+  std::optional<raft::device_matrix<T, uint32_t, raft::row_major>> vq_code_book_opt = std::nullopt;
+  if (ps.use_vq) {
+    vq_code_book_opt      = cuvs::neighbors::detail::train_vq<T>(res, ps, dataset);
+    vq_code_book_view_opt = raft::make_const_mdspan(vq_code_book_opt.value().view());
+  } else {
+    vq_code_book_opt = raft::make_device_matrix<T, uint32_t, raft::row_major>(res, 0, 0);
+  }
+  auto empty_codes = raft::make_device_matrix<uint8_t, int64_t, raft::row_major>(res, 0, 0);
+  if (ps.use_subspaces) {
```
tfeher: It would be great to use the PQ quantizer to do the quantization for IVF-PQ. Currently the output layout is different, but in the future we should unify these code paths. Could you create an issue to track that?

tfeher (changes requested):
Thanks @lowener for the PR! I still have to go through the implementation details of the new kernels, but overall the PR looks good. Here are my first batch of comments.

cpp/src/neighbors/detail/vamana/vamana_build.cuh
lowener: I added it to this issue on follow-up of the PR:
https://github.com/rapidsai/cuvs/issues/1575#issuecomment-3627592544

cpp/include/cuvs/preprocessing/quantize/pq.hpp
lowener: I added support for host dataset 👍

cpp/include/cuvs/neighbors/common.hpp
lowener: I changed the params to inherit those from CAGRA-Q. `use_vq`  and `use_subspaces` are now specific to the PQ API. The only difference for applications already using that code is that it will use kmeans-balanced and 256 points per pq code.

cpp/src/preprocessing/quantize/detail/pq.cuh
lowener: I created this issue: https://github.com/rapidsai/cuvs/issues/1648

cpp/include/cuvs/preprocessing/quantize/pq.hpp
```diff
@@ -0,0 +1,152 @@
...
+  if (config.use_vq) {
+    return sizeof(LabelT) * (1 + raft::div_rounding_up_safe<int64_t>(config.pq_dim * config.pq_bits,
+                                                                     8 * sizeof(LabelT)));
+  } else {
+    return raft::div_rounding_up_safe<int64_t>(config.pq_dim * config.pq_bits, 8);
+  }
+}
+
+/**
+ * @brief Applies inverse quantization transform to given dataset
```
cjnolet, line 168:
```suggestion
 * @brief Applies inverse quantization transform to given set of encoded vectors
```

c/include/cuvs/cluster/kmeans.h
```diff
@@ -121,6 +121,11 @@ cuvsError_t cuvsKMeansParamsCreate(cuvsKMeansParams_t* params);
  */
 cuvsError_t cuvsKMeansParamsDestroy(cuvsKMeansParams_t params);
 
+/**
+ * @brief Type of k-means algorithm.
+ */
+typedef enum { KMeans = 0, KMeansBalanced = 1 } cuvsKMeansType;
```
cjnolet: Future task- we should update `cuvsKMeansParams_t` to use this and remove the `bool hierarchical` flag. Can you create an issue for this just so we don't forget it? It doesn't have to be done now, especially since we're striving to maintain ABI compatibility in our C APIs (we need to start deprecating breaking changes).

c/src/preprocessing/quantize/pq.cpp
```diff
@@ -0,0 +1,295 @@
...
+ */
+
+#include <cstdint>
+#include <dlpack/dlpack.h>
+
+#include <cuvs/core/c_api.h>
+#include <cuvs/preprocessing/quantize/pq.h>
+#include <cuvs/preprocessing/quantize/pq.hpp>
+#include "../../core/exceptions.hpp"
+#include "../../core/interop.hpp"
```
divyegala, line 13: Why relative includes?

cpp/include/cuvs/neighbors/common.hpp
```diff
@@ -65,8 +70,22 @@ struct vpq_params {
    * When zero, an optimal value is selected using a heuristic.
    */
   double pq_kmeans_trainset_fraction = 0;
+  /**
+   * Type of k-means algorithm for PQ training.
+   */
```
divyegala, line 80: I think it might be worth talking a little about when to choose which KMeans type.

cpp/src/neighbors/detail/vamana/vamana_codebooks.cuh
```diff
@@ -9,6 +9,7 @@
 
 #include <raft/core/error.hpp>
 
+#include <fstream>
```
divyegala, line 12: Why was this include needed?

cpp/src/neighbors/detail/vpq_dataset.cuh
```diff
@@ -262,52 +349,112 @@ __device__ auto compute_code(
...
+      }
+    }
+  }
+  if constexpr (VecElems <= 1) {
+    for (uint32_t i = threadIdx.x; i < n; i += blockDim.x) {
+      out[i] = in[i];
+    }
+  }
+}
+
```
divyegala: This code seems to be shared with ivf_flat interleaved scan. Please share it as a utility. But before that, have you profiled if a vectorized copy is actually needed?

cpp/src/neighbors/detail/vpq_dataset.cuh
```diff
@@ -384,56 +560,222 @@ auto vpq_convert_math_type(const raft::resources& res, vpq_dataset<OldMathT, Idx
...
+    for (; l + 3 * SubWarpSize < pq_book_size; l += 4 * SubWarpSize) {
+      MathT d0 = 0.0f, d1 = 0.0f, d2 = 0.0f, d3 = 0.0f;
+
+#pragma unroll 4
+      for (uint32_t k = 0; k < pq_len; k++) {
+        MathT x  = __shfl_sync(subwarp_mask, my_dataset_val, k, SubWarpSize);
+        MathT c0 = __ldg(&centers_ptr[l * pq_len + k]);
+        MathT c1 = __ldg(&centers_ptr[(l + SubWarpSize) * pq_len + k]);
+        MathT c2 = __ldg(&centers_ptr[(l + 2 * SubWarpSize) * pq_len + k]);
+        MathT c3 = __ldg(&centers_ptr[(l + 3 * SubWarpSize) * pq_len + k]);
```
divyegala: Why the need for intrinsics here? Did you profile this kernel as a bottleneck? I generally like to avoid premature optimization and keep the kernels as simple as possible.

cpp/src/preprocessing/quantize/detail/pq.cuh
```diff
@@ -0,0 +1,336 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2025-2026, NVIDIA CORPORATION.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#pragma once
+
+#include "../../../core/nvtx.hpp"
```
divyegala, line 8: Relative include for header file?

cpp/src/preprocessing/quantize/detail/pq_codepacking.cuh
```diff
@@ -0,0 +1,141 @@
...
+#include <raft/util/integer_utils.hpp>
+#include <raft/util/pow2_utils.cuh>
+
+namespace cuvs::preprocessing::quantize::detail {
+
+/**
+ * Bitfield reference for reading from const pointers.
+ * This type mimics `uint8_t` or `uint16_t` for reading bitfield values.
+ */
+struct bitfield_const_ref_t {
```
divyegala: Is it possible to simplify the code a bit by using templates to account for const and non-const pointers?

c/src/preprocessing/quantize/pq.cpp
lowener: Because those headers are in the `src` folder of the `c` code, not in the `include` folder. They are not a part of the public API (as the public API is only `*.h` headers).
Most c++ files in the C API have the same relative includes

cpp/src/neighbors/detail/vamana/vamana_codebooks.cuh
lowener: This line is probably supposed to be in the follow-up branch that I prepared to get Vamana to use the PQ API.
`std::ifstream` is used just below and decoupling Vamana and PQ exposed that the proper include is not present.

cpp/src/preprocessing/quantize/detail/pq.cuh
lowener: This header file is not a part of the public API, it is located in `src/core/nvtx.hpp`

cpp/src/neighbors/detail/vpq_dataset.cuh
```diff
@@ -262,52 +349,112 @@ __device__ auto compute_code(
...
+    CodeT code = compute_code<SubWarpSize, CodeT>(dataset,
+                                                  vq_centers,
+                                                  std::make_optional(pq_centers_smem_view),
+                                                  pq_centers,
+                                                  row_ix,
+                                                  j,
+                                                  vq_label);
     // TODO: this writes in global memory one byte per warp, which is very slow.
     //  It's better to keep the codes in the shared memory or registers and dump them at once.
     if (lane_id == 0) { code_view[j] = code; }
```
tarang-jain, line 393: This has been a TODO for a while. Can we fix it here? I think you can potentially use a 16 byte local register here (only lane_id == 0 needs to maintain that, which it writes to, so 16 bytes of extra register space per warp). In the end, we can dump that 128 bit chunk to global memory. Maybe your `copy_vectorized` machinery comes in handy here to copy to global, or even direct `reinterpret_cast` to do it with things like uint4?
However, I do understand that this is separate from the scope of this PR and can be addressed in the future.

cpp/src/neighbors/detail/vpq_dataset.cuh
```diff
@@ -262,52 +349,112 @@ __device__ auto compute_code(
   return code;
 }
 
+template <int VecBytes = 16, typename T>
+__device__ inline void copy_vectorized(T* out, const T* in, uint32_t n)
+{
```
tarang-jain: we have a copy_vectorized in ivf_flat_interleaved_scan.cuh. Looks like this is doing the same thing. Can we reuse?

cpp/src/neighbors/detail/vpq_dataset.cuh
```diff
@@ -383,57 +559,269 @@ auto vpq_convert_math_type(const raft::resources& res, vpq_dataset<OldMathT, Idx
...
+struct vec4_op<float> {
+  using type = float4;
+  __device__ __forceinline__ static type make(float x, float y, float z, float w)
+  {
+    return make_float4(x, y, z, w);
+  }
+
+  __device__ __forceinline__ static type load(const float* ptr)
+  {
+    return __ldg(reinterpret_cast<const type*>(ptr));
```
tarang-jain: Can we use raft::ldg here if possible?

cpp/src/neighbors/detail/vpq_dataset.cuh
```diff
@@ -383,57 +559,269 @@ auto vpq_convert_math_type(const raft::resources& res, vpq_dataset<OldMathT, Idx
...
+      MathT d0 = 0.0f, d1 = 0.0f, d2 = 0.0f, d3 = 0.0f;
+      process_4centers_vec4<SubWarpSize, MathT>(d0, d1, d2, d3, centers_ptr, l, pq_len, get_x_smem);
+      compute_4centers_and_update<SubWarpSize, MathT, CodeT>(min_dist, code, d0, d1, d2, d3, l);
+    }
+
+    // Tail: process 1 center at a time
+    for (; l < pq_book_size; l += SubWarpSize) {
+      MathT d0 = 0.0f;
+      for (uint32_t k = 0; k < pq_len; k++) {
+        MathT diff = smem_dataset[k] - __ldg(&centers_ptr[l * pq_len + k]);
```
tarang-jain: Here as well. Can we use raft::ldg.

cpp/src/neighbors/detail/vpq_dataset.cuh
```diff
@@ -383,57 +559,269 @@ auto vpq_convert_math_type(const raft::resources& res, vpq_dataset<OldMathT, Idx
...
+
+  // Tail loop
+  for (; k < pq_len; k++) {
+    MathT x  = get_x_func(k);
+    MathT c0 = __ldg(&centers_ptr[l * pq_len + k]);
+    MathT c1 = __ldg(&centers_ptr[(l + SubWarpSize) * pq_len + k]);
+    MathT c2 = __ldg(&centers_ptr[(l + 2 * SubWarpSize) * pq_len + k]);
+    MathT c3 = __ldg(&centers_ptr[(l + 3 * SubWarpSize) * pq_len + k]);
+
+    d0 = __fmaf_rn(x - c0, x - c0, d0);
```
tarang-jain: Nit: x - c0 is computed twice. Can we store it?

cpp/src/neighbors/detail/vpq_dataset.cuh
```diff
@@ -383,57 +559,269 @@ auto vpq_convert_math_type(const raft::resources& res, vpq_dataset<OldMathT, Idx
...
+      if (vq_centers) { x -= vq_centers.value()(vq_label, jk); }
+      my_dataset_val = x;
+    }
+
+    const uint32_t subwarp_idx = raft::Pow2<SubWarpSize>::div(raft::laneId());
+    const uint32_t subwarp_mask =
+      (SubWarpSize == 32) ? 0xffffffffu : (0xffffu << (subwarp_idx * SubWarpSize));
+
+    auto get_x_shuffle = [&](uint32_t k) {
+      return __shfl_sync(subwarp_mask, my_dataset_val, k, SubWarpSize);
```
tarang-jain: Again we have a bunch of warp level primitives already wrapped in raft. Lets just use those rather than directly calling cuda instrinsics:
https://github.com/rapidsai/raft/blob/main/cpp/include/raft/util/warp_primitives.cuh

cpp/src/neighbors/detail/vpq_dataset.cuh
```diff
@@ -383,57 +559,269 @@ auto vpq_convert_math_type(const raft::resources& res, vpq_dataset<OldMathT, Idx
...
+__device__ auto compute_code_subspaces(
+  raft::device_matrix_view<const DataT, IdxT, raft::row_major> dataset,
+  std::optional<raft::device_matrix_view<const MathT, uint32_t, raft::row_major>> vq_centers,
+  raft::device_matrix_view<const MathT, uint32_t, raft::row_major> pq_centers,
+  MathT* smem_dataset,
+  IdxT row_ix,
+  uint32_t j,
+  LabelT vq_label) -> CodeT
+{
+  static_assert(std::is_same_v<CodeT, uint8_t> || std::is_same_v<CodeT, uint16_t>,
```
tarang-jain, line 682: We should use methods like `std::enable_if_t` to compile only for the correct types rather than failing with a `static_assert` here.

cpp/src/neighbors/detail/vpq_dataset.cuh
tarang-jain: Oh looks like @divyegala already pointed this out above.

cpp/src/neighbors/detail/vpq_dataset.cuh
```diff
@@ -262,52 +349,112 @@ __device__ auto compute_code(
...
 template <typename MathT, typename IdxT, typename DatasetT>
-auto process_and_fill_codes(
+void process_and_fill_codes(
   const raft::resources& res,
   const vpq_params& params,
   const DatasetT& dataset,
-  raft::device_matrix_view<const MathT, uint32_t, raft::row_major> vq_centers,
-  raft::device_matrix_view<const MathT, uint32_t, raft::row_major> pq_centers)
-  -> raft::device_matrix<uint8_t, IdxT, raft::row_major>
+  std::optional<raft::device_matrix_view<const MathT, uint32_t, raft::row_major>> vq_centers,
```
tarang-jain: Lets keep optional arguments at the very end of the signature and default to std::nullopt

cpp/src/neighbors/detail/vpq_dataset.cuh
```diff
@@ -383,57 +559,269 @@ auto vpq_convert_math_type(const raft::resources& res, vpq_dataset<OldMathT, Idx
   return vpq_dataset<NewMathT, IdxT>{
     std::move(vq_code_book), std::move(pq_code_book), std::move(src.data)};
 }
+template <typename MathT>
+struct vec4_op {};
+
+template <>
+struct vec4_op<float> {
+  using type = float4;
+  __device__ __forceinline__ static type make(float x, float y, float z, float w)
```
tarang-jain: Nit: lets just make this the constructor rather than `make`

cpp/src/neighbors/detail/vpq_dataset.cuh
```diff
@@ -383,57 +559,269 @@ auto vpq_convert_math_type(const raft::resources& res, vpq_dataset<OldMathT, Idx
...
+  const DataT* __restrict__ dataset_ptr = dataset.data_handle() + row_ix * dataset.extent(1);
+  MathT min_dist                        = std::numeric_limits<MathT>::infinity();
+  CodeT code                            = 0;
+
+  if (pq_len <= SubWarpSize) {
+    // Each thread loads one dataset element and will share to others using shuffle
+    MathT my_dataset_val = 0.0f;
+    if (lane_id < pq_len) {
+      auto jk = j * pq_len + lane_id;
+      auto x  = data_mapping(dataset_ptr[jk]);
```
tarang-jain, line 699: Nit: These reads seem coaelesced, so I dont see a huge disadvantage or advantage to using __ldg. But lets just be consistent. Either we use it for all these coalesced reads from dataset or we dont.

cpp/src/neighbors/detail/vpq_dataset.cuh
divyegala: I disagree, `static_assert` provides a cleaner error message. That said we are on C++20 now and can use the `requires` clause or create a new `Concept` to encapsulated allowed byte types. https://en.cppreference.com/w/cpp/language/constraints.html

cpp/src/neighbors/detail/vpq_dataset.cuh
divyegala: +1 here. I still believe simplicity and readability is valuable and we should not have to rely on intrinsics for minor things like this. If this kernel is provably a bottleneck then it makes sense to optimize.

tarang-jain: yeah in general, its a bit too complex to follow, especially the part where four floats are handled in one go. But maybe that is a necessary optimization.

cpp/src/neighbors/detail/vpq_dataset.cuh
lowener: I created https://github.com/rapidsai/raft/pull/2900 to add it to RAFT.
After some profiling I noticed that the vectorized copy is not actually needed so I changed it to a regular copy to the shared mem.

cpp/src/neighbors/detail/vpq_dataset.cuh
lowener: The intrinsics is not needed there so I removed it. And yes this kernel was a bottleneck, but computing 4 centers in a row does actually much more for performance.

cpp/src/neighbors/detail/vpq_dataset.cuh
lowener: I removed all direct `__ldg` as they are not needed

cpp/src/neighbors/detail/vpq_dataset.cuh
lowener: I just tried to use a concept for it, but I found the error message to be more vague than the `static_assert`.
With a concept the error becomes `"cannot determine which instance of function template is intended"` when called with `uint32_t` as CodeT

cpp/src/neighbors/detail/vpq_dataset.cuh
lowener: I removed the intrinsics calls.
The four centers being handled in one go is the main driver for performance  improvement, and can lead to up to 2.5x faster transform

cpp/src/neighbors/detail/vpq_dataset.cuh
lowener: This struct is only providing static functions to manipulate float4 objects, so there is no need for a constructor

cpp/src/neighbors/detail/vpq_dataset.cuh
divyegala: Yes, static_assert is definitely the cleanest because we can control the error message.

cpp/src/neighbors/detail/vpq_dataset.cuh
lowener: Artem created a PR for it: https://github.com/rapidsai/cuvs/pull/1487
The bottleneck is not on the write operation, it is on the ALU and memory reading. I don't think I will change it on this PR as it still has no noticeable impact

cpp/src/neighbors/detail/vpq_dataset.cuh
```diff
@@ -263,51 +351,73 @@ __device__ auto compute_code(
...
+  const IdxT row_ix   = subwarp_align::div(IdxT{threadIdx.x} + IdxT{BlockSize} * IdxT{blockIdx.x});
   if (row_ix >= out_codes.extent(0)) { return; }
 
-  const uint32_t pq_dim = raft::div_rounding_up_unsafe(vq_centers.extent(1), pq_centers.extent(1));
+  const uint32_t pq_dim = raft::div_rounding_up_unsafe(dataset.extent(1), pq_centers.extent(1));
 
-  const uint32_t lane_id = raft::Pow2<kSubWarpSize>::mod(threadIdx.x);
-  const LabelT vq_label  = vq_labels(row_ix);
+  // Copy the pq_centers into shared memory for faster processing
+  const IdxT rows_in_shared_memory = shared_memory_size / (sizeof(MathT) * pq_centers.extent(1));
```
mfoerste4: Do we need to check against `pq_centers.extent(0)` here or in subsequent loop to protect against illegal access?

cpp/src/preprocessing/quantize/detail/pq.cuh
```diff
@@ -0,0 +1,336 @@
...
+  const ix_t n_rows       = out_vectors.extent(0);
+  const ix_t dim          = out_vectors.extent(1);
+  const ix_t pq_dim       = params.pq_dim;
+  const ix_t pq_bits      = params.pq_bits;
+  const ix_t pq_n_centers = ix_t{1} << pq_bits;
+
+  auto stream = raft::resource::get_cuda_stream(res);
+
+  constexpr ix_t kBlockSize  = 256;
+  const ix_t threads_per_vec = raft::WarpSize;
```
mfoerste4: Does this also hold for` pq_bits ==4 `where we have a subwarp scope?

cpp/src/neighbors/detail/vpq_dataset.cuh
tarang-jain: Sounds good. Marking as resolved.

cpp/cmake/thirdparty/get_faiss.cmake
```diff
@@ -31,7 +32,7 @@ function(find_and_configure_faiss)
     set(CPM_DOWNLOAD_faiss ON)
   endif()
 
-  include(cmake/modules/FindAVX)
+  include(${CMAKE_CURRENT_FUNCTION_LIST_DIR}/../modules/FindAVX.cmake)
```
tarang-jain, line 35: why are these changes needed?

cpp/src/neighbors/detail/vpq_dataset.cuh
lowener: When `pq_centers_smem`, it's boundaries are checked, and `pq_centers_smem.extent(0)` can't be superior to `pq_centers.extent(0)`. I edited the code to make it clearer:
```c++
rows_in_shared_memory = std::min<ix_t>(pq_centers.extent(0), kMaxSharedMemorySize / (sizeof(MathT) * pq_centers.extent(1)));
auto pq_centers_smem_view = raft::make_device_matrix_view<const MathT, uint32_t, raft::row_major>(
    pq_centers_smem, rows_in_shared_memory, pq_centers.extent(1));
```

cpp/src/preprocessing/quantize/detail/pq.cuh
lowener: Thanks for pointing that out, `threads_per_vec` is now consistent with the other functions:
```c++
threads_per_vec = std::min<ix_t>(raft::WarpSize, pq_n_centers);
```

cpp/cmake/thirdparty/get_faiss.cmake
lowener: It is not absolutely necessary.
While I was doing some extra benchmark on the side with Faiss I noticed that when using `get_faiss.cmake`, `FindAVX` can sometimes not be found depending on the current directory.
This fixes it and finds the `modules` directory just like we find the `patches` directory a few lines before (l. 18) in the same file, using `CMAKE_CURRENT_FUNCTION_LIST_DIR`

cpp/include/cuvs/neighbors/common.hpp
```diff
@@ -65,8 +70,25 @@ struct vpq_params {
    * When zero, an optimal value is selected using a heuristic.
    */
   double pq_kmeans_trainset_fraction = 0;
```
tfeher, line 74: I suggest to mark this as deprecated, and only use `max_train_points_per_pq_code`.

cpp/include/cuvs/neighbors/common.hpp
```diff
@@ -65,8 +70,25 @@ struct vpq_params {
...
+   */
+  cuvs::cluster::kmeans::kmeans_type pq_kmeans_type =
+    cuvs::cluster::kmeans::kmeans_type::KMeansBalanced;
+  /**
+   * The max number of data points to use per PQ code during PQ codebook training. Using more data
+   * points per PQ code may increase the quality of PQ codebook but may also increase the build
+   * time. We will use `pq_n_centers * max_train_points_per_pq_code` training
+   * points to train each PQ codebook.
+   */
+  uint32_t max_train_points_per_pq_code = 256;
```
tfeher, line 89: We should consider having the similar parameter  `max_train_points_per_vq_cluster`, and deprecate `vq_kmeans_traniset_fraction`.

c/include/cuvs/preprocessing/quantize/pq.h
```diff
@@ -0,0 +1,220 @@
...
+  /**
+   * The fraction of data to use during iterative kmeans building (VQ phase).
+   * When zero, an optimal value is selected using a heuristic.
+   */
+  double vq_kmeans_trainset_fraction;
+  /**
+   * The fraction of data to use during iterative kmeans building (PQ phase).
+   * When zero, an optimal value is selected using a heuristic.
+   */
+  double pq_kmeans_trainset_fraction;
```
tfeher: I would recommend that the C API only exposes the max_point_per_code/cluster parameters and not the fraction parameters.

python/cuvs/cuvs/preprocessing/quantize/pq/pq.pyx
```diff
@@ -0,0 +1,368 @@
...
+        When one, only product quantization is used.
+    kmeans_n_iters: int
+        specifies the number of iterations searching for kmeans centers
+    vq_kmeans_trainset_fraction: float
+        specifies the fraction of data to use during iterative kmeans building
+        for the vector quantizer. When zero, an optimal value is selected
+        using a heuristic.
+    pq_kmeans_trainset_fraction: float
+        specifies the fraction of data to use during iterative kmeans building
+        for the product quantizer. When zero, an optimal value is selected
```
tfeher: Same as above, I suggest to remove the `_fraction` parameters and have `max_train_points`

cpp/include/cuvs/neighbors/common.hpp
lowener: Marked it as deprecated, and it is no longer a part of the new C++ API for PQ.
At the moment the number of rows is selected as `min(trainset_fraction * n_rows, max_train_point_per_pq_code * pq_n_centers)` while users make their transition.

robertmaynard (changes requested):
~@cjnolet This isn't eligable for 26.02 if we want to keep ABI conformance with 25.12~

Read diffs wrong, unblocked

robertmaynard (approved):

c/include/cuvs/cluster/kmeans.h
benfred: This introduces two new symbols to the global namespace: `Kmeans` and `KmeansBalanced` - and could cause symbol collisions for people including this header.

Perhaps we should rename to something like`cuvsKmeansTypeKmeans` and `cuvsKmeansTypeKmeansBalanced`  ?  

(I realize that we haven't been doing this for most of our enums so far, but I think this is something that we should change)

benfred: Also - we are using a boolean parameter for this distinction in the `kMeansParams` object 
https://github.com/rapidsai/cuvs/blob/13fb5863f3e93b4d3b2fdce4dd59a9a11e725c2d/c/include/cuvs/cluster/kmeans.h#L95-L98

and i feel like we either should be using an enum for both or using a bool for both to be consistent

cpp/src/preprocessing/quantize/detail/pq.cuh
```diff
@@ -0,0 +1,360 @@
...
+  std::optional<raft::device_matrix_view<const MathT, uint32_t, raft::row_major>>
+    vq_code_book_view_opt = std::nullopt;
+  std::optional<raft::device_matrix<MathT, uint32_t, raft::row_major>> vq_code_book_opt =
+    std::nullopt;
+  if (filled_params.use_vq) {
+    vq_code_book_opt      = cuvs::neighbors::detail::train_vq<MathT>(res, vpq_params, dataset);
+    vq_code_book_view_opt = raft::make_const_mdspan(vq_code_book_opt.value().view());
+  } else {
+    vq_code_book_opt = raft::make_device_matrix<MathT, uint32_t, raft::row_major>(res, 0, 0);
+  }
```
tfeher: Nitpick, but all this optional construct make the code less readable. Why don't we just do the following?

(Also we should prefer `int64_t `as mdspan index type).

```suggestion
  auto vq_code_book = raft::make_device_matrix<MathT, int64_t, raft::row_major>>(res, 0, 0);
  if (filled_params.use_vq) {
    vq_code_book      = cuvs::neighbors::detail::train_vq<MathT>(res, vpq_params, dataset);
  }
```

Later we would just get an mdspan when we need. 
```c++
train_pq_subspaces<MathT>(res, filled_params, dataset, raft::make_const_mdspan(vq_code_book));
```
This requires to remove the `optional` wrapper from the function signature, which I would also recommend.

cpp/src/preprocessing/quantize/detail/pq.cuh
```diff
@@ -0,0 +1,360 @@
...
+    .max_train_points_per_pq_code    = params.max_train_points_per_pq_code,
+    .max_train_points_per_vq_cluster = params.max_train_points_per_vq_cluster};
+}
+
+template <typename MathT, typename DatasetT>
+auto train_pq_subspaces(
+  const raft::resources& res,
+  const cuvs::preprocessing::quantize::pq::params& params,
+  const DatasetT& dataset,
+  std::optional<const raft::device_matrix_view<const MathT, uint32_t, raft::row_major>> vq_centers)
```
tfeher: 1. Use int64_t index type
2. Nitpick but it would be more natural to leave out the `std::optional` wrapper, and use `vq_centers.empty()` to check whether we have `vq_centers`. This would simplify the code at the call site.

cpp/src/preprocessing/quantize/detail/pq.cuh
```diff
@@ -0,0 +1,360 @@
...
+  const ix_t pq_n_centers = ix_t{1} << pq_bits;
+  const ix_t pq_len       = raft::div_rounding_up_safe(dim, pq_dim);
+  RAFT_EXPECTS((dim % pq_dim) == 0, "Dimension must be divisible by pq_dim");
+  const ix_t n_rows_train =
+    std::min<ix_t>(n_rows, params.max_train_points_per_pq_code * pq_n_centers);
+  RAFT_EXPECTS(
+    n_rows_train >= pq_n_centers,
+    "The number of training samples must be equal to or greater than the number of PQ centers");
+
+  std::optional<raft::device_matrix<MathT, ix_t, raft::row_major>> pq_trainset = std::nullopt;
```
tfeher: Why don't we just initialize with zero size array?
```suggestion
  auto pq_trainset = raft::make_device_matrix<MathT, ix_t, raft::row_major>>(res, 0, 0);
```

cpp/src/preprocessing/quantize/detail/pq.cuh
```diff
@@ -0,0 +1,360 @@
...
+
+  // Subtract VQ centers
+  if (vq_centers) {
+    pq_trainset = std::make_optional(cuvs::util::subsample(res, dataset, n_rows_train));
+    auto vq_labels =
+      cuvs::neighbors::detail::predict_vq<uint32_t>(res, pq_trainset.value(), vq_centers.value());
```
tfeher: We should pass on pq_trainset.view() and not the owning mdarray.

cpp/src/preprocessing/quantize/detail/pq.cuh
```diff
@@ -0,0 +1,360 @@
...
+    auto stream = raft::resource::get_cuda_stream(res);
+    sub_labels_opt =
+      std::make_optional(rmm::device_uvector<uint32_t>(n_rows_train, stream, device_memory));
+    pq_cluster_sizes_opt =
+      std::make_optional(rmm::device_uvector<uint32_t>(pq_n_centers, stream, device_memory));
+    sub_labels_view_opt = std::make_optional(
+      raft::make_device_vector_view<uint32_t, ix_t>(sub_labels_opt.value().data(), n_rows_train));
+    pq_cluster_sizes_view_opt = std::make_optional(raft::make_device_vector_view<uint32_t, ix_t>(
+      pq_cluster_sizes_opt.value().data(), pq_n_centers));
+  }
```
tfeher: This is unnecessarily complex. We do not need `optional` wrappers. We should prefer `raft::device_vector` instead of `rmm_device_uvector`. We can pass the memory resource to `make_devie_mdarray`.

```suggestion
  auto trainset_ptr = !pq_trainset.empty() ? pq_trainset.data_handle() : dataset.data_handle();
  auto sub_labels = raft::make_device_vector<int32_t>(res, 0);
  auto pq_cluster_sizes = raft::make_device_vector<uint32_t>(res, 0);
  auto device_memory = raft::resource::get_workspace_resource(res);
  if (params.pq_kmeans_type == cuvs::cluster::kmeans::kmeans_type::KMeansBalanced) {
    auto stream = raft::resource::get_cuda_stream(res);
    sub_labels = raft::make_device_mdarray<uint32_t>(
        res, device_memory, raft::make_extents<int64_t>(n_rows_train));
    pq_cluster_sizes = raft::make_device_mdarray<uint32_t>(
        res, device_memory, raft::make_extents<int64_t>(pq_n_centers));
  }
```

cpp/src/preprocessing/quantize/detail/pq.cuh
```diff
@@ -0,0 +1,360 @@
...
+          typename CodeT,
+          typename DataT,
+          typename MathT,
+          typename IdxT,
+          typename LabelT>
+__launch_bounds__(BlockSize) RAFT_KERNEL reconstruct_vectors_kernel(
+  raft::device_matrix_view<const uint8_t, IdxT, raft::row_major> codes,
+  raft::device_matrix_view<DataT, IdxT, raft::row_major> dataset,
+  raft::device_matrix_view<const MathT, uint32_t, raft::row_major> pq_centers,
+  std::optional<raft::device_matrix_view<const DataT, IdxT, raft::row_major>> vq_centers,
```
tfeher: Please prefer to use `int64_t` as mdpan index type
```suggestion
  raft::device_matrix_view<const uint8_t, int64_t, raft::row_major> codes,
  raft::device_matrix_view<DataT, int64_t, raft::row_major> dataset,
  raft::device_matrix_view<const MathT, int64_t, raft::row_major> pq_centers,
  std::optional<raft::device_matrix_view<const DataT, int64_t, raft::row_major>> vq_centers,
```

cpp/src/preprocessing/quantize/detail/pq.cuh
```diff
@@ -0,0 +1,360 @@
...
+}
+
+template <typename DataT, typename MathT, typename IdxT, typename LabelT>
+auto reconstruct_vectors(
+  const raft::resources& res,
+  const params& params,
+  raft::device_matrix_view<const uint8_t, IdxT, raft::row_major> codes,
+  raft::device_matrix_view<const MathT, uint32_t, raft::row_major> pq_centers,
+  std::optional<raft::device_matrix_view<const DataT, uint32_t, raft::row_major>> vq_centers,
+  raft::device_matrix_view<DataT, IdxT, raft::row_major> out_vectors,
```
tfeher, line 255: prefer int64_t index tyes
```suggestion
  raft::device_matrix_view<const uint8_t, int64_t, raft::row_major> codes,
  raft::device_matrix_view<const MathT, int64_t, raft::row_major> pq_centers,
  std::optional<raft::device_matrix_view<const DataT, int64_t, raft::row_major>> vq_centers,
  raft::device_matrix_view<DataT, int64_t, raft::row_major> out_vectors,
```

cpp/src/neighbors/detail/vpq_dataset.cuh
```diff
@@ -383,57 +527,311 @@ auto vpq_convert_math_type(const raft::resources& res, vpq_dataset<OldMathT, Idx
...
+void process_and_fill_codes_subspaces(
   const raft::resources& res,
   const vpq_params& params,
   const DatasetT& dataset,
-  raft::device_matrix_view<const MathT, uint32_t, raft::row_major> vq_centers,
-  raft::device_matrix_view<const MathT, uint32_t, raft::row_major> pq_centers)
-  -> raft::device_matrix<uint8_t, IdxT, raft::row_major>
+  raft::device_matrix_view<const MathT, uint32_t, raft::row_major> pq_centers,
+  raft::device_matrix_view<uint8_t, IdxT, raft::row_major> codes,
+  std::optional<raft::device_matrix_view<const MathT, uint32_t, raft::row_major>> vq_centers =
```
tfeher:
```suggestion
  raft::device_matrix_view<const MathT, int64_t, raft::row_major> pq_centers,
  raft::device_matrix_view<uint8_t, int64_t, raft::row_major> codes,
  std::optional<raft::device_matrix_view<const MathT, int64_t, raft::row_major>> vq_centers =
```

cpp/src/neighbors/detail/vpq_dataset.cuh
```diff
@@ -383,57 +527,311 @@ auto vpq_convert_math_type(const raft::resources& res, vpq_dataset<OldMathT, Idx
   return vpq_dataset<NewMathT, IdxT>{
     std::move(vq_code_book), std::move(pq_code_book), std::move(src.data)};
 }
+template <typename MathT, uint8_t PqLenVectorSize>
+struct vec_op {
+  using type = void;
+};
+
+template <>
+struct vec_op<float, 4> {
```
tfeher: While this works, I have a feeling that it could have been simpler with [raft::txN_t](https://github.com/rapidsai/raft/blob/ac2b4ed12bb51aa0757a6f965e00b0033dc182f8/cpp/include/raft/util/vectorized.cuh#L281) and separate functions for the math operations.

cpp/src/preprocessing/quantize/detail/pq_codepacking.cuh
```diff
@@ -0,0 +1,112 @@
...
+
+/**
+ * Bitfield reference for reading/writing from non-const pointers.
+ * This type mimics `uint8_t&` or `uint16_t&` for the indexing operator.
+ */
+template <typename PtrT = uint8_t*>
+struct bitfield_ref_t {
+  PtrT ptr;
+  uint32_t offset;
+  uint32_t bits;
```
tfeher, line 30: The `bitfield_ref_t` is almost the same as `bitfield_ref_t` in ivf_pq_codepacking.cuh, except that here `bits` is a member variable instead of a template arg. Could mention this in the docstring? Additionally please document the meaning of `bits` - number of bits comprising the value.

cpp/src/preprocessing/quantize/detail/pq_codepacking.cuh
```diff
@@ -0,0 +1,112 @@
...
+  }
+};
+
+/**
+ * View a byte array as an array of unsigned integers of custom small bit size.
+ */
+template <typename PtrT = uint8_t*>
+struct bitfield_view_t {
+  PtrT raw;
+  uint32_t bits;
```
tfeher, line 99: Please add a docstring.

tfeher (changes requested):
I have finished my review. Overall looks good, a few minor comments below.

c/include/cuvs/cluster/kmeans.h
robertmaynard: I agree we should change these now before they get baked into the ABI and we have to wait ~6 months

cpp/src/preprocessing/quantize/detail/pq.cuh
lowener: Regarding the index type I think the current version makes sense: The public-facing API accepts only `int64_t` for the dataset to train/transform, and the vq/pq_centers are `uint32_t`.
For the centers, overflows are not a big worry since vq_centers is of max size `n_vq_centers` and pq_centers is of max_size `pq_dim * (1<<pq_bits)`. And batching is used to iterate over the dataset.

c/include/cuvs/cluster/kmeans.h
lowener: Done, I created https://github.com/rapidsai/cuvs/issues/1717 to track that problem

lowener: For the bool vs. enum I much prefer the enum since it would support much better adding a third algorithm for KMeans. But I don't want to modify kmeans codebase in this PR since its scope is already big enough

cpp/src/neighbors/detail/vpq_dataset.cuh
lowener: I updated that code to inherit from `raft::TxN_t` to make it more simple.

cpp/src/preprocessing/quantize/detail/pq.cuh
tfeher: Indeed overflow is not an issue. My problem is keeping track of where we need `uint32_t` and where `int64_t` as `mdspan` index type. Unless we have a good reason (e.g. significant perf or memory overhead) we should prefer the int64_t.

cjnolet:
@lowener looks like we still have some CI failures. Would love to get this merged before the weekend.

irina-resh-nvda:
Thank you @lowener  for this PR! parse_build_param function on line 198  in /cpp/bench/ann/src/cuvs/cuvs_ann_bench_param_parser.h also needs to pick up the new parameters

mythrocks:
I don't see changes to the Java code here, so I'm reviewing these from the ABI perspective.

Will we be addressing [this comment](https://github.com/rapidsai/cuvs/pull/1278#discussion_r2699764485) about enum names, in this PR/release?  I ask because I didn't see a PR attached to #1717.

lowener:
@mythrocks A force push triggered a request for re-review on all files.
The comment about ABI changes was addressed, and now the enums introduced in this PR have a name more specific that is not likely to cause symbol collision.
The issue #1717 is to fix it accross all the codebase but it is not within the scope of this PR

Issue #1279, opened 2025-08-23
[FEA] all neighbors python API


cjnolet:
The new all neighbors API is proving to be immensely useful. We should wrap it with C and Python to make it easier to use and debug downstream applications.

viclafargue:
Closed by #1282.

PR #1282, opened 2025-08-26
All neighbors C and Python bindings

viclafargue:
Answers https://github.com/rapidsai/cuvs/issues/1279

### DIFF
--- a/docs/source/neighbors/all_neighbors.rst
+++ b/docs/source/neighbors/all_neighbors.rst
@@ -65,6 +65,7 @@ Performance Characteristics
 
 - **Build Time**: Scales with dataset size and chosen algorithm
 - **Memory Usage**: Depends on cluster size and overlap factor
+- **Accuracy**: Varies by algorithm (brute_force, others) and also according to the number of clusters (n_clusters). n_clusters>1 will result in an approximation.
 - **Scalability**: Linear scaling with number of GPUs when n_clusters > 1
 
 The algorithm automatically chooses between single-GPU and multi-GPU execution based on the n_clusters parameter and available resources.


python/cuvs/cuvs/tests/test_all_neighbors.py
```diff
@@ -0,0 +1,166 @@
...
+    assert recall > 0.9
+
+
+@pytest.mark.parametrize("algo", ["nn_descent", "brute_force", "ivf_pq"])
+def test_all_neighbors_host_build_quality(algo):
+    """Test host build with quality validation against brute force ground
+    truth.
+    """
+    n_rows, n_cols, k = 7151, 64, 16
+    n_clusters = 1
```
jinsolp: Could you also add tests with larger `n_clusters`? Maybe just add `n_clusters=4` to check if it propagates properly (many correctness checks already being done at cpp level)

cpp/include/cuvs/neighbors/all_neighbors.h
```diff
@@ -0,0 +1,121 @@
...
+  cuvsIvfPqIndexParams_t ivf_pq_params;          ///< Parameters for IVF-PQ algorithm (when algo ==
+                                                 ///< CUVS_ALL_NEIGHBORS_ALGO_IVF_PQ)
+  cuvsNNDescentIndexParams_t nn_descent_params;  ///< Parameters for NN-Descent algorithm (when algo
+                                                 ///< == CUVS_ALL_NEIGHBORS_ALGO_NN_DESCENT)
+};
+
+typedef struct cuvsAllNeighborsIndexParams* cuvsAllNeighborsIndexParams_t;
+
+/**
+ * @brief Build an all-neighbors k-NN graph from a host-resident dataset (SNMG).
```
jinsolp: the host-resident dataset can also take a regular `raft::device_resources` and run single-gpu. Could you expose that as well 🙏

python/cuvs/cuvs/common/resources.pyx
```diff
@@ -80,6 +86,83 @@ cdef class Resources:
     def __dealloc__(self):
         check_cuvs(cuvsResourcesDestroy(self.c_obj))
 
+cdef class SNMGResources:
```
jinsolp: I didn't know we had a separate resource wrapper in cuVS. If there is no explicit reason to not import raft, I think we can just use `pylibraft.common.DeviceResourcesSNMG` (used like [this](https://github.com/rapidsai/cuml/blob/952e4a46c1dca1ea3484903923d4ca9e53887a81/python/cuml/tests/test_umap.py#L29) in cuML). 
I think we should be able to have `cuvsResources_t` to be the arg type in the c api which will dynamically dispatch to the multi gpu if the user passes `DeviceResourcesSNMG` as the handle.

python/cuvs/cuvs/neighbors/all_neighbors/all_neighbors.pyx
```diff
@@ -0,0 +1,393 @@
...
+# setting resources from within the function as data on host/device differ
+@auto_convert_output
+def build(dataset, k, params, *,
+          indices=None,
+          distances=None,
+          core_distances=None,
+          alpha=1.0,
+          resources=None):
+    """
+    SNMG all-neighbors: always uses multi-GPU resources. If resources is None,
```
jinsolp: I think the default should be the `raft::resources` (i.e. single-gpu), and we can add docs saying that if the user chooses to pass mutli-gpu resource as the handle, it runs multi-gpu.

python/cuvs/cuvs/tests/test_all_neighbors.py
viclafargue: Tested with `n_clusters=8` as it seems like it offered better accuracy in this instance.

cpp/include/cuvs/neighbors/all_neighbors.h
viclafargue: Fixed this.

python/cuvs/cuvs/common/resources.pyx
viclafargue: I removed the `cuvsSNMGResources_t` type to simplify the API. But, we need the user to be able to instantiate a multi-GPU resource from C. This also enables other bindings. When it comes to handling RAFT handles, this is definitely something we would want to do, but I feel like this may be out of the scope of this PR. Implementing it only in All Neighbors would be inconsistent with the rest of the cuVS Python API. Maybe others would have different insights though. cc @cjnolet

python/cuvs/cuvs/neighbors/all_neighbors/all_neighbors.pyx
```diff
@@ -0,0 +1,382 @@
...
+    Parameters for all-neighbors k-NN graph building.
+
+    Parameters
+    ----------
+    algo : str or cuvsAllNeighborsAlgo
+        Algorithm to use for local k-NN graph building.
+        Options: "brute_force", "ivf_pq", "nn_descent"
+    overlap_factor : int, default=2
+        Number of clusters each point is assigned to (must be < n_clusters)
+    n_clusters : int, default=1
```
Intron7, line 88: Please mention that n_clusters> 1 is needed for multi-GPU.

cpp/include/cuvs/core/c_api.h
```diff
@@ -110,6 +110,24 @@ cuvsError_t cuvsStreamSync(cuvsResources_t res);
  * @return cuvsError_t
  */
 cuvsError_t cuvsDeviceIdGet(cuvsResources_t res, int* device_id);
+
+/**
+ * @brief Create a SNMG resources handle with all available GPUs
+ * @param[out] res output handle
+ * @return cuvsError_t
+ */
+cuvsError_t cuvsSNMGResourcesCreate(cuvsResources_t* res);
```
cjnolet: Can we please follow the standard we created for the multi-gpu APIs and use "MutliGpu" here? I understand the C++ piece uses SNMG, and honestly I think we could probably update that as well. The closer I get with day to day users and partners, the more I'm realizing our acronyms are confusing them.

cpp/include/cuvs/neighbors/all_neighbors.h
```diff
@@ -0,0 +1,121 @@
...
+ * @param[in] res             Single-GPU resources
+ * @param[in] params          Build parameters (see cuvsAllNeighborsIndexParams). For device
+ * datasets, `n_clusters` must be 1 (no batching); `overlap_factor` is ignored.
+ * @param[in] dataset         2D tensor [num_rows x dim] on device (CUDA/CUDAManaged)
+ * @param[out] indices        2D tensor [num_rows x k] on device (int64)
+ * @param[out] distances      Optional 2D tensor [num_rows x k] on device (float32); can be NULL
+ * @param[out] core_distances Optional 1D tensor [num_rows] on device (float32); can be NULL
+ * @param[in] alpha           Mutual-reachability scaling; used only when core_distances is provided
+ */
+cuvsError_t cuvsAllNeighborsBuildDevice(cuvsResources_t res,
```
cjnolet: Can we please merge these into a single function? Since we're using DLManagedTensor, we can dispatch inside of the appropriate functions without having to introduce the complexity of calling a different API.

cpp/src/core/c_api.cpp
```diff
@@ -291,3 +302,22 @@ extern "C" cuvsError_t cuvsMatrixSliceRows(cuvsResources_t res,
...
+
+extern "C" cuvsError_t cuvsSNMGResourcesCreate(cuvsResources_t* res)
+{
+  return cuvs::core::translate_exceptions([=] {
+    auto res_ptr = new raft::device_resources_snmg{};
+    *res         = reinterpret_cast<uintptr_t>(res_ptr);
+  });
+}
+
+extern "C" cuvsError_t cuvsSNMGResourcesCreateWithDevices(cuvsResources_t* res,
```
cjnolet: Nitpick: Can we call this `cuvsMultiGpuResourcesCreateWithDeviceIds`? Also, please accept a host-level dlmanaged tensor here. one of the primary reasons we accept dlmanaged tensor now instead of the raw pointers to arrays (even on host) is that we can avoid having to pass in an array + "number of elements in array". The dlmanagedtensor already carries this information forward.

docs/source/indexes/all_neighbors.rst
```diff
@@ -0,0 +1,71 @@
+All-neighbors
```
cjnolet: I love the docs, but all-neighbros isn't really an index type, it's more-so a way to take a series of dense vectors and construct a knn graph from them (it's really a way to convert dense vectors into a corresponding knn graph when you think about it). 
That said, I think this documentation is immensely useful, I'm just proposing we put it in a different spot in the dots (that better reflects its purpose) or rename the "indexes" directory/rst file to be something else- like `neighbors`.

python/cuvs/cuvs/common/resources.pyx
cjnolet: Nitpick- can we please rename `deviceSNMG*` to `DeviceMultiGpu` all the way down to the C++ layer? I'm realizing this naming is becoming confusing to users. 

Any C/C++ public APIs should absolutely not leak RAFT types. This is actually for more reasons than may be immediately obvious- it turns out the way symbol lookups are resolved in binaries are non-deterministic, and so they can causes nasty segfaults when, say, a symbol is looked up in cugraph or cuml, instantiated, and then later looked up in cuVS (where the instantiated symbol resolves to an uninstantiated one, thus causing a memory illegal access). 

I think it's okay to use RAFT types in the Python layer only, so long as the symbol is being exposed through pylibraft's cython/python layer and not being wrapped through cuVS' cython.

python/cuvs/cuvs/common/resources.pyx
```diff
@@ -118,3 +206,31 @@ def auto_sync_resources(f):
         resources_docstring=_resources_param_string
     )
     return wrapper
+
+
+def auto_sync_snmg_resources(f):
```
cjnolet: The more I think about this, the more I realize we should just be using pylibraft for this instead of wrapping a new type just in cuvs. That would allow cuml and others to benefit from this as well.

python/cuvs/cuvs/neighbors/all_neighbors/all_neighbors.pyx
```diff
@@ -0,0 +1,384 @@
...
+    def metric(self):
+        """Distance metric used for graph construction."""
+        # Reverse lookup in DISTANCE_TYPES
+        for name, value in DISTANCE_TYPES.items():
+            if value == self.params.metric:
+                return name
+        return self.params.metric
+
+
+# @auto_sync_snmg_resources
```
cjnolet: Make sure we uncomment this. 

I don't think we want to assume snmg here all the time, right? When we're not using snmg, we want to just have all-neighbors function as a single gpu.

python/cuvs/cuvs/common/resources.pyx
viclafargue: I agree that it would be neat to allow the use of RAFT handles. But, I think that this would have to be a dedicated follow-up work across the full API. Otherwise, it would need a lot of rework and would make the cuVS API inconsistent.

docs/source/indexes/all_neighbors.rst
viclafargue: What would you like the exact change to be regarding documentation?

python/cuvs/cuvs/common/resources.pyx
cjnolet: @viclafargue if you haven't done so already, can you please create an issue for this so it doesn't get lost in the noise of all other things going on?

viclafargue: https://github.com/rapidsai/cuvs/issues/1318

docs/source/indexes/all_neighbors.rst
jinsolp: I put the cpp doc [here](https://github.com/rapidsai/cuvs/blob/branch-25.10/docs/source/cpp_api/neighbors_all_neighbors.rst) so maybe this should go under `docs/source/c_api`

cjnolet: First change I would make is to not put it in the "indexes" docs (`[docs/source/indexes/all_neighbors.rst](https://github.com/rapidsai/cuvs/pull/1282/files#diff-1e4386723cfd6adfaa094ed3a0969c1611fd46a956966ae04b8f4d2dc7e49cfe)).

python/cuvs/cuvs/common/mg_resources.pyx
```diff
@@ -0,0 +1,153 @@
...
+)
+from cuvs.common.cydlpack cimport DLManagedTensor, dlpack_c
+
+import numpy as np
+from pylibraft.common.cai_wrapper import wrap_array
+
+from cuvs.common.exceptions import check_cuvs
+
+
+cdef class MultiGpuResources:
```
cjnolet, line 36: I haven't checked yet, but do we have a Github issue for this one? Can you please reference it in the TODO here so that we don't lose sight of it? Having a `DeviceResources` from raft and `MultiGpuResources` is not consistent and it's ultimately going to affect the user experience. We should make that fix asap (though we don't need to hold up this PR).

cpp/src/neighbors/all_neighbors_c.cpp
```diff
@@ -0,0 +1,254 @@
...
+}
+
+static void ensure_optional_core_distance_dtype(DLManagedTensor* core_distances)
+{
+  if (core_distances == nullptr) { return; }
+  auto dtype = core_distances->dl_tensor.dtype;
+  RAFT_EXPECTS(dtype.code == kDLFloat && dtype.bits == 32,
+               "core_distances must be float32 output tensor");
+}
+
```
jinsolp: Can we also add a check to ensure that the indices, optional distances, and optional core distances are device side pointers (if they are not nullptr)?

cpp/include/cuvs/neighbors/all_neighbors.h
```diff
@@ -0,0 +1,103 @@
...
+#include <cuvs/neighbors/nn_descent.h>
+#include <dlpack/dlpack.h>
+#include <stdint.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/**
+ * @defgroup all_neighbors_c_build All-neighbors C-API build (SNMG only)
```
jinsolp: I think we can remove the SNMG only : )

cpp/include/cuvs/neighbors/all_neighbors.h
```diff
@@ -0,0 +1,103 @@
...
+ * @{
+ *
+ * All-neighbors constructs an approximate k-NN graph for all vectors in a dataset.
+ * This SNMG C API can be used with a multi-GPU resources handle (`cuvsResources_t`)
+ * created from `raft::device_resources_snmg` to distribute work across GPUs.
```
jinsolp: Would be nice to add that it can be used with a regular single-gpu resource!

python/cuvs/cuvs/tests/test_all_neighbors.py
```diff
@@ -0,0 +1,225 @@
...
+        if n_features == 1
+        else np.column_stack(
+            (x, y, r.normal(size=(n_samples, max(0, n_features - 2))))
+        )
+    )
+    return X, y
+
+
+@pytest.mark.parametrize("algo", ["nn_descent", "brute_force", "ivf_pq"])
+@pytest.mark.parametrize("cluster", ["single_cluster", "multi_cluster"])
```
jinsolp, line 49: I believe the multi cluster build wouldn't run with device data 👀

python/cuvs/cuvs/tests/test_all_neighbors.py
```diff
@@ -0,0 +1,225 @@
...
+
+    assert indices.shape == (n_rows, k)
+    assert indices.dtype == cupy.int64
+    assert distances.shape == (n_rows, k)
+    assert distances.dtype == cupy.float32
+
+    recall = calc_recall(indices_host, bf_indices_host)
+
+    if snmg:
+        assert recall > 0.85
```
jinsolp: Do we experience recall drops with snmg?

jinsolp (commented):
Thanks for the work Victor! Left a few comments : )

python/cuvs/cuvs/common/mg_resources.pyx
jinsolp: I think Victor left an issue [here](https://github.com/rapidsai/cuvs/issues/1318). Please correct me if this isn't the right one @viclafargue

viclafargue: Thanks @jinsolp, an issue has indeed been opened: https://github.com/rapidsai/cuvs/issues/1318.

docs/source/indexes/all_neighbors.rst
viclafargue: The `all_neighbors.rst` does not document the C API though. In the end, what made the most sense was to rename the `indexes` directory to `neighbors` and `Nearest Neighbor Indexes` to `Nearest Neighbor`. I also updated the links.

python/cuvs/cuvs/tests/test_all_neighbors.py
viclafargue: In this test All Neighbors is built with `X_host`. `X_device` is used for the bruteforce comparison.

python/cuvs/cuvs/tests/test_all_neighbors.py
viclafargue: The recall for multi-GPU IVF-PQ was a bit low. I updated the `n_lists` parameters for the multi-GPU case. I set a unique threshold at 0.83 to be resilient to minor testing inconsistencies.

jinsolp: Got it. yeah ivfpq seemed to be very sensitive to `n_lists`.

python/cuvs/cuvs/tests/test_all_neighbors.py
jinsolp: Oh I see so `n_clusters` here is used for the number of clusters in the blobs dataset

python/cuvs/cuvs/tests/test_all_neighbors.py
```diff
@@ -0,0 +1,222 @@
...
+
+    if metric == "cosine":
+        X, _ = make_cosine(
+            n_samples=n_rows, n_features=n_cols, random_state=42
+        )
+    else:
+        X, _ = make_blobs(
+            n_samples=n_rows,
+            n_features=n_cols,
+            centers=n_clusters,
```
jinsolp: can we distinguish this `n_clusters` from the `n_clusters` for the all neighbors params? I think it's confusing to use the same value for the data generation/params value. Maybe we can just hardcode this to something like 5 in make_blobs?

docs/source/neighbors/all_neighbors.rst
```diff
@@ -0,0 +1,71 @@
...
+- Batch processing of large datasets across multiple GPUs
+
+All-neighbors supports multiple underlying algorithms:
+- **Brute Force**: Exact nearest neighbors computation
+- **IVF-PQ**: Approximate nearest neighbors using inverted file with product quantization
+- **NN-Descent**: Approximate nearest neighbors using graph-based descent
+
+The algorithm partitions the dataset into clusters and distributes the work across multiple GPUs when possible, making it suitable for large-scale graph construction tasks.
+
+[ :doc:`C++ API <../cpp_api/neighbors_all_neighbors>` | :doc:`Python API <../python_api/neighbors_all_neighbors>` ]
```
jinsolp: I think we should add the C API as well by adding a `neighbors_all_neighbors.rst` like we have for other index types here: https://github.com/rapidsai/cuvs/tree/branch-25.10/docs/source/c_api

docs/source/neighbors/all_neighbors.rst
```diff
@@ -0,0 +1,71 @@
...
+- **n_clusters**: Total number of clusters for work distribution
+- **metric**: Distance metric for graph construction
+- **algorithm-specific parameters**: IVF-PQ or NN-Descent specific settings
+
+Performance Characteristics
+-------------------------
+
+- **Build Time**: Scales with dataset size and chosen algorithm
+- **Memory Usage**: Depends on cluster size and overlap factor
+- **Accuracy**: Varies by algorithm (brute_force = exact, others = approximate)
```
jinsolp: brute force is not 100% exact when we have `n_clusters>1` because using batching is already an approximation. Can we add this information here? 👀 Maybe like "always approximate when `n_clusters>1`?

jinsolp (commented):
Looking great @viclafargue ! A few final minor comments about the docs 🙂

python/cuvs/cuvs/tests/test_all_neighbors.py
viclafargue: i used centers=10 since it seemed to give a better recall.

docs/source/neighbors/all_neighbors.rst
viclafargue: I added the C API and its documentation.

PR #1283, opened 2025-08-26
[Review][Java] Expand `CagraIndex#search` to accept more types (int8 and int32)

ldematte:
With the introduction of `CuVSMatrix` as the type to provide an input dataset, we expanded the types that can be used in indexing.
With this PR we aim to do the same query/search:
- Use a `CuVSMatrix` for `CagraQuery` too
- Use the correct types for sizes in `CagraIndex#search`

There is more we can do there, e.g. use `CuVSDeviceMatrix` for storing the results so we don't have to explicitly copy them (all duplicated code). That would be follow-up work.


ldematte:
Changing this back to draft as we need https://github.com/rapidsai/cuvs/pull/1361 as preliminary work

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/CagraIndexImpl.java
```diff
@@ -254,80 +251,90 @@ public SearchResults search(CagraQuery query) throws Throwable {
...
                 prepareTensor(
                     localArena, prefilterDP.handle(), prefilterShape, kDLUInt(), 32, kDLCUDA());
 
             cuvsFilter.type(prefilter, 1);
             cuvsFilter.addr(prefilter, prefilterTensor.address());
           }
 
-          returnValue = cuvsStreamSync(cuvsRes);
-          checkCuVSError(returnValue, "cuvsStreamSync");
+          // TODO: do we need this stream sync here?
```
benfred, line 308: I don't think you need to do this sync here - since cuvsCagraSearch is operating on the same stream

benfred (approved):

PR #1286, opened 2025-08-26
ScaNN: Overlapped gather for AVQ

rmaschal:
Adds a class cluster_loader for AVQ that enables overlapping the gather operation and HtoD copy with GPU computation.

There are two scenarios:
1. dataset on device: This is identical to the previous code, using raft::matrix::gather to perform the gather on device.
2. dataset on host: cluster_loader allocates to pinned buffers in host for fast (and possibly async) copies of cluster vectors DtoH. The actual gather operation is performed on cpu, into the pinned buffer. Copies can be overlapped with GPU work (namely AVQ update of the previous cluster) if scheduled on a separate stream.

### DIFF
--- a/cpp/src/neighbors/scann/detail/scann_avq.cuh
+++ b/cpp/src/neighbors/scann/detail/scann_avq.cuh
@@ -68,7 +68,8 @@ __global__ void build_clusters(
 template <typename LabelT, typename IdxT>
 void compute_cluster_offsets(raft::resources const& dev_resources,
                              raft::device_vector_view<const LabelT, IdxT> clusters,
-                             raft::device_vector_view<LabelT, int64_t> cluster_sizes)
+                             raft::device_vector_view<LabelT, int64_t> cluster_sizes,
+                             raft::host_scalar_view<int64_t> max_cluster_size)
 {
   cudaStream_t stream = raft::resource::get_cuda_stream(dev_resources);
   rmm::device_async_resource_ref device_memory =
@@ -103,10 +104,33 @@ void compute_cluster_offsets(raft::resources const& dev_resources,
                                       clusters.extent(0),
                                       stream);
 
-  temp_storage_bytes = 0;
+  int num_items = cluster_sizes.extent(0);
+
+  // Compute max cluster size
+  auto d_max_cluster_size = raft::make_device_scalar<int64_t>(dev_resources, 0);
+  temp_storage_bytes      = 0;
+
+  cub::DeviceReduce::Max(nullptr,
+                         temp_storage_bytes,
+                         cluster_sizes.data_handle(),
+                         d_max_cluster_size.data_handle(),
+                         num_items);
+
+  rmm::device_uvector<int64_t> temp_storage_max(temp_storage_bytes, stream, device_memory);
+
+  cub::DeviceReduce::Max(temp_storage_max.data(),
+                         temp_storage_bytes,
+                         cluster_sizes.data_handle(),
+                         d_max_cluster_size.data_handle(),
+                         num_items);
+
+  raft::copy(max_cluster_size.data_handle(),
+             d_max_cluster_size.data_handle(),
+             d_max_cluster_size.size(),
+             stream);
   // Scan to sum cluster sizes and get cluster start ptrs in flat array
   // Done in place
-  int num_items = cluster_sizes.extent(0);
+  temp_storage_bytes = 0;
 
   cub::DeviceScan::ExclusiveSum(nullptr,
                                 temp_storage_bytes,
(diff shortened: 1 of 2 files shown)


cpp/src/neighbors/scann/detail/scann_avq.cuh
```diff
@@ -396,63 +604,63 @@ void apply_avq(raft::resources const& res,
...
 
-  cudaStream_t stream = raft::resource::get_cuda_stream(res);
-  auto cluster_ptrs   = raft::make_device_vector<uint32_t, int64_t>(res, centroids_view.extent(0));
-  auto clusters       = raft::make_device_vector<uint32_t, int64_t>(res, dataset.extent(0));
-
-  compute_cluster_offsets(res, labels_view, cluster_ptrs.view());
+  cudaStream_t stream  = raft::resource::get_cuda_stream(res);
+  auto cluster_offsets = raft::make_device_vector<uint32_t, int64_t>(res, centroids_view.extent(0));
+  auto clusters        = raft::make_device_vector<uint32_t, int64_t>(res, dataset.extent(0));
+  auto max_cluster_size = raft::make_host_scalar<int64_t>(0);
```
tfeher: Why do we need a `host_scalar` instead of simply an `int64_t` variable? If there is no specific reason, then I would prefer simplicity, and recommend an `int64_t` variable.

cpp/src/neighbors/scann/detail/scann_avq.cuh
```diff
@@ -103,10 +104,33 @@ void compute_cluster_offsets(raft::resources const& dev_resources,
...
+
+  rmm::device_uvector<int64_t> temp_storage_max(temp_storage_bytes, stream, device_memory);
+
+  cub::DeviceReduce::Max(temp_storage_max.data(),
+                         temp_storage_bytes,
+                         cluster_sizes.data_handle(),
+                         d_max_cluster_size.data_handle(),
+                         num_items);
+
+  raft::copy(max_cluster_size.data_handle(),
```
tfeher: If you change the last argument to `int64_t max_cluster_size`, then this would simplify to 
```suggestion
  max_cluster_size = d_max_cluster_size.value(stream);
```

cpp/src/neighbors/scann/detail/scann_avq.cuh
```diff
@@ -396,63 +604,63 @@ void apply_avq(raft::resources const& res,
                raft::mdspan<const T, raft::matrix_extent<IdxT>, raft::row_major, Accessor> dataset,
                raft::device_matrix_view<T, IdxT> centroids_view,
                raft::device_vector_view<const LabelT, IdxT> labels_view,
-               float eta)
+               float eta,
+               cudaStream_t gather_stream)
```
bkarsin: If the gather primarily gathers from host into pinned memory, why do we need a dedicated stream?

cpp/src/neighbors/scann/detail/scann_build.cuh
```diff
@@ -179,8 +179,12 @@ index<T, IdxT> build(
...
 
   // AVQ update of KMeans centroids
-  apply_avq(
-    res, dataset, centroids_view, raft::make_const_mdspan(labels_view), params.partitioning_eta);
+  apply_avq(res,
+            dataset,
+            centroids_view,
+            raft::make_const_mdspan(labels_view),
+            params.partitioning_eta,
+            copy_stream);
```
bkarsin, line 187: Do we need all 3 separate streams to optimally overlap execution and gather? Any issues that could arise from RMM by using these 3 dedicated streams?

cpp/src/neighbors/scann/detail/scann_avq.cuh
```diff
@@ -368,6 +392,190 @@ void rescale_avq_centroids(raft::resources const& dev_resources,
...
+    auto cluster_ids = raft::make_device_vector_view<const LabelT, int64_t>(
+      cluster_ids_.data_handle() + h_cluster_offsets_(cluster_idx), size);
+
+    auto cluster_vectors =
+      raft::make_device_matrix_view<float, int64_t>(d_cluster_copy_buf_.data_handle(), size, dim_);
+
+    if (needs_copy_) {
+      // htod
+      auto h_cluster_ids =
+        raft::make_pinned_vector_view<LabelT, int64_t>(cluster_ids_buf_.data_handle(), size);
```
bkarsin, line 541: Is there a limit on what `size` could be? Seems to be based on cluster sizes only. If very few clusters are used, will pinned memory allocation be very costly?

cpp/src/neighbors/scann/detail/scann_avq.cuh
```diff
@@ -368,6 +392,190 @@ void rescale_avq_centroids(raft::resources const& dev_resources,
...
+                 int64_t max_cluster_size,
+                 int64_t h_buf_size,
+                 raft::host_vector_view<LabelT> h_cluster_offsets,
+                 raft::device_vector_view<LabelT> cluster_ids,
+                 bool needs_copy,
+                 cudaStream_t stream)
+    : dim_(dim),
+      n_rows_(n_rows),
+      dataset_ptr_(dataset_ptr),
+      cluster_buf_(raft::make_pinned_matrix<T, int64_t>(res, h_buf_size, dim)),
```
bkarsin, line 445: Any limit on `h_buf_size`? Could be very slow - may want a maximum pinned buffer size or check that it is reasonably small

cpp/src/neighbors/scann/detail/scann_avq.cuh
rmaschal: I also had to change d_max_cluster_size to type rmm::scalar

cpp/src/neighbors/scann/detail/scann_avq.cuh
rmaschal: There isn't a hard limit, but the expectation is that it's fairly small, something like sqrt(n) for a two-level tree and size n dataset.

cpp/src/neighbors/scann/detail/scann_avq.cuh
rmaschal: Same comment above, the expecatation is that the allocation is small. However, this is not an allocation, just making a view over the buffer allocated in the constructor.

cpp/src/neighbors/scann/detail/scann_avq.cuh
rmaschal: The gather stream is for copying, not gathering. I changed the name to copy_stream for clarity/consistency with scann_build

cpp/src/neighbors/scann/detail/scann_avq.cuh
rmaschal: For some reason I thought that was needed with raft::copy. I'm not sure why. int64_t is much more ergonomic. There are other places I should make this change.

cpp/src/neighbors/scann/detail/scann_build.cuh
rmaschal: Technically just two streams, one for copy, one for compute. I think the confusion is that copy_stream is passed as the argument "gather_stream" in apply_avq. Per the above, I changed the argument name to copy_stream for clarity/consistency.

cpp/src/neighbors/scann/detail/scann_avq.cuh
tfeher: Thanks for simplifying it!

cpp/src/neighbors/scann/detail/scann_avq.cuh
bkarsin: Ah, makes sense.

cpp/src/neighbors/scann/detail/scann_avq.cuh
bkarsin: Ok, so sqrt(n) should be sufficiently small for reasonably size dataset that can be used here.

bkarsin:
PR looks good to me, thanks for the clarification on my questions.

bkarsin (approved):
PR looks good to me, thanks (and very sorry for the mistaken close/open)

stic:
Hi @bkarsin, 
The assumption is that there is one cluster or clusters will be of same / similar size? Asking as if I've read it correctly max for buffers will be per biggest cluster?

PR #1296, opened 2025-08-27
[REVIEW] [Java] Option to build fat-jars with native dependencies included

### LINKED ISSUE #1293: [BUG] [Java] Race condition between building `src/main/java` and `src/main/java22`
I have observed/reported in the past that compiling the Java sources sometimes fails with the following error:
```
[ERROR] COMPILATION ERROR : 
[INFO] -------------------------------------------------------------
[ERROR] /home/mithunr/workspace/dev/cuvs/1/java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/common/Util.java:[160,18] cannot find symbol
  symbol:   method computeCapability()
  location: variable gpuInfo of type com.nvidia.cuvs.GPUInfo
[ERROR] /home/mithunr/workspace/dev/cuvs/1/java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/common/Util.java:[161,21] cannot find symbol
  symbol:   method totalMemory()
  location: variable gpuInfo of type com.nvidia.cuvs.GPUInfo
[ERROR] /home/mithunr/workspace/dev/cuvs/1/java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/common/Util.java:[203,13] cannot find symbol
  symbol:   constructor GPUInfo(int,java.lang.String,long,long,float)
  location: class com.nvidia.cuvs.GPUInfo
```
While working on the fat-jar builds, I'm now seeing other failures that are similar:
```
[ERROR] /home/mithunr/workspace/dev/cuvs/1/java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/BruteForceIndexImpl.java:[41,23] cannot find symbol
[ERROR]   symbol:   class BruteForceIndex
[ERROR]   location: package com.nvidia.cuvs
[ERROR] /home/mithunr/workspace/dev/cuvs/1/java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/BruteForceIndexImpl.java:[42,23] cannot find symbol
[ERROR]   symbol:   class BruteForceIndexParams
[ERROR]   location: package com.nvidia.cuvs
[ERROR] /home/mithunr/workspace/dev/cuvs/1/java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/BruteForceIndexImpl.java:[43,23] cannot find symbol
[ERROR]   symbol:   class BruteForceQuery
[ERROR]   location: package com.nvidia.cuvs
```
I have ruled out Java/JDK mix-ups.  The problem also persists for me with `maven-3.9.x`.  
Note that the failures are not consistent:  The compile fails about 50% of the time.  

This seems indicative of a race condition between `compile` and `compile-java-22`.  I think we need a way to ensure that the compile for `src/main/java` completes ahead of `compile-java-22`.

I am exploring moving the compiles to different maven phases (`compile` vs `process-classes`, etc.)

mythrocks:
This commit introduces an _option_ to include the native libraries as part of a new Java JAR artifact.

In addition, this commit also adds scripts to build the libraries included in the fat-jars using `gcc-toolset`, to allow the libraries to be portable across several Linux / `libstdc++` versions.  (This was earlier attempted in #1264, but will now reside in this commit.)

Note that for the initial cut, the "fat" jars will include only the following libraries:
1. `libcuvs.so`
2. `libcuvs_c.so`
3. `librmm.so`
4. `librapids_logger.so`

The resultant JARs will still be dependent on `LD_LIBRARY_PATH` for other dependencies (`cublas`, `cusparse`, `cusolver`, `nccl`, etc.).  

Two new profiles have been introduced in the `pom.xml`:
1. `x86_64-cuda12`
5. `x86_64-cuda13` (Although this is more of an example than anything.)

The main JAR artifact (`cuvs-java-25.x.x.jar`) remains unmodified.  But when `-P x86_64-cuda12` is employed, an additional `cuvs-java-25.x.x-x86_64-cuda12.jar` is produced, containing `libcuvs.so`, `libcuvs_c.so`, and some (minimal) additional dependencies.

The idea is that the JAR artifact build for `x86_64` + `cuda12` would look something like:
```bash
# On an x86 build box, using the cuda-12 conda env, from the project root directory:

# Build `libcuvs.so` first.
LIBCUVS_BUILD_DIR=`pwd`/cpp/build/cuda12 ./build.sh libcuvs

# Now the Java build.
cd java/
CMAKE_PREFIX_PATH=`pwd`/../cpp/build/cuda12 ./build.sh
```

The `java/build.sh` detects the CPU platform and the CUDA version to automatically choose the profile (`x86-cuda12`).

Note that there are tangential changes to the `pom.xml`:
1. Fixes #1293: The Java 22 portion of the build will now follow the Java 21, to prevent races between the builds.
2. The `maven-compiler-plugin` version has been dropped to `3.11.0`, to prevent build errors regarding a "0-byte module-info.class".  This is apparently a known issue in `3.13`, to be fixed in `3.14`.


mythrocks:
Note that the work in #1264 will depend on these changes.

java/cuvs-java/src/assembly/native-with-deps.xml
```diff
@@ -0,0 +1,52 @@
...
+            <useProjectArtifact>true</useProjectArtifact>
+            <unpack>true</unpack>
+            <scope>runtime</scope>
+        </dependencySet>
+    </dependencySets>
+
+    <!-- Include native libraries from separate directory -->
+    <fileSets>
+        <fileSet>
+            <directory>${project.build.directory}/native-libs</directory>
```
mythrocks, line 44: This is the assembly file that packages native-libraries into the native-jar artifact.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/spi/JDKProvider.java
```diff
@@ -31,6 +31,10 @@
 
 final class JDKProvider implements CuVSProvider {
 
+  static {
+    OptionalNativeDependencyLoader.loadLibraries();
```
mythrocks, line 35: This is the call into the optional native-dependency loader.  

If the jar includes native dependency libraries, they will be loaded at startup.  If not, the load is skipped, and it runs as before (i.e. depending on `$LD_LIBRARY_PATH`, etc.)

java/cuvs-java/src/main/java22/com/nvidia/cuvs/spi/OptionalNativeDependencyLoader.java
```diff
@@ -0,0 +1,82 @@
...
+
+import java.io.*;
+import java.net.URL;
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * A class that loads native dependencies if they are available in the jar.
+ */
+public class OptionalNativeDependencyLoader {
```
mythrocks, line 25: Implementation for the native dependency loader.

java/cuvs-java/pom.xml
```diff
@@ -108,10 +109,11 @@
             <plugin>
                 <groupId>org.apache.maven.plugins</groupId>
                 <artifactId>maven-compiler-plugin</artifactId>
-                <version>3.13.0</version>
+                <version>3.11.0</version>
```
mythrocks, line 112: The version of the `maven-compiler-plugin` had to be dropped to 3.11, to avoid build errors where Java22 might be built ahead of Java21.

java/cuvs-java/pom.xml
```diff
@@ -134,6 +136,9 @@
                                 <compileSourceRoot>${project.basedir}/src/main/java22</compileSourceRoot>
                             </compileSourceRoots>
                             <multiReleaseOutput>true</multiReleaseOutput>
+                            <excludes>
+                                <exclude>module-info.java</exclude>
```
mythrocks, line 140: `module-info.java` is skipped only for the Java22 build.  It's already built as part of the Java21 compile step.

This is to avoid errors during the Java 22 build, where `module-info.class` from Java 21 is clobbered.

java/cuvs-java/pom.xml
```diff
@@ -155,29 +160,6 @@
...
-                <executions>
-                    <execution>
-                        <id>assemble-all</id>
-                        <phase>package</phase>
-                        <goals>
-                            <goal>single</goal>
-                        </goals>
-                    </execution>
-                </executions>
-            </plugin>
```
mythrocks, line 180: This is no longer required. 

The fat-jar is built using `maven-resources-plugin`.

java/cuvs-java/pom.xml
```diff
@@ -277,4 +259,160 @@
...
+        <profile>
+            <id>x86-cuda12</id>
+            <activation>
+                <property>
+                    <name>cuda.version</name>
+                    <value>12</value>
+                </property>
+            </activation>
+            <properties>
+                <native.classifier>x86-cuda12</native.classifier>
```
mythrocks: Classifier for x86-cuda12.

java/cuvs-java/pom.xml
```diff
@@ -277,4 +259,160 @@
...
+        <profile>
+            <id>x86-cuda13</id>
+            <activation>
+                <property>
+                    <name>cuda.version</name>
+                    <value>13</value>
+                </property>
+            </activation>
+            <properties>
+                <native.classifier>x86-cuda13</native.classifier>
```
mythrocks: `x86-cuda13`.

java/build.sh
```diff
@@ -7,10 +7,22 @@ set -e -u -o pipefail
...
+# Identify CUDA major version.
+CUDA_VERSION_FROM_NVCC=$(nvcc --version | grep -oP 'release [0-9]+' | awk '{print $2}')
+CUDA_MAJOR_VERSION=${CUDA_VERSION_FROM_NVCC:-12}
+
+# Identify architecture.
+if [[ $(uname -m) == "aarch64" ]]; then
+  ARCH="arm64"
+else
+  ARCH="x86"
+fi
```
mythrocks: @jameslamb, @robertmaynard, @cjnolet:  I've referred to these as `x86` and `arm64` instead of `x86_64` and `aarch64` respectively.

This is the convention we used in the Spark RAPIDS JNI project: 
https://repo1.maven.org/maven2/com/nvidia/spark-rapids-jni/25.08.0/

I think it is easier to refer to the resultant Java libraries as `cuvs-java-25.10.0-x86-cuda12.jar`, instead of `cuvs-java-25.10.0-x86_64-cuda12.jar`.  But if this is a _faux pas_, I'd be happy to correct it.

java/build.sh
```diff
@@ -33,12 +45,12 @@ fi
 
 # Build the java layer
 if [ -z ${LD_LIBRARY_PATH+x} ]
-then export LD_LIBRARY_PATH=${CURDIR}/../cpp/build
-else export LD_LIBRARY_PATH=${CURDIR}/../cpp/build:${LD_LIBRARY_PATH}
+then export LD_LIBRARY_PATH=$CMAKE_PREFIX_PATH
+else export LD_LIBRARY_PATH=$CMAKE_PREFIX_PATH:${LD_LIBRARY_PATH}
```
mythrocks, line 45: This was changed so that we don't make assumptions on the build paths.

When building Java artifacts for `cuda12` or `cuda13`, I would like the option of putting the binaries in `cpp/build/cuda12` and `cpp/build/cuda13`.  This way, we don't lose progress when we switch environments.

In #1264, the Java build will look like:
```bash
# First, build libcuvs.so.
LIBCUVS_BUILD_DIR=`pwd`/cpp/build/cuda13 build.sh libcuvs
# Next, build Java, using the libraries built above.
CMAKE_PREFIX_PATH=`pwd`/cpp/build/cuda13 build.sh java
```

java/cuvs-java/src/main/java22/com/nvidia/cuvs/spi/OptionalNativeDependencyLoader.java
```diff
@@ -0,0 +1,82 @@
...
+  private static void loadLibrariesImpl() throws IOException {
+    if (!loaded) {
+      String os = System.getProperty("os.name");
+      String arch = System.getProperty("os.arch");
+      List<File> files = new ArrayList<>();
+      for (String file : filesToLoad) {
+        files.add(createFile(os, arch, file));
+      }
+      for (File file : files) {
+        System.out.println("Loading native dependency: " + file.getName());
```
mythrocks: TODO: This needs to be removed before merge.

java/build.sh
jameslamb: It'd be better to use the output of `$(arch)` (which would be `x86_64`), I think, so that every script handling these artifacts doesn't need to do the same type of `if ... else ...` you've introduced here.

But why do we need a CPU architecture specifier like that at all? I'm not very familiar with Java packaging, but it looks to me like Kafka and Spark, for example, are able to publish files with names without CPU architecture in the file name, like:

image

* https://repo1.maven.org/maven2/org/apache/kafka/kafka-clients/4.0.0/
* https://repo1.maven.org/maven2/org/apache/spark/spark-core_2.13/4.0.0/

Are those projects able to build multi-architecture binaries, and should we be doing the same here?

mythrocks: > But why do we need a CPU architecture specifier like that at all? ... Are those projects able to build multi-architecture binaries...

Good question.

You'll notice that neither Spark nor Kafka jars hold any linux-native executable code.  No DSOs, no .exe, nothing.  It's all pure Java class files, which are portable to multiple hardware platforms.  The same Spark jars will work on x86 and arm64.

That does not hold for the `cuVS` jars we're building here.  We're including the DSOs built for `x86` / `arm64`.  We need to distinguish between the jars for different platforms, lest the user deploy `arm64` binaries on `x86` boxen.

So should we consider a single jar that ships both `x86` and `arm64` binaries in the same jar file, in different subdirectories?  Maybe the custom dependency loader could examine the runtime and load the appropriate library?  One might have considered it if the native libraries weren't 1GB per.  We need to keep the jar files under 1GB in size to pass Maven Central's upload criteria.

mythrocks: On second thought, I might have missed what you're asking.

I know that you know we'd have different cuVS tarballs for x86 and arm64, and for cuda12/cuda13.  The fat-jars would be similar.  So it's possible I haven't understood the question properly.

jameslamb: No no you've understood correctly! This is perhaps just showing my ignorance of Java:

* I assumed that jar files were architecture-dependent even if they didn't contain DSOs (because I thought they had precompiled JVM bytecode, and I thought that was CPU-architecture-dependent)
* I did not look at the contents of those Kafka / Spark jars
* I did not know / remember the 1GB size limit for Maven Central

Thanks to your explanation, the separation of artifacts you've described makes sense to me. I still think it'd be better for the CPU architecture to be encoded as the output of `$(arch)` but don't feel strongly about that... consistency with Spark RAPIDS conventions is helpful too, because we're likely to borrow more from that project for this and future Java projects in RAPIDS.

mythrocks: In fairness, the `cuVS` packages aren't in complete sync with Spark RAPIDS JNI either.

The x86 packages for Spark RAPIDS JNI don't use the `x86` classifier, but [the ARM ones do mention `arm64`](https://repo1.maven.org/maven2/com/nvidia/spark-rapids-jni/25.08.0/) (as "arm64").

mythrocks: I've addressed this now.  `x86_64` is more of a mouthful, but explicit.

I couldn't find a ruling on the naming conventions for JNI fat-jars.  AFAICT, this should be acceptable.

mythrocks:
I've added a `README.md` for the `java/docker-build` directory that describes how these scripts are to be used to build for different scenarios. (E.g. cuda12/cuda13, local GPU arch vs `--allgpuarch`, etc.)

Also, I have just tested the build with `CUDA_VERSION=13.0.0`.  It's looking good so far.  The fat jars are 800MB each.  It's better than I'd hoped, honestly.  Only slightly bigger than [`spark-rapids`](https://repo1.maven.org/maven2/com/nvidia/rapids-4-spark_2.13/25.08.0/).

java/build.sh
```diff
@@ -33,12 +41,12 @@ fi
...
 fi
 
-export LD_LIBRARY_PATH=${CURDIR}/../cpp/build:${LD_LIBRARY_PATH}
 cd cuvs-java
-mvn verify "${MAVEN_VERIFY_ARGS[@]}" \
+mvn clean verify "${MAVEN_VERIFY_ARGS[@]}" -P "$BUILD_PROFILE" \
   && mvn install:install-file -Dfile=./target/cuvs-java-$VERSION.jar -DgroupId=$GROUP_ID -DartifactId=cuvs-java -Dversion=$VERSION -Dpackaging=jar \
+  && mvn install:install-file -Dfile=./target/cuvs-java-$VERSION-"$BUILD_PROFILE".jar -DgroupId=$GROUP_ID -DartifactId=cuvs-java -Dversion=$VERSION -Dclassifier="$BUILD_PROFILE" -Dpackaging=jar \
```
ldematte, line 51: Nice, and thank you for keeping the slim jar too!

java/cuvs-java/src/main/java22/com/nvidia/cuvs/spi/OptionalNativeDependencyLoader.java
ldematte: Reporting what we discussed offline and expanding it: 
I like the idea; as you see, I've gone down a very similar route in #1316 to check if `libcuvs` can be loaded, and if not why, before proceeding.
I think it would be great if we can merge the efforts, and have an `OptionalNativeDependencyLoader` that works for both cases (either a conditional implementation -- we can introspect the jar to see if it's the slim or fat version, or with 2 different implementations).

I think we should merge this as-is; I have verified that with these changes the slim jar keeps working and nothing breaks. However, if you don't mind, I have opened an issue to keep track of the follow up work, so we can have both this loader (which is necessary for the fat-jar to work) and another similar to/derived from #1316 for the slim jar (which is necessary for the slim-jar to work).

ldematte (approved):
I have concentrated my review and comments on the Java part, but I've also tried build-in-docker with success. LGTM

java/cuvs-java/pom.xml
```diff
@@ -124,7 +126,7 @@
                     </execution>
                     <execution>
                         <id>compile-java-22</id>
-                        <phase>compile</phase>
+                        <phase>process-classes</phase>
```
ldematte: I think this could be very problematic: this change breaks IntelliJ and makes very difficult to navigate code and use the IDE. 
Why was this changed?

java/cuvs-java/src/main/java22/com/nvidia/cuvs/spi/OptionalNativeDependencyLoader.java
```diff
@@ -0,0 +1,81 @@
...
+/**
+ * A class that loads native dependencies if they are available in the jar.
+ */
+public class OptionalNativeDependencyLoader {
+
+  private static final ClassLoader loader = JDKProvider.class.getClassLoader();
+
+  private static boolean loaded = false;
+
+  private static String[] filesToLoad = {
```
ldematte: Nit: this can be final. 
Also, following the convention of the rest of the codebase, I would rename it to `LIBRARIES_TO_LOAD`

java/cuvs-java/pom.xml
mythrocks: The reason I have this is because of the bug I reported before:  When building java22, I see occasions failures indicating that some symbols in the java21 build are missing. (Often in the base classes, or sometimes in Util.)

I finally tracked that down to the java22 build occasionally running before java21.  This change attempts to explicitly build 22 after 21. 

I will try rolling this part back to see if the problem occurs again.

ldematte: >This change attempts to explicitly build 22 after 21. 

It kind of makes sense, but messing with maven phases feels very hacky to me, and it's very strange it does not work as intended.
If you really have to do it, maybe you could use `process-sources` and `compile`, so at least the IDE still sees them both as source roots, but I'd call this a last resort.

mythrocks: I've tried building in different scenarios (within and outside the docker image).  Things seem to be holding.

I'm rolling this change back.  Thank you for the scrutiny here, @ldematte.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/spi/OptionalNativeDependencyLoader.java
mythrocks: Agreed.  This has now been fixed.

mythrocks:
I'm considering merging this tomorrow, unless @ldematte and @jameslamb have objections.  I think I've addressed the concerns we had for the moment.

jameslamb:
> _unless @ldematte and @jameslamb have objections_

Sorry, I'm not able to review this soon. Please talk with @robertmaynard about it, and don't wait on me.

Some quick notes on what I see:

* nothing in the `docker-build/` directory added here appears to being run in CI... and I think you'll have a hard time doing docker-in-docker when you try to get that running in CI in future PRs
* I still feel that it'd be easier to maintain and produce more-portable packages to instead use the existing Rocky Linux 8 images the rest of RAPIDS uses to build Python wheels, as I started prototyping in #1311

mythrocks:
Thank you, @jameslamb.  I'll wait to see if @robertmaynard has any objections to the contents of `docker-build/`.  (The other stuff, I'll leave in @ldematte's able hands.)

Re docker-in-docker:  ACK. That will need resolving, but I agree with your assessment:  It would be good to bake this into the CUVS CI container.  Devs could then use that.  I'll pick that thread up in a couple of weeks.

ldematte:
> I'm considering merging this tomorrow, unless @ldematte and @jameslamb have objections. I think I've addressed the concerns we had for the moment.

On the strictly Java (non-docker) part: GTG for me. We have plans on how extend it, but we'll address it separately after this and #1314 are merged.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/spi/OptionalNativeDependencyLoader.java
ldematte: We'll address this in a follow-up, no need to waste cycles for such a small nit!

robertmaynard:
Some follow-up work yall should think about is to update the C++ build flags used once https://github.com/rapidsai/cuvs/pull/1317 is merged.  That would allow you to build rmm/logger/fmt/spdlog statically and only have to bundle `libcuvs.so` and `libcuvs_c.so`

mythrocks:
> That would allow you to build rmm/logger/fmt/spdlog statically and only have to bundle `libcuvs.so` and `libcuvs_c.so`

Thank you for the heads-up.  Agreed.  That's would be to C++ libs what the fat-jar is for Java.

Come to that, I'm not quite sure why we have a separate `libcuvs.so` and a `libcuvs_c.so`.  The latter can't be used without the former.  

This is worth investigating further.

java/build.sh
robertmaynard: In general I recommend C++ builds be installed before being used downstream, instead of being used from the build directory. Ensure that downstream projects only use public API....

java/docker-build/build-in-docker
```diff
@@ -0,0 +1,38 @@
...
+export CUDA_VERSION=${CUDA_VERSION:-12.9.1}
+# CUDA_CLASSIFIER=cuda${CUDA_VERSION%%.*}  # Currently unused, but may be needed for future CUDA version handling
+
+# Set env for arm64 build, The possible values of 'uname -m' : [x86_64/i386/aarch64/mips/...]
+if [ "$(uname -m)" == "aarch64" ]; then
+  # ARM-specific settings.
+  echo "Set ARM Specific settings"
+fi
+
+# TODO: Settings to build for different CUDA versions, different GPU archs, CPU archs, etc.
```
robertmaynard, line 37: We should either error out if no arguments are passed into `build-in-docker` or default to `libcuvs java --allgpuarch`

java/docker-build/Dockerfile
```diff
@@ -0,0 +1,92 @@
...
+# Act as default GCC toolset in the image
+ARG TOOLSET_VERSION=14
+ARG CMAKE_VERSION=3.30.4
+ARG CCACHE_VERSION=4.11.2
+
+# Default x86_64 from x86 build, aarch64 cmake for arm build
+ARG CMAKE_ARCH=x86_64
+
+### Install basic requirements
+RUN dnf --enablerepo=crb install -y scl-utils gcc-toolset-${TOOLSET_VERSION} python39 \
```
robertmaynard: The dnf command will need to change based on the OS version.

For Rocky 8 you want `--enablerepo=devel` instead of crb

java/docker-build/Dockerfile
```diff
@@ -0,0 +1,92 @@
...
+#
+
+###
+# Build the image for spark-rapids-jni development environment.
+#
+# Arguments: CUDA_VERSION=[12.X.Y], OS_RELEASE=[8, 9], TARGETPLATFORM=[linux/amd64, linux/arm64]
+#
+###
+ARG CUDA_VERSION=12.9.1
+ARG OS_RELEASE=9
```
robertmaynard, line 24: Why is 8 / 9 an option here? I would have thought we would always build on 8

mythrocks: I've carried this bit forward from Spark RAPIDS.  That's left as an option.

I've been testing this with Rocky 9 with `gcc-toolset-14`, as has Elastic.  Would you recommend that we remove the option and stick with Rocky 8?

mythrocks: Per discussion, I'll retain support to use 8 and 9.  Thank you for clarifying, @robertmaynard.

java/docker-build/Dockerfile
mythrocks: Fixed with a conditional.

java/docker-build/build-in-docker
mythrocks: Agreed.  I've been putting that off.

This is now fixed.

java/build.sh
```diff
@@ -7,10 +7,18 @@ set -e -u -o pipefail
 ARGS="$*"
 NUMARGS=$#
 
-CURDIR=$(cd "$(dirname "$0")"; pwd)
 VERSION="25.10.0" # Note: The version is updated automatically when ci/release/update-version.sh is invoked
 GROUP_ID="com.nvidia.cuvs"
 
+# Identify CUDA major version.
+CUDA_VERSION_FROM_NVCC=$(nvcc --version | grep -oP 'release [0-9]+' | awk '{print $2}')
+CUDA_MAJOR_VERSION=${CUDA_VERSION_FROM_NVCC:-12}
```
msarahan, line 15:
```suggestion
CUDA_MAJOR_VERSION=$(nvcc --version | grep -oP 'release [0-9]+' | awk '{print $2}')
```

Any reason to have two variables here? My inclination would be to error out if the inner expression fails. The pipefail setting at the top of the script takes care of propagating the status.

java/docker-build/Dockerfile
```diff
@@ -0,0 +1,104 @@
...
+ARG TARGETPLATFORM=linux/amd64
+
+# multi-platform build with: docker buildx build --platform linux/arm64,linux/amd64 <ARGS> on either amd64 or arm64 host
+# check available official arm-based docker images at https://hub.docker.com/r/nvidia/cuda/tags (OS/ARCH)
+FROM --platform=$TARGETPLATFORM nvidia/cuda:$CUDA_VERSION-devel-rockylinux$OS_RELEASE
+
+# Re-declare ARGs after FROM since they don't persist across FROM boundary
+ARG CUDA_VERSION=12.9.1
+ARG OS_RELEASE=9
+ARG TARGETPLATFORM=linux/amd64
```
msarahan, line 34:
```suggestion
ARG CUDA_VERSION
ARG OS_RELEASE
ARG TARGETPLATFORM
```

Redeclaring doesn't need default values

java/cuvs-java/src/main/java22/com/nvidia/cuvs/spi/JDKProvider.java
msarahan: Is there a way to pre-load specific paths? With the [conda-pack](https://conda.github.io/conda-pack/) option, we'd need to force pre-load the bundled libstdc++ library. I'm mentioning this because this looks like it might be the right place to do that pre-loading. If you just rely on LD_LIBRARY_PATH, you would likely end up with the system libstdc++, which will be too old for the conda packages.

java/docker-build/Dockerfile
```diff
@@ -0,0 +1,104 @@
...
+ARG DEV_BUILD=OFF
+
+# Dependency versions
+# Act as default GCC toolset in the image
+ARG TOOLSET_VERSION=14
+ARG CMAKE_VERSION=3.30.4
+ARG CCACHE_VERSION=4.11.2
+
+# Default x86_64 from x86 build, aarch64 cmake for arm build
+ARG CMAKE_ARCH=x86_64
```
msarahan, line 46: Would be better to consume docker platform, so that thing are not defined in multiple places and get out of sync. Code generated by cursor.

```suggestion
    # Extract architecture part (remove linux/ prefix and any variant suffix)
    local arch="${TARGETPLATFORM#linux/}"
    arch="${arch%%/*}"  # Remove variant suffix like /v8, /v1, etc.
    
    # Convert to CMake platform naming
    case "$arch" in
        "amd64"|"x86_64")
            echo "x86_64"
            ;;
        "arm64"|"aarch64")
            echo "aarch64"
            ;;
        *)
            echo "Error: Unsupported architecture '$arch' in platform '$platform'" >&2
            echo "Supported architectures: amd64, x86_64, arm64, aarch64" >&2
            return 1
            ;;
    esac```

File: java/docker-build/run-in-docker [hunk 21]
```diff
@@ -0,0 +1,161 @@
...
+# Initialize DOCKER_BUILD_EXTRA_ARGS as an array
+if [[ -n "${DOCKER_BUILD_EXTRA_ARGS:-}" ]]; then
+    # Convert existing string to array by splitting on whitespace
+    read -ra DOCKER_BUILD_EXTRA_ARGS <<< "$DOCKER_BUILD_EXTRA_ARGS"
+else
+    DOCKER_BUILD_EXTRA_ARGS=()
+fi
+
+if [ "$(uname -m)" == "aarch64" ]; then
+    DOCKER_BUILD_EXTRA_ARGS=(--build-arg TARGETPLATFORM=linux/arm64 --build-arg CMAKE_ARCH=aarch64 "${DOCKER_BUILD_EXTRA_ARGS[@]}")
```
msarahan, line 58: this could be cleaned up to one line with the code above to translate TARGETPLATFORM to CMAKE_ARCH. I would not base the TARGETPLATFORM on the host CPU type. Better to pass it in as a parameter/env var so you don't need specific hardware to build a docker image.

mythrocks:
@msarahan Would it be alright if we left further changes to this in a separate follow up PR?  I'm afraid I'm away, and will be unavailable for the next 3 weeks. There are PRs queued up behind this one.

benfred (approved):

java/docker-build/run-in-docker
mythrocks: I agree with your point.  Cross-compilers would break this.

I had to start somewhere.

Issue #1298, opened 2025-08-28
Add docs for spectral embedding


aamijar:
Add docs for the `cuvs::preprocessing::spectral_embedding` api.
It should be accessible here https://docs.rapids.ai/api/cuvs/nightly/cpp_api/preprocessing/

PR #1299, opened 2025-08-28
Docs for spectral embedding

aamijar:
Resolves https://github.com/rapidsai/cuvs/issues/1298


cjnolet (approved):

cjnolet:
@aamijar the rust failure happened on a second run. I don't think it's flaky. can you take a look at the error and see if you can make sense of it (and possibly fix it)? Also tagging @benfred in case this makes sense to you.

aamijar:
According to @bdice, this is because it was built with older packages and is testing with newer packages. Pushing an empty commit to see if it passes.

PR #1300, opened 2025-08-28
[WIP] [Doc Update] CAGRA Memory Footprint

singhmanas1:
Goal: Easy to follow peak memory estimation for users. 

Changes Made:
1. Provided a formulae for CAGRA index build and search memory footprint under various scenarios- 
   1. build_algo = 'IVF-PQ', 'NN-descent'
   2. memory_resource = 'cuda memmory resource', 'managed_memory_resources'
2. Provided example calculation for each step. 

WIP - Need help with writing the 'nn_descent' peak memory consumption.

### DIFF
--- a/docs/source/indexes/cagra.rst
+++ b/docs/source/indexes/cagra.rst
@@ -164,7 +164,10 @@ per-query candidate neighbors (IDs + distances). Max batch size is 1024.
 
 **Example** (batch = 1024, dim = 1024, intermediate\_degree = 128): **5.00 MiB**
 
-**NN-descent peak memory** — *TBD* (depends on implementation details and parameters).
+**NN-descent peak memory** 
+~~~~~~~~~~~~~~~~~~~~~~~
+
+*TBD* (depends on implementation details and parameters).
 
 Optimize phase (device)
 ~~~~~~~~~~~~~~~~~~~~~~~
--- a/docs/source/indexes/cagra.rst
+++ b/docs/source/indexes/cagra.rst
@@ -100,11 +100,14 @@ The 3 hyper-parameters that are most often tuned are `graph_degree`, `intermedia
 Memory footprint
 ================
 
-CAGRA stores the dataset (raw vectors) and nearest-neighbor graph (neighbor IDs) in memory.  
+CAGRA stores the dataset (raw vectors) and nearest-neighbor graph (neighbor IDs) in memory.
 The dataset can be on **host or device**; the graph is **pinned to host memory** (staged to device during search).
+Use the formulas below to estimate baseline and peak memory needs.
 
 Baseline sizes
 --------------
+
+These are the steady-state sizes of the stored data structures.
 
 .. math::
 
@@ -118,16 +121,72 @@ Baseline sizes
    \;=\;
    \text{number\_vectors} \times \text{graph\_degree} \times \operatorname{sizeof}\!\big(\mathrm{IdxT}\big)
 
-**Example** (1,000,000 vectors, dim = 1024, 4-byte precision, graph\_degree = 64, IdxT = int32):
+**Example** (1,000,000 vectors, dim = 1024, fp32, graph\_degree = 64, IdxT = int32):
 
-- ``dataset_size = 1,000,000 * 1024 * 4 bytes = 3891 MB``  (approx.)
-- ``graph_size  = 1,000,000 * 64 * 4 bytes = 246 MB``      (approx.)
+- dataset\_size = 4,096,000,000 B = **3906.25 MiB** (~4096 MB)
+- graph\_size   =   256,000,000 B = **244.14 MiB** (~256 MB)
+
+Build peak memory usage
+-----------------------
+
+Index build has two phases: (1) construct an **intermediate graph**, then (2) **optimize** it (prune/reorder).
+The initial graph can be built with **IVF-PQ** (supports out-of-core, so datasets larger than GPU memory can be streamed)
+or **NN-descent**. The steps below are sequential (*not additive* across steps).
+
+Out-of-core IVF-PQ (sequential; not additive)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+**IVF-PQ Build (centroid training)** — uses a training subset to compute cluster centroids and assignments.
+
+.. math::
+
+   \text{IVFPQ\_build\_peak}
+   \;=\;
+   \frac{n\_{\text{vectors}}}{\text{train\_set\_ratio}} \times \text{dim} \times 4
+   \;+\;
+   n\_{\text{clusters}} \times \text{dim} \times 4
+   \;+\;
+   \frac{n\_{\text{vectors}}}{\text{train\_set\_ratio}} \times \operatorname{sizeof}(\mathrm{uint32\_t})
+
+**Example** (n = 1e6; dim = 1024; n\_clusters = 1024; train\_set\_ratio = 10): **395.01 MiB** (~414.19 MB)
+
+**IVF-PQ Search (forms the intermediate graph)** — batches vectors to the GPU, finds nearest lists, and records
+per-query candidate neighbors (IDs + distances). Max batch size is 1024.
+
+.. math::
+
+   \text{IVFPQ\_search\_peak}
+   \;=\;
+   \text{batch\_size} \times \text{dim} \times 4
+   \;+\;
+   \text{batch\_size} \times \text{intermediate\_degree} \times \operatorname{sizeof}(\mathrm{uint32\_t})
+   \;+\;
+   \text{batch\_size} \times \text{intermediate\_degree} \times 4
+
+**Example** (batch = 1024, dim = 1024, intermediate\_degree = 128): **5.00 MiB** (~5.24 MB)
+
+**NN-descent peak memory** — *TBD* (depends on implementation details and parameters).
+
+Optimize phase (device)
+~~~~~~~~~~~~~~~~~~~~~~~
+
+Pruning/reordering the intermediate graph; peak scales linearly with intermediate degree.
+
+.. math::
+
+   \text{optimize\_peak}
+   \;=\;
+   n\_{\text{vectors}} \times
+   \Big( 4 + \big(\operatorname{sizeof}(\mathrm{IdxT}) + 1\big)\times \text{intermediate\_degree} \Big)
+
+**Example** (n = 1e6, intermediate\_degree = 128, IdxT = int32): **614.17 MiB** (~644 MB)
 
 Search peak memory usage
 ------------------------
 
-Assumes a single in-flight batch (reuse buffers). For concurrent/overlapped batches, add one
-``result_size`` (and any per-batch scratch) per extra batch. Distances computed in fp32 by default.
+During search, the dataset and graph are staged to the GPU and per-batch buffers hold the current queries
+and their top-*k* results. If multiple batches run concurrently/overlapped, add one **result\_size**
+(and any per-batch scratch) per extra in-flight batch. Distances are fp32 by default.
 
 .. math::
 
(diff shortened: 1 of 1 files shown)


docs/source/indexes/cagra.rst
```diff
@@ -98,53 +98,148 @@ Tuning Considerations
 The 3 hyper-parameters that are most often tuned are `graph_degree`, `intermediate_graph_degree`, and `itopk_size`.
 
 Memory footprint
-----------------
+================
 
-CAGRA builds a graph that ultimately ends up on the host while it needs to keep the original dataset around (can be on host or device).
+CAGRA stores the dataset (raw vectors) and nearest-neighbor graph (neighbor IDs) in memory.
```
cjnolet: What memory? This rewrite of the sentence adds some good additional details but seems to have lost some important information.

docs/source/indexes/cagra.rst
```diff
@@ -98,53 +98,148 @@ Tuning Considerations
...
+The dataset can be on **host or device**; the graph is **pinned to host memory**.
```
cjnolet: Help me understand this a bit- why is this saying the graph is in pinned host memory? I don't see indication in the code that this is true.

docs/source/indexes/cagra.rst
```diff
@@ -98,53 +98,148 @@ Tuning Considerations
...
 
-IVFPQ or NN-DESCENT can be used to build the graph (additions to the peak memory usage calculated as in the respective build algo above).
+Baseline Memory Footprint
+--------------
 
-Dataset on device (graph on host):
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+These are the steady-state sizes of the stored data structures.
```
cjnolet: Maybe we should try to avoid the term "steady-state" here. I think what you mean is the memory footprint of the index itself after being trained, right?

docs/source/indexes/cagra.rst
```diff
@@ -98,53 +98,148 @@ Tuning Considerations
...
 
-Index memory footprint (device): :math:`n\_index\_vectors * n\_dims * sizeof(T)`
+.. math::
+
+   \text{dataset\_size}
```
cjnolet: We seem to have dropped the fact that the dataset is on device here. We should be clarify where the dataset lives when it's passed in. unfortunately, today it's not so straightforward because we make a copy of the dataset nearly every time (to pad it). 

However, we should also point out that the dataset need not be attached to the index at all after being built. We did this so that the user could turn around and convert the dataset to HNSW without having to attach the dataset at all.

docs/source/indexes/cagra.rst
```diff
@@ -98,53 +98,148 @@ Tuning Considerations
...
+
+**Example** (1,000,000 vectors, dim = 1024, fp32, graph\_degree = 64, IdxT = int32):
+
+- dataset\_size = 4,096,000,000 B = **3906.25 MiB**
+- graph\_size   = 256,000,000 B = **244.14 MiB**
+
+Build peak memory usage
+-----------------------
+
+Index build has two phases: (1) construct an intermediate graph, then (2) optimize it (prune/reorder).
```
cjnolet:
```suggestion

Index build has two phases: (1) construct a knn graph, then (2) optimize it to remove redundant and unecessary paths.
```

docs/source/indexes/cagra.rst
```diff
@@ -98,53 +98,148 @@ Tuning Considerations
...
+The initial graph can be built with IVF-PQ (supports out-of-core, so datasets larger than GPU memory can be streamed)
```
cjnolet:
```suggestion
The initial knn graph can be built with IVF-PQ or nn-descent. IVF-PQ has the additional benefit that it supports out-of-core construction, allowing CAGRA to be trained on datasets larger than available GPU memory.
```

docs/source/indexes/cagra.rst
```diff
@@ -98,53 +98,148 @@ Tuning Considerations
...
+or NN-descent. The steps below are sequential with distinct peak memory consumption. The overall peak memory utilization depends on the configured memory resource.
```
cjnolet:
```suggestion
The steps below are sequential with distinct peak memory consumption. The overall peak memory utilization depends on the configured RMM memory resource.
```

docs/source/indexes/cagra.rst
```diff
@@ -98,53 +98,148 @@ Tuning Considerations
...
+
+Out-of-core IVF-PQ
+~~~~~~~~~~~~~~~~~~~~~~~
```
cjnolet: This line needs to be the same size as the text above it.

docs/source/indexes/cagra.rst
```diff
@@ -98,53 +98,148 @@ Tuning Considerations
...
+
+**IVF-PQ Build (centroid training)** — uses a training subset to compute cluster centroids and assignments.
```
cjnolet: Maybe briefly summarize the way these steps are used together to scale the construction of the knn graph? Another thing we can (and should) add is the use of the all-neighbors graph API to construct the KNN graph in a super scalable way and then calling `cagra::optimize()` on the knn graph to yield the final CAGRA graph. That should require only a few lines of code and would allow a cagra graph to be built super fast. Further, we should update this to include enabling the new ACE algorithm. Not sure if we should include the CAGRA iterative build yet, as that hasn't been deep tested quite yet.

docs/source/indexes/cagra.rst
```diff
@@ -98,53 +98,148 @@ Tuning Considerations
...
+   \frac{n\_{\text{vectors}}}{\text{train\_set\_ratio}} \times \text{dim} \times 4
+   \;+\;
+   n\_{\text{clusters}} \times \text{dim} \times 4
+   \;+\;
+   \frac{n\_{\text{vectors}}}{\text{train\_set\_ratio}} \times \operatorname{sizeof}(\mathrm{uint32\_t})
+
+**Example** (n = 1e6; dim = 1024; n\_clusters = 1024; train\_set\_ratio = 10): **395.01 MiB**
+
+**IVF-PQ Search (forms the intermediate graph)** — batches vectors to the GPU, finds nearest lists, and records
+per-query candidate neighbors (IDs + distances). Max batch size is 1024.
```
cjnolet: I think this is confusing. I would just say that it constructs the knn graph in batches by querying the IVFPQ index for the nearest neighbors of all the training points.

docs/source/indexes/cagra.rst
```diff
@@ -98,53 +98,148 @@ Tuning Considerations
...
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+**Example** (batch = 1024, dim = 1024, intermediate\_degree = 128): **5.00 MiB**
 
-Index memory footprint (host): :math:`n\_index\_vectors * n\_dims * sizeof(T) + graph\_degree * n\_index\_vectors * sizeof(T)`
+**NN-descent peak memory** 
+~~~~~~~~~~~~~~~~~~~~~~~
 
-Build peak memory usage:
-~~~~~~~~~~~~~~~~~~~~~~~~
+*TBD* (depends on implementation details and parameters).
```
cjnolet: Do we want this in the docs or is this just a TODO for yourself as a follow-up in this PR?

docs/source/indexes/cagra.rst
```diff
@@ -98,53 +98,148 @@ Tuning Considerations
...
+   \;=\;
+   n\_{\text{vectors}} \times
+   \Big( 4 + \big(\operatorname{sizeof}(\mathrm{IdxT}) + 1\big)\times \text{intermediate\_degree} \Big)
+
+**Example** (n = 1e6, intermediate\_degree = 128, IdxT = int32): **614.17 MiB**
+
+Overall Index Build peak (device)
+---------------------
+
+Depending on the selected rmm memory resource, the overall peak memory footprint on the device would be different. For ``cuda_memory_resource``, peak is the maximum allocation across each step; For ``managed_memory_resource memory``, the
```
cjnolet: RMM should be capitalized. We use `device_memory_resource` and `host_memory_resource` to denote the differences. `device_memory_resource` is a type of `cuda_memory_resource`, but we never use that explicitly.

docs/source/indexes/cagra.rst
```diff
@@ -98,53 +98,148 @@ Tuning Considerations
...
+   \;+\;
+   \text{optimize\_peak}
+
+**Example:** 3906.25 + 395.01 + 5.00 + 614.17 = **4920.43 MiB**
+
+Search peak memory usage
+------------------------
 
-   + n\_vectors / train\_set\_ratio * sizeof(uint32_t)    // labels, may be in managed mem
+During search, the dataset and graph are staged to the GPU and per-batch buffers hold the current queries and their top-k results. If multiple batches run concurrently/overlapped, add one result\_size
```
cjnolet: The graph isn't copied during search, though. That would be slow as heck! It's already in device memory by the time the search begins. The only reason it's ever brought to host memory is because there are some parts of optimize() which are done on host. Technically we could do them on the GPU (and Vikram's team has actually done some work to port those).

docs/source/indexes/cagra.rst
```diff
@@ -98,53 +98,148 @@ Tuning Considerations
...
+(and any per-batch scratch) per extra in-flight batch. Distances are fp32 by default.
```
cjnolet: We should probably define what "scratch" means a little more clearly. Something like "Temporary memory used as a workspace for computations".

docs/source/indexes/cagra.rst
```diff
@@ -98,53 +98,143 @@ Tuning Considerations
 The 3 hyper-parameters that are most often tuned are `graph_degree`, `intermediate_graph_degree`, and `itopk_size`.
 
 Memory footprint
-----------------
+================
 
-CAGRA builds a graph that ultimately ends up on the host while it needs to keep the original dataset around (can be on host or device).
+CAGRA builds a nearest-neighbor graph(neighbor IDs) that ultimately ends up on the host while it needs to keep the original dataset(raw vectors) around (can be on host or device)
```
cjnolet: Small improvement here- can we maybe mention briefly that "keeping the dataset around" can be optional when immediately converting CAGRA to a CPU-based format like HNSW for search?

docs/source/indexes/cagra.rst
```diff
@@ -98,53 +98,143 @@ Tuning Considerations
...
 
-IVFPQ or NN-DESCENT can be used to build the graph (additions to the peak memory usage calculated as in the respective build algo above).
+Baseline Memory Footprint
+--------------
```
cjnolet: These bars need to be as long as the text in their corresponding heading.

docs/source/neighbors/cagra.rst
```diff
@@ -98,53 +98,143 @@ Tuning Considerations
...
-CAGRA builds a graph that ultimately ends up on the host while it needs to keep the original dataset around (can be on host or device).
+CAGRA builds a nearest-neighbor graph(neighbor IDs) that ultimately ends up on the host while it needs to keep the original dataset(raw vectors) around (can be on host or device)
 
-IVFPQ or NN-DESCENT can be used to build the graph (additions to the peak memory usage calculated as in the respective build algo above).
+Baseline Memory Footprint
+--------------
 
-Dataset on device (graph on host):
-~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+.. math::
```
cjnolet, line 110: Before we merge this- just want to veirfy that you did build the docs locally and checked that the math looks as expected?

docs/source/indexes/cagra.rst
```diff
@@ -98,53 +98,143 @@ Tuning Considerations
...
-   n\_vectors / train\_set\_ratio * dim * sizeof_{float}   // trainset, may be in managed mem
+   \text{optimize\_peak}
+   \;=\;
+   n\_{\text{vectors}} \times
+   \Big( 4 + \big(\operatorname{sizeof}(\mathrm{IdxT}) + 1\big)\times \text{intermediate\_degree} \Big)
+
+**Example** (n = 1e6, intermediate\_degree = 128, IdxT = int32): **614.17 MiB**
+
+Overall Index Build peak (device)
+---------------------
```
cjnolet: THe "underline" needs to be the same width as the text

docs/source/indexes/cagra.rst
```diff
@@ -98,53 +98,143 @@ Tuning Considerations
...
 
-   + n\_vectors / train\_set\_ratio * sizeof(uint32_t)    // labels, may be in managed mem
+Depending on the selected memory resource, the overall peak memory footprint on the device would be different. For ``cuda_memory_resource``, peak is the maximum allocation across each step; For ``managed_memory_resource memory``, the
```
cjnolet: We don't recommend using the `managed_memory_resource` to oversubscribe the GPU. I think we should remove this. It's not a good way to do "out of core" and if you oversubscribe a small GPU by 2x or more, the severe thrashing will end up making the algorithm slower than CPU.

docs/source/indexes/cagra.rst
```diff
@@ -98,53 +98,143 @@ Tuning Considerations
...
+   \text{IVFPQ\_search\_peak}
+   \;+\;
+   \text{optimize\_peak}
+
+**Example:** 3906.25 + 395.01 + 5.00 + 614.17 = **4920.43 MiB**
+
+Search peak memory usage
+------------------------
+
+CAGRA search is performed by having the dataset and graph in GPU memory, and requiring memory to store the search results of a batch of queries.
```
cjnolet: This is not quite true w/ CAGRA -Q. i think we can word this slightly better, because if using CAGRA-Q, the original dataset will actually be in host memory.

docs/source/neighbors/cagra.rst
singhmanas1: [cagra_memory.html](https://github.com/user-attachments/files/24969283/cagra_memory.html)

Here is the built html page

PR #1305, opened 2025-08-29
Fix redundant memset

vinaydes:


vinaydes:
The failure does not seem related to this PR. Should we rerun the tests?

PR #1306, opened 2025-09-02
Add note about Linux-only support for cuVS pre-compiled packages

stic:
As I've dived in into cuVS from outside RAPIDS project I wasn't aware of target platform. Hence lost coffee break trying to troubleshoot (along hallucinating GPT insisting it's issue with latest CUDA toolkit).

Added one liner to reference to RAPIDS docs on WSL2 support


cjnolet:
Hi @stic , unless you have new commits to add, please refrain from updating the branch in your PR. Each time you do so, it resets the CI checkers and we have to run CI again. After so many commits have been merged to the release branch, we might have to update the beanch to get CI checkers to pass, but the reviewers will normally do that as needed.

cjnolet (approved):
LGTM. Thanks for the contribution!

PR #1307, opened 2025-09-02
MG Python API

viclafargue:


docs/source/python_api/neighbors.rst
```diff
@@ -5,13 +5,25 @@ Nearest Neighbors
    :language: python
    :class: highlight
 
+Single-GPU Algorithms
+######################
```
benfred:
```suggestion
Single-GPU Algorithms
#####################
```

cpp/src/neighbors/mg_ivf_pq_c.cpp
```diff
@@ -381,28 +427,8 @@ extern "C" cuvsError_t cuvsMultiGpuIvfPqDistribute(cuvsResources_t res,
...
-      index->addr       = reinterpret_cast<uintptr_t>(_mg_distribute<int8_t>(res, filename));
-    } else if (dtype.kind == 'u' && dtype.itemsize == 1) {
-      index->dtype.code = kDLUInt;
-      index->addr       = reinterpret_cast<uintptr_t>(_mg_distribute<uint8_t>(res, filename));
-    } else {
-      RAFT_FAIL("Unsupported index dtype");
-    }
+    index->dtype.code = kDLFloat;
+    index->dtype.bits = 32;
+    index->addr       = reinterpret_cast<uintptr_t>(_mg_distribute<float>(res, filename));
```
benfred, line 432: Is this change intentional? (do we only support float32 here, and you meant to remove support for the other dtypes?)

python/cuvs/cuvs/neighbors/mg_ivf_flat/mg_ivf_flat.pxd
```diff
@@ -0,0 +1,129 @@
+#
```
benfred: Rather than have modules like `cuvs.neighbors.mg_ivf_flat` and `cuvs.neighbors.mg_cagra` - maybe we can move all of the multi gpu code into a `cuvs.neighbors.mg` module and then have `cuvs.neighbors.mg.ivf_flat` and `cuvs.neighbors.mg.cagra` as submodules of that?

python/cuvs/cuvs/neighbors/mg_ivf_flat/__init__.py
```diff
@@ -0,0 +1,39 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+from .mg_ivf_flat import (
+    MultiGpuIndex,
+    MultiGpuIndexParams,
+    MultiGpuSearchParams,
```
benfred: Can we rename these to just: 

```suggestion
    Index,
    IndexParams,
    SearchParams,
``` 

?

We already know that this is in multi-gpu because of the module name

python/cuvs/cuvs/neighbors/mg_cagra/mg_cagra.pyx
```diff
@@ -0,0 +1,461 @@
...
+        if self.params != NULL:
+            check_cuvs(cuvsCagraIndexParamsDestroy(self.params))
+
+        # Create multi-GPU params which includes embedded base params
+        check_cuvs(cuvsMultiGpuCagraIndexParamsCreate(&self.mg_params))
+        # Replace base pointer with embedded base params
+        self.params = self.mg_params.base_params
+
+    def __dealloc__(self):
+        # Only destroy the mg_params, which will handle base_params cleanup
```
benfred: Does the `__dealloc__` from the parent class not also run here, and try to free up `self.params` ? I'm wondering if there is a double free on this

python/cuvs/cuvs/neighbors/mg_cagra/mg_cagra.pyx
```diff
@@ -0,0 +1,461 @@
...
+        check_cuvs(cuvsMultiGpuCagraSearchParamsDestroy(self.mg_params))
+        self.mg_params = NULL
+        self.params = <cuvsCagraSearchParams_t>NULL
+
+    def __init__(self, *, search_mode="load_balancer",
+                 merge_mode="merge_on_root_rank",
+                 n_rows_per_batch=1000, **kwargs):
```
benfred: Can you add getters for `search_mode`  / `merge_mode` and `n_rows_per_batch`  ?

cpp/src/neighbors/mg_ivf_pq_c.cpp
viclafargue: This functions deserializes a single-GPU index and distributes it across multiple GPUs. The IVF-PQ index does not have a `T` template parameter, only `IdxT` (which is always `int64_t`). Because of this, a file containing a serialized IVF-PQ index is not prepended with an index type (float, half, uint8_t, int8_t). This makes it impossible to find the index type necessary to instantiate a multi-GPU index. This is fundamentally an issue in the multi-GPU API that should be remediated. In the meantime, we should assume that for this specific case (the replication of a single-GPU IVF-PQ index), the user is planning to search with floats.

benfred:
lgtm!

PR #1311, opened 2025-09-03
[DO NOT MERGE] cuvs-java: fat jars with libcuvs.so (proof-of-concept)

jameslamb:
Proof of concept that the goal of #1264 (for #1248) could be accomplished without using docker-in-docker in CI, and by re-using the same images that the rest of RAPIDS uses to build portable manylinux Python wheels.

Created this based on the discussion that led to https://github.com/rapidsai/cuvs/pull/1264#issuecomment-3246938538

## Notes for Reviewers

This is only meant to demonstrate what's possible... it'd need relevant changes from #1296 + other changes related to artifact organization and naming before being merged.

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -26,7 +26,7 @@ jobs:
           # - '13.0.0'
     with:
       build_type: pull-request
-      node_type: "gpu-l4-latest-1"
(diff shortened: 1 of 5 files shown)


jameslamb:
This was here as a proof-of-concept but I'm not planning to take it any farther. Closing it.

PR #1313, opened 2025-09-04
Fix pq_compressed file created by SSD Vamana build

bkarsin:
Fix issue with pq_compressed files being incorrect when GPU vamana build was being provided opq codebooks as an input to the build. Bug was caused by vq_centers not being set to 0s, despite OPQ compression not using vq centers, and only using pq_centers during quantized vector encoding.

Also cleaned up vamana_example file to support other datatypes and function correctly when building the DiskANN index without quantization.


cpp/src/neighbors/detail/vamana/vamana_build.cuh
```diff
@@ -388,6 +388,7 @@ auto quantize_all_vectors(raft::resources const& res,
 {
   auto dim         = residuals.extent(1);
   auto vq_codebook = raft::make_device_matrix<float, uint32_t, raft::row_major>(res, 1, dim);
+  raft::linalg::map_offset(res, vq_codebook.view(), [] __device__(size_t i) { return 0; });
```
tarang-jain: Can we use `raft::matrix::fill` for this?

bkarsin: Sure, I'll update this and verify correctness.

tarang-jain:
There are some merge conflicts. Other than that this PR looks good!

PR #1314, opened 2025-09-05
[Java] Add reason(s) to UnsupportedProvider/UnsupportedOperationExceptions

ldematte:
This PR adds some precise and readable messages specifying why a CuVSProvider is unsupported (it throws UnsupportedOperationException), to help with diagnosing cuvs-java startup issues.

PR #1315, opened 2025-09-05
[WIP][Java] Add `libcuvs` version check

ldematte:
This PR extends https://github.com/rapidsai/cuvs/pull/1314 to add version check (version reported by libcuvs vs. cuvs-java build version)

PR #1316, opened 2025-09-05
[Review][Java] Add detailed error message for `libcuvs` load failure to UnsupportedProvider/UnsupportedOperationExceptions

ldematte:
This PR further extends #1296 and #1314 to give meaningful error messages in case libcuvs fails to load.

The `jextract` generated bindings we use in cuvs-java use `SymbolLookup#libraryLookup` to load the `cuvs_c` dynamic library; this uses `RawNativeLibraries#load` (see https://github.com/openjdk/jdk/blob/master/src/java.base/share/native/libjava/RawNativeLibraries.c#L58); `RawNativeLibraries#load` in turn calls `JVM_LoadLibrary`. 

`JVM_LoadLibrary` does a good job to put together a good error message (e.g. calling `dlerror`, trying to locate and inspect the file for platform mismatch, etc. Unfortunately, `RawNativeLibraries#load` calls it passing false to the `throwException` parameter, which means that the detailed error messages are not surfaced.

This PR follows the pattern introduced in #1296 and preloads libcuvs (and dependencies) using `JVM_LoadLibrary` directly with `throwException` true; preloading it will also cause the OS to look for and load all dependencies. 
In case of error we can see what's broken in better detail; e.g. if `libcuvs_c.so` is present, but `librmm.so` is missing:

```
java.lang.UnsupportedOperationException: cannot create JDKProvider: libcuvs_c.so: librmm.so: cannot open shared object file: No such file or directory
        at com.nvidia.cuvs@25.10.0/com.nvidia.cuvs.spi.UnsupportedProvider.newCuVSResources(UnsupportedProvider.java:35)
        at com.nvidia.cuvs@25.10.0/com.nvidia.cuvs.CuVSResources.create(CuVSResources.java:90)
        at com.nvidia.cuvs@25.10.0/com.nvidia.cuvs.CuVSResources.create(CuVSResources.java:79)
```

Fixes https://github.com/rapidsai/cuvs/issues/1321


java/cuvs-java/src/main/java22/com/nvidia/cuvs/spi/JDKProvider.java
```diff
@@ -28,11 +34,77 @@
...
+        // RawNativeLibraries::load in turn calls JVM_LoadLibrary. Unfortunately, it calls it with a
+        // JNI_FALSE parameter for throwException, which means that the detailed error messages are
+        // not surfaced.
+        // Here we try and load it again, with throwException true, so we can see what's broken
+        try (var localArena = Arena.ofConfined()) {
+          var name = localArena.allocateFrom(System.mapLibraryName("cuvs_c"));
+          Object lib = JVM_LoadLibrary$mh.invoke(name, true);
+          if (lib != null) {
+            // It wasn't a problem with library loading, so undo what we did
+            JVM_UnloadLibrary$mh.invoke(lib);
```
ChrisHegarty: @ldematte nice sleuthing. Let's try to use System.localLibrary earlier so that we can get similar improved error messages.

ldematte: I like the idea! Less reflection :)
A first test seems to indicate that `System.loadLibrary` gives a different and less precise message. I'll dig into that. 
In any case, trying to load the library _before_ seems best.

ldematte:
This is marked WIP as I will need to rebase this on top of ~~#1315~~ #1327 for it to merge cleanly. Will be ready for review once I do that.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/spi/NativeDependencyLoader.java
```diff
@@ -0,0 +1,163 @@
...
+      return false;
+    }
+  }
+
+  private static boolean loaded = false;
+
+  static void loadLibraries() throws ProviderInitializationException {
+    if (!loaded) {
+      try {
+        LOADER_STRATEGY.loadLibraries();
```
ldematte, line 67: @chatman I think pre-loading libraries this way will already solve the problem with cudart, but if it does not we can add the System.loadLibrary around here (e.g. in the implementation of `loadLibraries`), then also call `cudaRuntimeGetVersion` and compare it with `embeddedLibrariesCudaVersion` for an additional check. 
Or the other way around: use `embeddedLibrariesCudaVersion` to load a specific version of cudart

(All of this in a follow-up I think)

ChrisHegarty (approved):
LGTM

java/cuvs-java/src/main/java22/com/nvidia/cuvs/spi/NativeDependencyLoader.java
```diff
@@ -0,0 +1,163 @@
...
+ */
+class NativeDependencyLoader {
+
+  interface NativeDependencyLoaderStrategy {
+    void loadLibraries() throws ProviderInitializationException;
+  }
+
+  private static final NativeDependencyLoaderStrategy LOADER_STRATEGY = createLoaderStrategy();
+
+  private static NativeDependencyLoaderStrategy createLoaderStrategy() {
```
mythrocks, line 39: 👏

mythrocks:
Thank you for this change, @ldematte.  It's now been merged.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/spi/NativeDependencyLoader.java
ldematte: Thanks a lot for the quick review!

PR #1323, opened 2025-09-09
[Java] Make `cudaGetDeviceProperties` compatible with CUDA 12 and 13 based on symbol presence

ldematte:
In #1273 we addressed a signature change in CUDA 13 by binding different symbols based on a environment variable, `RAPIDS_CUDA_MAJOR`. That works, but forces users of cuvs-java with CUDA 12 to define this environment variable.

This PR improves on it by making the symbol lookup dynamic, looking for the CUDA 13 symbol name, and falling back to the CUDA 12 exported name if we fail to locate the first.


mythrocks (approved):
This is definitely the preferable way to address this.

PR #1328, opened 2025-09-13
[Java] Uniform toHost/toDevice to work across all CuVSMatrix classes

ldematte:
This PR moves `toHost`/`toDevice` operations to `CuVSMatrix` in order to uniform them; implementations then do not need to worry or perform checks on the input type, but simply accept any `CuVSMatrix` and then "request" the type they want or need (no ifs, custom cudaAlloc or cudaMemcpy operations needed):
```
CuVSDeviceMatrix matrix = input.toDevice(resources);
```
If the source matrix is a `CuVSHostMatrix`, device memory will be allocated and data will be copied; if the source matrix is a already a `CuVSDeviceMatrix`, a (wrapped) reference to the original data is returned (very fast, no copies, no additional allocations).

The wrapper is currently implemented to follow C++ `weak_ptr` semantics: it delegates everything to the wrapped matrix, the only difference is on `close()` (which is a no-op in the wrapper, as you'd expect). The lifecycle of the data follows the one of the original matrix, the "weak" delegate does not influence it in any way. This way the caller can handle it uniformly (e.g. with try-with-resources), without side effects on the original matrix (which lifecycle will be already handled by its owner).

An alternative would be to give this C++ `shared_ptr` semantics, making it reference counted. I think this would be more complex and not necessary, but let me know.

### DIFF
--- a/cpp/include/cuvs/neighbors/cagra.h
+++ b/cpp/include/cuvs/neighbors/cagra.h
@@ -356,7 +356,7 @@ cuvsError_t cuvsCagraIndexDestroy(cuvsCagraIndex_t index);
  * @param[out] dim return dimension of the index
  * @return cuvsError_t
  */
-cuvsError_t cuvsCagraIndexGetDims(cuvsCagraIndex_t index, int* dim);
+cuvsError_t cuvsCagraIndexGetDims(cuvsCagraIndex_t index, int64_t* dim);
 
 /**
  * @brief Get size of the CAGRA index
@@ -365,7 +365,7 @@ cuvsError_t cuvsCagraIndexGetDims(cuvsCagraIndex_t index, int* dim);
  * @param[out] size return number of vectors in the index
  * @return cuvsError_t
  */
-cuvsError_t cuvsCagraIndexGetSize(cuvsCagraIndex_t index, uint32_t* size);
+cuvsError_t cuvsCagraIndexGetSize(cuvsCagraIndex_t index, int64_t* size);
 
 /**
  * @brief Get graph degree of the CAGRA index
@@ -374,7 +374,7 @@ cuvsError_t cuvsCagraIndexGetSize(cuvsCagraIndex_t index, uint32_t* size);
  * @param[out] graph_degree return graph degree
  * @return cuvsError_t
  */
-cuvsError_t cuvsCagraIndexGetGraphDegree(cuvsCagraIndex_t index, uint32_t* graph_degree);
+cuvsError_t cuvsCagraIndexGetGraphDegree(cuvsCagraIndex_t index, int64_t* graph_degree);
 
 /**
  * @brief Returns a view of the CAGRA dataset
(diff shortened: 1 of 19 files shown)


java/cuvs-java/src/main/java/com/nvidia/cuvs/CuVSHostMatrix.java
```diff
@@ -20,4 +20,77 @@
...
+  }
+
+  default CuVSDeviceMatrix toDevice(CuVSResources resources) {
+    var deviceMatrix = CuVSMatrix.deviceBuilder(resources, size(), columns(), dataType()).build();
+    toDevice(deviceMatrix, resources);
+    return deviceMatrix;
+  }
+
+  class CuVSHostMatrixDelegate implements CuVSHostMatrix {
+    private final CuVSHostMatrix hostMatrix;
```
mythrocks, line 35: I've not sure I've complete grokked this pattern;  The `CuVSHostMatrixDelegate` already implements `CuVSHostMatrix`.  Why use composition alongside inheritance?

Specifically, would it be viable not to override `get()`, `size()`, `toArray()`, etc?  Instead, perhaps we could leave those unimplemented, and bank on the implementation in `CuVSHostMatrix`?  Maybe we could implement only `close()` as a NOOP.

mythrocks: Actually, the more I think about it, using composition and inheritance together makes sense.

I can't think of a way to construct a delegating-host-matrix without explicitly specifying the underlying object.  

It's a shame; it would have been nice to use the base-class's sub-object, and avoid the forwarding.

java/cuvs-java/src/main/java/com/nvidia/cuvs/CuVSDeviceMatrix.java
```diff
@@ -39,4 +30,67 @@ default CuVSHostMatrix toHost() {
...
+    }
+
+    @Override
+    public void toDevice(CuVSDeviceMatrix deviceMatrix, CuVSResources cuVSResources) {
+      this.deviceMatrix.toDevice(deviceMatrix, cuVSResources);
+    }
+
+    @Override
+    public void close() {
+      deviceMatrix.close();
```
mythrocks: Wait, shouldn't this be a NOOP, like with `CuVSHostMatrixDelegate`?

java/cuvs-java/src/main/java/com/nvidia/cuvs/CuVSHostMatrix.java
ldematte: I think you are onto something here: since this is a inner class, we could use the reference to the inner class. But it would be effectively the same. In fact, I'd rather keep this obvious and explicit, make this a static class / not a inner class and pass the reference to the inner matrix explicitly

java/cuvs-java/src/main/java/com/nvidia/cuvs/CuVSDeviceMatrix.java
ldematte: Good catch, you are 100% correct.

java/cuvs-java/src/main/java/com/nvidia/cuvs/CuVSHostMatrix.java
ldematte: Also, since we are specifying this here on the interface, it has to be a static class (it is implicitly a static class).

mythrocks: ACK. Closing this.

ldematte:
> This PR makes a good change. 👍

Thank you! :)

PR #1332, opened 2025-09-16
[Java] Buffered device matrix builder

ldematte:
The current `CuVSDeviceMatrix` builder implementation uses a non-native `MemorySegment` created directly from on-heap java arrays to avoid an intermediate allocation and copy to a native (off-heap) segment.

However, this means that copy from host (Java heap) memory to device memory happens row by row (vector by vector), which is not ideal.

This PR introduces a new builder implementation; it still returns a `CuVSDeviceMatrix` backed by managed RMM device memory, but it uses an intermediate buffer (pinned, implemented by a new `PinnedMemoryBuffer` class extracted/refactored from the existing code) to batch data before copying it to the GPU.

Even if this means an intermediate allocation and copy to a native (off-heap) segment (so, data is copied twice), using pinned memory is up to 2x faster; benchmarks reveal that copying data directly from on-heap memory and using this intermediate buffer have similar throughput: 

Before (heap to GPU, row-by-row)
```
Benchmark                                            (dims)  (size)   Mode  Cnt   Score   Error  Units
CuVSDeviceMatrixBenchmarks.matrixDeviceBuilder         2048   16384  thrpt    5  34.932 ± 0.574  ops/s
```

After (heap to native memory to GPU, batched)
```
Benchmark                                            (dims)  (size)   Mode  Cnt   Score   Error  Units
CuVSDeviceMatrixBenchmarks.matrixDeviceBuilder         2048   16384  thrpt    5  35.622 ± 0.304  ops/s
```

So essentially identical; however, I think that despite the added complexity the change is still worth it as this method significantly reduces the number of calls to the GPU, reducing the risk of device locking due to syncs.
But I'm open to feedback, comments and suggestions!

### DIFF
--- a/java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/common/CloseableRMMAllocation.java
+++ b/java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/common/CloseableRMMAllocation.java
@@ -79,6 +79,7 @@ public class CloseableRMMAllocation implements CloseableHandle {
   public void close() {
     if (mustClose()) {
       checkCuVSError(cuvsRMMFree(cuvsResourceHandle, pointer, numBytes), "cuvsRMMFree");
+      pointer = MemorySegment.NULL;
     }
   }
 
--- a/java/cuvs-java/src/test/java/com/nvidia/cuvs/CagraBuildAndSearchIT.java
+++ b/java/cuvs-java/src/test/java/com/nvidia/cuvs/CagraBuildAndSearchIT.java
@@ -16,15 +16,13 @@
 package com.nvidia.cuvs;
 
 import static com.carrotsearch.randomizedtesting.RandomizedTest.assumeTrue;
+import static com.nvidia.cuvs.CuVSMatrixIT.assertSame2dArray;
 import static org.junit.Assert.*;
 
 import com.carrotsearch.randomizedtesting.RandomizedRunner;
 import com.nvidia.cuvs.CagraIndexParams.CagraGraphBuildAlgo;
 import com.nvidia.cuvs.CagraIndexParams.CuvsDistanceType;
 import com.nvidia.cuvs.CagraMergeParams.MergeStrategy;
-import java.io.File;
-import java.io.FileInputStream;
-import java.io.FileOutputStream;
 import java.io.InputStream;
 import java.lang.foreign.Arena;
 import java.lang.foreign.Linker;
@@ -397,12 +395,12 @@ public class CagraBuildAndSearchIT extends CuVSTestCase {
             .withMetric(CuvsDistanceType.L2Expanded)
             .build();
 
-    try (CuVSResources resources = CheckedCuVSResources.create()) {
-      CagraIndex index =
-          CagraIndex.newBuilder(resources)
-              .withDataset(dataset)
-              .withIndexParams(indexParams)
-              .build();
+    try (CuVSResources resources = CheckedCuVSResources.create();
+        CagraIndex index =
+            CagraIndex.newBuilder(resources)
+                .withDataset(dataset)
+                .withIndexParams(indexParams)
+                .build()) {
 
       // No prefilter (all points allowed)
       CagraSearchParams searchParams = new CagraSearchParams.Builder().build();
@@ -514,6 +512,10 @@ public class CagraBuildAndSearchIT extends CuVSTestCase {
     loadedIndex.close();
   }
 
+  /**
+   * Tests that an index built starting from a native MemorySegment is identical to one built from
+   * Java heap arrays
+   */
   @Test
   public void testNativeDatasetEquivalent() throws Throwable {
     float[][] sampleData = createSampleData();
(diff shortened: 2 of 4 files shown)


java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/common/PinnedMemoryBuffer.java
```diff
@@ -0,0 +1,81 @@
...
+    }
+  }
+
+  private static void destroyPinnedBuffer(MemorySegment bufferSegment) {
+    checkCudaError(cudaFreeHost(bufferSegment), "cudaFreeHost");
+  }
+
+  public MemorySegment address() {
+    if (hostBuffer == MemorySegment.NULL) {
+      //      System.out.println("Creating a buffer of size " + hostBufferBytes);
```
mythrocks: This might be safe to remove now.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/common/PinnedMemoryBuffer.java
```diff
@@ -0,0 +1,81 @@
...
+
+public class PinnedMemoryBuffer implements AutoCloseable {
+
+  private static final int CHUNK_BYTES =
+      8 * 1024 * 1024; // Based on benchmarks, 8MB seems the minimum size to optimize PCIe bandwidth
+  private final long hostBufferBytes;
+
+  private MemorySegment hostBuffer = MemorySegment.NULL;
+
+  public PinnedMemoryBuffer(long size, long columns, ValueLayout valueLayout) {
```
mythrocks: Nit: 
```suggestion
  public PinnedMemoryBuffer(long numRows, long columns, ValueLayout valueLayout) {
```

mythrocks:
Using pinned memory for a bounce buffer makes sense to do.  Good change, this.

Regarding the perf numbers posted in the description of the PR:  I was wondering how the results might scale for, say, millions of rows rather than 16K rows.

ldematte:
> Regarding the perf numbers posted in the description of the PR: I was wondering how the results might scale for, say, millions of rows rather than 16K rows.

I'll do some runs with higher-end hardware and bigger numbers

mythrocks (approved):

mythrocks:
How odd.  There are compile failures here:
```bash
Error:  Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.11.0:compile (compile-java-22) on project cuvs-java: Compilation failure: Compilation failure: 
Error:  /__w/cuvs/cuvs/java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/common/PinnedMemoryBuffer.java:[20,1] cannot find symbol
Error:    symbol:   static cudaFreeHost
Error:    location: class com.nvidia.cuvs.internal.panama.headers_h_1
Error:  /__w/cuvs/cuvs/java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/common/PinnedMemoryBuffer.java:[21,1] cannot find symbol
Error:    symbol:   static cudaMallocHost
Error:    location: class com.nvidia.cuvs.internal.panama.headers_h_1
```
Complaining about `cudaMallocHost` and `cudaFreeHost`, referenced from `PinnedMemoryBuffer`.  These seem to be resolved for me locally, from `headers_h_1`.  I'm not sure why the build is complaining.

benfred:
@mythrocks it's also working for the cuda 12.9 build in ci - but failing for on cuda 13.0, which is a bit weird (also, what version of cuda are you running locally?).

It looks like the previous code was also referencing cudaMallocHost, but was pulling in with https://github.com/rapidsai/cuvs/blob/7da2e183d7abe9a26669a737276ac0209345bfd6/java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/CuVSDeviceMatrixImpl.java#L19

instead of `headers_h_1` (not sure if this makes a difference or not?)

mythrocks:
@benfred:  Yeah, that does appear to be the difference.  

There should be a way to instruct `jextract` to put these symbols in a specific class/Java file.  

I think this is the result of using a single `headers.h` which includes every header we need.  The sheer number of symbols might be causing them to spill over to a different Java class (`headers_h_1` vs `headers_h`), non-deterministically.  I'm speculating.

cc @ChrisHegarty, @chatman for expertise.

ldematte:
That's really weird...
I also have CUDA 12.9 locally. Let me try with CUDA 13.0

> It looks like the previous code was also referencing cudaMallocHost, but was pulling in with 

I can definitely change the import. But it's kind of a problem that we have different imports for different versions -- it's a complication, as signatures do not change, so it would be nice to keep things independent of CUDA version.

ldematte:
Changing the import to `import static com.nvidia.cuvs.internal.panama.headers_h.*; ` works for both cuda12 and 13, so I'm going to make the change.

ldematte:
Just to close the loop: I run some benchmarks with bigger sizes (1M rows) and on enterprise hardware (A10G) and got very similar results -- we use around 1/4 of the PCIe bandwidth in both cases.

So I dug a little bit and measured each piece separately: it seems that the row-by-row copy to the pinned buffer balances out the effect of (slightly) faster copy from batches of vectors from pinned memory.
Still, since we cut down the number of copy operations and device syncs by 3 orders of magnitudes, the device should be free to parallelize other operations (computation or other copies).
Let me know if you have other ideas of things to try.

PR #1335, opened 2025-09-18
test ci

aamijar:
Check cuvs CI with https://github.com/rapidsai/raft/pull/2793


aamijar:
https://github.com/rapidsai/raft/pull/2793 was merged.

PR #1336, opened 2025-09-18
Improve performance of assigning clusters in batched all_neighbors

jinsolp:
Turns out we were using the wrong resource to create a per-device vector.


cjnolet (approved):
LGTM!

PR #1340, opened 2025-09-18
[Java] Adding tests to use CuVSDeviceMatrix (device memory) directly as a CagraIndex input dataset

ldematte:
When we introduced `CuVSDeviceMatrix` in #1232, we made it possible to use device-memory-backed dataset as an input for index build: since we accept a `CuVSMatrix`, and we have correct `toTensor` implementations for CPU and GPU, and the underlying functions in libcuvs support different memory types and sizes (through DLManagedTensor information), this became supported "naturally".
However, we never tested this explicitly. 

This PR adds tests to check and show that using CuVSDeviceMatrix (device memory) directly as a CagraIndex input dataset works as intended.

(similar tests for other index types will be added as follow-ups)


java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/common/CloseableRMMAllocation.java
```diff
@@ -79,6 +79,7 @@ private boolean mustClose() {
   public void close() {
     if (mustClose()) {
       checkCuVSError(cuvsRMMFree(cuvsResourceHandle, pointer, numBytes), "cuvsRMMFree");
+      pointer = MemorySegment.NULL;
```
ldematte, line 82: This is a small enhancement that came up while testing: it protects us from a double close().
I thought it was worth adding it.

ChrisHegarty (approved):
LGTM

java/cuvs-java/src/test/java/com/nvidia/cuvs/CuVSMatrixIT.java
```diff
@@ -254,10 +250,32 @@ public void testFloatDatasetCopy() {
...
+
+  static void assertSame2dArray(long rows, long cols, float[][] array1, float[][] array2) {
+    assertEquals(rows, array1.length);
+    assertEquals(cols, array1[0].length);
+    assertEquals(rows, array2.length);
+    assertEquals(cols, array2[0].length);
+
+    for (int n = 0; n < rows; ++n) {
+      for (int i = 0; i < cols; ++i) {
+        assertEquals(array1[n][i], array2[n][i], DELTA);
```
mythrocks, line 265: This is the sort of thing that C++ does nicely. :]

`assertSame2dArray` would be a function template.  After the element type `T` of `array1`  is deduced, line#265 would be a call to `compareEquals<T>`, which would have type-specific specializations.  

`compareEquals<float>` would use `DELTA` for the epsilon-compare.

mythrocks: All that said, this changed version does remove quite a bit of code repetition. 👏

ldematte: > This is the sort of thing that C++ does nicely. :]

Yep.
I have planned another iteration of test cleanup/extension, as I have spotted other places where we are uncovered/where we can improve things.
Let me think if I can further de-duplicate here.

mythrocks: > Let me think if I can further de-duplicate here.

I don't think that's necessary.  It's pretty sleek right now.  We can add polish later, if required.

mythrocks:
This change has been merged.  Thank you for the contribution, @ldematte.

PR #1345, opened 2025-09-19
[Java] Support row strides in CuVSMatrix

ldematte:
This PR adds full support for row strides to `CuVSMatrix`.

There was already some support for them in `CuVSDeviceMatrix`, as we needed to at least to save them to handle matrices built from tensors (e.g. resulting from a cuvs computation) and/or to create tensors for that kind of data (again, to pass them to the C API).
With this PR we introduce ways to:
- allocate host and device matrices with row stride != from the number of columns (e.g. padding/additional info at the end of row vectors)
- map a pre-allocated memory region to a `CuVSMatrix` taking row stride into account
- copy between different matrices with different strides
- populate data taking the target matrix stride into account
- read data from matrices (e.g. read rows or copy to heap arrays) taking the source matrix stride into account

The API changes include also the introduction of column strides ("gaps" between elements withing the same vector); however this is not implemented yet. To the best of my knowledge, cuvs functions do not support these _yet_. I wanted to introduce them as we need them in some scenarios (e.g. round-tripping from `DLTensor`s), and I did not want to change the API once again if/when we are going to support them.

However, I am open to drop `columnStride` if we do not want it. Let me know.


ldematte:
Note: this is partially done on top of https://github.com/rapidsai/cuvs/pull/1332; will rebase once it gets merged.

ldematte:
Another note:
I used "stride" because the C API (and [raft](https://github.com/rapidsai/raft/blob/c12777506b609275681b95fb147d8625fc452345/cpp/include/raft/util/cudart_utils.hpp#L169)) exposes it this way: `rowStride` is the width in memory _in number of elements_ of the 2D array.
The CUDA functions expose it as "pitch", i.e. width _in bytes_ (see [cudaMemcpy2DAsync](https://developer.download.nvidia.com/compute/DevZone/docs/html/C/doc/html/group__CUDART__MEMORY_gedfe4cb63f692bde9dcbfc9a74d9a70c.html)).

Personally, I would have preferred this way, but that would clash with the C and C++ APIs and make our code more complex (as we'd need to use CUDA functions directly instead of the more "friendly" cuvs C API).

java/cuvs-java/src/main/java/com/nvidia/cuvs/CuVSMatrix.java
```diff
@@ -99,23 +99,39 @@ interface Builder<T extends CuVSMatrix> {
   }
 
   /**
-   * Returns a builder to create a new instance of a dataset
+   * Returns a builder to create a new instance of a host-memory matrix
    *
-   * @param size     Number of vectors in the dataset
-   * @param columns  Size of each vector in the dataset
+   * @param size      Number of rows (e.g. vectors in a dataset)
+   * @param columns   Number of columns (e.g. dimension of each vector in the dataset)
```
mythrocks, line 105: 👏

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/common/PinnedMemoryBuffer.java
```diff
@@ -0,0 +1,81 @@
...
+import static com.nvidia.cuvs.internal.common.LinkerHelper.C_POINTER;
+import static com.nvidia.cuvs.internal.common.Util.checkCudaError;
+import static com.nvidia.cuvs.internal.panama.headers_h_1.cudaFreeHost;
+import static com.nvidia.cuvs.internal.panama.headers_h_1.cudaMallocHost;
+
+import java.lang.foreign.Arena;
+import java.lang.foreign.MemorySegment;
+import java.lang.foreign.ValueLayout;
+
+public class PinnedMemoryBuffer implements AutoCloseable {
```
mythrocks, line 26: This part of the diff will go away when #1332 is merged, right?  They look identical to my eye.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/CuVSDeviceMatrixRMMImpl.java
```diff
@@ -79,11 +79,27 @@ public static CuVSDeviceMatrixImpl create(
...
+    final long rowSize;
+    if (rowStride <= 0) {
+      rowSize = columns * elementSize;
+    } else if (rowStride >= columns) {
+      rowSize = rowStride * elementSize;
+    } else {
+      throw new IllegalArgumentException("Row stride cannot be less than the number of columns");
+    }
+    if (columnStride != -1) {
+      throw new IllegalArgumentException(
```
mythrocks: Ah, I see.  This is the part regarding column-striding that was mentioned in the description.

Nitpick:  Should this be an `UnsupportedOperationException`?

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/CuVSHostMatrixImpl.java
```diff
@@ -162,12 +214,13 @@ public void close() {}
 
   @Override
   public int get(int row, int col) {
-    return (int) accessor$vh.get(memorySegment, 0L, (long) row * columns + col);
+    return (int) accessor$vh.get(memorySegment, 0L, (long) row * rowStride + col);
```
mythrocks: The value of `rowStride` could be `-1` at this point, right?  Shouldn't we only use `rowStride` if it's been set to a valid value?

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/common/PinnedMemoryBuffer.java
ldematte: Yep, exactly! I'll take care of rebasing this on top of #1332

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/CuVSDeviceMatrixRMMImpl.java
ldematte: Yes, probably it's a better fit. I'll change that, thank you!

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/CuVSHostMatrixImpl.java
ldematte: Good catch! Elsewhere I used rowSize, but here the accessor already uses the layout for the element type, so I have to use either columns or rowStride.
I'll fix that. And add a test too.

ldematte:
@mythrocks failure is the same I addressed in #1332 ; I'll rebase this PR on #1332 as soon as it's merged, sorry for the noise

mythrocks:
> @mythrocks failure is the same I addressed in #1332 ; I'll rebase this PR on #1332 as soon as it's merged, sorry for the noise

Thanks for the fix, @ldematte.  I'm just waiting on #1332 to pass CI.  We'll get that in shortly.

mythrocks:
#1332 has been merged.  This should now be safe to rebase.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/CuVSHostMatrixImpl.java
```diff
@@ -214,7 +214,8 @@ public void close() {}
 
   @Override
   public int get(int row, int col) {
-    return (int) accessor$vh.get(memorySegment, 0L, (long) row * rowStride + col);
+    var rowPitch = rowStride > 0 ? rowStride : columns;
+    return (int) accessor$vh.get(memorySegment, 0L, (long) row * rowPitch + col);
```
mythrocks, line 228: :clap:

ldematte:
Thanks @mythrocks! I rebased it, and the duplicated changes are now gone (as well as the CUDA 13 compilation issue)

ldematte:
There was one CI failure, unrelated and reproducible in main. I created https://github.com/rapidsai/cuvs/issues/1387 to track it.

mythrocks:
It's occurred again.  I'm re-kicking the failed tests.

ldematte:
> It's occurred again. I'm re-kicking the failed tests.

That's kind of worrying tough. I think we need to investigate and find why it's failing -- and fix it.

mythrocks:
This time around, there are some flaky Python tests. :/

https://github.com/rapidsai/cuvs/actions/runs/18054200786/job/51382385024

PR #1346, opened 2025-09-19
Allow compilation when OpenMP is disabled

robertmaynard:
Since we compile CUVS with warnings as errors, we need to disable the unknown pragma warning generated by our OpenMP pragmas so we can still compile.


robertmaynard:
Do: merge

PR #1326, opened 2025-09-09
update dependencies: use cuda-toolkit wheels

jameslamb:
## Description

Contributes to https://github.com/rapidsai/build-planning/issues/213

* switches uses of `nvidia-cuda-*-cu{12,13}` wheels to https://pypi.org/project/cuda-toolkit/

### Benefits of these changes

Using the `cuda-toolkit` metapackage means we don't need to maintain our own pins for mutually compatible versions of CTK sub-components (which each follow their own different versioning conventions).

PR #1350, opened 2025-09-22
Treat warnings as errors in doc builds + MG ANNs Python API doc update

viclafargue:
This PR :
- updates doc build to treat warnings as errors
- fixes an issue in the MG ANNs Python API doc
- fixes several doc issues that resulted in warnings

Closes https://github.com/rapidsai/cuvs/issues/1382


benfred (approved):
Changes look good to me - 

I tried to make the docs build fail on warnings w/ #1301  so that these kind of errors wouldn't happen in the future. I think I just updated `./build.sh` in that PR - and it looks like I also needed to update `ci\build_docs.sh` in the same way (` sphinx-build -W`) to get CI to treat docs warnings as errors. 

 Can we also change the `ci/docs_build.sh` to enable warnings as errors in this PR, so that this kind of error won't happen again in the future?

benfred (approved):

ci/build_docs.sh
```diff
@@ -48,7 +48,7 @@ popd
 
 rapids-logger "Build Python docs"
 pushd docs
-sphinx-build -b dirhtml source _html
+sphinx-build -W -b dirhtml source _html
```
jameslamb: Consider putting this into the `Makefile` and using `make dirhtml` in any scripts where you want to build the docs:

https://github.com/rapidsai/cuvs/blob/c07ab70786d5553d4bb726a9d25aa643352ef946/docs/Makefile#L5

That way, the settings would apply still if you switched from `dirhtml` to `html`, and you wouldn't have to keep this in sync with this other use:

https://github.com/rapidsai/cuvs/blob/c07ab70786d5553d4bb726a9d25aa643352ef946/build.sh#L520

See how `cudf` does this for an example:

https://github.com/rapidsai/cudf/blob/1504ddc2ebb39b3be0847e680b8d76e8cc37a7a1/docs/cudf/Makefile#L5

https://github.com/rapidsai/cudf/blob/1504ddc2ebb39b3be0847e680b8d76e8cc37a7a1/ci/build_docs.sh#L56

jameslamb (approved):
Left one suggestion for your consideration.

Please also update the PR title and add a short description, so others finding this from search and the git history can understand what its purpose was.

I think it is something like "treat warnings as errors in doc builds".

Approving so you don't need another review from me to merge, but please `@` me if you have any questions.

ci/build_docs.sh
cjnolet: @viclafargue do you mind creating a Github issue for this?

cjnolet (approved):

ci/build_docs.sh
viclafargue: Sure, I created https://github.com/rapidsai/cuvs/issues/1382.

jameslamb:
One Java test failed:

```text
Error:  Failures: 
Error:    CagraRandomizedIT.testResultsTopKWithRandomValues:55->tmpResultsTopKWithRandomValues:181->CuVSTestCase.compareResults:114 Not found in expected list: 7106
```

([build link](https://github.com/rapidsai/cuvs/actions/runs/18032615689/job/51315637945?pr=1350#step:11:10333))

That's unlikely to be related to these changes, and I've seen that test be flaky before (https://github.com/rapidsai/cuvs/pull/1326#issuecomment-3276112293), so I re-ran that job for you. Hopefully we'll see it pass.

PR #1354, opened 2025-09-22
[Review] ScaNN: Add option for AVQ/Noise Shaping to bfloat16 quantization

rmaschal:
This PR adds support for AVQ loss/Noise shaping to the BFloat16 dataset quantization. 

AVQ loss is a modified version of L2 loss which separately penalizes the components of the residual vector which are parallel and perpendicular to the original vector. Quantizing vectors with AVQ loss rather than L2 loss gives a better approximation of the inner product, and thus performs better in Maximal Innter Product Search (https://arxiv.org/abs/1908.10396).

Math:
x : original vector
x_q : quantized vector
r = x - x_q : residual vector
r_para = < r , x > x / || x ||^2 : parallel component of the residual
r_perp = r - r_para : perpendicular compoent of the residual
eta >= 1 : AVQ parameter
AVQ loss = eta * || r_para ||^2 +   || r_perp ||^2

For a float vector x, the goal is to find a bfloat16 vector x_q which minimizes the AVQ loss for a given eta. Unlike L2, AVQ loss is not separable (e.g. ||r_para||^2 contains cross terms from the inner product), so we cannot optimize individual dimensions in parallel and expect convergence. Instead, we use coordinate descent to optimize dimensions of x_q one at a time,  until convergence. 

This coordinate descent happens in the new kernel "quantize_bfloat16_noise_shaped_kernel". For efficient memory accesses and compute, one warp is assigned to optimize each dataset vector. The computation of avq loss is algebraically separated into two pieces: those which can be computed in parallel (i.e. those only depending on local information for the assigned dimension) and those which require global information (namely those depending on < r , x >). Finally threads in a warp serialize to compute the final cost for their dimension, update the quantized value and value of < r , x > (if applicable), and broadcast the updated value of < r, x > for other threads. This continues in blocks of 32 dimensions, until convergence (or a maximum of 10 iterations). 

I've found this strategy does a good job taking advantage of the inherently row_major structure of the dataset/index for efficient coalesced accesses, while still making good use of compute resources (hitting >90% compute throughput on an A6000). 

Besides the coordinate descent kernel, this PR adds some helper functions for the above, refactors the existing bfloat16 to take advantage of them, and adds configuration for the AVQ eta (code uses normal bfloat16 quantization when avq threshold is NaN).

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -42,8 +42,7 @@ ENV HISTFILE="/home/coder/.cache/._bash_history"
 ENV AWS_ROLE_ARN="arn:aws:iam::279114543810:role/nv-gha-token-sccache-devs"
 ENV SCCACHE_REGION="us-east-2"
 ENV SCCACHE_BUCKET="rapids-sccache-devs"
-# 2hr (1 minute longer than sccache-dist request timeout)
-ENV SCCACHE_IDLE_TIMEOUT=7200
+ENV SCCACHE_IDLE_TIMEOUT=0
 
 ###
 # sccache-dist configuration
@@ -54,13 +53,8 @@ ENV DEVCONTAINER_UTILS_ENABLE_SCCACHE_DIST=1
 ENV SCCACHE_DIST_FALLBACK_TO_LOCAL_COMPILE=true
 # Retry transient errors 4 times (for a total of 5 attempts)
 ENV SCCACHE_DIST_MAX_RETRIES=4
-ENV SCCACHE_DIST_CONNECT_TIMEOUT=30
-ENV SCCACHE_DIST_CONNECTION_POOL=false
 # 1hr 59min (to accommodate debug builds)
 ENV SCCACHE_DIST_REQUEST_TIMEOUT=7140
-ENV SCCACHE_DIST_KEEPALIVE_ENABLED=true
-ENV SCCACHE_DIST_KEEPALIVE_INTERVAL=20
-ENV SCCACHE_DIST_KEEPALIVE_TIMEOUT=600
 ENV SCCACHE_DIST_URL="https://${TARGETARCH}.linux.sccache.rapids.nvidia.com"
 
 # Build as much in parallel as possible
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -3,7 +3,7 @@ name: build
 on:
   push:
     branches:
-      - "branch-*"
+      - "main"
     tags:
       - v[0-9][0-9].[0-9][0-9].[0-9][0-9]
   workflow_dispatch:
--- a/RAPIDS_BRANCH
+++ b/RAPIDS_BRANCH
@@ -1,1 +1,1 @@
-branch-25.12
+main
--- a/c/CMakeLists.txt
+++ b/c/CMakeLists.txt
@@ -106,7 +106,29 @@ set_target_properties(
              EXPORT_NAME c_api
 )
 
-target_compile_definitions(cuvs_c PUBLIC $<$<BOOL:${BUILD_CAGRA_HNSWLIB}>:CUVS_BUILD_CAGRA_HNSWLIB>)
+# Encode what optional components we are building
+set(CUVS_BUILD_CAGRA_HNSWLIB ${BUILD_CAGRA_HNSWLIB})
+set(CUVS_BUILD_MG_ALGOS ${BUILD_MG_ALGOS})
+# cmake-lint: disable=E1126
+file(
+  CONFIGURE
+  OUTPUT
+  "${CMAKE_CURRENT_BINARY_DIR}/include/cuvs/core/c_config.h"
+  CONTENT
+  [=[
+  #pragma once
+  #ifndef CUVS_BUILD_CAGRA_HNSWLIB
+    #cmakedefine CUVS_BUILD_CAGRA_HNSWLIB
+  #endif
+  #ifndef CUVS_BUILD_MG_ALGOS
+    #cmakedefine CUVS_BUILD_MG_ALGOS
+  #endif
+]=]
+)
+target_compile_definitions(
+  cuvs_c PUBLIC $<$<BOOL:${BUILD_CAGRA_HNSWLIB}>:CUVS_BUILD_CAGRA_HNSWLIB>
+                $<$<BOOL:${BUILD_MG_ALGOS}>:CUVS_BUILD_MG_ALGOS>
+)
 
 target_compile_options(cuvs_c PRIVATE "$<$<COMPILE_LANGUAGE:CXX>:${CUVS_CXX_FLAGS}>")
 
@@ -114,6 +136,7 @@ target_include_directories(
   cuvs_c
   PUBLIC "$<BUILD_INTERFACE:${DLPACK_INCLUDE_DIR}>"
          "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>"
+         "$<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}/include>"
   INTERFACE "$<INSTALL_INTERFACE:include>"
 )
 
@@ -159,6 +182,11 @@ if(PROJECT_IS_TOP_LEVEL)
     DIRECTORY include/cuvs
     COMPONENT cuvs_c
     DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}
+  )
+  install(
+    FILES ${CMAKE_CURRENT_BINARY_DIR}/include/cuvs/core/c_config.h
+    COMPONENT cuvs_c
+    DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/cuvs/core/
   )
 endif()
 
--- a/c/include/cuvs/core/all.h
+++ b/c/include/cuvs/core/all.h
@@ -1,0 +1,53 @@
+/*
+ * Copyright (c) 2025, NVIDIA CORPORATION.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+// This is a list of required header files that cuvs_c rust/go/java
+// will bind to.
+
+#pragma once
+
+#include <cuvs/core/c_config.h>
+#include <cuvs/core/c_api.h>
+
+#include <cuvs/cluster/kmeans.h>
+
+#include <cuvs/distance/distance.h>
+#include <cuvs/distance/pairwise_distance.h>
+
+#include <cuvs/neighbors/all_neighbors.h>
+#include <cuvs/neighbors/brute_force.h>
+#include <cuvs/neighbors/cagra.h>
+#include <cuvs/neighbors/common.h>
+#include <cuvs/neighbors/ivf_flat.h>
+#include <cuvs/neighbors/ivf_pq.h>
+#include <cuvs/neighbors/nn_descent.h>
+#include <cuvs/neighbors/refine.h>
+#include <cuvs/neighbors/tiered_index.h>
+#include <cuvs/neighbors/vamana.h>
+
+#ifdef CUVS_BUILD_CAGRA_HNSWLIB
+  #include <cuvs/neighbors/hnsw.h>
+#endif
+
+#ifdef CUVS_BUILD_MG_ALGOS
+  #include <cuvs/neighbors/mg_cagra.h>
+  #include <cuvs/neighbors/mg_common.h>
+  #include <cuvs/neighbors/mg_ivf_flat.h>
+  #include <cuvs/neighbors/mg_ivf_pq.h>
+#endif
+
+#include <cuvs/preprocessing/quantize/binary.h>
+#include <cuvs/preprocessing/quantize/scalar.h>
--- a/c/include/cuvs/neighbors/cagra.h
+++ b/c/include/cuvs/neighbors/cagra.h
@@ -214,14 +214,14 @@ enum cuvsCagraSearchAlgo {
   /** For small batch sizes. */
   MULTI_CTA    = 1,
   MULTI_KERNEL = 2,
-  AUTO         = 3
+  AUTO         = 100
 };
 
 /**
  * @brief Enum to denote Hash Mode used while searching CAGRA index
  *
  */
-enum cuvsCagraHashMode { HASH = 0, SMALL = 1, AUTO_HASH = 2 };
+enum cuvsCagraHashMode { HASH = 0, SMALL = 1, AUTO_HASH = 100 };
 
 /**
  * @brief Supplemental parameters to search CAGRA index
--- a/c/include/cuvs/neighbors/common.h
+++ b/c/include/cuvs/neighbors/common.h
@@ -16,7 +16,6 @@
 
 #pragma once
 
-#include <cuvs/distance/distance.h>
 #include <stdint.h>
 
 #ifdef __cplusplus
--- a/c/include/cuvs/neighbors/mg_common.h
+++ b/c/include/cuvs/neighbors/mg_common.h
@@ -16,7 +16,6 @@
 
 #pragma once
 
-#include <cuvs/core/c_api.h>
 #include <stdint.h>
 
 #ifdef __cplusplus
--- a/c/include/cuvs/preprocessing/quantize/binary.h
+++ b/c/include/cuvs/preprocessing/quantize/binary.h
@@ -30,7 +30,11 @@ extern "C" {
  * and sampling_median thresholds are calculated separately for each dimension.
  *
  */
-enum cuvsBinaryQuantizerThreshold { ZERO, MEAN, SAMPLING_MEDIAN };
+enum cuvsBinaryQuantizerThreshold {
+  ZERO=0,
+  MEAN=1,
+  SAMPLING_MEDIAN=2
+};
 
 /**
  * @brief Binary quantizer parameters.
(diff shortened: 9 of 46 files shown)


cpp/include/cuvs/neighbors/scann.hpp
```diff
@@ -74,8 +75,12 @@ struct index_params : cuvs::neighbors::index_params {
   uint32_t pq_train_iters = 10;
 
   /** whether to apply bf16 quantization of dataset vectors **/
-  bool bf16_enabled = false;
+  bool reordering_bf16_enabled = false;
```
cjnolet: Small nitpick- the fact that this is a boolean kind of suggests the `_enabled` part. Can we just call this `reorder_bf16`?

cpp/src/neighbors/scann/detail/scann_build.cuh
```diff
@@ -313,16 +313,9 @@ index<T, IdxT> build(
...
-        bf16_dataset.data_handle(),
-        batch_view.data_handle(),
-        batch_view.size(),
-        [] __device__(T x) {
-          nv_bfloat16 val = __float2bfloat16(x);
-          return reinterpret_cast<int16_t&>(val);
-        },
-        resource::get_cuda_stream(res));
+    if (params.reordering_bf16_enabled) {
+      quantize_bfloat16(
```
cjnolet, line 317: Is this somethig we can use more broadly and expose in our preprocessing API?

cpp/src/neighbors/scann/detail/scann_build.cuh
```diff
@@ -340,7 +333,7 @@ index<T, IdxT> build(
                quantized_soar_residuals.size(),
                stream);
 
-    if (params.bf16_enabled) {
+    if (params.reordering_bf16_enabled) {
```
cjnolet: I'm looking over your implementation here trying to figure out the expectations from the user. What I'm gathering is that this is taking a dataset in float and quantizing to bfloat16. 

Is there a benefit to doing the quantization inside the build process instead of the user quantizing up front and then passing the quantized dataset into build()? Is it because it's quantizing independently for each IVF list rather than quantizing the dataset as a whole?

rmaschal: It's quantizing the dataset as a whole, but the full precision dataset is used for all other steps of the build process, not the quantized dataset. So the user quantizing first and passing that in wouldn't make sense. 

I do the quant is this particular loop mostly for locality reasons. We have a batch on gpu for the other soar/quant steps, might as well do the bfloat16 too.

For regular bfloat16 quant (basically chop off the last 16 bits) it's not really important that we do it right here. In fact, the extra DtoH copy might be more expensive than doing regular bfloat16 right on cpu. I haven't tried yet. 

For bfloat16 w/ AVQ (introduced in this PR), we definitely want to use gpu, and I don't think it makes sense to introduce another HtoD copy of the dataset to do it elsewhere (e.g. outside of build)

cpp/tests/neighbors/ann_scann.cuh
```diff
@@ -227,7 +227,7 @@ inline auto big_dims_all_pq_bits() -> test_cases_t
 inline auto bf16() -> test_cases_t
 {
   scann_inputs ts;
-  ts.index_params.bf16_enabled = true;
+  ts.index_params.reordering_bf16_enabled = true;
```
cjnolet: Please rename according to my comment above.

cpp/include/cuvs/neighbors/scann.hpp
```diff
@@ -74,8 +75,12 @@ struct index_params : cuvs::neighbors::index_params {
...
 
+  /** Threshold for computing AVQ eta va Theorem 3.4 in https://arxiv.org/abs/1908.10396
+   * If the threshold is NAN, AVQ is not performed during bfloat16 quant
```
cjnolet, line 89: Can you please explain in more detail what this is so that a lay person can understand the attribute without following the link, please? The link is helpful but should not be required. Why is there a noise threshold? What is the AVQ eta?

cpp/include/cuvs/neighbors/scann.hpp
```diff
@@ -74,8 +75,12 @@ struct index_params : cuvs::neighbors::index_params {
...
+   */
+  float reordering_noise_shaping_threshold = NAN;
   // TODO - add other scann build params
```
cjnolet: Please create a Github issue for this and reference it here.As a general rule, we don't just drop todos in comments without having a corresponding Github issue for you or someone else to follow up.

cpp/src/neighbors/scann/detail/scann_build.cuh
cjnolet: > For bfloat16 w/ AVQ (introduced in this PR), we definitely want to use gpu, and I don't think it makes sense to introduce another HtoD copy of the dataset to do it elsewhere (e.g. outside of build)

The question here is more about reuse than it is about whether or not you should be doing this in the build. While I'm not particularly fond of the idea of maintaining copies in device memory, I think I can accept that more. What I'm concerned about is having different implementations of the same sorts of computations, transformations, and computational functions duplicated throughout the codebased. That's why I ask the question "Can this be it's own API or is it inherently coupled to SCaNN?" 

Just as a heads up, if we end up finding that it's more reusable, we're going to want to pull it out into its own thing, and if that means introducing an addtional copy or small perf hit as a result, that's a trade-off we are generally wiling to make in order to keep the codebase clean and maintainable. However, I'm fine keeping it where it is in the meantime if it's the only place in the codebase that needs to do it. Does that make sense?

cpp/src/neighbors/scann/detail/scann_quantize.cuh
```diff
@@ -267,6 +269,231 @@ void unpack_codes(raft::resources const& res,
...
+    bf16_dataset(row_idx, i) = float_to_bfloat16(dataset(row_idx, i));
+
+    float residual = dataset(row_idx, i) - bfloat16_to_float(bf16_dataset(row_idx, i));
+    residual_dot += dataset(row_idx, i) * residual * inv_norm;
+  }
+
+  // reduce  and broadcast residual_dot across warp
+  for (uint32_t offset = 16; offset > 0; offset >>= 1) {
+    residual_dot += raft::shfl_xor(residual_dot, offset, 32);
+  }
```
tfeher, line 368: Note that we have a [raft::warpReduce](https://github.com/rapidsai/raft/blob/bea333e76bf230bf1e4d94e23ab87ef6e89c1133/cpp/include/raft/util/reduction.cuh#L75) helper

The raft implementation unrolls the loop, I am not sure if that works for you in terms of register usage. Therefore just treat this as an FYI.

cpp/src/neighbors/scann/detail/scann_quantize.cuh
```diff
@@ -267,6 +269,231 @@ void unpack_codes(raft::resources const& res,
...
+
+  // 1 / ||x||
+  float inv_norm = 1 / sqrtf(sq_norms[row_idx]);
+  float eta      = compute_avq_eta(dim, sq_norms[row_idx], noise_shaping_threshold);
+
+  // < r, x >
+  float residual_dot = 0.0;
+
+  for (int i = lane_id; i < dim; i += 32) {
+    bf16_dataset(row_idx, i) = float_to_bfloat16(dataset(row_idx, i));
```
tfeher, line 359: Is the dataset padded to 32 dim? If yes please mention this in the docstring of the function. Otherwise please guard for out of bound access (here and other places as well.

tfeher (commented):
Hi Robert, thank you for the PR, I am looking through it here are my first two comments.

cpp/src/neighbors/scann/detail/scann_quantize.cuh
```diff
@@ -267,6 +269,231 @@ void unpack_codes(raft::resources const& res,
...
+  for (int round = 0; round < kMaxRounds && round_changes; round++) {
+    round_changes = false;
+
+    for (int i = lane_id; i < dim; i += 32) {
+      // coaleseced reads of required data
+      float original    = dataset(row_idx, i);
+      int16_t quantized = bf16_dataset(row_idx, i);
+
+      float old_residual    = original - bfloat16_to_float(quantized);
+      int16_t quantized_new = bfloat16_next_delta(old_residual, quantized);
```
tfeher, line 382: It is not clear to me what is the range of the input data values? Dataset is not normalized right? So it could have any range, including -1..1. Isn't it a problem that bfloat16_next_delta changes the values by ±1 only?

rmaschal: Although normalization does seem to be impactful overall, it's not expected to be normalized in anyway, so the range of possible values is all floats. 

The bfloat16_next_delta increments/decrements the encoded int16_t value. This is the same as incrementing/decrementing the mantissa, which could carry-over and impact the exponent similarly (in fact I should probably check it doesn't impact the sign-bit). In terms on the float repr it's not the same as just ±1

cpp/src/neighbors/scann/detail/scann_quantize.cuh
rmaschal: It is not padded. Out-of-bounds is guarded by the loop conditional for i, and at the start of the function for row_idx. The main guard is for the "sync" portion of this loop below this, where we limit the broadcast to only active threads.

cpp/src/neighbors/scann/detail/scann_quantize.cuh
rmaschal: Sounds nice, I'll look into it as a follow-up. IMU the compiler will unroll this loop itself.

cpp/src/neighbors/scann/detail/scann_build.cuh
rmaschal: Makes sense to me. In that case, your concern applies to several things I've added and will add to ScaNN (anything related to AVQ, SOAR, etc.). I've generally added these under ScaNN for my convenience, but in principle they could all be generalized. There is no inherit coupling with "ScaNN" (which is a library, not an algorithm). The coupling is only with the query distance function (should be dot product). 

The perf issue i've seen show up with other functions (the predict function for vpq comes to mind). I think this could be solved by providing a single batch device api, and a batched (host or device) api which calls the single batch api. I can use the first in my build func since I want control of the batching (e.g. for good locality) and the batch apis are more general/ergonomic. I feel like this out-of-scope for this PR, but this pattern makes sense I think.

cpp/src/neighbors/scann/detail/scann_quantize.cuh
tfeher: Ah, I missed that it is actually int16_t. Could you add this explanation to the docstring of bfloat16_next_delta?

cpp/include/cuvs/neighbors/scann.hpp
```diff
@@ -74,9 +75,20 @@ struct index_params : cuvs::neighbors::index_params {
   uint32_t pq_train_iters = 10;
 
   /** whether to apply bf16 quantization of dataset vectors **/
-  bool bf16_enabled = false;
-
-  // TODO - add other scann build params
+  bool reordering_bf16 = false;
```
bkarsin, line 78: I may be missing something - why is it called `reordering`? Does employing bf16 quant reorder the dimensions?

cpp/src/neighbors/scann/detail/scann_quantize.cuh
```diff
@@ -267,6 +269,233 @@ void unpack_codes(raft::resources const& res,
...
+ * representation, it is the mantissa that is being incremented/decremented,
+ * which could carryover to the exponent
+ *
+ * @param res the float residual
+ * @param current the current quantized dimension
+ * @return the other possible quantized value
+ */
+__device__ inline int16_t bfloat16_next_delta(float& res, int16_t& current)
+{
+  uint32_t res_sign  = ((int32_t)res & (1u << 31) >> 31);
```
bkarsin, line 328: Nitpick, but would this be simpler as just `(int32_t)res >= 0`?

cpp/src/neighbors/scann/detail/scann_quantize.cuh
```diff
@@ -267,6 +269,233 @@ void unpack_codes(raft::resources const& res,
...
+    float residual = dataset(row_idx, i) - bfloat16_to_float(bf16_dataset(row_idx, i));
+    residual_dot += dataset(row_idx, i) * residual * inv_norm;
+  }
+
+  // reduce  and broadcast residual_dot across warp
+  for (uint32_t offset = 16; offset > 0; offset >>= 1) {
+    residual_dot += raft::shfl_xor(residual_dot, offset, 32);
+  }
+
+  constexpr uint32_t kMaxRounds = 10;
```
bkarsin, line 370: There a reason for this max of 10 rounds? May be worth adding a little note of why

cpp/src/neighbors/scann/detail/scann_quantize.cuh
```diff
@@ -267,6 +269,233 @@ void unpack_codes(raft::resources const& res,
...
+
+  if (res_sign == curr_sign) { return current - 1; }
+
+  return current + 1;
+}
+
+template <uint32_t BlockSize, typename IdxT>
+__launch_bounds__(BlockSize) RAFT_KERNEL
+  quantize_bfloat16_noise_shaped_kernel(raft::device_matrix_view<const float, IdxT> dataset,
+                                        raft::device_matrix_view<int16_t, IdxT> bf16_dataset,
```
bkarsin, line 339: Why are we calling this `bf16` if it's an `int16` matrix? It seems you're casting it to `bf16` or `float` throughput, but wouldn't it still always fall in the range of `-2^15 to 2^15`? Sorry if I'm missing the main `bf16` benefit here.

cpp/include/cuvs/neighbors/scann.hpp
rmaschal: I'm trying to name params consistent with steps in the ScaNN pipeline. E.g. partitioning, pq, reordering. The "reordering" step is similar to itopk, we keep track of a larger list of candidates with which we perform an exact distance computation. In this case, the bfloat16 quantized dataset is used for the "exact" computation.

cpp/src/neighbors/scann/detail/scann_quantize.cuh
rmaschal: To be consistent with cpu algo.

cpp/src/neighbors/scann/detail/scann_quantize.cuh
rmaschal: It's convenient to use int16_t for the datatype in the index. OSS ScaNN expects an int16 matrix, so we can directly use the serialized result without any post processing. That's where bf16_dataset comes from, a view into that large int16_t matrix. With one exception, it doesn't really matter whether we use __nv_bfloat16 or int16_t. I use the former  for the float <-> __nv_bfloat16 conversion functions, but I just store the bits reinterpreted as int16_t for convenience.

The only place it matters is bfloat16_next_delta. There I'm using the IEEE representation of a bfloat16 together with arithmetic operations on int16_t to generate the next bfloat16 number that is larger or smaller than the given value (as explained above, I do this by incrementing/decrementing the mantissa, which is equivalent to inc/dec the int16_t representation). I'm not aware of a bfloat16 version of std::nextafter, which is similar to what blfoat16_next_delta does.

cpp/include/cuvs/neighbors/scann.hpp
bkarsin: So "reordering" is the step that ranks all candidates from a particular partition using the quantized distances?

cpp/src/neighbors/scann/detail/scann_quantize.cuh
bkarsin: I see, thanks for clarifying. Out of curiosity, would there be any perf advantage to this over using the original `int16_t` throughout (and casting it to `float` where needed)?

rmaschal: Hmm maybe, I'm not sure

cpp/include/cuvs/neighbors/scann.hpp
rmaschal: No, a configurable number of partitions/leafs will be searched, and the "itopk" closest populate an intermediate list which will be reranked using the exact (or bfloat16) distances

bkarsin: Gotcha, thanks for explaining

bkarsin (approved):
Thanks for answering my questions. PR looks good to me.

cpp/src/neighbors/scann/detail/scann_quantize.cuh
```diff
@@ -267,6 +269,233 @@ void unpack_codes(raft::resources const& res,
...
+    // reduce round_changes across warp
+    for (uint32_t offset = 16; offset > 0; offset >>= 1) {
+      round_changes |= raft::shfl_xor(round_changes, offset, 32);
+    }
+  }
+}
+
+/**
+ * @brief Quantized a float dataset as bfloat16, with noise shaping (AVQ)
+ *
```
tfeher, line 439: It would be useful to explain in the docstring what is actually being done. Here is my understanding:

```suggestion
* During quantization we replace each input vector coordinate `f` of type float32 with a bfloat16
* coordinate `b`. One way to do this would be to simply assign the nearest bfloat16 value to
* each coordinate. This would be the best way to quantize if we want to minimize the L2
* distance between the quantized and the original vector.
*
* In the AVQ method, we use a different cost function. To minimize that, we consider nearest 
* representable bfloat16 values (`b1`, `b2`) around `f`, and select the one that minimizes the AVQ
* cost function. In two dimensions we need to consider the four neighboring quantized vectors:
* b1       b2
*        f
* b3      b4
* 
* In N dimension we will select one the vertices of an N dimensional hypercube as the quantized
* vector. To find the minimum without enumerating all the combinations, a coordinate descent
* method is used.
```

tfeher (changes requested):
He @rmaschal the PR looks good overall. I have one suggestion to improve the documentition.

cpp/src/neighbors/scann/detail/scann_quantize.cuh
rmaschal: Yes I think this looks like a good summary, thank you :).

tfeher (approved):
Thanks @rmaschal for the updates! The PR looks good to me.

cpp/src/neighbors/scann/detail/scann_quantize.cuh
cjnolet: @rmaschal your explanation here for why we use an integral type is great. Can we add that to the code for future eyes please?

rmaschal: @cjnolet the part about relating arithmetic on int16_t to finding the next smaller/larger bfloat16 is in the description of bfloat16_next_delta(..). But I added the full reasoning behind using int16_t (OSS ScaNN expects it + this avq specific convo) as comment in the index def in scann.hpp

PR #1348, opened 2025-09-19
Bug/support installing minimal libcuvs c

robertmaynard:

PR #1357, opened 2025-09-23
Refactor libcuvs_c header and source locations

robertmaynard:
In support of building a stable C interface for libcuvs I have moved all `C` code into a separate directory for both the headers and sources. This will make it easier to install just 'C' language headers, make it easier to apply code rules for C headers, and in the future add more CI checks for our C API.

### DIFF
--- a/java/cuvs-java/src/main/java/com/nvidia/cuvs/CagraSearchParams.java
+++ b/java/cuvs-java/src/main/java/com/nvidia/cuvs/CagraSearchParams.java
@@ -83,7 +83,7 @@ public class CagraSearchParams {
     /**
      * AUTO_HASH
      */
-    AUTO_HASH(100);
+    AUTO_HASH(2);
 
     /**
      * The value for the enum choice.


robertmaynard:
By moving this into a separate header it will reduce the complexity of reviewing https://github.com/rapidsai/cuvs/pull/1348

cpp/CMakeLists.txt
```diff
@@ -717,30 +717,26 @@ target_compile_definitions(cuvs::cuvs INTERFACE $<$<BOOL:${CUVS_NVTX}>:NVTX_ENAB
...
+      $<$<BOOL:${BUILD_MG_ALGOS}>:src/c/neighbors/mg_ivf_flat.cpp>
+      $<$<BOOL:${BUILD_MG_ALGOS}>:src/c/neighbors/mg_cagra.cpp>
+      src/c/neighbors/nn_descent.cpp
+      src/c/neighbors/refine.cpp
+      src/c/neighbors/tiered_index.cpp
+      src/c/neighbors/all_neighbors.cpp
+      src/c/preprocessing/quantize/binary.cpp
+      src/c/preprocessing/quantize/scalar.cpp
+      src/c/distance/pairwise_distance.cpp
       ${CUVS_C_MG_ALGOS}
```
KyleFromNVIDIA: Please remove this line as it is no longer needed.

```suggestion
```

cpp/tests/CMakeLists.txt
```diff
@@ -329,7 +329,7 @@ if(TARGET cuvs::c_api)
   endif()
 
   add_executable(cuvs_c_test core/c_api.c)
-  target_link_libraries(cuvs_c_test PUBLIC cuvs::c_api)
+  target_link_libraries(cuvs_c_test PUBLIC cuvs::c_api CUDA::cudart)
```
KyleFromNVIDIA: Why has this suddenly been added when it wasn't needed before?

robertmaynard: Missed this was part of my cherry-pick from #1348, will remove

robertmaynard: Removed

benfred:
Rather than put in the c headers in`cpp/include/cuvs/c/`  - how about we create a top level `c` directory like we have for `cpp`, `go`, `rust`, `python`, and put the C headers and source in there rather than in the `cpp/` directory? (like `c/include/cuvs/neighbors/cagra.h` instead of `cpp/include/cuvs/c/neighbors/cagra.h` )

Right now the user has to add an extra `/c/` to the path when including the header files - and I'd rather avoid changing what the include path looks like .

robertmaynard:
> Rather than put in the c headers in`cpp/include/cuvs/c/` - how about we create a top level `c` directory like we have for `cpp`, `go`, `rust`, `python`, and put the C headers and source in there rather than in the `cpp/` directory? (like `c/include/cuvs/neighbors/cagra.h` instead of `cpp/include/cuvs/c/neighbors/cagra.h` )
> 
> Right now the user has to add an extra `/c/` to the path when including the header files - and I'd rather avoid changing what the include path looks like .

This opens us to incorrect installs of libcuvs or libcuvs_c since when we install we flatten everything to the same install location of `include/cuvs/`. That means we would have to either manually validate that we don't introduce any `.h` files under `/cpp/include` or use some of the tooling from package managers like conda to warn us.

The current model of the source tree layout is that the `cpp`, `rust`, `go` are all roots for build invcocations. If we elevate `c` to that level ( happy to do so ) we should also seriously consider making it a seperate build instead of a component of the `cpp` build.

robertmaynard:
I am happy to try moving everything over to `cuvs/c/src` and `cuvs/c/include` while we discuss if we like that layout / side-effects

Issue #1358, opened 2025-09-23
[BUG] KMeans Multi-GPU Race Condition in k-means++ Initialization


csadorf:
**Describe the bug**
A race condition in the NCCL broadcast operation during k-means++ initialization causes intermittent hanging in multi-GPU KMeans operations. Different ranks calculate different root rank values (`rp`) due to inconsistent random number generator states, leading to deadlocks in collective communication operations.

**Steps/Code to reproduce bug**
The issue can be reproduced using the minimal test case provided in [cuML Issue #7245](https://github.com/rapidsai/cuml/issues/7245). The test script exhibits intermittent hanging behavior when running on multiple GPUs, typically hanging after approximately 4 attempts.

**Expected behavior**
The script should consistently complete in under 10 seconds for all runs.

**Actual behavior**
- **Normal runs**: Complete in ~7-8 seconds
- **Hanging runs**: Hangs indefinitely with no progress
- **Frequency**: Hangs occur approximately every 4th run when using 2 GPUs

**Root Cause**
The issue occurs in `initKMeansPlusPlus` function in `cpp/src/cluster/detail/kmeans_mg.cuh` (lines 178-180) where each rank independently calculates the root rank:

```cpp
std::mt19937 gen(params.rng_state.seed);
std::uniform_int_distribution<> dis(0, n_rank - 1);
int rp = dis(gen);
```

With no seed set, there's no guarantee that the `std::mt19937` generators are in identical states across ranks. If ranks calculate different `rp` values, the subsequent NCCL broadcast becomes inconsistent - some ranks expect to broadcast data while others expect to receive it, causing a deadlock. The actual hang occurs on [line 207](https://github.com/rapidsai/cuvs/blob/8fc4a2ad134f2d2d7763b026d0965b328ecc8db8/cpp/src/cluster/detail/kmeans_mg.cuh#L207) during the broadcast operation.

**Workaround**
The issue can be avoided by restricting the script to use only a single GPU, e.g., by setting `CUDA_VISIBLE_DEVICES=0`.

**Environment details (please complete the following information):**

**Additional context**
This bug has likely been present for a long time, but has only now become more observable due to the recent change in cuML's KMeans defaults from `random_state=1` to `random_state=None` ([cuML PR #6884](https://github.com/rapidsai/cuml/pull/6884)). With the fixed seed, the random number generator states were more likely to be synchronized across ranks, masking the race condition.

[...]

PR #1359, opened 2025-09-23
Fix k-means++ root rank race condition

csadorf:
Fixes https://github.com/rapidsai/cuvs/issues/1358


cpp/src/cluster/detail/kmeans_mg.cuh
```diff
@@ -175,9 +175,21 @@ void initKMeansPlusPlus(const raft::resources& handle,
...
+    std::uniform_int_distribution<> dis(0, n_rank - 1);
+    rp_host = dis(gen);
+  }
+  {
+    rmm::device_scalar<int> rp_d(stream);
+    raft::copy(rp_d.data(), &rp_host, 1, stream);
+    comm.bcast<int>(rp_d.data(), 1, /*root=*/KMEANS_COMM_ROOT, stream);
+    raft::copy(&rp, rp_d.data(), 1, stream);
+    raft::resource::sync_stream(handle, stream);
+  }
```
divyegala: Can we simplify this a bit?
```suggestion
  int rp      = 0;
  if (my_rank == KMEANS_COMM_ROOT) {
    std::mt19937 gen(params.rng_state.seed);
    std::uniform_int_distribution<> dis(0, n_rank - 1);
    rp = dis(gen);
  }
  {
    rmm::device_scalar<int> rp_d(stream);
    raft::copy(rp_d.data(), &rp, 1, stream);
    comm.bcast<int>(rp_d.data(), 1, /*root=*/KMEANS_COMM_ROOT, stream);
    raft::copy(&rp, rp_d.data(), 1, stream);
    raft::resource::sync_stream(handle);
  }
```

csadorf: How would you propose to simplify? To my (limited) understanding, rp needs to be on device for communication, no?

divyegala: Yes, I kept `rp_d` in my suggestion for communication. I just merged the variables `rp` and `rp_host`.

PR #1361, opened 2025-09-23
[Review][Java] Refactor: extract interface from CuVSMatrixBaseImpl

ldematte:
This PR extract an internal interface that is used as a base to implement all internal CuVSMatrix types; the interface introduces commonly used field accessors like `memorySegment()` and `toTensor()` that we do not want/could not appear on the public interface (e.g. because they expose or require Panama types or internal types).

The new interface is implemented by all concrete matrix types, closing a gap that we had in https://github.com/rapidsai/cuvs/pull/1328 (which I realized while working on separate PRs like https://github.com/rapidsai/cuvs/pull/1283)

Follow-up of https://github.com/rapidsai/cuvs/pull/1328


java/cuvs-java/src/main/java/com/nvidia/cuvs/CuVSDeviceMatrix.java
```diff
@@ -30,67 +30,4 @@ default CuVSHostMatrix toHost() {
     toHost(hostMatrix);
     return hostMatrix;
   }
-
-  default CuVSDeviceMatrix toDevice(CuVSResources resources) {
-    return new CuVSDeviceMatrixDelegate(this);
-  }
-
-  class CuVSDeviceMatrixDelegate implements CuVSDeviceMatrix {
```
ldematte, line 38: These got moved to internal implementation classes, for better encapsulation and to use the internal interface type.

java/cuvs-java/src/main/java/com/nvidia/cuvs/CuVSHostMatrix.java
```diff
@@ -21,76 +21,9 @@
...
-    return new CuVSHostMatrixDelegate(this);
-  }
-
   default CuVSDeviceMatrix toDevice(CuVSResources resources) {
     var deviceMatrix = CuVSMatrix.deviceBuilder(resources, size(), columns(), dataType()).build();
     toDevice(deviceMatrix, resources);
     return deviceMatrix;
   }
-
-  class CuVSHostMatrixDelegate implements CuVSHostMatrix {
```
ldematte, line 34: Same as above

PR #1366, opened 2025-09-24
[Review][Java] Test indexing and serialization with integral (byte) dataset

ldematte:
Some additional tests to increase our test coverage.

Good to review, but we might want to expand this after https://github.com/rapidsai/cuvs/pull/1283 (or vice-versa, whichever comes first).

### DIFF
--- a/java/cuvs-java/src/test/java/com/nvidia/cuvs/CagraBuildAndSearchIT.java
+++ b/java/cuvs-java/src/test/java/com/nvidia/cuvs/CagraBuildAndSearchIT.java
@@ -16,6 +16,7 @@
 package com.nvidia.cuvs;
 
 import static com.carrotsearch.randomizedtesting.RandomizedTest.assumeTrue;
+import static com.carrotsearch.randomizedtesting.RandomizedTest.randomIntBetween;
 import static org.junit.Assert.*;
 
 import com.carrotsearch.randomizedtesting.RandomizedRunner;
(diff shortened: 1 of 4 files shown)


mythrocks:
Looks like this is going to need a rebase.

@cjnolet will keep me honest:  Given the code-freeze, maybe we should retarget this to 25.12.

ldematte:
@mythrocks @cjnolet I can do that, but at the moment it seems that 25.12 is _behind_ 25.10, so you probably want to incorporate all changes from 25.10 before I rebase, right?

java/cuvs-java/src/test/java/com/nvidia/cuvs/CagraBuildAndSearchIT.java
```diff
@@ -133,18 +134,19 @@ public void testIndexingAndSearchingFlow() throws Throwable {
...
+          var indexPath = serializeOnce(index);
+          try (var loadedIndex = deserializeOnce(indexPath, resources)) {
+            queryAndCompare(
+                index,
+                loadedIndex,
+                SearchResults.IDENTITY_MAPPING,
+                queries,
+                expectedResults,
+                resources);
+            Files.deleteIfExists(indexPath);
```
mythrocks, line 147: 👍

mythrocks (approved):
This looks good to me.

I've removed a bunch of aliases from the reviewer-list.  They might have been added as an artifact of rebasing this PR.

I'll merge this once the CI passes.

ldematte:
Hey @mythrocks , test failure says:
```
Caused by: java.lang.AssertionError: Exception while executing runnable: java.lang.RuntimeException: cuvsCagraBuild returned 0[RAFT failure at file=/tmp/conda-bld-output/bld/rattler-build_libcuvs/work/cpp/src/neighbors/detail/cagra/graph_core.cuh line=1417: Could not generate an intermediate CAGRA graph because the initial kNN graph contains too many invalid or duplicated neighbor nodes. This error can occur, for example, if too many overflows occur during the norm computation between the dataset vectors.
```
This is due to the fact that we generate data randomly. I wonder if there is a way to avoid this? The error mentions too many overflows; maybe limit the range of the generated floats? Keep them between a given min and max?

mythrocks:
> The error mentions too many overflows; maybe limit the range of the generated floats? Keep them between a given min and max?

Yes, I think so.  This test seems flaky.

I'm re-kicking this PR after updating.

ldematte:
>Yes, I think so. This test seems flaky.

Any suggestion on a sensible range to use as input? E.g. keep them between 0 and 1.0 or ...?

mythrocks:
Let's pick up the flaky test in another issue.

PR #1375, opened 2025-09-25
[Review][C] Export the ability to get/set the log level to the C API

### LINKED ISSUE #1310: [BUG] Non-controllable logging from libcuvs/raft
Today cuvs (and raft underneath) do not offer any way to control log output and level in a programmatic way, and the generated logs can be massive and interfere with the application usage of stdout/strerr.

The only workaround ATM is to `export RAFT_DEBUG_LOG_FILE=/dev/null`

We should offer a programmatic way to control the log level (Possibly related: https://github.com/rapidsai/cuvs/issues/576) or, as a short term solution, turn off logging by default.

ldematte:
CuVS uses raft underneath, which in turns uses `rapids_logger`. Users of the C++ API can access directly the logger used by raft via `raft::default_logger()`, and use it to get and set the log level. 
This is not possible for users of the C API (and of any API based on it, like the C API).
This PR adds specific C functions to close this gap.

Closes #1310

### DIFF
--- a/cpp/include/cuvs/core/c_api.h
+++ b/cpp/include/cuvs/core/c_api.h
@@ -20,6 +20,7 @@
 #include <dlpack/dlpack.h>
 #include <stdbool.h>
 #include <stdint.h>
+#include <rapids_logger/log_levels.h>
 
 #ifdef __cplusplus
 extern "C" {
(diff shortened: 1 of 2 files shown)


c/include/cuvs/core/c_api.h
```diff
@@ -48,6 +49,35 @@ void cuvsSetLastErrorText(const char* error);
 
 /** @} */
 
+/**
+ * @defgroup log_c cuVS Logging
```
benfred, line 53: Can you add an entry for this to the C docs in `docs/source/c_api/core_c_api.rst` ? (following the same pattern as the error handling / `error_c` section should be sufficient)

ChrisHegarty:
Do we need this in 25.10? How do we control the log level without it?

mythrocks:
@benfred, @cjnolet:  Should this change be on `branch-25.12`?  Or is it safe to go into the current release?

I would have thought the former, given that the ability to change log levels hasn't been available in previous releases either. 

But I'm Johnny-come-lately.

cjnolet:
Unfortunately now that the code is frozen we have to push it forward, so it'll need to be in 25.12.

mythrocks:
@cjnolet, thank you for confirming. 

@ldematte:  I'm 👍 on this change too.  Let's retarget to `branch-25.12`.

ldematte:
@mythrocks @benfred I have rebased this onto branch-25.12
I think it's ready to be merged.

mythrocks:
Pardon me, but I don't think the changes are complete:
```
jextract downloaded to /__w/cuvs/cuvs/java/jextract-22
c_api.h:21:10: error: 'rapids_logger/log_levels.h' file not found
__stddef_max_align_t.h:22:15: warning: Skipping __clang_max_align_nonce2 (type LongDouble is not supported)
```
We'll need to change the Java build script, and change the `jextract` line to include the `rapids_logger` header directory.

ldematte:
I have that change in https://github.com/rapidsai/cuvs/pull/1376, along with the Java changes, but it seems the way we use jextract breaks. I'll move the change to the scripts to this PR.

ldematte:
@mythrocks In general, I find a bit odd that when you change the C api you need to update something related to the Java API. I understand the mechanics of it (jextract process a C header which includes headers from the C API, so in turn changing a C API header may have effects), but it's counterintuitive. Imagine if I was a developer purely focusing on the C side of things, with no knowledge of jextract etc.

I think we need a more robust fix for this; I think jextract should use the same include directories as the C compiler. No idea how to achieve that though, ATM.

mythrocks:
> I think jextract should use the same include directories as the C compiler.

I agree in principle.  But given that these are disparate build systems, I'm not sure how we'd unify these. :/

PR #1376, opened 2025-09-25
[REVIEW][Java] Log level api

ldematte:
Java API for getting/setting the CuVS/raft logging

Depends on https://github.com/rapidsai/cuvs/pull/1375

### DIFF
--- a/java/cuvs-java/src/main/java/com/nvidia/cuvs/spi/CuVSProvider.java
+++ b/java/cuvs-java/src/main/java/com/nvidia/cuvs/spi/CuVSProvider.java
@@ -1,17 +1,6 @@
 /*
- * Copyright (c) 2025, NVIDIA CORPORATION.
- *
- * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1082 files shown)


cjnolet (approved):

java/cuvs-java/src/main/java/com/nvidia/cuvs/spi/CuVSProvider.java
```diff
@@ -143,6 +143,10 @@ default CagraIndex mergeCagraIndexes(CagraIndex[] indexes, CagraMergeParams merg
   /** Returns a {@link GPUInfoProvider} to query the system for GPU related information */
   GPUInfoProvider gpuInfoProvider();
 
+  void setLogLevel(java.util.logging.Level level);
+
+  java.util.logging.Level getLogLevel();
```
mythrocks, line 148: Trivial nit:  Should we consider an import here, for consistency with `UnsupportedProvider` and `JDKProvider`?

PR #1377, opened 2025-09-25
Regroup overloads in docs

lowener:
This PR introduce a custom js script to regroup overloads into a single link on the right side navigation bar.
Another alternative would be to have the brief of every function in there instead of grouping them, so that it could look like
```
- fit(float, int)
- fit(double, int)
- fit(float, int64)
- ...
```
Before:
grafik

After:
grafik

### DIFF
--- a/docs/source/cpp_api/cluster_agglomerative.rst
+++ b/docs/source/cpp_api/cluster_agglomerative.rst
@@ -1,5 +1,5 @@
 Agglomerative
-=======
+=============
 
 .. role:: py(code)
    :language: c++


docs/source/cpp_api/cluster_agglomerative.rst
```diff
@@ -1,4 +1,4 @@
-Cluster
+Agglomerative
 =======
```
benfred:
```suggestion
Agglomerative
=============
```

benfred (approved):

PR #1391, opened 2025-09-29
Add the example for building the CARGA index in a streaming fashion

### LINKED ISSUE #1146: [FEA] Establish streaming incremental micro-batched index build for CAGRA

yan-zaretskiy:
This example shows how to build a CAGRA graph index by streaming host batches:
1) Stage the first few batches to train an IVF-PQ index.
2) Incrementally extend the IVF-PQ index with every batch.
3) Run IVF-PQ search over the full dataset to form an intermediate k-NN graph.
4) Optimize that graph into the final fixed-degree CAGRA graph.

Closes #1146

### DIFF
--- a/examples/cpp/src/cagra_streaming_example.cu
+++ b/examples/cpp/src/cagra_streaming_example.cu
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2024-2025, NVIDIA CORPORATION.
+ * Copyright (c) 2025, NVIDIA CORPORATION.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -17,22 +17,14 @@
 #include <algorithm>
 #include <cstdint>
 #include <iostream>
-#include <random>
-#include <stdexcept>
 
-#include <thrust/sequence.h>
+#include "cagra_streaming_example.cuh"
+#include "common.cuh"
 
 #include <cuvs/neighbors/cagra.hpp>
 #include <cuvs/neighbors/cagra_optimize.hpp>
 #include <cuvs/neighbors/ivf_pq.hpp>
-#include <raft/core/device_mdarray.hpp>
-#include <raft/core/device_mdspan.hpp>
 #include <raft/core/device_resources.hpp>
-#include <raft/core/host_mdarray.hpp>
-#include <raft/core/host_mdspan.hpp>
-#include <raft/core/mdspan.hpp>
-#include <raft/core/resource/cuda_stream.hpp>
-#include <raft/core/resource/thrust_policy.hpp>
 
 #include <rmm/mr/device/device_memory_resource.hpp>
 #include <rmm/mr/device/pool_memory_resource.hpp>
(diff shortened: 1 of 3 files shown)


examples/cpp/src/cagra_streaming_example.cu
```diff
@@ -0,0 +1,310 @@
+/*
+ * Copyright (c) 2024-2025, NVIDIA CORPORATION.
```
cjnolet: Since this is a new file, we should only use the current year

examples/cpp/src/cagra_streaming_example.cu
```diff
@@ -0,0 +1,310 @@
...
+3) Run IVF-PQ search over the full dataset to form an intermediate k-NN graph.
+4) Optimize that graph into the final fixed-degree CAGRA graph.
+
+The dataset is kept on the host the whole time to mimic a scenario where the
+full corpus cannot live on the device.
+*/
+
+namespace {
+
+void make_host_dataset(raft::host_matrix_view<float, int64_t, raft::row_major> dataset)
```
cjnolet: Can we maybe call this "generate host dataset"? Honestly, I think I'd prefer to use `make_blobs` from `raft` than to just generate completely random uniform vectors. make_blobs at least has some locality to the vector space that can be exploited by IVFPQ and CAGRA. It would also look a little cleaner to have just a simple function call to "make_blobs" and then a copy to host.

examples/cpp/src/cagra_streaming_example.cu
```diff
@@ -0,0 +1,310 @@
...
+  raft::device_matrix<float, int64_t> kmeans_staging_buffer;
+  raft::device_matrix<float, int64_t> batch_device_buffer;
+  raft::device_vector<int64_t, int64_t> batch_id_buffer;
+
+  std::vector<HostBatch> batches;
+};
+
+}  // namespace
+
+void streaming_cagra_build_example(
```
cjnolet, line 46: Can we pull everything above this point into a corresponding `cagra_streaming_example.hpp` header for readability? Ideally the user would be able to follow through the meat of the example first, and then refer to the header for all the implementation details. 

The other benefit to this approach is that they essentially can drop the header into their own project and ideally just copy/paste the relevant blocks from the source file into their own applications.

yan-zaretskiy: Sure thing. I was trying to follow the existing examples as closely as possible, but I like your idea.

cjnolet:
Just making a note here publicly- we aren't seeing a great speedup in this example, partly because the building of the ivfpq index itself isn't the bottleneck. We're currently investigating some other methods for incrementally building a cagra graph with massive paralleilsm to be able to offset and overlap computation with data transfer. 

Tagging @tfeher and @anaruse for awareness.

cjnolet:
@yan-zaretskiy given that we have determined this wasn't offsetting enough of the bottleneck to hide the transfer latencies, I think we should close this one and open a new one when we have a good solution. 

But- as always, please feel free to reopen if you disagree.

PR #1329, opened 2025-09-13
Update to CCCL 3.1.x.

bdice:
This PR updates to CCCL 3.1.x. Do not merge until all of RAPIDS is ready to update.

Part of https://github.com/rapidsai/build-planning/issues/217.

PR #1398, opened 2025-09-30
Patch FAISS for missing Thrust includes

bdice:
This PR patches FAISS for some missing Thrust includes. This is needed to support building with CCCL 3.1.x. RAPIDS 25.10 will still use CCCL 3.0.x but we want to make it possible to build with 3.1.x in 25.10.

xref: #1329 (This backports a fix from that testing PR, which is targeting 25.12)

I am upstreaming the patch here: https://github.com/facebookresearch/faiss/pull/4597

### DIFF
--- a/docs/source/c_api/neighbors.rst
+++ b/docs/source/c_api/neighbors.rst
@@ -16,3 +16,4 @@ Nearest Neighbors
    neighbors_hnsw_c.rst
    neighbors_mg.rst
    neighbors_all_neighbors_c.rst
+   neighbors_vamana_c.rst


cpp/cmake/patches/faiss_override.json
```diff
@@ -1,9 +1,16 @@
...
-      }
+  "packages" : {
+    "faiss" : {
+      "version": "1.12.0",
+      "git_url": "https://github.com/facebookresearch/faiss.git",
+      "git_tag": "v1.12.0",
+      "patches" : [
+        {
+          "file" : "${current_json_dir}/faiss/fix-missing-includes-for-thrust-copy.patch",
+          "issue" : "Fix missing Thrust includes",
```
bdice:
```suggestion
          "issue" : "Fix missing Thrust includes. https://github.com/facebookresearch/faiss/pull/4597",
```

tarang-jain (approved):
Thanks! We should revert this once https://github.com/facebookresearch/faiss/pull/4597 is merged. However we still do pin Faiss to 1.12.0, so this PR needs to be merged until 1.13.0 is released (trying to avoid building nightly Faiss in our CI)

divyegala (approved):

PR #1414, opened 2025-10-08
[DO NOT MERGE] Test cuVS against raft #2835

achirkin:
Testing if cuVS uses copy constructors on any mdarrays  https://github.com/rapidsai/raft/pull/2835


cpp/cmake/thirdparty/get_raft.cmake
```diff
@@ -55,8 +55,8 @@ endfunction()
 # To use a different RAFT locally, set the CMake variable
 # CPM_raft_SOURCE=/path/to/local/raft
 find_and_configure_raft(VERSION  ${RAFT_VERSION}.00
-        FORK                     ${RAFT_FORK}
-        PINNED_TAG               ${RAFT_PINNED_TAG}
+        FORK                     achirkin
+        PINNED_TAG               enh-consistent-container-policy
```
tfeher, line 59: Don't forget to revert.

achirkin:
> Changes look good on the host transform function. Is the gpu transform function correct?

Apparently so, because the grid goes over the `dataset_dim` rather than over `minimul_out_dim` and hence was the failing test when the host- and device- produced outputs didn't match.
Actually I've been wondering if it would be more appropriate to not touch the extra elements and modify the tests instead (and the gpu version)?..

Issue #1370, opened 2025-09-24
[BUG] cuVS Nearest Neighbors recall lower than expected for some datasets


jinsolp:
**Describe the bug**
cuVS brute force does not 100% match python-computed brute force results.
cuVS NN Descent seems to have low recall for some datasets (especially for smaller dimensions)

**Steps/Code to reproduce bug**
```
from pynndescent import NNDescent
from sklearn import datasets
from sklearn.neighbors import NearestNeighbors
import numpy as np
from cuvs.neighbors import nn_descent, brute_force
import cupy as cp

def knn_recall(pred: np.ndarray, gt: np.ndarray) -> float:
    N, k = pred.shape
    correct = 0

    for i in range(N):
        pred_set = set(pred[i])
        gt_set = set(gt[i])
        correct += len(pred_set & gt_set)

    return correct / (N * k)

def run_exp(X, data_tag, k=16):
    # CPU Brute Force
    nn = NearestNeighbors(n_neighbors=k, algorithm="brute", metric="euclidean")
    nn.fit(X)
    cpu_bf_distances, cpu_bf_indices = nn.kneighbors(X)

    # CPU NN Descent
    index = NNDescent(X, metric="euclidean")
    cpu_nnd_indices, cpu_nnd_distances = index.query(X, k=k)

    # GPU Brute Force
    X_cp = cp.asarray(X)
    index = brute_force.build(X_cp, metric="euclidean")
    gpu_bf_distances, gpu_bf_indices = brute_force.search(index, X_cp, k)
    gpu_bf_indices = gpu_bf_indices.copy_to_host()
    gpu_bf_distances = gpu_bf_distances.copy_to_host()

    # GPU NN Descent
    build_params = nn_descent.IndexParams(metric="euclidean", return_distances=True, graph_degree=k)
    index = nn_descent.build(build_params, X)
    gpu_nnd_indices = index.graph
    gpu_nnd_distances = index.distances

    print(f"cpu bf vs cpu nnd on {data_tag}: {knn_recall(cpu_bf_indices, cpu_nnd_indices)}")
    print(f"cpu bf vs gpu nnd on {data_tag}: {knn_recall(cpu_bf_indices, gpu_nnd_indices)}")
    print(f"cpu bf vs gpu bf on {data_tag}: {knn_recall(cpu_bf_indices, gpu_bf_indices)}\n")


X, y = datasets.make_blobs(n_samples=10_000, n_features=16, centers=5, random_state=42)
[...]

Issue #195, opened 2024-06-18
[FEA] Calculating distances with FP32 in NN Descent 


jinsolp:
I hope NN Descent can support calculating distances in `local_join_kernel` using fp32 instead of fp16 (`__half`).

Due to precision issues, it is difficult to grab the distances calculated by NN Descent and compare it against fp32-calculated distances for additional operations outside of NN Descent such as making predictions.

jinsolp:
Summarizing details after investigating and comparing this to our brute force implementation.

TLDR;
The problem with our NN Descent implementation is that given FP32 data, we always convert it to FP16 to run wmma.
This causes problems when compared against the brute force implementation, which directly uses the FP32 data.
To solve this, we need to keep the FP32 data as FP32.

This writeup demonstrates this by showing
1. NN Descent vs Brute force show similar results when given fp16 data
2. NN Decent vs Brute force shows similar behavior when we modifying our Brute force implementation to do a fp32->fp16 conversion like NN Descent  


# FP16 data
Given fp16 data, our NN Descent calculations are done exactly the same as the brute force implementation
- Calculate norms in fp32 by converting fp16 to fp32 on the fly
- Calculate dot product in fp32 by converting fp16 to fp32 on the fly, accumulator is fp32.

```
from sklearn.datasets import make_moons
from cuvs.neighbors import brute_force, nn_descent
import numpy as np

# Data in fp16
X, y = make_moons(n_samples=1_000_000, noise=0.1, random_state=42)
X = X.astype(np.float16)
k = 16
metric="l2"

# NN Descent
build_params = nn_descent.IndexParams(metric=metric, return_distances=True, graph_degree=k)
index = nn_descent.build(build_params, X)
gpu_nnd_indices = index.graph
gpu_nnd_distances = index.distances

# Brute Force
cp_X = cp.asarray(X)
index = brute_force.build(cp_X, metric=metric)
gpu_bf_distances, gpu_bf_indices = brute_force.search(index, cp_X, k)

# Add self-loops: discard last column, shift right, and add row index as first column
def add_self_loops(indices):
    n_rows = indices.shape[0]
    # Create array with self-loops in first column
    self_loops = np.arange(n_rows).reshape(-1, 1)
    # Drop last column and concatenate with self-loops at the beginning
    return np.concatenate([self_loops, indices[:, :-1]], axis=1)

def knn_recall(pred: np.ndarray, gt: np.ndarray) -> float:
    # print(pred[0], gt[0])
[...]

PR #1415, opened 2025-10-08
Dispatch to use fp32 distance computation in NN Descent depending on data dimensions

jinsolp:
Closes https://github.com/rapidsai/cuvs/issues/1370
Closes https://github.com/rapidsai/cuvs/issues/195

This PR adds an option to use fp32 distance computation.

~~(Outdated) From heuristics, chose dim=16 as the threshold for dispatching to a fp32 distance kernel.
We do manual computation, but since we only target small dimensions, fp32 dispatching ends up being slightly faster end to end with much better recall for small dimensions.~~

All number below are run on L40 machine and AMD EPYC CPU with 128 cores. Perf and recall is averaged over 5 runs and all time is in seconds. Baseline knn graph is computed using `sklearn.neighbors.NearestNeighbors` brute for method.

### Max iters=20
Screenshot 2025-10-08 at 10 56 17 AM

For larger dimensions there is an inherent issue with the NN Descent algorithm itself that makes the recall low. This can be improved slightly with more iterations.
Also notice that the e2e time taken is similar or slightly less for using fp32.

### Max iters=100
Screenshot 2025-10-08 at 10 58 26 AM

Notice how the blue part, the recall doesn't get better compared to the table above even with more iterations (i.e. why we need the fp32 appraoch for this part)

# Perf impact on different architectures

### H100
Screenshot 2025-11-20 at 10 15 25 AM

### L40
Screenshot 2025-11-20 at 10 15 50 AM

### DIFF
--- a/ci/build_standalone_c.sh
+++ b/ci/build_standalone_c.sh
@@ -5,6 +5,7 @@
 set -euo pipefail
 
 TOOLSET_VERSION=14
+NINJA_VERSION=v1.13.1
 
 BUILD_C_LIB_TESTS="OFF"
 if [[ "${1:-}" == "--build-tests" ]]; then
@@ -35,6 +36,7 @@ if ! command -V ninja >/dev/null 2>&1; then
     rm /tmp/ninja-linux.zip
 fi
 
+source rapids-install-sccache
 source rapids-configure-sccache
 source rapids-date-string
 


jinsolp:
Maybe we might need to have fp16 vs fp32 as an option (and default to fp16) instead of depending on data dimensions. WIP investigating

cpp/src/neighbors/detail/nn_descent.cuh
```diff
@@ -476,6 +478,326 @@ __device__ __forceinline__ void remove_duplicates(
   }
 }
 
+template <typename Index_t, typename Data_t, typename DistEpilogue_t>
+__device__ __forceinline__ void calculate_metric(float* s_distances,
```
jinsolp, line 486: This is taken out as a separate function from the original kernel because it's commonly used for fp32 and fp16

cpp/src/neighbors/detail/nn_descent.cuh
```diff
@@ -476,6 +478,326 @@ __device__ __forceinline__ void remove_duplicates(
...
+        int tmp_row = i / SKEWED_MAX_NUM_BI_SAMPLES;
+        int tmp_col = i % SKEWED_MAX_NUM_BI_SAMPLES;
+        if (tmp_row < list_new_size && tmp_col < list_new_size) {
+          float acc = 0.0f;
+          for (int d = 0; d < num_load_elems; d++) {
+            acc += s_nv[tmp_row][d] * s_nv[tmp_col][d];
+          }
+          s_distances[i] += acc;
+        }
+      }
```
jinsolp, line 676: This matmul part is different from the fp16 kernel

jinsolp:
Thanks for the feedback @divyegala, changing this to target 26.02 for now because Corey suggested we do further investigation. Force-pushing after rebasing

cpp/src/neighbors/detail/nn_descent.cuh
```diff
@@ -1039,6 +1316,20 @@ GNND<Data_t, Index_t>::GNND(raft::resources const& res, const BuildConfig& build
 {
   static_assert(NUM_SAMPLES <= 32);
 
+  using input_t = typename std::remove_const<Data_t>::type;
+  if (std::is_same_v<input_t, float> && build_config.dataset_dim <= 16) {
```
jinsolp: dispatch logic based on dimensions

jinsolp:
Blue is better  (e2e time or kernel time)
- FP32 kernel is usually faster for small dimensions
- FP32 e2e time is sometimes slightly longer for small dimensions despite shorter kernel time. Believe this small gap comes from more data transfers (4bytes vs 2bytes per data element)

## L40 GPU
Screenshot 2025-12-02 at 6 28 48 PM
Screenshot 2025-12-02 at 6 28 58 PM

## H100 GPU
Screenshot 2025-12-02 at 6 30 47 PM
Screenshot 2025-12-02 at 6 30 59 PM

jinsolp:
@divyegala based on the latest benchmarks (and confirming from HDBSCAN that we don't have quality issues from using fp16 for larger dimensions), the mechanism is back to dispatching based on the number of dimensions 
(chose 16 as a threshold).

> I would like to see a unit test, if possible, where you can demonstrably prove distance computation improvements in fp32 over fp16.

Not sure how to add this as a test. Any suggestions?

divyegala:
> Not sure how to add this as a test. Any suggestions?

@jinsolp is it possible to generate a dataset that shows degraded recall for fp16 and good recall for fp32? The way I envision the assertion is `ASSERT(fp16_recall < <upperbound_value> && fp32_recall >= <lowerbound_value>)`.

jinsolp:
We can (e.g. the blobs data with dim=2 which shows 0.4 recall for fp16 vs 0.99 recall for fp32)
, but if we run through the current code, it will always take the fp32 path.

Should I hardwire the results of fp16 and add that as a vector to the test?

divyegala:
@jinsolp we should allow the user an override in-case the heuristic fails. Let users choose fp32 explicitly if they want, and exporting that option will also let you test the separate paths.

jinsolp:
Oh okay, so let the user choose an option, and if none is given fall back to the heuristic dim=16 threshold?

divyegala:
The other way around. Use the heuristic eagerly, but let user override.

jinsolp:
@divyegala we now have a `dist_comp_dtype` option defaulting to `auto`. can override by using `fp32` or `fp16`.
Added a test as well.

c/src/neighbors/nn_descent.cpp
divyegala, line 1: Please default initialize the index creation with this new parameter.

cpp/include/cuvs/neighbors/nn_descent.hpp
```diff
@@ -37,13 +47,19 @@ namespace cuvs::neighbors::nn_descent {
...
+ * valid when data type is fp32. Use `FP16` for better performance and memory usage at the cost of
+ * precision.
  */
 struct index_params : cuvs::neighbors::index_params {
   size_t graph_degree              = 64;
   size_t intermediate_graph_degree = 128;
   size_t max_iterations            = 20;
   float termination_threshold      = 0.0001;
   bool return_distances            = true;
+  DIST_COMP_DTYPE dist_comp_dtype  = DIST_COMP_DTYPE::AUTO;
```
jinsolp, line 62: @divyegala It already defaults to AUTO here, which would be taken in `nn_descent.cpp`

divyegala: That file is the source of C API, and this default is for the C++ API.

jinsolp: I believe all the defaults of cpp are passed when we call `cuvsNNDescentIndexParamsCreate`
https://github.com/rapidsai/cuvs/blob/1959a0dda1f1d1e49124860f1a766793e1598706/c/src/neighbors/nn_descent.cpp#L167-L182

C doesn't support default struct init and so nothing is initialized in the c struct:
https://github.com/rapidsai/cuvs/blob/1959a0dda1f1d1e49124860f1a766793e1598706/c/include/cuvs/neighbors/nn_descent.h#L38-L46

divyegala: I haven't kept up with the changes to the C API, my apologies. We used to explicitly default initialize in the create function previously.

PR #1416, opened 2025-10-09
Deallocation should be noexcept

bdice:
Update for compatibility with https://github.com/rapidsai/rmm/pull/2077.

### DIFF
--- a/cpp/bench/ann/src/common/cuda_huge_page_resource.hpp
+++ b/cpp/bench/ann/src/common/cuda_huge_page_resource.hpp
@@ -1,5 +1,5 @@
 /*
- * Copyright (c) 2023-2024, NVIDIA CORPORATION.
+ * Copyright (c) 2023-2025, NVIDIA CORPORATION.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


cpp/bench/ann/src/common/cuda_huge_page_resource.hpp
```diff
@@ -79,7 +79,7 @@ class cuda_huge_page_resource final : public rmm::mr::device_memory_resource {
    *
    * @param p Pointer to be deallocated
    */
-  void do_deallocate(void* ptr, std::size_t size, rmm::cuda_stream_view) override
+  void do_deallocate(void* ptr, std::size_t size, rmm::cuda_stream_view) noexcept override
   {
     if (munmap(ptr, size) == -1) { RAFT_FAIL("huge_page_resource::munmap"); }
```
bdice, line 84: This function must be made `noexcept`. Throwing in `noexcept` code forces `std::terminate`, so this `RAFT_FAIL` will trigger that. I don't know if this error message will be displayed, but my suspicion is that it won't show up. The right solution is probably to adopt a macro that outputs to `std::cerr` before throwing/terminating rather than relying on the C++ exception message to show? We have had to do similar tricks with [RMM_ASSERT_CUDA_SUCCESS](https://github.com/rapidsai/rmm/blob/f3bb0b822b02432e76476d5d9174843bfea81334/cpp/include/rmm/detail/error.hpp#L177-L218) because it can't throw in `noexcept` functions.

Since this fix is probably necessary to unblock builds (and I need to file similar fixes for other repos), I am leaving this as a follow-up task for the cuVS team to handle whichever way they prefer.

divyegala: Can you capture this in an issue?

achirkin: https://github.com/rapidsai/cuvs/pull/1417

PR #1425, opened 2025-10-14
Spectral Clustering

### LINKED ISSUE #1201: Spectral Clustering
Create the spectral clustering algorithm in `cuvs::cluster::spectral`

aamijar:
Resolves #1201

This PR introduces the Spectral Clustering algorithm which builds on top of Spectral Embedding.

This PR will enable us to remove the legacy and _inaccurate_ spectral clustering algorithm called `partition`

This PR also adds `double` precision support for spectral embedding and spectral clustering. The new templates require a refactor of the source files to `.cuh` and the template instantiations in `.cu`

Gtests have been added to this PR with synthetic data in a similar fashion as kmeans gtests.
I have tested this PR through python bindings and by comparing the plots with sklearn's SpectralClustering.

Here is a plot of sklearn vs ours.
Screenshot 2025-10-14 at 2 53 01 PM


aamijar:
Update:
I have added the current version of the gtests, which has a similar structure to how `kmeans.cu` does it.
I had to reduce the ARI threshold score.
I was able to setup this same test in python as well to check what happens with sklearn SpectralClustering when you pass in this synthetic data. However, it seems that sklearn gives a perfect 1.0 ARI. So there must be something going wrong in our implementation. I am digging through to see what the possible issue could be.

Update:
So it turns out that the solver is working fine, but we just need to increase the precision using smaller tolerance. I also confirmed that the laplacian step output matrix between sklearn and ours is the same.

aamijar:
Update:
I'm running into some problems with gpu KMeans potentially not being consistent compared to cpu KMeans. I wonder if that is expected? For example if I replace the KMeans in sklearn spectral clustering with cuml.cluster KMeans then it gives a much lower ARI.
This was solved by `oversampling_factor=0`

Update:
Key findings
`oversampling_factor=0` k-means++ init method
`increasing solver precision, tol=1e-5->0`
`fix to remove duplicate entries in laplacian`
`max_iterations = 10 * n_samples` instead of fixed `1000`
are able to improve the quality.
we are able to achieve near perfect scores compared to ground truth in the gtests, mean of 0.97 ARI across 20 runs.

cpp/include/cuvs/cluster/spectral.hpp
```diff
@@ -0,0 +1,38 @@
+/*
+ * Copyright (c) 2025, NVIDIA CORPORATION.
```
cjnolet: I propose we name this file `spectral.hpp` instead of `spectral_clustering`. The clustering is already in the namespace, so it's redundant.

cpp/include/cuvs/preprocessing/spectral_embedding.hpp
```diff
@@ -174,4 +174,9 @@ void transform(raft::resources const& handle,
  * @}
  */
 
+void create_connectivity_graph(raft::resources const& handle,
```
cjnolet, line 175: This is not part of the standard "fit()" "predict()" APIs that one would normally use day to day, so please add it to the `cuvs::preprocessing::spectral_embedding::helpers` namespace.

cpp/include/cuvs/cluster/spectral.hpp
aamijar: Addressed in https://github.com/rapidsai/cuvs/pull/1425/commits/f1ab56a5a4ffdf7ac0cfda7eae27f6e87c116a08

cpp/include/cuvs/preprocessing/spectral_embedding.hpp
aamijar: Addressed in https://github.com/rapidsai/cuvs/pull/1425/commits/f1ab56a5a4ffdf7ac0cfda7eae27f6e87c116a08

cpp/src/cluster/spectral.cu
```diff
@@ -0,0 +1,70 @@
...
+
+  cuvs::preprocessing::spectral_embedding::transform(
+    handle, spectral_embedding_config, connectivity_graph, embedding_col_major.view());
+
+  raft::linalg::transpose(handle,
+                          embedding_col_major.data_handle(),
+                          embedding_row_major.data_handle(),
+                          n_samples,
+                          config.n_components,
+                          raft::resource::get_cuda_stream(handle));
```
viclafargue: I guess that it is necessary to double VRAM usage and perform a transposition here, but would be great if we could find a workaround in the long term.

aamijar: Thanks for the review! Yes, we have to switch from col to row major since spectral embedding needs col major and kmeans needs row major. Yes, we are using `int` as nnz type currently until we resolve the scale up work in spectral embedding.

cpp/include/cuvs/cluster/spectral.hpp
```diff
@@ -0,0 +1,38 @@
...
+#include <raft/core/resources.hpp>
+
+namespace cuvs::cluster::spectral {
+
+struct params {
+  int n_clusters;
+  int n_components;
+  int n_init;
+  int n_neighbors;
+  uint64_t seed;
```
jnke2016: Can we expose the `rng_state` instead of the seed?

cpp/include/cuvs/cluster/spectral.hpp
```diff
@@ -0,0 +1,38 @@
...
+};
+
+void fit_predict(raft::resources const& handle,
+                 params config,
+                 raft::device_coo_matrix_view<float, int, int, int> connectivity_graph,
```
jnke2016, line 25: `raft::device_coo_matrix_view<double, int, int, int>` is not supported while cuGraph has templates for both `float` and `double`. Can we add support for the latter as well. Otherwise, i get the below error
```
error: no suitable user-defined conversion from "raft::device_coo_matrix_view<double, int, int, int>" to "raft::device_coo_matrix_view<float, int, int, int>" exists
```

jnke2016:
I tested this PR in cuGraph and provided few suggestions. A cuGraph [PR](https://github.com/rapidsai/cugraph/pull/5326) incorporating the cuVS Spectral Clustering API is ready for review and will be merged as soon as this PR is. @aamijar

cpp/include/cuvs/cluster/spectral.hpp
aamijar: Addressed in https://github.com/rapidsai/cuvs/pull/1425/commits/d9d4bc8d89f412b6076a0670ffb35fe8b30b21ed

cpp/include/cuvs/cluster/spectral.hpp
aamijar: I have added `double` support in https://github.com/rapidsai/cuvs/pull/1425/commits/5c608f3d9a96f17ef78d2c80e97aa3e5fb8d0511

cjnolet: Do we really want `int` here? Specifically for the nnz, we probably want an int64 for that. Otherwise, you are limited to very small graphs where the total number of nonzeros is < 2B (which is very small).

aamijar: We will be able to support int64_t once the spectral scale up work is resolved https://github.com/rapidsai/cuml/issues/7225

cjnolet: Can you open a cuvs issue specifically for this API please, so that it oesn't get left behind once we're able to shift to the int_64 in other places?

cjnolet: Also, please link the issue in the code here in a todo that we're going to be changing the ints to int_64 soon.

aamijar: done in https://github.com/rapidsai/cuvs/pull/1425/commits/40ac5c2313bd84015fcb807c91d5a491c35ccd23

cjnolet (approved):

jnke2016 (approved):
Thanks for addressing the changes I requested and it looks good to me. I tested this PR in cuGraph and all tests succeeded. The cuGraph [PR](https://github.com/rapidsai/cugraph/pull/5326) will be merged once this one is.

PR #1427, opened 2025-10-14
Fix table of content documentation

lowener:
Use javascript to not display duplicates in doxygen doc


lowener:
Before:
grafik

After:
grafik

PR #1428, opened 2025-10-14
Expose pq list data to python

benfred:
Expose pq list data to python. Also change cuvsMatrixCopy to handle more than just 2D tensors - so that things like `pq_centers` can be converting to host memory without needing to involve cupy .

### DIFF
--- a/c/src/core/c_api.cpp
+++ b/c/src/core/c_api.cpp
@@ -1,5 +1,5 @@
 /*
- * SPDX-FileCopyrightText: Copyright (c) 2024-2025, NVIDIA CORPORATION.
+ * SPDX-FileCopyrightText: Copyright (c) 2024-2026, NVIDIA CORPORATION.
  * SPDX-License-Identifier: Apache-2.0
  */
 
--- a/c/include/cuvs/neighbors/ivf_pq.h
+++ b/c/include/cuvs/neighbors/ivf_pq.h
@@ -1,5 +1,5 @@
 /*
- * SPDX-FileCopyrightText: Copyright (c) 2024-2025, NVIDIA CORPORATION.
+ * SPDX-FileCopyrightText: Copyright (c) 2024-2026, NVIDIA CORPORATION.
  * SPDX-License-Identifier: Apache-2.0
  */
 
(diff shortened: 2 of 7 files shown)


c/src/core/c_api.cpp
```diff
@@ -240,18 +240,33 @@ void _copy_matrix(cuvsResources_t res, DLManagedTensor* src_managed, DLManagedTe
...
+  } else {
+    // Otherwise use cudaMemcpyAsync - and assert that we don't have strided data
+    RAFT_EXPECTS(src.strides == nullptr, "cuvsCopyMatrix only supports strides with 2D inputs");
+    RAFT_EXPECTS(dst.strides == nullptr, "cuvsCopyMatrix only supports strides with 2D inputs");
+
+    size_t elements = 1;
+    for (int64_t i = 0; i < src.ndim; ++i) {
+      elements *= src.shape[i];
+    }
+    cudaMemcpyAsync(dst.data, src.data, elements * sizeof(T), cudaMemcpyDefault, stream);
```
lowener: Check return value of that CUDA API call, or use raft::copy (https://github.com/rapidsai/raft/blob/main/cpp/include/raft/util/cudart_utils.hpp#L136)

c/src/core/c_api.cpp
```diff
@@ -240,18 +240,33 @@ void _copy_matrix(cuvsResources_t res, DLManagedTensor* src_managed, DLManagedTe
...
+                         dst_row_stride,
+                         static_cast<const T*>(src.data),
+                         src_row_stride,
+                         src.shape[1],
+                         src.shape[0],
+                         stream);
+  } else {
+    // Otherwise use cudaMemcpyAsync - and assert that we don't have strided data
+    RAFT_EXPECTS(src.strides == nullptr, "cuvsCopyMatrix only supports strides with 2D inputs");
+    RAFT_EXPECTS(dst.strides == nullptr, "cuvsCopyMatrix only supports strides with 2D inputs");
```
lowener, line 285: Can [this raft::copy](https://github.com/rapidsai/raft/blob/main/cpp/include/raft/core/copy.cuh#L46) be used to suppport strides in non-2D output?

PR #1430, opened 2025-10-14
Test RAFT with CCCL mdspan

bdice:
Testing https://github.com/rapidsai/raft/pull/2836/.


bdice:
The build errors here seem to stem from a bunch of places in the source code where the extents type is left as default (`uint32_t`) or defined incorrectly. The previous mdspan implementation was willing to implicitly convert something like `raft::make_device_matrix_view<const int64_t, uint64_t>` to `raft::make_device_matrix_view<const int64_t, int64_t>`, _unsafely_ casting `uint64` to `int64` in the extents type. The CCCL mdspan does not allow that. I am iterating through build errors to ensure that we have extent types properly defined.

bdice:
Replaced by https://github.com/rapidsai/cuvs/pull/1605, which is now complete.

Issue #1322, opened 2025-09-08
[BUG] CMake build flag DISABLE_OPENMP doesn't actually disable OpenMP


robertmaynard:
**Describe the bug**
The OpenMP package is searched for unconditionally on https://github.com/rapidsai/cuvs/blob/branch-25.10/cpp/CMakeLists.txt#L585 and therefore enabling `DISABLE_OPENMP` doesn't actually do anything

**Steps/Code to reproduce bug**
Build with `DISABLE_OPENMP` enabled, and run `ldd` on the generated libcuvs.so. You will still see an openmp  dependency

robertmaynard:
re-opening this as multiple cuvs source files call `omp_get_max_threads`, `omp_get_num_procs`, `omp_get_num_threads`, and `omp_get_thread_num` all which can't occur when `DISABLE_OPENMP` is set.

PR #1435, opened 2025-10-16
Properly guard usage of openmp function calls 

robertmaynard:
Proper support to disable OpenMP also requires that any calls to functions like `omp_get_max_threads` need to be guarded by a `_OPENMP` check.

This is done by adding openmp wrapper functions into `cuvs/core` that properly handle the `_OPENMP` guards and behave as required when OpenMP is disabled. 

Required to fix https://github.com/rapidsai/cuvs/issues/1322

### DIFF
--- a/cpp/src/core/omp_wrapper.cpp
+++ b/cpp/src/core/omp_wrapper.cpp
@@ -22,7 +22,7 @@
 
 namespace cuvs::core::omp {
 
-bool is_omp_enabled()
+constexpr bool is_omp_enabled()
 {
 #if defined(_OPENMP)
   return true;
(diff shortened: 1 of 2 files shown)


cpp/src/core/omp_wrapper.cpp
```diff
@@ -0,0 +1,87 @@
...
+
+#include <omp.h>
+
+#include <raft/core/logger.hpp>
+
+#include "omp_wrapper.hpp"
+
+namespace cuvs::core::omp {
+
+bool is_omp_enabled()
```
divyegala:
```suggestion
constexpr bool is_omp_enabled()
```

cpp/src/core/omp_wrapper.cpp
```diff
@@ -0,0 +1,87 @@
...
+}
+
+int get_max_threads()
+{
+#if defined(_OPENMP)
+  return omp_get_max_threads();
+#else
+  return 1;
+#endif
+}
```
divyegala:
```suggestion
constexpr int get_max_threads()
{
  return is_omp_enabled() ? omp_get_max_threads() : 1;
}
```

robertmaynard: Great improvement! 

I have refactored all the function implementations to use this pattern

Issue #1418, opened 2025-10-09
Using a wrapper `byte_array` to support both `uint8_t` and `int8_t`


divyegala:
Currently, `libcuvs` instantiates both `uint8_t` and `int8_t` which causes binary size bloat. Given the relative similarity of both types and the ease of normalization of values that are encompassed by them, I propose that we combine the two types using a wrapper `cuvs::detail::byte_array`.

The public API remains the same:
```cpp
namespace cuvs {

void foo(raft::device_matrix_view<uint8_t, ...>(...);

void foo(raft::device_matrix_view<int8_t, ...>(...);

}
```

The definition of `cuvs::detail::byte_array`, which is a struct that would always normalize `int8_t` values to `uint8_t` when returning and de-normalize when assigning:
```cpp
namespace cuvs::detail {

struct byte_array {
  void* data = nullptr;
  bool is_signed = false;

  byte_array(void* ptr, bool signed_flag)
      : data(ptr), is_signed(signed_flag) {}

  // Proxy that references an element in the array
  struct byte {
    byte_array* parent = nullptr;
    int64_t idx = -1;
    uint8_t value = 0;   // used for detached proxies

    // Constructor for live proxy
    byte(byte_array& p, int64_t i) : parent(&p), idx(i) {}

    // Copy constructor: detached copy stores the current value
    byte(const byte& other)
        : parent(nullptr), idx(-1), value(static_cast<uint8_t>(other)) {}

    // Copy assignment: detached copy stores value
    byte& operator=(const byte& other) {
        parent = nullptr;
        idx = -1;
        value = static_cast<uint8_t>(other);
        return *this;
    }

    // Deleted move operations
    byte(byte&& other) = delete;
    byte& operator=(byte&& other) = delete;

    // Conversion to uint8_t
    operator uint8_t() const {
        if (parent) {
            if (parent->is_signed) {
                int8_t val = reinterpret_cast<int8_t*>(parent->data)[idx];
                return static_cast<uint8_t>(static_cast<int16_t>(val) + 128);
            } else {
                return reinterpret_cast<uint8_t*>(parent->data)[idx];
            }
        } else {
            return value;  // return local value if detached
        }
    }

    // Assignment from uint8_t
    byte& operator=(uint8_t normalized_value) {
        if (parent) {
            if (parent->is_signed) {
                reinterpret_cast<int8_t*>(parent->data)[idx] =
[...]

PR #1437, opened 2025-10-18
Using `byte_arithmetic_ptr` in `interleaved_scan_kernel`

divyegala:
Building on top of idea in https://github.com/rapidsai/cuvs/issues/1418

This PR reduces the binary size of CUDA 12.9 `libcuvs.so` from **1175.32 MB to 1127.43 MB**.

### Benchmark Parameter Sweep

The benchmarks were run across the following parameter combinations:

- **Data types**: `int8`, `uint8`
- **Distance metrics**: `inner_product`, `sqeuclidean`
- **Dataset sizes (n_rows)**: 25,000, 100,000, 250,000, 500,000, 1,000,000
- **Dimensions (n_cols)**: 64, 128, 256, 512
- **Query batch sizes (n_queries)**: 10, 100, 1,000, 10,000
- **k values**: 32, 64, 128, 256

**Total configurations tested**: 1,280 (per implementation)

### Overall Performance Summary

| dtype | metric | tests | avg speedup | median speedup | avg overhead | baseline QPS | hybrid QPS |
|-------|--------|-------|-------------|----------------|--------------|--------------|------------|
| int8 | sqeuclidean | 320 | 0.9579x | 0.9661x | +4.94% | 556.8K | 549.6K |
| int8 | inner_product | 320 | 1.0118x | 1.0097x | -1.05% | 586.9K | 589.3K |
| uint8 | sqeuclidean | 320 | 0.9988x | 0.9808x | +0.62% | 545.4K | 557.2K |
| uint8 | inner_product | 320 | 0.9745x | 0.9910x | +3.11% | 170.9K | 167.9K |

### Performance Distribution

| dtype | metric | min speedup | 25th % | 50th % | 75th % | max speedup |
|-------|--------|-------------|--------|--------|--------|-------------|
| int8 | sqeuclidean | 0.593x | 0.926x | 0.966x | 1.000x | 1.087x |
| int8 | inner_product | 0.878x | 1.000x | 1.010x | 1.029x | 1.174x |
| uint8 | sqeuclidean | 0.829x | 0.957x | 0.981x | 1.046x | 1.272x |
| uint8 | inner_product | 0.755x | 0.954x | 0.991x | 1.007x | 1.164x |


cpp/include/cuvs/core/byte_array.hpp
```diff
@@ -0,0 +1,120 @@
...
+ * limitations under the License.
+ */
+
+#pragma once
+
+#include <cstdint>
+
+namespace cuvs::detail {
+
+struct byte_array {
```
achirkin: What do you think about renaming this to something like `byte_ptr` or `byte_arithmetic_ptr`? The reason I'm asking is that this structure semantically looks more like a pointer (dereferencing, pointer arithmetic) rather than a container (no size attribute).

cpp/include/cuvs/core/byte_arithmetic_ptr.hpp
```diff
@@ -0,0 +1,120 @@
...
+
+#pragma once
+
+#include <cstdint>
+
+namespace cuvs::detail {
+
+struct byte_array {
+  void* data     = nullptr;
+  bool is_signed = false;
```
achirkin, line 25: Passing this on gpu will probably require three words, right? This could have the register usage consequence in some edge cases. If only could we rely on the `data` being aligned - then we could use the lowest bit of it as the signedness tag and pass this struct everywhere in place of the pointers.

divyegala: The data being aligned doesn't guarantee that the lowest bit will be set in the int8_t case, right? If the first int8 value is positive then the sign bit won't be set.

Unless I misunderstood what you are saying?

We could always try to pass this as a reference to device functions instead, if that's helpful.

achirkin: The pointers will almost surely be aligned, but technically one can pass an unaligned pointer since the data is one byte granularity. My question is would it be acceptable for us to always assume it's at least two-byte aligned (or make an alignment check somewhere and throw an error otherwise)?

divyegala: I think it should be acceptable for us to check that. In the case of alignment, how do you propose we check for signed-ness?

cpp/include/cuvs/core/byte_array.hpp
divyegala: That's a good idea! I like `byte_arithmetic_ptr`

cpp/include/cuvs/core/byte_arithmetic_ptr.hpp
achirkin: I'd do something along the lines of this:
```C++
struct byte_arithmetic_ptr {
 private:
    constexpr static uintptr_t kSignMask = 0x1;
    uintptr_t value_;
  public:
    byte_arithmetic_ptr(uint8_t* ptr): value_(reinterpret_cast<uintptr_t>(ptr) {}
    byte_arithmetic_ptr(int8_t* ptr): value_(reinterpret_cast<uintptr_t>(ptr | kSignMask) {}
   
    constexpr void* get_data() {
      return reinterpret_cast<void*>(value_ & ~kSignMask);
    }
    constexpr bool is_signed() const {
      return (value_ & kSignMask) == kSignMask;
    }
};
```

divyegala: Unfortunately, it looks like even our gtests don't pass. Here's an int8 address: `139761184407553` and it is not 2-byte aligned.

achirkin: Do we have these pointers passed down from the public api anywhere?

divyegala: Sorry, what do you mean? The `ivf_flat::search` public API just passes these pointers along to `ivf_flat_interleaved_scan` function

cpp/src/neighbors/ivf_flat/ivf_flat_interleaved_scan.cuh
```diff
@@ -202,335 +217,356 @@ struct loadAndComputeDist {
...
 #pragma unroll
     for (int j = 0; j < kUnroll; ++j) {
       uint32_t encV[veclen_int];
-      raft::ldg(
-        encV,
-        reinterpret_cast<unsigned const*>(data) + loadIndex + j * kIndexGroupSize * veclen_int);
+      const int byte_offset = (loadIndex + j * kIndexGroupSize * veclen_int) * 4;
+#pragma unroll
+      for (int i = 0; i < veclen_int; i++) {
+        encV[i] = pack_bytes_uint32(data[byte_offset + i * 4 + 0],
```
tarang-jain: Isn't there a better way to do this? I think we can retain the `raft::ldg` mechanism of fast loads into `encV`. Then do the +128 arithmetic (if need be) on `encV` in place. We should be able to avoid so many (as many as 16?) global reads. You can iterate over the `veclen_int` values of encV and extract 4 bytes from each value with suitable bit manipulations.

cpp/src/neighbors/ivf_flat/ivf_flat_interleaved_scan.cuh
```diff
@@ -202,335 +217,356 @@ struct loadAndComputeDist {
...
-    constexpr int veclen_int = uint8_veclen / 4;  // converting uint8_t veclens to int
-    uint32_t queryReg =
-      (lane_id < 8) ? reinterpret_cast<unsigned const*>(query + baseLoadIndex)[lane_id] : 0;
+    constexpr int veclen_int     = uint8_veclen / 4;  // converting uint8_t veclens to int
+    const bool is_signed         = data.is_signed;
+    constexpr int32_t offset_reg = 0x80808080;  // 128 in each byte position
+    uint32_t queryReg            = 0;
+    if (lane_id < 8) {
+      const int byte_offset = baseLoadIndex + lane_id * 4;
+      queryReg              = pack_bytes_uint32(query[byte_offset + 0],
```
tarang-jain: Same argument as above -- why not first do the fast loads and then the arithmetic? Perhaps the `byte_arithmetic_ptr` class can have another function to provide a raw (reinterpret_cast-ed) uint8_t pointer.

cpp/src/neighbors/ivf_flat/ivf_flat_interleaved_scan.cuh
divyegala: So you want to reinterpret the `int8` as `uint8` during read with `raft::ldg` and then manipulate with the arithmetic? Hmmm that's a good idea and perhaps it is possible. Let me try to implement that.

cpp/src/neighbors/ivf_flat/ivf_flat_interleaved_scan.cuh
```diff
@@ -202,335 +217,356 @@ struct loadAndComputeDist {
...
       raft::lds(queryRegs,
                 reinterpret_cast<unsigned const*>(query_shared + shmemIndex) + j * veclen_int);
 #pragma unroll
       for (int k = 0; k < veclen_int; k++) {
-        compute_dist(dist, queryRegs[k], encV[k]);
-        if constexpr (ComputeNorm) {
-          norm_query = raft::dp4a(queryRegs[k], queryRegs[k], norm_query);
-          norm_data  = raft::dp4a(encV[k], encV[k], norm_data);
+        if constexpr (is_inner_prod_dist<Lambda>::value) {
+          int32_t q = is_signed ? (static_cast<int32_t>(queryRegs[k]) - offset_reg)
```
tarang-jain: Is a `static_cast` correct here? That would change the underlying bit string, isn't it?

tarang-jain: Actually -- why do we even need this subtraction? Cant we just use efficient loading (with perhaps `reinterpret_cast` into a uint32_t array with `raft::ldg` as before, and just `reinterpret_cast` it here as int32_t? At the end of the day, we just need to bit string to be correct, isn't it?

divyegala: No, the value of `queryRegs` is always positive

cpp/include/cuvs/core/byte_arithmetic_ptr.hpp
achirkin: Ah, so that's the query pointer what causes the problem (the ivf lists are aligned). Then it's debatable whether it makes sense to require it to be aligned. Could be good for performance, but (1) loading queries isn't a bottleneck, (2) it would require changes to the code where we increment by potentially odd offset https://github.com/rapidsai/cuvs/blob/ef65e5aa7f8c0d3a9e0a2691b02a8617a78829d3/cpp/src/neighbors/ivf_flat/ivf_flat_interleaved_scan.cuh#L926

divyegala: I think we can keep this in mind for later, in case performance becomes an issue? I just posted some performance #s in the description body and IMO they look pretty good :)

achirkin:
@divyegala Could you please look into the worst-case scenario where you reported almost two-fold slowdown? (int8, euclidean distance). 
 - Is it one or few consistent cases or random fluctuation?
 - Is it one of the small-data-size case that are latency bound, or a more relevant big-data-size case where the throughput impacted due to e.g. accidentally broken vectorized loads?

divyegala:
@achirkin here's the top 10 worst cases. To me the first one looks like possibly a one-off, not sure what happened there exactly. For the other cases, I think 20% is an acceptable perf drop keeping the context in mind that the raw numbers are a few hundred microseconds as well average drop only being ~5% meaning we make up the numbers elsewhere.

#### Top 10 Worst Cases

| Rank | n_rows | n_cols | n_queries | k | Baseline (ms) | Hybrid (ms) | Speedup | Overhead (μs/query) |
|------|--------|--------|-----------|---|---------------|-------------|---------|---------------------|
| 1 | 1,000,000 | 64 | 100 | 128 | 0.786 | 1.324 | 0.593x | 5.4 |
| 2 | 100,000 | 512 | 10 | 32 | 0.447 | 0.590 | 0.758x | 14.3 |
| 3 | 250,000 | 64 | 10 | 32 | 0.419 | 0.535 | 0.783x | 11.6 |
| 4 | 25,000 | 512 | 10 | 128 | 0.444 | 0.564 | 0.786x | 12.1 |
| 5 | 100,000 | 128 | 100 | 128 | 0.449 | 0.571 | 0.787x | 1.2 |
| 6 | 500,000 | 128 | 10 | 128 | 0.432 | 0.546 | 0.790x | 11.5 |
| 7 | 25,000 | 512 | 10 | 32 | 0.437 | 0.550 | 0.794x | 11.3 |
| 8 | 25,000 | 512 | 100 | 64 | 0.465 | 0.579 | 0.803x | 1.1 |
| 9 | 25,000 | 512 | 10 | 64 | 0.430 | 0.532 | 0.807x | 10.3 |
| 10 | 250,000 | 64 | 10 | 64 | 0.420 | 0.520 | 0.808x | 10.0 |

achirkin:
Are you doing micro-benchmarks or a full ann-bench search? Don't these translate to QPS 1-1? If so, 20% is not so little tbh

divyegala:
@achirkin microbenchmarks via the Python API. I think the understanding is that we are okay with the tradeoff of a lesser performant `int8` for binary size savings. I think the overall impact of 5% for `int8` + `sqeuclidean` is not bad at all, given that this is also a temporary measure. As mentioned offline:
1. CUDA 13 will support the full type matrix via JIT/LTO
2. We will reverse these changes for CUDA 12 when we upgrade to driver r580
3. `int8` is a lesser used type than `uint8`

cc @cjnolet

achirkin:
Ok, thanks for the clarification. Could you please still run the micro-benchmarks with more iterations/warmup or full benchmarks for the found worst-case scenario of 0.593x speedup? Just to be on the safe side

divyegala:
Yes. Can you suggest which dataset you would like to see? I'll run cuvs-bench for it.

divyegala:
@achirkin I also ran the worst-case benchmark a few more times and found the median time to go from 1.324 ms to 1.006 ms, which brings it to about a speedup of 0.781x.

cpp/src/neighbors/ivf_flat/ivf_flat_interleaved_scan.cuh
```diff
@@ -216,24 +270,31 @@ struct loadAndComputeDist<kUnroll, Lambda, uint8_veclen, uint8_t, uint32_t, Comp
...
       raft::lds(queryRegs,
                 reinterpret_cast<unsigned const*>(query_shared + shmemIndex) + j * veclen_int);
 #pragma unroll
       for (int k = 0; k < veclen_int; k++) {
+        // Normalize int8_t bytes to uint8_t for Euclidean distance (translation-invariant)
+        if (is_signed) {
+          encV[k]      = normalize_int8_packed(encV[k]);
+          queryRegs[k] = normalize_int8_packed(queryRegs[k]);
+        }
         compute_dist(dist, queryRegs[k], encV[k]);
```
tarang-jain, line 287: Aren't we supposed to normalize only when the metric is L2Expanded?

divyegala: Yes, this code-path is only instantiated for `int8` when the metric is euclidean-based https://github.com/rapidsai/cuvs/pull/1437/files#diff-403c2fcd55246356fc13c2f369a109027c0983f6fafa31109d56f6f9cb273439R978-R980

cpp/src/neighbors/ivf_flat/ivf_flat_interleaved_scan.cuh
```diff
@@ -86,6 +90,56 @@ __device__ inline void copy_vectorized(T* out, const T* in, uint32_t n)
   }
 }
 
+// Specialization for byte_arithmetic_ptr -> uint8_t* (for int8_t normalization)
+__device__ inline void copy_vectorized(uint8_t* out,
```
lowener, line 83: Can those function be in an other file that can be included? That way the other algorithms switching to byte array can use them

lowener (commented):
The changes look good!
Modify `generate_ivf_flat.py` too to reflect this change.

cpp/src/neighbors/ivf_flat/ivf_flat_interleaved_scan.cuh
divyegala: I believe @tarang-jain did a scan and these functions are unused elsewhere.

tarang-jain: Thats correct! copy_vectorized is not being used elsewhere.

divyegala:
> Modify generate_ivf_flat.py too to reflect this change.

@lowener we don't use `generate_ivf_flat.py` to generate the TUs changed in this PR.

PR #1438, opened 2025-10-20
Draft: Show how we can build libcuvs_c.tar.gz in CI

robertmaynard:


ci/test_standalone_c.sh
```diff
@@ -0,0 +1,21 @@
+#!/bin/bash
+# SPDX-FileCopyrightText: Copyright (c) 2022-2025, NVIDIA CORPORATION.
+# SPDX-License-Identifier: Apache-2.0
+
+set -euo pipefail
+
+
+# Download the standalone C library artifact
+pkg_name="libcuvs_c_${RAPIDS_CUDA_VERSION}.tar.gz.zip"
```
gforsyth:
```suggestion
pkg_name="libcuvs_c_${RAPIDS_CUDA_VERSION}.tar.gz"
```

I believe the artifact _name_ is suffixed with `tar.gz`, where the actual file contained also has the `.zip` suffix. Fun times.

You can see here the name that has been used with the successful upload: https://github.com/rapidsai/cuvs/actions/runs/19074700862/job/54487081522?pr=1438#step:12:129

gforsyth:
CI isn't running because of an error in `pr.yaml`

```

Invalid workflow file: .github/workflows/pr.yaml#L188
You have an error in your yaml syntax on line 188
```

gforsyth:
Workflow yaml is invalid again: https://github.com/rapidsai/cuvs/actions/runs/19084677724/workflow

PR #1441, opened 2025-10-21
[Java] One PinnedMemoryBuffer per CuVSResourcesImpl

ldematte:
While profiling cuvs-java, we found that allocating a `PinnedMemoryBuffer` for each host->device or device->host memory copy was unnecessary and wasteful.
This PR moves the allocation of a `PinnedMemoryBuffer` to `CuVSResourcesImpl`, so that the buffer can be cached and reused. Since `CuVSResources` are already meant to be per-thread, this is safe, as the `PinnedMemoryBuffer` will never be used concurrently.
In order to do it cleanly, we introduced two named `ScopedAccess` classes and a helper method that will always find its way to the internal `MemorySegment` used by native functions to access the buffer, without the need to expose it via the public interface.

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,3 +1,6 @@
+# SPDX-FileCopyrightText: Copyright (c) 2023-2025, NVIDIA CORPORATION. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
 # syntax=docker/dockerfile:1.5
 
 ARG BASE
@@ -42,6 +45,7 @@ ENV HISTFILE="/home/coder/.cache/._bash_history"
 ENV AWS_ROLE_ARN="arn:aws:iam::279114543810:role/nv-gha-token-sccache-devs"
 ENV SCCACHE_REGION="us-east-2"
 ENV SCCACHE_BUCKET="rapids-sccache-devs"
+ENV SCCACHE_S3_USE_PREPROCESSOR_CACHE_MODE=true
 ENV SCCACHE_IDLE_TIMEOUT=0
 
 ###
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -39,8 +39,10 @@ jobs:
       build_type: ${{ inputs.build_type || 'branch' }}
       branch: ${{ inputs.branch }}
       date: ${{ inputs.date }}
+      node_type: cpu16
       script: ci/build_cpp.sh
       sha: ${{ inputs.sha }}
+      sccache-dist-token-secret-name: GIST_REPO_READ_ORG_GITHUB_TOKEN
 
   rocky8-clib-standalone-build:
     secrets: inherit
@@ -63,6 +65,7 @@ jobs:
       artifact-name: "libcuvs_c_${{ matrix.cuda_version }}.tar.gz"
       file_to_upload: "libcuvs_c.tar.gz"
       sha: ${{ inputs.sha }}
+      sccache-dist-token-secret-name: GIST_REPO_READ_ORG_GITHUB_TOKEN
   rust-build:
     needs: cpp-build
     secrets: inherit
@@ -84,6 +87,7 @@ jobs:
       node_type: "gpu-l4-latest-1"
       script: "ci/build_rust.sh"
       sha: ${{ inputs.sha }}
+      sccache-dist-token-secret-name: GIST_REPO_READ_ORG_GITHUB_TOKEN
   go-build:
     needs: cpp-build
     secrets: inherit
@@ -127,6 +131,7 @@ jobs:
       artifact-name: "cuvs-java-cuda${{ matrix.cuda_version }}"
       file_to_upload: "java/cuvs-java/target/"
       sha: ${{ inputs.sha }}
+      sccache-dist-token-secret-name: GIST_REPO_READ_ORG_GITHUB_TOKEN
   python-build:
     needs: [cpp-build]
     secrets: inherit
@@ -137,6 +142,7 @@ jobs:
       date: ${{ inputs.date }}
       script: ci/build_python.sh
       sha: ${{ inputs.sha }}
+      sccache-dist-token-secret-name: GIST_REPO_READ_ORG_GITHUB_TOKEN
   upload-conda:
     needs: [cpp-build, python-build]
     secrets: inherit
@@ -173,11 +179,13 @@ jobs:
       branch: ${{ inputs.branch }}
       sha: ${{ inputs.sha }}
       date: ${{ inputs.date }}
+      node_type: cpu16
       script: ci/build_wheel_libcuvs.sh
       # build for every combination of arch and CUDA version, but only for the latest Python
       matrix_filter: group_by([.ARCH, (.CUDA_VER|split(".")|map(tonumber)|.[0])]) | map(max_by(.PY_VER|split(".")|map(tonumber)))
       package-name: libcuvs
       package-type: cpp
+      sccache-dist-token-secret-name: GIST_REPO_READ_ORG_GITHUB_TOKEN
   wheel-publish-libcuvs:
     needs: wheel-build-libcuvs
     secrets: inherit
@@ -198,8 +206,10 @@ jobs:
       branch: ${{ inputs.branch }}
       sha: ${{ inputs.sha }}
       date: ${{ inputs.date }}
+      node_type: cpu8
       script: ci/build_wheel_cuvs.sh
       package-name: cuvs
+      sccache-dist-token-secret-name: GIST_REPO_READ_ORG_GITHUB_TOKEN
       package-type: python
   wheel-publish-cuvs:
     needs: wheel-build-cuvs
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -144,6 +144,7 @@ jobs:
       build_type: pull-request
       node_type: cpu16
       script: ci/build_cpp.sh
+      sccache-dist-token-secret-name: GIST_REPO_READ_ORG_GITHUB_TOKEN
   conda-cpp-tests:
     needs: [conda-cpp-build, changed-files]
     secrets: inherit
@@ -152,6 +153,7 @@ jobs:
     with:
       build_type: pull-request
       script: ci/test_cpp.sh
+      sccache-dist-token-secret-name: GIST_REPO_READ_ORG_GITHUB_TOKEN
   conda-cpp-checks:
     needs: conda-cpp-build
     secrets: inherit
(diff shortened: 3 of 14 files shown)


java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/common/PinnedMemoryBuffer.java
```diff
@@ -10,35 +10,18 @@
...
-      this.hostBufferBytes = rowBytes * rowCount;
-    }
-  }
-
-  private static MemorySegment createPinnedBuffer(long bufferBytes) {
+  private static MemorySegment createPinnedBuffer() {
     try (var localArena = Arena.ofConfined()) {
       MemorySegment pointer = localArena.allocate(C_POINTER);
-      checkCudaError(cudaMallocHost(pointer, bufferBytes), "cudaMallocHost");
+      checkCudaError(cudaMallocHost(pointer, PinnedMemoryBuffer.CHUNK_BYTES), "cudaMallocHost");
```
mythrocks, line 24: Please correct me if I'm wrong.  

There were a couple of advantages to the previous approach that now seem to be missing:
1. If the total number of bytes did not exceed `CHUNK_BYTES`, we would make a smaller allocation.  By removing this check, we're guaranteeing 8MB allocation per thread just for pinned memory.  Is that wise?
2. There was protection against a very large row-size (8MB).  Now we ignore the possibility.  Wouldn't that mean a buffer-overrun when a single row is copied?  Or is that not deemed possible now?

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/CuVSResourcesImpl.java
```diff
@@ -65,11 +61,25 @@ public void close() {
...
   }
 
   @Override
   public Path tempDirectory() {
     return tempDirectory;
   }
+
+  public static MemorySegment getHostBuffer(ScopedAccess access) {
+
+    while (access instanceof DelegatingScopedAccess delegatingScopedAccess) {
```
mythrocks, line 75: TIL.  I wasn't aware of this idiom.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/ScopedAccessWithHostBuffer.java
```diff
@@ -0,0 +1,41 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package com.nvidia.cuvs.internal;
+
+import com.nvidia.cuvs.CuVSResources;
+import java.lang.foreign.MemorySegment;
+
+class ScopedAccessWithHostBuffer implements CuVSResources.ScopedAccess {
```
mythrocks, line 10: Should we consider renaming this class to `ScopedAccessWithPinnedMemoryBuffer`?

mythrocks: Hmm.  I see that the `PinnedMemoryBuffer` used to construct this is actually part of `CuvsResourcesImpl`.  And this class is only used with pinned-memory.

I'm trying to work out if there's a different way of arranging this.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/common/PinnedMemoryBuffer.java
ldematte: > If the total number of bytes did not exceed CHUNK_BYTES, we would make a smaller allocation. By removing this check, we're guaranteeing 8MB allocation per thread just for pinned memory. Is that wise?

We discussed this with @achirkin and think this is worth it. It's true we are allocating 8MB per-resource, but it's also true we do this on-demand, and the benefit of not having to re-allocate every time (which is costly, due to the lock on CUDA context) is worth it.

> There was protection against a very large row-size (8MB). Now we ignore the possibility. Wouldn't that mean a buffer-overrun when a single row is copied? Or is that not deemed possible now?

It's true. It's very unlikely, I would say impossible in our scenarios, but it's something missing. Let me add this protection back (although it will need to be in a different place)

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/ScopedAccessWithHostBuffer.java
ldematte: I'm OK with renaming this class.

>I see that the `PinnedMemoryBuffer` used to construct this is actually part of `CuvsResourcesImpl`. And this class is only used with pinned-memory. 

I agree this is probably not ideal, but it's the best compromise I could think of.
Alternatives I considered and discarded:
- the `PinnedMemoryBuffer` is held by `ScopedAccessWithHostBuffer` directly ; that would require explicit cleanup. Also, it implies that we create just one ScopedAccess per `CuvsResourcesImpl`, which is _currently_ the case, but might change in the future
- the `PinnedMemoryBuffer` is held by the resources, and it's _also_ accessed directly from the resources.

I prefer to have the `PinnedMemoryBuffer` owned by `CuvsResourcesImpl`, to make it clear it's one buffer-per-resource and that its lifetime is tied to the resources.
And I prefer to have it accessed via the `ScopedAccess`, to make it consistent and explicitly "controlled" (no access to the resources directly means these are controlled, e.g. that you cannot access them from multiple threads, which is why we introduced it and it is definitely needed here).

One possible improvement is to move these classes: `PinnedMemoryBuffer` can be moved to this package and made package-private; we can even go one step further and make both `ScopedAccessWithHostBuffer` and `PinnedMemoryBuffer` private static classes inside `CuvsResourcesImpl`. I do like the encapsulation, but it could bloat `CuvsResourcesImpl` and make it difficult to read.

But if you prefer that or find a better way, let me know :)

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/CuVSResourcesImpl.java
ldematte: This whole pattern is inspired by Lucene, which has the same problem of not exposing some implementation classes (e.g. pre-Java 22 FFI classes) externally, but the need to access them internally. So we have a "Access" interface (that we can inspect and cast to) that tags a class with a capability (in this case, the presence of a memory segment that can be used as a buffer).

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/ScopedAccessWithHostBuffer.java
mythrocks: >  I prefer to have the `PinnedMemoryBuffer` owned by `CuvsResourcesImpl`, to make it clear it's one buffer-per-resource **_and that its lifetime is tied to the resources_**.

This is a good point.  I agree with this arrangement.

mythrocks:
I've rebased this PR and resolved a conflict.  
But it looks like we're going to need to fix the copyright headers on the changed files.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/common/PinnedMemoryBuffer.java
mythrocks: I'm 👍 on the change so far.  

> It's very unlikely, I would say impossible in our scenarios, but it's something missing. Let me add this protection back...

I'll take another look when this changes.

ldematte: Yes please :) I'll ping you when I'll push the change!

mythrocks:
> /ok to test [d792fce](https://github.com/rapidsai/cuvs/commit/d792fce7e25716b3363cb9f72a6a1d8294728959)

@benfred:  This PR will likely continue to fail CI until the copyright headers are fixed for this change.  Plus, there's another change pending (for handling cases where the row is super-large).

benfred:
@mythrocks the copyright issue should be fixed with c4f0570 (but CI is still failing here on `CagraMultiThreadStabilityIT.testQueryingUsingMultipleThreadsWithPrivateResources:62->testQueryingUsingMultipleThreads:162 MultiThreaded stablity test failed: null` errors in the java unittests =(

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/common/PinnedMemoryBuffer.java
ldematte: @mythrocks I have added back the protection against a very large (> buffer) row-size (8MB).
See CuVSDeviceMatrixImpl and JDKProvider for the 2 different strategies

ldematte:
@benfred @mythrocks I fixed the test to address the issue

I also addressed the issue with "too large rows" that won't fit in the buffer, falling back to a strategy that does direct copies (no buffering) in those cases (one for device -> host, the other for host -> device).

ldematte:
@mythrocks I have rebased this one onto release/25.12 too

mythrocks:
CI failures are a result of https://github.com/rapidsai/raft/pull/2813.  Should be resolved by https://github.com/rapidsai/raft/pull/2881.

java/cuvs-java/src/main/java/com/nvidia/cuvs/DelegatingScopedAccess.java
```diff
@@ -0,0 +1,29 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2025, NVIDIA CORPORATION.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+package com.nvidia.cuvs;
+
+public final class DelegatingScopedAccess implements CuVSResources.ScopedAccess {
```
mythrocks, line 7: It would be good to add a javadoc here to describe the purpose of this class.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/CuVSDeviceMatrixImpl.java
```diff
@@ -15,15 +15,55 @@
...
+  private class DirectRowAccessStrategy implements RowAccessStrategy {
+    @Override
+    public RowView getRow(long row) {
+      try (var access = resources.access()) {
+        var memorySegment = Arena.ofAuto().allocate(size * valueByteSize);
+        populateBuffer(access, row, row + 1, memorySegment);
+        return new SliceRowView(memorySegment, columns, valueLayout, dataType, valueByteSize);
+      }
+    }
+  }
```
mythrocks, line 66: 👏

Issue #1265, opened 2025-08-19
Implement cuvs::neighbors::hnsw::to_cagra_param function in cuvs/java


mayya-sharipova:
Expose the function that converts HNSW params to Cagra params in cuvs/java

```cpp
cagra::index_params cuvs::neighbors::hnsw::to_cagra_params(
    raft::matrix_extent<int64_t> dataset,
    int M,
    int ef_construction,
    cuvs::distance::DistanceType metric = cuvs::distance::DistanceType::L2Expanded);
```

PR #1448, opened 2025-10-22
Improved CAGRA build parameter heuristics

achirkin:
Changes to the build parameter heuristics:
  - Move the code from HNSW namespace to CAGRA namespace to avoid depending on HNSW target
  - Add one more variant of the heuristics: allow generating smaller graph to better match the performance of the HNSW-generated graph
  - Implement automatic switch between NN-Descent and IVF-PQ as the graph-build algorithms depending on the dataset size: NN-Descent tends to perform better on smaller-scale datasets

PR also include C and java bindings.
Resolves https://github.com/rapidsai/cuvs/issues/1265

### DIFF
--- a/c/CMakeLists.txt
+++ b/c/CMakeLists.txt
@@ -99,6 +99,17 @@ set_target_properties(
              EXPORT_NAME c_api
 )
 
+# Setup the VERSION and SOVERSION of our library
+#
+# Long term goal is that every release that breaks the C ABI will update `cuvs_abi_version` allowing
+# us to have multiple releases of libcuvs_c that don't need consumers to rebuild against
+if(PROJECT_IS_TOP_LEVEL)
+  set(cuvs_abi_version "1")
+  set(cuvs_version "${cuvs_abi_version}.${RAPIDS_VERSION}")
+  set_target_properties(cuvs_c PROPERTIES VERSION "${cuvs_version}")
+  set_target_properties(cuvs_c PROPERTIES SOVERSION "${cuvs_abi_version}")
+endif()
+
 # Encode what optional components we are building
 set(CUVS_BUILD_CAGRA_HNSWLIB ${BUILD_CAGRA_HNSWLIB})
 set(CUVS_BUILD_MG_ALGOS ${BUILD_MG_ALGOS})
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -349,6 +349,7 @@ if(NOT BUILD_CPU_ONLY)
     src/cluster/kmeans_transform_double.cu
     src/cluster/kmeans_transform_float.cu
     src/cluster/single_linkage_float.cu
+    src/cluster/spectral.cu
     src/core/bitset.cu
     src/core/omp_wrapper.cpp
     src/distance/detail/kernels/gram_matrix.cu
--- a/cpp/cmake/patches/faiss/refactor-rmm-memory-resources.patch
+++ b/cpp/cmake/patches/faiss/refactor-rmm-memory-resources.patch
@@ -1,0 +1,105 @@
+From 0aa03d6eaf4bacc4d669254499966d892453961d Mon Sep 17 00:00:00 2001
+From: Bradley Dice <bdice@bradleydice.com>
+Date: Wed, 29 Oct 2025 12:37:57 -0500
+Subject: [PATCH] Refactor RMM memory resources
+
+---
+ faiss/gpu/StandardGpuResources.cpp | 14 +++++++-------
+ faiss/gpu/StandardGpuResources.h   |  7 ++++---
+ 2 files changed, 11 insertions(+), 10 deletions(-)
+
+diff --git a/faiss/gpu/StandardGpuResources.cpp b/faiss/gpu/StandardGpuResources.cpp
+index 649b7cb5c..765fdb3d0 100644
+--- a/faiss/gpu/StandardGpuResources.cpp
++++ b/faiss/gpu/StandardGpuResources.cpp
+@@ -24,8 +24,8 @@
+ #if defined USE_NVIDIA_CUVS
+ #include <raft/core/device_resources.hpp>
+ #include <rmm/mr/device/managed_memory_resource.hpp>
++#include <rmm/mr/pinned_host_memory_resource.hpp>
+ #include <rmm/mr/device/per_device_resource.hpp>
+-#include <rmm/mr/host/pinned_memory_resource.hpp>
+ #include <memory>
+ #endif
+
+@@ -93,7 +93,7 @@ StandardGpuResourcesImpl::StandardGpuResourcesImpl()
+         :
+ #if defined USE_NVIDIA_CUVS
+           mmr_(new rmm::mr::managed_memory_resource),
+-          pmr_(new rmm::mr::pinned_memory_resource),
++          pmr_(new rmm::mr::pinned_host_memory_resource),
+ #endif
+           pinnedMemAlloc_(nullptr),
+           pinnedMemAllocSize_(0),
+@@ -164,7 +164,7 @@ StandardGpuResourcesImpl::~StandardGpuResourcesImpl() {
+
+     if (pinnedMemAlloc_) {
+ #if defined USE_NVIDIA_CUVS
+-        pmr_->deallocate(pinnedMemAlloc_, pinnedMemAllocSize_);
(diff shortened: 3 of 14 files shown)


c/include/cuvs/neighbors/cagra.h
```diff
@@ -156,6 +156,48 @@ cuvsError_t cuvsCagraCompressionParamsCreate(cuvsCagraCompressionParams_t* param
...
+ *
+ * @param[out] params The CAGRA index params to populate
+ * @param[in] n_rows Number of rows in the dataset
+ * @param[in] dim Number of dimensions in the dataset
+ * @param[in] M HNSW index parameter M
+ * @param[in] ef_construction HNSW index parameter ef_construction
+ * @param[in] metric Distance metric to use
+ * @return cuvsError_t
+ */
+cuvsError_t cuvsCagraIndexParamsFromHnswHardM(cuvsCagraIndexParams_t params,
```
cjnolet: Instead of introducing two new functions in the public APIs, can we just make the "hard" vs "soft" a user option that the user can set?

ldematte: Personally I prefer the 2 distinct functions, but I think this is going to be a very personal preference :)

achirkin: @cjnolet sure it's trivial and not a strong preference, so I'm happy change it to one function if you'd like.
However, I made it into two functions in the originally because I thought it would give more clarity to a user: it's clear that the two heuristics are sufficiently different, each can have its own documentation, and we have a room to add more/alternative heuristics in future.
So, shall I change it?

c/include/cuvs/neighbors/cagra.h
```diff
@@ -156,6 +156,48 @@ cuvsError_t cuvsCagraCompressionParamsCreate(cuvsCagraCompressionParams_t* param
  */
 cuvsError_t cuvsCagraCompressionParamsDestroy(cuvsCagraCompressionParams_t params);
 
+/**
+ * @brief Create CAGRA index parameters similar to an HNSW index with hard M constraint
+ *
+ * This factory function creates CAGRA parameters that yield a graph of the same size
```
tfeher: Is it approximately same size in `bytes`, or logically same size (M==graph_degree/2)?

c/include/cuvs/neighbors/cagra.h
tfeher: On the C++ side, I think from a user perspective it is better to have a single function with a selector arg that has a default value. For the C API I am not sure.

cpp/src/neighbors/cagra.cpp
```diff
@@ -0,0 +1,71 @@
...
+  return params;
+}
+
+cagra::index_params index_params::from_hnsw_soft_m(raft::matrix_extent<int64_t> dataset,
+                                                   int M,
+                                                   int ef_construction,
+                                                   cuvs::distance::DistanceType metric)
+{
+  cagra::index_params params;
+  params.graph_degree              = 2 + M * 2 / 3;
```
tfeher: The hard am variant sets `graph_degree =  2 * M`, it is surprising to see that soft variant can lead to similar search performance with `graph_degree < M`. The benchmarks for 768 and 1536 dimension looked good. Was it also tested for smaller dimensional datasets?

tfeher (approved):
Thanks Artem for the PR, it looks good to me!

cpp/src/neighbors/cagra.cpp
achirkin: Yes, I've tested it on DEEP-100M and glove datasets. The hard-M variant actually shows much higher recall and lower throughput for the same search 'ef' parameter (the QPS-recall curve is close to HNSW, but all points on it are 'shifted' towards higher recall and lower throughput).

c/include/cuvs/neighbors/cagra.h
achirkin: Ok I went on with merging the two functions with a enum parameter

c/src/neighbors/cagra.cpp
```diff
@@ -665,6 +665,80 @@ extern "C" cuvsError_t cuvsCagraCompressionParamsDestroy(cuvsCagraCompressionPar
   return cuvs::core::translate_exceptions([=] { delete params; });
 }
 
+// Helper function to populate C IVF-PQ params from C++ params
+static void populate_c_ivf_pq_params(cuvsIvfPqParams* c_ivf_pq,
```
robertmaynard: This should be in an unnamed namespace

robertmaynard (changes requested):

java/cuvs-java/src/main/java/com/nvidia/cuvs/spi/CuVSProvider.java
```diff
@@ -151,4 +151,12 @@ default CagraIndex mergeCagraIndexes(CagraIndex[] indexes, CagraMergeParams merg
   static CuVSProvider provider() {
     return CuVSServiceProvider.Holder.INSTANCE;
   }
+
+  CagraIndexParams cagraIndexParamsFromHnswParams(
```
mythrocks, line 171: Nit:  For the new API function, it would be good to add a brief javadoc, with `@param` descriptions.

java/cuvs-java/src/main/java/com/nvidia/cuvs/CagraIndexParams.java
```diff
@@ -45,6 +51,61 @@ public enum CagraGraphBuildAlgo {
     private CagraGraphBuildAlgo(int value) {
       this.value = value;
     }
+
+    private static final Map<Integer, CagraGraphBuildAlgo> VALUES =
+        Arrays.stream(CagraGraphBuildAlgo.values())
+            .collect(Collectors.toUnmodifiableMap(x -> x.value, Function.identity()));
+
```
mythrocks, line 58: Yep, nice phrasing.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/spi/JDKProvider.java
```diff
@@ -370,8 +478,10 @@ public void addVector(int[] vector) {
   }
 
   /**
-   * This {@link CuVSDeviceMatrix} builder implementation returns a {@link CuVSDeviceMatrix} backed by managed RMM
-   * device memory. It uses a {@link PinnedMemoryBuffer} to batch data before copying it to the GPU.
+   * This {@link CuVSDeviceMatrix} builder implementation returns a
+   * {@link CuVSDeviceMatrix} backed by managed RMM
+   * device memory. It uses a {@link PinnedMemoryBuffer} to batch data before
+   * copying it to the GPU.
```
mythrocks: Sorry, why did this need reflowing?

robertmaynard (approved):

c/include/cuvs/neighbors/cagra.h
```diff
@@ -37,6 +37,40 @@ enum cuvsCagraGraphBuildAlgo {
...
+   * graph (2M being the max graph degree) and CAGRA produces a fixed-degree graph, there's always a
+   * difference in the performance of the two.
+   *
+   * This function attempts to produce such a graph that the QPS and recall of the two graphs being
+   * searched by HNSW are close for any search parameter combination. The CAGRA-produced graph tends
+   * to have a "longer tail" on the low recall side (that is being slightly faster and less
+   * precise).
+   *
+   */
+  SIMILAR_SEARCH_PERFORMANCE = 0,
```
benfred: For new enums in C - can we prefix the values with something like `CUVS_CAGRA_HEURISTIC_TYPE` or something similar? 

I realize its verbose (and we haven't been doing this consistently as of now), but with this change the symbol `SIMILAR_SEARCH_PERFORMANCE` goes into the global namespace and we could get namespace collisions

java/cuvs-java/src/main/java22/com/nvidia/cuvs/spi/JDKProvider.java
achirkin: Sorry must have C++ auto-formatting accidentally applied.

java/cuvs-java/src/main/java/com/nvidia/cuvs/CagraIndexParams.java
achirkin: @ldematte 👍

c/include/cuvs/neighbors/cagra.h
achirkin: Good point, totally agree!

PR #1459, opened 2025-10-24
[BUG] Check if dynamic batching conf is null

tarang-jain:
This PR adds pytests for `cuvs_ivf_pq` algorithm in cuvs-bench.

Bug explanation: CAGRA with IVF-PQ as the `graph_build_algo` gives an error when dynamic batching params are not set for IVF-PQ:
`ERROR OCCURRED: '[json.exception.type_error.306] cannot use value() with null'`
Ths PR fixes that and also adds pytests for `cuvs_ivf_pq` algorithm in cuvs-bench.


achirkin (approved):
Hi @tarang-jain thanks for the fix! I've seen the offline discussion of the issue, but could you please add an explanation of the issue this PR solves in the PR description? This helps a lot when searching for changes in the commit history.

tarang-jain:
@achirkin done! I updated the PR description.

PR #1460, opened 2025-10-24
[FEA] Add missing pytests for cuvs-bench algorithms

tarang-jain:


cjnolet (approved):

cpp/src/neighbors/detail/vamana/vamana_serialize.cuh
```diff
@@ -65,13 +65,20 @@ void serialize_dataset(raft::resources const& res,
...
-                 raft::make_host_vector_view(h_dataset.data_handle(),
-                                             strided_dataset->n_rows() * strided_dataset->dim()),
-                 raft::make_device_vector_view(strided_dataset->view().data_handle(),
-                                               strided_dataset->n_rows() * strided_dataset->dim()));
+      auto nrows     = strided_dataset->n_rows();
+      auto dim       = strided_dataset->dim();
+      auto stride    = strided_dataset->stride();
+      auto d_data    = strided_dataset->view();
+      auto h_dataset = raft::make_host_matrix<T, int64_t>(nrows, dim);
+      RAFT_CUDA_TRY(cudaMemcpy2DAsync(h_dataset.data_handle(),
```
bkarsin: Preferable to use `raft::copy` to direct cudaMemcpy calls?

bkarsin (commented):
Looks good to me, just the 1 note about using `raft::copy` vs. cudaMemcpy.

cjnolet:
Correct- we shoukd not be directly calling cudaMemCpy anywhere. (There are
a few places where we do it,  but those need to be corrected, and not
continue to use it.

cpp/src/neighbors/detail/vamana/vamana_serialize.cuh
tarang-jain: That is a strided copy (cudaMemcpy2DAsync). We currently do not have a raft wrapper for it. We are calling this directly in other places of the codebase too.

cjnolet:
Please make a wrapper for this.

PR #1462, opened 2025-10-24
Migrate hash strategy to use the new cuco::static_map

PointKernel:
This PR migrates usage from the legacy static_map to the new design, with no algorithmic changes.

### DIFF
--- a/go/brute_force/brute_force.go
+++ b/go/brute_force/brute_force.go
@@ -69,7 +69,7 @@ func BuildIndex[T any](Resources cuvs.Resource, Dataset *cuvs.Tensor[T], metric 
 // * `queries` - Tensor in device memory to query for
 // * `neighbors` - Tensor in device memory that receives the indices of the nearest neighbors
 // * `distances` - Tensor in device memory that receives the distances of the nearest neighbors
-func SearchIndex[T any](resources cuvs.Resource, index BruteForceIndex, queries *cuvs.Tensor[T], neighbors *cuvs.Tensor[int64], distances *cuvs.Tensor[T]) error {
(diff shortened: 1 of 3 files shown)


cpp/src/distance/detail/sparse/coo_spmv_strategies/hash_strategy.cuh
```diff
@@ -32,11 +36,19 @@ namespace sparse {
...
-  using find_type =
-    typename cuco::legacy::static_map<value_idx, value_t, cuda::thread_scope_block>::device_view;
+  static constexpr value_idx empty_key_sentinel = value_idx{-1};
+  static constexpr value_t empty_value_sentinel = value_t{0};
+  using probing_scheme_type = cuco::linear_probing<1, cuco::murmurhash3_32<value_idx>>;
+  using storage_ref_type = cuco::bucket_storage_ref<cuco::pair<value_idx, value_t>, 1, cuco::extent<int>>;
+  using map_type = cuco::static_map_ref<
+                                       value_idx,
+                                       value_t,
+                                       cuda::thread_scope_block,
```
divyegala: For this scope, is it okay to use triple chevron launcher still or we must migrate to `cudaLaunchCooperativeKernel` any time we use cooperative groups?

PointKernel: <<<>>> is fine here: no grid‑wide cooperative sync, only block‑local use of cooperative groups. `thread_scope_block` is correct because the map lives in shared memory per block.

divyegala (approved):

PR #1465, opened 2025-10-25
Use an explicit std::min for byte alignment calculation

maxwbuckley:
Use explicit std::min removing compiler ambiguity. This is imported at the top of the file "<algorithm>" and used elsewhere in this file

### DIFF
--- a/cpp/cmake/thirdparty/get_cutlass.cmake
+++ b/cpp/cmake/thirdparty/get_cutlass.cmake
@@ -31,19 +31,17 @@ function(find_and_configure_cutlass)
   include("${rapids-cmake-dir}/cpm/package_override.cmake")
   rapids_cpm_package_override("${CMAKE_CURRENT_FUNCTION_LIST_DIR}/../patches/cutlass_override.json")
 
-  include("${rapids-cmake-dir}/cpm/detail/package_details.cmake")
-  rapids_cpm_package_details(cutlass version repository tag shallow exclude)
-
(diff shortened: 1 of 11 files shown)


cpp/src/distance/detail/pairwise_matrix/dispatch_layout.cuh
```diff
@@ -57,7 +57,7 @@ int determine_vec_len(pairwise_matrix_params<IdxT, DataT, OutT, FinOpT> params)
 {
   size_t align_x        = alignment_of_2d_array(params.x, params.ldx);
   size_t align_y        = alignment_of_2d_array(params.y, params.ldy);
-  size_t byte_alignment = min(align_x, align_y);
+  size_t byte_alignment = std::min(align_x, align_y);
```
mythrocks, line 49: Silly question:  Should we consider converting this to `cuda::std::min` instead?

mythrocks: (I ask because this is function template in a `.cuh` file.)

tarang-jain: Even though its a `.cuh` file, the function itself is still a purely host function, so that is not necessary.

mythrocks: Ok, thanks for confirming that this template isn't used in `__device__` code. 👍

cjnolet:
Thanks so much for helping us improve the codebase @maxwbuckley. These changes LGTM!

PR #1473, opened 2025-10-29
Fix binary quantizer host transform bounds and stream order

achirkin:
Fix two issues with the binary quantizer's host transform:

1. Synchronize the stream after async GPU operations and before a host-only omp loop;
2. Fill the bytes between `minimul_out_dim` and `out_dim` with zeroes to match device transform behavior.

This PR is a prerequisite to merging https://github.com/rapidsai/raft/pull/2835 (otherwise the tests fail due to host mdarray not initializing with zeroes anymore).


lowener (approved):
LGTM

PR #1480, opened 2025-10-30
Add libaio to cuvs-bench-cpu, improve bench dependencies

tarang-jain:
libaio is used directly at run time by the diskann_ssd benchmarking wrapper in cuvs-bench-cpu.

### DIFF
--- a/conda/environments/bench_ann_cuda-129_arch-aarch64.yaml
+++ b/conda/environments/bench_ann_cuda-129_arch-aarch64.yaml
@@ -5,7 +5,6 @@ channels:
 - rapidsai-nightly
 - conda-forge
 dependencies:
-- benchmark>=1.8.2
 - c-compiler
 - clang-tools==20.1.4
 - clang==20.1.4
--- a/conda/environments/bench_ann_cuda-129_arch-x86_64.yaml
+++ b/conda/environments/bench_ann_cuda-129_arch-x86_64.yaml
@@ -5,7 +5,6 @@ channels:
 - rapidsai-nightly
 - conda-forge
 dependencies:
-- benchmark>=1.8.2
 - c-compiler
 - clang-tools==20.1.4
 - clang==20.1.4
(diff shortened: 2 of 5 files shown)


conda/recipes/cuvs-bench/recipe.yaml
```diff
@@ -57,13 +57,14 @@ requirements:
     - rmm =${{ minor_version }}
     - scikit-learn>=1.5
     - if: linux64
-      then: mkl =2023
+      then:
+        - libaio  # Required for diskann_ssd at runtime
```
bdice: Is `libaio` also needed to build the package? Do we need an entry in `host` for that?

The entry in `ignore_run_exports` wasn't doing anything because no run exports were being added. Run exports would only be present if `libaio` was in `host` to begin with.

bdice (changes requested):

conda/recipes/cuvs-bench/recipe.yaml
tarang-jain: Its needed to build DiskANN and we are directly linking the wrapper to it here:
https://github.com/rapidsai/cuvs/blob/d8fdd7d460f8793bbbf692200b0b68028ae49975/cpp/bench/ann/CMakeLists.txt#L336

tarang-jain: Looks like libaio is in `host` for `cuvs-bench-cpu`, right?

bdice (approved):
Discussed with @tarang-jain over a call. We found some other improvements to conda recipes:
- We can remove all C++ dependencies in `cuvs-bench` recipe because it is pure Python. Those dependencies are already coming from `libcuvs-bench-ann` and don't need to be respecified in Python.
- We don't need to use `benchmark` conda packages anywhere anymore, we always use CPM to [fetch and statically link Google Benchmark here](https://github.com/rapidsai/cuvs/blob/c9072594391272bf2aeb0b3bb8330a75f6d5a4c9/cpp/CMakeLists.txt#L199-L200) (we've already made this change elsewhere in RAPIDS)
- Small fix to `get_nlohmann_json.cmake` -- the log message was for `glog` by mistake

PR #1483, opened 2025-10-31
[FEA] IVF-PQ Build Factories for Precomputed Centroids and Codebooks

tarang-jain:
IVF-PQ Index Build API Enhancements and PIMPL Refactoring
## Summary
This PR adds new build APIs for IVF-PQ indices using precomputed centroids and implements a complete PIMPL refactoring with owning/view semantics
## Key Changes
1. New Build APIs for Precomputed Centroids
Added `cuvs::neighbors::ivf_pq::build()` overloads that accept precomputed cluster centroids, PQ codebooks, and rotation matrices
Enables building indices from pre-trained models without re-training
Supports both device and host input data with automatic memory transfer
2. PIMPL Refactoring with Owning/View Semantics
`owning_impl`: Owns centroid and codebook data (traditional behavior)
`view_impl`: References external centroid data without copying
Maintains identical search behavior with zero data copying
3. Enhanced Helper Functions
Removed mutator functions that directly modify the state of the index. Instead we have helpers for the user to fetch and own the transformed data

View indices avoid copying large centroid arrays
Backward Compatibility: All existing APIs work unchanged

[Updates] (as of 12/23/2025):
The non-const getters that allow direct modification of the state of the index:
`pq_centers()`, `centers()`, `centers_rot()` and `rotation_matrix()` have been removed from the interface and the user-facing class.


cpp/include/cuvs/neighbors/ivf_pq.hpp
```diff
@@ -23,6 +23,13 @@
 
 namespace cuvs::neighbors::ivf_pq {
 
+template <typename IdxT>
```
cjnolet: One of the benefits to this design is that the only thing the users should have to know about in the public APIs is the `pimpl` class. That's what decouples all of the internals and allows us to code the headers to a common interface contract. 

Given that- I think we should move the implementations for the impls (and iface), as well as `build_owning` and `build_view` into the source files / detail namespace, because the user shouldn't need to know they exist.

cpp/include/cuvs/neighbors/ivf_pq.hpp
```diff
@@ -2877,6 +3182,39 @@ void set_centers(raft::resources const& res,
...
+ * Note: This function requires the index to be empty (no data added yet).
+ * The centers will be copied to device memory and the rotated centers
+ * will be computed if a rotation matrix exists.
+ *
+ * @param[in] res raft resources handle
+ * @param[inout] index pointer to the IVF-PQ index
+ * @param[in] cluster_centers new cluster centers on host memory [n_lists, dim] or [n_lists,
+ * dim_ext]
+ */
+void set_centers(raft::resources const& res,
```
cjnolet: I think the `build()` that allows them to specify the centers and codebooks should be enough, right? Is this something we expect users to call directly? If not, it shouldn't be in the public API header. It's very important that we make this distinction, especially as we're gearing up to start stabilizing ABI at the C layer- we need to treat everythign in the public API headers as being something we'd expect users to invoke. Any internal code should go into source files or at least be placed in `src/`.

tarang-jain: Yeah that makes sense. The only situation it would not make sense is when the user has custom ivf centroids but does not have custom pq centroids.

cjnolet: Can we have another overload for the `build()` that accepts centroids but not codebooks? Or would that make things too complicated w/ the number of underlying impl classes (that'd probably require another impl with centroids as an mdarray and everything else as a view). 

I can't immediately think of why a user would want this, though. I think if they are wanting to use IVFPQ to "transform()" vectors, they'd want both the coarse and fine populated. I say for now we just skip this and implement it if users ultimate end up asking for it.

tarang-jain: > I can't immediately think of why a user would want this, though.

That is actually true. I was thinking that Faiss needs to use it, but Faiss will also set its own pq centroids so it can very well use such a build API that allows it to set both.

So I'll just remove set_centers from the public APIs and simply use the build API.

> (that'd probably require another impl with centroids as an mdarray and everything else as a view)

Thats not how I am doing it currently. Either we don't own anything (centroids, pq centers and rotation matrix) or we own everything if even one of those is missing or of inconsistent extents. on the other hand if by 'another impl' you were referring to another build API and not another impl class, I can just make pq centers optional in the current build API.

cpp/include/cuvs/neighbors/ivf_pq.hpp
```diff
@@ -996,6 +1066,129 @@ void build(raft::resources const& handle,
...
+ * All parameters must be provided with correct extents. The caller is responsible for ensuring
+ * the lifetime of the input data exceeds the lifetime of the returned index.
+ *
+ * The index_params must be consistent with the provided matrices. Specifically:
+ * - index_params.codebook_kind determines the expected shape of pq_centers
+ * - index_params.metric will be stored in the index
+ * - index_params.conservative_memory_allocation will be stored in the index
+ * The function will verify consistency between index_params, dim, and the matrix extents.
+ *
+ * @tparam IdxT Type of indices (default: int64_t)
```
jinsolp: Believe we don't have an `IdxT` type here

cpp/include/cuvs/neighbors/ivf_pq.hpp
```diff
@@ -996,6 +1066,129 @@ void build(raft::resources const& handle,
...
```
jinsolp: same as above

.devcontainer/Dockerfile
```diff
@@ -45,7 +42,6 @@ ENV HISTFILE="/home/coder/.cache/._bash_history"
 ENV AWS_ROLE_ARN="arn:aws:iam::279114543810:role/nv-gha-token-sccache-devs"
 ENV SCCACHE_REGION="us-east-2"
 ENV SCCACHE_BUCKET="rapids-sccache-devs"
-ENV SCCACHE_S3_USE_PREPROCESSOR_CACHE_MODE=true
```
cjnolet, line 48: Seems like we have some additional changes in here?

cpp/include/cuvs/neighbors/ivf_pq.hpp
```diff
@@ -278,6 +278,72 @@ constexpr typename list_spec<SizeT, IdxT>::list_extents list_spec<SizeT, IdxT>::
...
+  virtual uint32_t get_list_size_in_bytes(uint32_t label) const noexcept = 0;
+
+  virtual std::vector<std::shared_ptr<list_data<IdxT>>>& lists() noexcept             = 0;
+  virtual const std::vector<std::shared_ptr<list_data<IdxT>>>& lists() const noexcept = 0;
+
+  virtual raft::device_vector_view<uint32_t, uint32_t, raft::row_major> list_sizes() noexcept = 0;
+  virtual raft::device_vector_view<const uint32_t, uint32_t, raft::row_major> list_sizes()
+    const noexcept = 0;
+
+  virtual raft::device_vector_view<uint8_t*, uint32_t, raft::row_major> data_ptrs() noexcept = 0;
```
cjnolet, line 308: Whya re all the possible type comtinations separate virtuals? Are we not able to have templates for these? This seems unnecessary to have these defined on the iface for all types at the same time.

.devcontainer/Dockerfile
tarang-jain: I actually verified that this file is consistent with release/25.12 as well as with main. Not sure why git is showing this diff.

cpp/include/cuvs/neighbors/ivf_pq.hpp
```diff
@@ -996,6 +1066,129 @@ void build(raft::resources const& handle,
            const cuvs::neighbors::ivf_pq::index_params& index_params,
            raft::host_matrix_view<const uint8_t, int64_t, raft::row_major> dataset,
            cuvs::neighbors::ivf_pq::index<int64_t>* idx);
+
+/**
+ * @brief Build a view-type IVF-PQ index from device memory centroids and codebook.
+ *
+ * This function creates a non-owning index that references the provided device data directly.
```
cjnolet:
```suggestion
 * This function creates a non-owning index that stores a reference to the provided device data.
```

.devcontainer/Dockerfile
jinsolp: have you rebased this branch?

cpp/include/cuvs/neighbors/ivf_pq.hpp
tarang-jain: Added the missing docs

tarang-jain:
/ok to test https://github.com/rapidsai/cuvs/commit/8d4998e3f7efa4a5a3968adc1df9967d5c6dd27b

cpp/src/neighbors/ivf_pq_index.cu
```diff
@@ -14,6 +15,349 @@
...
+template <typename IdxT>
+raft::device_matrix_view<const float, uint32_t, raft::row_major>
+owning_impl<IdxT>::rotation_matrix() const noexcept
+{
+  return rotation_matrix_.view();
+}
+
+template <typename IdxT>
+raft::device_mdspan<float, pq_centers_extents, raft::row_major>
+view_impl<IdxT>::pq_centers() noexcept
```
lowener: On the `view_impl` class I think the non-const accessor should be disabled: The user is passing a const array and we shouldn't remove the constness to modify it. This would apply for pq_centers, centers, centers_rot, ...

cpp/include/cuvs/neighbors/ivf_pq.hpp
```diff
@@ -996,6 +1067,124 @@ void build(raft::resources const& handle,
...
+ * dim]
+ *
+ * @return A view-type ivf_pq index that references the provided data
+ */
+auto build(raft::resources const& handle,
+           const cuvs::neighbors::ivf_pq::index_params& index_params,
+           const uint32_t dim,
+           raft::device_mdspan<const float, raft::extent_3d<uint32_t>, raft::row_major> pq_centers,
+           raft::device_matrix_view<const float, uint32_t, raft::row_major> centers,
+           raft::device_matrix_view<const float, uint32_t, raft::row_major> centers_rot,
```
lowener, line 1095: Why is `centers_rot` and `rotation_matrix` optional only in the host API?

cpp/include/cuvs/neighbors/ivf_pq.hpp
```diff
@@ -2847,35 +3037,50 @@ void make_rotation_matrix(raft::resources const& res,
...
+ * extended centers suitable for coarse search operations. The output has dimensions
+ * [n_centers, dim_ext] where dim_ext = round_up(dim + 1, 8).
  *
  * @param[in] res raft resource
- * @param[inout] index pointer to IVF-PQ index
- * @param[in] cluster_centers new cluster centers [index.n_lists(), index.dim()]
+ * @param[in] centers cluster centers [n_centers, dim]
+ * @param[out] padded_centers padded centers with norms [n_centers, dim_ext]
  */
-void set_centers(raft::resources const& res,
```
lowener, line 2874: This function getting removed might break downstream libs.
I checked milvus and didn't see it used there, but FAISS is using it: https://github.com/facebookresearch/faiss/blob/338a5fbf8b5aea3008fac68501956cc8a525a468/faiss/gpu/impl/CuvsIVFPQ.cu#L152.
Maybe it is worth to keep it for compatibility, or do the change in downstream libs.

cpp/src/neighbors/ivf_pq_impl.hpp
```diff
@@ -0,0 +1,188 @@
...
+ */
+
+#pragma once
+
+#include <cuvs/neighbors/ivf_pq.hpp>
+
+namespace cuvs::neighbors::ivf_pq {
+
+template <typename IdxT>
+class index_impl : public index_iface<IdxT> {
```
lowener, line 13: Shouldn't this class be abstract too, with a pure virtual function? It seems to be used as an interface, while the concrete classes are `owning_impl` and `view_impl`

cpp/include/cuvs/neighbors/ivf_pq.hpp
tarang-jain: yes that is the reason I have updated the Faiss patch in this PR, so it wont break our CI. It wont break the Faiss repo, because that is tied to cuvs=25.10 at the moment. I'll push the updates to Faiss to make this compatible when we are migrating Faiss to use cuvs=26.02.

cpp/src/neighbors/ivf_pq_impl.hpp
tarang-jain: This is an implementation class, which is not fully abstract, because it contains data_ptrs, list_sizes etc -- the methods that are common to both the view_impl and owning_impl. The distinction between view_impl and owning_impl is only limited to ownership of centroids and codebooks.

cpp/include/cuvs/neighbors/ivf_pq.hpp
tarang-jain: For the device side API we have to guarantee creation of the view_impl if the user provides device mdspans. That is the reason we have helpers such as pad_centers_with_norms which give the user the necessary tools to create the missing pieces. The idea is that if the user is supplying device_mdspans, they are responsible for the lifetime of those mdspans and we wont assume ownership of those things or create copies. We are also intentionally getting rid of mutators to change the state of the index directly -- if the user really wants to do that, they must assume ownership of the data and call the view-type build factory. This is in accordance with the design proposed by @cjnolet

cpp/src/neighbors/ivf_pq_index.cu
tarang-jain: The idea is for the interface (`index_iface`) to be purely virtual, so we must override those. Yes the user can use the non-const getters to intentionally modify the state of objects held by the index, but that kind of usage is discouraged. We want to encourage the use of only the factory APIs and helpers. Furthermore, we do need the non-const getters in some places for exactly this reason when the calling program cannot own those objects -- for example in Faiss. But that is 100% intentional and not meant for end users.

tarang-jain: So we are getting rid of helpers such as `set_centers` which directly modify the sate of the index.

lowener: I don't see how it is being discouraged right now. If it's not going to be disabled statically then add a warning on the non-const accessor before the const_cast, or/and write a comment about that in the documentation.
In the current version an end user could modify it's const data by mistake, without a warning or an error or any documentation on our side... And the only way to know about that would be to look at the implementation code

tarang-jain: > On the view_impl class I think the non-const accessor should be disabled: The user is passing a const array and we shouldn't remove the constness to modify it. This would apply for pq_centers, centers, centers_rot, ...

I can actually address this for the view_impl. I think your idea is correct that we should not const_cast a ptr passed by the user. What I was trying to say was more intended for the owning type.

tarang-jain: Fixed this in latest commit a09cbab. Marking as resolved.

cpp/include/cuvs/neighbors/ivf_pq.hpp
tarang-jain: This is done. Marking as resolved.

tarang-jain:
@lowener the current approach is causing test failures. By default, when doing something like index.pq_centers(), the non-const getters get called, and we throw errors there in the `view_impl`. Furthermore, I have had to change the noexcept status in the interface for the non-const getters. I personally don't see a problem with allowing the user to access and modify their own data. I think something like this:
`std::as_const(index).pq_centers()` would force the const getters to get called, but having to do this for only view type indexes would beat purpose of the polymorphism.

PR #1485, opened 2025-11-01
Introduce `libcuvs-static` as a separate conda package

divyegala:
This creates a new conda output for `libcuvs-static` by making `cuvs_static` target its own separate component in CMake. We noticed that `libcuvs_static.a` was being bundled along in the `libcuvs` conda package and making the package more inflated than necessary.

This PR also separates the installation of the C header files from the conditional `BUILD_C_LIBRARY` because the C++ targets (`cuvs` and `cuvs_static`) need the C headers to be installed whether the C target (`c_api`) is built or not.

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -746,24 +746,24 @@ target_compile_definitions(cuvs::cuvs INTERFACE $<$<BOOL:${CUVS_NVTX}>:NVTX_ENAB
     DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}
   )
 
-  if(BUILD_C_LIBRARY)
-    install(
-      DIRECTORY ../c/include/cuvs
-      COMPONENT cuvs
-      DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}
-    )
+  install(
+    DIRECTORY ../c/include/cuvs
+    COMPONENT cuvs
+    DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}
+  )
 
+  install(
+    FILES ${CUVS_C_BINARY_DIR}/include/cuvs/core/c_config.h
+    COMPONENT cuvs
+    DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/cuvs/core/
+  )
+
+  if(BUILD_C_LIBRARY)
     install(
       TARGETS cuvs_c
       DESTINATION ${lib_dir}
       COMPONENT c_api
       EXPORT cuvs-c-exports
-    )
-
-    install(
-      FILES ${CUVS_C_BINARY_DIR}/include/cuvs/core/c_config.h
-      COMPONENT cuvs
-      DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/cuvs/core/
     )
   endif()
 
@@ -773,6 +773,30 @@ target_compile_definitions(cuvs::cuvs INTERFACE $<$<BOOL:${CUVS_NVTX}>:NVTX_ENAB
       DESTINATION ${lib_dir}
       COMPONENT cuvs_static
       EXPORT cuvs-static-exports
+    )
+
+    install(
+      DIRECTORY include/cuvs
+      COMPONENT cuvs_static
+      DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}
+    )
+
+    install(
+      DIRECTORY ../c/include/cuvs
+      COMPONENT cuvs_static
+      DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}
+    )
+
+    install(
+      FILES ${CMAKE_CURRENT_BINARY_DIR}/include/cuvs/version_config.h
+      COMPONENT cuvs_static
+      DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/cuvs
+    )
+
+    install(
+      FILES ${CUVS_C_BINARY_DIR}/include/cuvs/core/c_config.h
+      COMPONENT cuvs_static
+      DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/cuvs/core/
     )
   endif()
 


cpp/CMakeLists.txt
```diff
@@ -746,23 +746,56 @@ target_compile_definitions(cuvs::cuvs INTERFACE $<$<BOOL:${CUVS_NVTX}>:NVTX_ENAB
     DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}
   )
 
+  install(
+    DIRECTORY ../c/include/cuvs
+    COMPONENT cuvs
+    DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}
+  )
+
+  install(
```
robertmaynard, line 755: This won't work if `BUILD_C_LIBRARY` is disabled as the file won't be generated.

Currently `BUILD_C_LIBRARY` is a false option, as we can't build without it. So we need to refactor the `c_config.h` genreation to a function and call that from the `cpp` CMakeLists.txt when `BUILD_C_LIBRARY` is disabled.

cpp/CMakeLists.txt
```diff
@@ -746,23 +746,56 @@ target_compile_definitions(cuvs::cuvs INTERFACE $<$<BOOL:${CUVS_NVTX}>:NVTX_ENAB
...
   if(BUILD_C_LIBRARY)
+    install(
+      TARGETS cuvs_c
+      DESTINATION ${lib_dir}
+      COMPONENT c_api
+      EXPORT cuvs-c-exports
+    )
+  endif()
+
+  if(NOT CUVS_COMPILE_DYNAMIC_ONLY)
```
robertmaynard, line 770: please us `cuvs_compile_mode` like the rest of the file

cpp/CMakeLists.txt
```diff
@@ -779,13 +812,19 @@ target_compile_definitions(cuvs::cuvs INTERFACE $<$<BOOL:${CUVS_NVTX}>:NVTX_ENAB
     set(CUVS_C_TARGET cuvs_c)
   endif()
 
+  if(TARGET cuvs_static)
+    list(APPEND cuvs_components cuvs_static)
+    list(APPEND cuvs_export_sets cuvs-static-exports)
+    set(CUVS_STATIC_TARGET cuvs_static)
```
robertmaynard, line 818: No need for this variable, it just makes the code harder to read

robertmaynard: `GLOBAL_TARGETS` will ignore targets that don't exist. So we should just pass it `cuvs cuvs_c cuvs_static`

cpp/CMakeLists.txt
```diff
@@ -746,23 +746,56 @@ target_compile_definitions(cuvs::cuvs INTERFACE $<$<BOOL:${CUVS_NVTX}>:NVTX_ENAB
...
-      TARGETS cuvs_c
-      DESTINATION ${lib_dir}
-      COMPONENT c_api
-      EXPORT cuvs-c-exports
+      FILES ${CMAKE_CURRENT_BINARY_DIR}/include/cuvs/version_config.h
+      COMPONENT cuvs_static
+      DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/cuvs
     )
 
     install(
```
robertmaynard, line 796: This entire `c_config` install should be removed. This file is already installed by the `cuvs` component which is non-optional

cpp/CMakeLists.txt
```diff
@@ -746,23 +746,56 @@ target_compile_definitions(cuvs::cuvs INTERFACE $<$<BOOL:${CUVS_NVTX}>:NVTX_ENAB
...
+      COMPONENT cuvs_static
       DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}
     )
 
     install(
-      TARGETS cuvs_c
-      DESTINATION ${lib_dir}
-      COMPONENT c_api
-      EXPORT cuvs-c-exports
+      FILES ${CMAKE_CURRENT_BINARY_DIR}/include/cuvs/version_config.h
```
robertmaynard, line 791: This entire version_config install should be removed. This file is already installed by the `cuvs` component which is non-optional.

cpp/CMakeLists.txt
```diff
@@ -746,23 +746,56 @@ target_compile_definitions(cuvs::cuvs INTERFACE $<$<BOOL:${CUVS_NVTX}>:NVTX_ENAB
...
+      EXPORT cuvs-static-exports
+    )
+
+    install(
+      DIRECTORY include/cuvs
+      COMPONENT cuvs_static
+      DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}
+    )
+
     install(
```
robertmaynard, line 784: This  install rule should be removed. These files are already installed by the cuvs component which is non-optional.

cpp/CMakeLists.txt
```diff
@@ -746,23 +746,56 @@ target_compile_definitions(cuvs::cuvs INTERFACE $<$<BOOL:${CUVS_NVTX}>:NVTX_ENAB
...
+
+  if(NOT CUVS_COMPILE_DYNAMIC_ONLY)
+    install(
+      TARGETS cuvs_static
+      DESTINATION ${lib_dir}
+      COMPONENT cuvs_static
+      EXPORT cuvs-static-exports
+    )
+
+    install(
```
robertmaynard, line 778: This install rule should be removed. These files are already installed by the cuvs component which is non-optional.

cpp/CMakeLists.txt
```diff
@@ -734,7 +734,7 @@ target_compile_definitions(cuvs::cuvs INTERFACE $<$<BOOL:${CUVS_NVTX}>:NVTX_ENAB
   include(CPack)
 
   install(
-    TARGETS ${_cuvs_lib_targets}
+    TARGETS cuvs
```
robertmaynard, line 737: This fails when `BUILD_SHARED_LIBS` is OFF

robertmaynard: This needs to be guarded by `NOT cuvs_compile_mode STREQUAL static_only`

conda/recipes/libcuvs/recipe.yaml
```diff
@@ -134,6 +134,61 @@ outputs:
       license: ${{ load_from_file("python/libcuvs/pyproject.toml").project.license.text }}
       summary: ${{ load_from_file("python/libcuvs/pyproject.toml").project.description }}
 
+  - package:
+      name: libcuvs-static
```
robertmaynard, line 138: We can't have libcuvs-static also install headers as we would have 2 projects that both install the same set of headers

divyegala: The expectation is that no user will install both libcuvs and libcuvs-static conda packages at the same time, right?

cjnolet:
> This is the only correct approach I see if we want to make cuvs-static optional.

Given the push to reduce package sizes (in addition to binary sizes) as much as possible, this sounds like the only logical approach to take.

divyegala:
@robertmaynard @cjnolet should I close this PR, then?

robertmaynard:
> @robertmaynard @cjnolet should I close this PR, then?

Entirely up to you. What ever you think would be easier

divyegala:
@robertmaynard if we're going to do it the way you described for the long term solution then I don't think I have the CMake-chops to make it happen without a fair amount of guidance from your side.

I can give it a shot or you may prefer to do it yourself?

robertmaynard:
> @robertmaynard if we're going to do it the way you described for the long term solution then I don't think I have the CMake-chops to make it happen without a fair amount of guidance from your side.
> 
> I can give it a shot or you may prefer to do it yourself?

Give it a shot and I can help out via reviews

PR #1487, opened 2025-11-03
VPQ Dataset: write codes one word at a time during building

achirkin:
Improve the efficiency of `process_and_fill_codes_kernel` by writing the codes in larger chunks.


cjnolet (approved):

achirkin:
Moving this to draft: although the write efficiency improves by 2-5x according to the nsight profiler (less store instructions), the overall kernel runtime barely changes at all, because the bottleneck is data reading and ALU (encoding). So the value of the PR is in question.

aamijar:
Hi @achirkin, should we close this PR for now if we don't need it?

PR #1488, opened 2025-11-03
Decouple C++ library from C library

divyegala:
Currently, the C++ library needs C headers to build and be installed at runtime (as it was leaking those headers). This is unnecessary circular dependency and as the only C header needed by the C++ library was that of `cuvs/distance/distance.h`, it was easily resolved. As precedent has been set already, we declare an `enum` first in C++ and then create a duplicate of it in C, and cast between the two when calling the C++ API from the C API.

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -754,8 +754,8 @@ target_compile_definitions(cuvs::cuvs INTERFACE $<$<BOOL:${CUVS_NVTX}>:NVTX_ENAB
 
     install(
       DIRECTORY ../c/include/cuvs/
-      component c_api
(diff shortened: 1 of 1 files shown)


c/src/neighbors/refine.cpp
```diff
@@ -11,6 +11,8 @@
 #include <raft/core/resources.hpp>
 
 #include <cuvs/core/c_api.h>
+#include <cuvs/distance/distance.h>
+#include <cuvs/distance/distance.hpp>
```
tarang-jain, line 15: Why exactly do we need both the headers here?

tarang-jain: Oh I think I understand what you are trying to do. Thanks!

tarang-jain (approved):

Issue #1447, opened 2025-10-21
`eigen_tol` parameter in spectral embedding


aamijar:
It can be useful to expose the `eigen_tol` parameter in spectral embedding, so that the user can tune the value based on their needs. sklearn also does this.

PR #1493, opened 2025-11-04
eigen `tolerance` option Spectral Embedding

aamijar:
Resolves https://github.com/rapidsai/cuvs/issues/1447

This PR exposes the `tolerance` parameter so that users can set it according to their needs in Spectral Embedding and Spectral Clustering.
We will be able to expose this option on the cuml estimator side as well.

### DIFF
--- a/cpp/src/preprocessing/spectral/detail/spectral_embedding.cuh
+++ b/cpp/src/preprocessing/spectral/detail/spectral_embedding.cuh
@@ -100,7 +100,7 @@ void compute_eigenpairs(raft::resources const& handle,
   config.n_components   = spectral_embedding_config.n_components;
   config.max_iterations = 10 * n_samples;
   config.ncv            = std::min(n_samples, std::max(2 * config.n_components + 1, 20));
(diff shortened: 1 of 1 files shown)


cpp/include/cuvs/preprocessing/spectral_embedding.hpp
```diff
@@ -45,6 +45,14 @@ struct params {
    */
   bool drop_first;
 
+  /**
+   * @brief Tolerance for the eigenvalue solver.
+   *
+   * The tolerance for the eigenvalue solver. This is used to determine
+   * when to stop the eigenvalue solver.
+   */
+  float eigen_tolerance{1e-5f};
```
cjnolet: Nitpick: Can we just calling this "tolerance" or "eps"?

aamijar: Addressed in https://github.com/rapidsai/cuvs/pull/1493/commits/07262d2294735a0f0815fa358058b4f9ec6b7702

PR #1494, opened 2025-11-04
Introduce `libcuvs-headers` and `libcuvs-static` CMake components and conda packages

divyegala:
The PR adds a new target `cuvs-cpp-headers` that is part of the main `cuvs-exports` set. It also separates out `cuvs_static` as its own optional component.

Two new conda packages are added to reflect the aforementioned changes, namely `libcuvs-headers` and `libcuvs-static`.

### DIFF
--- a/cpp/cmake/modules/generate_cuvs_export.cmake
+++ b/cpp/cmake/modules/generate_cuvs_export.cmake
@@ -31,8 +31,9 @@ function(generate_cuvs_export)
       [=[
 
   set(_cuvs_implicit_comp_names cuvs_cpp_headers cuvs_shared cuvs_static)
-  foreach(_cuvs_comp IN LISTS _cuvs_implicit_comp_names)
-    if(NOT TARGET cuvs::${_cuvs_comp})
+  set(_cuvs_implicit_target_names cuvs_cpp_headers cuvs cuvs_static)
+  foreach(_cuvs_comp _cuvs_target IN ZIP_LISTS _cuvs_implicit_comp_names _cuvs_implicit_target_names)
+    if(NOT TARGET cuvs::${_cuvs_target})
       file(GLOB cuvs_component_dep_files LIST_DIRECTORIES FALSE
             "${CMAKE_CURRENT_LIST_DIR}/cuvs-${_cuvs_comp}*-dependencies.cmake")
       foreach(f IN LISTS  cuvs_component_dep_files)


robertmaynard (commented):
Working through the testing the changes locally, my biggest thought so far is that we can't have an option to disable the headers install. They are a required component if we are providing the shared or static version of libcuvs.so

divyegala:
@robertmaynard that makes sense, I did not like introducing that option either.

The other problem (maybe okay for this release) is that we'll need a separate target for the C library if we want to build only once and ship the C++ shared and static libraries separately. As it stands currently, I believe the C library can only link to the C++ shared library.

robertmaynard:
> The other problem (maybe okay for this release) is that we'll need a separate target for the C library if we want to build only once and ship the C++ shared and static libraries separately. As it stands currently, I believe the C library can only link to the C++ shared library.

Yeah for the conda / wheel packages the C library will need a dependency on the shared library. For things like java builds and the prebuilt tarball the C library links against the static version ( https://github.com/rapidsai/cuvs/pull/1524 ).

c/CMakeLists.txt
```diff
@@ -158,30 +162,15 @@ if(PROJECT_IS_TOP_LEVEL)
...
   # Add CUDAToolkit as an export dependency
-  rapids_export_package(INSTALL CUDAToolkit cuvs-c-exports)
-  rapids_export_package(BUILD CUDAToolkit cuvs-c-exports)
+  rapids_export_package(INSTALL CUDAToolkit cuvs-exports)
+  rapids_export_package(BUILD CUDAToolkit cuvs-exports)
 
   install(
     TARGETS cuvs_c
     DESTINATION ${lib_dir}
     COMPONENT cuvs_c
```
divyegala: Is the COMPONENT name mismatch intentional? The cpp CMakeLists.txt installs this with COMPONENT name `c_api`

conda/recipes/libcuvs/recipe.yaml
```diff
@@ -145,6 +145,7 @@ outputs:
       script:
         content: |
           cmake --install cpp/build --component cuvs
```
divyegala, line 147: This is not a component anymore, right? `cuvs_shared` should install the `cuvs` target?

cpp/CMakeLists.txt
```diff
@@ -212,17 +208,15 @@ endif()
...
-  if(INSTALL_CPP_HEADERS)
-    add_library(cuvs_cpp_headers INTERFACE)
-    add_library(cuvs::cuvs_cpp_headers ALIAS cuvs_cpp_headers)
-    target_include_directories(
-      cuvs_cpp_headers
-      INTERFACE "$<BUILD_INTERFACE:${CMAKE_CURRENT_SOURCE_DIR}/include>"
-                "$<BUILD_INTERFACE:${CMAKE_CURRENT_BINARY_DIR}/include>"
-                "$<INSTALL_INTERFACE:include>"
-    )
-    target_link_libraries(cuvs_cpp_headers INTERFACE raft::raft rmm::rmm)
```
divyegala: Why did you remove these links? We leak rmm and raft from cuvs headers.

c/CMakeLists.txt
robertmaynard: That was an oversight, corrected.

conda/recipes/libcuvs/recipe.yaml
robertmaynard: The `cuvs-config.cmake` and other support CMake files are still in the `cuvs` target. This is due to allow us to support generating installs that don't have a `cuvs_shared`

cpp/CMakeLists.txt
robertmaynard: An oversight since each target that consumes `cuvs_cpp_headers` also links to raft and rmm as well.

robertmaynard:
Realized I need to also make `cuvs::cuvs_static` also be automatically imported when it exists. Adding that now

robertmaynard:
> Realized I need to also make `cuvs::cuvs_static` also be automatically imported when it exists. Adding that now

Addressed in the latest commit

conda/recipes/libcuvs/recipe.yaml
```diff
@@ -78,6 +78,61 @@ cache:
           - mkl-devel =2023
 
 outputs:
+  - package:
+      name: libcuvs-headers
```
bdice, line 82: This is very close to what we would want in a `-dev` package but perhaps reversed -- where `libcuvs` would be only runtime libs while `libcuvs-dev` would be runtime libs plus headers. See proposal in https://github.com/rapidsai/build-planning/issues/46.

What is our main goal in splitting out the headers? I am curious if that approach for `-dev` packages still makes sense or not. It might be different in a world with JIT where runtime always requires headers.

robertmaynard: We need the headers for both the cuvs and cuvs-static packages to work in a `-dev` sense, but don't want to have two conda packages that clobber the same headers so that is why they are in `cpp-headers`.

The extra complexity is that I am also planning for splitting out the `C` API from the libcuvs conda package. That would allow us to have `libcuvs-c-dev` which only installs the `libcuvs.so`, `libcuvs_c.so` and the `C` headers

>  In most cases, ${lib}-dev will likely just be a metapackage that pulls in ${lib} and ${lib}-headers

It looks like we match the proposal in 46

bdice (approved):

robertmaynard (approved):

gforsyth:
Run dependencies for the two new `conda` packages look good:

```
 │ │ Finalized run dependencies (libcuvs-headers-25.12.00a148-cuda13_251124_6f21d3b8):
 │ │ ╭──────────────────┬────────────────────────────────────────────────────────╮
 │ │ │ Name             ┆ Spec                                                   │
 │ │ ╞══════════════════╪════════════════════════════════════════════════════════╡
 │ │ │ Run dependencies ┆                                                        │
 │ │ │ cuda-version     ┆ >=13,<14.0a0 (PC)                                      │
 │ │ │ libraft-headers  ┆ 25.12.*                                                │
 │ │ │ librmm           ┆ 25.12.*                                                │
 │ │ │ nccl             ┆                                                        │
 │ │ │ cuda-cudart      ┆                                                        │
 │ │ │ libcublas        ┆                                                        │
 │ │ │ libcurand        ┆                                                        │
 │ │ │ libcusolver      ┆                                                        │
 │ │ │ libcusparse      ┆                                                        │
 │ │ │ __glibc          ┆ >=2.28,<3.0.a0 (RE of [build: sysroot_linux-64])       │
 │ │ │ __glibc          ┆ >=2.28,<3.0.a0 (RE of [cache-build: sysroot_linux-64]) │
 │ │ │ libgcc           ┆ >=14 (RE of [cache-build: gcc_linux-64])               │
 │ │ │ libstdcxx        ┆ >=14 (RE of [cache-build: gxx_linux-64])               │
 │ │ │ libgcc           ┆ >=14 (RE of [cache-build: gxx_linux-64])               │
 │ │ ╰──────────────────┴────────────────────────────────────────────────────────╯
```

```
 │ │ Finalized run dependencies (libcuvs-static-25.12.00a148-cuda13_251124_6f21d3b8):
 │ │ ╭──────────────────┬────────────────────────────────────────────────────────╮
 │ │ │ Name             ┆ Spec                                                   │
 │ │ ╞══════════════════╪════════════════════════════════════════════════════════╡
 │ │ │ Run dependencies ┆                                                        │
 │ │ │ cuda-version     ┆ >=13,<14.0a0 (PC)                                      │
 │ │ │ libcuvs-headers  ┆ ==25.12.0a148 cuda13_251124_6f21d3b8 (PS)              │
 │ │ │ libraft-headers  ┆ 25.12.*                                                │
 │ │ │ librmm           ┆ 25.12.*                                                │
 │ │ │ nccl             ┆                                                        │
 │ │ │ cuda-cudart      ┆                                                        │
 │ │ │ libcublas        ┆                                                        │
 │ │ │ libcurand        ┆                                                        │
 │ │ │ libcusolver      ┆                                                        │
 │ │ │ libcusparse      ┆                                                        │
 │ │ │ __glibc          ┆ >=2.28,<3.0.a0 (RE of [build: sysroot_linux-64])       │
 │ │ │ libstdcxx        ┆ >=14 (RE of [cache-build: gxx_linux-64])               │
 │ │ │ libgcc           ┆ >=14 (RE of [cache-build: gxx_linux-64])               │
 │ │ │ libgcc           ┆ >=14 (RE of [cache-build: gcc_linux-64])               │
 │ │ │ __glibc          ┆ >=2.28,<3.0.a0 (RE of [cache-build: sysroot_linux-64]) │
 │ │ ╰──────────────────┴────────────────────────────────────────────────────────╯
```

PR #1496, opened 2025-11-05
Add filter for cagra::merge

benfred:
Add an optional filter argument to cagra::merge, so that we can compact the index to remove deleted items when merging.

Also remove the 'LOGICAL_MERGE' strategy from the `cagra::merge` api, since this is only supported with the composite merge api. (The `cagra::merge` function ignore this value if set, and only does a logical merge).


cpp/include/cuvs/neighbors/cagra_index_wrapper.hpp
```diff
@@ -12,11 +23,43 @@
...
+   * @brief Constructs merge parameters with given index parameters.
+   * @param params Parameters for creating the output index.
+   */
+  explicit merge_params(const cagra::index_params& params) : output_index_params(params) {}
+
+  /// Parameters for creating the output index.
+  cagra::index_params output_index_params;
+
+  /// Strategy for merging. Defaults to `MergeStrategy::MERGE_STRATEGY_PHYSICAL`.
+  cuvs::neighbors::MergeStrategy merge_strategy =
```
cjnolet, line 41: Should we just remove the merge params altogether? I don't know if we need them anymore. Maybe we do- that's why I'm asking. Might help to think forward to when we have a SMART merge option (maye you've thought of this already). But maybe we don't need a "merge strategy" for that? Maybe we just have a flag or parameter for "rebuild" vs "combine"?

c/src/neighbors/cagra.cpp
```diff
@@ -669,16 +633,16 @@ extern "C" cuvsError_t cuvsCagraMerge(cuvsResources_t res,
...
     } else if (dtype.code == kDLFloat && dtype.bits == 16) {
       output_index->addr =
-        reinterpret_cast<uintptr_t>(_merge<half>(res, *params, indices, num_indices));
+        reinterpret_cast<uintptr_t>(_merge<half>(res, *params, indices, num_indices, filter));
     } else if (dtype.code == kDLInt && dtype.bits == 8) {
       output_index->addr =
-        reinterpret_cast<uintptr_t>(_merge<int8_t>(res, *params, indices, num_indices));
+        reinterpret_cast<uintptr_t>(_merge<int8_t>(res, *params, indices, num_indices, filter));
     } else if (dtype.code == kDLUInt && dtype.bits == 8) {
       output_index->addr =
```
cjnolet, line 723: Ah- I was trying to figure out how you were going to support allowing a user to pass in the "index" to populate in the C layer while only have the option to return a newly created index in the C++ layer. Not opposed to this solution at all. Just good that we're not doing a copy!

cpp/src/neighbors/detail/cagra/cagra_merge.cuh
```diff
@@ -84,29 +96,74 @@ index<T, IdxT> merge(raft::resources const& handle,
...
+        cagra::build(handle, params, raft::make_const_mdspan(filtered_dataset.view()));
+      if (!merged_index.data().is_owning() && params.attach_dataset_on_build) {
+        using matrix_t           = decltype(updated_dataset);
+        using layout_t           = typename matrix_t::layout_type;
+        using container_policy_t = typename matrix_t::container_policy_type;
+        using owning_t           = owning_dataset<T, int64_t, layout_t, container_policy_t>;
+        auto out_layout          = raft::make_strided_layout(filtered_dataset.view().extents(),
+                                                    std::array<int64_t, 2>{stride, 1});
+
+        merged_index.update_dataset(handle, owning_t{std::move(filtered_dataset), out_layout});
```
cjnolet, line 130: We really need to prioritize having the indexes all accept "Dataset" instead of mdspans. This is going to get out of control :-(

cpp/include/cuvs/neighbors/cagra_index_wrapper.hpp
```diff
@@ -12,11 +23,43 @@
 namespace cuvs::neighbors::cagra {
 template <typename T, typename IdxT>
 struct index;
-struct merge_params;
+struct index_params;
```
cjnolet, line 15: These include files have gotten super lightweight now that they are only including prototypes and declarations. I'm not a huge fan of all of the CAGRA headers we're creating. Can we just consolidate this into the main header? 

Also- for some reason, we have a header for `cagra_optimize.hpp` and I think this was a misunderstanding. Would you be able to copy that into the main cagra.hpp in this PR? It should be a straightforward copy/paste... just trying to keep the number of headers manageable. The whole benefit to no longer being header only is that the headers are super super lightweight.

cpp/tests/neighbors/ann_cagra.cuh
```diff
@@ -943,6 +943,242 @@ class AnnCagraFilterTest : public ::testing::TestWithParam<AnnCagraInputs> {
...
+      EXPECT_TRUE(eval_neighbours(indices_naive,
+                                  indices_Cagra,
+                                  distances_naive,
+                                  distances_Cagra,
+                                  ps.n_queries,
+                                  ps.k,
+                                  0.006,
+                                  min_recall));
+
+      /* TODO: eval_distances doesn't work, potentially because of id translation mismatch
```
benfred, line 1150: fwiw, This needs either fixed (by adding id translations) - or removed before this PR should be merged

cjnolet (approved):

robertmaynard:
@benfred We are going to have to open a new PR and introduce `cuvsCagraMerge_v2` with this new signature as this breaks ABI between 26.02 and 26.04

benfred:
@robertmaynard this change is in 26.02 - I messed up the merge on this pr, but we merged into 26.02 with #1755

robertmaynard:
> @robertmaynard this change is in 26.02 - I messed up the merge on this pr, but we merged into 26.02 with #1755

@benfred That is great news! Thanks

Issue #1467, opened 2025-10-28
[BUG] [Java] `CagraBuildAndSearchIT.testFloatIndexing` fails randomly


mythrocks:
One of the tests introduced by #1366 seems to be failing randomly in CI.  

For instance, from [the CI run for #1459](https://github.com/rapidsai/cuvs/actions/runs/18854301203/job/53801261133?pr=1459):
```
Error:  Tests run: 17, Failures: 0, Errors: 2, Skipped: 0, Time elapsed: 7.993 s <<< FAILURE! -- in com.nvidia.cuvs.CagraBuildAndSearchIT
Error:  com.nvidia.cuvs.CagraBuildAndSearchIT.testFloatIndexing -- Time elapsed: 0.186 s <<< ERROR!
java.util.concurrent.ExecutionException: 
java.lang.AssertionError: Exception while executing runnable: java.lang.RuntimeException: cuvsCagraBuild returned 0[RAFT failure at file=/tmp/conda-bld-output/bld/rattler-build_libcuvs/work/cpp/src/neighbors/detail/cagra/graph_core.cuh line=1406: Could not generate an intermediate CAGRA graph because the initial kNN graph contains too many invalid or duplicated neighbor nodes. This error can occur, for example, if too many overflows occur during the norm computation between the dataset vectors.
Obtained 6 stack frames
#1 in /opt/conda/envs/java/lib/jvm/lib/server/../../../libcuvs.so(+0x56826d) [0x7a570b4db26d]
#2 in /opt/conda/envs/java/lib/jvm/lib/server/../../../libcuvs.so: void cuvs::neighbors::cagra::detail::graph::optimize<unsigned int, raft::host_device_accessor<std::experimental::default_accessor<unsigned int>, (raft::memory_type)0> >(raft::resources const&, std::experimental::mdspan<unsigned int, std::experimental::extents<long, 18446744073709551615ul, 18446744073709551615ul>, std::experimental::layout_right, raft::host_device_accessor<std::experimental::default_accessor<unsigned int>, (raft::memory_type)0> >, std::experimental::mdspan<unsigned int, std::experimental::extents<long, 18446744073709551615ul, 18446744073709551615ul>, std::experimental::layout_right, raft::host_device_accessor<std::experimental::default_accessor<unsigned int>, (raft::memory_type)0> >, bool, bool) +0x23f9 [0x7a570bf36c19]
[...]

PR #1503, opened 2025-11-06
[Java] Disable flaky test `testFloatSerialization`

mythrocks:
This commit disables a flaky test
`CagraBuildAndSearchIT.testFloatSerialization`, to reduce churn in CI pipelines.

The test fails occasionally (e.g. with `tests.seed=42FC5CC6B4C6BA8E`), producing similar crashes to those listed in #1467.  The root cause is likely to be the same.

This test will be re-enabled once #1467 is resolved.

### DIFF
--- a/docs/source/build.rst
+++ b/docs/source/build.rst
@@ -98,7 +98,7 @@ Conda environment scripts are provided for installing the necessary dependencies
 
 .. code-block:: bash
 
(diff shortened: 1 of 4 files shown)


mythrocks:
CI seems to be failing with CI runners receiving 403s from Github.  I'll back off and try again in a bit.

PR #1504, opened 2025-11-06
Deduplicate `KMeans` instantiations

### LINKED ISSUE #1514: Remove kernel duplication in KMeans
Some kernels related to `kmeans` are used internally several times while others in `kmeans_balanced` are instantiated again by CAGRA/IVF-PQ/SCANN.

divyegala:
Closes #1514 

CUDA 13 binary size from 464 MB to 437 MB.
CUDA 12 binary size from 1165 MB to 1102 MB.


cpp/src/cluster/kmeans.cuh
```diff
@@ -1,11 +1,10 @@
 /*
- * SPDX-FileCopyrightText: Copyright (c) 2020-2024, NVIDIA CORPORATION.
+ * SPDX-FileCopyrightText: Copyright (c) 2020-2025, NVIDIA CORPORATION.
  * SPDX-License-Identifier: Apache-2.0
- */
+ */ \
```
lowener:
```suggestion
 */
```

cpp/include/cuvs/cluster/kmeans.hpp
lowener, line 1: Add `fit_predict` for the new data type half and uint8

cpp/include/cuvs/cluster/kmeans.hpp
```diff
@@ -819,6 +887,138 @@ void predict(const raft::resources& handle,
...
+ *   cuvs::cluster::kmeans::balanced_params params;
+ *   int64_t n_features = 15, n_clusters = 8;
+ *   auto centroids = raft::make_device_matrix<float, int64_t>(handle, n_clusters, n_features);
+ *
+ *   kmeans::fit(handle,
+ *               params,
+ *               X,
+ *               centroids.view());
+ *   ...
+ *   auto labels = raft::make_device_vector<int, int64_t>(handle, X.extent(0));
```
lowener:
```suggestion
 *   auto labels = raft::make_device_vector<uint32_t, int64_t>(handle, X.extent(0));
```
This fix should be applied to all the `predict()` examples probably. `@copydoc` should also be considered to simplify this header file, but I don't know if it will be compatible with having two different `fit()` functions, for kmeans and kmeans_balanced

cpp/include/cuvs/cluster/kmeans.hpp
divyegala: Actually, I'd prefer to move away from `fit_predict` conventions entirely. It's really syntactic sugar and a scikit-learn convention. Users can explicitly call `fit` then `predict` if they so desire.

cjnolet: Is there a perf implication to this? Usually if we know the user is interested in fitting for intention of prediction, we've already done the calculation for the prediction so that should be a no-op. If that's not the case, we need to revisit this, because that means we're missing out some perf gains. Cluster assignment can be more expensive than training in many scenarios.

divyegala: As of now, there's no shared state between fit and predict. fit_predict was just calling into the individual functions.

cjnolet (approved):

PR #1505, opened 2025-11-06
Remove duplicate VPQ code of scann

lowener:
Scann is using `process_and_fill_codes_subspaces` but this code is already present in `vpq_dataset.cuh`.
The code in `vpq_dataset.cuh` is more recent because it supports batching.
This PR removes the duplicated code.


lowener:
Instead of creating this independant compilation unit, we can use the PQ public API I am preparing in #1278 
Created #1575 to track that

PR #1510, opened 2025-11-07
[Java] Fix `CagraBuildAndSearchIT` concurrent deallocation errors

ldematte:
Fixes https://github.com/rapidsai/cuvs/issues/1467 (see issue for details)

I added some diagnostic (debug level) that I kept in the test, as it might be useful for future investigations.


java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/CuVSMatrixBaseImpl.java
```diff
@@ -39,6 +39,11 @@ protected CuVSMatrixBaseImpl(
     this.columns = columns;
   }
 
+  @Override
+  public String toString() {
+    return String.format("%dx%d %s @ 0x%016X", size, columns, dataType, memorySegment.address());
```
mythrocks, line 44: 👍

java/README.md
```diff
@@ -41,6 +41,16 @@ Also, ensure that your panama bindings are up-to-date. They can be re-generated
 ./panama-bindings/generate-bindings.sh
 ```
 
+Tests run using a randomized runner. Specific failures can be reproduced running a test suite with a specific seed,
+by passing `-Dtests.seed=42FC5CC6B4C6BA8E` (where `42FC5CC6B4C6BA8E` has to be
+replaced with your specific seed). It also possible to re-run a single test, but
+in this case it's necessary to pass the extended seed (suite:method), e.g.
```
mythrocks, line 47: 👍

mythrocks:
I feel a little silly not to have suspected this being the problem, and barking up the wrong tree.  Good catch, @ldematte!  This is a decent fix. 

In the short term, we can keep this to be the fix.  In the longer term, I think this could benefit from a reference-counted close() mechanism.  That's the method used in other projects I've worked on.

I'm open to other ideas, of course.

PR #1495, opened 2025-11-04
Use `sccache-dist` build cluster for conda and wheel builds

trxcllnt:
## Description
RAPIDS has deployed an autoscaling cloud build cluster that can be used to accelerate building large RAPIDS projects. 

This PR updates the conda and wheel builds to use the build cluster.

This contributes to https://github.com/rapidsai/build-planning/issues/228.

PR #1524, opened 2025-11-10
Extend CI to build and test x86 libcuvs_c tarballs

robertmaynard:
Adds new `rocky8-clib-standalone-build` and  `rocky8-clib-tests` PR jobs that validate that the C api binaries can be built and run all C tests correctly.

Also adds a new nightly build job that produces the C api binaries.


.github/workflows/pr.yaml
```diff
@@ -148,41 +174,71 @@ jobs:
...
         cuda_version:
           - &latest_cuda12 '12.9.1'
           - &latest_cuda13 '13.0.2'
+    with:
+      build_type: pull-request
+      arch: "amd64"
+      date: ${{ inputs.date }}_c
+      container_image: "rapidsai/ci-wheel:25.12-cuda${{ matrix.cuda_version }}-rockylinux8-py3.10"
+      # requires_license_builder: false
+      script: "ci/build_standalone_c.sh  --build-tests"
```
bdice:
```suggestion
      script: "ci/build_standalone_c.sh --build-tests"
```

.github/workflows/build.yaml
```diff
@@ -41,6 +41,28 @@ jobs:
...
+    strategy:
+      fail-fast: false
+      matrix:
+        cuda_version:
+          - &latest_cuda12 '12.9.1'
+          - &latest_cuda13 '13.0.2'
+    with:
+      build_type: ${{ inputs.build_type || 'branch' }}
+      branch: ${{ inputs.branch }}
+      arch: "amd64"
```
bdice, line 57: Are we only supporting amd64 with the C library?

.github/workflows/build.yaml
```diff
@@ -41,6 +41,28 @@ jobs:
...
+  rocky8-clib-standalone-build:
+    secrets: inherit
+    uses: rapidsai/shared-workflows/.github/workflows/custom-job.yaml@main
+    strategy:
+      fail-fast: false
+      matrix:
+        cuda_version:
+          - &latest_cuda12 '12.9.1'
+          - &latest_cuda13 '13.0.2'
+    with:
```
bdice, line 54: I would specify the `node_type` explicitly. It defaults to `cpu8`. Same in `pr.yaml`.

ci/build_standalone_c.sh
```diff
@@ -0,0 +1,90 @@
...
+      cmake -S cpp -B cpp/build/ \
+            -DCMAKE_CUDA_HOST_COMPILER=/opt/rh/gcc-toolset-${TOOLSET_VERSION}/root/usr/bin/gcc \
+            -DCMAKE_CUDA_ARCHITECTURES=RAPIDS \
+            -DBUILD_SHARED_LIBS=OFF \
+            -DCUTLASS_ENABLE_TESTS=OFF \
+            -DDISABLE_OPENMP=OFF \
+            -DBUILD_TESTS=OFF \
+            -DBUILD_SHARED_LIBS=ON \
+            -DCUVS_STATIC_RAPIDS_LIBRARIES=ON
+cmake --build cpp/build -j9
```
bdice: Why `-j9`? We usually use `-j${PARALLEL_LEVEL}` which is set by `rapids-configure-sccache`.

https://github.com/rapidsai/gha-tools/blob/1814d7d1ab6742da1a38c3e1082065050d279890/tools/rapids-configure-sccache#L13

On an 8-core node, `-j9` might oversubscribe, which tends to crash the GitHub Actions logging process.

ci/test_standalone_c.sh
```diff
@@ -0,0 +1,37 @@
+#!/bin/bash
+# SPDX-FileCopyrightText: Copyright (c) 2022-2025, NVIDIA CORPORATION.
```
bdice: New file, new copyright. Same in other files.
```suggestion
# SPDX-FileCopyrightText: Copyright (c) 2025, NVIDIA CORPORATION.
```

ci/build_standalone_c.sh
```diff
@@ -0,0 +1,90 @@
+#!/bin/bash
+# SPDX-FileCopyrightText: Copyright (c) 2022-2025, NVIDIA CORPORATION.
+# SPDX-License-Identifier: Apache-2.0
+
+set -euo pipefail
+
+TOOLSET_VERSION=14
+CMAKE_VERSION=3.31.8
```
bdice, line 8: Why not use CMake 4.x?

.github/workflows/build.yaml
robertmaynard: I have on my todo to extend the matrix to arm + x86 but that isn't a hard requirement for 25.12.

If you know how to easily extend the matrix in a custom job I would happily integrate it into this PR.
When I was looking at other yaml files it seemed complicated ( dumping to file for some reason )

ci/build_standalone_c.sh
robertmaynard: I wasn't aware of `PARALLEL_LEVEL`, thanks!

ci/build_standalone_c.sh
robertmaynard: Updated to 4.1.2

ci/test_standalone_c.sh
robertmaynard: Updated both files

.github/workflows/build.yaml
robertmaynard: I see that our wheel builds use `cpu16` and our conda builds use `cpu8` be default. So would you recommend cpu16 to match the wheel builds?

.github/workflows/build.yaml
bdice: Let's defer this until later. You can just add a `matrix` entry here like already exists for the CUDA keys and then use `${{ matrix.arch }}` in the `arch:` value but let's get one arch done before we expand.

.github/workflows/build.yaml
bdice: I think we'll be standardizing on `cpu8` for both conda and wheels once the sccache-dist build cluster work is done. However, if there's a lot of code to compile here and we're _not_ using the sccache-dist build cluster in this job, we can use `cpu16`.

bdice: Approving without this -- feel free to add it if desired, but this isn't blocking.

robertmaynard: It does compile all of cuvs, so I will go with `cpu16` and we can change once https://github.com/rapidsai/cuvs/pull/1495 is merged

PR #1525, opened 2025-11-10
random orthogonal transformation preprocess

enp1s0:
This PR introduces a random orthogonal transformation as a preprocessor for CAGRA-Q, etc. By using this preprocessor, we can achieve a higher recall in the CAGRA-Q search depending on the datasets. This implementation generates an orthogonal matrix through the QR decomposition of a random matrix and then multiplies it by the dataset matrix.

figure-recall


cpp/include/cuvs/preprocessing/linear_transform/random_orthogonal.hpp
```diff
@@ -0,0 +1,477 @@
...
+ * @tparam T data element type
+ *
+ */
+template <typename T>
+struct transformer {
+  raft::device_matrix<T, int64_t> orthogonal_matrix;
+};
+
+/**
+ * @brief Initializes a random orthogonal transoformation to be used later for transforming the
```
cjnolet:
```suggestion
 * @brief Initializes a random orthogonal transformation to be used later for transforming the
```

cjnolet:
@enp1s0 this looks like a good capability to have in cuVS. The recall improvements in your benchmarks look good, but this is also targeting CAGRA-Q improvement, which I think we would want to also support at scale. Have you evaluated this approach yet at scale? Do you have a sense for the limits that can be achieved with this approach (in terms of max dataset size on a single GPU, impact to end-to-end GPU build perf)?  Another option could be to use random projections, but that approach also has a limitation to scale (optimal number of dimensions increases with number of vectors to preserve quality).

PR #1526, opened 2025-11-10
SNMG ANN build with OpenMP nested parallelism

viclafargue:
The SNMG ANN build spawns threads with `#pragma omp parallel for`, but CAGRA build also uses OpenMP internally. With nested parallelism disabled by default, CAGRA's internal parallel regions ran serially on a single thread, causing severe performance degradation.

- Enable nested parallelism with `omp_set_nested(1)`
- Limit outer loop to `num_ranks` threads (one per GPU)
- Inside each rank thread, allocate `threads_per_rank` for CAGRA's internal parallelism
- Restore original thread count after parallel region

### DIFF
--- a/.devcontainer/Dockerfile
+++ b/.devcontainer/Dockerfile
@@ -1,3 +1,6 @@
+# SPDX-FileCopyrightText: Copyright (c) 2023-2025, NVIDIA CORPORATION. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+
 # syntax=docker/dockerfile:1.5
 
 ARG BASE
@@ -42,6 +45,7 @@ ENV HISTFILE="/home/coder/.cache/._bash_history"
 ENV AWS_ROLE_ARN="arn:aws:iam::279114543810:role/nv-gha-token-sccache-devs"
 ENV SCCACHE_REGION="us-east-2"
 ENV SCCACHE_BUCKET="rapids-sccache-devs"
+ENV SCCACHE_S3_USE_PREPROCESSOR_CACHE_MODE=true
 ENV SCCACHE_IDLE_TIMEOUT=0
 
 ###
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -39,8 +39,10 @@ jobs:
       build_type: ${{ inputs.build_type || 'branch' }}
       branch: ${{ inputs.branch }}
       date: ${{ inputs.date }}
+      node_type: cpu16
       script: ci/build_cpp.sh
       sha: ${{ inputs.sha }}
+      sccache-dist-token-secret-name: GIST_REPO_READ_ORG_GITHUB_TOKEN
 
   rocky8-clib-standalone-build:
     secrets: inherit
@@ -63,6 +65,7 @@ jobs:
       artifact-name: "libcuvs_c_${{ matrix.cuda_version }}.tar.gz"
       file_to_upload: "libcuvs_c.tar.gz"
       sha: ${{ inputs.sha }}
+      sccache-dist-token-secret-name: GIST_REPO_READ_ORG_GITHUB_TOKEN
   rust-build:
     needs: cpp-build
     secrets: inherit
@@ -84,6 +87,7 @@ jobs:
       node_type: "gpu-l4-latest-1"
       script: "ci/build_rust.sh"
       sha: ${{ inputs.sha }}
+      sccache-dist-token-secret-name: GIST_REPO_READ_ORG_GITHUB_TOKEN
   go-build:
     needs: cpp-build
     secrets: inherit
@@ -127,6 +131,7 @@ jobs:
       artifact-name: "cuvs-java-cuda${{ matrix.cuda_version }}"
       file_to_upload: "java/cuvs-java/target/"
       sha: ${{ inputs.sha }}
+      sccache-dist-token-secret-name: GIST_REPO_READ_ORG_GITHUB_TOKEN
   python-build:
     needs: [cpp-build]
     secrets: inherit
@@ -137,6 +142,7 @@ jobs:
       date: ${{ inputs.date }}
       script: ci/build_python.sh
       sha: ${{ inputs.sha }}
+      sccache-dist-token-secret-name: GIST_REPO_READ_ORG_GITHUB_TOKEN
   upload-conda:
     needs: [cpp-build, python-build]
     secrets: inherit
@@ -173,11 +179,13 @@ jobs:
       branch: ${{ inputs.branch }}
       sha: ${{ inputs.sha }}
       date: ${{ inputs.date }}
+      node_type: cpu16
       script: ci/build_wheel_libcuvs.sh
       # build for every combination of arch and CUDA version, but only for the latest Python
       matrix_filter: group_by([.ARCH, (.CUDA_VER|split(".")|map(tonumber)|.[0])]) | map(max_by(.PY_VER|split(".")|map(tonumber)))
       package-name: libcuvs
       package-type: cpp
+      sccache-dist-token-secret-name: GIST_REPO_READ_ORG_GITHUB_TOKEN
   wheel-publish-libcuvs:
     needs: wheel-build-libcuvs
     secrets: inherit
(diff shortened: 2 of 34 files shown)


cpp/src/neighbors/mg/snmg.cuh
```diff
@@ -107,8 +121,18 @@ void build(const raft::resources& clique,
     RAFT_LOG_DEBUG("SHARDED BUILD: %d*%drows", index.num_ranks_, n_rows_per_shard);
 
     index.ann_interfaces_.resize(index.num_ranks_);
+
+    // Enable nested parallelism: build uses OpenMP internally
+    int saved_omp_threads = omp_get_max_threads();
+    int threads_per_rank  = saved_omp_threads / index.num_ranks_;
```
tarang-jain: This seems risky. If saved_omp_threads is less than index.num_threads_, threads_per_rank will be set to zero. Instead we can do this:
`threads_per_rank = max(1, saved_omp_threads / index.num_ranks_);`

cpp/src/neighbors/mg/snmg.cuh
```diff
@@ -119,6 +143,9 @@ void build(const raft::resources& clique,
       cuvs::neighbors::build(dev_res, ann_if, index_params, partition);
```
tarang-jain, line 149: Can you confirm if `cuvs::neighbors::build` does not access environment variables such as `omp_get_max_threads`? If that happens this nested approach breaks with oversubscription of threads.

cpp/src/neighbors/mg/snmg.cuh
```diff
@@ -119,6 +143,9 @@ void build(const raft::resources& clique,
...
       resource::sync_stream(dev_res);
     }
+
+    // Restore original thread count
+    omp_set_num_threads(saved_omp_threads);
```
tarang-jain: I'm a bit skeptical of modifying variables like this based on env variables for the entire process. Rather we can specify threads for a particular loop such as `#pragma omp parallel for num_threads(n)` (this is for the main loop) in this PR.

cpp/src/neighbors/mg/snmg.cuh
viclafargue: Yes, `cuvs::neighbors::build` for `IVF-Flat`, `IVF-PQ` and `CAGRA` do not ever use `omp_get_max_threads`.

cpp/src/neighbors/mg/snmg.cuh
```diff
@@ -92,13 +93,25 @@ void build(const raft::resources& clique,
     RAFT_LOG_DEBUG("REPLICATED BUILD: %d*%drows", index.num_ranks_, n_rows);
 
     index.ann_interfaces_.resize(index.num_ranks_);
-#pragma omp parallel for
+
+    // Enable nested parallelism
+    int saved_omp_threads = omp_get_max_threads();
+    int threads_per_rank  = std::max(1, saved_omp_threads / index.num_ranks_);
+    omp_set_nested(1);
```
robertmaynard: All usage of OpenMP functions need to go through https://github.com/rapidsai/cuvs/blob/main/cpp/src/core/omp_wrapper.hpp so that we can properly support builds without OpenMP enabled

robertmaynard (approved):

cpp/src/core/omp_wrapper.cpp
```diff
@@ -31,6 +31,12 @@ void set_nested(int v)
   if constexpr (is_omp_enabled()) { omp_set_nested(v); }
 }
 
+void set_num_threads(int v)
+{
+  (void)v;
```
tarang-jain, line 37: why is this line needed? Can we remove it?
`(void)v;`

cpp/src/neighbors/mg/snmg.cuh
```diff
@@ -92,13 +92,28 @@ void build(const raft::resources& clique,
...
+
+    // Enable nested parallelism
+    int saved_omp_threads = cuvs::core::omp::get_max_threads();
+    int threads_per_rank  = std::max(1, saved_omp_threads / index.num_ranks_);
+    cuvs::core::omp::set_nested(1);
+
+    const int& requirements = index.num_ranks_;
+    cuvs::core::omp::check_threads(requirements);
+
+#pragma omp parallel for num_threads(index.num_ranks_)
```
tarang-jain, line 104: Ok this seems problematic again. If num_ranks is greater than saved_omp_threads, there is again oversubscription. We should add a check here -- can you see how to do that within the pragma itself?

tarang-jain: Actually this might not be needed -- the check_threads clause will warn the user if that happens.

cpp/src/neighbors/mg/snmg.cuh
```diff
@@ -107,8 +122,20 @@ void build(const raft::resources& clique,
...
 
     index.ann_interfaces_.resize(index.num_ranks_);
-#pragma omp parallel for
+
+    // Enable nested parallelism
+    int saved_omp_threads = cuvs::core::omp::get_max_threads();
+    int threads_per_rank  = std::max(1, saved_omp_threads / index.num_ranks_);
+    cuvs::core::omp::set_nested(1);
+
+    const int& requirements = index.num_ranks_;
```
tarang-jain: Nit: Is the requirements variable needed at all? Can you pass index.num_ranks_ directly?

cpp/src/neighbors/mg/snmg.cuh
```diff
@@ -92,13 +92,28 @@ void build(const raft::resources& clique,
...
+      cuvs::core::omp::set_num_threads(threads_per_rank);
+
       const raft::resources& dev_res = raft::resource::set_current_device_to_rank(clique, rank);
       auto& ann_if                   = index.ann_interfaces_[rank];
       cuvs::neighbors::build(dev_res, ann_if, index_params, index_dataset);
       resource::sync_stream(dev_res);
     }
+
+    // Restore original thread count
+    cuvs::core::omp::set_num_threads(saved_omp_threads);
```
tarang-jain: this restoration is a bit confusing. the pool of threads is set inside the for loop above, but its restored after the for loop? Can you verify if this logic is correct?

cpp/src/neighbors/mg/snmg.cuh
```diff
@@ -92,13 +92,28 @@ void build(const raft::resources& clique,
...
+    int threads_per_rank  = std::max(1, saved_omp_threads / index.num_ranks_);
+    cuvs::core::omp::set_nested(1);
+
+    const int& requirements = index.num_ranks_;
+    cuvs::core::omp::check_threads(requirements);
+
+#pragma omp parallel for num_threads(index.num_ranks_)
     for (int rank = 0; rank < index.num_ranks_; rank++) {
+      // Set thread limit for this rank's nested OpenMP regions
+      cuvs::core::omp::set_num_threads(threads_per_rank);
```
tarang-jain, line 107: Actually this also seems problematic. This is a global variable being modified by all the threads. Can lead to race conditions. Lets avoid the set_num_threads altogether -- its unnecessary risk. We can add a pragma like:
`#pragma omp parallel num_threads(threads_per_rank)`

cpp/src/neighbors/mg/snmg.cuh
```diff
@@ -119,6 +146,9 @@ void build(const raft::resources& clique,
       cuvs::neighbors::build(dev_res, ann_if, index_params, partition);
       resource::sync_stream(dev_res);
     }
+
```
tarang-jain, line 155: you must also reset the behavior for nested loops. set_nested(1) also seems to globally modify the state of the process.

cpp/src/core/omp_wrapper.cpp
viclafargue: The `(void)v;` statement suppresses "unused parameter" warning. The below if statement is evaluated at compile-time and depends on the presence of OpenMP.

cpp/src/neighbors/mg/snmg.cuh
viclafargue: I agree. This is actually not necessary. My worry was that the main thread could be one of the threads in the loop. ~But, after checking it appears like OpenMP is designed in such a way that `omp_set_num_threads` calls inside of a parallel region only affect how many threads each thread can use in nested parallel region. The main thread is unaffected, and its number of threads is actually preserved.~ I made sure that each thread sees its number of thread restored back to the global maximal number of threads within the parallel region.

cpp/src/neighbors/mg/snmg.cuh
viclafargue: `num_threads(threads_per_rank)` would have to be placed within the CAGRA build loops. The variable that sets the number of threads available for use is actually thread local. This should not be an issue.

cpp/src/neighbors/mg/snmg.cuh
viclafargue: I also added the same OpenMP usage for the `extend` function.

PR #1527, opened 2025-11-10
Remove need to have rapids_logger headers installed to use clib

robertmaynard:
We can't require users of prebuilt versions of libcuvs_c to install rapids-logger headers. 

The current strategy in https://github.com/rapidsai/cuvs/pull/1524 is to build rapids-logger statically which makes it even more non-obvious to users why they need rapids-logger headers

### DIFF
--- a/.devcontainer/cuda12.9-conda/devcontainer.json
+++ b/.devcontainer/cuda12.9-conda/devcontainer.json
@@ -11,7 +11,7 @@
   "runArgs": [
     "--rm",
     "--name",
-    "${localEnv:USER:anon}-rapids-${localWorkspaceFolderBasename}-25.12-cuda12.9-conda",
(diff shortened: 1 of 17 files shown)


robertmaynard:
> Before we merge this, can we see if there's something that rapids-logger could/should do to make this easier? If the problem for the C API is only the enums, would it be sufficient to move the enums into a separate CMake target? Basically I'm thinking if we can help resolve this with rapids-logger changes that don't require API changes (or new APIs, like a C API) but just build changes, we might as well do that now since we know we're going to need this for other libraries eventually.

This is a packaging requirement, and not something we can fix on the rapids-logger side.

The pre-built libcuvs packages are required to come with all headers needed to run them out of the box except for dlpack.
In these cases we are building rapids-logger statically, so instead of having to re-package / install the rapids-logger header(s)  which has issues, I went with removing it from our public includes

Issue #1367, opened 2025-09-24
[FEA] cuvs-bench needs to measure index conversion time


cjnolet:
Currently, cuvs-bench does not measure the conversion time for appropriate indexes. This is ultimately causing users to get an unrealistic picture of the index build times, especially for Vamana where the conversion time can grow quite large. We need to report the conversion time. 

cc @tarang-jain

cjnolet:
@tarang-jain and @achirkin have pointed out that in order to implement this, we'll need to update the conversion so that it doesn't require serializing to disk each time. We will also need to update HNSW so that it doesn't need to serialize to disk each time. 

Related: https://github.com/rapidsai/cuvs/issues/1482

achirkin:
Hi @cjnolet, actually we can measure the index conversion time as a part or regular index build using NVTX markers once #1529 is merged (see my comment to that PR)

PR #1529, opened 2025-11-11
ANN_BENCH: integrate NVTX statistics

achirkin:
Add the aggregate reporting of NVTX ranges in the output of benchmark executable.

### Usage
```bash
# Measure the CPU and GPU runtime of all NVTX ranges
nsys launch --trace=cuda,nvtx <ANN_BENCH with arguments>
# Measure only the CPU runtime of all NVTX ranges
nsys launch --trace=nvtx <ANN_BENCH with arguments>
# Do not measure/report any NVTX ranges
<ANN_BENCH with arguments>
# Do not measure/report any NVTX ranges within benchmark, but use nsys profiling as usual
nsys profile ... <ANN_BENCH with arguments>
```

### Implementation

The PR adds a single module `nvtx_stats.hpp` to the benchmark executable; there are no changes to the library at all.
The program leverages NVIDIA Nsight Systems CLI to collect and export NVTX statistics and then SQLite API to aggregate it into the benchmark state:

  1. Detect if run via `nsys launch`; if so, call `nsys start` / `nsys stop` around benchmark loop; otherwise do nothing.
  2. If the report is generated, read it and query all NVTX events and the GPU correlation data using SQLite
  3. Aggregate the NVTX events by their short names (without arguments to reduce the number of columns)
  4. Add them to the benchmark performance counters with the same averaging strategy as the global CPU/GPU runtime.

### Performance cost
If the benchmark is **not** run using `nsys launch`, there's virtually zero overhead in the new functionality.
Otherwise, there are overheads:
  1. Usual nsys profiling overheads (minimized by disabling unused information via `nsys start` CLI internally). This affects the reported performance the same way as normal nsys profiling does (especially if cuda tracing is enabled).
  2. One or more data collection/exporting events per benchmark case. These add some extra time to the benchmark time, but do not affect the counters (they are not the part of the benchmark loop)
 
Closes https://github.com/rapidsai/cuvs/issues/1367


achirkin:
As an example, I've build CAGRA-for-HNSW using only CPU events with a wiki-1M dataset, exported a CSV, and generated the breakdown bar plot using [Google Sheets](https://docs.google.com/spreadsheets/d/1aGE_tf5swj4suo-_IwS8ZqXHpcpwklYfZAumfBVYogM/edit?usp=sharing):
```bash
nsys launch --trace=nvtx ./cpp/build/bench/ann/ANN_BENCH \
  --build --force \
  --benchmark_min_time=20s \
  --benchmark_min_warmup_time=0.1 \
  --benchmark_counters_tabular \
  --benchmark_out_format=csv \
  --benchmark_out=cagra-hnsw-build.csv \
  --benchmark_filter=cuvs_cagra_hnsw.* \
  --override_kv=dataset_memory_type:\"host\" \
  wiki-1M-cagra-hnsw.json
```
CAGRA HNSW build time breakdown

tfeher (approved):
Thanks Artem for the PR! This is neat, this will be very useful for understanding more details about how time is spent during benchmarks. The PR looks good to me.

vyasr:
This seems like it could be quite useful for other RAPIDS libraries too. I wonder if we could put this somewhere that's easy to reuse in others too, WDYT?

achirkin:
In theory yes, it's rather modular and pluggable in any executable. I'm just not sure where one would keep the code if not just copy-pasting

PR #1533, opened 2025-11-12
Improve CAGRA-Q performance and add support for PQ_LEN=8

enp1s0:
This PR:
1. Introduces the E5M2 data type as the internal storage type to improve performance
1. Adds support for PQ_LEN = 8 to achieve a higher compression ratio

## E5M2 as smem data type
Using a lower-precision data type helps reduce shared memory bank conflicts and can improve throughput.
Since the quantization error from VQ+PQ is typically larger than the representation error of E5M2, the impact on search recall is expected to be negligible.

---

## Support for PQ_LEN=8

The current cuVS implementation supports only `PQ_LEN = 2` (4 bits per vector element) and `4` (2 bits per vector element).
This PR adds support for `PQ_LEN = 8` to enable a higher compression ratio (1 bit per vector element).


irina-resh-nvda (approved):
Reviewed mostly the inner nest in `compute_distance_vpq_worker` (lines 311–417): the loops, the FP8 unpack via `as_half2(bi)`, and the codebook-transpose math in setup_workspace_vpq. Built a minimal Compiler-Explorer reproducer of the consumer-side lds + unpack chain and inspected the SASS on sm_100. Confirmed one `LDS.64` per call (with the cbook/query offset folded into the immediate), one `F2FP.F16.E5M2.UNPACK_B `per `as_half2`, and no `LDL`/`STL`.

Learned a lot from this one. LGTM.

cpp/src/neighbors/detail/cagra/jit_lto_kernels/compute_distance_impl.cuh
```diff
@@ -115,12 +122,17 @@ _RAFT_DEVICE RAFT_DEVICE_INLINE_FUNCTION auto compute_distance_vpq_worker_impl(
...
 #pragma unroll
     for (std::uint32_t e = 0; e < nelem; e++) {
       const std::uint32_t k = e * kTeamVLen + elem_offset + laneId * vlen;
       if (k >= n_subspace) break;
-      device::ldg_cg(pq_codes[e], reinterpret_cast<const std::uint32_t*>(dataset_ptr + 4 + k));
+      if constexpr (std::is_same_v<PQ_CODEBOOK_LOAD_T, uint32_t>) {
+        device::ldg_cg(pq_codes[e],
+                       reinterpret_cast<const PQ_CODEBOOK_LOAD_T*>(dataset_ptr + 4 + k));
+      } else {
+        pq_codes[e] = *reinterpret_cast<const PQ_CODEBOOK_LOAD_T*>(dataset_ptr + 4 + k);
```
achirkin: Is there a specific reason to not use ldg_cg? If it's just the overload what's missing, we can that overload in https://github.com/rapidsai/cuvs/blob/main/cpp/src/neighbors/detail/cagra/device_memory_ops.hpp

cpp/src/neighbors/detail/cagra/jit_lto_kernels/cagra_planner_base.hpp
```diff
@@ -106,45 +113,205 @@ struct CagraPlannerBase : AlgorithmPlanner {
...
+                                               uint32_t dataset_block_dim,
+                                               Lambda&& l)
+  {
+    switch (team_size) {
+      case 8:
+        switch (dataset_block_dim) {
+          case 128: std::forward<Lambda>(l).template operator()<8u, 128u>(); return;
+          case 256: std::forward<Lambda>(l).template operator()<8u, 256u>(); return;
+          case 512: std::forward<Lambda>(l).template operator()<8u, 512u>(); return;
+          default: break;
```
achirkin: Would be nice to reduce these repeated lines (e.g. by introducing a template struct for each template parameter like we do in https://github.com/rapidsai/cuvs/blob/main/cpp/src/neighbors/ivf_pq/ivf_pq_compute_similarity_impl.cuh).
But you follow an already established pattern in this file, so I don't insist we must do it here.

achirkin (approved):
Thanks for working on this and sorry for such a long delay in the review process!

Also I like the packed structure; maybe we should consolidate it with the vectorized types in raft in future.

cpp/src/neighbors/detail/cagra/jit_lto_kernels/compute_distance_impl.cuh
enp1s0: Fixed it

achirkin (approved):

bdice (approved):
Approving CMake.

Issue #1454, opened 2025-10-23
[BUG] Segfault after using cuvsRMMPoolMemoryResourceEnable/cuvsRMMMemoryResourceReset


ldematte:
**Describe the bug**

When using the C API cuvsRMMPoolMemoryResourceEnable/cuvsRMMMemoryResourceReset, cuvsBruteForceSearch fails with a SEGFAULT for a null pointer:
```
siginfo: si_signo: 11 (SIGSEGV), si_code: 1 (SEGV_MAPERR), si_addr: 0x0000000000000000
```

Here is the complete stack trace for the core dump.
Notice there is some JVM stuff in the way (catching and rethrowing segfault), but it should be pretty clear nonetheless that this happens when `cuvsBruteForceSearch` tries to allocate some memory (`rmm::device_buffer::allocate_async`) after `cuvsRMMMemoryResourceReset` was called.

```
Program terminated with signal SIGABRT, Aborted.
#0  __pthread_kill_implementation (threadid=<optimized out>, signo=signo@entry=6, no_tid=no_tid@entry=0) at ./nptl/pthread_kill.c:44
44      ./nptl/pthread_kill.c: No such file or directory.
[Current thread is 1 (Thread 0x7fbdeabff6c0 (LWP 1338041))]
(gdb) bt
#0  __pthread_kill_implementation (threadid=<optimized out>, signo=signo@entry=6, no_tid=no_tid@entry=0) at ./nptl/pthread_kill.c:44
#1  0x00007fbdec30cf4f in __pthread_kill_internal (signo=6, threadid=<optimized out>) at ./nptl/pthread_kill.c:78
#2  0x00007fbdec2bdfb2 in __GI_raise (sig=sig@entry=6) at ../sysdeps/posix/raise.c:26
#3  0x00007fbdec2a8472 in __GI_abort () at ./stdlib/abort.c:79
#4  0x00007fbdeaea4179 in os::abort(bool, void const*, void const*) [clone .cold] () from /usr/lib/jvm/jdk-24.0.1/lib/server/libjvm.so
#5  0x00007fbdebbc3718 in VMError::report_and_die(int, char const*, char const*, __va_list_tag*, Thread*, unsigned char*, void const*, void const*, char const*, int, unsigned long) ()
   from /usr/lib/jvm/jdk-24.0.1/lib/server/libjvm.so
[...]

PR #1540, opened 2025-11-14
[C] Fix: `cuvsRMMMemoryResourceReset` sets a valid resource

ldematte:
In `cuvsRMMMemoryResourceReset`, we currently call `rmm::mr::set_current_device_resource(nullptr)`; but this generates segfaults (more specifically, a tentative to dereference a NULL pointer)

We should either save the original resource we get inside `cuvsRMMPoolMemoryResourceEnable` (returning some state, e.g. via a new `cuvsRMMMemoryResource_t` -- which may be get rid of the thread local too, if we find it convenient), and pass it back to `cuvsRMMMemoryResourceReset` so it can restore it.

Or reset the resource to some initial, valid instance. 
The `new rmm::mr::reset_current_device_resource_ref()` (see https://github.com/rapidsai/rmm/blob/f965b7f98805e96ccd4fb3e7774f9b8e38ad2bdb/cpp/include/rmm/mr/per_device_resource.hpp#L466) does that.

This PR follows the second approach and replicates `reset_current_device_resource_ref` behaviour in `cuvsRMMMemoryResourceReset`, by resetting the current device resource to `rmm::mr::detail::initial_resource()`.

Fixes https://github.com/rapidsai/cuvs/issues/1454

### DIFF
--- a/c/include/cuvs/core/c_api.h
+++ b/c/include/cuvs/core/c_api.h
@@ -7,7 +7,6 @@
 
 #include <cuda_runtime.h>
 #include <dlpack/dlpack.h>
-#include <rapids_logger/log_levels.h>
 #include <stdbool.h>
 #include <stdint.h>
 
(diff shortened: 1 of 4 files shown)


examples/c/src/bruteforce_c_example.c
```diff
@@ -0,0 +1,159 @@
+
+#include <cuvs/neighbors/brute_force.h>
```
ldematte, line 6: @benfred I decided to retain the example program I added to repro the issue, but let me know if you want to remove it or move it somewhere else.

examples/c/src/bruteforce_c_example.c
```diff
@@ -0,0 +1,159 @@
...
+#include <stdint.h>
+
+static const char dataset[] = {
```
mythrocks: Just a moment, shouldn't this be:

```suggestion
static const float dataset[] = {
```

mythrocks: (Am I reading this incorrectly?)

mythrocks: It hadn't occurred to me that this would compile in C-lang.  I'd expect the C++ compiler to catch this.

ldematte: You are right, it should be float, but it should also be valid C... well, let's not be obscure here :D

examples/c/src/bruteforce_c_example.c
```diff
@@ -0,0 +1,159 @@
...
+#include <stdint.h>
+
+static const float dataset[] = {
+    0.74021935f, 0.9209938f,
+    0.03902049f, 0.9689629f,
+    0.92514056f, 0.4463501f,
+    0.6673192f, 0.10993068f
+  };
+  
+static const char queries[] = {
```
ldematte: This needs changing too
```suggestion
static const float queries[] = {
```

mythrocks:
Ah, I'm not on @rapidsai/cuvs-cpp-codeowners.  @benfred:  This might need a quick gander.

ldematte:
@robertmaynard @benfred Rebased on release/25.12, now the changeset is clean

mythrocks:
Hi, @ldematte.  

I'll mention this here before @cjnolet does:  For RAPIDS projects, the recommendation is not to do force pushes, to prevent review comments from disappearing, etc.

ldematte:
@mythrocks I usually don't do that, but in case of a rebase to another branch (main -> 25.12 in this case) this is necessary (I think? If you know a better way, please correct me!)

mythrocks:
> in case of a rebase to another branch...

Fair point.  And in this case, the prior review comments remained intact, so there's no adverse consequence.

mythrocks:
Argh.  It looks like we need to run `pre-commit` to fix the formatting in the C test file.

ldematte:
That's very strange, I have installed the pre-commit hook. Maybe it got updated? Anyway, will run it

PR #1544, opened 2025-11-17
[Java] Relax cuVS version matching

ldematte:
As discussed during a sync, we want to relax version matching/checking between cuvs-java and the cuVS C library.
The rationale is that the C API should be pretty stable, and the cuvs-java bindings to the C API should not be affected by the cuVS version, provided that the API does not change.
For these reasons, we do still want to check that
- the C library version is at least as recent as the Java module
- the C library is not "too far" in the future; we do not want unlimited forward compatibility, we want to be able to change the C API at some point (after e.g. a deprecation cycle).

This PR introduces these checks, allowing cuvs-java to start and bind with C libraries up to 3 more cuVS releases in the future.

### DIFF
--- a/java/cuvs-java/src/main/java22/com/nvidia/cuvs/spi/JDKProvider.java
+++ b/java/cuvs-java/src/main/java22/com/nvidia/cuvs/spi/JDKProvider.java
@@ -444,7 +444,10 @@ final class JDKProvider implements CuVSProvider {
   @Override
   public CuVSMatrix.Builder<CuVSDeviceMatrix> newDeviceMatrixBuilder(
       CuVSResources resources, long size, long columns, CuVSMatrix.DataType dataType) {
-    return new DeviceMatrixBuilder(resources, size, columns, dataType);
+    var rowBytes = columns * dataType.bytes();
+    return rowBytes > PinnedMemoryBuffer.CHUNK_BYTES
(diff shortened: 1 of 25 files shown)


java/cuvs-java/src/main/java22/com/nvidia/cuvs/spi/JDKProvider.java
```diff
@@ -34,6 +34,73 @@
...
+      }
+
+      return cuvsVersion.major < maxVersion.major
+          || (cuvsVersion.major == maxVersion.major && cuvsVersion.minor <= maxVersion.minor);
+    }
+
+    private static CuVSVersion addReleases(CuVSVersion currentVersion, int numberOfReleases) {
+      short candidateMinor = (short) (currentVersion.minor + numberOfReleases * 2);
+      short releaseMinor = (short) (candidateMinor % 12);
+      short releaseMajor = (short) (currentVersion.major + (candidateMinor / 12));
```
ChrisHegarty, line 89: CalVer cyclic !

mythrocks (approved):

mythrocks:
CI failures are a result of https://github.com/rapidsai/raft/pull/2813.  Should be resolved by https://github.com/rapidsai/raft/pull/2881.

PR #1561, opened 2025-11-18
Deduplicate KMeans instantiations

divyegala:
Closes https://github.com/rapidsai/cuvs/issues/1514

CUDA 13 binary size from 464 MB to 437 MB.
CUDA 12 binary size from 1165 MB to 1102 MB.


cjnolet:
@divyegala this is a critical change AFAIC. Did you close this because of the 26.02 issue? Can we get a new one open if you don't intent to continue on this branch?

PR #1564, opened 2025-11-18
Forward merge 25.12

vyasr:
A redo of #1562. No diff, but we need the commits integrated onto main.


rockhowse:
Ack that this should NOT be squashed merged 🫡

PR #1566, opened 2025-11-18
Update FAISS patch for RMM memory resource header migration

bdice:
Updates FAISS patch for RMM memory resource header migration.

xref: https://github.com/rapidsai/rmm/issues/2141


cpp/cmake/patches/faiss_override.json
```diff
@@ -6,8 +6,8 @@
       "git_tag": "v1.12.0",
```
tarang-jain, line 6: Actually they just released v1.13.0. Can we tag that one?

bdice: Discussed offline. We're going to merge this as-is to get `main` unblocked, then tackle 1.13 in a follow-up PR.

PR #1570, opened 2025-11-19
Add arm64 builds to the libcuvs_c rocky8 matrix

robertmaynard:
Extend the `rocky8-clib-standalone-build` to include arm64 builds

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -218,9 +218,7 @@ jobs:
       arch: "${{matrix.arch}}"
       date: ${{ inputs.date }}_c
       container_image: "rapidsai/ci-wheel:26.02-cuda${{ matrix.cuda_version }}-rockylinux8-py3.10"
-      env: |
-        RAPIDS_TEST_ARCH=${{ matrix.arch }}
-      script: "ci/test_standalone_c.sh"
+      script: "ci/test_standalone_c.sh libcuvs_c_${{ matrix.cuda_version }}_${{ matrix.arch }}.tar.gz"
       sha: ${{ inputs.sha }}
   conda-java-build-and-tests:
     needs: [conda-cpp-build, changed-files]
--- a/ci/test_standalone_c.sh
+++ b/ci/test_standalone_c.sh
@@ -7,17 +7,23 @@ set -euo pipefail
 rapids-pip-retry install cmake
 pyenv rehash
 
+INSTALL_PREFIX="${PWD}/libcuvs_c_install"
+mkdir -p "${INSTALL_PREFIX}"
+
 # Download the standalone C library artifact
-payload_name="libcuvs_c_${RAPIDS_CUDA_VERSION}_${RAPIDS_TEST_ARCH}.tar.gz"
+if [ -z "$1" ]; then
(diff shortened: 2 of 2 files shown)


.github/workflows/pr.yaml
```diff
@@ -12,21 +12,21 @@ jobs:
...
-      - conda-cpp-build
-      - conda-cpp-tests
-      - conda-cpp-checks
-      - conda-python-build
-      - conda-python-tests
+      # - conda-cpp-build
+      # - conda-cpp-tests
+      # - conda-cpp-checks
+      # - conda-python-build
+      # - conda-python-tests
```
KyleFromNVIDIA: Why has this and everything else been commented out?

robertmaynard: That needs to be dropped. I had commented out everything else to reduce CI load

ci/test_standalone_c.sh
```diff
@@ -17,17 +17,20 @@ if [ ! -e "/usr/local/bin/cmake" ]; then
...
-DOWNLOAD_LOCATION=$(rapids-download-from-github "${payload_name}")
-
-# Extract the artifact to a staging directory
 INSTALL_PREFIX="${PWD}/libcuvs_c_install"
 mkdir -p "${INSTALL_PREFIX}"
-ls -l "${DOWNLOAD_LOCATION}"
-tar -xf "${DOWNLOAD_LOCATION}/${pkg_name}" -C "${INSTALL_PREFIX}"
+
+# Download the standalone C library artifact
+if [[ -n "${1}" ]]; then
```
KyleFromNVIDIA: Since this argument is effectively required due to the `ctest` invocation below, I believe it should be an error if the argument doesn't exist.

.github/workflows/pr.yaml
```diff
@@ -212,7 +212,7 @@ jobs:
       arch: "amd64"
       date: ${{ inputs.date }}_c
       container_image: "rapidsai/ci-wheel:26.02-cuda${{ matrix.cuda_version }}-rockylinux8-py3.10"
-      script: "ci/test_standalone_c.sh"
+      script: "ci/test_standalone_c.sh libcuvs_c_${{ matrix.cuda_version }}_${{ matrix.arch }}.tar.gz"
```
jameslamb, line 221: This doesn't accomplish what you want. Notice that `arch: "amd64"` is hard-coded a few lines up, and as an input not in the `matrix:`.

You'll need to move `arch` up into the `strategy.matrix:` block, like this:

```yaml
matrix:
  cuda_version:
    - *latest_cuda12
    - *latest_cuda13
  arch:
     - amd64
     - arm64
```

GitHub Actions will take care of generating the 4 different combinations of those inputs.

ci/test_standalone_c.sh
robertmaynard: Good point. Updated to error out when no arguments are provided

.github/workflows/pr.yaml
robertmaynard: Updated, how does that look now?

jameslamb: yep that's looking like what I'd expected, nice.

PR #1597, opened 2025-11-29
Add HNSW ACE build method

julianmi:
This PR adds a direct `hnsw::build` API that uses the ACE (Augmented Core Extraction) algorithm to build HNSW indexes on the GPU. ACE enables building HNSW indexes for datasets too large to fit in GPU memory by partitioning the data and building sub-indexes.

CC @tfeher 

**C++ API**
- Added `hnsw::build()` function with ACE parameters for direct HNSW index construction. This serializes an HNSW index to disk if `use_disk` is true.
- Added `hnsw::graph_build_params::ace_params` struct with configurable options:
  - `npartitions` - number of partitions for parallel build
  - `ef_construction` - index quality parameter
  - `build_dir` - directory for disk-based build artifacts
  - `use_disk` - force disk-based storage mode
- Implemented proper serialization/deserialization for disk-backed HNSW indexes
- Added C++ tests in `ann_hnsw_ace.cuh`

**C API**
- Added `cuvsHnswBuild` function with ACE parameters
- Added C tests in `ann_hnsw_ace.cu`

**Python**
- Added `hnsw.AceParams` class for configuring ACE builds
- Added Python tests in `test_hnsw_ace.py`

**Java**
- Added `HnswAceParams` class
- Added Java tests in `HnswAceBuildAndSearchIT.java`

**Documentation**
- Added `cuvs_hnsw` section to the parameter tuning guide with ACE parameters

**Example**
- Added `hnsw_ace_example.cu` demonstrating the build → deserialize → search workflow


cpp/include/cuvs/neighbors/hnsw.hpp
```diff
@@ -51,6 +56,30 @@ struct index_params : cuvs::neighbors::index_params {
       is parallelized with the help of CPU threads.
    */
   int num_threads = 0;
+
+  /** HNSW M parameter: number of bi-directional links per node (used when building with ACE).
+   *  graph_degree = m * 2, intermediate_graph_degree = m * 3.
```
tfeher: I propose to avoid discussing `ACE` and `graph_degree` / `intermediate_graph_degree` params here. These are implementation details.

cpp/include/cuvs/neighbors/hnsw.hpp
```diff
@@ -51,6 +56,30 @@ struct index_params : cuvs::neighbors::index_params {
...
+   */
+  size_t m = 32;
+
+  /** Parameters for graph building (ACE algorithm).
+   *
+   * Set ace_params to configure ACE (Augmented Core Extraction) parameters for building
+   * a GPU-accelerated HNSW index. ACE enables building indexes for datasets too large
+   * to fit in GPU memory.
```
tfeher: I prefer if we have something like that:

> Parameters to fine tune GPU graph building. By default we select the parameters based on dataset shape and HNSW build parameters. We can also adjust these parameters (refer to cagra build docs).

cpp/include/cuvs/neighbors/hnsw.hpp
```diff
@@ -158,6 +187,107 @@ struct extend_params {
   int num_threads = 0;
 };
 
+/**
+ * @}
+ */
+
+/**
+ * @defgroup hnsw_cpp_index_build Build HNSW index using ACE algorithm
```
tfeher:
```suggestion
 * @defgroup hnsw_cpp_index_build Build HNSW index on the GPU
```

examples/cpp/src/hnsw_ace_example.cu
```diff
@@ -11,34 +11,37 @@
...
-  index_params.intermediate_graph_degree = 128;
-  index_params.graph_degree              = 64;
+  // HNSW index parameters
+  hnsw::index_params hnsw_params;
+  hnsw_params.metric    = cuvs::distance::DistanceType::L2Expanded;
+  hnsw_params.hierarchy = hnsw::HnswHierarchy::GPU;
 
-  // ACE index parameters
-  auto ace_params = cagra::graph_build_params::ace_params();
+  // ACE index parameters for building HNSW directly
```
tfeher:
```suggestion
  // Parameters for GPU accelerated HNSW index building
```

examples/cpp/src/hnsw_ace_example.cu
```diff
@@ -51,90 +54,65 @@ void cagra_build_search_ace(raft::device_resources const& dev_resources,
...
+  hnsw_params.graph_build_params = ace_params;
+
+  // Build the HNSW index using ACE
+  std::cout << "Building HNSW index using ACE" << std::endl;
+  auto hnsw_index =
+    hnsw::build(dev_resources, hnsw_params, raft::make_const_mdspan(dataset_host.view()));
+
+  // For disk-based indexes, the build function serializes the index to disk
+  // We need to deserialize it before searching
+  std::string hnsw_index_path = "/tmp/hnsw_ace_build/hnsw_index.bin";
```
tfeher: Can we query the file path from `hnsw_index`?

examples/cpp/src/hnsw_ace_example.cu
```diff
@@ -51,90 +54,65 @@ void cagra_build_search_ace(raft::device_resources const& dev_resources,
...
+  std::cout << "Deserializing HNSW index from disk for search" << std::endl;
+  hnsw::index<float>* hnsw_index_deserialized = nullptr;
+  hnsw::deserialize(dev_resources,
```
tfeher, line 79: This would fail if we have too large index, that does not fit host mem, right? Could you add error handling?

cpp/include/cuvs/neighbors/hnsw.hpp
```diff
@@ -158,6 +187,107 @@ struct extend_params {
...
+ * @{
+ */
+
+/**
+ * @brief Build an HNSW index using the ACE (Augmented Core Extraction) algorithm
+ *
+ * ACE enables building HNSW indices for datasets too large to fit in GPU memory by:
```
tfeher:
```suggestion
 * @brief Build an HNSW index on the GPU
 *
 * The resulting graph is compatible for HNSW search, but is not an exact equivalent of the graph 
 * built by the HNSW.
 *
 * The HNSW index construction parameters `M` and `ef_construction` are the main parameters to
 * control the graph degree and graph quality.  We have additional options that can be used to fine 
 * tune graph building on the GPU (`link cagra doc`). In case the index does not fit the host or GPU
 * memory,  we would use disk as temporary storage. In such cases it is important to set
 * `ace_params.build_dir` to a fast disk with sufficient storage size.
 
```

examples/cpp/src/hnsw_ace_example.cu
```diff
@@ -11,34 +11,37 @@
...
+  raft::copy(dataset_host.data_handle(),
+             dataset.data_handle(),
+             dataset.extent(0) * dataset.extent(1),
+             raft::resource::get_cuda_stream(dev_resources));
+  raft::resource::sync_stream(dev_resources);
 
-  // CAGRA index parameters
-  cagra::index_params index_params;
-  index_params.intermediate_graph_degree = 128;
-  index_params.graph_degree              = 64;
```
tfeher, line 38: Please set parameter `M` explicitly, to show how the graph degree can be controlled.

examples/cpp/src/hnsw_ace_example.cu
tfeher, line 1: I would prefer if we keep a `cagra_hnsw_ace_example.cu`, where we discuss how to fine tune cagra graph building, and a simpler `hnsw_example.cu` where we set HNSW params and temp file location. I am not sure about showing `npartitions` in the `hnsw_example.cu`, since we will want to automate setting that.

examples/cpp/src/hnsw_ace_example.cu
```diff
@@ -51,90 +54,65 @@ void cagra_build_search_ace(raft::device_resources const& dev_resources,
   ace_params.ef_construction = 120;
```
tfeher: Somewhat confusing that we have two ways to set this. I would prefer to use `hnsw::index_params.ef_construction`.

cpp/include/cuvs/neighbors/hnsw.hpp
```diff
@@ -158,6 +187,107 @@ struct extend_params {
...
+ *   hnsw::serialize(res, "index.bin", *hnsw_index);
+ * @endcode
+ */
+std::unique_ptr<index<float>> build(
+  raft::resources const& res,
+  const index_params& params,
+  raft::host_matrix_view<const float, int64_t, raft::row_major> dataset);
+
+/**
+ * @brief Build an HNSW index using the ACE algorithm (half precision)
```
tfeher: Don't we need to repeat the full docstring so that we have the documentation available for each input type?

cpp/tests/neighbors/ann_hnsw_ace.cuh
```diff
@@ -0,0 +1,226 @@
...
+        // Clean up deserialized index
+        delete deserialized_index;
+      } else {
+        hnsw::search(handle_,
+                     search_params,
+                     *hnsw_index,
+                     queries_host.view(),
+                     indexes_hnsw_host.view(),
+                     distances_hnsw_host.view());
+      }
```
tfeher: I recommend to test two separate things:
1. In memory index is searchable
2. We can serialize, deserialize, search regardless whether disk was used for index building

```suggestion
      if (!ps.use_disk) {
        hnsw::search(handle_,
                     search_params,
                     *hnsw_index,
                     queries_host.view(),
                     indexes_hnsw_host.view(),
                     distances_hnsw_host.view());
          // check results
      }
        tmp_index_file index_file;
        hnsw::serialize(handle_, index_file.filename, *hnsw_index);
        
        hnsw::index<DataT>* deserialized_index = nullptr;
        hnsw::deserialize(handle_, hnsw_params, index_file.filename, ps.dim, ps.metric, &deserialized_index);
        ASSERT_NE(deserialized_index, nullptr);
        hnsw::search(handle_,
                     search_params,
                     *deserialized_index,
                     queries_host.view(),
                     indexes_hnsw_host.view(),
                     distances_hnsw_host.view());

        // Clean up deserialized index
        delete deserialized_index;

```

cpp/include/cuvs/neighbors/hnsw.hpp
```diff
@@ -158,6 +187,107 @@ struct extend_params {
...
+ *
+ *   // Create index parameters with ACE configuration
+ *   hnsw::index_params params;
+ *   params.metric = cuvs::distance::DistanceType::L2Expanded;
+ *   params.hierarchy = hnsw::HnswHierarchy::GPU;
+ *
+ *   // Configure ACE parameters
+ *   auto ace_params = hnsw::graph_build_params::ace_params();
+ *   ace_params.npartitions = 4;
+ *   ace_params.ef_construction = 120;
```
tfeher: Can we pick this up from `hnsw::index_params` `ef_constructions`?

tfeher (changes requested):
Thanks Julian for the PR! I think it is a nice improvement, but I would still recommend to simplify further along the following lines
- Treat CAGRA parameters as implementation details. We want to keep the usage and the documentation simple for users familiar with HNSW index building.
- Still provide a way to adjust these parameters.
- Clarify that the results are not an exact equivalent to a HNSW graph, but in practice the graph is a good replacement that works well in HNSW search.

cpp/include/cuvs/neighbors/hnsw.hpp
julianmi: This makes sense to me. I removed the details but kept the general description of the `M` parameter.

cpp/include/cuvs/neighbors/hnsw.hpp
julianmi: Done.

examples/cpp/src/hnsw_ace_example.cu
julianmi: Yes, with `hnsw_index->file_path()`. I have changed this.

examples/cpp/src/hnsw_ace_example.cu
julianmi: I have added an OOM check for `hnsw::deserialize()`.

examples/cpp/src/hnsw_ace_example.cu
julianmi: Done.

examples/cpp/src/hnsw_ace_example.cu
julianmi: I have restored `cagra_hnsw_ace_example.cu`. I kept `npartitions` for now. I will remove it in the next PR adding the heuristic.

examples/cpp/src/hnsw_ace_example.cu
julianmi: Done.

cpp/include/cuvs/neighbors/hnsw.hpp
julianmi: Good point. I have copied the full docstring here and below.

cpp/include/cuvs/neighbors/hnsw.hpp
julianmi: Done.

cpp/tests/neighbors/ann_hnsw_ace.cuh
julianmi: Applied.

cpp/include/cuvs/neighbors/hnsw.hpp
```diff
@@ -51,6 +56,31 @@ struct index_params : cuvs::neighbors::index_params {
       is parallelized with the help of CPU threads.
    */
   int num_threads = 0;
+
+  /** HNSW M parameter: number of bi-directional links per node (used when building with ACE).
+   */
+  size_t m = 32;
```
tfeher: I suggest to keep this capitalized

```suggestion
  size_t M = 32;
```

cpp/src/neighbors/detail/hnsw.hpp
```diff
@@ -1196,12 +1212,69 @@ void deserialize(raft::resources const& res,
...
+  }
+}
+
+/**
+ * @brief Build an HNSW index using the ACE algorithm
+ *
+ * This function builds an HNSW index using ACE (Augmented Core Extraction) by:
+ * 1. Converting HNSW parameters to CAGRA parameters with ACE configuration
+ * 2. Building a CAGRA index using ACE
+ * 3. Converting the CAGRA index to HNSW format
```
tfeher:
```suggestion
 * @brief Build an HNSW index on the GPU using CAGRA graph building algorithm
 *
 * This function builds an HNSW index
 * 1. Converting HNSW parameters to CAGRA parameters (ACE configuration by default)
 * 2. Building a CAGRA index
 * 3. Converting the CAGRA index to HNSW format
```

cpp/src/neighbors/detail/hnsw.hpp
```diff
@@ -1196,12 +1212,69 @@ void deserialize(raft::resources const& res,
...
+ */
+template <typename T>
+std::unique_ptr<index<T>> build(raft::resources const& res,
+                                const index_params& params,
+                                raft::host_matrix_view<const T, int64_t, raft::row_major> dataset)
+{
+  common::nvtx::range<common::nvtx::domain::cuvs> fun_scope("hnsw::build<ACE>");
+
+  // Validate that ACE parameters are set
+  RAFT_EXPECTS(std::holds_alternative<graph_build_params::ace_params>(params.graph_build_params),
```
tfeher: The default value given to `graph_build_params` in `include/neighbors/hnsw.hpp` is `std::monostate`. This means the program execution would fail with default parameters here. That should not happen. I would recommend to set up default `ace_params` in case `std::monostate` is used, and continue.

c/include/cuvs/neighbors/hnsw.h
```diff
@@ -49,6 +98,17 @@ struct cuvsHnswIndexParams {
       is parallelized with the help of CPU threads.
   */
   int num_threads;
+  /** HNSW M parameter: number of bi-directional links per node (used when building with ACE).
+   *  graph_degree = m * 2, intermediate_graph_degree = m * 3.
+   */
+  size_t m;
```
tfeher:
```suggestion
  size_t M;
```

cpp/include/cuvs/neighbors/hnsw.hpp
```diff
@@ -51,6 +56,31 @@ struct index_params : cuvs::neighbors::index_params {
...
+   * a GPU-accelerated HNSW index. ACE enables building indexes for datasets too large
+   * to fit in GPU memory.
+   *
+   * @code{.cpp}
+   * hnsw::index_params params;
+   * // Configure ACE parameters
+   * params.graph_build_params = hnsw::graph_build_params::ace_params();
+   * auto& ace = std::get<hnsw::graph_build_params::ace_params>(params.graph_build_params);
+   * ace.npartitions = 4;
+   * ace.ef_construction = 120;
```
tfeher: We have `hnsw::index_params::ef_construction` and `ace_params::ef_construction`. We need to clarify which one to use. I recommend to use `hnsw::index_params::ef_construction`.

cpp/src/neighbors/detail/hnsw.hpp
```diff
@@ -1196,12 +1212,69 @@ void deserialize(raft::resources const& res,
...
+  // Create CAGRA index parameters from HNSW parameters
+  cuvs::neighbors::cagra::index_params cagra_params;
+  cagra_params.metric                    = params.metric;
+  cagra_params.intermediate_graph_degree = params.m * 3;
+  cagra_params.graph_degree              = params.m * 2;
+
+  // Configure ACE parameters for CAGRA
+  cuvs::neighbors::cagra::graph_build_params::ace_params cagra_ace_params;
+  cagra_ace_params.npartitions     = ace_params.npartitions;
+  cagra_ace_params.ef_construction = ace_params.ef_construction;
```
tfeher:
```suggestion
  cagra_ace_params.ef_construction = params.ef_construction;
```
It can still lead to confusion that we are taking the value from `hnsw::index_params`. Maybe we should check whether `ace_params.ef_construction` value is different than the default `ef_construction` value, and inform the user that it is not the right way to change it?

python/cuvs/cuvs/neighbors/hnsw/hnsw.pyx
```diff
@@ -50,20 +116,34 @@ cdef class IndexParams:
         NOTE: When hierarchy is `gpu`, while the majority of the work is done
         on the GPU, initialization of the HNSW index itself and some other
         work is parallelized with the help of CPU threads.
+    m : int, default = 32 (optional)
```
tfeher:
```suggestion
    M : int, default = 32 (optional)
```

python/cuvs/cuvs/neighbors/hnsw/hnsw.pyx
```diff
@@ -347,6 +445,94 @@ def from_cagra(IndexParams index_params, cagra.Index cagra_index,
...
+    >>>
+    >>> n_samples = 50000
+    >>> n_features = 50
+    >>> dataset = np.random.random_sample((n_samples, n_features),
+    ...                                   dtype=np.float32)
+    >>>
+    >>> # Create ACE parameters
+    >>> ace_params = hnsw.AceParams(
+    ...     npartitions=4,
+    ...     ef_construction=120,
```
tfeher: `ef_construction` value should be set at the right place.

tfeher (changes requested):
Thanks @julianmi for the updates! There are only two issues remaining:
1. Redundancy with `ef_construction` parameter (see comments below)
2. Build - deserialize - search workflow

Considering the second point, currently the resulting index can be either in memory or stored in a file. (Although #1604 removes the in-memory ACE path, we might still decide later to enable other in-memory build methods.)
```c++
auto hnsw_index = hnsw::build(res, hnsw_params, dataset);
// index is saved to disk at hnsw_index->file_path(), the hnsw_index structure just a wrapper around the file name and index params
hnsw::search(res, search_params, *hnsw_index, queries, indices, distances);
```
This would fail with the following error message if the index does not fit memory. 
```
Searching HNSW index
terminate called after throwing an instance of 'raft::logic_error'
  what():  RAFT failure at file=/home/scratch.tfeher_gpu_2/cuvs_1597/cpp/src/neighbors/detail/hnsw.hpp line=1140: Cannot search an HNSW index that is stored on disk. The index must be deserialized into memory first using hnsw::deserialize().
Obtained 7 stack frames
```

Instead, when the index is on disk, then we need to use the following step
```c++
auto hnsw_index = hnsw::build(res, hnsw_params, dataset);
hnsw::deserialize(res, hnsw_params, hnsw_index->file_path(), dataset.extent(1), hnsw_params.metric, &hnsw_index_deserialized);
hnsw::search(res, search_params, *hnsw_index_deserialized, queries, indices, distances);
```

Do we have to dictate such usage patterns? Could we instead let `hnsw::search` load the index from file if needed?

cpp/include/cuvs/neighbors/hnsw.hpp
julianmi: Capitalized `M` throughout.

cpp/src/neighbors/detail/hnsw.hpp
julianmi: Applied.

cpp/src/neighbors/detail/hnsw.hpp
julianmi: Applied. This will require automatically deriving the number of partitions in #1603 though.

cpp/include/cuvs/neighbors/hnsw.hpp
julianmi: Sorry I missed some references in `ace_params`. Only `index_params::ef_construction` should be used.

cpp/src/neighbors/detail/hnsw.hpp
julianmi: I have removed `ace_params.ef_construction` altogether.

python/cuvs/cuvs/neighbors/hnsw/hnsw.pyx
julianmi: Done.

tfeher (approved):
Thanks @julianmi for the updates, the PR looks good to me!

c/include/cuvs/neighbors/cagra.h
```diff
@@ -127,10 +127,10 @@ typedef struct cuvsIvfPqParams* cuvsIvfPqParams_t;
 
 /**
  * Parameters for ACE (Augmented Core Extraction) graph build.
- * ACE enables building indices for datasets too large to fit in GPU memory by:
+ * ACE enables building indexes for datasets too large to fit in GPU memory by:
```
mythrocks, line 130: 👍

c/include/cuvs/neighbors/cagra.h
```diff
@@ -251,22 +251,6 @@ cuvsError_t cuvsAceParamsCreate(cuvsAceParams_t* params);
...
- */
-cuvsError_t cuvsAceParamsCreate(cuvsAceParams_t* params);
-
-/**
- * @brief De-allocate ACE params
- *
- * @param[in] params
- * @return cuvsError_t
- */
-cuvsError_t cuvsAceParamsDestroy(cuvsAceParams_t params);
```
robertmaynard, line 268: I don't believe we should be removing the public definition of these. For example how does someone delete `cuvsAceParams_t` without `cuvsAceParamsDestroy`?

robertmaynard (changes requested):

c/include/cuvs/neighbors/cagra.h
julianmi: Thanks for flagging this. I think these were duplicated by mistake (see lines above) and I removed them accordingly.

robertmaynard (approved):

divyegala (approved):
Python approval only

mythrocks (approved):
Approving the Java bit.

PR #1598, opened 2025-12-01
[Improvement] Allow Configurable DRAM in DiskANN SSD Wrapper

tarang-jain:
Notes: During the final param parsing to DiskANN index building API, we convert the dram budgets to GB


cpp/bench/ann/src/diskann/diskann_wrapper.h
```diff
@@ -178,6 +180,8 @@ class diskann_ssd : public algo<T> {
     uint32_t QD                   = 192;
     std::string dataset_base_file = "";
     std::string index_file        = "";
+    uint32_t build_dram_budget    = std::numeric_limits<uint32_t>::max();
+    uint32_t search_dram_budget   = std::numeric_limits<uint32_t>::max();
```
mythrocks: Is the budget to be specified in megabytes?  I'm not familiar with this, but we seem to be dividing by `1024.0f` to get to a gigabyte value.

It would be good to include the units in the variable name.  `search_dram_budget_megabytes` or similar.

cjnolet: I 100% agree with this. And we really should be documenting each and every parameter as much as possible (both for users and for future developer eyes).

tarang-jain: Fixed in 5aee439

PR #1600, opened 2025-12-01
Ivf_flat extends golang APIs

cpegeric:
Add golang ivf_flat Apis for GetNLists, GetDim and GetCenters.


go/brute_force/brute_force.go
```diff
@@ -57,6 +58,8 @@ func BuildIndex[T any](Resources cuvs.Resource, Dataset *cuvs.Tensor[T], metric
 		return err
 	}
 	index.trained = true
+	runtime.KeepAlive(Dataset.C_tensor)
```
ajit283: `C_tensor` is memory managed by C (through `malloc`/`free`), not by the Go runtime. So this should not have any effect. Did you add this due to some error/bug?

cpegeric: No, this is just for safety.  If you think this is not necessary, you can ignore it. thanks

PR #1603, opened 2025-12-02
Automatic Partition Count Derivation for ACE 

julianmi:
This PR adds automatic partition count derivation for ACE (Augmented Core Extraction) graph builds based on available system memory.

Previously, users had to manually calculate and specify the number of partitions based on their dataset size and available memory. This was error-prone and required understanding the internal memory requirements of the ACE algorithm. With auto-derivation, users get optimal partitioning out of the box while still having the option to override if needed.

**Changes**
- When `npartitions` is set to `0` (new default), ACE automatically calculates the optimal number of partitions based on available host and GPU memory.
- When `npartitions` is set to a positive value, the specified count is used but may be automatically increased if it would exceed memory limits.
- Added `max_host_memory_gb` and `max_gpu_memory_gb` parameters to allow users to constrain memory usage (useful for shared systems or testing).

This builds on top of PR #1597, which should be merged first.


KyleFromNVIDIA (approved):
Approved CMake changes

cpp/src/neighbors/detail/cagra/cagra_build.cuh
```diff
@@ -792,6 +796,241 @@ void ace_load_partition_dataset_from_disk(
...
+      "Per-partition breakdown: dataset %.2f GiB, graph %.2f GiB",
+      to_gib(disk_mode_host_required),
+      to_gib(mem.available_host_memory),
+      to_gib(mem.sub_dataset_size),
+      to_gib(mem.sub_graph_size));
+
+    // Calculate suggested number of partitions for host memory
+    host_suggested_partitions = static_cast<size_t>(
+      std::ceil(disk_mode_host_required / (usable_cpu_memory_fraction * mem.available_host_memory) *
+                n_partitions));
```
tfeher: `disk_mode_host_required` includes partition labels and id_mapping size estimate. I believe multiplying that those terms with n_partitions is not necessary. 
```suggestion
    std::ceil((mem.sub_dataset_size + mem.sub_graph_size) * n_partitions /  
        (usable_cpu_memory_fraction * mem.available_host_memory - mem.partition_labels_size -
         mem.id_mapping_size));
```
We should probable check that the result is positive (i.e. denominator is positive).

cpp/src/neighbors/detail/cagra/cagra_build.cuh
```diff
@@ -792,6 +796,241 @@ void ace_load_partition_dataset_from_disk(
...
+    host_suggested_partitions = std::max(host_suggested_partitions, n_partitions + 1);
```
tfeher, line 1023: Why `n_partitions + 1`?

cpp/src/neighbors/detail/cagra/cagra_build.cuh
```diff
@@ -792,6 +796,241 @@ void ace_load_partition_dataset_from_disk(
...
+  // Check GPU memory requirements in disk mode
+  // GPU memory model for disk mode (per-partition processing):
+  //   - Per-partition dataset on GPU: 4 * (dataset_size / n_partitions) * dataset_dim * sizeof(T)
+  //   - Per-partition graph on GPU: (dataset_size / n_partitions) * (intermediate + final) *
+  //   sizeof(IdxT)
+  //   - CAGRA build also requires workspace memory. Conservative estimate: dataset size.
+  size_t gpu_sub_dataset_size =
+    imbalance_factor * 2 * (dataset_size / n_partitions) * dataset_dim * sizeof(T);
+  size_t gpu_sub_graph_size = imbalance_factor * 2 * (dataset_size / n_partitions) *
+                              (intermediate_degree + graph_degree) * sizeof(IdxT);
```
tfeher: These are the same as `mem.sub_dataset_size` and `mem.sub_graph_size`, right? Let's use them instead of recalculating.

cpp/src/neighbors/detail/cagra/cagra_build.cuh
```diff
@@ -792,6 +796,241 @@ void ace_load_partition_dataset_from_disk(
...
+  size_t gpu_workspace_size     = gpu_sub_dataset_size;
+  size_t disk_mode_gpu_required = gpu_sub_dataset_size + gpu_sub_graph_size + gpu_workspace_size;
```
tfeher: Why do we need a workspace that is equivalent with `sub_datset_size`?

We should also keep in mind that memory requirements depend on the build algorithm that we use. Having `gpu_sub_dataset_size + gpu_sub_graph_size` is a good upper limit now. Just for reference, I expect that we have the following actual memory usage
- IVF-PQ `max(pq_compressed_sub_dataset_size, gpu_sub_graph_size)`
- NN descent `gpu_sub_dataset_size_fp16 + gpu_sub_graph_size`
- Iterative solver `max(gpu_sub_dataset_size, gpu_sub_graph_size)`

cpp/src/neighbors/detail/cagra/cagra_build.cuh
```diff
@@ -792,6 +796,241 @@ void ace_load_partition_dataset_from_disk(
...
+      to_gib(disk_mode_gpu_required),
+      to_gib(mem.available_gpu_memory),
+      to_gib(gpu_sub_dataset_size),
+      to_gib(gpu_sub_graph_size),
+      to_gib(gpu_workspace_size));
+
+    gpu_suggested_partitions = static_cast<size_t>(
+      std::ceil(disk_mode_gpu_required / (usable_gpu_memory_fraction * mem.available_gpu_memory) *
+                n_partitions));
+    gpu_suggested_partitions = std::max(gpu_suggested_partitions, n_partitions + 1);
```
tfeher, line 1050: Why +1?

cpp/src/neighbors/detail/cagra/cagra_build.cuh
```diff
@@ -792,6 +796,241 @@ void ace_load_partition_dataset_from_disk(
...
+  // Check if GPU has enough memory for the final graph or use disk mode instead.
+  // TODO: Extend model or use managed memory if running out of GPU memory.
+  if (max_gpu_memory_gb.has_value() && max_gpu_memory_gb.value() > 0) {
+    mem.available_gpu_memory = static_cast<size_t>(max_gpu_memory_gb.value() * (1ULL << 30));
+    RAFT_LOG_INFO("ACE: Using overridden GPU memory limit: %.2f GiB", max_gpu_memory_gb.value());
+  } else {
+    mem.available_gpu_memory = rmm::available_device_memory().second;
+  }
+  bool gpu_memory_limited = static_cast<size_t>(usable_gpu_memory_fraction *
+                                                mem.available_gpu_memory) < mem.cagra_graph_size;
```
tfeher: The whole CAGRA graph does not need to fit GPU mem (if `n_partitions` > 1). I think we need something like that
```suggestion
  bool gpu_memory_limited = static_cast<size_t>(usable_gpu_memory_fraction *
                                                mem.available_gpu_memory) < std::max(mem.sub_graph_size, mem.sub_dataset_size);
```

python/cuvs/cuvs/neighbors/hnsw/hnsw.pyx
```diff
@@ -38,16 +38,33 @@ cdef class AceParams:
...
-        per partition is recommended depending on available memory.
+    npartitions : int, default = 0 (optional)
+        Number of partitions for ACE partitioned build. When set to 0 (default),
+        the number of partitions is automatically derived based on available
+        host and GPU memory to maximize partition size while ensuring the build
+        fits in memory.
+
+        Small values might improve recall but potentially degrade performance.
+        100k - 5M vectors per partition is recommended depending on available
+        memory.
```
tfeher: I am not sure if we should make these recommendations. We are still exploring what are the best values.

tfeher (changes requested):
Thanks Julian for the PR, it is great to have the `n_partition` parameter automatically determined. I have reviewed C/C++/Python changes. I have suggestions for the formulas used for memory estimation, but in general the PR is in a good shape.

cpp/src/neighbors/detail/cagra/cagra_build.cuh
```diff
@@ -865,74 +1104,23 @@ index<T, IdxT> build_ace(raft::resources const& res,
   try {
```
tfeher, line 1168: Commenting for line 1079:
```C++
  RAFT_EXPECTS(n_partitions > 0, "ACE: npartitions must be greater than 0");
```
Now `n_partitions == 0` is a valid input so we can just set n_partitions = 2 and let the heuristic below figure out the right value.

cpp/src/neighbors/detail/cagra/cagra_build.cuh
julianmi: Good catch, thanks. I have update the code.

cpp/src/neighbors/detail/cagra/cagra_build.cuh
julianmi: This is inside the `host_memory_insufficient` check meaning that with `n_partitions` we expect to run out of memory. `n_partitions + 1` guarantees that we lower the memory requirements even in the edge case of rounding `host_suggested_partitions` to `n_partitions` and running out of memory again. Does this make sense?

cpp/src/neighbors/detail/cagra/cagra_build.cuh
julianmi: Done.

cpp/src/neighbors/detail/cagra/cagra_build.cuh
julianmi: Let me know if you have a better estimate of `gpu_workspace_size` please. The limiter I have found seems to be `build_knn_graph` calling `sort_knn_graph` which creates a copy of the dataset ([graph_core.cuh#L532-L537](https://github.com/rapidsai/cuvs/blob/main/cpp/src/neighbors/detail/cagra/graph_core.cuh#L532-L537)).

Thanks for these limits. This is very helpful.

cpp/src/neighbors/detail/cagra/cagra_build.cuh
julianmi: See above.

cpp/src/neighbors/detail/cagra/cagra_build.cuh
julianmi: Good catch, thanks.

python/cuvs/cuvs/neighbors/hnsw/hnsw.pyx
julianmi: I have removed these recommendations throughout until we have more data.

cpp/src/neighbors/detail/cagra/cagra_build.cuh
julianmi: Done.

cpp/src/neighbors/detail/cagra/cagra_build.cuh
tfeher: Sounds good, thanks!

cpp/src/neighbors/detail/cagra/cagra_build.cuh
tfeher: Indeed sorting would need the dataset. But that is done after KNN building, so we reuse the space that we reserved for the dataset during that phase. Therefore I would not include dataset size in workspace.

I went through `optimize`, and here is the workspace memory usage. Could you define a helper function, and include this in the memory estimate?

```python
def optimize_workspace_size(N, deg, ideg, S, mst_opt = False):
    """ Claculates CAGRA optimize memory usage

    This is the working memory on top of the input/output host memory usage (N * (deg + ideg) * S).

    N - number of rows in the dataset
    deg - graph degree
    ideg - intermediate graph degree
    S - graph type size
    """

    mst_host = N * S                       # mst_graph_num_edges
    if mst_opt:
        mst_host += N * deg * S            # mst_graph allocated in optimize
        mst_host += N * deg * S            # mst_graph allocated in mst_optimize
        mst_host += N * S * 7              # vectors with _max_edges suffix
        mst_host += (deg-1) * (deg-1) * S  #  iB_candidates 
  

    prune_host = N * ideg * 1  # detour count

    prune_dev = N * ideg * 1   # detour count
    prune_dev += N * 4         # d_num_detour_edges
    prune_dev += N * ideg * S  # d_input_graph
    # We neglect 8 bytes (both on host and device) for stats

    rev_host = N * deg * S     # rev_graph
    rev_host += N * 4          # rev_graph_count
    rev_host += N * S          # dest_nodes

    rev_dev = N * deg * S      # d_rev_graph
    rev_dev += N * 4           # d_rev_graph_count
    rev_dev += N * 4           # d_dest_nodes

    # Memory for merging graphs
    combine_host = (N * 4 + deg * 4) / 1e9  # in_edge_count + hist

    mst_host /= 1e9
    prune_host /= 1e9;
    prune_dev /= 1e9;
    rev_host /= 1e9;
    rev_dev /= 1e9;
    
    print("Prune host {:4.2f} GB, dev {:4.2f} GB".format(prune_host, prune_dev))
    print("Rev   host {:4.2f} GB, dev {:4.2f} GB".format(rev_host, rev_dev))
    print("MST   host {:4.2f} GB".format(mst_host))

    
    total_host = mst_host + max(prune_host, rev_host, combine_host)
    total_dev =  max(prune_dev, rev_dev)
    print("Total host {:4.2f} GB, dev {:4.2f} GB".format(total_host, total_dev))

    return total_host, total_dev
```

tfeher:
Thanks Julian for the update, I had one more comment for the workspace size estimate, otherwise the code looks good.

cpp/src/neighbors/detail/cagra/cagra_build.cuh
julianmi: Thank you, this is very helpful. I have added a helper which is used in host and device memory calculations.

tfeher (changes requested):
Thanks Julian for the updates! Although the memory size estimates could be still improved, I believe it is better to tweak them in a follow up PR. The rest of the PR look good to me.

tfeher (approved):
My last review meant to approve the PR.

python/cuvs/cuvs/neighbors/cagra/cagra.pyx
```diff
@@ -169,13 +184,17 @@ cdef class AceParams:
             check_cuvs(cuvsAceParamsDestroy(self.params))
 
     def __init__(self, *,
-                 npartitions=1,
+                 npartitions=0,
                  ef_construction=120,
                  build_dir="/tmp/ace_build",
```
benfred, line 189: not related to this PR: but I don't think we should be hardcoding the `build_dir` like this, and instead should be using something like `tempfile.TemporaryDirectory` .  The user might not have `/tmp` be a valid directory, and this could cause issues with concurrency when two programs both try to access the same directory.

benfred (approved):
python/java changes look good!

PR #1605, opened 2025-12-02
Use CCCL's mdspan implementation

divyegala:
Based on https://github.com/rapidsai/raft/pull/2836

### DIFF
--- a/ci/build_go.sh
+++ b/ci/build_go.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-# SPDX-FileCopyrightText: Copyright (c) 2024, NVIDIA CORPORATION.
+# SPDX-FileCopyrightText: Copyright (c) 2024-2025, NVIDIA CORPORATION.
 # SPDX-License-Identifier: Apache-2.0
 
 set -euo pipefail
--- a/ci/build_rust.sh
+++ b/ci/build_rust.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-# SPDX-FileCopyrightText: Copyright (c) 2024, NVIDIA CORPORATION.
+# SPDX-FileCopyrightText: Copyright (c) 2024-2025, NVIDIA CORPORATION.
 # SPDX-License-Identifier: Apache-2.0
 
 set -euo pipefail


ci/build_cpp.sh
```diff
@@ -23,6 +23,7 @@ export RAPIDS_PACKAGE_VERSION
 RAPIDS_ARTIFACTS_DIR=${RAPIDS_ARTIFACTS_DIR:-"${PWD}/artifacts"}
 mkdir -p "${RAPIDS_ARTIFACTS_DIR}"
 export RAPIDS_ARTIFACTS_DIR
+source ./ci/use_conda_packages_from_prs.sh
```
bdice: Reminder to revert before merging.

dantegd (approved):

cpp/include/cuvs/neighbors/ivf_pq.hpp
```diff
@@ -2896,7 +2894,7 @@ void set_centers(raft::resources const& res,
  */
 void extract_centers(raft::resources const& res,
                      const index<int64_t>& index,
-                     raft::device_matrix_view<float, uint32_t, raft::row_major> cluster_centers);
+                     raft::device_matrix_view<float, int64_t, raft::row_major> cluster_centers);
```
tarang-jain, line 2897: We might need to change the signatures of ivf-pq getters to int64_t instead of uint32_t, right? For example things like `list_sizes`.

divyegala: Can you share a link reference?

tarang-jain: yes so we have `n_lists()` for example https://github.com/rapidsai/cuvs/blob/ba67db12dcd2b4f894e9b0693c11e85dff4c3626/cpp/include/cuvs/neighbors/ivf_pq.hpp#L406

Do we want to change this signature to `int64_t`?

tarang-jain: In other words, what I am trying to say is that we might be considering an index-wide migration to `int64_t` from `uint32_t` for all extent types in IVF-PQ.

tarang-jain: @tfeher had made of aware of plans for this, but maybe that is outside the scope of this PR.

divyegala: `n_lists()` returns `lists_.size()` and `lists_` is an `std::vector` https://github.com/rapidsai/cuvs/blob/ba67db12dcd2b4f894e9b0693c11e85dff4c3626/cpp/include/cuvs/neighbors/ivf_pq.hpp#L501

It should be returning `size_t`. Anyway, the extent type updates in this PR are not related to `index<IdxT>` where `IdxT` is the index type, not the extent type.

PR #1607, opened 2025-12-02
[FEA] IVF-PQ to Write Flat PQ Codes

tarang-jain:
This PR brings new params to ivf_pq: an option for the user to choose the layout of the ivf lists. The lists can be flat (no interleaving) or interleaved (current default). Flat codes allows building the index in a CPU-compatible format.

[UPDATE as of 12/19/2025]:
After #1278 is merged, we can unify IVF-PQ and PQ API codepaths.

[UPDATE 01/08/2026]:
This PR can be merged before #1278. The flat code-writing can potentially be reverted once #1278 is merged (so we can later use the PQ preprocessing API directly). However that will come naturally as a part if a broader unification of IVF-PQ and PQ codepaths.

[Benchmarks 01/15/2026]:
## IVF-PQ Layout Benchmark Results

**Dataset**: 1,000,000 vectors × 128 dimensions | **pq_dim**: 32

pq_bits | Code Size | Direct FLAT Build (ms) | INTERLEAVED Build (ms) | Convert INTERLEAVED to FLAT with Codepacker (ms) | Total time for INTERLEAVED build + Conversion to FLAT with Codepacker (unpack) (ms) | Overhead |
|:-------:|:---------:|:---------------:|:----------------------:|:-------------------:|:----------------------:|:--------:|
| 8 | 32 bytes | 372.46 | 385.86 | 985.28 | 1371.13 | 3.68× |
| 6 | 24 bytes | 298.83 | 300.99 | 961.82 | 1262.82 | 4.23× |
| 5 | 20 bytes | 283.25 | 281.95 | 795.43 | 1077.38 | 3.80× |
| 4 | 16 bytes | 270.63 | 271.01 | 489.73 | 760.75 | 2.81× |

### DIFF
--- a/go/ivf_pq/index_params.go
+++ b/go/ivf_pq/index_params.go
@@ -21,8 +21,8 @@ const (
 )
 
 var cCodebookKinds = map[codebookKind]int{
-	Subspace: C.PER_SUBSPACE,
-	Cluster:  C.PER_CLUSTER,
+	Subspace: C.CUVS_IVF_PQ_CODEBOOK_GEN_PER_SUBSPACE,
+	Cluster:  C.CUVS_IVF_PQ_CODEBOOK_GEN_PER_CLUSTER,
 }
 
 // Creates a new IndexParams


tarang-jain:
@lowener I believe https://github.com/rapidsai/cuvs/pull/1278 can be used directly to do this, rather than writing a new kernel, correct? If so, we can wait for #1278 to be merged and then just call those APIs from IVF-PQ. I also so the comments in https://github.com/rapidsai/cuvs/issues/1648 related to this.

cpp/src/neighbors/ivf_pq/ivf_pq_codepacking.cuh
```diff
@@ -165,11 +167,62 @@ __device__ void write_vector(
...
+  const uint32_t lane_id = raft::Pow2<SubWarpSize>::mod(threadIdx.x);
+
+  uint8_t* out_vec_ptr = out_flat_codes + out_ix * bytes_per_vector;
+
+  bitfield_view_t<PqBits> code_view{out_vec_ptr};
+  for (uint32_t j = 0; j < pq_dim; j++) {
+    uint8_t code = action(in_ix, j);
+    if (lane_id == 0) { code_view[j] = code; }
+  }
+}
```
tarang-jain, line 218: Small optimization opportunity here: We can do the `lane_id == 0` check at the very beginning and simply return if that condition is not true. Because the code writing is done only by the thread with lane_id = 0.

tarang-jain:
/ok to test https://github.com/rapidsai/cuvs/commit/3bf9018fdc606a665c2fa028eba40abbd4a59d3c

c/include/cuvs/neighbors/ivf_pq.h
```diff
@@ -114,6 +123,14 @@ struct cuvsIvfPqIndexParams {
...
    */
   uint32_t max_train_points_per_pq_code;
+  /**
+   * Memory layout of the IVF-PQ list data.
+   *
+   * - FLAT: Codes are stored contiguously, one vector's codes after another.
+   * - INTERLEAVED: Codes are interleaved for optimized search performance.
+   *   This is the default and recommended for search workloads.
+   */
+  enum list_layout codes_layout;
```
cjnolet: Just making a note here that this is going to break the C ABI. No action needed, but we will need to track these things in the future. I'll share the POR with you (and the team) for more info.

cpp/include/cuvs/neighbors/common.hpp
```diff
@@ -683,11 +683,50 @@ template <typename IdxT>
...
+
+  /** Get the indices pointer. */
+  virtual index_type* indices_ptr() noexcept             = 0;
+  virtual const index_type* indices_ptr() const noexcept = 0;
+
+  /** Get the current size (number of records). */
+  virtual size_type get_size() const noexcept = 0;
+
+  /** Set the current size (number of records). */
+  virtual void set_size(size_type new_size) noexcept = 0;
```
cjnolet, line 716: I suppose as long as users are never direclty mutating this object then we shouldn't have to be too vigilent. Would be nice if we could make this internal, though, so that only cuVS funcitons are capable of changing it. Just concerned for users shooting themselves in the foot.

cpp/src/neighbors/ivf_pq/ivf_pq_codepacking.cuh
lowener: No, all the threads are used to compute that code in parallel (see [encode_vectors function](https://github.com/rapidsai/cuvs/blob/efe7c978d0aac628cbbf2222de26c525d5393311/cpp/src/neighbors/ivf_pq/ivf_pq_process_and_fill_codes.cuh#L115-L125)). At the end the minimum of all the threads is shuffled so that the lane 0 gets the smallest one.
Returning at the beginning will mean that just 1 out of the 32 threads of the warp will do the work,

c/include/cuvs/neighbors/ivf_pq.h
```diff
@@ -28,6 +28,15 @@ enum codebook_gen {  // NOLINT
   PER_CLUSTER  = 1,  // NOLINT
 };
 
+/**
+ * @brief A type for specifying the memory layout of IVF-PQ list data
+ *
+ */
+enum list_layout {   // NOLINT
+  FLAT        = 0,   // NOLINT
```
robertmaynard: We should prefix this with CUVS. @benfred

tarang-jain: do we want to prefix the enum itself?
`enum cuvsListLayout`?

cjnolet (approved):
LGTM, pending addressing @robertmaynard's feedback. LanceDB is chomping at the bit to try this out, so we can continue to improve over time if we need to (and also use the upcoming preprocessing API directly in IVFPQ in the future).

c/include/cuvs/neighbors/ivf_pq.h
tarang-jain: enums are fixed.

cpp/include/cuvs/neighbors/common.hpp
tarang-jain: Tracking here https://github.com/rapidsai/cuvs/issues/1726

PR #1608, opened 2025-12-02
Add rust bindings for vamana index

benfred:


bkarsin (approved):
Looks good, but the SSD codebook functionality is missing (i.e., `deserialize_codebooks`). This is also missing from the C API, though, so I created a separate issue for this (https://github.com/rapidsai/cuvs/issues/1629)

cjnolet (approved):

Issue #1578, opened 2025-11-20
Reduce Binary Size by deduplicating `{unpack/pack}_list_data_kernel`


jinsolp:
There are multiple instantiations of `cuvs::neighbors::ivf_pq::detail::unpack_list_data_kernel` and `cuvs::neighbors::ivf_pq::detail::pack_list_data_kernel` kernels.

PR #1609, opened 2025-12-03
Deduplicate `{unpack/pack}_list_data_kernel`

jinsolp:
Closes https://github.com/rapidsai/cuvs/issues/1578

This PR refactors the code so that we have pre-compiled code for launching `{unpack/pack}_list_data_kernel`.

`.cuh`: with declarations for including in other files
`*_impl.cuh`, `*.cu`: actual implementation

CUDA 12: 1104.44 MB -> 1100.26 MB
CUDA 13: 437.47 MB -> 435.32 MB

### DIFF
--- a/cpp/src/neighbors/ivf_pq/ivf_pq_build.cuh
+++ b/cpp/src/neighbors/ivf_pq/ivf_pq_build.cuh
@@ -10,7 +10,7 @@
 #include "../ivf_list.cuh"
 #include "ivf_pq_codepacking.cuh"
 #include "ivf_pq_contiguous_list_data.cuh"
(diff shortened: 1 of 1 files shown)


cpp/src/neighbors/ivf_pq/ivf_pq_list_data.hpp
divyegala, line 1: This can be a `.hpp` file as it does not contain any CUDA compilation elements.

jinsolp: done!

PR #1612, opened 2025-12-03
Script to generate billion scale synthetic data

singhmanas1:
This script generates billion-scale synthetic data. Following is the procedure-
Dataset Generation-
1. Generate centroids using k-means over randomly generated 1M vectors.
2. Create a Gaussion blob for each centroid using varying standard deviation.
Ground Truth Generation-
1. Randomly sample query vectors from dataset (from each cluster) -- remove these IDs from the dataset.
2. Use cuvs brute force to identify 4 closest centroids
3. Do a brute force search over the 4 clusters to identify top-k ground truth.

User Configurations-
1. Total vectors
2. Total vectors per cluster
3. Vector dim and dtype
4. Top-k (for ground truth)


singhmanas1:
Time to generate 10B vectors with 1 L4 (AWS g6.16xlarge)
Screenshot 2025-12-04 at 3 22 14 AM

cjnolet:
Closing in favor of @jinsolp's ongoing work.

PR #1614, opened 2025-12-03
[FEA]  Enforce tighter link restrictions on libcuvs_c

robertmaynard:
This makes sure we don't leak unneeded dependencies in our `PUBLIC` target_link_libraries for cuvs_c

### DIFF
--- a/examples/c/CMakeLists.txt
+++ b/examples/c/CMakeLists.txt
@@ -32,20 +32,30 @@ set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} -Wall -Wextra -Werror")
 
 add_executable(CAGRA_C_EXAMPLE src/cagra_c_example.c)
 target_include_directories(CAGRA_C_EXAMPLE PUBLIC "$<BUILD_INTERFACE:${DLPACK_INCLUDE_DIR}>")
-target_link_libraries(CAGRA_C_EXAMPLE PRIVATE cuvs::c_api CUDA::cudart $<TARGET_NAME_IF_EXISTS:conda_env>)
+target_link_libraries(
+  CAGRA_C_EXAMPLE PRIVATE cuvs::c_api CUDA::cudart $<TARGET_NAME_IF_EXISTS:conda_env>
+)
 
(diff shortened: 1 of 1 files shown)


c/tests/CMakeLists.txt
```diff
@@ -44,6 +44,8 @@ function(ConfigureTest)
   add_executable(${TEST_NAME} ${_CUVS_TEST_PATH})
   target_link_libraries(
     ${TEST_NAME} PRIVATE cuvs::c_api GTest::gtest GTest::gtest_main
+                         $<$<BOOL:${CUVSC_STATIC_CUVS_LIBRARY}>:$<COMPILE_ONLY:cuvs::cuvs_static>>
+                         $<$<NOT:$<BOOL:${CUVSC_STATIC_CUVS_LIBRARY}>>:cuvs::cuvs>
```
KyleFromNVIDIA:
```suggestion
                         $<IF:$<BOOL:${CUVSC_STATIC_CUVS_LIBRARY}>,$<COMPILE_ONLY:cuvs::cuvs_static>,cuvs::cuvs>
```

c/CMakeLists.txt
```diff
@@ -155,8 +155,10 @@ target_include_directories(
 
 target_link_libraries(
   cuvs_c
-  PUBLIC $<IF:$<BOOL:${CUVSC_STATIC_CUVS_LIBRARY}>,cuvs::cuvs_static,cuvs::cuvs>
-  PRIVATE raft::raft $<TARGET_NAME_IF_EXISTS:hnswlib::hnswlib>
+  PRIVATE # we don't want any dependencies exported
+    $<IF:$<BOOL:${CUVSC_STATIC_CUVS_LIBRARY}>,cuvs::cuvs_static,cuvs::cuvs> #
```
divyegala: Trailing comment tag

KyleFromNVIDIA (changes requested):

c/CMakeLists.txt
robertmaynard: it is to stop the formatter from collapsing it with the next line which should be separate as the comment only applies to it

PR #1617, opened 2025-12-04
Use HNSW GPU Hierarchy by Default

julianmi:
Change the default HNSW hierarchy from `NONE` to `GPU` to enable search with Hnswlib. Searching an HNSW index built with `HnswHierarchy::NONE` using Hnswlib fails with a segmentation fault. This change should go into 25.12 to prevent such issues and improve the interoperability between cuVS and Hnswlib.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -53,7 +53,7 @@ jobs:
       matrix:
         cuda_version:
           - &latest_cuda12 '12.9.1'
-          - &latest_cuda13 '13.1.0'
+          - &latest_cuda13 '13.1.1'
         arch:
           - amd64
           - arm64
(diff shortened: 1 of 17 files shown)


tfeher (approved):
Thanks Julian for fixing this, LGTM!

divyegala:
@tfeher this is a breaking change, and it's also missing an update to the C API default.

julianmi:
> @tfeher this is a breaking change, and it's also missing an update to the C API default.

I have changed the C API default as well.

> Instead of a segmentation fault, can you put an error checker to fail gracefully if user tries to search a NONE hierarchy HNSW index?

Sorry to not me more specific. Our internal search works but a separate search with Hnswlib fails. Changing the default would enable the interoperability out of the box.

tfeher:
> @tfeher this is a breaking change

Thanks @divyegala for catching my mistake. Indeed, users might be relying on the default value, so this is a breaking change. I have updated the labels.

Issue #1595, opened 2025-11-26
[BUG] CAGRA extend operation does not support float16 data type


TomaszGrabowskiPawlak:
## Bug Description

CAGRA `extend` operation does not support `float16` data type, while other CAGRA operations already support it.

## Related Issues/PRs

- #1419 - Initial request for float16 support in CAGRA
- #1429 - Added float16 support for CAGRA build/search operations

## Problem

PR #1429 added `float16` support to CAGRA `build` and `search` operations, but the `extend` operation was not updated. This creates an inconsistency where users can build and search indices with `float16`, but cannot extend them.

## Solution

Possible solution: https://github.com/TomaszGrabowskiPawlak/cuvs/pull/1/files

TomaszGrabowskiPawlak:
@lowener , would you be so kind to have a look at this issue ? I've noticed that cagra doesn't support extend with float16 type and vibe-coded a patch ( https://github.com/TomaszGrabowskiPawlak/cuvs/pull/1/files ) , which I am currently using in my pipelines. However, not being experienced with cpp and rapids, I don't think it's acceptable for me to create a pull request.

lowener:
It looks good, I created a PR starting from your work, and added some tests #1620

PR #1620, opened 2025-12-08
Add float16 support for CAGRA extend

lowener:
Related issue: #1595
CAGRA support extend for float32, int8, uint8. This PR adds float16. I conserved the commits from the original authors and added some tests on C++ and python sides.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - "main"
+      - "release/*"
     tags:
       - v[0-9][0-9].[0-9][0-9].[0-9][0-9]
   workflow_dispatch:
@@ -63,7 +64,7 @@ jobs:
       date: ${{ inputs.date }}
       container_image: "rapidsai/ci-wheel:26.04-cuda${{ matrix.cuda_version }}-rockylinux8-py3.10"
       node_type: "cpu16"
-      # requires_license_builder: false
+      requires_license_builder: true
       script: "ci/build_standalone_c.sh"
       artifact-name: "libcuvs_c_${{ matrix.cuda_version }}_${{ matrix.arch }}.tar.gz"
       file_to_upload: "libcuvs_c.tar.gz"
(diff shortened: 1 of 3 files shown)


cjnolet:
Thanks for spotting and fixing this @lowener. Just for completeness, can you share the increase in binary size that this change causes? Likely not. a blocker, but helps us for tracking to keep that info close to the changes. Thanks!

cpp/include/cuvs/neighbors/cagra.hpp
```diff
@@ -1242,7 +1242,83 @@ void extend(
  *
  * Usage example:
  * @code{.cpp}
- *   using namespace cuvs::neighbors;
+ *   using namespace raft::neighbors;
```
dantegd:
```suggestion
 *   using namespace cuvs::neighbors;
```

This should stay as cuvs, no?

cpp/include/cuvs/neighbors/cagra.hpp
```diff
@@ -1242,7 +1242,83 @@ void extend(
...
+  cuvs::neighbors::cagra::index<half, uint32_t>& idx,
+  std::optional<raft::device_matrix_view<half, int64_t, raft::layout_stride>>
+    new_dataset_buffer_view                                                        = std::nullopt,
+  std::optional<raft::device_matrix_view<uint32_t, int64_t>> new_graph_buffer_view = std::nullopt);
+
+/** @brief Add new vectors to a CAGRA index
+ *
+ * Usage example:
+ * @code{.cpp}
+ *   using namespace raft::neighbors;
```
lowener:
```suggestion
 *   using namespace cuvs::neighbors;
```

cpp/include/cuvs/neighbors/cagra.hpp
```diff
@@ -1242,7 +1242,83 @@ void extend(
...
```
lowener:
```suggestion
 *   using namespace cuvs::neighbors;
```

lowener:
Increase in binary size for all architecture: libcuvs.so went from 1096M to 1098M. So +2M

PR #1621, opened 2025-12-08
Only test hnsw cagra functions when hnsw is enabled

robertmaynard:
Builds with `BUILD_CAGRA_HNSWLIB=OFF` failed to link due to `CAGRA_C_TEST` assuming hnswlib was always usable, which is incorrect. 

Move the HNSWLIB related test to a new test unit


robertmaynard:
> Combine with ann_hnsw_c.cu

Done

PR #1627, opened 2025-12-09
Default to static linking of libcudart

bdice:
## Summary
- Enable static linking of libcudart by default (`CUDA_STATIC_RUNTIME=ON`)
- Remove `cuda-cudart` from conda recipe run requirements (no longer needed when statically linked)

This is part of a RAPIDS-wide effort to switch to static CUDA runtime linking. See https://github.com/rapidsai/build-planning/issues/235 for tracking.

## Changes
- `cpp/CMakeLists.txt`: Change `CUDA_STATIC_RUNTIME` default from OFF to ON
- `conda/recipes/cuvs/recipe.yaml`: Remove `cuda-cudart` from run deps
- `conda/recipes/libcuvs/recipe.yaml`: Remove `cuda-cudart` from run deps (4 outputs)

Note: Python builds already use `CUDA_STATIC_RUNTIME=ON` (set in `python/libcuvs/CMakeLists.txt`).

### DIFF
--- a/c/tests/CMakeLists.txt
+++ b/c/tests/CMakeLists.txt
@@ -49,7 +49,6 @@ function(ConfigureTest)
             GTest::gtest_main
             $<IF:$<BOOL:${CUVSC_STATIC_CUVS_LIBRARY}>,$<COMPILE_ONLY:cuvs::cuvs_static>,cuvs::cuvs>
             $<TARGET_NAME_IF_EXISTS:conda_env>
-            rmm::rmm
   )
   set_target_properties(
     ${TEST_NAME}


conda/recipes/libcuvs/recipe.yaml
```diff
@@ -79,7 +79,6 @@ cache:
       - libraft-headers =${{ minor_version }}
       - nccl ${{ nccl_version }}
       - cuda-version =${{ cuda_version }}
-      - cuda-cudart-dev
```
bdice, line 82: I would have expected that we still need the `-dev` package for its headers, but I don't remember what I did with other RAPIDS libraries.

dependencies.yaml
```diff
@@ -357,6 +358,11 @@ dependencies:
           - matrix:
             packages:
               - cuda-toolkit[cublas,curand,cusolver,cusparse,nvjitlink]>=12,<14
+  depends_on_cudart:
```
bdice, line 361: Why not tack this onto the `cuda` list?

conda/recipes/libcuvs/recipe.yaml
KyleFromNVIDIA: Indeed we do need it, I'll put it back in.

dependencies.yaml
KyleFromNVIDIA: Only Rust and Go need it at build time, I didn't want to put it in the other environments.

KyleFromNVIDIA: Looking at it now though, I see that `dlpack.go` uses stuff from `cuda_runtime_api.h`. Perhaps the dependency should be on `cuda-cudart-dev` instead.

cpp/CMakeLists.txt
```diff
@@ -746,6 +744,12 @@ if(NOT BUILD_CPU_ONLY)
   if(NOT cuvs_compile_mode STREQUAL "static_only")
     add_library(cuvs SHARED $<TARGET_OBJECTS:cuvs_objs> $<TARGET_OBJECTS:cuvs-cagra-search>)
     add_library(cuvs::cuvs ALIAS cuvs)
+    get_property(
+      rt
+      TARGET cuvs
+      PROPERTY CUDA_RUNTIME_LIBRARY
+    )
+    message(STATUS "cuvs CUDA_RUNTIME_LIBRARY: ${rt}")
```
KyleFromNVIDIA: Remove debug statement:

```suggestion
```

cpp/CMakeLists.txt
```diff
@@ -808,6 +812,12 @@ SECTIONS
   if(NOT cuvs_compile_mode STREQUAL "shared_only")
     add_library(cuvs_static STATIC $<TARGET_OBJECTS:cuvs_objs> $<TARGET_OBJECTS:cuvs-cagra-search>)
     add_library(cuvs::cuvs_static ALIAS cuvs_static)
+    get_property(
+      rt
+      TARGET cuvs_static
+      PROPERTY CUDA_RUNTIME_LIBRARY
+    )
+    message(STATUS "cuvs_static CUDA_RUNTIME_LIBRARY: ${rt}")
```
KyleFromNVIDIA: Remove debug statement:

```suggestion
```

KyleFromNVIDIA:
`libcuvs.so` is still getting `CUDA::cudart` rather than `CUDA::cudart_static` due to a leak from `rmm::rmm`. Relevant excerpt from `lib/cmake/rmm/rmm-targets.cmake`:

```cmake
# Create imported target rmm::rmm
add_library(rmm::rmm SHARED IMPORTED)

set_target_properties(rmm::rmm PROPERTIES
  INTERFACE_INCLUDE_DIRECTORIES "${_IMPORT_PREFIX}/include"
  INTERFACE_LINK_LIBRARIES "CUDA::cudart;CCCL::CCCL;dl;nvtx3::nvtx3-cpp;rapids_logger::rapids_logger"
  INTERFACE_POSITION_INDEPENDENT_CODE "ON"
)
```

rust/cuvs-sys/build.rs
```diff
@@ -15,8 +15,12 @@ fn main() {
         "cargo:rustc-link-search=native={}/lib",
         cuvs_build.display()
     );
+    println!(
+        "cargo:rustc-link-search=native={}/lib",
+        env::var("CONDA_PREFIX").unwrap()
+    );
```
benfred: The `unwrap` here will panic if `CONDA_PREFIX` environment variable isn't set .  How about something like 

```suggestion
    if let Ok(conda_prefix) = env::var("CONDA_PREFIX") {
        println!("cargo:rustc-link-search=native={}/lib", conda_prefix);
    }
```

instead?

KyleFromNVIDIA: Done

c/tests/CMakeLists.txt
```diff
@@ -49,6 +49,7 @@ function(ConfigureTest)
             GTest::gtest_main
             $<IF:$<BOOL:${CUVSC_STATIC_CUVS_LIBRARY}>,$<COMPILE_ONLY:cuvs::cuvs_static>,cuvs::cuvs>
             $<TARGET_NAME_IF_EXISTS:conda_env>
+            rmm::rmm
```
KyleFromNVIDIA: It's not entirely clear to me why this is needed now. AFAICT, we've always been building with `CUVSC_STATIC_CUVS_LIBRARY=ON`, so the [link error](https://github.com/rapidsai/cuvs/actions/runs/23020725689/job/66963224901#step:13:1004) should have been triggered before this PR.

KyleFromNVIDIA: And now we're getting runtime link errors due to missing `librmm.so`: https://github.com/rapidsai/cuvs/actions/runs/23055523149/job/66970964103?pr=1627

Issue #1243, opened 2025-08-12
Support `nnz_t` for spectral_embedding


aamijar:
In order to support larger datasets it would be useful to have `nnz_t` type instead of `int`

Hopefully, this will help solve https://github.com/rapidsai/cuml/issues/6910.

PR #1628, opened 2025-12-10
Spectral Embedding `nnz_t`

aamijar:
Resolves https://github.com/rapidsai/cuvs/issues/1243. Depends on https://github.com/rapidsai/raft/pull/2891.

This PR adds a `NNZType` to the spectral embedding public api with precomputed connectivity graph.
The transform api for the precomputed connectivity graph has been switched to use the COO codepath all the way through the algorithm.

The spectral embedding api which passes in a dataset also has been switched to use the COO codepath and use int64_t by default.

### DIFF
--- a/cpp/src/preprocessing/spectral/detail/spectral_embedding.cuh
+++ b/cpp/src/preprocessing/spectral/detail/spectral_embedding.cuh
@@ -30,43 +30,6 @@
 
 namespace cuvs::preprocessing::spectral_embedding::detail {
 
-template <typename DataT>
-raft::device_csr_matrix_view<DataT, int, int, int> coo_to_csr_matrix(
-  raft::resources const& handle,
-  const int n_samples,
-  raft::device_vector_view<int> sym_coo_row_ind,
-  raft::device_coo_matrix_view<DataT, int, int, int> sym_coo_matrix_view)
-{
-  auto stream = raft::resource::get_cuda_stream(handle);
-
-  raft::sparse::op::coo_sort<DataT>(n_samples,
-                                    n_samples,
-                                    sym_coo_matrix_view.structure_view().get_nnz(),
-                                    sym_coo_matrix_view.structure_view().get_rows().data(),
-                                    sym_coo_matrix_view.structure_view().get_cols().data(),
-                                    sym_coo_matrix_view.get_elements().data(),
-                                    stream);
-
-  raft::sparse::convert::sorted_coo_to_csr(sym_coo_matrix_view.structure_view().get_rows().data(),
-                                           sym_coo_matrix_view.structure_view().get_nnz(),
-                                           sym_coo_row_ind.data_handle(),
-                                           n_samples,
-                                           stream);
-
-  auto sym_coo_nnz = sym_coo_matrix_view.structure_view().get_nnz();
-  raft::copy(sym_coo_row_ind.data_handle() + sym_coo_row_ind.size() - 1, &sym_coo_nnz, 1, stream);
-
-  auto csr_matrix_view = raft::make_device_csr_matrix_view<DataT, int, int, int>(
-    const_cast<DataT*>(sym_coo_matrix_view.get_elements().data()),
-    raft::make_device_compressed_structure_view<int, int, int>(
-      const_cast<int*>(sym_coo_row_ind.data_handle()),
-      const_cast<int*>(sym_coo_matrix_view.structure_view().get_cols().data()),
-      n_samples,
-      n_samples,
-      sym_coo_matrix_view.structure_view().get_nnz()));
-  return csr_matrix_view;
-}
-
 template <typename DataT, typename OutSparseMatrixType, typename InSparseMatrixViewType>
 OutSparseMatrixType create_laplacian(raft::resources const& handle,
                                      params spectral_embedding_config,


cpp/src/preprocessing/spectral/detail/spectral_embedding.cuh
```diff
@@ -160,23 +159,79 @@ void compute_eigenpairs(raft::resources const& handle,
...
   const int n_samples = connectivity_graph.structure_view().get_n_rows();
 
   auto sym_coo_row_ind = raft::make_device_vector<int>(handle, n_samples + 1);
   auto diagonal        = raft::make_device_vector<DataT, int>(handle, n_samples);
 
-  auto csr_matrix_view =
-    coo_to_csr_matrix(handle, n_samples, sym_coo_row_ind.view(), connectivity_graph);
-  auto laplacian =
-    create_laplacian(handle, spectral_embedding_config, csr_matrix_view, diagonal.view());
+  // raft::print_device_vector("connectivity_graph_vals", connectivity_graph.get_elements().data(),
```
cjnolet: Please don't leave commented out code in your changes

cpp/cmake/thirdparty/get_raft.cmake
```diff
@@ -60,8 +60,8 @@ endfunction()
 # To use a different RAFT locally, set the CMake variable
 # CPM_raft_SOURCE=/path/to/local/raft
 find_and_configure_raft(VERSION  ${RAFT_VERSION}.00
-        FORK                     ${RAFT_FORK}
-        PINNED_TAG               ${RAFT_PINNED_TAG}
+        FORK                     aamijar
+        PINNED_TAG               laplacian-coo-input
```
aamijar: Undo later

cpp/src/preprocessing/spectral/detail/spectral_embedding.cuh
```diff
@@ -160,23 +154,29 @@ void compute_eigenpairs(raft::resources const& handle,
...
   auto diagonal        = raft::make_device_vector<DataT, int>(handle, n_samples);
 
-  auto csr_matrix_view =
-    coo_to_csr_matrix(handle, n_samples, sym_coo_row_ind.view(), connectivity_graph);
-  auto laplacian =
-    create_laplacian(handle, spectral_embedding_config, csr_matrix_view, diagonal.view());
+  auto laplacian = create_laplacian<DataT, raft::device_coo_matrix<DataT, int, int, NNZType>>(
+    handle, spectral_embedding_config, connectivity_graph, diagonal.view());
+
+  raft::sparse::op::coo_sort<DataT>(n_samples,
```
viclafargue:
```suggestion
  raft::sparse::op::coo_sort<DataT, int, NNZType>(n_samples,
```
Would be nice to be explicit here, just in case.

cpp/src/preprocessing/spectral/detail/spectral_embedding.cuh
```diff
@@ -264,10 +264,10 @@ void transform(raft::resources const& handle,
...
   auto csr_matrix_view =
     coo_to_csr_matrix<float>(handle, n_samples, sym_coo_row_ind.view(), sym_coo_matrix.view());
-  auto laplacian =
-    create_laplacian<float>(handle, spectral_embedding_config, csr_matrix_view, diagonal.view());
+  auto laplacian = create_laplacian<float, raft::device_csr_matrix<float, int, int, int>>(
+    handle, spectral_embedding_config, csr_matrix_view, diagonal.view());
   compute_eigenpairs<float>(
-    handle, spectral_embedding_config, n_samples, laplacian, diagonal.view(), embedding);
+    handle, spectral_embedding_config, n_samples, laplacian.view(), diagonal.view(), embedding);
 }
```
viclafargue: It looks like the connectivity graph in the transform function that takes the dataset as argument is assumed to have a nnz of type int. Is this intentional? Will it be updated in a follow-up PR?

aamijar: Good point, I'll try to change it so that it defaults to int64_t.

aamijar: Addressed in https://github.com/rapidsai/cuvs/pull/1628/commits/30ebe8ed587888197d1c64776a0b6815376b7c8b

viclafargue (commented):
Looks good for the most part. But, some checks might be necessary.

Could you review the following for risks of integer overflow?

`nnz` overflow :
[thrust::tabulate](https://github.com/rapidsai/cuvs/blob/a12e0091b01c041def9421c1b6ed7906e1391d7d/cpp/src/preprocessing/spectral/detail/spectral_embedding.cuh#L208-L211)

`n_samples*k_search` overflow :
[d_indices and d_distances allocation](https://github.com/rapidsai/cuvs/blob/a12e0091b01c041def9421c1b6ed7906e1391d7d/cpp/src/preprocessing/spectral/detail/spectral_embedding.cuh#L191-L192) (extents are provided as integers which may cause an overflow internally before allocation)

less likely `n_samples` overflow :
[config.max_iterations](https://github.com/rapidsai/cuvs/blob/a12e0091b01c041def9421c1b6ed7906e1391d7d/cpp/src/preprocessing/spectral/detail/spectral_embedding.cuh#L102C3-L102C24)

RAFT operators that may use container extents internally for indexing :

- raft::linalg::matrix_vector_op
- raft::matrix::gather
- raft::linalg::unary_op
- raft::matrix::fill

It looks like the `coo_to_csr_matrix` function is not used anymore. Should we delete it or make it compatible with larger nnz? In this case `sym_coo_row_ind` would probably have to be of the nnz type.

Also, why do we keep two versions of the function (for the two nnz types)? Is this for legacy support or are there some performance implications? If there is no performance implication we should probably only use the `uin64_t` nnz in cuML.

cpp/src/preprocessing/spectral/detail/spectral_embedding.cuh
```diff
@@ -88,11 +89,11 @@ raft::device_csr_matrix<DataT, int, int, int> create_laplacian(
   return laplacian;
 }
 
-template <typename DataT>
+template <typename DataT, typename A>
```
tarang-jain: Please use a more informative name here instead of `A`.

cpp/src/preprocessing/spectral/spectral_embedding.cu
```diff
@@ -1,25 +1,29 @@
...
+                 params config,                                                             \
+                 raft::device_coo_matrix_view<DataT, int, int, NNZType> connectivity_graph, \
+                 raft::device_matrix_view<DataT, int, raft::col_major> embedding)           \
+  {                                                                                         \
+    detail::transform<DataT, NNZType>(handle, config, connectivity_graph, embedding);       \
   }
 
-CUVS_INST_SPECTRAL_EMBEDDING(float);
-CUVS_INST_SPECTRAL_EMBEDDING(double);
+CUVS_INST_SPECTRAL_EMBEDDING(float, int);
```
tarang-jain, line 23: Do we need the `int` instantiations here? Or can we skip them and stick to `int64_t` only?

cpp/src/preprocessing/spectral/detail/spectral_embedding.cuh
aamijar: Addressed in https://github.com/rapidsai/cuvs/pull/1628/commits/3d962a302391122c37e064ae922627e35652822f

cpp/src/preprocessing/spectral/spectral_embedding.cu
aamijar: We can keep the `int` ones to avoid breaking cuml and remove them later.

aamijar:
Hi @viclafargue, thanks for the review! I have addressed your `int` overflow concerns in https://github.com/rapidsai/cuvs/pull/1628/commits/c81cb15fa410777b6ee244cfa8f8324e89bfbf59.

Yes, we can remove the `coo_to_csr` since that is no longer used. I was keeping it around for debugging purposes. Removed in https://github.com/rapidsai/cuvs/pull/1628/commits/54e3133fe107bdd9b8963e0308d333d6deeff839
Yes, we can only instantiate the `int64_t` nnz type functions and drop `int` but we should keep it for now to avoid breaking cuml.

aamijar:
> It could be interesting to see if we could handle the edge case of very large n_samples * config.n_components (matrix_vector_op & gather use) maybe as a follow-up PR though.

Hi @viclafargue, so that would mean we need to change the output `embedding` to be indexed with `int64_t` right? I think we can address that in a follow up PR. Lanczos solver uses `int` for embedding output currently.

cpp/src/preprocessing/spectral/spectral_embedding.cu
tarang-jain: Yes sounds good.

aamijar: Tracking here https://github.com/rapidsai/cuvs/issues/1695

PR #1640, opened 2025-12-15
Shorten the Test Duration of test_cagra_ace.py

julianmi:
I was made aware that the CAGRA ACE Python test was taking long during CI. This shortens the test duration by reducing the test combinations. Instead, variations of `npartitions`, `ef_construction`, and `hierarchy` are tested separately to keep the test coverage high. I have also aligned the test defaults with the existing CAGRA test.


bdice (approved):
Thanks for this! We should always aim for maximum coverage but GPU test runtime is also a significant constraint for team velocity and CI resource management.

tfeher (approved):
LGTM.

benfred (approved):

PR #1641, opened 2025-12-15
Remove Flaky Python ACE Example

julianmi:
The Python doc test shows sporadic segfaults in CI during `cagra.build`. This PR removes the problematic example. The CAGRA ACE build method is thoroughly tested in `test_cagra_ace.py`.

Additionally, the number of queries is used during `cagra.search`.

### DIFF
--- a/python/cuvs/cuvs/neighbors/cagra/cagra.pyx
+++ b/python/cuvs/cuvs/neighbors/cagra/cagra.pyx
@@ -511,26 +511,13 @@ def build(IndexParams index_params, dataset, resources=None):
     ...                                   dtype=cp.float32)
     >>> build_params = cagra.IndexParams(metric="sqeuclidean")
     >>> index = cagra.build(build_params, dataset)
+    >>> queries = cp.random.random_sample((n_queries, n_features),
+    ...                                   dtype=cp.float32)
     >>> distances, neighbors = cagra.search(cagra.SearchParams(),
-    ...                                      index, dataset,
+    ...                                      index, queries,
     ...                                      k)
     >>> distances = cp.asarray(distances)
     >>> neighbors = cp.asarray(neighbors)
-
-    >>> # ACE example with host data
-    >>> import numpy as np
-    >>> dataset_host = np.random.random_sample(
-    ...     (n_samples, n_features)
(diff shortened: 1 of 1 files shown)


python/cuvs/cuvs/neighbors/cagra/cagra.pyx
```diff
@@ -511,26 +511,13 @@ def build(IndexParams index_params, dataset, resources=None):
...
-    >>> dataset_host = np.random.random_sample(
-    ...     (n_samples, n_features)
-    ... ).astype(np.float32)
-    >>> ace_params = cagra.AceParams(
-    ...     npartitions=4, use_disk=True, build_dir="/tmp/ace"
-    ... )
-    >>> build_params = cagra.IndexParams(
-    ...     metric="sqeuclidean",
-    ...     build_algo="ace",
-    ...     ace_params=ace_params
```
tfeher, line 531: It would make sense to restore the original example without ACE

https://github.com/rapidsai/cuvs/blob/55a31fd5df23fdbc044a36507497fd4e8eadf080/python/cuvs/cuvs/neighbors/cagra/cagra.pyx#L355-L359

julianmi: It looks to me like the original author intended to use less queries during search given the definition of `n_queries`. This would also follow typical search behavior and would shorten the test time. Anyways, I did restore it to the example before CAGRA ACE was introduced.

tfeher: Hmm, sorry I was under the impression that removing the ACE build example removes all our build examples. Now I see that it is not the case. You are right, using a smaller query set was a actually a good change.

tfeher (approved):
LGTM.

Created an issue for tracking the failing code example: https://github.com/rapidsai/cuvs/issues/1647

python/cuvs/cuvs/neighbors/cagra/cagra.pyx
julianmi: Added back the changes which added the smaller query set.

benfred (approved):

Issue #1486, opened 2025-11-03
[FEA] CAGRA ACE Follow up tracker


tfeher:
#1404 implements a new build method for CAGRA indices. Here we list follow up issues that are out of scope of the original PR.

- [ ] Default build config should select ACE build when regular build would go OOM
- [ ] cuvs-bench param parser to accept `graph_build_algo: "ACE"` (and set other params automatically if needed)
- [ ]  Set `ace_params.npartitions `default value 0 for auto select?
- [ ] do we need to expose `build_knn_graph` for `ivf_pq` method only?
- [ ] does `get_device_for_address` (used in `src/neighbors/cagra.cuh`) work as intended on systems with AST or HMM?
- [ ] `hnsw::from_cagra` does only serialize the index when it is `on_disk`. Is this the right behavior?
- [ ] python API tests
- [ ] #1461 
- [ ] writing numpy headers for files is repetitive in cagra_build.cuh is repetitive, create better helpers.

julianmi:
I have removed the outdated Java and Python interfaces. If Java and Python support for ACE builds is needed, we will need to implement them based on the final APIs and add tests.

Edit Nov. 6: I have added interfaces that match the C interface and respective tests.

julianmi:
> hnsw::from_cagra does only serialize the index when it is on_disk. Is this the right behavior?

I propose to move to `hnsw::build` in https://github.com/rapidsai/cuvs/pull/1597. This would build a CAGRA index on disk using CAGRA ACE followed by serializing an HNSW index to disk. A user would deserialize this index to make it searchable. However, users might expect `hnsw::build` to be searchable. We could rename the API to something more explicit such as `hnsw::build_and_serialize`. @tfeher What do you think?

julianmi:
> Default build config should select ACE build when regular build would go OOM

We might want to wait for the ACE implementation to mature before switching automatically.

> cuvs-bench param parser to accept graph_build_algo: "ACE" (and set other params automatically if needed)

This was done in the original PR #1404.

> Set ace_params.npartitions default value 0 for auto select?

This is addressed in #1603.

> writing numpy headers for files is repetitive in cagra_build.cuh is repetitive, create better helpers.

This is addressed in #1645.

[...]

PR #1645, opened 2025-12-16
Add Helper to Create NumPy Files

julianmi:
This PR introduces `create_numpy_file` in the file_io.hpp header. It creates NumPy files with pre-allocated space and writes the appropriate header based on the specified shape and data type. This removes duplicated code in CAGRA ACE in `cagra_build.cuh`. It addresses one follow-up item of https://github.com/rapidsai/cuvs/issues/1486.

CC @tfeher


tfeher (approved):
Thanks Julian for simplifying the code. LGTM.

PR #1652, opened 2025-12-17
Fix libclang download for Rust, CUDA initialization for C tests

KyleFromNVIDIA:
Searching in `/opt/conda` for libclang was causing it to sometimes find libclang in the package directory instead of the environment. Search the environment only.

Also initialize CUDA architectures for C tests in case `$CUDAARCHS` is set to `RAPIDS`.

### DIFF
--- a/c/CMakeLists.txt
+++ b/c/CMakeLists.txt
@@ -201,6 +201,8 @@ endif()
 # ##################################################################################################
 # * build tests ----------------------------------------------------
 if(BUILD_TESTS)
+  include(rapids-cuda)
+  rapids_cuda_init_architectures(CUVS_C)
   enable_language(CUDA)
   find_package(CUDAToolkit REQUIRED)
 
--- a/.github/workflows/publish-rust.yaml
+++ b/.github/workflows/publish-rust.yaml
@@ -44,7 +44,7 @@ jobs:
           conda activate rust
           set -eu
 
-          LIBCLANG_PATH=$(dirname "$(find /opt/conda -name libclang.so | head -n 1)")
+          LIBCLANG_PATH=$(dirname "$(find "$CONDA_PREFIX" -name libclang.so | head -n 1)")
           export LIBCLANG_PATH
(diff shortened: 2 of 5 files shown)


KyleFromNVIDIA:
These changes were already tested at https://github.com/rapidsai/cuvs/actions/runs/20316882452/job/58364555189?pr=1405, so they definitely work.

c/CMakeLists.txt
```diff
@@ -201,6 +201,8 @@ endif()
 # ##################################################################################################
 # * build tests ----------------------------------------------------
 if(BUILD_TESTS)
+  include(rapids-cuda)
+  rapids_cuda_init_architectures(CUVS_C)
```
bdice, line 205: Does this need to be called before the `project()` call? I thought that was a requirement but I might be misinterpreting the docs. Seems like there's an "either ... _or_" missing. https://docs.rapids.ai/api/rapids-cmake/nightly/command/rapids_cuda_init_architectures/

> Used before enabling the CUDA language either via [`project()`](https://cmake.org/cmake/help/latest/command/project.html#command:project) to establish the CUDA architectures to be compiled for.

dantegd (approved):
cahnges lgtm

c/CMakeLists.txt
KyleFromNVIDIA: I think those docs should read:

> Used before enabling the CUDA language either via [`project()`](https://cmake.org/cmake/help/latest/command/project.html#command:project) or [`enable_language()`](https://cmake.org/cmake/help/latest/command/project.html#command:enable_language) to establish the CUDA architectures to be compiled for.

KyleFromNVIDIA: Here's a similar example that I worked off of:

https://github.com/rapidsai/rmm/blob/main/cpp/CMakeLists.txt#L138-L145

KyleFromNVIDIA: I've opened https://github.com/rapidsai/rapids-cmake/pull/956 to fix this.

PR #1655, opened 2025-12-18
Fix optional seed assignment in spectral embedding

bdice:
Fixes optional seed assignment in `spectral_embedding.cuh`.

The `spectral_embedding_config.seed` is `std::optional<uint64_t>` but `config.seed` is `uint64_t`. This fix checks if the optional has a value before assigning it.


aamijar:
Isn't `config.seed` also `std::optional<uint64_t>`?

cjnolet (approved):

bdice:
This is no longer needed. @aamijar changed the type in https://github.com/rapidsai/raft/commit/f4a331ecf1d2d20635fa86bf5c466e781edc2253.

PR #1656, opened 2025-12-18
[WIP] Add pytest Logging Output

julianmi:
Do not review / merge.

This PR adds capturing of the pytest output to narrow down a flaky segmentation fault we see in `test_doctests.py`. I cannot reproduce the issue locally.


julianmi:
We found that this was an issue in the multi-GPU Python doctest (mg/cagra) and not in the regular CAGRA build doctest. https://github.com/rapidsai/cuvs/pull/1671 disables the multi-GPU tests since they are not supported in the current CI. This is not needed anymore.

PR #1657, opened 2025-12-19
Deduplicate `calc_chunk_indices_kernel`

### LINKED ISSUE #1577: Reduce Binary Size by deduplicating `calc_chunk_indices_kernel` kernel
There are multiple instantiations (62) of the `cuvs::neighbors::ivf::detail::calc_chunk_indices_kernel` kernel.

jinsolp:
Closes https://github.com/rapidsai/cuvs/issues/1577

Reduces binary size by deduplicating `calc_chunk_indices_kernel`. This PR reduces instantiations from 62 -> 1 for each template (`BlockDim`=32, 64, ..., 1024)

### Binary Size Changes
CUDA 12.9: 1096.15MB -> 
CUDA 13: 432.98 MB->


divyegala:
@robertmaynard can you review this PR as well? I was under the impression that launching a kernel across TUs would not work in CUDA whole compilation mode but here it seems to be working. Aren't kernels supposed to have their symbols hidden too?

robertmaynard (changes requested):
This approach is invalid and goes against the guidance in https://developer.nvidia.com/blog/cuda-c-compiler-updates-impacting-elf-visibility-and-linkage/

This currently only work in cuvs as we have failed to remove: https://github.com/rapidsai/cuvs/blob/main/cpp/cmake/modules/ConfigureCUDA.cmake#L38

jinsolp:
Ohh I see okay 🥲

cpp/src/neighbors/ivf_common.cu
divyegala, line 1: The kernel needs to be launched in the same TU as which it is defined. We *can* (but should ideally avoid) pass the pointer around to other TUs but they shouldn't be attempting to launch the kernel.

divyegala (approved):

jinsolp:
Hi @robertmaynard , can you check this PR? I've added the changes to follow the guidelines

robertmaynard (approved):

PR #1660, opened 2025-12-19
Only allow undefined (runtime error) kernel behavior in ivf_pq_search

robertmaynard:
Context: https://developer.nvidia.com/blog/cuda-c-compiler-updates-impacting-elf-visibility-and-linkage/

The current ivf_pq_search kernels rely on abusing behavior that causes runtime errors. Instead of allowing that behavior across all of CUVS we now restrict to just these kernels


robertmaynard:
Closing as this PR has been replaced by: https://github.com/rapidsai/cuvs/pull/1663

PR #1668, opened 2026-01-05
Remove faiss patch

aamijar:
Since this has been merged https://github.com/facebookresearch/faiss/pull/4731, we no longer need the patch introduced in https://github.com/rapidsai/cuvs/pull/1654


aamijar:
Looks like we are building faiss 1.13.0 (https://github.com/rapidsai/cuvs/actions/runs/20724359535/job/59496384768?pr=1668#step:11:1126) so its not picking up the latest faiss changes and the CI build is failing.

Let's build faiss from `main` since the changes haven't been tagged in a new release, probably will be available in 1.13.3

PR #1670, opened 2026-01-05
Update CUVS C binding ABI version.

robertmaynard:
Since we don't have the tooling online to validate if ABI changes have occured we will update the SOVERSION to be safe.


cjnolet:
Will this mean that code built against cuVS 25.12 will no longer work against 26.02? We've committed to Elastic that we would begin deprecation of breaking ABI changes starting in 26.02, rather than breaking it unexpectedly.

robertmaynard:
> Will this mean that code built against cuVS 25.12 will no longer work against 26.02? 

Yes any project that linked explicitly against a pre-built standalone version of libcuvs_c.so will now fail to run with 26.02 ( with this change ).

Currently we don't have the infrastructure in place to validate that no ABI change has occured between 25.12 and 26.02. So the safest way is to assume an ABI change has occured.

We could validate that no ABI change has occured during burndown, but after that point we would have to enforce stricter code reviews / manual validation for any PR that modifies the C API.

robertmaynard:
Closing as https://github.com/rapidsai/cuvs/pull/1708 replaces this

Issue #1647, opened 2025-12-16
[BUG] Python docstring code example for CAGRA build with ACE fails in CI


tfeher:
https://github.com/rapidsai/cuvs/pull/1641 removes the python example for CAGRA ACE building from our docstring, because that lead to CI failures.

We were not able to reproduce such errors locally. 

TODO: confirm whether this is an error in the way our docstrings are tested.

julianmi:
We found that this was an issue in the multi-GPU Python doctest (`mg/cagra`) and not in the regular CAGRA build doctest. #1671 disables the multi-GPU tests since they are not supported in the current CI. @tfeher I think this can be closed.

tfeher:
I agree that this can be closed.

PR #1671, opened 2026-01-06
Skip Python Multi-GPU Doctests

julianmi:
This test is flaky and difficult to reproduce. Disable until fully triaged. See https://github.com/rapidsai/cuvs/issues/1647.


julianmi:
The flaky tests results showed up around when we added the ACE build method. It was assumed that this was the issues. Enabling debugging in the Python doctest runs (#1656) showed the multi-GPU tests to fail: https://github.com/rapidsai/cuvs/actions/runs/20742345304/job/59552791803?pr=1656. I suspect that these should be skipped similarly to the regular MG Python tests.

bdice (approved):

divyegala (approved):

PR #1688, opened 2026-01-08
Fix Faiss Warning in CI

tarang-jain:
Not sure if we can get rid of the RESCAN, but it seems that `faiss_gpu_objs` is an OBJECT library.
Link to warning related to this: https://github.com/rapidsai/cuvs/actions/runs/20800604992/job/59744590582?pr=1664


tarang-jain:
Looks like this does not work, and requires the RESCAN.

tarang-jain:
This does not work and we need the RESCAN. Closing

PR #1691, opened 2026-01-09
[FEA] Out of Core MiniBatchKmeans Algorithm

tarang-jain:
Merge after #1886 

This PR adds support for out of core (dataset on host) kmeans clustering. The idea is simple:
1. Batched accumulation of centroid updates: Data is processed in batches and batch-wise means and cluster counts are accumulated until all the batches i.e., the full dataset pass has completed.
2. `MiniBatchKmeans`: Randomly sampled batches on host + online learning rate, which changes with each batch.

[01/09/2026]
1. Mathematically equivalent to current kmeans. This PR just brings a batch-size parameter to load and compute cluster assignments and (weighted) centroid adjustments on batches of the dataset. The final centroid 'updates' i.e. a single kmeans iteration only completes when all these accumulated sums are averaged once the whole dataset pass has completed. Distinction from miniBatchKmeans: The centroid updates are done for each batch in miniBatchKmeans (faster to converge).
2. Binary size:
While the header is different for the batched approach, I've put the batched fit functions into the same TU as kmeans_fit. Common functions such as `minClusterAndDistance` with the same template params SHOULD not recompile.

[02/17/2026]
This PR also adds a GPU version of sklearn's [MiniBatchKMeans](https://scikit-learn.org/stable/modules/generated/sklearn.cluster.MiniBatchKMeans.html) so centroids are updated at every batch (potentially faster convergence, but incurs the cost of preparing randomly sampled batches on CPU).


c/include/cuvs/cluster/kmeans.h
```diff
@@ -212,6 +242,39 @@ cuvsError_t cuvsKMeansClusterCost(cuvsResources_t res,
...
+ *
+ *   This function processes data from HOST memory in batches, streaming
+ *   to the GPU. Useful when the dataset is too large to fit in GPU memory.
+ *
+ * @param[in]     res           opaque C handle
+ * @param[in]     params        Parameters for KMeans model.
+ * @param[in]     X             Training instances on HOST memory. The data must
+ *                              be in row-major format.
+ *                              [dim = n_samples x n_features]
+ * @param[in]     batch_size    Number of samples to process per batch.
```
tarang-jain, line 272: for the public API, the batch_size and centroid update mode can be wrapped into another struct called `batch_params`, which in fact becomes a part of `params`, so all we pass to this API is an updated `params` struct

python/cuvs/cuvs/tests/test_kmeans.py
```diff
@@ -69,3 +77,98 @@ def test_cluster_cost(n_rows, n_cols, n_clusters, dtype):
...
+        update_mode="mini_batch",
+    )
+    centroids_cuvs, _, _ = fit_batched(
+        params,
+        X_host,
+        batch_size=256,
+        centroids=device_ndarray(initial_centroids_host.copy()),
+    )
+    centroids_cuvs = centroids_cuvs.copy_to_host()
+    print(centroids_cuvs)
```
dantegd: Remove debug print

c/src/cluster/kmeans.cpp
```diff
@@ -17,16 +17,20 @@ namespace {
...
-  kmeans_params.n_clusters = params.n_clusters;
-  kmeans_params.max_iter   = params.max_iter;
-  kmeans_params.tol        = params.tol;
+  auto kmeans_params                = cuvs::cluster::kmeans::params();
+  kmeans_params.metric              = static_cast<cuvs::distance::DistanceType>(params.metric);
+  kmeans_params.init = static_cast<cuvs::cluster::kmeans::params::InitMethod>(params.init);
+  kmeans_params.n_clusters          = params.n_clusters;
+  kmeans_params.max_iter            = params.max_iter;
+  kmeans_params.tol                 = params.tol;
+  kmeans_params.n_init              = params.n_init;
```
dantegd, line 26: `n_init` is not currently used by `fit_batched`, right? i.e. it always does a single iteration, would be good to document that limitation of out of core KMeans.

c/src/cluster/kmeans.cpp
```diff
@@ -17,16 +17,20 @@ namespace {
...
   kmeans_params.oversampling_factor = params.oversampling_factor;
   kmeans_params.batch_samples       = params.batch_samples;
   kmeans_params.batch_centroids     = params.batch_centroids;
   kmeans_params.inertia_check       = params.inertia_check;
+  kmeans_params.final_inertia_check = params.final_inertia_check;
```
dantegd: The `final_inertia_check` field is in the C struct but not in the  Cython .pxd declaration. Cython accesses the C struct via extern, so existing field offsets are still correct. But Python users have no way to set `final_inertia_check = true` to request a final inertia computation, I think it'd be good to add `final_inertia_check` to the .pxd and expose it. 

If we don't want that, at least we should document that batched inertia is skipped from Python.

python/cuvs/cuvs/tests/test_kmeans.py
```diff
@@ -69,3 +77,98 @@ def test_cluster_cost(n_rows, n_cols, n_clusters, dtype):
...
+    centroids_batched, _, _ = fit_batched(
+        params,
+        X_host,
+        batch_size=batch_size,
+        centroids=device_ndarray(initial_centroids_host.copy()),
+    )
+    centroids_batched = centroids_batched.copy_to_host()
+
+    assert np.allclose(
+        centroids_regular, centroids_batched, rtol=0.05, atol=0.05
```
dantegd: Question: the margins (0.0.5) here should come mainly from some randomness and floating point operation discrepancies, right? I wonder if the margins could be tighter in fp64 inputs, and in general wanted to ask if you checked that 0.05 is the correct tightness

python/cuvs/cuvs/tests/test_kmeans.py
```diff
@@ -69,3 +77,98 @@ def test_cluster_cost(n_rows, n_cols, n_clusters, dtype):
...
+        params,
+        X_host,
+        batch_size=256,
+        centroids=device_ndarray(initial_centroids_host.copy()),
+    )
+    centroids_cuvs = centroids_cuvs.copy_to_host()
+    print(centroids_cuvs)
+
+    assert np.allclose(
+        centroids_sklearn, centroids_cuvs, rtol=0.3, atol=0.3
```
dantegd: I agree with this margin in general, this should give us an indication that we're in the same ballpark so to speak. But the test could also compare inertia to see how close we are to sklearn's and be a form of documenting differences there as well.

cpp/src/cluster/detail/kmeans_batched.cuh
```diff
@@ -0,0 +1,695 @@
...
+        IdxT sample_idx = batch_indices.data_handle()[i];
+        std::memcpy(host_batch_buffer.data_handle() + i * n_features,
+                    X.data_handle() + sample_idx * n_features,
+                    n_features * sizeof(T));
+      }
+
+      raft::copy(batch_data.data_handle(),
+                 host_batch_buffer.data_handle(),
+                 current_batch_size * n_features,
+                 stream);
```
dantegd: We are copying each batch synchronously, I wonder if for large datasets  overlapping the next H2D transfer with the current batch's compute (double-buffering) would improve throughput. 

More of an optimization potential opportunity than a correctness comment really.

cpp/src/cluster/detail/kmeans_batched.cuh
```diff
@@ -0,0 +1,695 @@
...
+{
+  cudaStream_t stream = raft::resource::get_cuda_stream(handle);
+  auto n_samples      = X.extent(0);
+  auto n_features     = X.extent(1);
+  auto n_samples_out  = X_sample.extent(0);
+
+  std::mt19937 gen(seed);
+  std::vector<IdxT> indices(n_samples);
+  std::iota(indices.begin(), indices.end(), 0);
+  std::shuffle(indices.begin(), indices.end(), gen);
```
dantegd: Another potential optimization: we are creating an array of size `n_samples` which could be very large for datasets that warrant out-of-core (particularly for large GPUs) therefore this vector could be gigabytes of host memory, and then we shuffle the entire thing. Meanwhile, `n_samples_out` is siginificantly smaller ( even say with 100,000 clusters, it is `min(n_samples, max(3 * n_clusters, 10000))` which is 300,000, much smaller than the number of samples).

A simple optimization would be to generate`n_samples_out` random indices directly with rejection sampling (using an `unordered_set` to avoid duplicates for example). Since `n_samples_out` << `n_samples`, collisions should be negligible and it's `O(n_samples_out)` in both time and memory instead of `O(n_samples)`.

dantegd (changes requested):
Very solid addition and implementation, very glad we finally have batched/mini-batched KMeans in RAPIDS!

Had a few comments, but overall the PR is very good and all major things look fine to me.

cpp/src/cluster/detail/kmeans_batched.cuh
tarang-jain: Good catch! I'll add double buffering.

python/cuvs/cuvs/tests/test_kmeans.py
tarang-jain: I think it is floating point operation discrepancies. I did an experiment, wherein I pre-normalized the data with L1 norms, and I could reduce atol and rtol to as low as 0.001.

tarang-jain:
Visualization of Clusters On wiki_all_1M (1M * 768 fp32) 256 centroids + CPU UMAP for dimensionality reduction:

Regular KMeans:
image

Full batch (LLoyd's algo with batched accumulation of centroid updates):
batch_size = 100k
image

MiniBatchKmeans:
Minibatch size = 100k with early stopping criteria = stop when 10 consecutive batches produce no improvement in inertia.
image

tarang-jain:
Benchmarks on wiki_all_1M (1M * 768 fp32 dataset):
| Method                | n_clusters | batch_size | max_iter | Time (s) | Number of Times Centroids are Updated | Inertia       |
|------------------------|-------------|-------------|-----------|----------|--------|----------------|
| Regular fit            | 256         | —           | 100       | 1.401    | 101    | 2.748459e+06   |
| Batched FULL-BATCH     | 256         | 100,000     | 100       | 32.224   | 101    | 2.751807e+06   |
| Batched MINI-BATCH     | 256         | 100,000     | 100       | 7.671    | 147    | 2.844728e+06   |

tarang-jain:
Fit time with different batch sizes (wiki_all_1M 1M * 768):
image

cpp/src/cluster/detail/kmeans_batched.cuh
tarang-jain: Ok but double buffering would mean that we need space on the GPU to stream two batches of data at a time, right? I can update the batch loading to use our own batch_load_iterator, that has optional prefetch mode instead of iterating and creating batches on our own. Maybe we can make it an optional flag to prefetch?

python/cuvs/cuvs/tests/test_kmeans.py
tarang-jain: I have added inertia check

c/src/cluster/kmeans.cpp
tarang-jain: yeah I intentionally left it out. It might still make sense for minibatch mode. I'll recheck what sklearn does for this.

tarang-jain: I have added the n_init related stuff. Its pretty neat! For full_batch mode we can do the same thing that we were doing before, although it will be slow so it is the user's choice. For minibatch mode, we hold aside a validation set and validate our n_init initializations on the validation set with inertia (idea borrowed directly from sklearn).

tarang-jain:
Closing because this has been deprioritized at the moment.

PR #1692, opened 2026-01-10
Avoid installing CCCL headers in wheels

divyegala:
Uncompressed size of CCCL headers: 19M
No of headers per CCCL library:
```
cudax/include:
208
cub/cub:
172
libcudacxx/include/cuda:
950
thrust/thrust:
729
```

Before this PR:
* compressed size: 0.406G
* uncompressed size: 0.452G

After this PR:
* compressed size: 0.403G
* uncompressed size: 0.438G


python/libcuvs/pyproject.toml
```diff
@@ -69,6 +69,7 @@ sdist.reproducible = true
 wheel.install-dir = "libcuvs"
 wheel.packages = ["libcuvs"]
 wheel.py-api = "py3"
+install.components = ["cuvs_cpp_headers", "cuvs", "cuvs_shared", "c_api", "hnswlib"]
```
vyasr, line 72: Will we ever be able to stop installing hnswlib? Do we have to expose its headers publicly to users?

divyegala: I'm going to do this as a follow-on because I didn't want to overpopulate the diff, but we don't need this component at all.

cjnolet (approved):

PR #1693, opened 2026-01-11
Spectral Embedding with `all_neighbors`

aamijar:
Instead of using `cuvs::neighbors::brute_force::build` and `cuvs::neighbors::brute_force::search` we can consolidate to use the `cuvs::neighbors::all_neighbors::build` API.

### DIFF
--- a/cpp/src/preprocessing/spectral/detail/spectral_embedding.cuh
+++ b/cpp/src/preprocessing/spectral/detail/spectral_embedding.cuh
@@ -147,8 +147,8 @@ void create_connectivity_graph(
 
   auto stream = raft::resource::get_cuda_stream(handle);
 
-  auto d_indices   = raft::make_device_matrix<int64_t>(handle, n_samples, k_search);
-  auto d_distances = raft::make_device_matrix<float>(handle, n_samples, k_search);
+  auto d_indices   = raft::make_device_matrix<int64_t, int64_t>(handle, n_samples, k_search);
+  auto d_distances = raft::make_device_matrix<float, int64_t>(handle, n_samples, k_search);
 
   cuvs::neighbors::all_neighbors::all_neighbors_params all_neighbors_params{
     .graph_build_params =


cpp/src/preprocessing/spectral/detail/spectral_embedding.cuh
```diff
@@ -192,18 +192,17 @@ void create_connectivity_graph(
...
   auto d_distances = raft::make_device_matrix<float>(handle, n_samples, k_search);
 
-  auto index =
-    cuvs::neighbors::brute_force::build(handle, index_params, raft::make_const_mdspan(dataset));
-
-  cuvs::neighbors::brute_force::search(
-    handle, search_params, index, dataset, d_indices.view(), d_distances.view());
+  cuvs::neighbors::all_neighbors::all_neighbors_params all_neighbors_params{};
+  all_neighbors_params.graph_build_params =
+    cuvs::neighbors::all_neighbors::graph_build_params::brute_force_params{};
```
jinsolp: Question: is it correct that the metrics used is always `L2SqrtExpanded`? (from what was deleted above)

can you set the metrics to these to match whatever metric we should be using?
```
all_neighbors_params.metric = ...

bf_params = cuvs::neighbors::all_neighbors::graph_build_params::brute_force_params{};
bf_params.metric = ...
all_neighbors_params.graph_build_params = bf_params
```

aamijar: Yes, the metric I want to use is `L2SqrtExpanded`, same as the old code. I can explicitly set the metric yes, but it looks like the default is already `L2SqrtExpanded`?

aamijar: Addressed in https://github.com/rapidsai/cuvs/pull/1693/changes/c093ed933be24eec6a3fa84a8f0bda7f526222b4

cpp/src/preprocessing/spectral/detail/spectral_embedding.cuh
```diff
@@ -192,18 +192,19 @@ void create_connectivity_graph(
...
   auto d_indices   = raft::make_device_matrix<int64_t>(handle, n_samples, k_search);
   auto d_distances = raft::make_device_matrix<float>(handle, n_samples, k_search);
 
-  auto index =
-    cuvs::neighbors::brute_force::build(handle, index_params, raft::make_const_mdspan(dataset));
-
-  cuvs::neighbors::brute_force::search(
-    handle, search_params, index, dataset, d_indices.view(), d_distances.view());
+  cuvs::neighbors::all_neighbors::all_neighbors_params all_neighbors_params{
+    .graph_build_params =
```
tarang-jain, line 154: not well-versed with this way of initializing a struct, but apart from that LGTM.

tarang-jain (approved):

cpp/src/preprocessing/spectral/detail/spectral_embedding.cuh
aamijar: Yeah, I just found out about it too, its called designated initializers in cpp 20

jinsolp (approved):
LGTM!

cpp/src/preprocessing/spectral/detail/spectral_embedding.cuh
tarang-jain: Wait, but then why do you have metric twice? Is it because one is in the `graph_build_params` and one is in the `build_params`?

aamijar: Yes, one is for all_neighbors_params and one is for bf_params just like https://github.com/rapidsai/cuvs/pull/1693#discussion_r2684194858

viclafargue (approved):

PR #1704, opened 2026-01-14
Remove unused variable n_rows

maxwbuckley:
Removed unused variable n_rows in ivf_flat_serialize.cuh.


cjnolet (approved):

aamijar:
Hi @maxwbuckley thanks for the PR. The reason CI is failing is because we need to keep the first `raft::deserialize_scalar`. We have to deserialize the entire input stream in order to properly assign to the correct variable types, hence we can't skip any of them.

PR #1708, opened 2026-01-15
pre-built libcuvs_c.so now use the new ABI major/minor values

robertmaynard:
Based on the in-development cuVS ABI stability docs we now encode the C API's ABI stability in the SOVERSION

### DIFF
--- a/cpp/cmake/modules/determine_cuvs_abi_version.cmake
+++ b/cpp/cmake/modules/determine_cuvs_abi_version.cmake
@@ -79,7 +79,7 @@ function(determine_cuvs_abi_version cal_ver)
     math(EXPR first_year_count "(12-${abi_base_month})/2")
     math(EXPR extra_years "(${cal_ver_major} - ${abi_base_year} - 1) * 6")
     math(EXPR this_year_count "(${cal_ver_minor})/2")
-    math(EXPR computed_abi_minor "${first_year_count} + ${extra_years} + +${this_year_count}")
+    math(EXPR computed_abi_minor "${first_year_count} + ${extra_years} + ${this_year_count}")
   endif()
 
   set(${_CUVS_RAPIDS_MAJOR}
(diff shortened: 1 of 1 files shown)


cpp/cmake/modules/determine_cuvs_abi_version.cmake
```diff
@@ -0,0 +1,93 @@
...
+  # compute the abi version
+  if(cal_ver_major STREQUAL abi_base_year)
+    # If we are in the same year is is pretty easy to compute our abi break
+    math(EXPR computed_abi_minor "(${cal_ver_minor}-${abi_base_month})/2")
+  else()
+    #
+    math(EXPR first_year_count "(12-${abi_base_month})/2")
+    math(EXPR extra_years "(${cal_ver_major} - ${abi_base_year} - 1) * 6")
+    math(EXPR this_year_count "(${cal_ver_minor})/2")
+    math(EXPR computed_abi_minor "${first_year_count} + ${extra_years} + +${this_year_count}")
```
dantegd:
```suggestion
    math(EXPR computed_abi_minor "${first_year_count} + ${extra_years} + ${this_year_count}")
```

Unsure about this so more of a question, is the double `+ +` here on purpose? I think it's valid, but it doesn't do anything, right?

cpp/cmake/modules/determine_cuvs_abi_version.cmake
```diff
@@ -0,0 +1,93 @@
...
+  endif()
+
+  set(${_CUVS_RAPIDS_MAJOR}
+      ${computed_abi_major}
+      PARENT_SCOPE
+  )
+  set(${_CUVS_RAPIDS_MINOR}
+      ${compute_abi_minor}
```
dantegd:
```suggestion
      ${computed_abi_minor}
```

cpp/cmake/modules/determine_cuvs_abi_version.cmake
robertmaynard: That was a typo, removed

cpp/cmake/modules/determine_cuvs_abi_version.cmake
robertmaynard: corrected

PR #1710, opened 2026-01-16
wheel builds: react to changes in pip's handling of build constraints

mmccarty:
## Description
Contributes to https://github.com/rapidsai/build-planning/issues/242

Modifying `ci/build_wheel.sh` to

- pass`--build-constraint="${PIP_CONSTRAINT}"` unless build isolation is enabled.
- unset `PIP_CONSTRAINT` (set by rapids-init-pip)... it doesn't affect builds as of pip 25.3, and results in an error from `pip wheel` when set and `--build-constraint` is also passed

### DIFF
--- a/ci/build_wheel_cuvs.sh
+++ b/ci/build_wheel_cuvs.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-# SPDX-FileCopyrightText: Copyright (c) 2023-2025, NVIDIA CORPORATION.
+# SPDX-FileCopyrightText: Copyright (c) 2023-2026, NVIDIA CORPORATION.
 # SPDX-License-Identifier: Apache-2.0
 
 set -euo pipefail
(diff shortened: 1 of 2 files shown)


ci/build_wheel_cuvs.sh
```diff
@@ -13,8 +13,8 @@ RAPIDS_PY_CUDA_SUFFIX="$(rapids-wheel-ctk-name-gen "${RAPIDS_CUDA_VERSION}")"
 # Downloads libcuvs wheels from this current build,
 # then ensures 'cuvs' wheel builds always use the 'libcuvs' just built in the same CI run.
 #
-# Using env variable PIP_CONSTRAINT is necessary to ensure the constraints
-# Using env variable PIP_CONSTRAINT (initialized by 'rapids-init-pip') is necessary to ensure the constraints
+# env variable 'PIP_CONSTRAINT' is set up by rapids-init-pip. It constrains all subsequent
```
jameslamb, line 16: Thanks for fixing these duplicate lines, probably artifacts from some bad merge conflict resolution in the past.

PR #1713, opened 2026-01-20
Fix multi-GPU All Neighbors memory coherence issue on older platforms

viclafargue:
Closes https://github.com/rapidsai/cuml/issues/7210


cpp/src/neighbors/all_neighbors/all_neighbors_batched.cuh
```diff
@@ -528,6 +531,39 @@ void batch_build(
...
+    // Ampere (SM 8.0+) and newer architectures handle this correctly.
+    int num_ranks         = raft::resource::get_num_ranks(handle);
+    bool needs_workaround = false;
+    for (int rank = 0; rank < num_ranks; rank++) {
+      raft::resource::set_current_device_to_rank(handle, rank);
+      int device_id;
+      RAFT_CUDA_TRY(cudaGetDevice(&device_id));
+      int major = 0;
+      RAFT_CUDA_TRY(cudaDeviceGetAttribute(&major, cudaDevAttrComputeCapabilityMajor, device_id));
+      if (major < 8) { needs_workaround = true; }  // Turing is 7.x, Ampere is 8.x
```
jinsolp:
```suggestion
      if (major < 8) { needs_workaround = true; break;}  // Turing is 7.x, Ampere is 8.x
```

cpp/src/neighbors/all_neighbors/all_neighbors_batched.cuh
```diff
@@ -528,6 +531,39 @@ void batch_build(
 
   reset_global_matrices(handle, params.metric, global_neighbors.view(), global_distances.view());
 
+  // For multi-GPU: sync the stream to ensure fill completes before other GPUs access
+  // the managed memory.
+  if (raft::resource::is_multi_gpu(handle)) {
+    raft::resource::sync_stream(handle);
+
```
jinsolp: I think we need to grab the current device here, and then reset it after the for loop below. (since we call `set_current_device_to_rank` in the for loop)

cpp/src/neighbors/all_neighbors/all_neighbors_batched.cuh
```diff
@@ -528,6 +531,39 @@ void batch_build(
...
+    int num_ranks         = raft::resource::get_num_ranks(handle);
+    bool needs_workaround = false;
+    for (int rank = 0; rank < num_ranks; rank++) {
+      raft::resource::set_current_device_to_rank(handle, rank);
+      int device_id;
+      RAFT_CUDA_TRY(cudaGetDevice(&device_id));
+      int major = 0;
+      RAFT_CUDA_TRY(cudaDeviceGetAttribute(&major, cudaDevAttrComputeCapabilityMajor, device_id));
+      if (major < 8) { needs_workaround = true; }  // Turing is 7.x, Ampere is 8.x
+    }
```
dantegd, line 551: Small thing, but after iterating through ranks to check capabilities, the current CUDA device is left at the last rank's device. While `multi_gpu_batch_build` will set the device correctly in its parallel region, this is still problematic because of code smell, it violates principle of least surprise.

I'd recommend we save and restore the device context.

cpp/src/neighbors/all_neighbors/all_neighbors_batched.cuh
```diff
@@ -528,6 +531,39 @@ void batch_build(
...
+    if (needs_workaround) {
+      cudaMemLocation cpu_location = {cudaMemLocationTypeHost, 0};
+      RAFT_CUDA_TRY(cudaMemAdvise(global_neighbors.data_handle(),
+                                  num_rows * k * sizeof(IdxT),
+                                  cudaMemAdviseSetPreferredLocation,
+                                  cpu_location));
+      RAFT_CUDA_TRY(cudaMemAdvise(global_distances.data_handle(),
+                                  num_rows * k * sizeof(T),
+                                  cudaMemAdviseSetPreferredLocation,
+                                  cpu_location));
```
dantegd, line 564: Not necessarily a blocker or a big thing, but the current order of operations in the code goes through:

1. Create managed memory
2. Fill with initial values (`reset_global_matrices`)
3. Sync
4. Set preferred location to CPU

Setting cudaMemAdviseSetPreferredLocation to host after filling the data may trigger page migration from device to host. It might be better to do the location before filling, something like:

```C++
auto global_neighbors = raft::make_managed_matrix<IdxT, IdxT>(handle, num_rows, k);
auto global_distances = raft::make_managed_matrix<T, IdxT>(handle, num_rows, k);

if (raft::resource::is_multi_gpu(handle)) {
    // Check capabilities and set preferred location first
}

reset_global_matrices(handle, params.metric, global_neighbors.view(), global_distances.view());
```

cpp/src/neighbors/all_neighbors/all_neighbors_batched.cuh
```diff
@@ -528,6 +531,39 @@ void batch_build(
...
+      raft::resource::set_current_device_to_rank(handle, rank);
+      int device_id;
+      RAFT_CUDA_TRY(cudaGetDevice(&device_id));
+      int major = 0;
+      RAFT_CUDA_TRY(cudaDeviceGetAttribute(&major, cudaDevAttrComputeCapabilityMajor, device_id));
+      if (major < 8) { needs_workaround = true; }  // Turing is 7.x, Ampere is 8.x
+    }
+
+    if (needs_workaround) {
+      cudaMemLocation cpu_location = {cudaMemLocationTypeHost, 0};
```
dantegd, line 556:
```suggestion
      RAFT_LOG_DEBUG("Applying managed memory workaround for pre-Ampere GPU architecture");
      cudaMemLocation cpu_location = {cudaMemLocationTypeHost, 0};
```

Logging when this happens might be useful for debugging later on.

dantegd (changes requested):
Change looks good, just some minor suggestions alongside @jinsolp review

PR #1716, opened 2026-01-21
Temporarily disable memory pool usage in MG tests

viclafargue:


viclafargue:
The memory pool usage in MG tests is not involved in CI issues. Closing as it is not relevant.

PR #1718, opened 2026-01-21
Better handling of batching of search in MG replicated mode

viclafargue:
Answers #1720

In multi-GPU replicated mode, the search query is divided in batches. These batches are ran in parallel with OpenMP. In some cases, there may be more batches than available GPUs causing a thread safety issue (at least for CAGRA indices). This change solves the issue. Each rank gets its own thread, that thread handles all batches for that rank sequentially. This prevents concurrent access to the same GPU from multiple threads.

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -43,7 +43,6 @@ jobs:
       node_type: cpu16
       script: ci/build_cpp.sh
       sha: ${{ inputs.sha }}
-      sccache-dist-token-secret-name: GIST_REPO_READ_ORG_GITHUB_TOKEN
 
   rocky8-clib-standalone-build:
     secrets: inherit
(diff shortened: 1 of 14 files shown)


tfeher:
I would expect CAGRA to be thread safe. We expect that we can search using multiple threads to achieve large throughput for concurrent small batch size queries. We should fix this in CAGRA.

cjnolet:
@viclafargue cam you also link the issue in a comment in the code before this is merged?

bdice:
@divyegala @achirkin This is failing to merge because it is targeting `release/26.02`. (Is it a hotfix?) If that's correct, an admin-merge is needed from the ops team. Otherwise it should be retargeted to `main`.

viclafargue:
> @divyegala @achirkin This is failing to merge because it is targeting release/26.02. (Is it a hotfix?) If that's correct, an admin-merge is needed from the ops team. Otherwise it should be retargeted to main.

This is a mistake. I re-targeted to the `main` branch.

PR #1719, opened 2026-01-21
Auto select CAGRA build algorithm for hnsw::build

tfeher:
Configuring HNSW graph build using CAGRA is complicated, because CAGRA offers multiple build algorithms. This PR implements an automatic algorithm selection. The goal is to have a simplified API, where the user needs to set only two parameters that control graph size and quality (`M` and `ef_construction` respectively). This shall be familiar for HNSW users, and allows easier adaption of cuvs accelerated HNSW graph building.

```c++
  hnsw::index_params params;
  params.M               = 24;
  params.ef_construction = 200;
  params.hierarchy       = cuvs::neighbors::hnsw::HnswHierarchy::GPU;

  auto hnsw_index = hnsw::build(res, params, dataset_host_view);
  cuvs::neighbors::hnsw::serialize(res, "hnsw_index.bin", *hnsw_index);
```

If we have enough memory (host and GPU) to do both the KNN graph building and optimization in memory, then we choose in memory build, and let `cagra::index_params::from_hnsw_params` derive the additional configuration parameters.

If the build would require more memory then available, then we choose ACE method and let the number of partitions derived using #1603.

For host we query the os for available memory, for GPU it is assumed that the whole device memory is available.


cpp/src/neighbors/detail/cagra/cagra_helpers.cpp
```diff
@@ -0,0 +1,132 @@
...
+  auto [host_workspace_size, gpu_workspace_size] =
+    cuvs::neighbors::cagra::helpers::optimize_workspace_size(
+      n_rows, graph_degree, intermediate_graph_degree, sizeof(uint32_t));
+  float host_workspace_gb = host_workspace_size / 1e9;
+  float gpu_workspace_gb  = gpu_workspace_size / 1e9;
+
+  size_t total_host =
+    graph_host_mem + host_workspace_size + 2e9;  // added 2 GB extra workspace (IVF-PQ search)
+  size_t total_dev =
+    std::max(dataset_gpu_mem, gpu_workspace_size) + 1e9;  // addet 1 GB extra workspace size
```
tfeher: This is still optimistic, we need to update and test before the PR can be merged.

examples/cpp/src/hnsw_openai_example.cu
```diff
@@ -0,0 +1,102 @@
...
+  auto dataset_host_view = raft::make_host_matrix_view<const float, int64_t, raft::row_major>(
+    reinterpret_cast<float*>(dataset_ptr + header_size), n_rows, shape[1]);
+
+  int64_t topk = 12;
+
+  // HNSW index parameters
+  hnsw::index_params params;
+  params.M               = 24;
+  params.ef_construction = 200;
+  params.hierarchy       = cuvs::neighbors::hnsw::HnswHierarchy::GPU;
```
tfeher: We should make this default #1617

examples/cpp/src/hnsw_openai_example.cu
```diff
@@ -0,0 +1,102 @@
...
+  }
+  uint32_t n_rows = shape[0];
+  // n_rows = 1000000;
+  auto dataset_host_view = raft::make_host_matrix_view<const float, int64_t, raft::row_major>(
+    reinterpret_cast<float*>(dataset_ptr + header_size), n_rows, shape[1]);
+
+  int64_t topk = 12;
+
+  // HNSW index parameters
+  hnsw::index_params params;
```
tfeher, line 64: We need to figure out how to handle `ace.build_dir` in this setup: the user does not set ace_params. The algorithm is automatically selected. But if we happen to choose ace, then we need to know which disk space to use. Do you have suggestions @julianmi?

julianmi: I think this is a somewhat challenging problem. In general, I agree with @benfred's [comment](https://github.com/rapidsai/cuvs/pull/1603#discussion_r2709472119) that hardcoding it is not a good approach.

I think the two most important properties are available disk space and speed. How about a layered approach:

1. Environment variable. E.g., `CUVS_ACE_BUILD_DIR`.
2. We could read `/sys/block` on Linux to query for fast disks (NVMe > SSD > HDD) and check for free size.
3. Fall back to `/tmp` or a generated temporary path like @benfred suggested.

cpp/bench/ann/src/cuvs/cuvs_cagra_hnswlib_wrapper.h
```diff
@@ -85,38 +85,9 @@ void cuvs_cagra_hnswlib<T, IdxT>::build(const T* dataset, size_t nrow)
...
-  auto& cagra_index = *cagra_wrapper.get_index();
-
-  // pass the dataset directly to HNSW if it's on the host
-  std::optional<raft::host_matrix_view<const T, int64_t>> opt_dataset_view = std::nullopt;
-  if (dataset_is_on_host) {
-    opt_dataset_view.emplace(
-      raft::make_host_matrix_view<const T, int64_t>(dataset, nrow, this->dim_));
-  }
-
+  auto dataset_view = raft::make_host_matrix_view<const T, int64_t>(dataset, nrow, this->dim_);
```
mfoerste4, line 88: Is the data expected to always reside in host memory?

cpp/include/cuvs/neighbors/cagra.hpp
```diff
@@ -3180,6 +3180,36 @@ void build_knn_graph(raft::resources const& res,
                      raft::host_matrix_view<uint32_t, int64_t, raft::row_major> knn_graph,
                      cuvs::neighbors::cagra::graph_build_params::ivf_pq_params build_params);
 
+namespace helpers {
+/** Calculates the workspace for graph optimization
+ *
+ * @param[in] n_rows number of rows in the dataset (or number of points in the grapt)
```
mfoerste4:
```suggestion
 * @param[in] n_rows number of rows in the dataset (or number of points in the graph)
```

mfoerste4: Also `mst_optimize `is not documented

examples/cpp/src/hnsw_openai_example.cu
```diff
@@ -0,0 +1,102 @@
...
+  // rmm::mr::set_current_device_resource(&pool_mr);
+
+  // Alternatively, one could define a pool allocator for temporary arrays (used within RAFT
+  // algorithms). In that case only the internal arrays would use the pool, any other allocation
+  // uses the default RMM memory resource. Here is how to change the workspace memory resource to
+  // a pool with 2 GiB upper limit.
+  raft::resource::set_workspace_to_pool_resource(res, 2 * 1024 * 1024 * 1024ull);
+
+  // ACE build and search example.
+  cagra_build_search_ace(res);
```
mfoerste4, line 95: maybe we want to rename this to something generic now that the selection is hidden from the user.

examples/cpp/src/hnsw_openai_example.cu
mfoerste4: I agree with a layered approach using only (1) and (3) which are both agreed on by the user to write to -- I don't think we can just write to a drive without user consent.

cpp/src/neighbors/detail/hnsw.hpp
```diff
@@ -1280,33 +1298,61 @@ std::unique_ptr<index<T>> build(raft::resources const& res,
...
+
+    // Configure ACE parameters for CAGRA
+    cuvs::neighbors::cagra::graph_build_params::ace_params cagra_ace_params;
+    cagra_ace_params.npartitions        = ace_params.npartitions;
+    cagra_ace_params.ef_construction    = params.ef_construction;
+    cagra_ace_params.build_dir          = ace_params.build_dir;
+    cagra_ace_params.use_disk           = ace_params.use_disk;
+    cagra_ace_params.max_host_memory_gb = ace_params.max_host_memory_gb;
+    cagra_ace_params.max_gpu_memory_gb  = ace_params.max_gpu_memory_gb;
+    cagra_params.graph_build_params     = cagra_ace_params;
```
mfoerste4, line 1595: Are you planning to add a heuristic for _npartitions_ depending on dimensions here as well?

mfoerste4 (commented):
I did not go over all memory estimates in detail but suggest to align predictions with real data. 

Is autotuning of ACE params part of a different PR? Besides the open question on the file location we might want to at least set the number of partitions dynamically.

examples/cpp/src/hnsw_openai_example.cu
julianmi: I agree that (2) is problematic. But (3) has similar problems, right? `/tmp` could be very small or on a slow disk.

Another approach I see would be to fail with a helpful message that a disk will be used given the graph size and the user should provide a suitable directory.

In general, I am not sure if environment variables are a good approach since the project does not use them a lot. @tfeher, @cjnolet What do you think?

cpp/src/neighbors/detail/hnsw.hpp
julianmi: We have added heuristics in #1603.

cpp/bench/ann/src/cuvs/cuvs_cagra_hnswlib_wrapper.h
julianmi: ACE only supports host memory right now. The main reasons is that we expect the data size to be large and memory-mapped. Further, we do the partitioning and reordering on the host since there is no benefit of moving it to the GPU only to write it to disk afterwards.

Anyways, I think we can support device datasets easily since these should not end up using ACE with this heuristic. @tfeher What do you think?

cpp/src/neighbors/ivf_pq_index.cu
```diff
@@ -788,6 +789,44 @@ raft::device_matrix_view<const half, uint32_t, raft::row_major> index<IdxT>::cen
...
+size_t compressed_dataset_size(raft::resources const& res,
+                               raft::matrix_extent<int64_t> dataset,
+                               cuvs::neighbors::ivf_pq::index_params params)
+{
+  auto idx = cuvs::neighbors::ivf_pq::index<int64_t>(res, params, (uint32_t)dataset.extent(1));
+  size_t pq_book_size                         = 1 << params.pq_bits;
+  constexpr static uint32_t kIndexGroupSize   = 32;
+  constexpr static uint32_t kIndexGroupVecLen = 16;
+
+  std::cout << "pq_dim " << params.pq_dim << ", pq_bits " << params.pq_bits << ", n_lists"
```
julianmi: Is there a specific reason not to use `RAFT_LOG_INFO` here and in the following lines?

cpp/src/neighbors/detail/hnsw.hpp
```diff
@@ -1265,6 +1266,23 @@ void deserialize(raft::resources const& res,
   }
 }
 
+inline std::pair<size_t, size_t> get_available_memory(
```
julianmi: Should this helper be placed in `cuvs::util::` like `get_free_host_memory`?

cpp/src/neighbors/detail/cagra/cagra_build.cuh
```diff
@@ -1640,7 +1597,7 @@ void build_knn_graph(
     return std::string(model_name);
   }();
 
-  RAFT_LOG_DEBUG("# Building IVF-PQ index %s", model_name.c_str());
+  RAFT_LOG_INFO("# Building IVF-PQ index %s", model_name.c_str());
```
julianmi: Was this and the following logging changes intentionally? Logging every 10 seconds might write a lot of output on a large run.

cpp/src/neighbors/detail/cagra/cagra_helpers.cpp
```diff
@@ -0,0 +1,132 @@
...
+
+  std::cout << "IVF-PQ build memory requirements\ndataset_gpu " << dataset_gpu_mem / 1e9 << " GB"
```
julianmi: Similar comment about using `RAFT_LOG_INFO`.

cpp/src/neighbors/detail/cagra/cagra_helpers.cpp
julianmi:
```suggestion
    std::max(dataset_gpu_mem, gpu_workspace_size) + 1e9;  // added 1 GB extra workspace size
```

cpp/include/cuvs/neighbors/ivf_pq.hpp
```diff
@@ -3124,6 +3124,18 @@ void make_rotation_matrix(
   raft::resources const& res,
   raft::device_matrix_view<float, uint32_t, raft::row_major> rotation_matrix,
   bool force_random_rotation);
+
+/** Calculate the size of the compressed dataset.
+ *
+ * @param[in] res raft resource
+ * @param[in] dataset shape of the dataset
+ * @param[in] param ivf-pq compression pramas
```
julianmi:
```suggestion
 * @param[in] param ivf-pq compression params
```

julianmi (commented):
I did not get a chance to fully review the memory heuristics yet. I wonder how we can test it though. Should `max_host_memory_gb` and  `max_gpu_memory_gb` be optional HNSW parameters that we could use to test that the expected algorithm is used based on memory limits set?

cpp/src/neighbors/detail/cagra/cagra_helpers.cpp
```diff
@@ -0,0 +1,132 @@
...
+  auto [host_workspace_size, gpu_workspace_size] =
+    cuvs::neighbors::cagra::helpers::optimize_workspace_size(
+      n_rows, graph_degree, intermediate_graph_degree, sizeof(uint32_t));
+  float host_workspace_gb = host_workspace_size / 1e9;
+  float gpu_workspace_gb  = gpu_workspace_size / 1e9;
+
+  size_t total_host =
+    graph_host_mem + host_workspace_size + 2e9;  // added 2 GB extra workspace (IVF-PQ search)
+  size_t total_dev =
+    std::max(dataset_gpu_mem, gpu_workspace_size) + 1e9;  // added 1 GB extra workspace size
```
achirkin: For high-dimensional data and large enough trainset ratio, kmeans training takes a non-negligible amount of device memory:
```suggestion
  // The kmeans trainset is a large temporary float buffer allocated during IVF-PQ training.
  // It is freed before the extend phase, so peak GPU = max(training_peak, extend_peak).
  size_t kmeans_trainset_ratio = std::max<size_t>(
    1,
    n_rows / std::max<size_t>(params.build_params.kmeans_trainset_fraction * n_rows,
                              params.build_params.n_lists));
  size_t kmeans_n_rows  = n_rows / kmeans_trainset_ratio;
  size_t kmeans_gpu_mem = kmeans_n_rows * dataset.extent(1) * sizeof(float);

  size_t total_host =
    graph_host_mem + host_workspace_size + 2e9;  // added 2 GB extra workspace (IVF-PQ search)
  size_t total_dev = std::max({kmeans_gpu_mem, dataset_gpu_mem, gpu_workspace_size}) + 1e9;
```

mfoerste4:
@julianmi , please have another look at the PR. Some remarks:
* I added 2GB static memory consumption for both GPU & Host. This is probably too conservative given that we already reduce the available usage to 80% of the actual value. Especially on small host memory machines this increases the number of partitions drastically. We might want to get rid of one of the limits. On the GPU I would prefer the constant 2GB (e.g. for workspace memory), but on the host side the percentage seems more natural. What do you think?
* the from_cagra hnsw index creation now also has the disk serialization fallback in case the cagra index is in memory
* main difference in ace-partition estimates are the 2GB workspace, removal of sub_graph on device, optimize taking the same vector estimate as data&graph and split into static/dynamic memory requirement.

examples/cpp/CMakeLists.txt
```diff
@@ -32,9 +34,11 @@ include(../cmake/thirdparty/get_cuvs.cmake)
 add_executable(BRUTE_FORCE_EXAMPLE src/brute_force_bitmap.cu)
 add_executable(CAGRA_EXAMPLE src/cagra_example.cu)
 add_executable(CAGRA_HNSW_ACE_EXAMPLE src/cagra_hnsw_ace_example.cu)
+add_executable(CAGRA_HNSW_ACE_BUILD_EXAMPLE src/hnsw_cagra_ace_build.cu)
```
huuanhhuyn: nit: the binary name and the .cu file name doesn't match. I would suggest to rename the .cu file instead of the binary file

cpp/src/neighbors/detail/cagra/cagra_build.cuh
```diff
@@ -1011,47 +978,58 @@ void ace_validate_disk_mode_partitions(size_t& n_partitions,
       to_gib(mem.available_host_memory),
       to_gib(mem.sub_dataset_size),
       to_gib(mem.sub_graph_size),
-      to_gib(host_workspace_size));
+      to_gib(host_workspace_size_total));
```
huuanhhuyn, line 1013: I got this message ` ACE: Host memory insufficient for disk mode. Required: 1.88 GiB, available: 2.33 GiB. Per-partition breakdown: dataset 0.00 GiB, graph 0.01 GiB, workspace 0.00 GiB`, which is counter-intuitive because we have 2.33GB > 1.88GB, which should mean sufficient memory.

I would suggest to add `usable_cpu_memory_fraction` (80% currently) into the message so that it is more clear why it failed.

python/cuvs/cuvs/tests/test_hnsw_ace.py
```diff
@@ -233,14 +233,15 @@ def test_hnsw_ace_tiny_memory_limit_triggers_disk_mode():
     dataset = generate_data((n_rows, n_cols), dtype)
 
     with tempfile.TemporaryDirectory() as temp_dir:
-        # Set ACE parameters with tiny memory limits (0.001 GiB = ~1 MB)
+        # Set ACE parameters with memory limits slightly above the minimum required
+        # Input GiB values * 80% usage factor --> requires 2GB static memory overhead
```
huuanhhuyn: nit: The minimum required is more precisely 2.329 GB instead of 2GB. I would change it to help explain the magic number 2.329 below and in several other places.

cpp/src/neighbors/detail/cagra/cagra_build.cuh
huuanhhuyn: I believe the available host memory when the parameter `max_host_memory_gb` is given should instead be the min between the param value and the actual `cuvs::util::get_free_host_memory()`). We perhaps didn't catch this because there wasn't any test case for this afaik

cpp/src/neighbors/detail/cagra/cagra_helpers.cpp
```diff
@@ -0,0 +1,151 @@
...
+  // It is freed before the extend phase, so peak GPU = max(training_peak, extend_peak).
+  size_t kmeans_trainset_ratio = std::max<size_t>(
+    1,
+    n_rows / std::max<size_t>(params.build_params.kmeans_trainset_fraction * n_rows,
+                              params.build_params.n_lists));
+  size_t kmeans_n_rows  = n_rows / kmeans_trainset_ratio;
+  size_t kmeans_gpu_mem = kmeans_n_rows * dataset.extent(1) * sizeof(float);
+
+  size_t total_host =
+    graph_host_mem + host_workspace_size + 2e9;  // added 2 GB extra workspace (IVF-PQ search)
```
julianmi: nit: naming the 1e9 and 2e9 constants might be clearer.

examples/build.sh
```diff
@@ -93,7 +93,7 @@ build_example() {
   # Configure
   cmake -S "${example_dir}" -B "${build_dir}" \
   -DCMAKE_BUILD_TYPE=${BUILD_TYPE} \
-  -DCUVS_NVTX=OFF \
+  -DCUVS_NVTX=ON \
```
julianmi, line 96: Was this intentional?

cpp/bench/ann/src/cuvs/cuvs_cagra_hnswlib.cu
```diff
@@ -42,36 +42,7 @@ auto parse_build_param(const nlohmann::json& conf) ->
   ::parse_build_param<T, IdxT>(conf, cagra_params);
   // If the users provides parameter M, we can use the CAGRA-HNSW heuristics to find optimal
   // parameters for the dataset and HNSW reference.
```
julianmi: This comment can probably be removed since the `from_hnsw_params` logic was moved.

cpp/src/neighbors/detail/cagra/cagra_helpers.cpp
```diff
@@ -0,0 +1,151 @@
...
+    cuvs::neighbors::cagra::helpers::optimize_workspace_size(
+      n_rows, graph_degree, intermediate_graph_degree, sizeof(uint32_t));
+
+  // The kmeans trainset is a large temporary float buffer allocated during IVF-PQ training.
+  // It is freed before the extend phase, so peak GPU = max(training_peak, extend_peak).
+  size_t kmeans_trainset_ratio = std::max<size_t>(
+    1,
+    n_rows / std::max<size_t>(params.build_params.kmeans_trainset_fraction * n_rows,
+                              params.build_params.n_lists));
+  size_t kmeans_n_rows  = n_rows / kmeans_trainset_ratio;
```
julianmi, line 126: nit: is this rounding fine or should this be ceil?

cpp/src/neighbors/detail/cagra/cagra_build.cuh
```diff
@@ -883,25 +836,28 @@ bool ace_check_use_disk_mode(bool use_disk,
...
+  size_t sub_partition_size =
+    static_cast<size_t>(imbalance_factor * vector_expansion_factor * (dataset_size / n_partitions));
+  auto [opt_host_ws_total, opt_dev_ws_total, opt_host_ws_fixed, opt_dev_ws_fixed] =
+    helpers::optimize_workspace_size(
+      sub_partition_size, graph_degree, intermediate_degree, sizeof(IdxT), guarantee_connectivity);
 
   // Optimistic memory model: focus on largest arrays, assumes all partitions are of equal size
   // For memory path:
   //   - Partition labels (core + augmented): 2 * dataset_size * sizeof(IdxT)
   //   - Backward ID mapping arrays (core + augmented): 2 * dataset_size * sizeof(IdxT)
```
julianmi:
```suggestion
//   - Partition labels (core + augmented): vector_expansion_factor * dataset_size * sizeof(IdxT)
//   - Backward ID mapping arrays (core + augmented): vector_expansion_factor * dataset_size * sizeof(IdxT)
```

cpp/src/neighbors/detail/cagra/cagra_build.cuh
```diff
@@ -883,25 +836,28 @@ bool ace_check_use_disk_mode(bool use_disk,
...
-  //   - Avg. per-partition graph during build: 2 * (dataset_size / n_partitions) * (intermediate +
-  //   final)
-  //   * sizeof(IdxT)
+  //   - Avg. per-partition dataset: vector_expansion_factor * (dataset_size / n_partitions) *
+  //   dataset_dim * sizeof(T)
+  //   - Avg. per-partition graph during build: vector_expansion_factor * (dataset_size /
+  //   n_partitions) * (intermediate + final) * sizeof(IdxT)
   //   - Final assembled graph: dataset_size * graph_degree * sizeof(IdxT)
   mem.partition_labels_size = 2 * dataset_size * sizeof(IdxT);
   mem.id_mapping_size       = 2 * dataset_size * sizeof(IdxT);
```
julianmi:
```suggestion
mem.partition_labels_size = vector_expansion_factor * dataset_size * sizeof(IdxT);
mem.id_mapping_size       = vector_expansion_factor * dataset_size * sizeof(IdxT);
```

cpp/src/neighbors/detail/cagra/cagra_build.cuh
```diff
@@ -883,25 +836,28 @@ bool ace_check_use_disk_mode(bool use_disk,
...
-  mem.sub_dataset_size =
-    imbalance_factor * 2 * (dataset_size / n_partitions) * dataset_dim * sizeof(T);
-  mem.sub_graph_size = imbalance_factor * 2 * (dataset_size / n_partitions) *
-                       (intermediate_degree + graph_degree) * sizeof(IdxT);
+  mem.sub_dataset_size      = sub_partition_size * dataset_dim * sizeof(T);
+  mem.sub_graph_size   = sub_partition_size * (intermediate_degree + graph_degree) * sizeof(IdxT);
   mem.cagra_graph_size = dataset_size * graph_degree * sizeof(IdxT);
   mem.total_size       = mem.partition_labels_size + mem.id_mapping_size + mem.sub_dataset_size +
-                   mem.sub_graph_size + mem.cagra_graph_size;
+                   mem.sub_graph_size + mem.cagra_graph_size + opt_host_ws_total + 2e9;
```
julianmi: Name these magic numbers

cpp/src/neighbors/detail/cagra/cagra_build.cuh
```diff
@@ -980,10 +942,10 @@ void ace_validate_disk_mode_partitions(size_t& n_partitions,
   // Host memory model for disk mode:
   //   - Partition labels (core + augmented): 2 * dataset_size * sizeof(IdxT)
   //   - ID mapping arrays (core + augmented): 2 * dataset_size * sizeof(IdxT)
```
julianmi:
```suggestion
//   - Partition labels (core + augmented): vector_expansion_factor * dataset_size * sizeof(IdxT)
//   - ID mapping arrays (core + augmented): vector_expansion_factor * dataset_size * sizeof(IdxT)
```

cpp/src/neighbors/detail/cagra/cagra_helpers.cpp
```diff
@@ -0,0 +1,151 @@
...
+               "intermediate_degree must be greater than or equal to graph_degree");
+
+  // MST optimization memory (host only)
+  size_t mst_host       = 0;
+  size_t mst_host_fixed = 0;
+  if (mst_optimize) {
+    mst_host = n_rows * index_size;                  // mst_graph_num_edges
+    mst_host += n_rows * graph_degree * index_size;  // mst_graph allocated in optimize
+    mst_host += n_rows * graph_degree * index_size;  // mst_graph allocated in mst_optimize
+    mst_host += n_rows * index_size * 7;             // vectors with _max_edges suffix
```
julianmi: Why 7?

cpp/src/neighbors/detail/cagra/cagra_helpers.cpp
```diff
@@ -0,0 +1,151 @@
...
+    mst_host_fixed += (graph_degree - 1) * (graph_degree - 1) * index_size;  // iB_candidates
+    mst_host += mst_host_fixed;
+  }
+
+  // batchsize for both prune and combine stages
+  size_t batch_size = std::min(static_cast<size_t>(256 * 1024), n_rows);
```
julianmi: Should the batch size be a global constant to prevent updating it only in one place in the future?

cpp/src/neighbors/detail/cagra/cagra_helpers.cpp
```diff
@@ -0,0 +1,151 @@
...
+
+  // Prune stage memory
+  // We neglect 8 bytes (both on host and device) for stats
+  size_t prune_dev_fixed = batch_size * intermediate_degree * 1;  // detour count (uint8_t)
```
julianmi: nit: I think we can either leave out the factor `1` or make it explicit:
```suggestion
  size_t prune_dev_fixed = batch_size * intermediate_degree * sizeof(uint8_t);  // detour count (uint8_t)
```

cpp/src/neighbors/detail/hnsw.hpp
```diff
@@ -1112,6 +1250,74 @@ std::unique_ptr<index<T>> from_cagra(
...
+  if (params.hierarchy != HnswHierarchy::CPU) {
+    int64_t n_rows       = dataset.has_value() ? dataset->extent(0) : cagra_index.size();
+    int64_t dim          = dataset.has_value() ? dataset->extent(1) : cagra_index.dim();
+    int graph_degree_int = static_cast<int>(cagra_index.graph().extent(1));
+
+    // Instantiate a size-1 dummy to read the exact per-element host footprint.
+    auto dummy_index = std::make_unique<index_impl<T>>(dim, cagra_index.metric(), params.hierarchy);
+    auto dummy_algo  = std::make_unique<hnswlib::HierarchicalNSW<typename hnsw_dist_t<T>::type>>(
+      dummy_index->get_space(), 1, (graph_degree_int + 1) / 2, params.ef_construction);
+    // The base level dominates; add ~12% for hierarchy linklists and transients.
```
julianmi: We might want to explain the 12%.

cpp/src/neighbors/detail/hnsw.hpp
```diff
@@ -1112,6 +1250,74 @@ std::unique_ptr<index<T>> from_cagra(
...
+    RAFT_LOG_INFO(
+      "hnsw::from_cagra - in-memory HNSW requires ~%4.1f GB host mem, available %4.1f GB",
+      required_host / 1e9,
+      available_host / 1e9);
+
+    if (required_host >= available_host) {
+      RAFT_LOG_INFO("Not enough host memory for in-memory HNSW. Spilling HNSW index to disk.");
+
+      // Determine output directory from ACE build_dir if configured, else /tmp.
+      std::string index_directory = "/tmp";
```
julianmi: Should we create a random sub-directory instead to prevent clashes?

cpp/src/neighbors/detail/hnsw.hpp
```diff
@@ -1269,9 +1475,10 @@ void deserialize(raft::resources const& res,
  * @brief Build an HNSW index on the GPU using CAGRA graph building algorithm
  *
  * This function builds an HNSW index
- * 1. Converting HNSW parameters to CAGRA parameters (ACE configuration by default)
- * 2. Building a CAGRA index
- * 3. Converting the CAGRA index to HNSW format
+ * 1. Converting HNSW parameters to CAGRA parameters
+ * 2. inspect memory usage and decide to use ACE build or in-memory build
+ * 3. Building a CAGRA index  (ACE configuration by default if memory is limited)
+ * 4. Converting the CAGRA index to HNSW format (disk-backed or in-memory)
```
julianmi:
```suggestion
 * 2. Inspect memory requirements (fall back to ACE algorithm if memory constrained)
 * 3. Building a CAGRA index with the chosen algorithm in (2)
 * 4. Converting the CAGRA index to HNSW format (in-memory or ACE disk-backed)
```

cpp/src/neighbors/detail/hnsw.hpp
```diff
@@ -1280,34 +1487,69 @@ std::unique_ptr<index<T>> build(raft::resources const& res,
...
+      res, dataset.extents(), sizeof(T), cagra_params);
+    auto [available_host, available_dev] = get_available_memory();
+
+    RAFT_LOG_INFO("CAGRA in memory build, required host mem %4.1f GB, GPU mem %4.1f GB",
+                  required_host / 1e9,
+                  required_dev / 1e9);
+    RAFT_LOG_INFO("Available                       host mem %4.1f GB, GPU mem %4.1f GB",
+                  available_host / 1e9,
+                  available_dev / 1e9);
+    if (required_host < available_host && required_dev < available_dev) {
```
julianmi, line 1573: Shouldn't we decide between ACE in-memory or disk based? In-memory might be beneficial if we have enough host memory but not device memory. We don't have to spill to the disk then.

cpp/src/neighbors/detail/hnsw.hpp
```diff
@@ -1280,34 +1487,69 @@ std::unique_ptr<index<T>> build(raft::resources const& res,
...
+      cagra_params.intermediate_graph_degree = params.M * 3;
+      cagra_params.graph_degree              = params.M * 2;
+    }
+
+    // Configure ACE parameters for CAGRA
+    cuvs::neighbors::cagra::graph_build_params::ace_params cagra_ace_params;
+    cagra_ace_params.npartitions        = ace_params.npartitions;
+    cagra_ace_params.ef_construction    = params.ef_construction;
+    cagra_ace_params.build_dir          = ace_params.build_dir;
+    cagra_ace_params.use_disk           = ace_params.use_disk;
```
julianmi, line 1592: Wouldn't this use the default if `ace_params` is unset, i.e. `use_disk = false`, even when we have would need a disk-backed ACE approach based on the memory requirements?

examples/cpp/CMakeLists.txt
```diff
@@ -16,6 +16,8 @@ include(rapids-export)
 include(rapids-find)
 
 # ------------- configure project --------------#
+set(CMAKE_CXX_STANDARD 20)
+set(CMAKE_CUDA_STANDARD 20)
```
julianmi: Was this intential?

julianmi (commented):
Thanks @mfoerste4. These are great improvements!

cpp/src/neighbors/detail/cagra/cagra_build.cuh
julianmi: I understand your reasoning and I think it make sense to change it. To provide some historical context: this was only meant for debugging and testing purposes. I'm not sure if there are any use cases where one wants to limit the max. memory artificially above the actual free memory. It could happen by mistake though.

julianmi:
> I added 2GB static memory consumption for both GPU & Host. This is probably too conservative given that we already reduce the available usage to 80% of the actual value. Especially on small host memory machines this increases the number of partitions drastically. We might want to get rid of one of the limits. On the GPU I would prefer the constant 2GB (e.g. for workspace memory), but on the host side the percentage seems more natural. What do you think?

I agree that the host might stick to the percentage. We have much less control over the host which has many other processes running. Also, high memory pressure reduces the performance significantly. We might want to test with increasing the limit to 90% given the added 2 GB static memory.

python/cuvs/cuvs/tests/test_hnsw_ace.py
mfoerste4: I updated the criteria for host to not have a static but only dynamic cutoff which allows resetting this hard-to-explain change.

examples/build.sh
mfoerste4: Yes, this is required for the memory tracker to fetch nvtx scopes.

cpp/src/neighbors/detail/cagra/cagra_helpers.cpp
mfoerste4: Moved it as constant to cagra.hpp

cpp/src/neighbors/detail/hnsw.hpp
mfoerste4: Default is false, ace build then decides whether it is required.

examples/cpp/src/hnsw_openai_example.cu
tfeher: We keep the default for now as `/tmp/ace_build`

cpp/src/neighbors/detail/cagra/cagra_helpers.cpp
tfeher: By counting the size of the relevant arrays (temporary arrays in mst_optimization)

cpp/src/neighbors/detail/hnsw.hpp
tfeher: Build ace decides to use the disk if needed.

cpp/src/neighbors/detail/cagra/cagra_helpers.cpp
julianmi: I'm mostly arguing against magic numbers and better commenting. I feel like these are difficult to maintain.

cpp/src/neighbors/detail/hnsw.hpp
julianmi: Rereading this, I understand why we split the logic. It still feels a bit strange to check the memory requirements here but not make the disk decision. We might want to add a comment explaining this.

cpp/src/neighbors/detail/hnsw.hpp
```diff
@@ -1280,34 +1487,69 @@ std::unique_ptr<index<T>> build(raft::resources const& res,
...
+    cagra_ace_params.max_host_memory_gb = ace_params.max_host_memory_gb;
+    cagra_ace_params.max_gpu_memory_gb  = ace_params.max_gpu_memory_gb;
+    cagra_params.graph_build_params     = cagra_ace_params;
+
+    RAFT_LOG_INFO(
+      "hnsw::build - Building HNSW index using ACE with %zu partitions, ef_construction=%zu",
+      cagra_ace_params.npartitions,
+      cagra_ace_params.ef_construction);
```
julianmi: I realize that you didn't add this code but given the `nparitions` heuristic does it make sense to move this logging to when we've finalized the number of partitions in CAGRA ACE build?

mfoerste4: Yes, the logging is not helpful as there is plenty ACE output afterwards regarding npartitions. I will remove it.

cpp/src/neighbors/detail/cagra/cagra_build.cuh
mfoerste4: I have chosen the minimum and added a warning.

cpp/include/cuvs/neighbors/cagra.hpp
```diff
@@ -3245,6 +3245,39 @@ namespace neighbors {
...
+ * @param[in] dtype_size size of dataset datatype in bytes
+ * @param[in] input_is_float whether the dataset element type is `float`
+ * @param[in] cparams CAGRA index building parameters
+ *
+ * @return pair of [host_size, device_size] memory sizes in bytes
+ */
+std::pair<size_t, size_t> cagra_build_mem_usage(raft::resources const& res,
+                                                raft::matrix_extent<int64_t> dataset,
+                                                size_t dtype_size,
+                                                bool input_is_float,
```
achirkin: Maybe it would make sense to use a single `cudaDataType_t` in place of these two like we do in some parameter structs?

cpp/include/cuvs/neighbors/cagra.hpp
```diff
@@ -3268,6 +3301,9 @@ void optimize(raft::resources const& handle,
               raft::host_matrix_view<uint32_t, int64_t, raft::row_major> knn_graph,
               raft::host_matrix_view<uint32_t, int64_t, raft::row_major> new_graph);
 
+/** The batch size for the CAGRA optimize stage. */
+constexpr static size_t kOptimizeBatchSize = 256 * 1024;
```
achirkin: Does this constant have to be in the public headers?

cpp/src/neighbors/detail/cagra/cagra_build.cuh
```diff
@@ -874,34 +828,52 @@ bool ace_check_use_disk_mode(bool use_disk,
...
+    } else {
+      RAFT_LOG_INFO("ACE: Using overridden host memory limit: %.2f GiB",
+                    max_host_memory_gb.value());
+      mem.available_host_memory = configured_host_memory;
+    }
   } else {
     mem.available_host_memory = cuvs::util::get_free_host_memory();
   }
+  size_t sub_partition_size =
+    static_cast<size_t>(imbalance_factor * vector_expansion_factor * (dataset_size / n_partitions));
```
achirkin: Should we use `raft::div_rounding_up_safe` here in place of rounding-down division for an extra safety? It probably doesn't make a big difference though - because the error is eaten by the two heuristics-based factors on the left.

cpp/src/neighbors/detail/cagra/cagra_build.cuh
```diff
@@ -811,57 +811,11 @@ struct ace_memory_requirements {
...
-    combine_dev += 2 * batch_size * graph_degree * index_size;  // d_mst_graph(2*batch)
-    combine_dev += 2 * batch_size * sizeof(uint32_t);           // d_mst_graph_num_edges(2*batch)
-  }
-
-  size_t total_host = mst_host + combine_host;
-  size_t total_dev  = std::max(prune_dev, rev_dev + combine_dev);
-
-  return std::make_pair(total_host, total_dev);
-}
+constexpr double vector_expansion_factor    = 2.0;
```
achirkin: Nitpick: could you please add short one-line description comments to each constant? I'm not sure what's the logic behind the `vector_expansion_factor`, for example.

cpp/src/neighbors/detail/cagra/cagra_build.cuh
```diff
@@ -914,17 +886,35 @@ bool ace_check_use_disk_mode(bool use_disk,
...
-  bool gpu_memory_limited =
-    static_cast<size_t>(usable_gpu_memory_fraction * mem.available_gpu_memory) <
-    std::max(mem.sub_graph_size, mem.sub_dataset_size);
+
+  // what we need is maximum of:
+  // * IVF-PQ on partition  (sub_dataset_size, uncompressed upper bound)
+  // * optimize workspace (opt_dev_ws_total)
+  // + some extra workspace (IVF-PQ search, ...)
+  size_t gpu_memory_required =
+    std::max(mem.sub_dataset_size, opt_dev_ws_total) + extra_gpu_workspace_size;
```
achirkin, line 914: From this logic, I think we can replace the `extra_gpu_workspace_size` constant with the [raft::resource::get_workspace_total_bytes(res)](https://github.com/rapidsai/raft/blob/1dde2731d93c08444e75f9e12bf4308dbc2ce8a0/cpp/include/raft/core/resource/device_memory_resource.hpp#L220). The function returns the allocation limit for the workspace resource, the logic is to make sure the workspace resource always has the GPU memory to back up to its limit of bytes (for ivf_pq::search and others to not fail with OOM).

cpp/src/neighbors/detail/cagra/graph_core.cuh
```diff
@@ -842,11 +844,8 @@ void merge_graph_gpu(
...
   auto d_check_num_protected_edges = raft::make_device_scalar<uint32_t>(res, 1u);
 
-  // The batchsize is statically set to 256 * 1024 which corresponds to 256MB for a graph
-  // degree of 128 and 16byte index type. This is a trade-off between memory usage and performance.
-  // When choosing dynamically based on available memory, we would also need to modify the static
-  // size assumption in the cagra_build.cuh::optimize_workspace_size function.
-  uint32_t batch_size      = static_cast<uint32_t>(std::min<uint64_t>(graph_size, 256 * 1024));
+  uint32_t batch_size =
+    static_cast<uint32_t>(std::min<uint64_t>(graph_size, helpers::kOptimizeBatchSize));
   const uint32_t num_batch = (graph_size + batch_size - 1) / batch_size;
```
achirkin:
```suggestion
  const uint32_t num_batch = raft::div_rounding_up_safe(graph_size, batch_size);
```

cpp/src/neighbors/detail/hnsw.hpp
```diff
@@ -1280,34 +1519,64 @@ std::unique_ptr<index<T>> build(raft::resources const& res,
...
+        ? std::get<graph_build_params::ace_params>(params.graph_build_params)
+        : graph_build_params::ace_params{};
+
+    // Only override CAGRA params if user explicitly provided ACE params;
+    // otherwise preserve the values from from_hnsw_params computed above.
+    if (std::holds_alternative<graph_build_params::ace_params>(params.graph_build_params)) {
+      cagra_params.metric                    = params.metric;
+      cagra_params.intermediate_graph_degree = params.M * 3;
+      cagra_params.graph_degree              = params.M * 2;
+    }
```
achirkin: Why do we need this? As far as I recall, cagra::hnsw_heuristic_type::SAME_GRAPH_FOOTPRINT already gives exactly this mapping.

achirkin (changes requested):
Looks good overall. A few small suggestions below.

cpp/src/neighbors/detail/hnsw.hpp
tfeher: We can end up in this branch not only if the system memory is insufficient but also when the user explicitly sets `params.graph_build_params` to `ace_params`. Therefore it makes sense to keep the value that is (potentially) provided by the user.

Malte raised the point that we could move the memory checks to `cagra_build`. We can re-evaluate this in a follow up PR.

cpp/src/neighbors/ivf_pq_index.cu
```diff
@@ -817,6 +818,44 @@ raft::device_matrix_view<const half, uint32_t, raft::row_major> index<IdxT>::cen
   return impl_->centers_half(res);
 }
 
+namespace helpers {
+size_t compressed_dataset_size(raft::resources const& res,
+                               raft::matrix_extent<int64_t> dataset,
+                               cuvs::neighbors::ivf_pq::index_params params)
+{
+  auto idx = cuvs::neighbors::ivf_pq::index<int64_t>(res, params, (uint32_t)dataset.extent(1));
```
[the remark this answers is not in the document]
tfeher: The memory usage for an empty index is minimal. In the scenario where this does not fit into memory we cannot build an index even in fallback mode.

cpp/src/neighbors/ivf_pq_index.cu
```diff
@@ -817,6 +818,44 @@ raft::device_matrix_view<const half, uint32_t, raft::row_major> index<IdxT>::cen
...
+
+  size_t pq_centers = idx.pq_len() * pq_book_size * params.pq_dim * sizeof(float);
+  size_t pq_dataset = raft::ceildiv<size_t>(dataset.extent(0), kIndexGroupSize) * kIndexGroupSize *
+                      raft::ceildiv<size_t>(params.pq_dim, pq_chunk) * kIndexGroupVecLen;
+  size_t indices         = dataset.extent(0) * sizeof(int64_t);
+  size_t rotation_matrix = idx.rot_dim() * idx.rot_dim() * sizeof(float);
+  size_t list_offsets    = (params.n_lists + 1) * sizeof(int64_t);
+  size_t list_sizes      = params.n_lists * sizeof(int64_t);
+  size_t centers         = params.n_lists * idx.dim_ext() * sizeof(float);
+  size_t centers_rot     = params.n_lists * idx.rot_dim() * sizeof(float);
```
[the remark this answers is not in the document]
tfeher: Fixed both.

examples/cpp/CMakeLists.txt
tfeher: Leftover from previous tests, removed.

cpp/src/neighbors/detail/cagra/cagra_helpers.cpp
tfeher: I have updated the comment.

We all agree that this will be difficult to maintain. Plan is to deprecate these soon, when https://github.com/rapidsai/raft/pull/2961 becomes available.

tfeher (commented):
Fixed few smaller issues.

examples/cpp/CMakeLists.txt
```diff
@@ -33,9 +33,11 @@ add_executable(BRUTE_FORCE_EXAMPLE src/brute_force_bitmap.cu)
 add_executable(CAGRA_EXAMPLE src/cagra_example.cu)
 add_executable(CAGRA_FILTER_UDF_EXAMPLE src/cagra_filter_udf_example.cu)
 add_executable(CAGRA_HNSW_ACE_EXAMPLE src/cagra_hnsw_ace_example.cu)
+add_executable(CAGRA_HNSW_ACE_BUILD_EXAMPLE src/cagra_hnsw_ace_build.cu)
```
KyleFromNVIDIA, line 40: Please move this one line above to maintain alphabetical ordering.

KyleFromNVIDIA (changes requested):
Just one small request, otherwise looks good

cpp/include/cuvs/neighbors/cagra.hpp
mfoerste4: I did not find a better place as it is drawn in by a host-only `cagra_helpers.cpp` file which prevents other internal places like` utils.hpp` that draw cuda includes.

cpp/src/neighbors/detail/hnsw.hpp
tfeher: We can land here if the user explicitly defined ace params
```c++ graph build params
hnsw::index_params hnsw_params;
// define ace params
hnsw_params.graph_build_params = ace_params;
```
In that case `cagra_params` is just a default params struct defined in line 1522, and we still need to define it's fields. @mfoerste4  We should probably initialize that with the default heuristics, and override `graph_build_params` when needed.

KyleFromNVIDIA (approved):

examples/cpp/CMakeLists.txt
```diff
@@ -49,13 +51,18 @@ target_link_libraries(
   CAGRA_FILTER_UDF_EXAMPLE PRIVATE cuvs::cuvs $<TARGET_NAME_IF_EXISTS:conda_env>
 )
 target_link_libraries(CAGRA_HNSW_ACE_EXAMPLE PRIVATE cuvs::cuvs $<TARGET_NAME_IF_EXISTS:conda_env>)
+target_link_libraries(
+  CAGRA_HNSW_ACE_BUILD_EXAMPLE PRIVATE cuvs::cuvs $<TARGET_NAME_IF_EXISTS:conda_env>
+)
```
KyleFromNVIDIA, line 60: This should also be one line up to maintain alphabetical order.

KyleFromNVIDIA (changes requested):

KyleFromNVIDIA (approved):
Approving as I will be out for the first few days of next week and don't want to hold this up. Please address my above comment before merging.

cpp/include/cuvs/neighbors/cagra.hpp
mfoerste4: I created a host only header `cagra_helpers.hpp` to hold the constant.

achirkin (approved):
Thanks for the updates! lgtm

dantegd (approved):

PR #1732, opened 2026-01-26
Add a `transform` function to ivf_pq

benfred:
This adds a `transform` api to ivf_pq, letting you pq encode vectors without adding them to the ivf-pq index. 

Unlike the preprocessing pq-encode api, this takes into account the ivf cluster centroids just the same as the ivf-pq index does . The pq-encoded vectors will have the same values as if added to the index,  but won't be stored in an inverted list format . This means that you pass in a dataset, and get back a vector with the label of which cluster the vector would be added to - as well as a matrix containing the pq-encoded dataset.

### DIFF
--- a/cpp/include/cuvs/neighbors/ivf_pq.hpp
+++ b/cpp/include/cuvs/neighbors/ivf_pq.hpp
@@ -1975,8 +1975,8 @@ void search(raft::resources const& handle,
  * @param[in] dataset a device matrix view to a row-major matrix [n_rows, index.dim()]
  * @param[out] output_labels a device vector view [n_rows] that will get populaterd with the
  * cluster ids (labels) for each vector in the input dataset
- * @param[out] output_dataset a device matrix view [n_rows, pq_dim] that will get populated with
- * the pq-encoded dataset
+ * @param[out] output_dataset a device matrix view [n_rows, ceildiv(index.pq_dim() *
+ * index.pq_bits(), 8)]] that will get populated with the pq-encoded dataset
  */
 void transform(raft::resources const& handle,
                const cuvs::neighbors::ivf_pq::index<int64_t>& index,
--- a/cpp/src/neighbors/ivf_pq/ivf_pq_transform.cuh
+++ b/cpp/src/neighbors/ivf_pq/ivf_pq_transform.cuh
@@ -107,8 +107,10 @@ void transform(raft::resources const& res,
   IdxT n_rows = dataset.extent(0);
   RAFT_EXPECTS(output_labels.extent(0) == n_rows, "incorrect number of rows in output_labels");
   RAFT_EXPECTS(output_dataset.extent(0) == n_rows, "incorrect number of rows in output_dataset");
-  RAFT_EXPECTS(output_dataset.extent(1) == index.pq_dim(),
-               "incorrect number of cols in output_dataset");
+
+  RAFT_EXPECTS(
+    output_dataset.extent(1) == raft::ceildiv<uint32_t>(index.pq_dim() * index.pq_bits(), 8),
+    "incorrect number of cols in output_dataset");
 
   raft::common::nvtx::range<cuvs::common::nvtx::domain::cuvs> fun_scope(
     "ivf_pq::transform(n_rows = %u, dim = %u)", n_rows, dataset.extent(1));
--- a/python/cuvs/cuvs/neighbors/ivf_pq/ivf_pq.pyx
+++ b/python/cuvs/cuvs/neighbors/ivf_pq/ivf_pq.pyx
@@ -1021,11 +1021,13 @@ def transform(Index index, input_dataset, output_labels=None, output_dataset=Non
     _check_input_array(output_labels_ai, [np.dtype('uint32')],
                        exp_rows=n_samples)
 
+    n_output_cols = int(np.ceil(index.pq_dim * index.pq_bits / 8))
+
     if output_dataset is None:
-        output_dataset = device_ndarray.empty((n_samples, pq_dim), dtype='uint8')
+        output_dataset = device_ndarray.empty((n_samples, n_output_cols), dtype='uint8')
     output_dataset_ai = wrap_array(output_dataset)
     _check_input_array(output_dataset_ai, [np.dtype('uint8')],
-                       exp_rows=n_samples, exp_cols=pq_dim)
+                       exp_rows=n_samples, exp_cols=n_output_cols)
 
     cdef cuvsResources_t res = <cuvsResources_t>resources.get_c_obj()
 
--- a/python/cuvs/cuvs/tests/test_ivf_pq.py
+++ b/python/cuvs/cuvs/tests/test_ivf_pq.py
@@ -339,10 +339,10 @@ def test_build_precomputed(codebook_kind, metric):
 @pytest.mark.parametrize("codebook_kind", ["subspace", "cluster"])
 @pytest.mark.parametrize("dtype", [np.float32, np.float16, np.int8, np.uint8])
 @pytest.mark.parametrize("n_rows", [500, 100000])
-def test_transform(codebook_kind, dtype, n_rows):
+@pytest.mark.parametrize("pq_bits", [6, 8])
+def test_transform(codebook_kind, dtype, n_rows, pq_bits):
     n_cols = 32
     n_lists = 50
-    pq_bits = 8
     pq_dim = 8
 
     # build the ivf-pq index
@@ -363,7 +363,9 @@ def test_transform(codebook_kind, dtype, n_rows):
 
     # collect the pq-encoded data from the index for the expected data
     labels_exp = np.zeros((n_rows), dtype="uint32")
-    transformed_exp = np.zeros((n_rows, pq_dim), dtype="uint8")
+    transformed_exp = np.zeros(
+        (n_rows, int(np.ceil(index.pq_dim * index.pq_bits / 8))), dtype="uint8"
+    )
     for cluster_id in range(n_lists):
         ids = index.list_indices(cluster_id).copy_to_host()
         labels_exp[ids] = cluster_id


cpp/src/neighbors/ivf_pq/ivf_pq_contiguous_list_data_impl.cuh
```diff
@@ -1,6 +1,7 @@
 /*
  * SPDX-FileCopyrightText: Copyright (c) 2025-2026, NVIDIA CORPORATION.
  * SPDX-License-Identifier: Apache-2.0
+ * SPDX-Lice/se-Identifier: Apache-2.0
```
tarang-jain: Typo?

cpp/src/neighbors/ivf_pq/ivf_pq_transform.cuh
```diff
@@ -0,0 +1,129 @@
...
+
+  auto encoder =
+    encode_vectors<kSubWarpSize, IdxT>{pq_centers, dataset_residual, codebook_kind, cluster_ix};
+  const uint32_t pq_dim = dataset_residual.extent(1) / pq_centers.extent(1);
+
+  for (uint32_t j = 0; j < pq_dim; j++) {
+    uint8_t code = encoder(row_ix, j);
+    if (lane_id == 0) { output_dataset(row_ix, j) = code; }
+
+    // TODO: do we want one code per byte (like here) or do we want tightly packed matrices?
```
tarang-jain: Can we use write_vector_flat here? (https://github.com/rapidsai/cuvs/blob/6f2ee7d254fc01130e589c67e7da6f0b422d5060/cpp/src/neighbors/ivf_pq/ivf_pq_codepacking.cuh#L202)

I have tightly packed in that kernel. Not sure what the preference for this API is, but Faiss does tight packing of codes for CPU indexes.

c/src/neighbors/ivf_pq.cpp
```diff
@@ -569,3 +591,33 @@ extern "C" cuvsError_t cuvsIvfPqIndexGetListIndices(cuvsIvfPqIndex_t index,
...
+      // Verify dtypes of inputs
+      auto& input_dataset_dl = input_dataset->dl_tensor;
+      RAFT_EXPECTS(input_dataset_dl.dtype.code == kDLFloat && input_dataset_dl.dtype.bits == 32,
+                   "input_dataset must have a float32 dtype");
+      auto& output_labels_dl = output_labels->dl_tensor;
+      RAFT_EXPECTS(output_labels_dl.dtype.code == kDLUInt && output_labels_dl.dtype.bits == 32,
+                   "output_labels must have a uint32 dtype ");
+      auto& output_dataset_dl = output_dataset->dl_tensor;
+      RAFT_EXPECTS(output_dataset_dl.dtype.code == kDLUInt && output_dataset_dl.dtype.bits == 8,
+                   "output_labels must have a uint8 dtype");
```
dantegd:
```suggestion
                   "output_dataset must have a uint8 dtype");
```

cpp/include/cuvs/neighbors/ivf_pq.hpp
```diff
@@ -1959,6 +1959,12 @@ void search(raft::resources const& handle,
             const cuvs::neighbors::filtering::base_filter& sample_filter =
               cuvs::neighbors::filtering::none_sample_filter{});
 
+void transform(raft::resources const& handle,
+               const cuvs::neighbors::ivf_pq::index<int64_t>& index,
+               raft::device_matrix_view<const float, int64_t, raft::row_major> dataset,
+               raft::device_vector_view<uint32_t, int64_t> output_labels,
+               raft::device_matrix_view<uint8_t, int64_t> output_dataset);
```
dantegd, line 1985: We should also add the other types here, no? 


```suggestion
void transform(raft::resources const& handle,
               const cuvs::neighbors::ivf_pq::index<int64_t>& index,
               raft::device_matrix_view<const float, int64_t, raft::row_major> dataset,
               raft::device_vector_view<uint32_t, int64_t> output_labels,
               raft::device_matrix_view<uint8_t, int64_t> output_dataset);
               
void transform(raft::resources const& handle,
               const cuvs::neighbors::ivf_pq::index<int64_t>& index,
               raft::device_matrix_view<const half, int64_t, raft::row_major> dataset,
               raft::device_vector_view<uint32_t, int64_t> output_labels,
               raft::device_matrix_view<uint8_t, int64_t> output_dataset);
               
void transform(raft::resources const& handle,
               const cuvs::neighbors::ivf_pq::index<int64_t>& index,
               raft::device_matrix_view<const int8_t, int64_t, raft::row_major> dataset,
               raft::device_vector_view<uint32_t, int64_t> output_labels,
               raft::device_matrix_view<uint8_t, int64_t> output_dataset);
               
void transform(raft::resources const& handle,
               const cuvs::neighbors::ivf_pq::index<int64_t>& index,
               raft::device_matrix_view<const uint8_t, int64_t, raft::row_major> dataset,
               raft::device_vector_view<uint32_t, int64_t> output_labels,
               raft::device_matrix_view<uint8_t, int64_t> output_dataset);                                                                             
```

python/cuvs/cuvs/tests/test_ivf_pq.py
```diff
@@ -334,3 +334,45 @@ def test_build_precomputed(codebook_kind, metric):
...
+        labels_exp[ids] = cluster_id
+        transformed_exp[ids] = index.list_data(cluster_id).copy_to_host()
+
+    # call transform to get actual results, and make sure we're equal to expected values
+    labels, transformed = ivf_pq.transform(index, dataset_device)
+
+    np.testing.assert_array_equal(labels.copy_to_host(), labels_exp)
+    np.testing.assert_allclose(
+        transformed.copy_to_host(), transformed_exp, rtol=1e-6
+    )
```
dantegd: shouldn't we use `assert_array_equal` here, since the PQ codes will be uint8 values?


```suggestion
    np.testing.assert_array_equal(
        transformed.copy_to_host(), transformed_exp
    )
```

python/cuvs/cuvs/neighbors/ivf_pq/ivf_pq.pyx
```diff
@@ -981,3 +981,68 @@ def extend(Index index, new_vectors, new_indices, resources=None):
         ))
 
     return index
+
+
+@auto_sync_resources
+def transform(Index index, input_dataset, output_labels=None, output_dataset=None, resources=None):
+    """
+    Transform a dataset by appyling pq-encoded to the vectors.
```
dantegd:
```suggestion
    Transform a dataset by applying pq-encoded to the vectors.
```

cpp/src/neighbors/ivf_pq/ivf_pq_transform.cuh
```diff
@@ -0,0 +1,129 @@
...
+  }
+}
+
+template <typename T, typename IdxT>
+void transform(raft::resources const& res,
```
dantegd: We probably want  to add a NVTX range for profiling consistency with other IVFPQ functions

cpp/include/cuvs/neighbors/ivf_pq.hpp
```diff
@@ -1959,6 +1959,12 @@ void search(raft::resources const& handle,
             const cuvs::neighbors::filtering::base_filter& sample_filter =
               cuvs::neighbors::filtering::none_sample_filter{});
 
+void transform(raft::resources const& handle,
```
dantegd, line 1981: We should add a docstring to the function(s)

cpp/src/neighbors/ivf_pq/ivf_pq_transform.cuh
```diff
@@ -0,0 +1,129 @@
...
+  cuvs::cluster::kmeans_balanced::predict(res,
+                                          kmeans_params,
+                                          dataset,
+                                          raft::make_const_mdspan(cluster_centers.view()),
+                                          output_labels,
+                                          utils::mapping<float>{});
+
+  // Compute the residuals for each vector in the dataset
+  auto dataset_residuals =
+    raft::make_device_mdarray<float>(res, mr, raft::make_extents<IdxT>(n_rows, index.rot_dim()));
```
dantegd, line 65: This temporary buffer for residuals is roughly the same size as the input dataset for float inputs, but for int8/uint8 inputs the residuals would be ~4x larger than the input, no? 

This could cause unexpected OOMs for large datasets potentially, so a future batching improvement might be worth considering, but in the meantime it might be good to document this so it doesn't surprise users. Not a big concern, but wanted to note it for future reference as well (and see I'm not missing something).

cpp/src/neighbors/ivf_pq/ivf_pq_transform.cuh
benfred: done!

cpp/src/neighbors/ivf_pq/ivf_pq_transform.cuh
benfred: thanks for the pointer! Thats exactly the functionality I was looking for - I'm using write_vector_flat in the last commit

cpp/src/neighbors/ivf_pq/ivf_pq_transform.cuh
```diff
@@ -0,0 +1,127 @@
...
+
+  // The cluster centers in the index are stored padded, which is not acceptable by
+  // the kmeans_balanced::predict. Thus, we need the restructuring raft::copy.
+  auto stream           = raft::resource::get_cuda_stream(res);
+  const auto n_clusters = index.n_lists();
+
+  auto cluster_centers =
+    raft::make_device_mdarray<float>(res, mr, raft::make_extents<IdxT>(n_clusters, index.dim()));
+
+  RAFT_CUDA_TRY(cudaMemcpy2DAsync(cluster_centers.data_handle(),
```
tarang-jain: maybe you can use `cuvs::neighbors::ivf_pq::detail::extract_centers` here.

cpp/src/neighbors/ivf_pq/ivf_pq_transform.cuh
tarang-jain: +1. We can certainly add batching here, especially since we are expecting to 4x the input.

benfred: I've changed to use the `batch_load_iterator` in this commit https://github.com/rapidsai/cuvs/pull/1732/changes/7108d930c89af97de5993ab883999855d16ada8e -

cpp/src/neighbors/ivf_pq/ivf_pq_transform.cuh
benfred: done w/ 98d63b4661e351d889871130c795f7e25414449c

PR #1738, opened 2026-01-27
[WIP] Clarify dataset ownership and allocation semantics

### LINKED ISSUE #1571: [FEA] Indexes to accept `Dataset` directly
The intention here is to provide a layer of indirection between the actual mdspan / training matrices and the indexes. One of the immediate benefits is to faciliate consistent ownership agreement between the user and the APIs. For example, CAGRA today will copy the entire training dataset into a strided/padded form, which now blurs the line between ownership- most users don't realize this copy has been made or even that the dataset needed to be strided/padded at all, which causes an unexpected memory spike and means the CAGRA index is now responsible for owning the copied dataset. This starts to introduce "sometimes owning" semantics, which we should avoid doing implicitly at all costs. 

The solution here is to have a `PaddedDataset` that a user can construct with their data. The `PaddedDataset` would then be accepted directly (as a view) by CAGRA. This makes the contract between the index and user very explicit and puts the control back in the user's hands.

### LINKED ISSUE #1574: [FEA] Standardize Datasets API for public end-user use
Currently, the `Datasets` classes are embedded inside internal implementations and not meant for direct public consumption by users. However, because of this, significant drawbacks have been created to the overall UX and memory management inside some of the indexes. For example, when an mdspan containing CAGRA training vectors has not been previously padded/strided to a proper offset, CAGRA will take it upon itself to copy the vectors into a "padded dataset" instance on behalf of the user. This forfeits user control, makes the UI confusing, creates memory spikes that have proven to be very confusing for the users, and blurs the memory ownership model (now the CAGRA index has become "sometimes owning", which is always very challening maintain and explain to users). 

Instead, we've decided we should invest more heavily in the Datasets infrastructure for being able to populate the necessary fields and pre-conditions, thus in the process also self-documenting these pre-conditions for the user. For example, CAGRA should accept a "PaddedDataset" instead of just an mdspan. This additional pre-condition now means that a user will need to construct an instance of "PaddedDataset" and they will be able to better manage an immediately note (as documented in the underlying factory functions) the potential for a copy. But more than anything else, they will now have full ownership and control over the memory that gets passed into the CAGRA index, so it'll always be a view from CAGRA's perspective. 

This will help address #1571

seunghwak:
Initial attempt to address https://github.com/rapidsai/cuvs/issues/1574 and https://github.com/rapidsai/cuvs/issues/1571.

Currently, `index` constructors accept an `mdspan` and internally invoke `make_strided|aligned_dataset` to construct a `strided_dataset` from input data with potentially unknown alignment. When the input is properly aligned, a zero-copy view is used; otherwise, an aligned copy is created. This implicit copy can unexpectedly double GPU memory usage, which may be surprising to API users.

Current API
```
   auto dataset = raft::make_device_matrix<float, int64_t>(res, n_rows, n_cols);
   auto knn_graph = raft::make_device_matrix<uint32_t, int64_t>(res, n_rows, graph_degree);

   // custom loading and graph creation
   // load_dataset(dataset.view());
   // create_knn_graph(knn_graph.view());
   // Wrap the existing device arrays into an index structure
   cagra::index<T, IdxT> index(res, metric, raft::make_const_mdspan(dataset.view()),
                               raft::make_const_mdspan(knn_graph.view()));
```

New API (explicitly create `strided_dataset`)
```
   auto dataset_matrix = raft::make_device_matrix<float, int64_t>(res, n_rows, n_cols);
   auto knn_graph = raft::make_device_matrix<uint32_t, int64_t>(res, n_rows, graph_degree);
   // custom loading and graph creation
   // load_dataset(dataset_matrix.view());
   // create_knn_graph(knn_graph.view());
   // create strided dataset (zero-copy if aligned, throws if not)
   auto dataset = make_strided_dataset_zerocopy(dataset_matrix.view());
   or
   // create strided_dataset (create an aligned copy)
   auto dataset = make_strided_dataset_owning(dataset_matrix.view());
   // Wrap the existing strided dataset into an index structure
   cagra::index<T, IdxT> index(res, metric, *dataset,
                               raft::make_const_mdspan(knn_graph.view()));
```

We deprecate the `index` constructors that take device `mdspan` (`device_matrix_view`) to avoid surprising implicit copy.

In addition, this PR makes an initial attempt to update the `build` functions to accept `strided_dataset`, in addition to `host_matrix_view`. At present, the `build` overload that takes `strided_dataset` supports only iterative-CAGRA for initial graph construction. Supporting NN-descent or IVF-PQ is more involved, as these implementations assume a row-major layout; extending them to handle strided layouts would require more substantial code changes.

Closes #1571 
Closes #1574

### DIFF
--- a/cpp/include/cuvs/neighbors/common.hpp
+++ b/cpp/include/cuvs/neighbors/common.hpp
@@ -316,7 +316,7 @@ inline constexpr bool is_strided_dataset_v = is_strided_dataset<DatasetT>::value
  * @return non-owning (zero-copy) current-device-accessible strided matrix
  */
 template <typename SrcT>
-auto make_strided_dataset_zerocopy(const raft::resources& res, const SrcT& src, uint32_t required_stride)
+auto make_strided_dataset_view(const raft::resources& res, const SrcT& src, uint32_t required_stride)
   -> std::unique_ptr<strided_dataset<typename SrcT::value_type, typename SrcT::index_type>>
 {
   using extents_type = typename SrcT::extents_type;
@@ -405,8 +405,8 @@ auto make_strided_dataset_owning(const raft::resources& res, const SrcT& src, ui
 /**
  * @brief Contstruct a strided matrix from any mdarray or mdspan.
  *
- * @deprecated Use make_strided_dataset_zerocopy (if the input matrix is properly aligned) or
- * make_strided_dataset_copy (if the input matrix is not properly aligned) instead.
+ * @deprecated Use make_strided_dataset_view (if the input matrix is properly aligned) or
+ * make_strided_dataset_owning (if the input matrix is not properly aligned) instead.
  *
  * This function constructs a non-owning view if the input satisfied two conditions:
  *
@@ -597,7 +597,7 @@ auto make_strided_dataset(
  * @return non-owning (zero-copy) current-device-accessible strided matrix
  */
 template <typename SrcT>
-auto make_aligned_dataset_zerocopy(const raft::resources& res, SrcT src, uint32_t align_bytes = 16)
+auto make_aligned_dataset_view(const raft::resources& res, SrcT src, uint32_t align_bytes = 16)
   -> std::unique_ptr<strided_dataset<typename SrcT::value_type, typename SrcT::index_type>>
 {
   using source_type      = std::remove_cv_t<std::remove_reference_t<SrcT>>;
@@ -605,7 +605,7 @@ auto make_aligned_dataset_zerocopy(const raft::resources& res, SrcT src, uint32_
   constexpr size_t kSize = sizeof(value_type);
   uint32_t required_stride =
     raft::round_up_safe<size_t>(src.extent(1) * kSize, std::lcm(align_bytes, kSize)) / kSize;
-  return make_strided_dataset_zerocopy(res, std::forward<SrcT>(src), required_stride);
+  return make_strided_dataset_view(res, std::forward<SrcT>(src), required_stride);
 }
 
 /**
@@ -635,7 +635,7 @@ auto make_aligned_dataset_owning(const raft::resources& res, SrcT src, uint32_t 
 /**
  * @brief Contstruct a strided matrix from any mdarray or mdspan.
  *
- * @deprecated Use make_aligned_dataset_zerocopy or make_aligned_dataset_owning instead.
+ * @deprecated Use make_aligned_dataset_view or make_aligned_dataset_owning instead.
  *
  * A variant `make_strided_dataset` that allows specifying the byte alignment instead of the
  * explicit stride length.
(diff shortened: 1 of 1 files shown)


seunghwak:
@cjnolet Let me know whether this PR aligns with what you are thinking.

cjnolet:
Hey @seunghwak, thank you for giving this some initial thought and sharing here. 

I think this is a good start. It might also be worth me mentioning some of the other dataset types that this design is going to prepare us for- in addition to the `padded_dataset`, we also intend to provide a `vpq_dataset` and a `pq_dataset`. 

I wonder if maybe we should make the dataset memory type explicit too? (`device_padded_dataset`, `device_padded_dataset_view`, `host_padded_dataset_view`)

The nice thing about the dataset abstraction is that it'll allow us to have the pq_dataset contain additoinal state like trained centroids and codebooks, which need to be passed along w/ the pq-encoded vectors because they are used to compute the distances. 

It's nice to have these things abstracted behind the dataset class so that, for example, we aren't having to provide additional additional overloads for each different combination we support- we can support a `dataset_view` as input to the `build()` functions and keep supporting new dataset types in behind the scenes, while not having to keep changing the APIs. 

```c++
   // create strided dataset (zero-copy if aligned, throws if not)
   auto dataset = make_strided_dataset_zerocopy(dataset_matrix.view());
```

I wonder if maybe instead of `zerocopy` and `owning` maybe we could use `make_strided_dataset_view` and `make_strided_dataset`? We could just always accept the view. Or do you not think view conveys the proper message to the user?

seunghwak:
>I wonder if maybe we should make the dataset memory type explicit too? (device_padded_dataset, device_padded_dataset_view, host_padded_dataset_view)

Yes, if there is a need to support padded data set for host as well. I though the current strided_dataset is implicitly device-only, considering

https://github.com/rapidsai/cuvs/blob/fcc5031d37dbefb6574cc49caec5d0d6f3100d4e/cpp/include/cuvs/neighbors/common.hpp#L165
https://github.com/rapidsai/cuvs/blob/fcc5031d37dbefb6574cc49caec5d0d6f3100d4e/cpp/include/cuvs/neighbors/common.hpp#L267

but if you see a need to support padded dataset for host data as well, I will explicitly create a class (or an alias like `host_matrix_view` and `device_matrix_view`) for host data as well. For consistency, we may better copy how we handle mdarray and mdspan in designing the dataset API.

>It's nice to have these things abstracted behind the dataset class so that, for example, we aren't having to provide additional additional overloads for each different combination we support- we can support a dataset_view as input to the build() functions and keep supporting new dataset types in behind the scenes, while not having to keep changing the APIs.

You mean that we want to take a `dataset` class (https://github.com/rapidsai/cuvs/blob/main/cpp/include/cuvs/neighbors/common.hpp#L140) object and internally dynamic cast to support different dataset types? I will update the PR in this direction but please correct me if I mis-interpreted your intention.

>I wonder if maybe instead of zerocopy and owning maybe we could use make_strided_dataset_view and make_strided_dataset? We could just always accept the view. Or do you not think view conveys the proper message to the user?

No problem, I will update the function names.

seunghwak:
https://github.com/rapidsai/cuvs/pull/1846 will replace this PR.

PR #1746, opened 2026-01-27
Use PQ API in CAGRA-Q + SCANN

### LINKED ISSUE #1575: [FEA] Use PQ public API in Vamana and SCANN
Vamana and SCANN make use of some PQ functions taken from the VPQ codebase.
Ideally they would use the public PQ API when ready to minimize binary size and have a single entry point for PQ-related code.
Depends on #1278

### LINKED ISSUE #1747: [FEA] Improve tests for SCANN
Currently SCANN doesn't seem to be thoroughly tested. Only the extents of the built index are checked in the C++ tests, but no qualitative test exist.

lowener:
Follow-up to the PQ PR #1278 .
Closes #1575 
Closes #1747
This PR removes the need to compile multiple times the same code for PQ in CAGRA-Q and SCANN, removing code duplication and improving build time.
CAGRA-Q can't use the new public API since it is using half for its math type so an private API function is used.

A small test is added to SCANN to make sure that the returned index is not complete garbage but more testing should be done there. (Created issue #1747 to track this)

This PR saves ~2-3 Mb on libcuvs.so compiled on a single architecture (141 Mb -> 138Mb)


cpp/src/neighbors/detail/cagra/factory.cuh
```diff
@@ -11,6 +11,9 @@
 #include "search_plan.cuh"
 #include "search_single_cta.cuh"
 
+#include <raft/core/resource/custom_resource.hpp>
+#include <raft/util/cache.hpp>
```
tarang-jain, line 15: unused headers?

cpp/src/preprocessing/quantize/detail/pq.cuh
```diff
@@ -328,4 +328,67 @@ void inverse_transform(
...
+
+  raft::linalg::map(res,
+                    vq_code_book.view(),
+                    cuvs::spatial::knn::detail::utils::mapping<NewMathT>{},
+                    raft::make_const_mdspan(src.vq_code_book.view()));
+  raft::linalg::map(res,
+                    pq_code_book.view(),
+                    cuvs::spatial::knn::detail::utils::mapping<NewMathT>{},
+                    raft::make_const_mdspan(src.pq_code_book.view()));
+  return cuvs::neighbors::vpq_dataset<NewMathT, IdxT>{
```
tarang-jain: While this is in the detail namespace, I would still avoid such `std::move` semantics. The caller can pre-allocate the buffer for the new `vpq_dataset` and this can be just a void function.

cpp/src/preprocessing/quantize/detail/pq.cuh
```diff
@@ -328,4 +328,67 @@ void inverse_transform(
...
+    res,
+    ps,
+    dataset,
+    raft::make_const_mdspan(pq_code_book.view()),
+    raft::make_const_mdspan(vq_code_book.view()),
+    raft::make_device_vector_view<label_t, IdxT>(nullptr, 0),
+    codes.view(),
+    true);
+
+  return cuvs::neighbors::vpq_dataset<MathT, IdxT>{
```
tarang-jain, line 377: Same comment as above.

cpp/src/neighbors/detail/cagra/factory.cuh
lowener: cache is used below line 138:
https://github.com/lowener/cuvs/blob/e4a6798d54ab989cd5af8662300047747e7b64e4/cpp/src/neighbors/detail/cagra/factory.cuh#L138-L140

cpp/src/neighbors/scann/detail/scann_build.cuh
```diff
@@ -157,45 +158,17 @@ index<T, IdxT> build(
...
+  pq_build_params.use_vq                       = false;  // We already computed residuals
+  pq_build_params.kmeans_n_iters               = params.pq_train_iters;
+  pq_build_params.max_train_points_per_pq_code = pq_n_rows_train / num_clusters;
+  pq_build_params.pq_kmeans_type               = cuvs::cluster::kmeans::kmeans_type::KMeansBalanced;
 
-  // Set up quantization bits and params
-  cuvs::neighbors::vpq_params pq_params;
-  pq_params.pq_bits = params.pq_bits;
-  pq_params.pq_dim  = dataset.extent(1) / params.pq_dim;
+  auto pq_quantizer = cuvs::preprocessing::quantize::pq::build(
```
rmaschal, line 172: This looks a lot cleaner, thank you :) 

Have you looked at the perf implications of the new vs. old impl at all? The old impl essentially trained each subspace sequentially in a for loop, which is pretty inefficient for low subspace dims. The cost is amortized away for larger datasets, but I wonder if we get anything back here.

cpp/src/neighbors/scann/detail/scann_build.cuh
```diff
@@ -270,18 +269,6 @@ index<T, IdxT> build(
     // Prefetch next batch
     dataset_vec_batches.prefetch_next_batch();
```
rmaschal: Can you move this before the if-statement with the pq-code unpacking? The quantized residuals are currently stored in pageable host memory, meaning the above raft::copy are synchronous. Moving them before the prefetching eliminates the benefit of prefetching, and is a pretty big perf hit when dataset is in host (the most common use case). We'll still take a perf hit because the kernels for unpack_codes are no longer overlapped, but I imagine this is pretty small.

cpp/src/neighbors/scann/detail/scann_build.cuh
```diff
@@ -226,9 +199,11 @@ index<T, IdxT> build(
                                      batch_soar_labels_view,
                                      params.soar_lambda);
 
-    // Compute and quantize residuals
-    auto avq_quant = quantize_residuals<T, IdxT, uint32_t>(
-      res, raft::make_const_mdspan(avq_residuals.view()), full_codebook.view(), pq_params);
+    // Compute and quantize residuals using the public PQ API
+    int64_t codes_dim = cuvs::preprocessing::quantize::pq::get_quantized_dim(pq_build_params);
+    auto avq_quant    = raft::make_device_matrix<uint8_t, IdxT>(res, batch.size(), codes_dim);
+    cuvs::preprocessing::quantize::pq::transform(
```
rmaschal, line 207: This will have a negative impact on perf. The impl of process_and_fill_codes_subspaces in scann_quantize lacked batching via dataset_batch _loader compared to the impl in vpq_dataset. Correct use of dataset_batch_loader requires additional synchronization, which you can see in the vpq_impl: https://github.com/rapidsai/cuvs/blob/main/cpp/src/neighbors/detail/vpq_dataset.cuh#L879-L900

ScaNN already handles the batching in this loop, and the additional synchronization eliminates the benefit of overlapping/prefetching that we get below. 

Can we eliminate the batching in the impl of process_and_fill_codes_subspaces in vpq_dataset when the dataset is small enough, to avoid the extra sync?

rmaschal:
I know it's hard to test the recall of a ScaNN index, I apologize for that. I pulled this PR to test. Compared to main, recall looks overall the same, so no concern there, but there is a build time regression due to the two issues I mentioned in the comments.

Your prior PR https://github.com/rapidsai/cuvs/pull/1278 improved PQ training quite a lot for ScaNN, thank you for that :) For larger dim datasets like LAION and OpenAI, it shaves 2-4s from the overall build time.

cpp/src/neighbors/scann/detail/scann_build.cuh
lowener: I didn't look specifically in the context of ScANN, but the new impl is using faster than the previous in both training and transform. The training is still done sequentially in a for loop, but using a balanced kmeans.

cpp/src/neighbors/scann/detail/scann_build.cuh
lowener: I changed the logic of prefetching in PQ so that if the data is already available from device and doesn't require copy the prefetching will be deactivated, and the handle will not go through the extra sync.

rmaschal: Unfortunately this doesn't fix the sync issue. I was advocating for the removal of dataset_batch_loader in the single batch case. It also hides some synchronization: https://github.com/rapidsai/cuvs/blob/9e4bfae397562407da7f7f3d233b34d98e76d5e1/cpp/src/neighbors/detail/ann_utils.cuh#L418

cpp/src/preprocessing/quantize/detail/pq.cuh
tarang-jain: Actually I realized why this is needed. Perhaps this comment can be ignored for the time being.

cpp/src/neighbors/scann/detail/scann_build.cuh
lowener: I added a path in PQ transform that skip the batching on some conditions (n_rows within range and data already on device)

rmaschal: Thanks, LGTM

cpp/tests/neighbors/ann_scann.cuh
```diff
@@ -123,6 +123,48 @@ class scann_test : public ::testing::TestWithParam<scann_inputs> {
...
+    raft::resource::sync_stream(handle_);
+
+    bool all_zeros = true;
+    bool has_nan   = false;
+    // Check all codes are in valid range [0, num_pq_clusters]
+    auto n_vecs_to_check = std::min(ps.num_db_vecs, 50u);
+    for (IdxT i = 0; i < n_vecs_to_check * num_subspaces; i++) {
+      if (quant_res_host.data_handle()[i] != 0) { all_zeros = false; }
+      if (quant_soar_host.data_handle()[i] != 0) { all_zeros = false; }
+      if (std::isnan(quant_res_host.data_handle()[i])) {
```
irina-resh-nvda: std::isnan expects a float-type, I'm not sure what does NaN mean for uint8_t

cjnolet: Can we maybe do a test to see if the type is float or integral and act accordingly? I can imagine a really large uint8 could be a sign of an overflow, though I'm not sure that's the best way to test for it... maybe we just always assume integral types are valid?

lowener: True, the main objective was to check that the results was not complete garbage, but with uint8 checking for NaN is meaningless.
I changed it to a reconstruction test using `pq::inverse_transform` api

Issue #1318, opened 2025-09-05
[FEA] Allow pylibraft handles in Python cuVS API


viclafargue:
Currently the cuVS Python API wraps the resources in a [dedicated Resources object](https://github.com/rapidsai/cuvs/blob/d24564a0a8e0dd6fbd4bcec46b04e19dfc8f2564/python/cuvs/cuvs/common/resources.pyx#L33). It itself uses C functions to manage the lifetime of a RAFT resource. These C functions also serve in other programming languages bindings. But, in the specific case of Python, allowing pylibraft handles to serve as resources across the cuVS Python API would allow more uniformity with the rest of the RAPIDS ecosystem.

PR #1752, opened 2026-01-29
Allow pylibraft Handle to be used as resource for cuVS

viclafargue:
Closes https://github.com/rapidsai/cuvs/issues/1318


cjnolet:
Hey @viclafargue thank you for opening a PR for this and I think this is almost ready. We very intetionally wrapped the raft resources into the cuvs namespace so that the user is dealing only with the cuvs resources. Please see the way we did this for the current cuvs device resources objects. This should be done all th way through the C and C++ layers. It will be confusing to users to have to remember which namespace which resource types are coming from and thast's the reason we opted to just wrap them into the cuvs namespace. The other reason for this is that it's coding defensively- should there ever be a need to add, change, or fix a bug on the cuvs side when we're not readily able to fix it in the raft side yet, this affords us the additional layer of indirection to do so.

PR #1755, opened 2026-01-29
Add filter for cagra::merge 

benfred:
Add an optional filter argument to cagra::merge, so that we can compact the index to remove deleted items when merging.

Also remove the 'LOGICAL_MERGE' strategy from the `cagra::merge` api, since this is only supported with the composite merge api. (The `cagra::merge` function ignore this value if set, and only does a logical merge).


benfred:
I accidentally merged #1496 into main - when it should have been in release/26.02.  This PR is cherry-picking the commit into the 26.02 release to fix that mistake

bdice (approved):

PR #1759, opened 2026-02-01
Rename lanczos

aamijar:
Follow up to https://github.com/rapidsai/raft/pull/2936. This PR is required since we renamed the function on the raft side (to better reflect its functionality).


cpp/cmake/thirdparty/get_raft.cmake
```diff
@@ -60,8 +60,8 @@ endfunction()
 # To use a different RAFT locally, set the CMake variable
 # CPM_raft_SOURCE=/path/to/local/raft
 find_and_configure_raft(VERSION  ${RAFT_VERSION}.00
-        FORK                     ${RAFT_FORK}
-        PINNED_TAG               ${RAFT_PINNED_TAG}
+        FORK                     aamijar
+        PINNED_TAG               rename-lanczos
```
aamijar: Undo later

lowener (approved):

mythrocks:
@aamijar: Could we please consider updating the description for the PR with why this change is required?  It would help make the git log legible.

PR #1760, opened 2026-02-02
remove pip.conf migration code in CI scripts, update CI-skipping rules

jameslamb:
## Description

Follow-up to changes for https://github.com/rapidsai/build-planning/issues/241

* removes `rapids-init-pip` feature flags now that those have become the default behavior (https://github.com/rapidsai/gha-tools/pull/242)

Follow-up changes for https://github.com/rapidsai/build-planning/issues/243

* a few more `changed-files` modifications
  - *wheel tests shouldn't run on PRs that only change files in `conda/`*
  - *test jobs don't need to run on PRs that only change Doxygen files*
  - *docs builds shouldn't need to run on PRs that only change wheel-related files*


jameslamb:
The first CI run failed when 1 job timed out after 6 hours:

```text
wheel-tests-cuvs / 12.9.1, 3.11, amd64, ubuntu22.04, l4, latest-driver, oldest-deps
```

I just cancelled a second attempt of that job after 4 hours:

https://github.com/rapidsai/cuvs/actions/runs/21615292316/job/62360229013?pr=1760

Other configurations take only around 12-15 minutes to run, so I strongly suspect there's something wrong there. It's also very likely not related to the changes in this PR.

Next I'll try a full rebuild and re-run of everything triggered by an empty commit. Maybe pulling in newer versions of dependencies at build time will help.

PR #1764, opened 2026-02-03
[FEA] View Type PQ Preprocessor

tarang-jain:
Separate VPQ codebooks from encoded data by introducing `vpq_codebooks<MathT>` (PIMPL with owning/view variants). `vpq_dataset` is now a simple `{codebooks, data}` struct. `quantizer<T>` holds only `vpq_codebooks` since it doesn't need encoded data. Adds a view-type `pq::build()` overload for pre-computed external codebooks.


cpp/include/cuvs/preprocessing/quantize/pq.hpp
```diff
@@ -117,6 +119,42 @@ quantizer<float> build(raft::resources const& res,
...
+ *   where pq_n_centers = (1 << pq_bits), pq_len = dim / pq_dim
+ * @param[in] vq_centers VQ codebook on device memory [vq_n_centers, dim].
+ *   Pass an empty view if use_vq=false.
+ *
+ * @return A view-type quantizer that references the provided data
+ */
+quantizer<float> build(raft::resources const& res,
+                       const params params,
+                       raft::device_matrix_view<const float, uint32_t, raft::row_major> pq_centers,
+                       raft::device_matrix_view<const float, uint32_t, raft::row_major> vq_centers);
```
lowener: In the public API I think it's better to use std::optional instead of empty views.
To be consistent with the rest of the cuVS API.

cpp/src/preprocessing/quantize/detail/pq.cuh
```diff
@@ -178,8 +231,10 @@ void transform(
                "Output matrix doesn't have the correct number of columns");
   RAFT_EXPECTS(quantizer.params_quantizer.pq_bits >= 4 && quantizer.params_quantizer.pq_bits <= 16,
                "PQ bits must be within [4, 16]");
-  // Encode dataset
-  auto vq_centers     = raft::make_const_mdspan(quantizer.vpq_codebooks.vq_code_book.view());
+
+  // Use view accessors from vpq_dataset
```
lowener: Remove that comment

cpp/src/preprocessing/quantize/detail/pq.cuh
```diff
@@ -314,18 +369,20 @@ void inverse_transform(
...
-    res,
-    quant.params_quantizer,
-    codes,
-    raft::make_const_mdspan(quant.vpq_codebooks.pq_code_book.view()),
-    raft::make_const_mdspan(quant.vpq_codebooks.vq_code_book.view()),
-    vq_labels,
-    out,
-    quant.params_quantizer.use_subspaces);
+
+  // Use view accessors from vpq_dataset
```
lowener: Remove comment

cpp/include/cuvs/neighbors/common.hpp
```diff
@@ -394,89 +394,126 @@ auto make_aligned_dataset(const raft::resources& res, SrcT src, uint32_t align_b
...
+ * This class wraps the internal implementation (vpq_dataset_owning or vpq_dataset_view)
+ * and provides a stable API.
+ *
  * @tparam MathT the type of elements in the codebooks
  * @tparam IdxT type of the vector indices (represent dataset.extent(0))
- *
  */
 template <typename MathT, typename IdxT>
-struct vpq_dataset : public dataset<IdxT> {
+class vpq_dataset : public dataset<IdxT> {
```
lowener: Just noticed that vpq_dataset doesn't appear in the generated doc because it is not in a doxygen group.
Can you add it to the PQ doc with the directive `@addtogroup pq`

cpp/src/neighbors/vpq_dataset_impl.hpp
```diff
@@ -0,0 +1,211 @@
...
+
+#pragma once
+
+#include <cuvs/neighbors/common.hpp>
+
+#include <raft/util/integer_utils.hpp>
+
+#ifdef __cpp_lib_bitops
+#include <bit>
+#endif
```
lowener: Remove it from `neighbors/common.hpp` since it won't be used there anymore

cpp/include/cuvs/neighbors/common.hpp
lowener: And what's your opinion on moving the whole vpq_dataset structs from `cuvs::neighbors` to `cuvs::preprocessing::quantize::pq`.
I think it would fit well there

lowener (changes requested):

cpp/include/cuvs/preprocessing/quantize/pq.hpp
tarang-jain: This particular factory API that you are pointing out here is intended to be used only for the view type PQ preprocessor. We do have the other (original) build factory for regular (owning) type PQ preprocessor.

cpp/include/cuvs/neighbors/common.hpp
tarang-jain: I have created a new header with the namespace `cuvs::preprocessing::quantize::pq`

cpp/include/cuvs/preprocessing/quantize/pq.hpp
lowener: I understand that but in the comment just above you wrote "Pass an empty view if use_vq=false" for `vq_centers`. So I think that `vq_centers` should be an optional.

tarang-jain: Good catch! I have made it optional and also updated the docs.

cpp/include/cuvs/neighbors/common.hpp
tarang-jain: Also added to the PQ doc group.

cpp/src/preprocessing/quantize/detail/vpq_dataset_impl.hpp
```diff
@@ -0,0 +1,207 @@
...
+};
+
+/**
+ * @brief Owning VPQ dataset implementation - owns the codebooks and data.
+ *
+ * @tparam MathT the type of elements in the codebooks
+ * @tparam IdxT type of the vector indices
+ */
+template <typename MathT, typename IdxT>
+class vpq_dataset_owning : public vpq_dataset_impl<MathT, IdxT> {
```
lowener: Add non-const accessor for this owning struct

cpp/src/neighbors/detail/vpq_dataset.cuh
```diff
@@ -504,22 +505,34 @@ void process_and_fill_codes(
...
+                    src.pq_code_book());
+
+  // Copy the data from the source (data type is uint8_t, independent of MathT)
+  auto data_view = src.data();
+  auto data      = raft::make_device_matrix<uint8_t, IdxT, raft::row_major>(
+    res, data_view.extent(0), data_view.extent(1));
+  raft::copy(data.data_handle(),
+             data_view.data_handle(),
+             data_view.size(),
+             raft::resource::get_cuda_stream(res));
```
lowener: Can we avoid this copy by making `detail::vpq_build` return the 3 arrays independently from the vpq_dataset?

lowener: Or an alternative would be to move the call to`process_and_fill_codes` (in `detail::vpq_build`) upstream so that we write it directly to the correct output buffer.

cpp/src/preprocessing/quantize/detail/vpq_dataset_impl.hpp
tarang-jain: I think that is what we are trying to avoid. If someone really wants to modify the codebooks, they can do a <const_cast>. But the encouraged pathway is that if someone wants to modify the codebooks, they should do so externally and then use the view-type factory with the new codebooks.

tarang-jain: I did the same thing in ivf_pq.hpp

lowener: `ivf_pq::owning_impl` has non-const accessor for pq_centers, centers, etc.
It is currently impossible to modify the codebooks once built, even externally because it owns the mdarray privately and provide no way to modify it.

cpp/src/preprocessing/quantize/detail/pq.cuh
```diff
@@ -155,8 +158,63 @@ quantizer<MathT> build(
...
   return {filled_params,
-          cuvs::neighbors::vpq_dataset<MathT, int64_t>{
-            std::move(vq_code_book), std::move(pq_code_book), std::move(empty_codes)}};
+          cuvs::preprocessing::quantize::pq::vpq_dataset<MathT, int64_t>{
+            std::make_unique<cuvs::neighbors::vpq_dataset_owning<MathT, int64_t>>(
+              std::move(vq_code_book), std::move(pq_code_book), std::move(empty_codes))}};
+}
+
+/**
+ * @brief Build a view-type quantizer from pre-computed codebooks.
```
tfeher:
```suggestion
 * @brief Build a quantizer from pre-computed codebooks.
 * 
 * The quantizer does not own the codebook arrays
 
```

cpp/include/cuvs/preprocessing/quantize/pq.hpp
```diff
@@ -117,6 +120,48 @@ quantizer<float> build(raft::resources const& res,
                        const params params,
                        raft::host_matrix_view<const float, int64_t> dataset);
 
+/**
+ * @brief Creates a view-type product quantizer from pre-computed codebooks.
```
tfeher:
```suggestion
 * @brief Creates a product quantizer from pre-computed codebooks.
```

cpp/include/cuvs/neighbors/common.hpp
```diff
@@ -393,100 +393,6 @@ auto make_aligned_dataset(const raft::resources& res, SrcT src, uint32_t align_b
...
- *
- *   1. Vector Quantization
- *   2. Product Quantization of residuals
- *
- * @tparam MathT the type of elements in the codebooks
- * @tparam IdxT type of the vector indices (represent dataset.extent(0))
- *
- */
-template <typename MathT, typename IdxT>
-struct vpq_dataset : public dataset<IdxT> {
```
tfeher, line 409: This is moved to `vpq_dataset.hpp` and that is not included by common.hpp. I would consider this as a breaking change and update labels accordingly (or add an include above).

cpp/src/neighbors/detail/vpq_dataset.cuh
tfeher: `src.data()` is the whole dataset. Although it is already compressed, it can still be HUGE in size. We definitely do not want to copy it. Note that the function takes `src` as rvalue reference, so we can just move `src.data()` to the new `vpq_dataset` and take ownership of it.

tfeher (changes requested):
Hi Tarang, thank you for the PR! Overall looks good, but there is an issue with a copy operation that should be fixed before we can merge this.

Nitpick, but I find 'View Type PQ preprocessor' expression confusing. Could you please update the PR description to explain that this PR provides an owning and non owning variants of the vpq_dataset?

cpp/src/neighbors/detail/vpq_dataset.cuh
tarang-jain: std::move might give us problems with cagra build because it tries to use the same dataset matrix afterwards.

cpp/include/cuvs/neighbors/common.hpp
tarang-jain: Updated the PR label.

tarang-jain:
@tfeher I have updated the PR desc

cpp/src/preprocessing/quantize/detail/vpq_dataset_impl.hpp
tarang-jain: The ones in ivf_pq::owning_impl probably need to be removed then. The idea is to not allow modification once the codebooks are built. If one really wants to, they can do a `const_cast`. Or they do that externally and use the view type API to build a new `vpq_dataset` object.

cpp/src/neighbors/detail/vpq_dataset.cuh
tarang-jain: I got rid of the copy. Separated out the vpq_codebooks into a separate class. Will see if this approach works.

cpp/src/preprocessing/quantize/detail/vpq_dataset_impl.hpp
lowener: The ivf_pq::owning_impl is allocating memory, but it would offer no way to write to it? It doesn't make sense a lot of sense to me.

The ivf_pq::owning_impl is important: used [here](https://github.com/rapidsai/cuvs/blob/main/cpp/src/neighbors/ivf_pq/ivf_pq_build.cuh#L306-L316) it allows to build the pq centers directly from a transpose operation. Without being able to write to that allocated space, a temporary buffer would be necessary with an extra copy.

I agree that we want to prevent modification **after** the codebooks are built, so after the call to build() is done. I think as a general rule:
- Within the scope of build() we should still be able to write that data.
- In any other scope (`search()` for ivf_pq, `transform()`, `inverse_transform()` for pq) we should only manipulate "view-types".
- Outside of cuvs, users don't have even access to owning_impl since it's not in the public API.

Removing non-const accessor for owning_impl is making life harder for cuvs devs more than anything.

For vpq_dataset_owning it's a bit different because it accepts mdarrays in the constructor to std::move, but I don't see why we would want to remove that non-const getter

tarang-jain: I just added non const getters to the owning implementation.

cpp/tests/neighbors/ann_scann.cuh
```diff
@@ -182,12 +183,11 @@ class scann_test : public ::testing::TestWithParam<scann_inputs> {
...
       vq_codebook.data_handle(), idx.centers().data_handle(), idx.centers().size(), stream_);
-    auto empty_data = raft::make_device_matrix<uint8_t, int64_t, raft::row_major>(handle_, 0, 0);
-
     cuvs::preprocessing::quantize::pq::quantizer<float> quantizer{
       pq_params,
-      cuvs::neighbors::vpq_dataset<float, int64_t>{
-        std::move(vq_codebook), std::move(pq_codebook_copy), std::move(empty_data)}};
+      cuvs::preprocessing::quantize::pq::vpq_codebooks<float>{
+        std::make_unique<cuvs::preprocessing::quantize::pq::vpq_codebooks_owning<float>>(
+          std::move(vq_codebook), std::move(pq_codebook_copy))}};
```
lowener, line 190: Use the new pq::build() with views that you're introducing in this PR. That will also get rid of the relative include of src

cpp/src/neighbors/detail/vamana/vamana_build.cuh
```diff
@@ -646,10 +647,10 @@ index<T, IdxT> build(
...
     auto quantizer = cuvs::preprocessing::quantize::pq::quantizer<float>(
       pq_params,
-      cuvs::neighbors::vpq_dataset<float, int64_t>{
-        raft::make_device_matrix<float, uint32_t, raft::row_major>(res, 0, 0),
-        std::move(pq_codebook),
-        raft::make_device_matrix<uint8_t, int64_t, raft::row_major>(res, 0, 0)});
+      cuvs::preprocessing::quantize::pq::vpq_codebooks<float>{
+        std::make_unique<cuvs::preprocessing::quantize::pq::vpq_codebooks_owning<float>>(
+          raft::make_device_matrix<float, uint32_t, raft::row_major>(res, 0, 0),
+          std::move(pq_codebook))});
```
lowener, line 653: Use the new pq::build() with views

cpp/src/preprocessing/quantize/detail/vpq_dataset_impl.hpp
```diff
@@ -0,0 +1,92 @@
...
+
+namespace cuvs::preprocessing::quantize::pq {
+
+template <typename MathT>
+class vpq_codebooks_owning : public vpq_codebooks_iface<MathT> {
+ public:
+  using math_type = MathT;
+
+  vpq_codebooks_owning(raft::device_matrix<math_type, uint32_t, raft::row_major>&& vq_code_book,
+                       raft::device_matrix<math_type, uint32_t, raft::row_major>&& pq_code_book)
```
lowener, line 18: Nitpick: Would it make sense to have the vq as optional in both struct? It would follow the other functions parameter order too if it's by default nullopt.

lowener (commented):
Looks very good, thanks! Just have a remark on `build()`

cpp/src/preprocessing/quantize/detail/vpq_dataset_impl.hpp
tarang-jain: Good idea! Made it optional everywhere.

cpp/include/cuvs/preprocessing/quantize/vpq_dataset.hpp
```diff
@@ -101,7 +118,15 @@ class vpq_codebooks {
     return impl_->pq_code_book();
   }
 
-  [[nodiscard]] auto dim() const noexcept -> uint32_t { return impl_->dim(); }
+  [[nodiscard]] auto dim() const noexcept -> uint32_t
+  {
+    if (!impl_->vq_code_book().has_value()) {
+      RAFT_LOG_WARN(
+        "vpq_codebooks::dim() returns 0 when no VQ codebook is present; "
+        "the original vector dimension cannot be recovered from PQ codebooks alone.");
```
lowener: It could be computed by `pq_dim * pq_len`: https://github.com/rapidsai/cuvs/pull/1764/changes/cc2a9005587a340de1018771dd9f66d4282f2498..901c8bbf92fd7af71e65f86709d63f3f3f3d6325#r3170234803.
It might be better to do that and have a value that makes sense rather than returning 0.

tarang-jain: I thought of that. But pq_dim * pq_len could be different from the dim right? I mean we are allowed to have a `pq_dim` that does not divide `dim`. I have to double check this however. It is possible that dim() is actually the rotated dim, that is divisible by `pq_dim`.

tarang-jain: yes confirmed that pq_dim * pq_len wont always equal to dim (in cases when it is not divisible). That product is the padded rotated dim.

lowener (approved):

tarang-jain:
Since the new dataset API is moving towards templates instead of a base dataset class and inheritance, we might want to use the same pattern here to store the codebooks. The PIMPL might not be needed and this PR most probably needs to be refactored to create containers for the codebooks.

So we can either merge this now and undo the PIMPL in https://github.com/rapidsai/cuvs/pull/1846, or wait for https://github.com/rapidsai/cuvs/pull/1846 to be merged and then create the container here for the codebooks.

cpp/bench/ann/src/cuvs/cuvs_cagra_wrapper.h
```diff
@@ -17,6 +17,8 @@
 #include <cuvs/neighbors/dynamic_batching.hpp>
 #include <cuvs/neighbors/ivf_pq.hpp>
 #include <cuvs/neighbors/nn_descent.hpp>
+#include <cuvs/preprocessing/quantize/pq.hpp>
+#include <cuvs/preprocessing/quantize/vpq_dataset.hpp>
```
cjnolet, line 21: This isn't RAFT and the library isn't header-only. Items declared in the headers are much more light weight than in RAFT. I'd like to avoid having to define a header per class/abstraction. It gets confusing to users when they have to explicitly include a thousand headers to use common APIs. 

What's the main intention of the "dataset" here? Is this meant to be used internally or directly by users Are users interacting with it directly? If header sizes are what are getting out of control, I'd much rather break the implementation into the compiled source files and keep the prototype/definition in the header files and then instantiate for the types we support, rather than break each one into separate headers that become nebulus to the users.

cpp/include/cuvs/neighbors/common.hpp
```diff
@@ -393,100 +393,6 @@ auto make_aligned_dataset(const raft::resources& res, SrcT src, uint32_t align_b
     raft::round_up_safe<size_t>(src.extent(1) * kSize, std::lcm(align_bytes, kSize)) / kSize;
   return make_strided_dataset(res, std::forward<SrcT>(src), required_stride);
 }
-/**
- * @brief VPQ compressed dataset.
```
cjnolet, line 397: This looks an awful lot like what @HowardHuang1 and @divyegala have been working on. I think we should consolidate the two- we only need 1 way to represent pq compressed datasets. Ideally it would be done with composition in mind over inheritance, and we could use templates like "container policy" as we've done in other APIs (like the sparse apis) to make the apis extensible/pluggable.

tarang-jain:
Closing because this is now incorporated in the new Datasets API

PR #1768, opened 2026-02-04
[REVIEW] Improve 1-NN performance with split GEMM/reduction kernels on Blackwell

vinaydes:
cuVS currently implements 1-nearest neighbor using a fused-kernel approach, where the pairwise-distance GEMM and the subsequent reduction are combined into a single kernel. While this can be efficient, the fused implementation has limitations that prevent it from consistently achieving the best performance. Additionally, the separate-kernels implementation can be used for `half` and `int8` datatypes unlike the fused implementation which is restricted to `float` only.

This PR adds a separate-kernel path in which GEMM and reduction run as two distinct kernels. On Blackwell, the separate-kernel approach performs better for certain M, N, and K configurations (see results below).

In addition, this PR includes:
- A simple heuristic to choose between the fused and separate paths
- Unit tests covering both fused and separate execution paths
- A benchmark that compares fused vs. separate performance and also reports GEMM-only time for reference

**End-to-end benchmarks**

I ran the CUVS_IVF_PQ_ANN_BENCH on a dataset with 10 million vectors and with following build parameters
```
"build_param": {
        "pq_dim": 128,
        "pq_bits": 8,
        "nlist": 10000,
        "niter": 10,
        "ratio": 100
      },
```
Observed following performance improvements:

Fused:
image

Seperate:
image

**1-NN computing benchmark:**

Following table shows the performance of fused and separate computation of 1-NN for various sizes of M, N and K. The GEMM column shows the performance of pure GEMM for comparison. Higher is better here.

|   M   |   N  |  K  | Fused TFLOPS | Separate TFLOPS |  GEMM TFLOPS |
|:-----:|:----:|:---:|:-----:|:--------:|:-----:|
| 16384 | 4096 | 128 | 28.28 |   37.14  | 44.53 |
| 16384 | 4096 |  64 | 22.04 |   25.86  | 33.66 |
|  8192 | 2048 | 128 | 26.43 |   35.33  | 40.96 |
|  8192 | 2048 |  64 | 18.57 |   24.77  | 30.54 |


cpp/src/cluster/detail/kmeans_balanced.cuh
```diff
@@ -51,6 +53,27 @@ namespace cuvs::cluster::kmeans::detail {
...
 
+/**
+ * @brief Returns true is fused implementation should used.
+ */
+template <typename MathT, typename IdxT, typename LabelT>
+bool use_fused(const raft::resources& handle, IdxT m, IdxT n, IdxT k)
+{
+  cudaDeviceProp prop;
+  prop = raft::resource::get_device_properties(handle);
+  if (prop.major <= 8) {
```
tarang-jain: maybe we can use conditional compilation here with `__CUDA_ARCH__` ? For example: https://github.com/rapidsai/cuvs/blob/205c8144967b8350b98d1ed47fbc13bf74f47700/cpp/src/distance/detail/fused_distance_nn/simt_kernel.cuh#L75

cpp/src/distance/unfused_distance_nn.cuh
```diff
@@ -0,0 +1,296 @@
...
+                            cudaStream_t stream)
+{
+  cudaDataType_t xyType, zType;
+  cublasComputeType_t computeType;
+
+  const half h_alpha = 1.0, h_beta = 0.0;
+  const float f_alpha = 1.0, f_beta = 0.0;
+  const double d_alpha = 1.0, d_beta = 0.0;
+  const int32_t i32_alpha = 1, i32_beta = 0;
+  const int8_t i8_alpha = 1, i8_beta = 0;
```
tarang-jain: we can probably define these variables inside the if constexpr braces below rather than declaring all of the types upfront.

cpp/src/distance/unfused_distance_nn.cuh
```diff
@@ -0,0 +1,296 @@
...
+    handle, (AccT*)workspace, x, y, m, n, k, xn, yn, stream);
+
+  ASSERT((metric == DistanceType::CosineExpanded) || (metric == DistanceType::L2Expanded) ||
+           (metric == DistanceType::L2SqrtExpanded),
+         "Only cosine and L2 distance types are supported");
+
+  if (metric == DistanceType::L2Expanded) {
+    reduce_min<DataT, AccT, OutT, IdxT, DistanceType::L2Expanded>(
+      min, (AccT*)workspace, xn, yn, m, n, stream, is_sqrt, initOutBuffer);
+  } else if (metric == DistanceType::L2SqrtExpanded) {
```
tarang-jain, line 307: It appears reduce_min_kernel is doing the same thing for L2Expanded and L2SqrtExpanded. So we can maybe combine these two into a single condition

cpp/src/cluster/detail/kmeans_balanced.cuh
vinaydes: This is in the host code, so `__CUDA_ARCH__` is not available here. I measured the penalty of this run-time check and found it to be negligible.

cpp/src/distance/unfused_distance_nn.cuh
vinaydes: That would not be possible as pointer to correct `alpha` and `beta` are required beyond the `if-else` block. A way around this could to repeat the GEMM API call in each `if-else` block. I thought that would be an even uglier solution over this.

cpp/src/distance/unfused_distance_nn.cuh
vinaydes: I've no clue why two separate distance types `L2Expanded` and `L2SqrtExpanded` exists along with `is_sqrt` argument. IMO we should drop `is_sqrt` in future and then it would make sense to have two separate template parameters to the `reduce_min` kernel. I haven't done a full analysis on removing `is_sqrt` so keeping it as it is for now to keep parity with fused API.

cpp/src/distance/unfused_distance_nn.cuh
tarang-jain: Oh I see! `f_alpha`, `f_beta` etc need to be in scope so that `alpha` and `beta` are not dangling ptrs. How about this:
`const AccT alpha_val = 1, beta_val = 0;`

Simply using `AccT` should work, right?

cjnolet:
Thanks for this change @vinaydes. There's 2 things we'll need to cover in a bit more depth:

1. Are M and N tile sizes and K number of dims? 16k and 4096 are not particularly realistic nor representative for the types of data we encounter in practice- we're now seeing rows well into the millions and columns into the thousands (1536 and 2048 are becoming more and more common). We'll definitely need to demonstrate what this looks like in an algorithmn like kmeans. 

2. We're experiencing a lot of challenges maintaining a reasonable binary size, especially as we keep adding new kernels, which get compiled various supported architectures. Can you pleaee verify the impact to the binary size in your changes here? 


Overall, these are welcome changes, and I think these help address some of the perf gaps we're seeing on Blackwell!

cpp/src/distance/unfused_distance_nn.cuh
vinaydes: > Simply using `AccT` should work, right?

Technically no, because the constant type depends not only on accumulator type but also on compute type. See https://docs.nvidia.com/cuda/cublas/index.html#cublasgemmex for details.

However, in the case here the only conbination where accumulator type is not constant type is for A, B and C type of `half` and compute type of `CUBLAS_COMPUTE_32F`. I'll add an assert for detecting this case and for all other combinations, just use AccT for declaring the constants.

I also noticed the constant type for `int8_t` are passed incorrectly, I'll also fix that in the next push

cpp/tests/neighbors/distance_nn.cu
```diff
@@ -0,0 +1,215 @@
...
+
+    if constexpr (impl == ImplType::fused) {
+      workspace_size = n * sizeof(IdxT);
+    } else if constexpr (impl == ImplType::unfused) {
+      workspace_size = m * n * sizeof(AccT);
+    }
+
+    // Reset buffers
+    RAFT_CUDA_TRY(cudaMemsetAsync(out.data_handle(), 0, m * sizeof(OutT)));
+    RAFT_CUDA_TRY(cudaStreamSynchronize(stream));
```
tarang-jain: Lets use `raft::matrixx::fill` and `raft::resource::sync_stream(handle)` rather than direct cuda API calls.

cpp/src/distance/unfused_distance_nn.cuh
```diff
@@ -0,0 +1,296 @@
...
+  const void* alpha = nullptr;
+  const void* beta  = nullptr;
+  if constexpr (std::is_same_v<DataT, float>) {
+    xyType      = CUDA_R_32F;
+    zType       = CUDA_R_32F;
+    computeType = CUBLAS_COMPUTE_32F;
+    // computeType = CUBLAS_COMPUTE_32F_FAST_TF32;
+    // computeType = CUBLAS_COMPUTE_32F_FAST_16F;
+    // computeType = CUBLAS_COMPUTE_32F_FAST_16BF;
+    // computeType = CUBLAS_COMPUTE_32F_EMULATED_16BFX9;
```
tarang-jain: these comments need to be removed?

cpp/src/distance/unfused_distance_nn.cuh
vinaydes: Should be fixed now.

cpp/src/distance/unfused_distance_nn.cuh
vinaydes: Done.

cpp/tests/neighbors/distance_nn.cu
vinaydes: I've replaced the stream sync call. However the matrix fill does not seem to handle vector of key-value pair. Keeping memset call for now. Feel free to demonstrate how can I use the fill call.

vinaydes:
Hi @cjnolet,
> 1.  Are M and N tile sizes and K number of dims? 16k and 4096 are not particularly realistic nor representative for the types of data we encounter in practice- we're now seeing rows well into the millions and columns into the thousands (1536 and 2048 are becoming more and more common). We'll definitely need to demonstrate what this looks like in an algorithmn like kmeans.

The CUVS_IVF_PQ_ANN_BENCH benchmark I ran for the end-to-end results calls kmeans internally. Large M and N don’t significantly affect performance because they can be processed in batches, and cuVS kmeans already does this (see batching here: https://github.com/rapidsai/cuvs/blob/main/cpp/src/cluster/detail/kmeans_balanced.cuh#L434). The performance-critical parameter is K.

For sufficiently large K, GEMM time dominates, so the reduction time (whether fused or separate) becomes negligible. In that regime, using a separate reduction kernel is the straightforward choice because we can rely on cuBLAS for near-optimal GEMM performance. This is applicable for Blackwell or any future architecture.

The challenging regime is small K. When K is small enough, GEMM is largely bandwidth-bound (mostly writing the output matrix), so fusing the reduction with GEMM tends to make the most sense. In this exercise, we aimed to identify what “large” vs. “small” K means on Blackwell.

Empirically:
1. For K >= 512, the reduction kernel is a negligible fraction of total runtime. So use separate kernel
2. For K <= 128, there is room to develop Blackwell-specific fused kernels. The current fused kernels were designed for Ampere and perform worse than the separate-kernel approach introduced in this PR. Therefore using separate kernel is a better choice here as well

Also @tfeher can back me on this, but we currently aren't interested in K < 96.

I'll get back to you with impact of this PR on binary size.

vinaydes:
@cjnolet 
> 2. We're experiencing a lot of challenges maintaining a reasonable binary size, especially as we keep adding new kernels, which get compiled various supported architectures. Can you pleaee verify the impact to the binary size in your changes here?

I built the libcuvs with ` -DCMAKE_CUDA_ARCHITECTURES=RAPIDS` CMake option. I get following sizes for cuVS lib:

main - 324028968 B (~310M this is inline with nightly build size of `libcuvs.so`)
PR - 324433552 B 

That is ~395 KB of increase in the binary size or 0.13% increase.

cpp/src/cluster/detail/kmeans_balanced.cuh
```diff
@@ -105,22 +131,43 @@ inline std::enable_if_t<std::is_floating_point_v<MathT>> predict_core(
...
+          minClusterAndDistance.data_handle(),
+          dataset,
+          centers,
+          dataset_norm,
+          centroidsNorm.data_handle(),
+          n_rows,
+          n_clusters,
+          dim,
+          (void*)workspace.data_handle(),
+          (params.metric == cuvs::distance::DistanceType::L2Expanded) ? true : false,
```
dantegd:
```suggestion
          (params.metric == cuvs::distance::DistanceType::L2Expanded) ? false : true,
```

This is flipped from the original code, unless I'm mistaken this means we are applying (or not applying) `sqrt` for the wrong case.

cpp/tests/neighbors/distance_nn.cu
```diff
@@ -0,0 +1,242 @@
...
+    } else {
+      raft::random::uniform(handle, rng, x.data_handle(), m * k, DataT(-1.0), DataT(1.0));
+      raft::random::uniform(handle, rng, y.data_handle(), n * k, DataT(-1.0), DataT(1.0));
+    }
+
+    // Pre-compute norms
+    raft::linalg::rowNorm<raft::linalg::L2Norm, true>(
+      x_norm.data_handle(), x.data_handle(), k, m, stream);
+    raft::linalg::rowNorm<raft::linalg::L2Norm, true>(
+      y_norm.data_handle(), y.data_handle(), k, m, stream);
```
dantegd:
```suggestion
      y_norm.data_handle(), y.data_handle(), k, n, stream);
```

cpp/bench/prims/src/distance/distance_nn.cu
```diff
@@ -0,0 +1,341 @@
...
+  RAFT_CUDA_TRY(cudaMemsetAsync(workspace.data_handle(), 0, workspace_size, stream));
+  RAFT_CUDA_TRY(cudaMemsetAsync(out.data_handle(), 0, m * sizeof(OutT), stream));
+  RAFT_CUDA_TRY(cudaStreamSynchronize(stream));
+
+  cudaEvent_t start, stop;
+  cudaEventCreate(&start);
+  cudaEventCreate(&stop);
+
+  for (auto _ : state) {
+    cudaEventRecord(start, 0);
```
dantegd:
```suggestion
    cudaEventRecord(start, stream);
```

This should be in `stream` and not the default stream, right?

cpp/bench/prims/CMakeLists.txt
```diff
@@ -0,0 +1,118 @@
...
+# * Target function -------------------------------------------------------------
+
+function(ConfigurePrimsBench)
+
+  set(oneValueArgs NAME)
+  set(multiValueArgs PATH LINKS CXXFLAGS)
+
+  #if(NOT BUILD_CPU_ONLY)
+    set(GPU_BUILD ON)
+  #endif()
```
dantegd: Do we need the conditional/commented code?

cpp/tests/neighbors/distance_nn.cu
```diff
@@ -0,0 +1,242 @@
...
+  raft::device_vector<OutT, IdxT> ref_out;
+  size_t workspace_size;
+};
+
+template <typename IdxT>
+const std::vector<NNInputs<IdxT>> input_fp32 = {
+  {4096, 4096, 64, DistanceType::L2Expanded, false, uint64_t(31415926), 0.1},
+  {4096, 4096, 128, DistanceType::L2Expanded, true, uint64_t(31415926), 0.1},
+  {4096, 4096, 64, DistanceType::CosineExpanded, false, uint64_t(31415926), 0.1},
+  {4096, 4096, 128, DistanceType::CosineExpanded, true, uint64_t(31415926), 0.1},
```
dantegd: It might be worth adding a few test cases with m != n

dantegd (changes requested):
logic looks good to me, had a few very minor comments, and seems like the binary size impact is very small

cpp/tests/neighbors/distance_nn.cu
vinaydes: Thanks, good catch.

cpp/src/cluster/detail/kmeans_balanced.cuh
vinaydes: Good catch, thanks.

cpp/bench/prims/src/distance/distance_nn.cu
vinaydes: Yes, although it would not affect benchmark performance measurement as 0 or NULL stream force syncronizes on all other streams. So I don't think rerunning the benchmarks are necessary.

cpp/bench/prims/CMakeLists.txt
vinaydes: Probably not. This whole file is copied from ANN benchmark cmake file. Can someone with CMake experties take a look at it as a whole?

cpp/src/distance/unfused_distance_nn.cuh
```diff
@@ -0,0 +1,281 @@
...
+#pragma once
+
+#include <raft/core/resource/cublas_handle.hpp>
+#include <raft/core/resources.hpp>
+
+namespace cuvs {
+namespace distance {
+
+template <typename T>
+__host__ __device__ T max_val()
```
tarang-jain: Nit: Can we use _RAFT_HOST_DEVICE?

cpp/src/distance/unfused_distance_nn.cuh
```diff
@@ -0,0 +1,281 @@
...
+  }
+}
+
+template <typename AccT, typename IdxT>
+struct Reducer {
+  typedef raft::KeyValuePair<IdxT, AccT> KVType;
+
+  __device__ KVType operator()(const KVType& a, const KVType& b)
+  {
+    if ((a.value < b.value) || (a.value == b.value && a.key < b.key)) {
```
tarang-jain, line 45: Can we not use the comparator `kvp_cg_min_reduce_op` from the fused header here? The tiebreak characteristic (where values are equal) can be added to the same struct, right?

cpp/tests/neighbors/distance_nn.cu
```diff
@@ -0,0 +1,242 @@
...
+      y_norm.data_handle(), y.data_handle(), k, m, stream);
+
+    if constexpr (impl == ImplType::fused) {
+      workspace_size = n * sizeof(IdxT);
+    } else if constexpr (impl == ImplType::unfused) {
+      workspace_size = m * n * sizeof(AccT);
+    }
+
+    // Reset buffers
+    RAFT_CUDA_TRY(cudaMemsetAsync(out.data_handle(), 0, m * sizeof(OutT)));
```
tarang-jain: Another place for using `raft::matrix::fill`.

cpp/bench/prims/CMakeLists.txt
vinaydes: Removed the commented code. Keeping rest of the CMake file as it is not. May be we can revisit refactoring of benchmark files in future.

cpp/src/distance/unfused_distance_nn.cuh
vinaydes: Done.

cpp/tests/neighbors/distance_nn.cu
vinaydes: Figured out how to use `matrix::fill`. Done now.

cpp/src/distance/unfused_distance_nn.cuh
vinaydes: The idea here is to keep the fused and unfused path completely independent. We may remove the fused implementation (at least the current one) completely and it is near exclusive for Ampere at the point.

cpp/tests/neighbors/distance_nn.cu
vinaydes: Done.

dantegd (approved):
PR looks good to me, also I think it's a good idea to keep the benchmark on a new bench/prims directory separate of the ann ones

vinaydes:
@cjnolet @dantegd What is the next step for this PR? I have addressed all the comments. Does it need some kind of blessing for CI to run?

vinaydes:
@aamijar Thanks for running the CI. I am not sure, why I dont see the CI failures locally. May be because I am using 13.2 version of toolkit. I'll try to reprodue the failure locally with 13.1.

cpp/src/distance/unfused_distance_nn.cuh
```diff
@@ -0,0 +1,281 @@
...
+    xyType      = CUDA_R_64F;
+    zType       = CUDA_R_64F;
+    computeType = CUBLAS_COMPUTE_64F;
+  }
+
+  const AccT alpha = static_cast<AccT>(1);
+  const AccT beta  = static_cast<AccT>(0);
+
+  auto cublas_h = raft::resource::get_cublas_handle(handle);
+  RAFT_CUBLAS_TRY(cublasGemmEx(cublas_h,
```
tfeher, line 219: Let's use [raft::linalg::gemm](https://github.com/rapidsai/raft/blob/8d8e1efceacb2c2cd85577be9752485d20af2dd7/cpp/include/raft/linalg/gemm.cuh#L219) if possible. 

The raft API takes care of  stream ordering, and also caches the GEMM descriptors, which is beneficial if we repeat the gemm with same size.

cpp/src/distance/unfused_distance_nn.cuh
```diff
@@ -0,0 +1,281 @@
...
+        out[row] = block_result.value;
+      } else {
+        out[row] = block_result.value < out[row] ? block_result.value : out[row];
+      }
+    }
+  }
+}
+
+template <typename DataT, typename AccT, typename OutT, typename IdxT, DistanceType metric>
+void reduce_min(OutT* out,
```
tfeher, line 146: Please add a docstring that describes that here the distance is calculated by adding the terms with the norms, and the reduction is performed afterwards. Fusing  these two ops in the same kernel is the reason why we are not using the reduction functions from `raft`.

cpp/bench/prims/src/distance/distance_nn.cu
```diff
@@ -0,0 +1,351 @@
...
+  // Calculate the workspace size
+  // for fused it is m * sizeof(IdxT)
+  // for unfused and gemm it is m * n * sizeof(AccT);
+
+  size_t workspace_size = (algo == AlgorithmType::fused) ? m * sizeof(IdxT) : m * n * sizeof(AccT);
+
+  raft::device_vector<char, size_t> workspace =
+    raft::make_device_vector<char, size_t>(handle, workspace_size);
+
+  // Warm up
```
tfeher: gbench would add warm-up iterations if requested, and we can control that using command line arg. I would skip adding an explicit warm-up step here.

cpp/bench/prims/src/distance/distance_nn.cu
```diff
@@ -0,0 +1,351 @@
...
+  const int m = state.range(0);
+  const int n = state.range(1);
+  const int k = state.range(2);
+
+  raft::device_resources handle;
+  rmm::cuda_stream_view stream;
+
+  stream = raft::resource::get_cuda_stream(handle);
+
+  auto x      = raft::make_device_matrix<DataT, IdxT>(handle, m, k);
```
tfeher: Ideally we would want to have setup and teardown functions to run the initializations once per benchmark instead of once per benchmark iteration.

cpp/bench/prims/src/distance/distance_nn.cu
```diff
@@ -0,0 +1,351 @@
...
+    raft::matrix::fill(handle, raft::make_device_matrix_view(out.data_handle(), m, 1), OutT{0});
+  }
+  raft::resource::sync_stream(handle, stream);
+
+  cudaEvent_t start, stop;
+  cudaEventCreate(&start);
+  cudaEventCreate(&stop);
+
+  for (auto _ : state) {
+    cudaEventRecord(start, stream);
```
tfeher: We have a helper for cuda event based benchmark timing: https://github.com/rapidsai/cuvs/blob/51fb88e0cb826605d4f21ee183af4e567efe9507/cpp/bench/ann/src/common/util.hpp#L54

tfeher (changes requested):
Thanks Vinay for the PR, it is great to improve kmeans performance.

I would recommend to move the benchmark code to a separate PR, because there are a few issues there that need to be resolved, and it should not delay merging the improved 1-NN distance functions.

Regarding the new 1-NN distance function, please check if we could use raft's GEMM wrappers. Otherwise the code looks good.

cpp/bench/prims/src/distance/distance_nn.cu
vinaydes: Yeah, we should probably have a common utils file across prims and ann benchmark. I found using ann utils in prims a little bit awkward.

cpp/bench/prims/src/distance/distance_nn.cu
vinaydes: I can't think of how to do it when input, output data types changes for each benchmark. Anyway I am separating benchmark code for now.

cpp/src/distance/unfused_distance_nn.cuh
vinaydes: `raft::linalg::gemm` currently does not have an option for specifying the compute type. I have created a RAFT issue to track this https://github.com/rapidsai/raft/issues/2988

vinaydes:
@tfeher I have deleted the benchmarking code for now. We can reintroduce it after refactoring and addressing changes you suggested. Thanks.

cpp/src/cluster/detail/kmeans_balanced.cuh
```diff
@@ -104,22 +130,43 @@ inline std::enable_if_t<std::is_floating_point_v<MathT>> predict_core(
...
-        dim,
-        (void*)workspace.data_handle(),
-        (params.metric == cuvs::distance::DistanceType::L2Expanded) ? false : true,
-        false,
-        true,
-        params.metric,
-        0.0f,
-        stream);
+      if (should_use_fused) {
+        cuvs::distance::fusedDistanceNNMinReduce<MathT, raft::KeyValuePair<IdxT, MathT>, IdxT>(
```
tarang-jain: Rather than (or in addition to) doing this check here, can we do this in minClusterDistanceCompute.cu?. In that case, even regular kmeans will benefit from this change. I have an open PR https://github.com/rapidsai/cuvs/pull/2001 to call minClusterDistanceCompute from kmeans_balanced.cuh instead of directly instantiating fusedDistanceNNMinReduce.

cpp/src/cluster/detail/kmeans_common.cuh
```diff
@@ -52,6 +53,31 @@
...
+    return true;
+  } else if (prop.major == 9 && (m >= 4096 || n >= 4096)) {
+    // On Hopper if m, n are bigger than 4096, use fused
+    return true;
+  } else if (prop.major >= 10) {
+    // On Blackwell onwards, use unfused
+    return false;
+  }
+  return false;
+}
```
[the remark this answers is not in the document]
vinaydes: >That means every SM 10.x case returns false

This is expected/intented behavior. We will pass `k` because in future the heuristic might be based on it.

cpp/tests/neighbors/distance_nn.cu
```diff
@@ -0,0 +1,253 @@
...
+  {
+  }
+
+ protected:
+  void SetUp() override
+  {
+    raft::random::RngState rng{params_.rng_seed};
+    if constexpr (std::is_same_v<DataT, int8_t>) {
+      fill_int8<<<1000, 256, 0, stream>>>(x.data_handle(), m * k);
+      fill_int8<<<1000, 256, 0, stream>>>(y.data_handle(), n * k);
```
vinaydes: Good catch, fixed this by adding a grid-strided loop in the kernel.

cpp/tests/neighbors/distance_nn.cu
```diff
@@ -0,0 +1,253 @@
...
+        0.0,
+        stream);
+    }
+  }
+
+  void compare()
+  {
+    vector_compare(handle, ref_out.data_handle(), out.data_handle(), m, summary);
+    ASSERT_TRUE(summary.max_diff < params_.tol) << summary;
+  }
```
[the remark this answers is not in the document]
vinaydes: This is fine, `n_misses` can be more than zero for less precise data types like FP16 or INT8. Even if vectors are different only at one place the test should fail. So `max_diff` comparison is good enough.

cpp/src/cluster/detail/minClusterDistanceCompute.cu
```diff
@@ -50,24 +51,50 @@ void minClusterAndDistanceCompute(
...
+          n_samples,
+          n_clusters,
+          n_features,
+          (void*)workspace.data(),
+          metric != cuvs::distance::DistanceType::L2Expanded,
+          false,
+          true,
+          metric,
+          0.0f,
+          stream);
```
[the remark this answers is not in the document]
vinaydes: Fixed this by updating `calc_minibatch_size()`

cpp/src/distance/unfused_distance_nn.cuh
```diff
@@ -0,0 +1,311 @@
...
+      thread_min.key   = col;
+    }
+  }
+
+  typedef cub::BlockReduce<KVType, TPB> BlockReduceT;
+  __shared__ typename BlockReduceT::TempStorage temp_storage;
+  auto block_result = BlockReduceT(temp_storage).Reduce(thread_min, Reducer<AccT, IdxT>{});
+
+  if (threadIdx.x == 0) {
+    if (is_sqrt) { block_result.value = raft::sqrt(block_result.value); }
```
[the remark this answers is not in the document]
vinaydes: Added guards for zero and very small value norms.

cpp/tests/neighbors/distance_nn_helper.cuh
```diff
@@ -0,0 +1,206 @@
...
+  AccT v1v2    = AccT(0.0);
+
+  for (IdxT k = 0; k < K; k++) {
+    v1_norm += (AccT(v1[k]) * AccT(v1[k]));
+    v2_norm += (AccT(v2[k]) * AccT(v2[k]));
+    v1v2 += (AccT(v1[k]) * AccT(v2[k]));
+  }
+
+  return AccT(1.0) - (v1v2 / (v1_norm * v2_norm));
+}
```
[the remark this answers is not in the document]
vinaydes: Fixed this.

cpp/tests/neighbors/distance_nn.cu
```diff
@@ -0,0 +1,253 @@
...
+{
+  int tid = threadIdx.x + blockIdx.x * blockDim.x;
+  // Fill the buffer with pseudo-random int8_t values using a simple LCG
+  for (int i = tid; i < len; i += blockIdx.x * gridDim.x) {
+    // Simple LCG: x_n+1 = (a * x_n + c) % m
+    // Use tid as seed, constants chosen for decent distribution
+    int seed  = tid * 1103515245 + 12345;
+    buff[tid] = static_cast<int8_t>((seed >> 16) & 0xFF);
+  }
+}
```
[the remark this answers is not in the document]
vinaydes: Fixed all these issues.

cpp/tests/neighbors/distance_nn.cu
```diff
@@ -0,0 +1,253 @@
...
+  {
+  }
+
+ protected:
+  void SetUp() override
+  {
+    raft::random::RngState rng{params_.rng_seed};
+    if constexpr (std::is_same_v<DataT, int8_t>) {
+      fill_int8<<<1000, 256, 0, stream>>>(x.data_handle(), m * k);
+      fill_int8<<<1000, 256, 0, stream>>>(y.data_handle(), n * k);
```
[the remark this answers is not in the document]
vinaydes: Done.

cpp/src/cluster/detail/kmeans_balanced.cuh
vinaydes: Moved the fused/unfused branch to `minClusterDistanceCompute.cu`

tfeher (approved):
Thanks Vinay for fixing the issues. The PR looks good to me!

tfeher:
/ok to test [e8d6bd1](https://github.com/rapidsai/cuvs/pull/1768/commits/e8d6bd12070b48a662e066c1e62e59696e593abf)

tfeher:
/ok to test [63e0796](https://github.com/rapidsai/cuvs/pull/1768/commits/63e079605c045b95e96b71ca45021757deff520c)

vinaydes:
I think the failure on RTX Pro 6000 is because of extra memory required by unfused path is unavailable. I'll try to fix it.

vinaydes:
I tested on RTX Pro 6000 BSE and all test seems to be passing with latest merge of `main`. Can we try one more CI run?

tfeher:
/ok to test [4104e5f](https://github.com/rapidsai/cuvs/pull/1768/commits/4104e5f4c8a9107542aeab740d4d14efce4a986d)

vinaydes:
The CI failures don't look relevant to this PR. Not sure, what can be done next here.

PR #1498, opened 2025-11-05
Convert non-type template parameters to runtime parameters in CAGRA search to cut binary size

seunghwak:
This PR converts

MAX_ITOPK & MAX_CANDIDATES in single-CTA search and
MAX_ELEMENTS in multi-CTA search to runtime parameters.

Cut libcuvs.so size (CUDA 13, `build.sh --allgpuarch -n libcuvs`) from 459 MB to 350 MB.

For correctness testing,
ran
```
NEIGHBORS_ANN_CAGRA_FLOAT_UTIN32_TEST
NEIGHBORS_ANN_CAGRA_HALF_UTIN32_TEST
NEIGHBORS_ANN_CAGRA_UINT8_UTIN32_TEST
NEIGHBORS_ANN_CAGRA_INT8_UTIN32_TEST
NEIGHBORS_ANN_CAGRA_HELPER_TEST
NEIGHBORS_ANN_CAGRA_TEST_BUGS
```
To evaluate performance impact,
ran cuvs_bench with batch sizes (10, 100, 1000, 10000) and k (10 and 100) (default options for other parameters) and `deep-image-96-inner`
```
python -m cuvs_bench.run --dataset deep-image-96-inner --algorithms cuvs_cagra --batch_size 10|100|1000|10000 - k 10|100
```
for both single-CTA and multi-CTA searches.

Performance impacts varies based on MAX_ITOPK and MAX_CANDIDATES combinations but performance numbers were roughly comparable (slightly slower in average with the maximum slowdown around 10%).

Let me know if there are other benchmarks I need to run to test performance.

Some performance logs from the original code and the updated code for anyone interested (batch size = 10K, K=100).

Original (main)
```
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
Benchmark                                                                                                             Time             CPU   Iterations        GPU    Latency     Recall end_to_end items_per_second      itopk          k max_iterations  n_queries search_width total_queries
-----------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
cuvs_cagra.graph_degree32.intermediate_graph_degree32.graph_build_algoNN_DESCENT/0/process_time/real_time          3.62 ms         3.62 ms          194   3.61323m   3.62005m   0.484826   0.702289       2.76241M/s        128        100             20        10k            1         1.94M algo="single_cta"
[...]

PR #1771, opened 2026-02-04
Set `cudaFuncAttributeMaxDynamicSharedMemorySize` with thread-safety

mythrocks:
This commit ensures that when the setup runs for the CAGRA search kernel,
the calls to set `cudaFuncAttributeMaxDynamicSharedMemorySize` are thread-safe.

The CUDA-context in a process is shared across all (CPU) threads.

Consider 2 (CPU) threads A and B, attempting to set max-dynamic-shared-mem
sizes to 3KB and 2KB respectively.  Consider the following race condition:
1. Thread A successfully sets 3KB, but then gets switched before
   the CAGRA kernel is called.
2. Thread B sets 2KB, calls the kernel successfully, and exits.
3. Thread A calls its kernel, which attempts to use 3KB of smem. But the
   CUDA context's max-smem is currently set to 2KB.  This causes the kernel
   launch to fail.

The error manifests as listed in https://github.com/rapidsai/cuvs-lucene/issues/93:
```
CUDA error encountered at: file=/.../search_single_cta_kernel-inl.cuh line=2207: call='cudaPeekAtLastError()', Reason=cudaErrorInvalidValue:invalid argument
Obtained 8 stack frames
...
```

This commit adds logic to always launch the search kernel with the smem-size set through `cudaFuncAttributeMaxDynamicSharedMemorySize`.  The calls to `cudaFuncSetAttribute` and the kernel launch are put in a critical section.

Signed-off-by: MithunR <mithunr@nvidia.com>

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -49,7 +49,6 @@ jobs:
       id-token: write
     env:
       GH_TOKEN: ${{ github.token }}
-      SHARED_ACTIONS_REF: 'check-nightly-success'
     steps:
       - name: Get PR Info
         id: get-pr-info


mythrocks:
For the kind attention of @tfeher, @cjnolet.

I've refrained from introducing a new `.cu` / translation unit, and introduced the `std::mutex` as a function static.  Please let me know if object lifetimes are a concern.  By my reading of the code, different instantiations will live separately and not clash.

Note that `select_and_run` is a function template, with a different instantiation per `search_kernel_config` definition.
Note, also, that `cudaFuncSetAttribute` is itself a function template, parametrized on the function type.

The above should imply that the `cudaFuncAttributeMaxDynamicSharedMemorySize` is tracked separately for `search_single_cta_float_uint32`, `search_single_cta_int8_uint32`, etc.

mythrocks:
CI is failing, pending merge for #1759.

In the meantime, ~~I'll move~~ I have moved the other calls that set `cudaFuncAttributeMaxDynamicSharedMemorySize` to use the new utility.

cpp/src/neighbors/detail/smem_utils.cuh
```diff
@@ -0,0 +1,37 @@
...
+ * In the event of concurrent calls, we'd like to accommodate the largest requested size.
+ * @tparam KernelT The type of the kernel.
+ * @param smem_size The size of the dynamic shared memory to be set.
+ * @param kernel The kernel to be set.
+ */
+template <typename KernelT>
+void optionally_set_larger_max_smem_size(uint32_t smem_size, KernelT& kernel)
+{
+  static auto mutex                 = std::mutex{};
+  static auto running_max_smem_size = uint32_t{0};
```
viclafargue: I would maybe make `running_max_smem_size` be a `std::atomic` for more safety.

cpp/src/neighbors/detail/smem_utils.cuh
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * @brief Optionally set the larger max dynamic shared memory size for the kernel.
+ * This is required because `cudaFuncSetAttribute` is not thread-safe.
+ * In the event of concurrent calls, we'd like to accommodate the largest requested size.
+ * @tparam KernelT The type of the kernel.
+ * @param smem_size The size of the dynamic shared memory to be set.
+ * @param kernel The kernel to be set.
+ */
+template <typename KernelT>
```
viclafargue: Should work fine if the kernel type is different for each CUDA kernel instantiation. I am not familiar with this.

achirkin:
I'd suggest we hold off this PR for a little while before finishing the discussion in https://github.com/rapidsai/cuvs-lucene/issues/93 . While the explanation of the bug seems plausible, I'm not 100% convinced this is the real cause; and the extra sync introduced here may be just hiding the other issue by serializing execution.

Even if we confirm this is indeed the case, I'd like to do a little more benchmarking to better understand how this affects high-concurrency scenario of using the persistent CAGRA kernel. In such a scenario any, even the smallest, call to the CUDA api may cause a slowdown, because all calls to CUDA context are serialized (i.e. guarded by mutexes).

cpp/src/neighbors/detail/smem_utils.cuh
mythrocks: I don't _think_ that's necessary.  Per the C++ standard, a function static's initialization is threadsafe.  The updates are within the critical section.

Forgive me if I've missed the point of concern.

mythrocks:
Just an update:  Over on https://github.com/rapidsai/cuvs-lucene/issues/93, we [discussed](https://github.com/rapidsai/cuvs-lucene/issues/93#issuecomment-3856424382) whether #1498 might have addressed the issue.  From my testing, it appears that the race condition remains:  The test still fails without the mutex. :/  I'm repeating my tests to be doubly sure.

mythrocks:
> I though about it and it looks to me like registering a CUDA kernel for launch should be almost instantaneous. Therefore we could have something like this...

@viclafargue: The way I've phrased it with double-checked-locking, the mutex is only acquired in the event that we see a request for a larger smem-max.  The kernel invocation is currently not subject to the mutex _unless_ it needs a larger smem-max.  

I'm not sure I've understood the advantage of moving the kernel invocation into the critical section.  (We'd have to lose the double-checked lock to get it to work.)

Permission to keep the critical section minimal?

mythrocks:
@achirkin:

> Even if we confirm this is indeed the case, I'd like to do a little more benchmarking to better understand how this affects high-concurrency scenario of using the persistent CAGRA kernel. 

ACK.  Appreciated.  Given that this is in the performance path, extra scrutiny and diligence would be prudent.

> In such a scenario any, even the smallest, call to the CUDA api may cause a slowdown, because _all calls to CUDA context are serialized_ (i.e. guarded by mutexes).

Not quite *all* calls to CUDA context:
1. I've only intercepted the case where `MaxDynamicSharedMemorySize` is set.
2. I'm checking whether the high-water-mark is exceeded.  The condition is checked twice.  The mutex is acquired (and `cudaFuncSetAttribute` is called) _only_ if the high-water-mark is breached.  This should mean fewer calls to `cudaFuncSetAttribute`.  

The pathological case would be if we have concurrent threads where each thread successively ramps up `MaxDynamicSharedMemorySize`.  I'll defer to you on the likelihood.

Plus, we might need to examine the effects of ramping up the max-smem-size on occupancy.

viclafargue:
> I'm not sure I've understood the advantage of moving the kernel invocation into the critical section. (We'd have to lose the double-checked lock to get it to work.)

You are most probably right. I am just unsure how important it is to set an optimal `MaxDynamicSharedMemorySize` value (minimal to handle a query) to maximize occupancy and ensuring the main search kernel complete its work rapidly versus avoiding the small critical section containing a few CUDA calls. Since threads are not launched exactly at the same time, it may be unlikely that they enter the critical section at the same time. And as you mentioned, there is a risk that each thread would successively ramps up `MaxDynamicSharedMemorySize` to a value that makes the search kernel run slower.

mythrocks:
> Thank you for working on this. LGTM, apart from small comment.
> 
> The only issue that I can think of is that `cudaFuncAttributeMaxDynamicSharedMemorySize` can at one point be set to values much higher than needed causing performance issues (lower occupancy) as discussed.
> 
> I though about it and it looks to me like registering a CUDA kernel for launch should be almost instantaneous. Therefore we could have something like this :
> 
> ```
> static auto mutex = std::mutex{};
> {
>     auto guard = std::lock_guard<std::mutex>{mutex};
>     cudaFuncSetAttribute(kernel, cudaFuncAttributeMaxDynamicSharedMemorySize, smem_size);
>     kernel<<<block_dims, thread_dims, smem_size, stream>>>(...);
> }
> ```

@viclafargue:  It took @jamxia155's independent suggestion before I finally grasped what you were suggesting.  Sorry I didn't catch it the first time.

I've changed the helper function to also call the kernel.  I couldn't find a clean way to accommodate how the kernel is launched in the "persistent runner" case (via `cudaLaunchCooperativeKernel`.  I've had to resort to using a lambda for the launcher.  I hope that is acceptable.

mythrocks:
/ok to test b1b97cc
(Drat.  Forgot to sign a commit.)

cpp/src/neighbors/detail/smem_utils.cuh
```diff
@@ -0,0 +1,37 @@
...
+ * @tparam KernelT The type of the kernel.
+ * @tparam InvocationT The type of the invocation function.
+ * @param kernel The kernel function address (for whom the smem-size is specified).
+ * @param smem_size The size of the dynamic shared memory to be set.
+ * @param invoke_kernel The kernel invocation function/lambda.
+ */
+template <typename KernelT, typename InvocationT>
+void safely_invoke_kernel_with_smem_size(KernelT& kernel,
+                                         uint32_t smem_size,
+                                         InvocationT const& invoke_kernel)
```
jamxia155: I believe this version has achieved the functional requirements that we discussed [over Slack](https://nvidia.slack.com/archives/C06FHK4NVL7/p1769668803737269). Thanks for solving the issue!

I would like to make one more suggestion. Right now, the variables `kernel` and `invoke_kernel` are independent of each other so a user error where different kernels are passed to `safely_invoke_kernel_with_smem_size` and captured by `invoke_kernel` is possible and may be hard to debug. My idea is to require that `invoke_kernel` be defined to take a kernel as a parameter. That is, before calling `safely_invoke_kernel_with_smem_size`, we shall define `invoke_kernel` as something like:
```
auto const& invoke_kernel = [&](decltype(kernel) Kernel) {
    Kernel<<<...>>>; // or some other user-specified way of launching
};
```
Then inside `safely_invoke_kernel_with_smem_size` we call `invoke_kernel(kernel)` to make sure the right kernel gets invoked. WDYT?

P.S.: in such a design, `invoke_kernel` probably should be renamed to `launch` or some such to avoid redundancy.

mythrocks: Excellent!  Thank you. 👍  I couldn't think of a good way to tie them together. 

I suppose we still technically run the risk of the user not using the kernel that's passed to the launcher lambda.  But this version is tighter.

I'll check that this applies cleanly to the persistent kernel case, and update this PR shortly.

cpp/src/neighbors/detail/smem_utils.cuh
```diff
@@ -0,0 +1,37 @@
...
+{
+  static auto mutex = std::mutex{};
+  auto guard        = std::lock_guard<std::mutex>{mutex};
+  RAFT_CUDA_TRY(
+    cudaFuncSetAttribute(kernel, cudaFuncAttributeMaxDynamicSharedMemorySize, smem_size));
```
jamxia155: In the case where the specified `smem_size` is too large, `RAFT_CUDA_TRY` will terminate the program with `cudaErrorInvalidValue` which may be hard to diagnose. May I suggest a small change:
```
auto status = cudaFuncSetAttribute(kernel, cudaFuncAttributeMaxDynamicSharedMemorySize, smem_size);
RAFT_EXPECTS(status == cudaSuccess, "Failed to set max dynamic shared memory size to %zu bytes.", smem_size);
```

mythrocks: Makes sense to do.  Varies from how the original code functioned, but this will be for the better.

mythrocks:
> Would it be possible to create a small reproducer in cuvs tests and show this PR fixes the bug through that? 

Updating here for completeness:  One repro is available in https://github.com/rapidsai/cuvs-lucene/issues/93.  Finding a C++ repro might take more time than I can invest at the moment. I'm afraid I've burned through the time I had allocated to diagnosing this problem.

On the bright side, @jamxia155 seems to have run into this problem as well.  I wonder if he might have an easy repro for this.

> Sorry for being so picky, but this is a performance-cricital piece of code, so I want to be sure we also don't breaking/slowing down things much here.

Completely understandable; the rigour is warranted.  It might be worth running benchmarks against this patch to check the perf degradation.  

For correctness, IMHO modifying the CUDA-context for the device needs to be thread-safe, because it applies across all host threads in the process.  I've tried capturing the concern in this PR's description.  My uneducated _guess_ is that there shouldn't be a heavy cost, but it does need quantifying eventually.

cpp/src/neighbors/detail/smem_utils.cuh
mythrocks: @jamxia155:  I've fixed this as well.  If you'd like to take another gander.

viclafargue:
Thanks a lot for the great work @mythrocks!

> Would it be possible to create a small reproducer in cuvs tests and show this PR fixes the bug through that?

I do not have a cuVS reproducer, but this small reproducer definitely proves that `cudaFuncSetAttribute` is not thread-safe.

[repro.cu](https://github.com/user-attachments/files/25257676/repro.cu.txt)
`nvcc repro.cu -o repro -lpthread`

We should definitely investigate the performance implications of the fix further. But, I also think that it effectively solves the problem and that it should **in theory** be relatively lightweight performance-wise (uses optimal value for kernel launch + very small critical section).

cpp/src/neighbors/detail/smem_utils.cuh
```diff
@@ -0,0 +1,40 @@
...
+  auto guard        = std::lock_guard<std::mutex>{mutex};
+  auto launch_status =
+    cudaFuncSetAttribute(kernel, cudaFuncAttributeMaxDynamicSharedMemorySize, smem_size);
+  RAFT_EXPECTS(launch_status == cudaSuccess,
+               "Failed to set max dynamic shared memory size to %zu bytes",
+               smem_size);
+  launch(kernel);
+}
+
+}  // namespace cuvs::neighbors::detail
```
achirkin, line 62:
```suggestion
/*
 * SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION.
 * SPDX-License-Identifier: Apache-2.0
 */
#pragma once

#include <raft/core/error.hpp>

#include <atomic>
#include <cstdint>
#include <mutex>

namespace cuvs::neighbors::detail {

/**
 * @brief (Thread-)Safely invoke a kernel with a maximum dynamic shared memory size.
 * This is required because the sequence `cudaFuncSetAttribute` + kernel launch is not executed
 * atomically.
 *
 * Used this way, the cudaFuncAttributeMaxDynamicSharedMemorySize can only grow and thus
 * guarantees that the kernel is safe to launch.
 *
 * @tparam KernelT The type of the kernel.
 * @tparam InvocationT The type of the invocation function.
 * @param kernel The kernel function address (for whom the smem-size is specified).
 * @param smem_size The size of the dynamic shared memory to be set.
 * @param launch The kernel launch function/lambda.
 */
template <typename KernelT, typename KernelLauncherT>
void safely_launch_kernel_with_smem_size(KernelT const& kernel,
                                         uint32_t smem_size,
                                         KernelLauncherT const& launch)
{
  // the last smem size is parameterized by the kernel thanks to the template parameter.
  static std::atomic<uint32_t> current_smem_size{0};
  auto last_smem_size = current_smem_size.load(std::memory_order_relaxed);
  if (smem_size > last_smem_size) {
    // We still need a mutex for the critical section: actualize last_smem_size and set the
    // attribute.
    static auto mutex = std::mutex{};
    auto guard        = std::lock_guard<std::mutex>{mutex};
    if (!current_smem_size.compare_exchange_strong(
          last_smem_size, smem_size, std::memory_order_relaxed, std::memory_order_relaxed)) {
      // The value has been updated by another thread between the load and the mutex acquisition.
      if (smem_size > last_smem_size) {
        current_smem_size.store(smem_size, std::memory_order_relaxed);
      }
    }
    // Only update if the last seen value is smaller than the new one.
    if (smem_size > last_smem_size) {
      auto launch_status =
        cudaFuncSetAttribute(kernel, cudaFuncAttributeMaxDynamicSharedMemorySize, smem_size);
      RAFT_EXPECTS(launch_status == cudaSuccess,
                   "Failed to set max dynamic shared memory size to %zu bytes",
                   smem_size);
    }
  }
  // We don't need to guard the kernel launch because the smem_size can only grow.
  return launch(kernel);
}

}  // namespace cuvs::neighbors::detail

```

achirkin: Here are my thoughts:

1. Change the wording a little bit (it's not the cudaFuncSetAttribute not thread safe, but the cudaFuncSetAttribute + kernel launch is not atomic)
2. Cache the max smem_size attribute to update it only if we need to increase the memory size
3. Only lock to keep cudaFuncSetAttribute and the atomic in sync

Changing the reproducer test to run for longer (more concurrent indices and 10K search iterations) makes the difference visible: 33 seconds for the full lock vs 29 seconds with atomics.

mythrocks: I think there's a small misunderstanding (perhaps on my part):  the current solution no longer tracks the largest max-smem size.  It should no longer be ratcheting upwards. 

Since every kernel is called with its max smem setting, there should no longer be an impact on occupancy. 

@achirkin: You're not suggesting that we _should_ ratchet up the smem setting, are you?  Maybe I've misunderstood.

achirkin: Yes, that's correct!
I don't think there's any significant downside to keeping the `cudaFuncAttributeMaxDynamicSharedMemorySize` at the maximum requested value. I tried to dig through the developer guide and other docs and I'm still not sure if `cudaFuncSetAttribute` can affect the device carveout configuration; (a) if it ever does, then this also means blocking the device, which is very bad if you do it repeatedly (much worse than potential occupancy drop in _some_ of the calls in multi-threaded environment) (b) if it doesn't, then I don't see how a higher-than-needed value can affect performance at all.

NB: we don't set the actual dynamic shared memory per block via `cudaFuncSetAttribute` call - that one is set via kernel launch parameters (and cannot exceed `cudaFuncAttributeMaxDynamicSharedMemorySize`).

jamxia155: Taking this argument to the other extreme, could we get away with doing a `std::call_once` to set `cudaFuncAttributeMaxDynamicSharedMemorySize` to the maximum hardware-supported value and just forget about it thereafter?

achirkin: I think this is a question worth deeper investigation (maybe outside of this PR to make it finally through). If we look at the function safely_launch_kernel_with_smem_size alone, there are cases when we explicitly set the carveout (in IVF-PQ), so the maximum supported `cudaFuncAttributeMaxDynamicSharedMemorySize` actually changes with it (and increasing carveout when you don't need all of the requested shared memory actively hurts the occupancy for nothing).

mythrocks:
I have to say, @achirkin:  With your recent changes to this PR, it suddenly looks a whole lot more rigorous a solution.  Thank you for investing time to strengthen the fix.

mythrocks:
I've fixed the formatting breaks, and up-merged.

cpp/src/neighbors/detail/cagra/search_single_cta_kernel-inl.cuh
```diff
@@ -1979,10 +1980,6 @@ struct alignas(kCacheLineBytes) persistent_runner_t : public persistent_runner_b
     // initialize the dataset/distance descriptor
     auto* dd_dev_ptr = dd_host.dev_ptr(stream);
 
-    // set kernel attributes same as in normal kernel
-    RAFT_CUDA_TRY(
-      cudaFuncSetAttribute(kernel, cudaFuncAttributeMaxDynamicSharedMemorySize, smem_size));
-
     // set kernel launch parameters
```
achirkin, line 1987: Revert the change for the persistence codepath (as the helper as removed already): only one persistent kernel can run per device at the same time.
```suggestion
    // set kernel attributes same as in normal kernel
    RAFT_CUDA_TRY(
      cudaFuncSetAttribute(kernel, cudaFuncAttributeMaxDynamicSharedMemorySize, smem_size));

    // set kernel launch parameters
```

mythrocks:
Sorry for the delay on this PR. I'm investigating the failures now.

mythrocks:
I've fixed the silly error with the format string.  Watching the build now.

mythrocks:
The compile error is unrelated to the current change.  (This is again because of `thrust::make_counting_iterator` and `thrust::for_each`.)

I've pulled in the change from https://github.com/rapidsai/cuvs/pull/1825 to fix the build error.

mythrocks:
> I've pulled in the change from https://github.com/rapidsai/cuvs/pull/1825 to fix the build error.

~~Ok, #1825 doesn't address it.  I'll post a fix to the `make_counting_iterator` problem, and pull the fix in here.~~

Belay that.  #1825 does fix the problem, now that I've cherrypicked it correctly.  Waiting on CI.

mythrocks:
Passing CI, at last. :] 
@achirkin: Should we consider merging this one?

achirkin:
Yes, waiting on @rapidsai/cuvs-cmake-codeowners

PR #1772, opened 2026-02-04
Use verify-hardcoded-version pre-commit hook

KyleFromNVIDIA:

### DIFF
--- a/java/cuvs-java/src/test/java/com/nvidia/cuvs/spi/CuVSProviderIT.java
+++ b/java/cuvs-java/src/test/java/com/nvidia/cuvs/spi/CuVSProviderIT.java
@@ -112,8 +112,8 @@ public class CuVSProviderIT extends CuVSTestCase {
           Please upgrade your software, or install a previous version of libcuvs_c.\
           """,
           ex.getMessage());
-      // rapids-pre-commit-hooks: enable[verify-hardcoded-version]
       System.setProperty("cuvs.max_version", "26.12.0");
+      // rapids-pre-commit-hooks: enable[verify-hardcoded-version]
       checkCuVSVersionMatching("25.12.0", 26, 12, 0);
     } catch (ProviderInitializationException e) {
       throw new AssertionError(e);


python/cuvs_bench/cuvs_bench/plot/__main__.py
```diff
@@ -1,11 +1,12 @@
 #
-# SPDX-FileCopyrightText: Copyright (c) 2024-2025, NVIDIA CORPORATION.
+# SPDX-FileCopyrightText: Copyright (c) 2024-2026, NVIDIA CORPORATION.
 # SPDX-License-Identifier: Apache-2.0
 
 # This script is inspired by
 # 1: https://github.com/erikbern/ann-benchmarks/blob/main/plot.py
 # 2: https://github.com/erikbern/ann-benchmarks/blob/main/ann_benchmarks/plotting/utils.py  # noqa: E501
 # 3: https://github.com/erikbern/ann-benchmarks/blob/main/ann_benchmarks/plotting/metrics.py  # noqa: E501
+# rapids-pre-commit-hooks: disable-next-line[verify-hardcoded-version]
```
jameslamb, line 9: Why was this one necessary? Because at release time the `/main/` in this URL will be replaced with `release/26.04` or similar?

The hook should avoid raising errors for things like this, these cases is lines beginning with whitespace or `#` or `//` would never be replaced by dynamically reading the value of `VERSION`.

There are more cases like this:

https://github.com/search?q=org%3Arapidsai+%2Fhttps.*%5C%2Frapidsai%5C%2F.*%5C%2Fmain%2F&type=code

Can we adjust the hook to not require `disable-next-line` comments for cases like this?

rust/Cargo.toml
```diff
@@ -6,6 +6,7 @@ members = [
 resolver = "2"
 
 [workspace.package]
+# rapids-pre-commit-hooks: disable-next-line[verify-hardcoded-version]
```
jameslamb, line 9: TIL there's a long-running conversation about being able to set the version dynamically: https://github.com/rust-lang/cargo/issues/6583

Maybe worth subscribing to notifications there so you could come back and switch these to that in the future if it's supported.

jameslamb (commented):
Left one suggestion for that case of a URL in a code comment.

python/cuvs_bench/cuvs_bench/plot/__main__.py
KyleFromNVIDIA: > Because at release time the `/main/` in this URL will be replaced with release/26.04 or similar?

Yes.

> Can we adjust the hook to not require `disable-next-line` comments for cases like this?

It would potentially (though admittedly unlikely) have false negatives due to the fact that such comments could be inside a string. In order to eliminate these, we would need a full lexer (but not parser) for each programming language that we support.

Just use the directive. This is exactly what it's meant for.

rust/Cargo.toml
KyleFromNVIDIA: Subscribed

python/cuvs_bench/cuvs_bench/plot/__main__.py
jameslamb: Ok. I still think this is going to annoy people, and would prefer the heuristics be expanded. That case of the possible false negative does sound unlikely as you said, I wouldn't support involving a full lexer to work around it.

BUT I guess when you adjust my GitHub search to only Python code (https://github.com/search?q=org%3Arapidsai+%2Fhttps.*%5C%2Frapidsai%5C%2F.*%5C%2Fmain%2F+language%3APython&type=code), there aren't currently that many of these so it's fine if you want to go forward.

I'll approve and get out of the way.

PR #1779, opened 2026-02-06
Fix build.sh: build cuvs_c for examples and fix --gpu-arch parsing

achirkin:
- Add 'examples' target to the C++ build condition and include cuvs_c in CMAKE_TARGET so that libcuvs_c.so is built before the C examples that link against it.
- Fix greedy regex in gpuArch() that consumed all remaining arguments after --gpu-arch, causing the script to skip the cmake build step.

### DIFF
--- a/cpp/.clang-tidy
+++ b/cpp/.clang-tidy
@@ -4,7 +4,6 @@
 Checks:          'clang-diagnostic-*,clang-analyzer-*,modernize-*,-modernize-make-*,-modernize-raw-string-literal,google-*,-google-default-arguments,-clang-diagnostic-#pragma-messages,readability-identifier-naming,-*,modernize-*,-modernize-make-*,-modernize-raw-string-literal,google-*,-google-default-arguments,-clang-diagnostic-#pragma-messages,readability-identifier-naming'
 WarningsAsErrors: ''
 HeaderFilterRegex: ''
-AnalyzeTemporaryDtors: false
 FormatStyle:     none
 User:            snanditale
 CheckOptions:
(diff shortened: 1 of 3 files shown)


cjnolet (approved):

build.sh
```diff
@@ -224,7 +224,7 @@ function gpuArch {
 
     # Check for gpu-arch option
     if [[ -n $(echo "$ARGS" | { grep -E "\-\-gpu\-arch" || true; } ) ]]; then
-        GPU_ARCH_ARG=$(echo "$ARGS" | { grep -Eo "\-\-gpu\-arch=.+( |$)" || true; })
+        GPU_ARCH_ARG=$(echo "$ARGS" | { grep -Eo "\-\-gpu\-arch=[^ ]+" || true; })
```
jameslamb, line 227: I looked at the `conda-cpp-builds` here and saw the following:

CUDA 12.9

```text
│ │ -- Project CUVS is building for CUDA architectures:
 │ │   70-real
 │ │   75-real
 │ │   80-real
 │ │   86-real
 │ │   90a-real
 │ │   100f-real
 │ │   120a-real
 │ │   120
```

CUDA 13.1

```text
│ │ -- Project CUVS is building for CUDA architectures:
 │ │   75-real
 │ │   80-real
 │ │   86-real
 │ │   90a-real
 │ │   100f-real
 │ │   120a-real
 │ │   120
```

([build link](https://github.com/rapidsai/cuvs/actions/runs/21774734407/job/63167010594?pr=1779))

Which looks right. realized after that that the conda recipes don't pass `--gpu-arch` anyway, but it's good to confirm this wouldn't affect the cuvs conda packages (and I know it won't affect wheels because those don't use `build.sh` at all).

PR #1780, opened 2026-02-06
Fix potential OOB access in CAGRA search when graph size < dataset size

irina-resh-nvda:
Random seed node selection incorrectly used `dataset_desc.size` to generate indices. When `graph.extent(0) < dataset.size`, this caused OOB access by attempting to index graph rows that don't exist.

This PR fixes an out-of-bounds memory access bug in CAGRA search that occurs when the graph has fewer nodes than the dataset (e.g., during iterative CAGRA-Q builds with compression).

**Solution**

Thread `graph.extent(0)` through the search kernel call chain as a `graph_size` parameter. Random seeds are now correctly constrained to `[0, graph.extent(0))` instead of `[0, dataset.size)`.

Purely internal fix with no API changes. Existing behavior unchanged for normal CAGRA usage where graph and dataset sizes match.

**Testing**

Added test that builds an index on 5,000 points, expands the dataset to 10,000 points via `update_dataset()`, then runs search. This reproduces the exact scenario where the bug would occur. Test verifies both SINGLE_CTA and MULTI_CTA algorithms complete without OOB access.

### DIFF
--- a/c/include/cuvs/distance/pairwise_distance.h
+++ b/c/include/cuvs/distance/pairwise_distance.h
@@ -1,5 +1,5 @@
 /*
- * SPDX-FileCopyrightText: Copyright (c) 2024, NVIDIA CORPORATION.
+ * SPDX-FileCopyrightText: Copyright (c) 2024-2026, NVIDIA CORPORATION.
  * SPDX-License-Identifier: Apache-2.0
  */
 #pragma once
@@ -11,6 +11,11 @@
 #ifdef __cplusplus
 extern "C" {
 #endif
+
+/**
+ * @defgroup pairwise_distance_c C pairwise distance
+ * @{
+ */
 
 /**
  * @brief Compute pairwise distances for two matrices
@@ -46,6 +51,9 @@ cuvsError_t cuvsPairwiseDistance(cuvsResources_t res,
                                  DLManagedTensor* dist,
                                  cuvsDistanceType metric,
                                  float metric_arg);
+
+/** @} */
+
 #ifdef __cplusplus
 }
 #endif
--- a/c/include/cuvs/neighbors/cagra.h
+++ b/c/include/cuvs/neighbors/cagra.h
@@ -828,6 +828,15 @@ cuvsError_t cuvsCagraIndexFromArgs(cuvsResources_t res,
                                    cuvsCagraIndex_t index);
 
 /**
+ * @}
+ */
+
+/**
+ * @defgroup cagra_c_index_merge CAGRA C-API merge functions
+ * @{
+ */
+
+/**
  * @brief Merge multiple CAGRA indices into a single CAGRA index.
  *
  * All input indices must have been built with the same data type (`index.dtype`) and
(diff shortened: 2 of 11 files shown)


cpp/src/neighbors/detail/cagra/search_multi_cta_kernel-inl.cuh
```diff
@@ -627,7 +629,8 @@ void select_and_run(const dataset_descriptor_host<DataT, IndexT, DistanceT>& dat
                                                        ps.min_iterations,
                                                        ps.max_iterations,
                                                        num_executed_iterations,
-                                                       sample_filter);
+                                                       sample_filter,
+                                                       static_cast<uint32_t>(graph.extent(0)));
```
mfoerste4: Does this have to be IndexT as well?
```suggestion
                                                       static_cast<IndexT>(graph.extent(0)));
```

cpp/src/neighbors/detail/cagra/search_multi_cta_kernel-inl.cuh
```diff
@@ -193,7 +193,8 @@ RAFT_KERNEL __launch_bounds__(1024, 1) search_kernel(
   const uint32_t min_iteration,
   const uint32_t max_iteration,
   uint32_t* const num_executed_iterations, /* stats */
-  SAMPLE_FILTER_T sample_filter)
+  SAMPLE_FILTER_T sample_filter,
+  const typename DATASET_DESCRIPTOR_T::INDEX_T graph_size = 0)
```
mfoerste4, line 198: Is the default value used anywhere besides tests?

cpp/src/neighbors/detail/cagra/search_single_cta_kernel-inl.cuh
```diff
@@ -2341,7 +2347,8 @@ control is returned in this thread (in persistent_runner_t constructor), so we'r
                                                            hash_bitlen,
                                                            small_hash_bitlen,
                                                            small_hash_reset_interval,
-                                                           sample_filter);
+                                                           sample_filter,
+                                                           static_cast<uint32_t>(graph.extent(0)));
```
mfoerste4: Same as above

cpp/src/neighbors/detail/cagra/search_multi_cta_kernel-inl.cuh
irina-resh-nvda: the default is used everywhere but the tests and  thefuture iterative cagra q implementation

cpp/tests/neighbors/ann_cagra/bug_graph_smaller_than_dataset.cu
```diff
@@ -0,0 +1,160 @@
...
+class cagra_graph_smaller_than_dataset_test : public ::testing::Test {
+ public:
+  using data_type  = float;
+  using index_type = uint32_t;
+
+ protected:
+  void run()
+  {
+    // Create a dataset with 10000 points
+    constexpr int64_t n_dataset = 10000;
```
tfeher: Do we need 10k points to test this? If we ca make the test lighter (e.g 1k points) that would be great. 

Since this is a single test, it is not a big issue, but in general we test CAGRA for a wide range of parameters, therefore we should be conscious about the test time.

irina-resh-nvda: Of course! Done

PR #1781, opened 2026-02-09
Fix multi-threaded CAGRA search OOB

### LINKED ISSUE #1802: [BUG] OOB access in multi-threaded CAGRA search
```
========= Invalid __global__ read of size 8 bytes
=========     at void cuvs::neighbors::cagra::detail::single_cta_search::search_kernel<(unsigned int)64, (unsigned int)64, (unsigned int)1, cuvs::neighbors::cagra::detail::dataset_descriptor_base_t<float, unsigned int, float>, unsigned int, cuvs::neighbors::filtering::none_sample_filter>(unsigned long, T4::DISTANCE_T *, unsigned int, const T4*, const T4::DATA_T *, const T4::INDEX_T *, unsigned int, const T5 *, unsigned int, unsigned long, const T4::INDEX_T *, unsigned int, T4::INDEX_T *, unsigned int, unsigned int, unsigned int, unsigned int, unsigned int *, unsigned int, unsigned int, unsigned int, T6)+0x40
=========     by thread (0,0,0) in block (0,15,0)
=========     Access to 0x6290000020 is potentially made before memory is allocated
=========     and is inside the nearest allocation at 0x6290000000 of size 64 bytes
=========     Saved host backtrace up to driver entry point at kernel launch time
=========         Host Frame: cuLaunchKernel_ptsz [0x3534c4] in libcuda.so.1
=========         Host Frame:  [0x14e7c] in libcudart.so.13
=========         Host Frame: cudaLaunchKernel_ptsz [0x572b7] in libcudart.so.13
```

viclafargue:
Closes #1802

Fixes OOB that happens during multi-threaded CAGRA search.

```
========= Invalid __global__ read of size 8 bytes
=========     at void cuvs::neighbors::cagra::detail::single_cta_search::search_kernel<(unsigned int)64, (unsigned int)64, (unsigned int)1, cuvs::neighbors::cagra::detail::dataset_descriptor_base_t<float, unsigned int, float>, unsigned int, cuvs::neighbors::filtering::none_sample_filter>(unsigned long, T4::DISTANCE_T *, unsigned int, const T4*, const T4::DATA_T *, const T4::INDEX_T *, unsigned int, const T5 *, unsigned int, unsigned long, const T4::INDEX_T *, unsigned int, T4::INDEX_T *, unsigned int, unsigned int, unsigned int, unsigned int, unsigned int *, unsigned int, unsigned int, unsigned int, T6)+0x40
=========     by thread (0,0,0) in block (0,15,0)
=========     Access to 0x6290000020 is potentially made before memory is allocated
=========     and is inside the nearest allocation at 0x6290000000 of size 64 bytes
=========     Saved host backtrace up to driver entry point at kernel launch time
=========         Host Frame: cuLaunchKernel_ptsz [0x3534c4] in libcuda.so.1
=========         Host Frame:  [0x14e7c] in libcudart.so.13
=========         Host Frame: cudaLaunchKernel_ptsz [0x572b7] in libcudart.so.13
```


cpp/src/neighbors/detail/cagra/compute_distance.hpp
```diff
@@ -257,6 +259,10 @@ struct dataset_descriptor_host {
         dev_descriptor_t* ptr = nullptr;
         RAFT_CUDA_TRY(cudaMallocAsync(&ptr, size, stream));
         fun(ptr, stream);
+        // Record an event after initialization so that other streams can establish
+        // a GPU-side dependency without expensive host synchronization.
+        RAFT_CUDA_TRY(cudaEventCreateWithFlags(&init_event, cudaEventDisableTiming));
```
achirkin: Do you know if this call ever synchronizes the device or the context? 
In any case I think it could make sense to move the event creation above `cudaMallocAsync` (which sometimes does synchronize and sometimes doesn't).

achirkin (commented):
Thanks, @viclafargue , the solution looks nice. Fingers crossed it indeed solves the bug in question :)

cpp/src/neighbors/detail/cagra/compute_distance.hpp
viclafargue: > Do you know if this call ever synchronizes the device or the context?

`cudaEventCreateWithFlags` does not synchronize the device. It's a lightweight host-side allocation that only briefly acquires the CUDA context lock.

> In any case I think it could make sense to move the event creation above cudaMallocAsync (which sometimes does synchronize and sometimes doesn't).

Makes sense, I just updated this.

cpp/src/neighbors/detail/cagra/compute_distance.hpp
```diff
@@ -255,8 +257,12 @@ struct dataset_descriptor_host {
       if (std::holds_alternative<init_f>(value)) {
         auto& [fun, size]     = std::get<init_f>(value);
         dev_descriptor_t* ptr = nullptr;
+        RAFT_CUDA_TRY(cudaEventCreateWithFlags(&init_event, cudaEventDisableTiming));
```
mythrocks, line 242: Based on the feedback received on my PR (https://github.com/rapidsai/cuvs/pull/1771/changes#diff-52f864438c5274d5d365954ba71d9988bb00f8c415e2b3c081aa532f3a1a8f07R34-R36), one wonders if debugging might be easier if one used `RAFT_EXPECTS` with a message here instead?

What is the generally accepted convention?

viclafargue: Good question. It looks like [the convention is to use `RAFT_CUDA_TRY` for error handling of CUDA API calls](https://github.com/rapidsai/raft/blob/main/docs/source/developer_guide.md#error-handling) while `RAFT_EXPECTS` would most often be used for assertion testing. `RAFT_CUDA_TRY` should point out the file and line where the error happened. But, I guess that `RAFT_EXPECTS` may indeed come useful when one wants to display a specific error message.

cpp/src/neighbors/detail/cagra/compute_distance.hpp
```diff
@@ -237,8 +239,12 @@ struct dataset_descriptor_host {
...
         RAFT_CUDA_TRY(cudaMallocAsync(&ptr, size, stream));
         fun(ptr, stream);
+        // Record an event after initialization so that other streams can establish
+        // a GPU-side dependency without expensive host synchronization.
+        RAFT_CUDA_TRY(cudaEventRecord(init_event, stream));
         value = std::make_tuple(ptr, stream);
         ready.store(true, std::memory_order_release);
```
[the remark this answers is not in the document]
mythrocks: Hmm.  This is a good point, IMHO.  If the `cudaEventRecord` line throws, that's the last we'll see of the `ptr` allocation. :/

@viclafargue, do we have an RAII-wrapper specifically for cases like this one?

achirkin:
@viclafargue that's funny I completely forgot about this PR and that I reviewed it and I arrived at almost the same solution in https://github.com/NVIDIA/cuvs/pull/2237

viclafargue:
> @viclafargue that's funny I completely forgot about this PR and that I reviewed it and I arrived at almost the same solution in #2237

Seems like this may be the correct solution then. It looks like yours is already approved, so we can merge it.

achirkin:
Ok, sorry for the confusion! Closing this as the other one is merged.

Issue #1678, opened 2026-01-07
[Task] Remove current implementation of IndexWrapper


cjnolet:
We will eventually have a `composite_index`, but we don't need the Indexwrapper in the meantime.

PR #1792, opened 2026-02-11
Remove IndexWrapper

aamijar:
Resolves https://github.com/rapidsai/cuvs/issues/1678.

This PR removes the `cuvs::neighbors::IndexWrapper` and `cuvs::neighbors::IndexBase`.
For logical merge we can use composite_index and pass in the cagra::index types as input.
For physical merge we can use the cagra::merge and cagra::search free functions.


cpp/include/cuvs/neighbors/index_base.hpp
```diff
@@ -87,6 +88,45 @@ struct IndexBase {
...
+   * The merge strategy and parameters are determined by the specific merge_params implementation.
+   * Not all index types need to support merging, so this has a default implementation that
+   * throws an error.
+   *
+   * @param[in] handle RAFT resources for executing operations
+   * @param[in] params Merge parameters containing strategy and algorithm-specific settings
+   * @param[in] other_indices Vector of other indices to merge with this one
+   * @return Shared pointer to merged index
+   */
+  virtual std::shared_ptr<IndexBase<value_type, index_type, out_index_type>> merge(
```
cjnolet: Definitely don't want this to be in a class- cuVS is not an object-oriented API, it's a set of C-like free functions that act upon stateful POCOs.

aamijar: I see that `cagra_index_wrapper.hpp` and its corresponding definition `cagra_index_wrapper.cu` implement the above interface functions merge and convert_serach_params. 
So I guess we would need to remove those override functions as well? In that case, is the idea that users would directly call `cagra::merge` or create their own `cuvs::neighbors::composite::CompositeIndex`? Here is the relevant code that would need to be removed from `cagra_index_wrapper.cu`. 
https://github.com/rapidsai/cuvs/blob/b0dccf2f7efea83f88f3468cf350658b9ffc9b6b/cpp/src/neighbors/cagra_index_wrapper.cu#L51-L88

cjnolet: Yeah, so the "composite_index" would just be a lightweight wrapper (like tiered index) which is created said free functions. It can have accessors on it (ideally no public mutators, so it should be immutable once created). The idea is that it wraps itself around multiple existing indexes and hijacks the query, such that it does a logical "query" (kind of like the sharded multi gpu index)across all indexes it wraps. 

But to he honest, if the composite_index has not been defined / used in the current impl, we can just remove all of this wrapper api an impl.

Ideally we would have a "make_composite_index" in cagra.hpp which would accept multiple cagra indexes and logically wrap them like mentioned dive.

cpp/include/cuvs/neighbors/composite/index.hpp
```diff
@@ -1,76 +1,78 @@
...
+ *   size_t n_streams = 2;
+ *   raft::resource::set_cuda_stream_pool(handle,
+ *                                        std::make_shared<rmm::cuda_stream_pool>(n_streams));
+ *
+ *   composite.search(handle, search_params, queries, neighbors, distances);
+ * @endcode
  */
 template <typename T, typename IdxT, typename OutputIdxT = IdxT>
-class CompositeIndex : public IndexBase<T, IdxT, OutputIdxT> {
+class CompositeIndex {
```
cjnolet: We use snake case (for eg. composite_index) for all other class names. Let's continue to use the same style standards here for consistency.

aamijar: Addressed in https://github.com/rapidsai/cuvs/pull/1792/commits/2052e39a7a8ca12fc28871e299b19691b3e3e8ad

PR #1793, opened 2026-02-11
Add tarball build from source and installation instructions

singhmanas1:
## Description

Adds C library setup and verification tooling to the `notebooks/` directory:

| File | Purpose |
|------|---------|
| `setup_and_test_cuvs_c.sh` | Build, extract, and test libcuvs with auto-detected paths |
| `build_libcuvs_c.sh` | Build libcuvs and pack into tarball |
| `cuvs_setup_tutorial.md` | Step-by-step tutorial for C lib installation |

Scripts are intended to be run from `notebooks/` and infer paths relative to the repo root. They handle conda setup, install missing deps, and run Python and C API sanity checks.


notebooks/setup_and_test_cuvs_c.sh
```diff
@@ -0,0 +1,270 @@
...
+if d: os.environ["LD_LIBRARY_PATH"] = d + ":" + os.environ.get("LD_LIBRARY_PATH", "")
+try: ctypes.CDLL("libcuda.so.1")
+except OSError: fail("libcuda.so.1 not found")
+try: cuvs = ctypes.CDLL("libcuvs_c.so")
+except OSError as e: fail(f"Could not load libcuvs_c.so: {e}")
+cuvs.cuvsResourcesCreate.argtypes = [ctypes.POINTER(ctypes.c_void_p)]; cuvs.cuvsResourcesCreate.restype = ctypes.c_int
+cuvs.cuvsStreamSync.argtypes = [ctypes.c_void_p]; cuvs.cuvsStreamSync.restype = ctypes.c_int
+cuvs.cuvsResourcesDestroy.argtypes = [ctypes.c_void_p]; cuvs.cuvsResourcesDestroy.restype = ctypes.c_int
+r = ctypes.c_void_p()
+cuvs.cuvsResourcesCreate(ctypes.byref(r)) or fail("cuvsResourcesCreate")
```
dantegd, line 187: In cuVS C, `CUVS_SUCCESS` is typically 0, no?

Python's `or` will short-circuit on "truth-y" values: `0 or fail(...)` will end up calling `fail()` on success because `0` is "false-y".  Every successful API call will be treated as a failure, then, I think this should be: 


```suggestion
if cuvs.cuvsResourcesCreate(ctypes.byref(r)) != 0: fail("cuvsResourcesCreate")
```

notebooks/setup_and_test_cuvs_c.sh
```diff
@@ -0,0 +1,270 @@
...
+nvidia-smi &>/dev/null || fail "nvidia-smi failed. Install NVIDIA driver."
+nvidia-smi --query-gpu=name,driver_version --format=csv,noheader 2>/dev/null || nvidia-smi
+if ! ldconfig -p 2>/dev/null | grep -q "libcuda"; then
+  find "${CONDA_PREFIX:-/none}" -name "libcuda*.so*" 2>/dev/null | head -1 | grep -q . || fail "libcuda not found."
+  info "libcuda via conda path"
+fi
+echo ""
+
+# -----------------------------------------------------------------------------
+# Step 4: Check libcuvs_c.so dependencies via ldd; install librmm/pylibraft if missing
```
dantegd, line 151:
```suggestion
# Step 3: Check libcuvs_c.so dependencies via ldd; install librmm/pylibraft if missing
```

Step 3 is missing which is throwing off the comments of number of steps

notebooks/setup_and_test_cuvs_c.sh
```diff
@@ -0,0 +1,270 @@
...
+# -----------------------------------------------------------------------------
+info "Step 4: Verifying libcuvs_c dependencies..."
```
dantegd, line 153:
```suggestion
info "Step 3: Verifying libcuvs_c dependencies..."
```

notebooks/setup_and_test_cuvs_c.sh
```diff
@@ -0,0 +1,270 @@
...
+  set -u
+  export LD_LIBRARY_PATH="${INSTALL_PREFIX}/lib:${CONDA_PREFIX:-}/lib:${LD_LIBRARY_PATH:-}"
+  MISSING=$(ldd "${LIBCUVS_C}" 2>/dev/null | grep "not found" || true)
+  [[ -n "${MISSING}" ]] && echo "${MISSING}" && fail "Dependencies still missing. Try: micromamba create -n cuvs -f ${CUVS_YAML}"
+fi
+info "All dependencies resolve OK"
+echo ""
+
+# -----------------------------------------------------------------------------
+# Step 5: Python sanity check (ResourcesCreate, StreamSync, ResourcesDestroy)
```
dantegd, line 170:
```suggestion
# Step 4: Python sanity check (ResourcesCreate, StreamSync, ResourcesDestroy)
```

notebooks/setup_and_test_cuvs_c.sh
```diff
@@ -0,0 +1,270 @@
...
+# -----------------------------------------------------------------------------
+info "Step 5: Python sanity check..."
```
dantegd, line 172:
```suggestion
info "Step 4: Python sanity check..."
```

notebooks/setup_and_test_cuvs_c.sh
```diff
@@ -0,0 +1,270 @@
...
+cuvs.cuvsResourcesDestroy(r) or fail("cuvsResourcesDestroy")
+print("cuVS sanity check PASSED")
+SANITY_EOF
+export CUVS_LIB_DIR="${INSTALL_PREFIX}/lib"
+python3 "${CUVS_SANITY_PY}" || { rm -f "${CUVS_SANITY_PY}"; fail "Python sanity failed."; }
+rm -f "${CUVS_SANITY_PY}"
+echo ""
+
+# -----------------------------------------------------------------------------
+# Step 6: C API test (compile minimal program, link libcuvs_c, run)
```
dantegd, line 198:
```suggestion
# Step 5: C API test (compile minimal program, link libcuvs_c, run)
```

notebooks/setup_and_test_cuvs_c.sh
```diff
@@ -0,0 +1,270 @@
...
+# -----------------------------------------------------------------------------
+info "Step 6: C API test..."
```
dantegd, line 200:
```suggestion
info "Step 5: C API test..."
```

notebooks/setup_and_test_cuvs_c.sh
```diff
@@ -0,0 +1,270 @@
...
+  [[ -d "${d}" ]] && CUDA_LIB_FLAGS+=" -L${d} -Wl,-rpath,${d}"
+done
+gcc -o "${SCRIPT_DIR}/.cuvs_minimal_test" "${MINIMAL_TEST_SRC}" ${CUDA_INC_FLAGS} ${CUDA_LIB_FLAGS} \
+  -lcuvs_c -lcudart -lm
+"${SCRIPT_DIR}/.cuvs_minimal_test" || fail "C API test failed"
+rm -f "${MINIMAL_TEST_SRC}" "${SCRIPT_DIR}/.cuvs_minimal_test"
+echo ""
+
+# -----------------------------------------------------------------------------
+# Step 7: Optional L2 distance example (if source exists in cuvs repo)
```
dantegd, line 236:
```suggestion
# Step 6: Optional L2 distance example (if source exists in cuvs repo)
```

notebooks/setup_and_test_cuvs_c.sh
```diff
@@ -0,0 +1,270 @@
...
+# -----------------------------------------------------------------------------
+L2_SRC="${CUVS_REPO}/examples/c/src/L2_c_example.c"
+COMMON_SRC="${CUVS_REPO}/examples/c/src/common.c"
+if [[ -f "${L2_SRC}" ]] && [[ -f "${CUVS_REPO}/examples/c/src/common.h" ]]; then
+  info "Step 7: L2 example..."
```
dantegd, line 241:
```suggestion
  info "Step 6: L2 example..."
```

notebooks/setup_and_test_cuvs_c.sh
```diff
@@ -0,0 +1,270 @@
...
+# -----------------------------------------------------------------------------
+if [[ $DO_EXTRACT -eq 1 || $DO_TEST -eq 1 ]]; then
+  [[ ! -f "${TARBALL}" ]] && fail "Tarball not found: ${TARBALL}. Run with --build first."
+  info "Step 0: Extracting to ${INSTALL_PREFIX}..."
+  mkdir -p "${INSTALL_PREFIX}"
+  tar -xzf "${TARBALL}" -C "${INSTALL_PREFIX}"
+  [[ ! -f "${LIBCUVS_C}" ]] && fail "libcuvs_c.so not found after extract."
+  info "Located ${LIBCUVS_C}"
+  echo ""
+fi
```
dantegd, line 104: If a user runs `--test` alone, the script will always re-extract the tarball into `INSTALL_PREFIX` unconditionally, overwriting whatever is already installed. This could be surprising, a user running `--test` without build and already has the tarball might just want to verify an existing installation, not re-extract. The test phase should simply check if libcuvs_c.so exists already and use that one when it's the only flag.

notebooks/setup_and_test_cuvs_c.sh
```diff
@@ -0,0 +1,270 @@
...
+set -euo pipefail
+
+# -----------------------------------------------------------------------------
+# Configuration: paths auto-detect from script location (override via env vars)
+# -----------------------------------------------------------------------------
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+REPO_ROOT="$(cd "${SCRIPT_DIR}/.." && pwd)"
+CUVS_REPO="${CUVS_REPO:-${REPO_ROOT}}"
+TARBALL="${CUVS_REPO}/libcuvs_c.tar.gz"
+INSTALL_PREFIX="${CUVS_INSTALL_PREFIX:-/usr/local}"
```
dantegd, line 18: Defaulting to `/usr/local` is what makes us suggest sudo/root. Therefore, the tutorial mentions sudo as a normal path. What is the reasoning for this?  Would defaulting to a user-local path might be a safer  default? And then document both options?

notebooks/setup_and_test_cuvs_c.sh
```diff
@@ -0,0 +1,270 @@
...
+  [[ -n "${MISSING}" ]] && echo "${MISSING}" && fail "Dependencies still missing. Try: micromamba create -n cuvs -f ${CUVS_YAML}"
+fi
+info "All dependencies resolve OK"
+echo ""
+
+# -----------------------------------------------------------------------------
+# Step 5: Python sanity check (ResourcesCreate, StreamSync, ResourcesDestroy)
+# -----------------------------------------------------------------------------
+info "Step 5: Python sanity check..."
+CUVS_SANITY_PY="${SCRIPT_DIR}/.cuvs_sanity.py"
```
dantegd, line 173: Small nitpick: we write temp files to the `notebooks/` directory and for `MINIMAL_TEST_SRC`, right? We could create a subdirectory (called tmp or something) to put the files in there, in case they are not deleted for whatever reason, it would be very clear to users that those files are temporary instead of ending with random files in the `notebooks` folder.

notebooks/setup_and_test_cuvs_c.sh
```diff
@@ -0,0 +1,270 @@
...
+# -----------------------------------------------------------------------------
+# Configuration: paths auto-detect from script location (override via env vars)
+# -----------------------------------------------------------------------------
+SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+REPO_ROOT="$(cd "${SCRIPT_DIR}/.." && pwd)"
+CUVS_REPO="${CUVS_REPO:-${REPO_ROOT}}"
+TARBALL="${CUVS_REPO}/libcuvs_c.tar.gz"
+INSTALL_PREFIX="${CUVS_INSTALL_PREFIX:-/usr/local}"
+ENV_NAME="${CUVS_CONSUMER_ENV:-cuvs}"
+CUVS_YAML="${CUVS_YAML:-${CUVS_REPO}/conda/environments/all_cuda-131_arch-x86_64.yaml}"
```
dantegd, line 20: This assumes a 13.1 x86-64 system, making the CUDA version and arch configurable would be a nice usability improvement. 

Just remembered that that's hardcoded in other parts of the script, so adding a note that we would need to update those too if we go this route.

notebooks/setup_and_test_cuvs_c.sh
```diff
@@ -0,0 +1,270 @@
...
+fi
+# Compose CUDA include/lib flags supporting split CUDA layout on conda (targets/x86_64-linux)
+CUDA_INC_FLAGS="-I${INSTALL_PREFIX}/include"
+for d in "${CONDA_PREFIX:-}/include" "${CONDA_PREFIX:-}/targets/x86_64-linux/include"; do
+  [[ -d "${d}" ]] && CUDA_INC_FLAGS+=" -I${d}"
+done
+CUDA_LIB_FLAGS="-L${INSTALL_PREFIX}/lib -Wl,-rpath,${INSTALL_PREFIX}/lib"
+for d in "${CONDA_PREFIX:-}/lib" "${CONDA_PREFIX:-}/targets/x86_64-linux/lib"; do
+  [[ -d "${d}" ]] && CUDA_LIB_FLAGS+=" -L${d} -Wl,-rpath,${d}"
+done
```
dantegd, line 228: This same code is repeated at the end of the file in step 7, might be nice for mantainability to make it another helper function

notebooks/cuvs_setup_tutorial.md
```diff
@@ -0,0 +1,186 @@
...
+2. Create the conda env `cuvs` if it does not exist
+3. Activate the env and set `LD_LIBRARY_PATH`
+
+---
+
+## Section 5: Test Phase – NVIDIA Driver & CUDA
+
+**Purpose:** Verify the system has a working NVIDIA driver and CUDA runtime.
+
+**Checks:** `nvidia-smi`, GPU info, `libcuda` availability.
```
dantegd, line 122: Adding sample output might make these sections read nicer for users.

notebooks/cuvs_setup_tutorial.md
```diff
@@ -0,0 +1,191 @@
+# cuVS C Library Setup & Verification Tutorial
+
```
cjnolet, line 2: Please add build and install instructions to the build and install guide [here](https://github.com/rapidsai/cuvs/blob/main/docs/source/build.rst) (and please follow the format in those instructions. They should be precise, easy to copy/paste, and to the point. Not exhaustive).

notebooks/cuvs_setup_tutorial.md
```diff
@@ -0,0 +1,191 @@
...
+This tutorial walks you through installing and verifying the cuVS (CUDA Vector Search) C library using the setup script. It is organized into separate phases you can run independently or together.
+
+**Files in this folder:** Run `setup_and_test_cuvs_c.sh` and `build_libcuvs_c.sh` from the `notebooks/` directory of your cuVS clone. Paths auto-detect the repo root.
+
+---
+
+## Overview
+
+**Prerequisites:**
+- micromamba (or conda)
```
cjnolet, line 12: Conda and python are not required to install a tarball.

notebooks/setup_and_test_cuvs_c.sh
```diff
@@ -0,0 +1,266 @@
+#!/bin/bash
+# cuVS Installation & Environment Verification
+# Builds tarball, extracts, verifies deps, installs missing ones, runs sanity checks.
+# Usage: ./setup_and_test_cuvs_c.sh [--build|--extract|--test|--allgpuarch|--all] [tarball]
+# Prereqs: micromamba, Python 3.x, NVIDIA GPU
+#
```
cjnolet, line 6: Let's drop this script. We already have a test script in `ci/test_standalone_c.sh` so this tarball will be well tested in CI for all the variants. 

Let's instead put a couple test commands in the install guide so the user can do a quick sanity check. They don't need an exhaustive test.

notebooks/setup_and_test_cuvs_c.sh
cjnolet: We shouldn't be doing anything with python. The reason for most of our users not wanting to use the rapids enterprise tarball is that they don't want python on their system at all, so that's why they're opting to use a cuvs-specific tarball instead of conda. We also don't need to be providing extensive sanity checks for them- a couple bash lines to test that everything linked okay should be sufficient.

cjnolet:
@singhmanas1, 

From the reviews and offline conversations with @msarahan, I think there's some significant work to be done to make a truly portable build from source option. What do you say for now we include the tarball installation instructions ony, along with your sanity check Python script, and split the build from source out into a different PR? That way once we get the tarball uploaded, we can immediately merge the installation instructions into the docs. Meanwhile, we can continue to iterate on the best way to provide build from source instructions (and I'm really liking @msarahan's suggesiton that we should probably be providing a dockerfile to do it reproducibly and portably).

PR #1796, opened 2026-02-12
Update Faiss and DiskANN Patch to Use C++20

tarang-jain:


cjnolet (approved):

cjnolet:
Do we need to submit a specific patch to faiss and diskann as well, so that we aren't having to maintain this patch for an unbounded amount of time?

tarang-jain:
> Do we need to submit a specific patch to faiss and diskann as well, so that we aren't having to maintain this patch for an unbounded amount of time?

Yes. For Faiss we have one patch per <faiss_version, cuvs_version> pair. Looks like this PR needs to be updated to rename the patch correctly.

aamijar:
Faiss has already updated to cpp 20 for cuvs here https://github.com/facebookresearch/faiss/pull/4881. However, faiss has not had a release that includes that commit.

PR #1798, opened 2026-02-12
Docs installation tarball faq

singhmanas1:
# docs: Add tarball instructions, verification FAQ, and installation improvements

This PR expands the Installation guide (`docs/source/build.rst`) with tarball installation options, build-from-source instructions for creating a tarball, and a new FAQ section for verifying the installation and environment.

### 1. Pre-compiled Tarball Section

- Added a new **Tarball** subsection under Installing Pre-compiled Packages
- Describes what the tarball contains (C/C++ APIs, RAFT, RMM) and what it requires (CUDA Toolkit, NVIDIA driver, NCCL, OpenMP, GLIBC)
- Documents two pre-built tarball options:
  - **Open Source:** Download from [developer.nvidia.com/cuvs](https://developer.nvidia.com/cuvs)
  - **Enterprise Supported:** Download from [NVIDIA NGC](https://catalog.ngc.nvidia.com/) (requires NVAIE license)

### 2. Build from Source Tarball

- Added **Tarball (Build from Source)** subsection under C and C++ Libraries
- Documents the command to build a standalone tarball: `./build.sh standalone --allgpuarch -v`
- Explains that the resulting `libcuvs_c.tar.gz` is created in `cpp/build/`

### 3. Installation & Environment Verification FAQ

- New **Installation & Environment Verification FAQ** section with step-by-step verification:
  - **Step 1:** Locate `libcuvs_c.so` (ldconfig or find)
  - **Step 2:** Verify NVIDIA driver (`nvidia-smi`)
  - **Step 3:** Verify CUDA runtime availability
  - **Step 4:** Verify cuVS shared library dependencies (`ldd`)
  - **Step 5:** Python sanity check script (`cuvsResourcesCreate` → `cuvsStreamSync` → `cuvsResourcesDestroy`)
  - **Step 6:** Run the sanity test
- **Common Errors** table (e.g., `libcuda.so.1 not found`, `Could not load libcuvs_c.so`, etc.)
- **Optional:** Debug library loading with `LD_DEBUG=libs`
- **FAQ** answers (no compilation needed, runs on GPU, safe for production, container usage)
- **Support Checklist** for reporting issues

### 4. De-duplication

- FAQ Prerequisites now references the main installation requirements instead of repeating Linux/GPU

## Files Changed

- `docs/source/build.rst` (+249 lines)

### DIFF
--- a/docs/source/build.rst
+++ b/docs/source/build.rst
@@ -17,6 +17,8 @@ The cuVS software development kit provides APIs for C, C++, Python, and Rust lan
 
   * `C and C++ Libraries`_
 
+    * `Tarball (Build from Source)`_
+
     * `Building the Googletests`_
 
   * `Python Library`_
@@ -26,6 +28,8 @@ The cuVS software development kit provides APIs for C, C++, Python, and Rust lan
   * `Using CMake Directly`_
 
 - `Build Documentation`_
+
+- `Installation & Environment Verification FAQ`_
 
 
 Installing Pre-compiled Packages
@@ -76,6 +80,16 @@ The cuVS Python package can also be `installed through pip <https://docs.rapids.
     pip install cuvs-cu12 --extra-index-url=https://pypi.nvidia.com
 
 Note: these packages statically link the C and C++ libraries so the `libcuvs` and `libcuvs_c` shared libraries won't be readily available to use in your code.
+
+Tarball
+^^^^^^^
+
+The cuVS tarball includes the C and C++ APIs along with core dependencies (RAFT, RMM). It requires a compatible CUDA Toolkit, NVIDIA driver, and system libraries such as NCCL, OpenMP, and GLIBC.
+
+Two pre-built tarballs are available:
+
+1. **Open Source:** Download from `developer.nvidia.com <https://developer.nvidia.com/cuvs>`_.
+2. **Enterprise Supported:** Download from `NVIDIA NGC <https://catalog.ngc.nvidia.com/>`_ (requires NVAIE license).
 
 Build from source
 -----------------
@@ -130,6 +144,18 @@ To disable the multi-gpu features run :
 .. code-block:: bash
 
     ./build.sh libcuvs --no-mg
+
+
+Tarball (Build from Source)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+To build a standalone tarball (``libcuvs_c.tar.gz``) from source, use the ``standalone`` target with ``--allgpuarch`` to build for all supported GPU architectures:
+
+.. code-block:: bash
+
+    ./build.sh standalone --allgpuarch -v
+
+The tarball will be created in the ``cpp/build/`` directory.
 
 
 Building the Googletests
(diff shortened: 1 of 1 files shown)


docs/source/build.rst
```diff
@@ -17,6 +17,8 @@ The cuVS software development kit provides APIs for C, C++, Python, and Rust lan
 
   * `C and C++ Libraries`_
 
+    * `Tarball (Build from Source)`_
```
cjnolet, line 20: This should probably be at the same level is the "Python Library". Users aren't going to come here to see "C & C++" they are going to come here to see "how do I build the tarball?"

docs/source/build.rst
```diff
@@ -77,6 +81,16 @@ The cuVS Python package can also be `installed through pip <https://docs.rapids.
...
 Note: these packages statically link the C and C++ libraries so the `libcuvs` and `libcuvs_c` shared libraries won't be readily available to use in your code.
 
+Tarball
+^^^^^^^
+
+The cuVS tarball includes the C and C++ APIs along with core dependencies (RAFT, RMM). It requires a compatible CUDA Toolkit, NVIDIA driver, and system libraries such as NCCL, OpenMP, and GLIBC.
+
+Two pre-built tarballs are available:
+
+1. **Open Source:** Download from `developer.nvidia.com <https://developer.nvidia.com/cuvs>`_.
```
cjnolet, line 91: This is going to be on `downloads.nvidia.com`. Given that it could be another couple weeks before we have this available, I wonder if we should split this part into a new PR that we merge once the pre-built tarball is available?

docs/source/build.rst
```diff
@@ -132,6 +146,18 @@ To disable the multi-gpu features run :
     ./build.sh libcuvs --no-mg
 
 
+Tarball (Build from Source)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+To build a standalone tarball (``libcuvs_c.tar.gz``) from source, use the ``standalone`` target with ``--allgpuarch`` to build for all supported GPU architectures:
```
cjnolet, line 152: Small nitpick- can you briefly mention that `allgpuarch` is not required, and that if not specified, it'll only build the current architecture installed on the system? Also worth metnioning that the benefit to this is that the final package will be much smaller.

docs/source/build.rst
```diff
@@ -256,3 +282,226 @@ The documentation requires that the C, C++ and Python libraries have been built
 .. code-block:: bash
 
     ./build.sh libcuvs python docs
+
+
+Installation & Environment Verification FAQ
```
cjnolet, line 287: Could we split this out into a separate rust file and link it from the docs (and from the build and install build?). We could put a link in the relvant sections (for build and install) that says "Please see [installation verification] guide to test that cuVS has been installed and is working properly". That way we can keep the build and install guide short and sweet, but have a dedicated guide for th sanity check that we can link directly.

build.sh
```diff
@@ -472,6 +501,14 @@ if (( NUMARGS == 0 )) || hasArg libcuvs || hasArg docs || hasArg tests || hasArg
   fi
 fi
 
+# Create standalone tarball when requested
+if hasArg standalone; then
```
cjnolet, line 505: As @msarahan and i were discussing yesterday, I am abit concerned that this tarball isn't going to be portable enough. It's okay if it's not initially, but we'll need to point this out in the build guide. We should probably also create an issue to have a more portable and reproducible build of the tarball. I think Mike's idea that we could use a dockerfile and then invoke the necessary docker-compose commands inside the build.sh to make it painless would be most ideal. Again- don't need to make that change in this PR, but will want to capture the plan so we don't lose sight of it.

robertmaynard: The portability of a  standalone binary built by a user is entirely dependent on the OS + compiler toolchain that they have built with. 

Currently the binaries we build are done based on a minimal Rocky 8 image which has a controlled set of installed libraries.

I don't see any reason why we should try and add 'smarts' to build.sh to try and launch a container and build in that. Let `build.sh` do a singular job well and that is invoke the CMake build-system. The user can therefore control what dependencies they want, as they are in advanced user mode and know more about the goals they want.

docs/source/build.rst
```diff
@@ -256,3 +282,226 @@ The documentation requires that the C, C++ and Python libraries have been built
...
+
+**Expected result:** GPU is listed, driver version is shown, no error messages.
+
+**Common failure:** ``NVIDIA-SMI has failed because it couldn't communicate with the NVIDIA driver`` — indicates a driver installation or kernel mismatch.
+
+
+Step 3: Verify CUDA Runtime Availability
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+cuVS requires the CUDA runtime provided by the NVIDIA driver.
```
robertmaynard, line 358: CUDA runtime isn't provided by the driver. the `libcuda.so` is the CUDA driver and not the runtime.

libcuda.so -> Driver library, provided by the NVIDIA driver
libcudart.so -> Runtime library, provided by the CUDA toolkit

docs/source/build.rst
```diff
@@ -256,3 +282,226 @@ The documentation requires that the C, C++ and Python libraries have been built
...
+Step 4: Verify cuVS Shared Library Dependencies
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Once `libcuvs_c.so` has been located, run:
+
+.. code-block:: bash
+
+    ldd /full/path/to/libcuvs_c.so
+
+**Expected result:** No lines containing ``not found``; ``libcudart.so``, ``libstdc++.so`` resolve correctly.
```
robertmaynard, line 378: We shouldn't call out `libcudart.so` as we are removing dynamic linking of the CUDA runtime from all of RAPIDS

docs/source/build.rst
```diff
@@ -256,3 +282,226 @@ The documentation requires that the C, C++ and Python libraries have been built
...
+
+    # Check NVIDIA driver
+    try:
+        ctypes.CDLL("libcuda.so.1")
+    except OSError:
+        fail("libcuda.so.1 not found (NVIDIA driver missing or not loaded)")
+
+    # Load cuVS
+    try:
+        cuvs = ctypes.CDLL("libcuvs_c.so")
```
robertmaynard, line 413: You need to load the ABI versioned library, loading the unversioned is bad practise

docs/source/build.rst
```diff
@@ -256,3 +282,226 @@ The documentation requires that the C, C++ and Python libraries have been built
...
+
+Shows required shared library versions.
+
+
+Step 5: cuVS Runtime Sanity Check (Python)
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+This test dynamically loads cuVS, creates GPU resources, synchronizes a CUDA stream, and cleans up.
+
+Copy-paste script: ``cuvs_sanity.py``
```
robertmaynard, line 394: This shouldn't be a inline script but a standalone python file, that also is run in one of examples CI jobs to make sure it is kept up to date with any ABI changes

build.sh
```diff
@@ -472,6 +501,14 @@ if (( NUMARGS == 0 )) || hasArg libcuvs || hasArg docs || hasArg tests || hasArg
...
+    echo "Creating libcuvs_c.tar.gz in build/"
```
robertmaynard, line 506: This isn't at all like the CUVS pre-built binary and shouldn't be merged as is.
At a minimum you need to do the following:

* Pass the following flags when building libcuvs https://github.com/rapidsai/cuvs/blob/main/ci/build_standalone_c.sh#L62-L67
* Have a separate step for the C library which ingests the static libcuvs.a
https://github.com/rapidsai/cuvs/blob/main/ci/build_standalone_c.sh#L78-L79

docs/source/build.rst
```diff
@@ -132,6 +146,18 @@ To disable the multi-gpu features run :
...
+Tarball (Build from Source)
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+To build a standalone tarball (``libcuvs_c.tar.gz``) from source, use the ``standalone`` target with ``--allgpuarch`` to build for all supported GPU architectures:
+
+.. code-block:: bash
+
+    ./build.sh standalone --allgpuarch -v
+
+The tarball will be created in the ``cpp/build/`` directory.
```
dantegd, line 158:
```suggestion
The tarball will be created in the ``build/`` directory.
```

The tarball gets created in `tar czvf "${REPODIR}/build/libcuvs_c.tar.gz" -C "${INSTALL_PREFIX}" .` which is the repo root directory

docs/source/build.rst
```diff
@@ -256,3 +282,226 @@ The documentation requires that the C, C++ and Python libraries have been built
...
+
+    libcuvs_c.so (libc6,x86-64) => /usr/local/cuvs/lib/libcuvs_c.so
+
+Option B: Using find (if not registered)
+
+.. code-block:: bash
+
+    sudo find / -name "libcuvs_c.so" 2>/dev/null
+
+Once located, ensure the directory containing `libcuvs_c.so` is included in ``LD_LIBRARY_PATH``.
```
dantegd, line 340:
```suggestion
Once located, ensure the directory containing ``libcuvs_c.so`` is included in ``LD_LIBRARY_PATH``.
```

docs/source/build.rst
```diff
@@ -256,3 +282,226 @@ The documentation requires that the C, C++ and Python libraries have been built
...
+
+**Expected result:** ``libcuda.so.1 (libc6,x86-64) => /usr/lib/...``
+
+If nothing appears, the NVIDIA driver is not correctly installed.
+
+
+Step 4: Verify cuVS Shared Library Dependencies
+^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+
+Once `libcuvs_c.so` has been located, run:
```
dantegd, line 372:
```suggestion
Once ``libcuvs_c.so`` has been located, run:
```

docs/source/build.rst
```diff
@@ -256,3 +282,226 @@ The documentation requires that the C, C++ and Python libraries have been built
...
+
+**Expected result:** No lines containing ``not found``; ``libcudart.so``, ``libstdc++.so`` resolve correctly.
+
+**Optional (advanced):**
+
+.. code-block:: bash
+
+    objdump -p /full/path/to/libcuvs_c.so | grep NEEDED
+
+Shows required shared library versions.
```
dantegd, line 386:
```suggestion
Shows required shared library sonames with major version. 
```

The description is slightly misleading, the soname has the major version, but not the full version. (e.g., libcudart.so.12). Wording it something like this would be more accurate and reflect what users will see, e.g.

```bash
  NEEDED               libcublas.so.13
  NEEDED               libcusolver.so.12
  NEEDED               libcusparse.so.12
...
```

PR #1799, opened 2026-02-12
Update ANN_BENCH to use C++20

achirkin:


tarang-jain:
Faiss and DiskANN wrappers use C++17, so this might not work. Also, for the C API part, #1795 needs to be merged into this branch.

achirkin:
Good point, thanks! I see you already have https://github.com/rapidsai/cuvs/pull/1796 Maybe you could also try to change the standard in the benchmarks there too?

PR #1804, opened 2026-02-15
Introduce UDF Architecture

divyegala:
This PR introduces User-Defined-Functions supporting architecture in cuVS and uses JIT LTO to achieve it. The initial example is written for passing a metric UDF to IVF Flat search kernels.

When tested with native L2 metric and UDF L2 metric, we get native performance.
image

### DIFF
--- a/cpp/include/cuvs/neighbors/ivf_flat.hpp
+++ b/cpp/include/cuvs/neighbors/ivf_flat.hpp
@@ -3472,7 +3472,7 @@ template <typename T, typename U> inline constexpr bool is_same_v = is_same<T, U
 inline std::string instantiate_udf(char const* data_type, char const* acc_type, int veclen)
 {
   std::ostringstream oss;
-  oss << "\nnamespace cuvs { namespace neighbors { namespace ivf_flat { namespace detail {\n"
+  oss << "\nnamespace cuvs::neighbors::ivf_flat::detail {\n"
       << "template <typename AccT>\n"
       << "__device__ void compute_dist(AccT& acc, AccT x, AccT y) {\n"
       << "  compute_dist_udf_impl<" << data_type << ", " << acc_type << ", " << veclen
@@ -3480,7 +3480,7 @@ inline std::string instantiate_udf(char const* data_type, char const* acc_type, 
       << "}\n"
       << "template __device__ void compute_dist<" << acc_type << ">(" << acc_type << "&, "
       << acc_type << ", " << acc_type << ");\n"
-      << "}}}}\n";
+      << "}\n";
   return oss.str();
 }
 
@@ -3557,14 +3557,14 @@ struct )" #NAME R"( : metric_interface<T, AccT, Veclen> {                       
 )" #BODY R"(                                                                           \
 };                                                                                     \
                                                                                        \
-namespace cuvs { namespace neighbors { namespace ivf_flat { namespace detail {         \
+namespace cuvs::neighbors::ivf_flat::detail {                                        \
 template <typename T, typename AccT, int Veclen>                                       \
 __device__ __forceinline__ void compute_dist_udf_impl(AccT& acc, AccT x, AccT y)       \
 {                                                                                      \
   ::)" #NAME R"(<T, AccT, Veclen> metric;                                              \
   metric(acc, ::point<T, AccT, Veclen>(x), ::point<T, AccT, Veclen>(y));               \
 }                                                                                      \
-}}}}                                                                                   \
+}                                                                                    \
 )";                                                                                               \
     return result;                                                                                \
   }


cpp/src/detail/jit_lto/NVRTCLTOFragmentCompiler.cu
```diff
@@ -0,0 +1,89 @@
...
+                                                  opts.size(),   // numOptions
+                                                  opts.data());  // options
+
+  if (compileResult != NVRTC_SUCCESS) {
+    // Obtain compilation log from the program.
+    size_t log_size;
+    NVRTC_SAFE_CALL(nvrtcGetProgramLogSize(prog, &log_size));
+    std::unique_ptr<char[]> log{new char[log_size]};
+    NVRTC_SAFE_CALL(nvrtcGetProgramLog(prog, log.get()));
+    std::cerr << "nvrtrc compile error log: \n";
```
dantegd:
```suggestion
    std::cerr << "nvrtc compile error log: \n";
```

Critical to fix 😛

cpp/src/neighbors/ivf_flat/ivf_flat_search.cuh
```diff
@@ -377,6 +383,13 @@ void search_with_filtering(raft::resources const& handle,
   RAFT_EXPECTS(queries.extent(1) == index.dim(),
                "Number of query dimensions should equal number of dimensions in the index.");
 
+  // Save original metric and temporarily set to CustomUDF if using UDF
+  auto original_metric = index.metric();
+  if (params.metric_udf.has_value()) {
+    const_cast<std::remove_const_t<std::remove_reference_t<decltype(index)>>&>(index).set_metric(
+      cuvs::distance::DistanceType::CustomUDF);
```
dantegd: This has been in my mind for a couple of days now... so the index is logically `const` but is mutated to drive the UDF branch, then restored. It'll work fine but it has some brittle-ness to it. 

We should at least document that search may temporarily change the index’s metric when using UDF, and that callers must not rely on index.metric() during this specific search. 

I wonder if it'd be worth introducing something like an “effective metric” (e.g. `CustomUDF`) through the call stack via `search_params` or an extra parameter instead of mutating the index, so the index stays logically `const`, I can't remember if we talked about this specifically before...

cpp/src/neighbors/ivf_flat/ivf_flat_interleaved_scan_jit.cuh
```diff
@@ -153,9 +169,38 @@ void launch_kernel(const index<T, IdxT>& index,
...
+    metric_udf += type_name<AccT>();
+    metric_udf += ");\n";
+    // Include hash of UDF source in key to differentiate different UDFs
+    auto udf_hash           = std::to_string(std::hash<std::string>{}(metric_udf));
+    std::string metric_name = "metric_udf_" + udf_hash;
+    NVRTCLTOFragmentCompiler nrtc_lto_compiler;
+    std::string key =
+      metric_name + "_" + std::to_string(Veclen) + "_" +
+      make_fragment_key<decltype(get_data_type_tag<T>()), decltype(get_acc_type_tag<AccT>())>();
+    nrtc_lto_compiler.compile(key, metric_udf);
```
dantegd: This is a minor thing probably, `compile` correctly skips work when `has_fragment(key)` is true, but we still build `standard_compile_opts` (and touch the device for arch) on every UDF search if I'm not mistaken. 

We could consider caching `standard_compile_opts` (or if we want more completeness, a cached compiler instance?)

cpp/tests/neighbors/ann_ivf_flat/test_udf.cu
```diff
@@ -0,0 +1,387 @@
...
+#include <raft/core/resource/cuda_stream.hpp>
+#include <raft/core/resources.hpp>
+#include <rmm/device_uvector.hpp>
+
+#include <vector>
+
+namespace cuvs::neighbors::ivf_flat {
+
+// ============================================================================
+// Define custom metrics using the UDF macro
```
dantegd, line 26: I'm biased, but the macro looks really good in the PR 😁

cpp/include/cuvs/neighbors/ivf_flat.hpp
```diff
@@ -3047,4 +3051,424 @@ void recompute_internal_state(const raft::resources& res, index<uint8_t, int64_t
...
+
+/**
+ * @brief Define a custom distance metric with compile-time validation.
+ *
+ * This macro creates:
+ * 1. A struct that inherits from metric_interface (compile-time validation)
+ * 2. A function NAME_udf() that returns a metric_source for JIT compilation
+ *
+ * @param NAME The name of your metric (becomes struct name and function prefix)
+ * @param BODY The body of operator()(AccT& acc, point_type x, point_type y)
```
dantegd, line 3499: I'd suggest one extra sentence to clarify that the same `BODY` is stringified and compiled by NVRTC so it must be valid device code and cannot use host-only types or headers.

cpp/tests/neighbors/ann_ivf_flat/test_udf.cu
```diff
@@ -0,0 +1,387 @@
...
+// ============================================================================
+
+// Custom L2 (squared Euclidean) metric - should match built-in L2
+CUVS_METRIC(custom_l2, { acc += squared_diff(x, y); })
+
+// ============================================================================
+// Test data traits for different types
```
dantegd, line 40: Nice dataset coverage!

dantegd (commented):
Love this so much!

cpp/src/neighbors/ivf_flat/ivf_flat_search.cuh
divyegala: I dislike this part of the code too. As of now, I consider this feature unusable anyways because I don't think there is ever a scenario where you build the index with some metric but search it with another metric.

We can put it in the experimental namespace if you like that more?

cpp/src/neighbors/ivf_flat/ivf_flat_interleaved_scan_jit.cuh
divyegala: Ah yeah, the CUDA runtime call *could* be expensive (although the benchmarks don't show any of it). How about turning this into a singleton class? `compile` is almost a static function and doesn't need to know the state of the object once it has been populated.

cpp/tests/neighbors/ann_ivf_flat/test_udf.cu
divyegala: I think it looks fantastic!

cpp/include/cuvs/neighbors/ivf_flat.hpp
divyegala: Will do. Also it shouldn't include *any* headers.

cpp/src/detail/jit_lto/NVRTCLTOFragmentCompiler.cu
divyegala: Need to clean up it and remove `std::cerr` usage entirely 😁

cpp/include/cuvs/neighbors/ivf_flat.hpp
```diff
@@ -3047,4 +3051,424 @@ void recompute_internal_state(const raft::resources& res, index<uint8_t, int64_t
...
+
+#ifdef CUVS_ENABLE_JIT_LTO
+namespace udf {
+
+// ============================================================================
+// Real C++ implementations for compile-time validation
+// ============================================================================
+
+/**
+ * @brief Wrapper for vector elements that provides both packed and unpacked access.
```
jinsolp, line 3063: I think this can be future work, but from my understanding the current `point` and helper functions use scalar logic for `T=fp16`. For `Veclen>1` packed fp16 where `Acc=int32` (or whatever 4 byte acc) we might be able to exploit vectorized instructions for fp16?

cpp/src/neighbors/ivf_flat/ivf_flat_search.cuh
dantegd: yeah, experimental namespace is a great idea, it'll also pressure us to get to a final stage :)

cpp/src/neighbors/ivf_flat/ivf_flat_interleaved_scan_jit.cuh
dantegd: Singleton class would work nicely as well, either approach should be great

cpp/include/cuvs/neighbors/ivf_flat.hpp
divyegala: It already works that way from the UDF perspective! You can get the raw data from `point` and use vectorized instructions in your UDF instead of iterating through the point array.

jinsolp: I see! Thanks for clarifying!

cpp/include/cuvs/neighbors/ivf_flat.hpp
```diff
@@ -3047,4 +3051,425 @@ void recompute_internal_state(const raft::resources& res, index<uint8_t, int64_t
...
+  }
+}
+
+/** @brief Absolute difference: |x - y| */
+template <typename T, typename AccT, int V>
+__device__ __forceinline__ AccT abs_diff(point<T, AccT, V> x, point<T, AccT, V> y)
+{
+  if constexpr (std::is_same_v<T, uint8_t> && V > 1) {
+    auto diff = __vabsdiffu4(x.raw(), y.raw());
+    return ((diff >> 0) & 0xFF) + ((diff >> 8) & 0xFF) + ((diff >> 16) & 0xFF) +
```
tarang-jain: A zero bit shift is not needed, right?

cpp/CMakeLists.txt
```diff
@@ -213,6 +213,13 @@ endif()
 # ##################################################################################################
 # * cuvs ---------------------------------------------------------------------
 if(NOT BUILD_CPU_ONLY)
+  set(JIT_LTO_TARGET_ARCHITECTURE "")
+  set(JIT_LTO_COMPILATION OFF)
+  set(JIT_LTO_FILES "")
+  if(CMAKE_CUDA_COMPILER_VERSION VERSION_GREATER_EQUAL 13.0)
+    set(JIT_LTO_TARGET_ARCHITECTURE "75-real")
```
tarang-jain: Curious! Why are we setting this to a specific architecture (75-real)?

divyegala: It's just the lowest architecture RAPIDS currently supports!

cpp/include/cuvs/neighbors/ivf_flat.hpp
divyegala: Good catch. Yes, it shouldn't be needed.

dependencies.yaml
```diff
@@ -371,7 +372,7 @@ dependencies:
           - matrix:
               cuda: "12.*"
             packages:
-              - cupy-cuda12x>=13.6.0
+              - cupy-cuda12x>=13.6.0,<14.0
```
bdice: Is this necessary? Only on CUDA 12?

divyegala: There was some cupy 14 problems I was facing earlier. Let me remove and check if it's still necessary.

divyegala: Removing max pin still leads to failure in CUDA 12 tests. Looks like it has something to do with how nvrtc and cupy 14 interact in CUDA 12. Any suggestions? https://github.com/rapidsai/cuvs/actions/runs/22649431168/job/65647025645?pr=1804

bdice: I don't see anything immediately, the logs are very verbose. I would treat this as a blocker for this PR though, we need to support cupy >=14 for a variety of reasons.

bdice: @leofang suggested adding a dependency on `cuda-toolkit[...,cudart]` per the table in https://github.com/cupy/cupy/issues/9719#issuecomment-3922329852. I pushed that in 3f09b32, we'll see if it helps.

divyegala: Looks like that worked!

cpp/include/cuvs/detail/jit_lto/NVRTCLTOFragmentCompiler.hpp
```diff
@@ -0,0 +1,19 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2025-2026, NVIDIA CORPORATION.
```
mythrocks: If this is a new file, shouldn't this be only `2026`?

```suggestion
 * SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION.
```

cpp/include/cuvs/detail/jit_lto/FragmentDatabase.hpp
```diff
@@ -43,3 +48,7 @@ void registerFatbinFragment(std::string const& algo,
                             std::string const& params,
                             unsigned char const* blob,
                             std::size_t size);
+
+void registerNVRTCFragment(std::string const& key,
+                           std::unique_ptr<char[]>&& program,
```
mythrocks: I haven't gotten far in this change yet, but I wonder why `program` is not a `vector<char>` instead of a `unique_ptr<char[]>`.
Edit:  Or a `std::string`, really.

cpp/src/detail/jit_lto/NVRTCLTOFragmentCompiler.cu
```diff
@@ -0,0 +1,91 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2025-2026, NVIDIA CORPORATION.
```
mythrocks: Likewise here.  Should this only be `2026`?

```suggestion
 * SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION.
```

cpp/include/cuvs/detail/jit_lto/FragmentDatabase.hpp
divyegala: I don't think there's a particular reason. We need to use the C type char[] so it's just clearer IMO.

cpp/include/cuvs/neighbors/ivf_flat.hpp
```diff
@@ -3047,4 +3051,417 @@ void recompute_internal_state(const raft::resources& res, index<uint8_t, int64_t
...
+}
+
+/** @brief Absolute difference: |x - y| */
+template <typename T, typename AccT, int V>
+__device__ __forceinline__ AccT abs_diff(point<T, AccT, V> x, point<T, AccT, V> y)
+{
+  if constexpr (std::is_same_v<T, uint8_t> && V > 1) {
+    auto diff = __vabsdiffu4(x.raw(), y.raw());
+  } else if constexpr (std::is_same_v<T, int8_t> && V > 1) {
+    auto diff = __vabsdiffs4(x.raw(), y.raw());
```
jinsolp, line 3173: Shouldn't we be retuning the `diff` here?

cpp/include/cuvs/neighbors/ivf_flat.hpp
```diff
@@ -3047,4 +3051,417 @@ void recompute_internal_state(const raft::resources& res, index<uint8_t, int64_t
...
+/** @brief Squared difference: (x - y)² */
+template <typename T, typename AccT, int V>
+__device__ __forceinline__ AccT squared_diff(point<T, AccT, V> x, point<T, AccT, V> y)
+{
+  if constexpr (std::is_same_v<T, uint8_t> && V > 1) {
+    auto diff = __vabsdiffu4(x.raw(), y.raw());
+    return __dp4a(diff, diff, AccT{0});
+  } else if constexpr (std::is_same_v<T, int8_t> && V > 1) {
+    auto diff = __vabsdiffs4(x.raw(), y.raw());
+    return __dp4a(diff, diff, static_cast<uint32_t>(0));
```
jinsolp: I don't think this should lead to correctness issues but why is the first one using `AccT` as the accumulator and the second one `uint32_t`? Can't we do `return __dp4a(diff, diff, AccT{0});` for both?

divyegala: Yes! Nice catch.

cpp/include/cuvs/neighbors/ivf_flat.hpp
divyegala: Good catch again!

cpp/include/cuvs/neighbors/ivf_flat.hpp
divyegala: Oh we can't, because `__vabsdiffs4` returns a `uint32_t`.

jinsolp: oh okay no worries

cpp/include/cuvs/neighbors/ivf_flat.hpp
```diff
@@ -3047,4 +3051,421 @@ void recompute_internal_state(const raft::resources& res, index<uint8_t, int64_t
...
+  using element_type          = T;
+  using storage_type          = AccT;
+  static constexpr int veclen = Veclen;
+
+  storage_type data_;
+
+  point() = default;
+  __device__ __host__ explicit point(storage_type d) : data_(d) {}
+
+  __device__ __forceinline__ storage_type raw() const { return data_; }
```
achirkin, line 3091: Please use [raft macros](https://github.com/rapidsai/raft/blob/main/cpp/include/raft/core/detail/macros.hpp) for these where appropriate:
```suggestion
  RAFT_DEVICE_INLINE_FUNCTION storage_type raw() const { return data_; }
```

conda/recipes/libcuvs/recipe.yaml
```diff
@@ -75,6 +75,7 @@ cache:
       - if: cuda_major == "13"
         then:
           - libnvjitlink-dev
+          - cuda-nvrtc-dev
```
achirkin: I see the nvrtc introduction is shared between this PR and #1807 (all related changes in conda, cmake, and C++ headers). It looks to me like an important change set on its own. Could you please move it in a separate PR? This would reduce the diff of both PRs and will make the commit history in the main branch more granular (much more readable `git blame`).

cpp/include/cuvs/neighbors/ivf_flat.hpp
```diff
@@ -3047,4 +3051,421 @@ void recompute_internal_state(const raft::resources& res, index<uint8_t, int64_t
...
+ *
+ * For float/half: trivial wrapper around scalar values
+ * For int8/uint8 with Veclen > 1: wraps packed bytes in a 32-bit word
+ *
+ * @tparam T Data type (float, __half, int8_t, uint8_t)
+ * @tparam AccT Storage/accumulator type (float, __half, int32_t, uint32_t)
+ * @tparam Veclen Vector length (1, 2, 4, 8, 16)
+ */
+template <typename T, typename AccT, int Veclen>
+struct point {
```
achirkin, line 3074: You've probably considered this already, but could you give a small explanation (perhaps right in the docstring for future reference): why can't we use `TxN_t` or `IOType` from [raft/util/vectorized.cuh](https://github.com/rapidsai/raft/blob/bf258f39b670d65c60d8a681c8ec45d0ade5690f/cpp/include/raft/util/vectorized.cuh) here? Either directly or as a wrapped-in carrier type? (adding missing accessors to raft I think is a reasonable idea too)

cpp/include/cuvs/neighbors/ivf_flat.hpp
```diff
@@ -3047,4 +3051,421 @@ void recompute_internal_state(const raft::resources& res, index<uint8_t, int64_t
...
+template <typename T, typename AccT, int Veclen = 1>
+struct metric_interface {
+  using point_type = point<T, AccT, Veclen>;
+
+  virtual __device__ void operator()(AccT& acc, point_type x, point_type y) = 0;
+  virtual ~metric_interface()                                               = default;
+};
+
+// ============================================================
+// Helper Operations - Deduce Veclen from point type!
```
achirkin, line 3146: What does this relate to?

cpp/include/cuvs/neighbors/ivf_flat.hpp
```diff
@@ -3047,4 +3051,421 @@ void recompute_internal_state(const raft::resources& res, index<uint8_t, int64_t
...
+    return result;
+  } else {
+    auto a = x.raw();
+    auto b = y.raw();
+    return (a > b) ? a : b;
+  }
+}
+
+// ============================================================================
+// String versions for JIT compilation
```
achirkin, line 3236: This looks like it's going to be hard to maintain. What if we put all definitions subject to JIT in a separate header and then write a small utility module that would read that header as text at compile time and save it to strings?

cpp/src/neighbors/ivf_flat/ivf_flat_interleaved_scan.cuh
```diff
@@ -1293,6 +1293,7 @@ struct select_interleaved_scan_kernel {
  */
 template <typename T, typename AccT, typename IdxT, typename IvfSampleFilterT>
 void ivfflat_interleaved_scan(const index<T, IdxT>& index,
+                              const search_params& params,
```
achirkin: I think, if we pass `search_params` here, we can drop some other arguments for being duplicates, right? Probably these two: `n_probes`, `metric`.

cpp/src/detail/jit_lto/NVRTCLTOFragmentCompiler.cu
```diff
@@ -0,0 +1,91 @@
...
+  }
+
+NVRTCLTOFragmentCompiler::NVRTCLTOFragmentCompiler()
+{
+  int device = 0;
+  int major  = 0;
+  int minor  = 0;
+  cudaGetDevice(&device);
+  cudaDeviceGetAttribute(&major, cudaDevAttrComputeCapabilityMajor, device);
+  cudaDeviceGetAttribute(&minor, cudaDevAttrComputeCapabilityMinor, device);
```
achirkin: Let's have these wrapped in `RAFT_CUDA_TRY`  (but also my suggestion to move this to a separate PR stands).

cpp/src/neighbors/ivf_flat/ivf_flat_interleaved_scan_jit.cuh
```diff
@@ -153,9 +169,38 @@ void launch_kernel(const index<T, IdxT>& index,
...
+    metric_udf += type_name<T>();
+    metric_udf += ", ";
+    metric_udf += type_name<AccT>();
+    metric_udf += ">(";
+    metric_udf += type_name<AccT>();
+    metric_udf += "&, ";
+    metric_udf += type_name<AccT>();
+    metric_udf += ", ";
+    metric_udf += type_name<AccT>();
+    metric_udf += ");\n";
```
achirkin: Since we're C++20 now, I think using `std::format` is preferable for being more concise:
```C++
metric_udf += fmt::format(
    "\ntemplate void cuvs::neighbors::ivf_flat::detail::compute_dist"
    "<{}, {}, {}>({}&, {}, {});\n",
    Veclen, type_name<T>(), type_name<AccT>(),
    type_name<AccT>(), type_name<AccT>(), type_name<AccT>());
```

divyegala:
Thanks for your review @achirkin!

> First, I think the enabling-nvrtc part can be split into a separate PR. Can it?

I would prefer not to for a few reasons:
1. No way to test the enabling-nvrtc part. This PR helps establish an e2e test through our algorithms.
2. It has been tough to find timely reviews for my JIT LTO/UDF work, so I would like to maintain momentum on this PR while it has reviewers.

> Second, the JIT code generation

I will attempt to answer all your questions related to the input at the same time.

At the present moment, we only have two options: 
1. Pass a string to nvrtc -> generate LTO-IR -> link with nvjitlink
2. User passes LTO-IR -> link with nvjiitlink (we can enable this feature in a follow-up PR, I consider this a feature that would require some savviness from the user themselves)

Even though it appears that we are supporting two methods (header or string), we are actually only supporting method 1. The header is simply a convenience around avoiding using raw strings and letting our users figure out at compile time if their UDF has the right interface to work in our kernels.

Creating any other method to parse the UDF would involve a significant investment from our end. That said, we have other teams already working on a better UX - please check out this [nvbug](https://nvbugspro.nvidia.com/bug/5835194).

Keeping efforts of other teams in mind, I consider the current macro-style type checker to be a fair middle ground and low effort bar for us to help users enforce some compile time compatibility checks while we await further developments.

cpp/include/cuvs/neighbors/ivf_flat.hpp
divyegala: That's a little hard and would need us to plug it in with CMake, so that we can embed that generated string back into this header. That said, I do think that it will declutter this header nicely. Would you be okay with a follow-up?

cpp/include/cuvs/neighbors/ivf_flat.hpp
divyegala: It is so that the user does not have to know the value of `Veclen` to write their UDF, it is just a convenience class.

cpp/include/cuvs/neighbors/ivf_flat.hpp
divyegala: These are pre-existing template types used by `interleaved_scan_kernel` to express its internal distance computation functions. 

I would say it is orthogonal to this PR to update these types. Any decision to refactor `interleaved_scan_kernel` should come separately from the JIT/LTO feature.

cpp/include/cuvs/neighbors/ivf_flat.hpp
divyegala: We cannot use those here. We are not allowed to include any headers in the UDF strings.

cpp/src/neighbors/ivf_flat/ivf_flat_interleaved_scan_jit.cuh
divyegala: I don't think we have `fmt` available?

cpp/include/cuvs/neighbors/ivf_flat.hpp
achirkin: > It is so that the user does not have to know the value of Veclen to write their UDF, it is just a convenience class.

Could you please write just that in the file? When I stumbled upon this comment, it looked like an instruction to me or some sort of todo note :)

cpp/include/cuvs/neighbors/ivf_flat.hpp
achirkin: Ah sorry I didn't look at ivf-flat code for a while and missed when this structure was introduced

cpp/include/cuvs/neighbors/ivf_flat.hpp
```diff
@@ -3047,4 +3051,421 @@ void recompute_internal_state(const raft::resources& res, index<uint8_t, int64_t
  */
 
 }  // namespace helpers
+
+#ifdef CUVS_ENABLE_JIT_LTO
+namespace experimental::udf {
+
+// ============================================================================
+// Real C++ implementations for compile-time validation
```
achirkin: Could you please clarify how and what these validate at compile time? In the code commentary.

cpp/include/cuvs/neighbors/ivf_flat.hpp
achirkin: Please write this in bold everywhere where it's relevant. To an unprepared viewer like me, there's no indication if this is the case and why. Another developer will see this with no commentary whatsoever and write the same style outside of the nvrtc/udf context.
This is becoming especially a problem in light of AI coding agents who learn the style from the existing project code and replicate it not-so-thoughtfully.

divyegala:
@achirkin 

> Is it really not possible to make a simple nvrtc test with a dummy kernel in our gtests? This is concerning.

Hmm, it should be actually. We would have to embed a test kernel in `libcuvs.so` though, which I don't feel great about adding testing components to the main library.

> On the topic of nvrtc strings.

I like your proposed solution, and it definitely reads cleaner and more maintainable. It will take some time to figure out the right design to land on, which is why this feature is in the `experimental` namespace :). Can you please create an issue describing your idea? We can iterate on the design there and update in 26.06.

cpp/include/cuvs/detail/jit_lto/AlgorithmPlanner.hpp
```diff
@@ -5,29 +5,33 @@
...
 
   std::string entrypoint;
-  std::vector<const FragmentEntry*> fragments;
+  std::vector<std::unique_ptr<FragmentEntry>> fragments;
 
-  void add_fragment(const FragmentEntry& fragment);
+  template <typename T, typename = std::enable_if_t<std::is_convertible_v<T*, FragmentEntry*>>>
+  void add_fragment(std::unique_ptr<T> fragment)
+  {
+    fragments.push_back(std::unique_ptr<FragmentEntry>(std::move(fragment)));
```
KyleFromNVIDIA, line 38: Please split the dangling pointer fix into its own PR so it's not blocked by this one.

divyegala: Done https://github.com/rapidsai/cuvs/pull/1988

cpp/src/neighbors/ivf_flat/ivf_flat_search.cuh
```diff
@@ -383,6 +383,13 @@ void search_with_filtering(raft::resources const& handle,
...
```
dantegd: Had a few more thoughts here, if two threads search the same index concurrently with different metrics (one built-in, one UDF), they will trample each other's metric. 

Also, the save/restore pattern doesn't seem to be  exception-safe either, right? So if `search_with_filtering` throws, the metric is never restored

cpp/src/neighbors/ivf_flat_index.cpp
```diff
@@ -60,6 +60,12 @@ cuvs::distance::DistanceType index<T, IdxT>::metric() const noexcept
   return metric_;
 }
 
+template <typename T, typename IdxT>
+void index<T, IdxT>::set_metric(cuvs::distance::DistanceType metric)
+{
+  metric_ = metric;
+}
```
dantegd: This exists only to support the const_cast workaround in ivf_flat_search.cuh, no?  If so, it would be good to add a comment that if that workaround is removed (by threading the effective metric through the call chain), this setter should be removed too.

cpp/src/detail/jit_lto/NVRTCLTOFragmentCompiler.cpp
```diff
@@ -0,0 +1,94 @@
...
+
+  cache[key] = std::move(lto_ir);
+  return std::make_unique<UDFFatbinFragment>(key, cache[key]);
+}
+
+NVRTCLTOFragmentCompiler& nvrtc_compiler()
+{
+  static NVRTCLTOFragmentCompiler compiler;
+  return compiler;
+}
```
dantegd, line 108: NVRTCLTOFragmentCompiler is a process-wide singleton with a mutable std::unordered_map cache, no?  So concurrent calls to compile() from different threads will race on the cache reads and writes. It would be a good idea to use a std::shared_mutex (or std::mutex)

cpp/src/detail/jit_lto/NVRTCLTOFragmentCompiler.cpp
```diff
@@ -0,0 +1,94 @@
...
+                                                  opts.size(),   // numOptions
+                                                  opts.data());  // options
+
+  if (compileResult != NVRTC_SUCCESS) {
+    // Obtain compilation log from the program.
+    size_t log_size;
+    NVRTC_SAFE_CALL(nvrtcGetProgramLogSize(prog, &log_size));
+    std::unique_ptr<char[]> log{new char[log_size]};
+    NVRTC_SAFE_CALL(nvrtcGetProgramLog(prog, log.get()));
+    RAFT_FAIL("nvrtc compile error log: \n%s", log.get());
```
dantegd: I think there is a small leak here, if there is a compilation failure, then RAFT_FAIL throws an exception, so `nvrtcDestroyProgram` (which is only on the success path) is never called. Then the nvrtcProgram handle leaks. We could use an RAII wrapper or a try/catch block that destroys the program before re-throwing.

cpp/src/neighbors/ivf_flat/ivf_flat_interleaved_scan_jit.cuh
```diff
@@ -122,11 +138,27 @@ void launch_kernel(const index<T, IdxT>& index,
...
+                 "CustomUDF search requires search_params.metric_udf");
+    std::string metric_udf = params.metric_udf.value();
+    std::ostringstream oss;
+    oss << "\ntemplate void cuvs::neighbors::ivf_flat::detail::compute_dist<" << Veclen << ", "
+        << type_name<T>() << ", " << type_name<AccT>() << ">(" << type_name<AccT>() << "&, "
+        << type_name<AccT>() << ", " << type_name<AccT>() << ");\n";
+    metric_udf += oss.str();
+    auto udf_hash                   = std::to_string(std::hash<std::string>{}(metric_udf));
+    std::string const udf_cache_key = "metric_udf_" + udf_hash;
+    auto udf_fragment               = nvrtc_compiler().compile(udf_cache_key, metric_udf);
```
dantegd: I think we have a potentially silent and obscure bug here. If two different UDF strings hash to the same size_t value will return the wrong cached LTO IR, which silently would produce incorrect results. I wonder if we should use the full UDF source string as the cache key, or use a cryptographic/content-addressing hash (e.g., SHA-256). Given that UDF strings are typically small (< 1KB), using the full string as the key might perfectly fine?

cpp/src/neighbors/ivf_flat/ivf_flat_interleaved_scan_jit.cuh
```diff
@@ -260,6 +292,17 @@ void launch_with_fixed_consts(cuvs::distance::DistanceType metric, Args&&... arg
...
+      return launch_kernel<Capacity,
+                           Veclen,
+                           Ascending,
+                           false,
+                           T,
+                           AccT,
+                           IdxT,
+                           IvfSampleFilterTag,
+                           tag_metric_custom_udf<Veclen, T, AccT>,
+                           tag_post_identity>(std::forward<Args>(args)...);
```
dantegd: Have been thinking about this for a bit, UDFs are hardcoded to `Ascending` (min-close), `ComputeNorm = false`, and `tag_post_identity`. 

This means users cannot write similarity-based UDFs (which need max-close / select-max) or UDFs that need sqrt post-processing. We could consider adding a select_min and/or post_processing option to search_params when using UDFs, or at minimum document these constraints.

cpp/src/neighbors/refine/refine_device.cuh
```diff
@@ -101,18 +101,18 @@ void refine_device(
 
   using acc_t = typename cuvs::spatial::knn::detail::utils::config<data_t>::value_t;
 
+  cuvs::neighbors::ivf_flat::search_params refine_params;
+  refine_params.n_probes = 1;
```
dantegd: This is a small nitpick. 

Changing three individual parameters (metric, n_probes, select_min) with the full search_params struct below forces us to construct a synthetic search_params just to pass n_probes = 1. This suggests the internal scan function should perhaps take both a search_params (for UDF state) and still accept the pre-computed metric/n_probes/select_min from its callers. Or, pass only what's needed (a lightweight struct containing UDF state plus the resolved metric/n_probes/select_min).

cpp/src/detail/jit_lto/NVRTCLTOFragmentCompiler.cpp
```diff
@@ -0,0 +1,94 @@
...
+  // Use actual GPU architecture for optimal code generation
+  this->standard_compile_opts[i++] =
+    std::string{"-arch=sm_" + std::to_string((major * 10 + minor))};
+  this->standard_compile_opts[i++] = std::string{"-dlto"};
+  this->standard_compile_opts[i++] = std::string{"-rdc=true"};
+  this->standard_compile_opts[i++] = std::string{"-default-device"};
+  this->standard_compile_opts[i++] = std::string{"--gen-opt-lto"};
+  // Optimization flags - NVRTC uses different syntax than nvcc
+  this->standard_compile_opts[i++] = std::string{"--use_fast_math"};
+  this->standard_compile_opts[i++] = std::string{"--extra-device-vectorization"};
```
dantegd: Couldn't we use `push_back()` or use brace initialization to make this code a bit cleaner? 


```suggestion
this->standard_compile_opts = {
  "-arch=sm_" + std::to_string(major * 10 + minor),
  "-dlto",
  "-rdc=true",
  "-default-device",
  "--gen-opt-lto",
  "--use_fast_math",
  "--extra-device-vectorization"
};
```

cpp/src/neighbors/ivf_flat/ivf_flat_interleaved_scan_jit.cuh
```diff
@@ -52,6 +54,19 @@ constexpr auto get_idx_type_tag()
...
+constexpr const char* type_name()
+{
+  if constexpr (std::is_same_v<T, float>) { return "float"; }
+  if constexpr (std::is_same_v<T, __half>) { return "__half"; }
+  if constexpr (std::is_same_v<T, int8_t>) { return "int8_t"; }
+  if constexpr (std::is_same_v<T, uint8_t>) { return "uint8_t"; }
+  if constexpr (std::is_same_v<T, int32_t>) { return "int32_t"; }
+  if constexpr (std::is_same_v<T, uint32_t>) { return "uint32_t"; }
+  if constexpr (std::is_same_v<T, int64_t>) { return "int64_t"; }
+}
```
dantegd, line 92: If T doesn't match any of these, the function falls off the end without returning a value, which is undefined behavior. Add a `static_assert(false, ...)` or a final-else branch.

cpp/tests/neighbors/ann_ivf_flat/test_udf.cu
```diff
@@ -0,0 +1,387 @@
...
+#include <vector>
+
+namespace cuvs::neighbors::ivf_flat {
+
+// ============================================================================
+// Define custom metrics using the UDF macro
+// ============================================================================
+
+// Custom L2 (squared Euclidean) metric - should match built-in L2
+CUVS_METRIC(custom_l2, { acc += squared_diff(x, y); })
```
dantegd, line 30: After looking again here, hadn't though thtat the test only covers float and int8_t with a custom L2 metric. We could consider adding:

- A genuinely novel metric (e.g., Chebyshev/L-infinity) that differs from any built-in, to prove the UDF produces correct results
- An error case test (bad CUDA code should give a clear error, not a crash)
- uint8_t and half types
- Testing that the cache actually works (second search with same UDF should be faster)

cpp/src/detail/jit_lto/NVRTCLTOFragmentCompiler.cpp
```diff
@@ -0,0 +1,94 @@
...
+  std::size_t i = 0;
+  // Use actual GPU architecture for optimal code generation
+  this->standard_compile_opts[i++] =
+    std::string{"-arch=sm_" + std::to_string((major * 10 + minor))};
+  this->standard_compile_opts[i++] = std::string{"-dlto"};
+  this->standard_compile_opts[i++] = std::string{"-rdc=true"};
+  this->standard_compile_opts[i++] = std::string{"-default-device"};
+  this->standard_compile_opts[i++] = std::string{"--gen-opt-lto"};
+  // Optimization flags - NVRTC uses different syntax than nvcc
+  this->standard_compile_opts[i++] = std::string{"--use_fast_math"};
```
dantegd: We should document that --use-fast-math is used by default, and I wonder if we should make it configurable by users? Some users coul dbe surprised by the semantic changes this sometimes causes.

cpp/src/neighbors/ivf_flat/ivf_flat_search.cuh
divyegala: I do not think that is an acceptable pattern, because you technically are not allowed to search the index with different metrics than the one it was built with. This PR just works around that to prove that UDFs work.

cpp/src/detail/jit_lto/NVRTCLTOFragmentCompiler.cpp
divyegala: Actually, most of these are unnecessary. I forgot to remove them. I do not think we should let these be configurable by users yet, as we want to mimic our `nvcc` build time flags only.

cpp/src/neighbors/ivf_flat/ivf_flat_interleaved_scan_jit.cuh
divyegala: I will add that constraint to the docs for now.

cpp/src/detail/jit_lto/NVRTCLTOFragmentCompiler.cpp
divyegala: I mean, that is the problem with exceptions and pointers in general right? Do we ever guarantee cleanup of all pointers at the point of failure? We always just let the OS clean up the program.

KyleFromNVIDIA: Exceptions can be caught. This is why RAII exists.

divyegala: My point is that this is not the only place in the codebase where we throw an uncaught expression. I'm happy enough to make this change here, but we leak throughout the codebase.

cpp/src/neighbors/refine/refine_device.cuh
divyegala: This conflicts with a previous review from @achirkin. I do agree that `interleaved_scan` should not be coupled with `ivf_flat::search_params` in an ideal case. I will make the change.

dantegd: to be fair this is a small thing without an objectively ideal solution, so it's not surprising to have differing reviews:  :P

cpp/tests/neighbors/ann_ivf_flat/test_udf.cu
divyegala: I do not know how to write case 1, as the `build` cannot use UDFs. 

And for case 2, how do you verify that a clear error was received?

cpp/src/neighbors/refine/refine_device.cuh
divyegala: Now I made the change again, to remove search_params (because that's what I believe in).

achirkin: Out of curiosity, with which my review does this conflict? :) probably I miss the context

cpp/src/neighbors/ivf_flat/ivf_flat_search.cuh
dantegd: You are correct, I was just thinking of thread safety and not necessarily of correctness of the search. That said, the fix would simple and not change the architecture: compute the effective DistanceType as a local variable at the top of the function and pass it down, instead of mutating the index. Something like:

```C++
auto effective_metric = params.metric_udf.has_value()
    ? cuvs::distance::DistanceType::CustomUDF
    : index.metric();
```

Then pass effective_metric through to the places that currently read index.metric(). No const_cast, no mutation, no UB, no race, no exception-safety issue. The rest of the PR stays the same.

Since this is explicitly a works around that to prove UDFs work, this seems like the right time to do it cleanly before the pattern gets copied to other algorithms. What do you think?

cpp/tests/neighbors/ann_ivf_flat/test_udf.cu
dantegd: For case 1 that's true, but was thinking about search, not build. A test could look like:

- Build the index normally with L2 (no UDF needed at build time)
- Search with a Chebyshev UDF (e.g., acc = max(acc, abs(x - y)); with acc initialized to 0)
- To verify correctness, do a brute-force search on CPU using the same Chebyshev metric, and compare results

The tricky part is that the index clusters were built with L2, so recall won't be optimal for Chebyshev. But you can work around this by using n_probes = n_lists (search all clusters), which should guarantees you scan the entire dataset, making it equivalent to brute force. At that point, the returned distances should match a CPU brute-force Chebyshev computation within tolerance.

The value of this test is that a custom L2 UDF can accidentally "pass" even if the UDF plumbing is partially broken (since the native L2 path might help over bugs). A genuinely different metric is the only way to prove the UDF code is actually being executed and producing correct results.

For case 2, couldn't we do something like

```C++
// Bad CUDA code, should produce a clear NVRTC compilation error, not a segfault
std::string bad_udf = "this is not valid CUDA;";
search_params.metric_udf = bad_udf;
EXPECT_THROW(cuvs::neighbors::ivf_flat::search(...), raft::exception);
```

You could also optionally catch the exception and verify the message contains something like "nvrtc compile error" to confirm the error is descriptive.

cpp/src/neighbors/ivf_flat/ivf_flat_search.cuh
divyegala: This is so much cleaner, I am jealous I didn't think of it.

cpp/tests/neighbors/ann_ivf_flat/test_udf.cu
divyegala: Done! All great ideas, thanks a lot.

cpp/src/neighbors/ivf_flat/ivf_flat_interleaved_scan_jit.cuh
```diff
@@ -54,6 +58,39 @@ constexpr auto get_idx_type_tag()
...
+  } else if constexpr (std::is_same_v<U, int32_t> ||
+                       (std::is_same_v<U, long> && sizeof(long) == sizeof(int32_t))) {
+    return "int32_t";
+  } else if constexpr (std::is_same_v<U, uint32_t> || std::is_same_v<U, unsigned int> ||
+                       (std::is_same_v<U, unsigned long> && sizeof(unsigned long) == 4)) {
+    return "uint32_t";
+  } else if constexpr (std::is_same_v<U, int64_t>) {
+    return "int64_t";
+  } else {
+    static_assert(type_name_always_false_v<U>, "Unsupported type to create UDF");
```
KyleFromNVIDIA, line 89: Ugh, I wish C++ would allow you to `static_assert(false)` and only have it trigger when that `constexpr` branch is actually hit.

cpp/include/cuvs/neighbors/ivf_flat.hpp
```diff
@@ -3047,4 +3051,524 @@ void recompute_internal_state(const raft::resources& res, index<uint8_t, int64_t
...
+)";
+
+/**
+ * NVRTC/LTO fragment appended to custom metric UDF source. The UDF program is compiled alone, so it
+ * must supply the `compute_dist` definition that device_functions.cuh only declares.
+ */
+inline std::string instantiate_udf(char const* data_type, char const* acc_type, int veclen)
+{
+  std::ostringstream oss;
+  oss << "\nnamespace cuvs { namespace neighbors { namespace ivf_flat { namespace detail {\n"
```
KyleFromNVIDIA: Aren't we using C++20? Couldn't this be flattened into `namespace cuvs::neighbors::ivf_flat::detail`?

divyegala: Oh yes, and I think C++17 already had the nicer namespace convention.

cpp/src/neighbors/ivf_flat/ivf_flat_interleaved_scan_jit.cuh
divyegala: Honestly I was surprised too, I went with `static_assert(false)` to start with.

PR #1806, opened 2026-02-15
Use Specific CCCL Includes

divyegala:
This is needed downstream because cub injects an EmptyKernel symbol into every TU that includes either cub/cub.cuh or cub/device/* includes, and this causes an issue for cubins created using JIT-LTO.

This PR also fixes 2 bugs in JIT kernels: 
1. Filters out `cub::EmptyKernel` from JIT TUs specifically because they only expect 1 kernel but are finding 2 instead
2. Keeps a reference to `cudaLibrary_t` that loads the JIT kernel and unloading it at destruction

### DIFF
--- a/cpp/src/detail/jit_lto/AlgorithmLauncher.cu
+++ b/cpp/src/detail/jit_lto/AlgorithmLauncher.cu
@@ -9,7 +9,10 @@
 
 AlgorithmLauncher::AlgorithmLauncher(cudaKernel_t k, cudaLibrary_t lib) : kernel{k}, library{lib} {}
 
-AlgorithmLauncher::~AlgorithmLauncher() { RAFT_CUDA_TRY(cudaLibraryUnload(library)); }
+AlgorithmLauncher::~AlgorithmLauncher()
+{
+  if (library != nullptr) { (void)cudaLibraryUnload(library); }
+}
 
 AlgorithmLauncher::AlgorithmLauncher(AlgorithmLauncher&& other) noexcept
   : kernel{other.kernel}, library{other.library}
--- a/cpp/src/detail/jit_lto/AlgorithmPlanner.cu
+++ b/cpp/src/detail/jit_lto/AlgorithmPlanner.cu
@@ -139,7 +139,7 @@ std::shared_ptr<AlgorithmLauncher> AlgorithmPlanner::build()
     }
 
     // Only keep the kernel if it's not EmptyKernel
-    if (!is_empty_kernel) { valid_kernels.push_back(kernels.release()[i]); }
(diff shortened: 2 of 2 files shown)


cpp/src/detail/jit_lto/AlgorithmPlanner.cu
```diff
@@ -106,15 +106,44 @@ std::shared_ptr<AlgorithmLauncher> AlgorithmPlanner::build()
...
+      std::string kernel_name(func_name);
+      // Check if this is EmptyKernel
+      if (kernel_name.find(empty_kernel_name) != std::string::npos ||
+          kernel_name == empty_kernel_name) {
+        is_empty_kernel = true;
+      }
+    }
+
+    // Only keep the kernel if it's not EmptyKernel
+    if (!is_empty_kernel) { valid_kernels.push_back(kernels.release()[i]); }
```
dantegd:
```suggestion
    if (!is_empty_kernel) { valid_kernels.push_back(kernels[i]); }
```

In the first iteration (e.g. i == 0) this transfers ownership and sets kernels to nullptr. On the next iteration, kernels.release() returns nullptr, and nullptr[i] is undefined behavior. The unique_ptr can keep owning the array until the function returns; you only need to copy the cudaKernel_t handles into valid_kernels, no?

cpp/src/detail/jit_lto/AlgorithmLauncher.cu
```diff
@@ -7,7 +7,29 @@
 
 #include <raft/util/cuda_rt_essentials.hpp>
 
-AlgorithmLauncher::AlgorithmLauncher(cudaKernel_t k) : kernel{k} {}
+AlgorithmLauncher::AlgorithmLauncher(cudaKernel_t k, cudaLibrary_t lib) : kernel{k}, library{lib} {}
+
+AlgorithmLauncher::~AlgorithmLauncher() { RAFT_CUDA_TRY(cudaLibraryUnload(library)); }
```
dantegd:
```suggestion
AlgorithmLauncher::~AlgorithmLauncher() {
  if (library != nullptr) {
    (void)cudaLibraryUnload(library);  // ignore errors in destructor
  }
}
```

This might be one of those places where I'm rusty about C++ destructors, isn't throwing from destructors during stack unwinding undefined behavior?

divyegala: Ugh yeah. I'm surprised the compiler did not complain.

cpp/src/detail/jit_lto/AlgorithmPlanner.cu
divyegala: Good point, probably why the tests are failing.

PR #1808, opened 2026-02-16
PCA preprocessor

### LINKED ISSUE #1207: [FEA] Implement `cuvs::preprocessing::pca`
We have been asked for PCA preprocessing from several customers and partners (including Google vertex vector search). cuML conains a PCA impl that is fairly straightforward and fully composed of RAFT primitives. We should pull the PCA impl into RAFT and expose it through cuVS C++ APIs to satisfy these users.

aamijar:
Resolves #1207. Depends on https://github.com/rapidsai/raft/pull/2952
This PR introduces the `cuvs::preprocessing::pca` with `float` support. The following APIs are supported:
`fit`, `transform`, `fit_transform`, `inverse_transform`.

### DIFF
--- a/cpp/src/detail/jit_lto/AlgorithmLauncher.cpp
+++ b/cpp/src/detail/jit_lto/AlgorithmLauncher.cpp
@@ -37,17 +37,12 @@ AlgorithmLauncher& AlgorithmLauncher::operator=(AlgorithmLauncher&& other) noexc
 void AlgorithmLauncher::call(
   cudaStream_t stream, dim3 grid, dim3 block, std::size_t shared_mem, void** kernel_args)
 {
-  cudaLaunchAttribute attribute[1];
-  attribute[0].id = cudaLaunchAttributeProgrammaticStreamSerialization;
-  attribute[0].val.programmaticStreamSerializationAllowed = 1;
-
   cudaLaunchConfig_t config;
   config.gridDim          = grid;
   config.blockDim         = block;
   config.stream           = stream;
-  config.attrs            = attribute;
-  config.numAttrs         = 1;
   config.dynamicSmemBytes = shared_mem;
+  config.numAttrs         = 0;
 
   RAFT_CUDA_TRY(cudaLaunchKernelExC(&config, kernel, kernel_args));
 }


cpp/src/preprocessing/pca/detail/pca.cuh
```diff
@@ -0,0 +1,114 @@
...
+inline auto to_raft_params(params config, std::size_t n_rows, std::size_t n_cols)
+  -> raft::linalg::paramsPCA
+{
+  raft::linalg::paramsPCA prms;
+  prms.n_rows       = n_rows;
+  prms.n_cols       = n_cols;
+  prms.n_components = static_cast<std::size_t>(config.n_components);
+  prms.algorithm    = static_cast<raft::linalg::solver>(static_cast<int>(config.algorithm));
+  prms.tol          = config.tol;
+  prms.n_iterations = static_cast<std::uint32_t>(config.n_iterations);
```
aamijar: These casts should be removed. Need to clean up some things on the raft side.

cpp/include/cuvs/preprocessing/pca.hpp
```diff
@@ -0,0 +1,238 @@
...
+         params config,
+         raft::device_matrix_view<float, int64_t, raft::col_major> input,
+         raft::device_matrix_view<float, int64_t, raft::col_major> components,
+         raft::device_vector_view<float, int64_t> explained_var,
+         raft::device_vector_view<float, int64_t> explained_var_ratio,
+         raft::device_vector_view<float, int64_t> singular_vals,
+         raft::device_vector_view<float, int64_t> mu,
+         raft::device_scalar_view<float, int64_t> noise_vars,
+         bool flip_signs_based_on_U = false);
+
```
aamijar, line 109: Making a note here that I don't think the existing cuml implementation has the ability to tune the percentage of explained variance.
For example, in sklearn we can set `0 < n_components < 1` where the user can select a percentage of the explained variance to recover and the `n_components` is automatically determined by the algorithm in order to satisfy that.

We will have to build that piece out since it doesn't exist in the current implementation.

cpp/tests/preprocessing/pca.cu
```diff
@@ -0,0 +1,337 @@
...
+                          cuvs::CompareApprox<float>(params.tolerance),
+                          raft::resource::get_cuda_stream(handle)));
+}
+
+typedef PcaTest<double> PcaTestDataVecSmallD;
+TEST_P(PcaTestDataVecSmallD, Result)
+{
+  ASSERT_TRUE(devArrMatch(data.data(),
+                          data_back.data(),
+                          (params.n_col * params.n_col),
```
dantegd:
```suggestion
                          (params.n_row * params.n_col),
```

This took me a while to think through, I think we are checking a subset of the elements of the array with `n_col * n_col`

cpp/tests/preprocessing/pca.cu
```diff
@@ -0,0 +1,337 @@
...
+                          cuvs::CompareApprox<double>(params.tolerance),
+                          raft::resource::get_cuda_stream(handle)));
+}
+
+typedef PcaTest<float> PcaTestDataVecSmallF;
+TEST_P(PcaTestDataVecSmallF, Result)
+{
+  ASSERT_TRUE(devArrMatch(data.data(),
+                          data_back.data(),
+                          (params.n_col * params.n_col),
```
dantegd:
```suggestion
                          (params.n_row * params.n_col),
```

Same as above

cpp/tests/preprocessing/pca.cu
```diff
@@ -0,0 +1,337 @@
...
+                          cuvs::CompareApprox<double>(params.tolerance),
+                          raft::resource::get_cuda_stream(handle)));
+}
+
+typedef PcaTest<float> PcaTestDataVecF;
+TEST_P(PcaTestDataVecF, Result)
+{
+  ASSERT_TRUE(devArrMatch(data2.data(),
+                          data2_back.data(),
+                          (params.n_col2 * params.n_col2),
```
dantegd:
```suggestion
                          (params.n_row2 * params.n_col2),
```

cpp/tests/preprocessing/pca.cu
```diff
@@ -0,0 +1,337 @@
...
+                          cuvs::CompareApprox<float>(params.tolerance),
+                          raft::resource::get_cuda_stream(handle)));
+}
+
+typedef PcaTest<double> PcaTestDataVecD;
+TEST_P(PcaTestDataVecD, Result)
+{
+  ASSERT_TRUE(devArrMatch(data2.data(),
+                          data2_back.data(),
+                          (params.n_col2 * params.n_col2),
```
dantegd:
```suggestion
                          (params.n_row2 * params.n_col2),
```

cpp/src/preprocessing/pca/detail/pca.cuh
```diff
@@ -0,0 +1,113 @@
...
+  raft::linalg::paramsPCA prms;
+  prms.n_rows       = n_rows;
+  prms.n_cols       = n_cols;
+  prms.n_components = config.n_components;
+  prms.algorithm    = config.algorithm;
+  prms.tol          = config.tol;
+  prms.n_iterations = config.n_iterations;
+  prms.copy         = config.copy;
+  prms.whiten       = config.whiten;
+  return prms;
```
dantegd, line 27:
```suggestion
  prms.copy         = config.copy;
  prms.whiten       = config.whiten;
  prms.verbose      = config.verbose;  
  return prms;
```

We are missing verbose here, no?

cpp/tests/preprocessing/pca.cu
```diff
@@ -0,0 +1,337 @@
...
+    std::vector<T> explained_vars_ref_h = {6.338, 0.3287};
+    explained_vars_ref_h.resize(params.n_col);
+
+    raft::update_device(components_ref.data(), components_ref_h.data(), len_comp, stream);
+    raft::update_device(
+      explained_vars_ref.data(), explained_vars_ref_h.data(), params.n_col, stream);
+
+    cuvs::preprocessing::pca::params prms;
+    // prms.n_cols       = params.n_col;
+    // prms.n_rows       = params.n_row;
```
dantegd: Remove commented code

cpp/include/cuvs/preprocessing/pca.hpp
```diff
@@ -0,0 +1,227 @@
...
+ * @param[out] components principal components [n_components x n_cols] (col-major)
+ * @param[out] explained_var explained variances [n_components]
+ * @param[out] explained_var_ratio explained variance ratios [n_components]
+ * @param[out] singular_vals singular values [n_components]
+ * @param[out] mu column means [n_cols]
+ * @param[out] noise_vars noise variance (scalar)
+ * @param[in] flip_signs_based_on_U whether to determine signs by U (true) or V.T (false)
+ */
+void fit(raft::resources const& handle,
+         params config,
```
dantegd: Small thing but important for API consistency: `params config` is passed by value here and in all other functions in the PR. The cuVS convention afaik is `const params&` for example in `kmeans::fit`:

```C++
void fit(raft::resources const& handle,
         const cuvs::cluster::kmeans::params& params,
         raft::device_matrix_view<const float, int> X,
         std::optional<raft::device_vector_view<const float, int>> sample_weight,
         raft::device_matrix_view<float, int> centroids,
         raft::host_scalar_view<float> inertia,
         raft::host_scalar_view<int> n_iter);
```

This won't affect performance or correctness at all, but for consistency I would suggest changing to `const params& config` throughout. Applies to all 8 overloads in this header plus the detail layer.

cpp/tests/preprocessing/pca.cu
```diff
@@ -0,0 +1,337 @@
...
+      data_back(params.len, stream),
+      data2(params.len2, stream),
+      data2_back(params.len2, stream)
+  {
+    basicTest();
+    advancedTest();
+  }
+
+ protected:
+  void basicTest()
```
dantegd: Both `basicTest` and `advancedTest` set n_components = n_col, so the inverse transform should perfectly reconstruct the input. 

Consider adding a test with `n_components < n_col` that verifies the reconstruction error is bounded but non-zero, this would confirm the dimensionality reduction is actually working, not just passing data through unchanged and be a useful test case.

cpp/include/cuvs/preprocessing/pca.hpp
cjnolet: Tuning is not what's being asked for. Exposing the explained variance is what's being requested (it is used for tuning / selecting the number of components but that's something the user does, not something we need to do).

cpp/tests/preprocessing/pca.cu
divyegala: This is a great catch!

divyegala (commented):
Why were `double` instantiations needed? Where is the code intended to be used?

aamijar:
> Why were double instantiations needed? Where is the code intended to be used?

The cuml pca python interface supports double inputs. However, cuml will use the raft api, so therefore cuvs does not need double instantiations if we think its not valuable.

cpp/src/preprocessing/pca/detail/pca.cuh
aamijar: `verbose` was a unused parameter, removed it from `pca.hpp` in https://github.com/rapidsai/cuvs/pull/1808/commits/99f32fcdcaa809f6ca1f71c4e4147c36121ddc6e

cpp/tests/preprocessing/pca.cu
aamijar: Addressed in https://github.com/rapidsai/cuvs/pull/1808/commits/074fd96e6290997ac6325600b53016d485a3dbb9

cpp/include/cuvs/preprocessing/pca.hpp
aamijar: Addressed in https://github.com/rapidsai/cuvs/pull/1808/commits/c7c52a70ce4985bd7c473b554879193cf87c3f7d

divyegala:
> The cuml pca python interface supports double inputs. However, cuml will use the raft api, so therefore cuvs does not need double instantiations if we think its not valuable.

In that case, please remove `double`.

cpp/tests/preprocessing/pca.cu
aamijar: Addressed in https://github.com/rapidsai/cuvs/pull/1808/commits/948882ce636ec4bd9555e6c36c8cb34d84e8bbe3

divyegala (approved):

PR #1810, opened 2026-02-16
Iterative CAGRA-Q

irina-resh-nvda:
### Build CAGRA on PQ datasets with Iterative CAGRA-Q

Iterative cagra graph construction using CAGRA-Q search.

This PR improves the iterative CAGRA build method by enabling PQ compression: the dataset is compressed before the iterative search starts, and CAGRA-Q is used to iteratively update the KNN graph.

This is the first time we are introducing building CAGRA on (PQ) quantized datasets directly.

This PR also adds support for C and Python APIs for creating a PQ dataset and building a cagra graph on it. We also add support to update_dataset from dense->PQ or PQ->dense. Serializing the index is also supported.

### Example Runnable workflow

```python3
import cupy as cp

from cuvs.common import make_device_pq_dataset
from cuvs.neighbors import cagra, hnsw
from cuvs.preprocessing.quantize import pq

dataset = cp.random.random((5000, 64), dtype=cp.float32)
pq_dataset = make_device_pq_dataset(
    pq.PQDatasetParams(pq_dim=32), dataset
)

index = cagra.build(
    cagra.IndexParams(build_algo="iterative_cagra_search"), pq_dataset
)
cagra.save("index.bin", index, include_dataset=False)

loaded_index = cagra.Index()
cagra.load(loaded_index, "index.bin")
cagra.update_dataset(loaded_index, dataset)

queries = dataset[:100]
cagra.search(cagra.SearchParams(), loaded_index, queries, 10)

hnsw_index = hnsw.from_cagra(hnsw.IndexParams(), loaded_index)
hnsw.search(hnsw.SearchParams(), hnsw_index, cp.asnumpy(queries), 10)
```


aamijar:
Hi @irina-resh-nvda, thanks for the contribution! I've added some commits to clean up some small things in this PR:
1. Merge main into this branch and fix merge conflicts https://github.com/NVIDIA/cuvs/pull/1810/commits/278a4f3c082f4dfda0b3ab4107227997b599cc5c
2. Run `pre-commit run --all-files` to fix style checker https://github.com/NVIDIA/cuvs/pull/1810/commits/cc6529185191cad3ecf1bcf81bb43e89c706b002
3. There were some unrelated `.clangd` and `.gitignore` changes I saw so I reverted that https://github.com/NVIDIA/cuvs/pull/1810/commits/25e6d7dedd6f5dd53835d4f144cf625b431f0aa4
4. There was an unrelated warning change in cuvs_bench so I reverted that https://github.com/NVIDIA/cuvs/pull/1810/commits/6bca27512894cfcf22a4cd5af5d3d6e658145c40
5. The tests/CMakeLists.txt has a duplicate file added so I removed that here https://github.com/NVIDIA/cuvs/pull/1810/commits/9d44b2374fc3417099e135021956374acf353640
6. There were some unrelated changes from `auto` to explicit return type which I reverted https://github.com/NVIDIA/cuvs/pull/1810/commits/73c9bbcdbe23cdd9993e730b91b080b9c6f34d0e
7. Two commented out code lines were removed https://github.com/NVIDIA/cuvs/pull/1810/commits/23cf81575eca343de0fec701cf8d83ab062a638e

cpp/tests/neighbors/ann_cagra.cuh
```diff
@@ -1598,40 +1598,40 @@ inline std::vector<AnnCagraInputs> generate_inputs()
...
+  //   {false},
+  //   {true},
+  //   {true},
+  //   {0.995},
+  //   {std::optional<float>{std::nullopt}},
+  //   {std::optional<vpq_params>{std::nullopt}},
+  //   {std::optional<bool>{std::nullopt}},
+  //   {cuvs::neighbors::MergeStrategy::MERGE_STRATEGY_PHYSICAL,
+  //    cuvs::neighbors::MergeStrategy::MERGE_STRATEGY_LOGICAL});
+  // inputs.insert(inputs.end(), inputs2.begin(), inputs2.end());
```
aamijar: Why has this test been commented out?

cpp/tests/neighbors/ann_cagra.cuh
```diff
@@ -1598,40 +1598,40 @@ inline std::vector<AnnCagraInputs> generate_inputs()
...
-    {16},                                                    // k
-    {32},                                                    // degree
-    {graph_build_algo::IVF_PQ,
-     graph_build_algo::NN_DESCENT,
+    {1000000},
+    {768},  // dim
+    {16},   // k
+    {32},   // degree
+    {       // graph_build_algo::IVF_PQ,
+     // graph_build_algo::NN_DESCENT,
```
aamijar: Same question: what is the reason for the changes here?

cpp/tests/neighbors/ann_cagra.cuh
```diff
@@ -1645,7 +1645,7 @@ inline std::vector<AnnCagraInputs> generate_inputs()
     {false},
     {true},
     {false},
-    {0.995},
+    {0.01},
```
aamijar: Why has the recall changed to 1%?

cpp/tests/neighbors/ann_cagra.cuh
```diff
@@ -1712,30 +1712,30 @@ inline std::vector<AnnCagraInputs> generate_inputs()
...
+  //   {false, true},
+  //   {false},
+  //   {true},
+  //   {0.985},
+  //   {std::optional<float>{std::nullopt}},
+  //   {std::optional<vpq_params>{std::nullopt}},
+  //   {std::optional<bool>{std::nullopt}},
+  //   {cuvs::neighbors::MergeStrategy::MERGE_STRATEGY_PHYSICAL,
+  //    cuvs::neighbors::MergeStrategy::MERGE_STRATEGY_LOGICAL});
+  // inputs.insert(inputs.end(), inputs2.begin(), inputs2.end());
```
aamijar: Same question: why was this commented out?

cpp/tests/neighbors/ann_cagra/bug_graph_smaller_than_dataset.cu
```diff
@@ -38,8 +38,8 @@ class cagra_graph_smaller_than_dataset_test : public ::testing::Test {
  protected:
   void run()
   {
-    // Create a dataset with 1000 points
-    constexpr int64_t n_dataset = 1000;
+    // Create a dataset with 10000 points
+    constexpr int64_t n_dataset = 10000;
```
aamijar: What is the reason this test file has been changed to use n_dataset = 10,000?

cpp/src/neighbors/detail/cagra/utils.hpp
```diff
@@ -301,4 +317,397 @@ void copy_with_padding(
...
+  void writeback_from_device_to_host(T* dev_ptr, size_t dst_row_offset, size_t num_rows)
+  {
+    const size_t n_elem  = num_rows * host_view_.extent(1);
+    const size_t n_bytes = n_elem * sizeof(T);
+    // use memcpy instead of raft::copy to avoid strange behavior with HMM/ATS memory
+    RAFT_CUDA_TRY(cudaMemcpyAsync(host_view_.data_handle() + dst_row_offset * host_view_.extent(1),
+                                  dev_ptr,
+                                  n_bytes,
+                                  cudaMemcpyDeviceToHost,
+                                  writeback_stream_));
```
aamijar: We should have a raft function that does this already.
> // use memcpy instead of raft::copy to avoid strange behavior with HMM/ATS memory

If there is a problem here, we should try to root fix that in raft.

cpp/src/neighbors/detail/cagra/utils.hpp
```diff
@@ -301,4 +317,397 @@ void copy_with_padding(
...
+  void prefetch_from_host_to_device(T* dev_ptr, size_t src_row_offset, size_t num_rows)
+  {
+    const size_t n_elem  = num_rows * host_view_.extent(1);
+    const size_t n_bytes = n_elem * sizeof(T);
+    // use memcpy instead of raft::copy to avoid strange behavior with HMM/ATS memory
+    RAFT_CUDA_TRY(cudaMemcpyAsync(dev_ptr,
+                                  host_view_.data_handle() + src_row_offset * host_view_.extent(1),
+                                  n_bytes,
+                                  cudaMemcpyHostToDevice,
+                                  prefetch_stream_));
```
aamijar: Same as before: let's try to use the raft copy function.

cpp/src/neighbors/detail/cagra/utils.hpp
```diff
@@ -161,6 +161,22 @@ struct gen_index_msb_1_mask {
...
+  return attr.devicePointer != nullptr;
+}
+
+template <typename T>
+bool is_ptr_host_accessible(T* ptr)
+{
+  cudaPointerAttributes attr;
+  RAFT_CUDA_TRY(cudaPointerGetAttributes(&attr, ptr));
+  return attr.hostPointer != nullptr;
+}
```
aamijar: We should already have utilities somewhere to do check if a pointer is host or device accessible, either within in raft or cuvs.
I can try to find those functions for you.

aamijar (changes requested):

cpp/src/neighbors/detail/cagra/utils.hpp
```diff
@@ -301,4 +317,397 @@ void copy_with_padding(
...
+ * * synchronize on all previous operations / increments batch_id_
+ * * (optionally) write back the data of the previous batch to the host
+ * * (optionally) prefetch the data of the next batch
+ * * return the view of the current batch
+ *
+ * @tparam T The type of the data
+ * @tparam IdxT The type of the index
+ */
+template <typename T, typename IdxT>
+class batched_device_view_from_host {
```
tfeher: Can't we use [batch_load_iterator](https://github.com/NVIDIA/cuvs/blob/8ecfa65e34da02973da82bbb93a834e107972071/cpp/src/neighbors/detail/ann_utils.cuh#L478) for the same purpose?

irina-resh-nvda:
the new main merge introduced some crashes, im investigating

cpp/src/neighbors/detail/cagra/cagra_build.cuh
```diff
@@ -2136,71 +2413,135 @@ auto iterative_build_graph(
...
+    graph_scatter_index_op<IdxT> scatter_idx_op{dev_perm.data_handle(), degree};
+    auto output_indices =
+      thrust::make_transform_iterator(thrust::make_counting_iterator<int64_t>(0), scatter_idx_op);
+
+    thrust::scatter(raft::resource::get_thrust_policy(res),
+                    dev_graph.data_handle(),
+                    dev_graph.data_handle() + n_rows * degree,
+                    output_indices,
+                    dev_unshuffled_graph.data_handle());
+
```
achirkin: Would it be possible to reformulate this using [`raft::linalg::map_offset`](https://github.com/NVIDIA/raft/blob/aa9a6d5083fb445cd1948e7e47053abb7736cb8f/cpp/include/raft/linalg/map.cuh#L218) and/or [`raft::matrix::scatter`](https://github.com/NVIDIA/raft/blob/main/cpp/include/raft/matrix/scatter.cuh#L41)?
Normally we're trying to use raft primitives over anything else to stay in mdspan API and prepare for future enhancements, such as dry run functionality.

cpp/src/neighbors/detail/cagra/cagra_build.cuh
```diff
@@ -2100,28 +2280,125 @@ auto iterative_build_graph(
...
+      // This is used for:
+      // 1. Remapping neighbor values: neighbor j (shuffled) -> perm[j] (original)
+      // 2. Reordering rows: row i (for shuffled node i) -> position perm[i] (original node)
+      dev_perm = raft::make_device_vector<IdxT, int64_t>(res, n_rows);
+      cast_to_idx_op<IdxT> cast_op;
+      thrust::transform(raft::resource::get_thrust_policy(res),
+                        dev_perm_i64.data_handle(),
+                        dev_perm_i64.data_handle() + n_rows,
+                        dev_perm.data_handle(),
+                        cast_op);
```
achirkin:
```suggestion
      raft::linalg::map(res,
                        dev_perm.view(),
                        raft::cast_op<IdxT>{},
                        raft::make_const_mdspan(dev_perm_i64.view()));
```

cpp/src/neighbors/detail/cagra/cagra_build.cuh
```diff
@@ -2018,6 +2053,110 @@ struct mmap_owner {
...
+                         raft::device_matrix_view<float, int64_t> dev_distances,
+                         raft::device_matrix<IdxT, int64_t>& dev_output_graph,
+                         size_t curr_query_size,
+                         size_t next_graph_degree,
+                         size_t curr_topk,
+                         uint64_t max_chunk_size)
+{
+  auto stream = raft::resource::get_cuda_stream(res);
+
+  auto dev_knn_graph = raft::make_device_matrix<IdxT, int64_t>(res, curr_query_size, curr_topk);
```
achirkin, line 2079: How big the `curr_query_size` here - is it bounded? 
Consider using  `raft::resource::get_workspace_resource_ref(res)` or  `raft::resource::get_large_workspace_resource_ref(res)` depending on the answer  to correctly track this as a temporary allocation.

cpp/src/neighbors/detail/cagra/cagra_build.cuh
```diff
@@ -2018,6 +2053,110 @@ struct mmap_owner {
...
+                                   batch_dev_distances_view);
+
+    raft::copy(dev_knn_graph.data_handle() + batch.offset() * curr_topk,
+               batch_dev_neighbors_view.data_handle(),
+               batch.size() * curr_topk,
+               stream);
+  }
+
+  dev_output_graph =
+    raft::make_device_matrix<IdxT, int64_t>(res, curr_query_size, next_graph_degree);
```
achirkin: The same question about being bounded and using the workspace resource here.
Also, do you allocate it multiple times (i.e. by calling it multiple times)?
 - if no: maybe use the return semantics here for clarity (instead of passing the output variable by reference)?
 - if yes: it would make sense to first clear the variable, and then assign a new one - to avoid two large arrays being allocated at the same time (increasing the memory footprint of the whole thing). Perhaps, doing this outside of the `search_and_optimize` function and changing the function to return the new graph would make sense in this case too.

cpp/src/neighbors/detail/cagra/utils.hpp
achirkin: It looks like a merge/diff view artifact. I think @mfoerste4 has added this new class earlier in his refactoring and I asked exactly the same question. There was seemingly a good reason for it, but I forgot it :) @mfoerste4 could you please remind?

achirkin (changes requested):
Hi, a few suggestions here, mainly about the use of raft resources and helper utilities in cuvs.

cpp/tests/neighbors/ann_cagra.cuh
irina-resh-nvda: this is a leftover from my testing, i'll fix it

cpp/tests/neighbors/ann_cagra.cuh
irina-resh-nvda: sorry, testing artefact, will fix

cpp/src/neighbors/detail/cagra/cagra_build.cuh
irina-resh-nvda: I removed shuffle functionality altogether as per API changes (let the user own the data)

cpp/src/neighbors/detail/cagra/utils.hpp
irina-resh-nvda: i think this change is indeed some kind of diff artefact

cpp/src/neighbors/detail/cagra/cagra_build.cuh
irina-resh-nvda: It's bounded by the dataset size, so i will use `get_large_workspace_resource_ref` as you suggested

cpp/src/neighbors/detail/cagra/cagra_build.cuh
irina-resh-nvda: same N-limited boundedness. Will change to return, and release the old graph before allocating the new one to reduce peak memory use

aamijar:
It looks like a lot of files were edited to change the copyright header? I think we should revert those.
Update I fixed it in these commits:
355d240
8cd4191

This commit 6ad9234 reverted minor diff in a test file.

We should look through the merge conflicts due to Datasets API and resolve them.
Fyi a few cagra-q fixes are being worked on here https://github.com/NVIDIA/cuvs/pull/2413

cpp/src/neighbors/detail/cagra/utils.hpp
aamijar: Bumping this, if we don't need this utility we can remove a bunch of code from this PR from this file `utils.hpp` and delete the `test_batched_device_view_from_host.cu` test file.

mfoerste4: Yes, this seems to be a merge issue. The class and test do not exist on main anymore.

aamijar: I've removed the `utils.hpp` diff, `test_batched_device_view_from_host.cu`, and the `cmakelists` diff here [8a9fda5](https://github.com/NVIDIA/cuvs/pull/1810/commits/8a9fda5c02c1fae5b6ddf9d99476f42980ac00f4)

achirkin (dismissed):
Please undo all changes to `bench/ann/src/common`. Let's keep CAGRA-Q compressed dataset support as before - by letting the compressed dataset be a part of index file / folder.
We should leave a proper refactoring of the benchmark harness for discussion and implementation in follow up work (if and how we want to pass compressed datasets as first-class input to the benchmarks).

irina-resh-nvda:
> Please undo all changes to `bench/ann/src/common`. Let's keep CAGRA-Q compressed dataset support as before - by letting the compressed dataset be a part of index file / folder. We should leave a proper refactoring of the benchmark harness for discussion and implementation in follow up work (if and how we want to pass compressed datasets as first-class input to the benchmarks).

I created a standalone [PR for the vpq dataset serialisation](https://github.com/NVIDIA/cuvs/pull/2480). We can look through it and merge faster that this PR, and start working on the cuvsbench refactoring sooner.

I need the refactoring to rerun the benchmarks (the older performance numbers are obsolete now). I won't be comfortable merging this PR into main before the new benchmarks land

cpp/src/neighbors/detail/cagra/cagra_build.cuh
```diff
@@ -1994,51 +1992,178 @@ void optimize(
...
+  // workspace resource here would use an unpooled managed_memory_resource, paying a synchronous
+  // cudaMallocManaged/cudaFree every iteration for multi-GB buffers.
+  auto dev_knn_graph = raft::make_device_matrix<IdxT, int64_t>(res, curr_query_size, curr_topk);
+
+  // Scratch for one reconstructed or depadded chunk. Reused across chunks; safe because all
+  // reconstruct/search/copy work is serialized on `stream`.
+  auto batch_queries =
+    vpq_queries != nullptr || dev_query_view.extent(1) != query_dim
+      ? raft::make_device_matrix<T, int64_t>(res, static_cast<int64_t>(max_chunk_size), query_dim)
+      : raft::make_device_matrix<T, int64_t>(res, 0, 0);
```
tarang-jain, line 2086: Looks like CAGRA search will pad these queries to the aligned dim if they are not aligned right? So why not make the batch itself padded here -- will avoid an extra D2D copy during cagra search (it might need the reconstruct kernel to jump the padded dim).

tarang-jain (commented):
Thanks for the great work!

tarang-jain:
What do you think about putting the ANN bench changes in a separate PR (even though you want to benchmark this e2e)? That can be a follow up to this PR.

c/include/cuvs/core/dataset.h
```diff
@@ -72,6 +76,14 @@ CUVS_EXPORT cuvsError_t cuvsDatasetMakePadded(cuvsResources_t res,
                                               cuvsDatasetMemType_t target_mem_type,
                                               cuvsDataset_t* padded_dataset);
 
+/**
+ * @brief Compress a dense dataset into a device VPQ dataset.
+ */
+CUVS_EXPORT cuvsError_t cuvsDatasetMakeVpq(cuvsResources_t res,
+                                           cuvsCagraCompressionParams_t params,
```
tfeher: This function compresses the dataset into VPQ format. The user can decide later to use that for CAGRA, or for other algorithms. It would make sense to rename the params struct to something like `cuvsPqParams`.

c/include/cuvs/core/dataset.h
```diff
@@ -20,7 +20,8 @@ extern "C" {
  */
 typedef enum {
   CUVS_DATASET_LAYOUT_STANDARD = 0,
-  CUVS_DATASET_LAYOUT_PADDED   = 1
+  CUVS_DATASET_LAYOUT_PADDED   = 1,
+  CUVS_DATASET_LAYOUT_VPQ = 2
```
tfeher: I think Corey recommended to use PQ instead of VPQ

cpp/bench/ann/src/cuvs/cuvs_ann_bench_param_parser.h
```diff
@@ -406,6 +407,85 @@ void parse_build_param(const nlohmann::json& conf, cuvs::neighbors::cagra::index
...
+          arg.num_random_samplings = build_search_conf.at("num_random_samplings");
+        }
+        if (build_search_conf.contains("persistent")) {
+          arg.persistent = build_search_conf.at("persistent");
+        }
+        if (build_search_conf.contains("persistent_lifetime")) {
+          arg.persistent_lifetime = build_search_conf.at("persistent_lifetime");
+        }
+        if (build_search_conf.contains("persistent_device_usage")) {
+          arg.persistent_device_usage = build_search_conf.at("persistent_device_usage");
```
tfeher: I assume `build_search` is set of parameters used only during build. Is persistent mode used during build? I don't expect that, since with large batch search operations we have large throughput without the persistent kernel. My concern in exposing this as tunable build parameter is that it might motivate users to add these options to their build parameter scan, which is not productive. 

If the goal is completeness in parameter handling, then shouldn't we use the same function that parses search params?

cpp/src/neighbors/detail/cagra/cagra_serialize.cuh
```diff
@@ -401,7 +408,16 @@ void deserialize(
   std::unique_ptr<owner_t> dataset_owner{};
   if (has_dataset) {
     if (out_dataset == nullptr) {
-      cuvs::neighbors::detail::skip_dense_dataset<T, int64_t>(res, is);
+      // Dropping the rows leaves a searchable index for a dense view, whose dataset can be
+      // reattached from the caller's own copy, but not for a VPQ one: the compressed rows exist
```
tfeher: Why do we handle the pq compressed dataset differently? Can't the user reattach a pq compressed dataset to an index after the graph was loaded from file?

cpp/src/neighbors/detail/cagra/cagra_build.cuh
```diff
@@ -2059,57 +2184,76 @@ auto ensure_device_padded_for_iterative_search(
...
+    raft::make_device_matrix_view<const T, int64_t>(static_cast<const T*>(nullptr), 0, 0);
+  uint32_t logical_dim = dataset.dim();
+  uint64_t final_graph_size;
+  const cuvs::neighbors::device_vpq_dataset<half, int64_t>* vpq_dataset = nullptr;
+
+  if constexpr (cuvs::neighbors::is_device_vpq_f16_dataset_view_v<DatasetViewT>) {
+    final_graph_size = static_cast<uint64_t>(dataset.n_rows());
+    vpq_dataset      = &dataset.dset();
+  } else {
+    auto search_dataset = ensure_device_padded_for_iterative_search<T>(res, dataset, padded_own);
```
tfeher: I believe that using the new DataSet API, the guideline is that we do not make such copies for the user, but expect them to provide already padded data.

cpp/src/neighbors/detail/cagra/cagra_build.cuh
```diff
@@ -2166,86 +2316,90 @@ auto iterative_build_graph(raft::resources const& res,
...
+                                      static_cast<int64_t>(logical_dim),
+                                      flag_last && params.guarantee_connectivity);
+    } else {
+      auto dev_dataset_view = raft::make_device_matrix_view<const T, int64_t>(
+        dev_dataset.data_handle(), static_cast<int64_t>(curr_graph_size), dev_dataset.extent(1));
+      cuvs::neighbors::device_padded_dataset_view<T, int64_t> sub_padded(dev_dataset_view,
+                                                                         logical_dim);
+      auto idx = cuvs::neighbors::cagra::device_padded_index<T, IdxT>(res, params.metric);
+      idx.update_device_dataset_same_layout(res, sub_padded);
+      if (use_device_graph) {
```
tfeher: Expect of the first iteration, we always use `dev_graph`, right? Could you simply copy the initial `cagra_graph` to the device so that we can simplify the code here?

cpp/src/neighbors/detail/cagra/cagra_build.cuh
tfeher: Could you allocate the temporary arrays before you start the iterative build? (At the max size it will be used). This way we would fail early if mem is not enough.

tfeher (changes requested):
Thanks @irina-resh-nvda and @aamijar for this PR!  My main questions below are about naming of params / APIs related to PQ compressed input. Otherwise, the PR looks great overall.

cpp/src/neighbors/detail/cagra/cagra_build.cuh
tarang-jain: Done!

cpp/src/neighbors/detail/cagra/cagra_build.cuh
tarang-jain: so we basically error out if its not padded, right?

cpp/src/neighbors/cagra.cuh
```diff
@@ -296,13 +296,43 @@ template <typename DatasetViewT>
...
+
+    detail::check_graph_degree<T, IdxT>(effective_params.intermediate_graph_degree,
+                                        effective_params.graph_degree,
+                                        static_cast<size_t>(dataset.n_rows()));
+    auto cagra_graph = detail::iterative_build_graph<T, IdxT>(res, effective_params, dataset);
+
+    index_type idx(res, effective_params.metric);
+    idx.update_graph(res, raft::make_const_mdspan(cagra_graph.view()));
+    if (effective_params.attach_dataset_on_build) {
+      idx.update_device_dataset_same_layout(res, dataset);
```
aamijar: I thought we got rid of this function `update_device_dataset_same_layout`, probably a merge conflict artifact

c/include/cuvs/core/dataset.h
tarang-jain: Unfortunately the struct's constructor and desctructor are declared in the cagra header. Renaming it would be ABI breaking. I think we'd want to later combine this with `cuvsProductQuantizerParams`. We likely dont need two structs with more or less the same contents.

tarang-jain:
> Please undo all changes to bench/ann/src/common. Let's keep CAGRA-Q compressed dataset support as before - by letting the compressed dataset be a part of index file / folder.

This is done. Will open a separate PR for 26.12

cpp/src/neighbors/detail/cagra/cagra_serialize.cuh
tarang-jain: This comment has not been addressed yet. We will take care of this in https://github.com/NVIDIA/cuvs/pull/2480. I added a reference there to this comment.

cpp/bench/ann/src/cuvs/cuvs_ann_bench_param_parser.h
tarang-jain: Created a new issue to track this.

cpp/src/neighbors/detail/cagra/cagra_build.cuh
tfeher: yes

tarang-jain: Done! I removed that helper.

c/include/cuvs/core/dataset.h
aamijar: Should we add the `cuvsDatasetMemType_t target_mem_type,` argument here?
We don't have host resident PQ datasets allowed yet, but just for futureproofing ABI do you think it would be wise to do?

cpp/src/neighbors/detail/dataset_serialize.hpp
```diff
@@ -412,6 +412,40 @@ auto deserialize_host_dense(raft::resources const& res, std::istream& is)
...
+ * Write the payload of a VPQ dataset: six scalars followed by the two codebooks and the encoded
+ * rows.
+ *
+ * Stays on `raft::serialize_mdspan` rather than the `write_dense_bytes` scheme used by the dense
+ * path above, because `deserialize_vpq` reads with `raft::deserialize_mdspan`, which expects the
+ * NumPy header that helper embeds per matrix. The scalar types must also match the reader exactly:
+ * `n_rows` is `IdxT` and the remaining five are `uint32_t`.
+ */
+template <typename DataT, typename IdxT>
+void serialize_vpq(raft::resources const& res,
```
aamijar: It seems as though we are supporting PQ dataset serialization in this PR?
So we are supporting serialization and deserialization of index + attached PQ dataset in this PR?

c/src/neighbors/cagra.cpp
```diff
@@ -920,13 +1012,19 @@ void _serialize(cuvsResources_t res, const char *filename,
...
   with_index_by_layout<T, uint32_t,
                        true>(box, null_handle_err, "", [&](auto &idx) {
-    if (include_dataset) {
-      RAFT_EXPECTS(
-          idx.dataset().n_rows() > 0,
-          "cuvsCagraSerializeGraphAndDataset: index has no attached dataset");
+    using index_dataset_view_t = std::remove_cvref_t<decltype(idx.dataset())>;
+    if constexpr (cuvs::neighbors::is_vpq_dataset_view_v<index_dataset_view_t>) {
+      RAFT_FAIL(
+        "CAGRA index serialization is not supported for VPQ indices");
```
aamijar: I think we need to fix this for the C API? We are supporting index serialization in CPP but not in C?
Not sure if C API serialization is necessary for this PR.

c/src/neighbors/cagra.cpp
```diff
@@ -374,6 +390,77 @@ static void with_dataset_view(cuvsDataset_t dataset, Fn&& fn)
...
+  out.pq_dim                          = params.pq_dim;
+  out.vq_n_centers                    = params.vq_n_centers;
+  out.kmeans_n_iters                  = params.kmeans_n_iters;
+  out.vq_kmeans_trainset_fraction     = params.vq_kmeans_trainset_fraction;
+  out.pq_kmeans_trainset_fraction     = params.pq_kmeans_trainset_fraction;
+  return out;
+}
+
+template <typename T>
+static auto make_vpq_from_dense_dataset(raft::resources* res_ptr,
```
aamijar: Making a note to rename any new instances of VPQ to PQ instead.
The existing instances will be renamed as part of https://github.com/NVIDIA/cuvs/pull/2494.

cpp/include/cuvs/neighbors/cagra.hpp
```diff
@@ -4731,6 +4880,19 @@ auto update_dataset(
   device_vpq_dataset_view<half, int64_t> dataset)
   -> index<uint8_t, uint32_t, device_vpq_dataset_view<half, int64_t>>;
 
+auto update_dataset(raft::resources const& res,
+                    vpq_f16_index<float>&& cagra_index,
```
aamijar: We have a bunch of uses of the alias `vpq_f16_index`. I'm wondering whether we should change this alias to just be pq_index as part of this PR.

Or use the full namespace version:
```
cuvs::neighbors::cagra::index<
  float,                                                   // query/input type
  uint32_t,                                                // neighbor ID type
  cuvs::neighbors::device_vpq_dataset_view<half, int64_t> // dataset view
>
```
It is a bit verbose though.

cpp/include/cuvs/neighbors/cagra.hpp
```diff
@@ -967,6 +1000,27 @@ using cagra_index_t = index<cuvs::neighbors::cagra_view_element_type_t<DatasetVi
...
+ *
+ * @param[in] res raft resources
+ * @param[in] params CAGRA index build parameters
+ * @param[in] dataset device VPQ dataset view
+ * @return built `index<float, uint32_t, device_vpq_dataset_view<half, int64_t>>`
+ */
+auto build(raft::resources const& res,
+           const cuvs::neighbors::cagra::index_params& params,
+           cuvs::neighbors::device_vpq_dataset_view<half, int64_t> const& dataset)
+  -> index<float, uint32_t, cuvs::neighbors::device_vpq_dataset_view<half, int64_t>>;
```
aamijar: I just noticed we are supporting `half` but not `float` for vpq datasets. I think we do have a float version of vpq? Is that something we are supporting in Iterative CAGRA-Q? I think CAGRA-Q search kernels already support it?

cpp/include/cuvs/neighbors/cagra.hpp
```diff
@@ -4731,6 +4880,19 @@ auto update_dataset(
...
+                    device_vpq_dataset_view<half, int64_t> dataset) -> vpq_f16_index<float>;
+auto update_dataset(raft::resources const& res,
+                    vpq_f16_index<half>&& cagra_index,
+                    device_vpq_dataset_view<half, int64_t> dataset) -> vpq_f16_index<half>;
+auto update_dataset(raft::resources const& res,
+                    vpq_f16_index<int8_t>&& cagra_index,
+                    device_vpq_dataset_view<half, int64_t> dataset) -> vpq_f16_index<int8_t>;
+auto update_dataset(raft::resources const& res,
+                    vpq_f16_index<uint8_t>&& cagra_index,
+                    device_vpq_dataset_view<half, int64_t> dataset) -> vpq_f16_index<uint8_t>;
```
aamijar: We definitely need to add the overloads to go from PQ backed index to Dense backed index.
That is the main use case for us too. As seen from benchmarking searching on dense backed index is what gives us 99%+ recall potential.

cpp/src/neighbors/detail/cagra/cagra_build.cuh
```diff
@@ -2231,113 +2232,258 @@ void optimize(
...
 
-  [[nodiscard]] auto data() const -> void* { return ptr_; }
-  [[nodiscard]] auto size() const -> size_t { return size_; }
-
- private:
-  void* ptr_;
-  size_t size_;
-};
+template <typename T, typename MathT, typename IdxT>
+void reconstruct_vpq_queries(raft::resources const& res,
```
aamijar, line 2265: Thinking longer term, have we considered using the reconstruction functionality from the existing PQ APIs? I don't think it can be directly replaced right now because of differing format, but perhaps something to explore later.

cpp/include/cuvs/neighbors/cagra.hpp
aamijar: Addressed in https://github.com/NVIDIA/cuvs/pull/1810/commits/ac66cc13519083f74936e96fe245e2101784625c

cpp/include/cuvs/neighbors/cagra.hpp
aamijar: Addressed in https://github.com/NVIDIA/cuvs/pull/1810/commits/ac66cc13519083f74936e96fe245e2101784625c

c/src/neighbors/cagra.cpp
aamijar: Addressed in https://github.com/NVIDIA/cuvs/pull/1810/commits/ac66cc13519083f74936e96fe245e2101784625c and https://github.com/NVIDIA/cuvs/pull/1810/commits/b12deb943657f36c851972711edcff82ea82cd2c

c/include/cuvs/core/dataset.h
aamijar: We added that argument here https://github.com/NVIDIA/cuvs/pull/1810/commits/ac66cc13519083f74936e96fe245e2101784625c for consistency. As discussed with @tarang-jain offline, adding arguments later on also wouldn't be an issue because that doesn't break ABI. Deleting arguments does break ABI though.

cpp/cmake/patches/faiss_override.json
```diff
@@ -7,7 +7,7 @@
       "patches" : [
         {
           "file" : "${current_json_dir}/faiss-1.14-cuvs-26.08.diff",
-          "issue" : "Multiple fixes for cuVS compatibility. Update Faiss cuVS to be compatible with new Dataset API: update_device_dataset_same_layout now takes dataset_view and make_padded_dataset_view must be called beforehand. Loading an index built from a user-provided KNN graph passes dataset_view into cagra::index, not raw mdspan.",
+          "issue" : "Multiple fixes for cuVS compatibility. Update Faiss cuVS to be compatible with the new Dataset API: pass dataset_view and call make_padded_dataset_view beforehand. Loading an index built from a user-provided KNN graph passes dataset_view into cagra::index, not raw mdspan.",
```
aamijar: Note to remove this unrelated diff.

cpp/src/neighbors/detail/dataset_serialize.hpp
aamijar: Addressed in https://github.com/NVIDIA/cuvs/pull/1810/commits/b18bbdf1e9edb304631561c3ceab83ba99279f39.
In this PR we support serialization and deserialization of the device_pq_index only. Dataset serialization will be addressed in https://github.com/NVIDIA/cuvs/pull/2480.

c/tests/CMakeLists.txt
```diff
@@ -89,7 +89,9 @@ ConfigureTest(NAME IVF_FLAT_C_TEST PATH neighbors/run_ivf_flat_c.c neighbors/ann
 ConfigureTest(NAME IVF_PQ_C_TEST PATH neighbors/run_ivf_pq_c.c neighbors/ann_ivf_pq_c.cu)
 ConfigureTest(NAME IVF_SQ_C_TEST PATH neighbors/run_ivf_sq_c.c neighbors/ann_ivf_sq_c.cu)
 ConfigureTest(NAME CAGRA_C_TEST PATH neighbors/ann_cagra_c.cu)
-ConfigureTest(NAME MG_C_TEST PATH neighbors/run_mg_c.c neighbors/ann_mg_c.cu)
+if(BUILD_MG_ALGOS)
+  ConfigureTest(NAME MG_C_TEST PATH neighbors/run_mg_c.c neighbors/ann_mg_c.cu)
+endif()
```
aamijar: Note to remove this unrelated diff.

aamijar: Addressed in https://github.com/NVIDIA/cuvs/pull/1810/commits/89a5394f7733bef15bfc817ed82c575717f79501

cpp/cmake/patches/faiss_override.json
aamijar: Addressed in https://github.com/NVIDIA/cuvs/pull/1810/commits/89a5394f7733bef15bfc817ed82c575717f79501

c/src/neighbors/cagra.cpp
aamijar: Addressed in https://github.com/NVIDIA/cuvs/pull/1810/commits/db81e15a3831fda7d9335479ae3add5dc8aaad14

cpp/include/cuvs/neighbors/cagra.hpp
aamijar: Opened issue for it https://github.com/NVIDIA/cuvs/issues/2590

cpp/src/neighbors/detail/cagra/cagra_build.cuh
aamijar: Opened an issue for it https://github.com/NVIDIA/cuvs/issues/2591

cpp/src/neighbors/detail/cagra/cagra_build.cuh
```diff
@@ -2016,20 +2257,26 @@ __global__ void kern_reconstruct_vpq_queries(const uint8_t* encoded_data,
...
                 static_cast<float>(pq_codebook[static_cast<uint32_t>(pq_codes[j]) * pq_len + k]);
-    output[batch_idx * dim + d] = static_cast<T>(val);
+    output[batch_idx * output_ld + d] = static_cast<T>(val);
   }
 }
 
 template <typename T, typename MathT, typename IdxT>
 void reconstruct_vpq_queries(raft::resources const& res,
-                             const cuvs::neighbors::device_vpq_dataset<MathT, IdxT>& vpq_dset,
+                             cuvs::neighbors::device_vpq_dataset_view<MathT, IdxT> const& vpq_view,
```
tfeher: Is it intentional to pass the `dataset_view` as a reference? Normally I would expect views to be similar like a pointer that are cheap to pass by value.

aamijar:
/ok to test [100db4f](https://github.com/NVIDIA/cuvs/pull/1810/commits/100db4fa5e7a95949b817c7976f03bf1c67d5639)

irina-resh-nvda:
It looks good to me

aamijar (approved):

tarang-jain (approved):

aamijar:
/ok to test https://github.com/NVIDIA/cuvs/pull/1810/commits/19f73974c08dff17188bd7982d166635061c2647

cpp/src/neighbors/detail/cagra/cagra_serialize.cuh
aamijar: >Why do we handle the pq compressed dataset differently? Can't the user reattach a pq compressed dataset to an index after the graph was loaded from file?

We can do this with the current PR btw. User should be able to serialize and deserialize the graph. And they can attach whatever dataset they want later with update_dataset as usual.

cpp/src/neighbors/detail/cagra/cagra_build.cuh
```diff
@@ -2360,129 +2517,138 @@ auto iterative_build_graph(raft::resources const& res,
...
-    flag_last       = (curr_graph_size == final_graph_size);
-    curr_graph_size = next_graph_size;
+    flag_last            = (curr_graph_size == final_graph_size);
+    auto next_graph_size = curr_query_size;
+    curr_graph_size      = next_graph_size;
   }
 
+  auto cagra_graph =
+    raft::make_host_matrix<IdxT, int64_t>(dev_graph.extent(0), dev_graph.extent(1));
+  raft::copy(cagra_graph.data_handle(),
```
aamijar: Wait - this should be removed. We are doing a D2H of the entire graph. And then if you look at the caller we are doing H2D back again. `update_graph` member function does a copy to device. 

https://github.com/NVIDIA/cuvs/blob/df7730dfa8c2e4ec3e2c504fdb1bc00736f0265e/cpp/src/neighbors/cagra.cuh#L328-L331

https://github.com/NVIDIA/cuvs/blob/51917751f38663189a964bcfa342932b44095066/cpp/include/cuvs/neighbors/cagra.hpp#L704-L708

aamijar: Addressed in https://github.com/NVIDIA/cuvs/pull/1810/commits/ffd1a1dee1664fc8b26cbb4848b7f48b5a91b4a3

aamijar:
/ok to test https://github.com/NVIDIA/cuvs/pull/1810/commits/ffd1a1dee1664fc8b26cbb4848b7f48b5a91b4a3

cpp/include/cuvs/neighbors/cagra.hpp
```diff
@@ -964,9 +994,34 @@ using cagra_index_t = index<cuvs::neighbors::cagra_view_element_type_t<DatasetVi
...
+ * `attach_dataset_on_build` is true. The owning `device_vpq_dataset` must outlive the index.
+ *
+ * @param[in] res raft resources
+ * @param[in] params CAGRA index build parameters
+ * @param[in] dataset device VPQ dataset view
+ * @return built `index<float, uint32_t, device_vpq_dataset_view<half, int64_t>>`
+ */
+auto build(raft::resources const& res,
+           const cuvs::neighbors::cagra::index_params& params,
+           cuvs::neighbors::device_vpq_dataset_view<half, int64_t> const& dataset)
```
tfeher, line 1022: Is it intentional to pass the dataset_view as a reference? Normally I would expect views to be similar like a pointer that are cheap to pass by value.

aamijar: Yep, we should change all of these cases to pass by value. There is an issue for it here https://github.com/NVIDIA/cuvs/issues/2445. We can address them all at once in separate PR. We will also need to change `cagra_build_inst.cu.in`.

aamijar:
/ok to test https://github.com/NVIDIA/cuvs/pull/1810/commits/abd796c58640079bf4fe17eef83f8ec07af86d7a

cpp/src/neighbors/detail/cagra/cagra_serialize.cuh
tfeher: Yes, my comment here is only about skipping reading the dataset. If we can just skip reading a compressed dataset, instead of throwing an error, then let's do that. But if we would need #2480 to be finished, so that we know what is the data size that needs to be skipped, then we can postpone it for #2480.

aamijar:
/ok to test https://github.com/NVIDIA/cuvs/pull/1810/commits/e218f9ef0e1f909be12a46992aef7d6506e86703

cpp/src/neighbors/detail/cagra/cagra_build.cuh
```diff
@@ -2231,113 +2232,258 @@ void optimize(
...
+        std::min<int64_t>(static_cast<int64_t>(max_chunk_size), curr_query_size - offset);
+      reconstruct_vpq_queries<T, half, int64_t>(
+        res,
+        *vpq_queries,
+        static_cast<uint64_t>(offset),
+        static_cast<uint32_t>(batch_size),
+        raft::make_device_matrix_view<T, int64_t>(
+          reconstructed_batch_queries->data_handle(), batch_size, query_ld));
+      auto batch_query_view = raft::make_device_matrix_view<const T, int64_t>(
+        reconstructed_batch_queries->data_handle(), batch_size, query_ld);
```
tfeher, line 2379: Nitpick, but we can construct the view first and use in two places
```suggestion
      auto batch_query_view = raft::make_device_matrix_view<const T, int64_t>(
        reconstructed_batch_queries->data_handle(), batch_size, query_ld);
      reconstruct_vpq_queries<T, half, int64_t>(
        res,
        *vpq_queries,
        static_cast<uint64_t>(offset),
        static_cast<uint32_t>(batch_size),
        batch_query_view);
```

cpp/src/neighbors/detail/cagra/cagra_build.cuh
tfeher: The points here are not yet addressed, I have created a tracking issue https://github.com/NVIDIA/cuvs/issues/2610

tfeher (approved):
Thanks everyone for pushing it through the finish line, the PR looks good to me!

aamijar:
/ok to test https://github.com/NVIDIA/cuvs/pull/1810/commits/a96e09059bd5ed8c9ae907f33cbf32019697a437

aamijar:
/ok to test https://github.com/NVIDIA/cuvs/pull/1810/commits/a96e09059bd5ed8c9ae907f33cbf32019697a437

aamijar:
/ok to test https://github.com/NVIDIA/cuvs/pull/1810/commits/c3029e67c34a9740c9e2940ddaee24c6d465f67c

tfeher:
I marked this as breaking change since it is changed what type of datasets are supported by CAGRA iterative build. Prior to this PR host dataset was also accepted. After this PR, it is expected that dataset is searchable from device (has right alignment and device accessible). This is in line with the new dataset API. We expect that users explicitly prepare the dataset in the right format before passing it to cagra::build.

aamijar:
/ok to test https://github.com/NVIDIA/cuvs/pull/1810/commits/75b9c3dd3602ad0d3f2309c2ecb3440275126c86

aamijar:
/ok to test https://github.com/NVIDIA/cuvs/pull/1810/commits/e0d1e909bed3c564f292f06ed2cd09827b8309b3

irina-resh-nvda:
/ok to test [4ae6b9f](https://github.com/NVIDIA/cuvs/pull/1810/commits/4ae6b9f93d6f6b5645d8751169a3d3103f829f39)

aamijar:
/ok to test https://github.com/NVIDIA/cuvs/pull/1810/commits/4ae6b9f93d6f6b5645d8751169a3d3103f829f39

tfeher:
/ok to test [eec1063](https://github.com/NVIDIA/cuvs/pull/1810/commits/eec1063082aeed24367420d717de93df31205333)

python/cuvs/cuvs/preprocessing/quantize/pq/pq.pyx
```diff
@@ -22,6 +27,35 @@ PQ_KMEANS_TYPES = {
...
 
+
+cdef class PQDatasetParams:
+    """Parameters for creating a PQ dataset."""
+
+    cdef cuvsCagraCompressionParams* params
+
+    def __cinit__(self):
+        self.params = NULL
+        check_cuvs(cuvsCagraCompressionParamsCreate(&self.params))
```
tarang-jain: can we avoid using cagra compression params from the PQ class?

python/cuvs/cuvs/neighbors/cagra/cagra.pyx
```diff
@@ -582,26 +588,26 @@ def build(IndexParams index_params, dataset, resources=None):
...
     else:
-        source_array = padded_dataset
-        dataset_obj = make_device_padded_dataset(padded_dataset, resources=resources)
+        source_array = dataset
+        dataset_obj = make_device_padded_dataset(dataset, resources=resources)
 
     cdef cuvsDataset_t dataset_handle = _cagra_dataset_handle(dataset_obj)
-    if dataset_obj.layout != "padded":
-        raise TypeError("padded_dataset must have padded layout")
+    if dataset_obj.layout not in ("padded", "vpq"):
```
tarang-jain: change this to PQ

python/cuvs/cuvs/neighbors/cagra/cagra.pyx
```diff
@@ -530,7 +533,10 @@ def build(IndexParams index_params, dataset, resources=None):
                 dl_data_type_to_numpy(idx.index.dtype)).name
             idx._dataset_source = dataset_obj
 
-            if not is_ace_build:
+            if dataset_obj.layout == "vpq":
```
tarang-jain: change to "pq"

python/cuvs/cuvs/common/dataset.pyx
```diff
@@ -46,6 +47,8 @@ cdef class Dataset:
         if self.dataset == NULL:
             return None
         check_cuvs(cuvsDatasetGetLayout(self.dataset, &layout))
+        if layout == CUVS_DATASET_LAYOUT_PQ:
+            return "vpq"
```
tarang-jain: change to "pq"

tarang-jain (changes requested):

divyegala (approved):

tarang-jain (approved):

PR #1812, opened 2026-02-17
Refactor JIT LTO kernel generation

KyleFromNVIDIA:
Add an algorithm that computes a matrix product, and add a generic
CMake function that uses this algorithm to generate a matrix of kernels
with desired parameters.

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -442,6 +442,7 @@ if(NOT BUILD_CPU_ONLY)
       PRIVATE "${CMAKE_CURRENT_SOURCE_DIR}/include" "${CMAKE_CURRENT_SOURCE_DIR}/src"
               "${CMAKE_CURRENT_SOURCE_DIR}/../c/include"
     )
+    target_link_libraries(cuvs_jit_lto_kernels PRIVATE raft::raft)
     add_library(cuvs::cuvs_jit_lto_kernels ALIAS cuvs_jit_lto_kernels)
   endif()
 


cpp/cmake/modules/compute_matrix_product.py
```diff
@@ -0,0 +1,162 @@
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
dantegd:
```suggestion
# SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```

cpp/cmake/modules/compute_matrix_product.py
```diff
@@ -0,0 +1,162 @@
...
+
+
+IDENTIFIER_RE: re.Pattern = re.compile(r"^(?P<underscores>_*)(?P<rest>.*)$")
+
+
+def iterate_matrix_product(
+    matrix: "Matrix",
+    warn_unused=True,
+    warn_used=True,
+) -> "Generator[dict[str, MatrixValue]]":
```
dantegd, line 46: This is great, only very small suggestion I have is to write a short docstring describing/showing an example of the output and conventions for keys, will be useful for future mantainers

cpp/cmake/modules/generate_jit_lto_kernels.cmake
```diff
@@ -7,214 +7,157 @@
...
+    process_matrix_entry(
+      "${source_list_var}"
+      NAME_FORMAT "${_JIT_LTO_NAME_FORMAT}"
+      KERNEL_INPUT_FILE "${_JIT_LTO_KERNEL_INPUT_FILE}"
+      EMBEDDED_INPUT_FILE "${_JIT_LTO_EMBEDDED_INPUT_FILE}"
+      OUTPUT_DIRECTORY "${_JIT_LTO_OUTPUT_DIRECTORY}"
+      MATRIX_JSON_ENTRY "${matrix_json_entry}"
+      KERNEL_LINK_LIBRARIES ${_JIT_LTO_KERNEL_LINK_LIBRARIES}
     )
+    list(APPEND ${source_list_var} "${embedded_file}")
```
dantegd: `"${embedded_file}"` is not visible here, I think this line is not required and does nothing, the required things are done in `process_matrix_entry`

KyleFromNVIDIA: You're right, must be a goober left over from a previous iteration. Will fix on the next push.

cpp/cmake/modules/compute_matrix_product.py
```diff
@@ -0,0 +1,242 @@
...
+    """Computes a matrix product of a JSON document
+
+    This algorithm computes the product of a matrix in a more sophisticated
+    way than can be done with itertools.product(). Multiple related values
+    can be grouped together, and a dimension can have sub-dimensions. Given
+    the following JSON document:
+
+    .. code-block:: json
+
+      {
```
divyegala, line 56: Is `subgroup_dim` a missing key here?

cpp/cmake/modules/generate_jit_lto_kernels.cmake
```diff
@@ -7,214 +7,149 @@
...
+
+  cmake_parse_arguments(_JIT_LTO "${options}" "${one_value}" "${multi_value}" ${ARGN})
+
+  add_library(${kernel_target} OBJECT EXCLUDE_FROM_ALL "${_JIT_LTO_KERNEL_FILE}")
+  target_link_libraries(${kernel_target} PRIVATE ${_JIT_LTO_LINK_LIBRARIES})
+  target_compile_options(${kernel_target} PRIVATE -Xfatbin=--compress-all --compress-mode=size)
   set_target_properties(
-    ${_JIT_LTO_FATBIN_TARGET}
-    PROPERTIES CUDA_ARCHITECTURES ${JIT_LTO_TARGET_ARCHITECTURE}
-               CUDA_STANDARD 17
```
divyegala, line 42: Should we add CUDA and CXX_STANDARD 20 here?

divyegala (commented):
I think we can normalize how keys are named in the JSON file. If the key ends with:
1. `_name`: Actual name of the template
2. `_type`: Always used to instantiate the actual kernel function template
3. `_type_tag`: Template parameters for `registerAlgorithm`. The JSON should have the whole tag as a value `tag_type_ui`.
4. `_type_val`: Value for the types used in constructor of `registerAlgorithm` as `_name` + `_type_val`

cpp/cmake/modules/compute_matrix_product.py
KyleFromNVIDIA: No

cpp/cmake/modules/generate_jit_lto_kernels.cmake
KyleFromNVIDIA: We're already using CUDA 17 by calling `target_compile_features(jit_lto_kernel_usage_requirements INTERFACE cuda_std_17)` in `cpp/CMakeLists.txt`. If you prefer CUDA 20 and C++20, I can update it accordingly.

KyleFromNVIDIA:
> I think we can normalize how keys are named in the JSON file.

It's not entirely clear to me what this would look like, or if it would even be possible with the current algorithm. Please post a suggestion with your specific renaming recommendations for `interleaved_scan_matrix.json`.

cpp/cmake/modules/generate_jit_lto_kernels.cmake
```diff
@@ -7,214 +7,152 @@
...
-          FATBIN_TARGET "fatbin_${kernel_name}"
-          FATBIN_SOURCE "${filename}"
-          EMBEDDED_TARGET "${target}"
-          EMBEDDED_HEADER "${generated_kernels_dir}/metric_device_functions/${kernel_name}.h"
-          EMBEDDED_ARRAY "embedded_${kernel_name}"
-        )
-      endforeach()
-    endforeach()
-  endforeach()
+  find_package(Python3 REQUIRED COMPONENTS Interpreter)
```
jameslamb, line 129: @dantegd @cjnolet are the non-Python cuVS users like https://github.com/rapidsai/cuvs-lucene ok with now requiring a Python interpreter in the build environment to build cuVS?

(cc @robertmaynard @msarahan @mythrocks in case that affects anything around the libcuvs tarball packaging work)

cpp/cmake/modules/compute_matrix_product.py
```diff
@@ -0,0 +1,242 @@
...
+    pass
+
+
+IDENTIFIER_RE: re.Pattern = re.compile(r"^(?P<underscores>_*)(?P<rest>.*)$")
+
+
+def iterate_matrix_product(
+    matrix: "Matrix",
+    warn_unused=True,
+    warn_used=True,
```
jameslamb, line 45:
```suggestion
def iterate_matrix_product(
    *,
    matrix: "Matrix",
    warn_unused: bool,
    warn_used: bool,
```

Since this isn't general-purpose library code but instead very-specific-only-used-in-one-place code, I think it'd be helpful to be stricter here... enforce keyword-only arguments and remove defaults. Forcing values to the call site means there's one source of configuration to think about.

cpp/cmake/modules/compute_matrix_product.py
```diff
@@ -0,0 +1,242 @@
...
+        sys.stdin
+        if namespace.filename == "-"
+        else open(namespace.filename) as f
+    ):
+        matrix: "Matrix" = json.load(f)
+
+    json.dump(
+        list(
+            iterate_matrix_product(
+                matrix,
```
jameslamb:
```suggestion
                matrix=matrix,
```

Accompanies my suggestion to force keyword-only arguments.

python/cuvs/cuvs/tests/test_compute_matrix_product.py
```diff
@@ -0,0 +1,258 @@
...
+    compute_matrix_product_script = runpy.run_path(
+        str(
+            pathlib.Path(__file__).parent
+            / "../../../../cpp/cmake/modules/compute_matrix_product.py"
+        )
+    )
+except FileNotFoundError:
+    pytest.skip(
+        "Could not find compute_matrix_product.py", allow_module_level=True
+    )
```
jameslamb: This introduces a risk of the tests in this file silently no longer running (for example, if `compute_matrix_product.py` is moved or renamed and this test isn't updated).

Is there some specific situation where it's acceptable for `compute_matrix_product.py` to not be present?

If not, I think this should be stricter and the test should hard-fail if that script isn't found.

I also think it'd be better to move this into `cpp/tests`. `python/cuvs/` should be for the `cuvs` Python package, which these tests don't cover or need.

jameslamb (commented):
I didn't look closely at the source files in `src/neighbors/ivf_flat`, trusting other reviewers to be much more knowledgeable about those than me.

Left a few comments about the general setup though. Most are minor but one is big... are we SURE it's to require a Python interpreter to build `libcuvs`?

Leaving a non-blocking "Comment" review, I'll come back and approve once those questions are answered.

cpp/cmake/modules/generate_jit_lto_kernels.cmake
robertmaynard: We already had a dependency on python3 for the builds due to cutlass bringing it in ( https://github.com/NVIDIA/cutlass/blob/main/CMakeLists.txt#L126 ) so this isn't an issue

jameslamb: Oh ok, thanks.

This was my only truly blocking comment, I'll switch my review to "Approve".

KyleFromNVIDIA:
> are we SURE it's to require a Python interpreter to build libcuvs?

I already discussed this with @divyegala back in December. One of cuvs's dependencies already requires Python to build, so this isn't introducing a new dependency.

Other projects that want this matrix product algorithm may not have the same luxury. Once CMake 4.3 with its [`string(JSON)` improvements](https://cmake.org/cmake/help/latest/release/4.3.html#commands) comes out, and once we require CMake 4.3, it will be possible to port the algorithm to CMake script and ditch the Python script.

python/cuvs/cuvs/tests/test_compute_matrix_product.py
KyleFromNVIDIA: > Is there some specific situation where it's acceptable for `compute_matrix_product.py` to not be present?

None that I can think of.

> If not, I think this should be stricter and the test should hard-fail if that script isn't found.

You're right, I'll do that.

> I also think it'd be better to move this into `cpp/tests`. `python/cuvs/` should be for the `cuvs` Python package, which these tests don't cover or need.

In principle I agree, but for obvious reasons, there is not currently infrastructure to run Python tests in `cpp/tests`. I put it in `python/cuvs` because that was the fastest and easiest way to get the test running. Perhaps we can look at moving it in a follow-up PR, but I don't want that to hold up this one.

.pre-commit-config.yaml
```diff
@@ -25,12 +25,13 @@ repos:
...
       - repo: https://github.com/pre-commit/mirrors-mypy
-        rev: 'v0.971'
+        rev: 'v1.19.1'
         hooks:
               - id: mypy
                 additional_dependencies: [types-cachetools]
                 args: ["--config-file=pyproject.toml",
-                       "python/cuvs/cuvs"]
+                       "python/cuvs/cuvs",
+                       "cpp/cmake/modules/compute_matrix_product.py"]
```
bdice: Maybe broaden this for future expansions.
```suggestion
                       "cpp/cmake/modules"]
```

bdice (approved):

python/cuvs/cuvs/tests/test_compute_matrix_product.py
KyleFromNVIDIA: Please see https://github.com/rapidsai/cuvs/pull/1836 for this follow-up.

PR #1814, opened 2026-02-17
PoC for Dockerfile for building standalone

msarahan:
This is an attempt at providing an example that might serve as easy build instructions for people who want to build their own cuvs standalone C tarball.


robertmaynard (commented):
Missing the `ci/standalone_c/Dockerfile.standalone_c ` file.

Also I prefer `ci/standalone_c/Dockerfile` since the folder name explains what the dockerfile is for

cjnolet:
> Missing the ci/standalone_c/Dockerfile.standalone_c  file.
> Also I prefer ci/standalone_c/Dockerfile since the folder name explains what the dockerfile is for

@robertmaynard @msarahan I'm really excited about this approach, but I'd prefer that we pull this out of the `ci` directory altogether, especially if it's meant for users to interact directly with. We could call it `Docker.standalone` and put it next to the existing Dockerfile. Then we can update `build.sh` (and update `ci/build_standalone_c.sh` to invoke `build.sh`).

msarahan:
To be clear: you want the implementation to live in build.sh? If `ci/build_standalone_c.sh` is calling build.sh, either the implementation is in build.sh, or build.sh is calling something else. Please be very specific about what architecture you want to see. Or better yet, just make the change directly and push to my branch.

msarahan:
I pushed the missing files, just for completeness. Feel free to move them around or rename them as desired.

cjnolet:
Thanks for getting this started, @msarahan. We're certainly happy to take it from here.

PR #1815, opened 2026-02-17
[WIP] Add support for returning distances from `eps_nn`

jcrist:
This adds a new optional `dists` parameter to `cuvs::neighbors::ball_cover::eps_nn`. This takes a vector of the same length as `adj_ja`. If provided, the corresponding distances for all found neighbors in the result will be stored there.

This is part of an effort to add `radius_neighbors`/`radius_neighbors_graph` in `cuml`.


jcrist:
This is my first PR to cuvs. I _believe_ I've plumbed everything through properly here, but wouldn't be surprised if I missed something.

A few notes/questions:

- I only added this API for sparse outputs, since that's the case I need it for. I _think_ that's probably fine here, but happy to add dense support if necessary.
- Likewise, I added support for if both `max_k` and `dists` are provided, but for my use case I'd be equally happy with erroring in that case. I only need the case where `max_k=nullopt`.
- I haven't done any perf testing of this change yet. I can do this easily once I have everything wired through cuml python, unless there's a better existing bench harness in cuvs I should use.
- What's the best way to test this? I found the C++ test file (https://github.com/rapidsai/cuvs/blob/8fab5b372af349b465fcf39d351e12a583d14f48/cpp/tests/neighbors/epsilon_neighborhood.cu#L238), but am unsure how to easily validate the results in C++ (my C++ is also admittedly not great). Could do something as easy as a smoke test where we just check that all returned dists are <= eps, but we probably want something a bit more rigorous than that. I'd appreciate some tips on how best to accomplish a test here.

aamijar:
> What's the best way to test this? I found the C++ test file

One possible approach is to compute the distances between the query and the neighbors found. Then you can compare that with the output `dists` to see if they match.
I think it is also okay to hardcode the expected `dists` instead.

> just check that all returned dists are <= eps

This is also a welcome addition you can add to the gtests.

Once you expose the feature on the cuml side, you add more programatic pytest cases.

cpp/include/cuvs/neighbors/ball_cover.hpp
```diff
@@ -269,6 +269,8 @@ void eps_nn(raft::resources const& handle,
  *                     that it is not guarantueed to return the nearest neighbors.
  *                     Upon return max_k is overwritten with the actual max_k found during
  *                     computation.
+ * @param[out] dists   An optional vector (with same length as adj_ja). If provided, will be
+ *                     used to store the corresponding distances for the computed neighbors.
```
dantegd, line 276: If I'm not mistaken internally we compute `eps2 = eps * eps` and compare the raw squared distance `dist <= eps2`. The `dist` value written into `dists` is this squared distance, so we should clarify that in the docstring with something like:

 "Distances are squared Euclidean distances which avoid `sqrt` for performance"

cpp/src/neighbors/ball_cover.cuh
```diff
@@ -328,6 +331,9 @@ void eps_nn(raft::resources const& handle,
   int64_t* max_k_ptr = nullptr;
   if (max_k.has_value()) { max_k_ptr = max_k.value().data_handle(); }
 
+  value_t* dists_ptr = nullptr;
+  if (dists.has_value()) { dists_ptr = dists.value().data_handle(); }
```
dantegd, line 335: I think adding an assert for the unsupported case would be fine here

```
  ASSERT(!(max_k.has_value() && dists.has_value()),
         "dists output is not supported when max_k is set");

**Review by dantegd — COMMENTED** [review]
PR looks fine to me, my comments are super minor. Answers to questions from my POV:

1. I think adding sparse support only is perfectly fine, if dense support is needed then it can easily be a follow up, and we can assess if there's user demand for it. 
2. The easiest way to test this that I can think is to add a new test case in `cpp/tests/neighbors/epsilon_neighborhood.cu` alongside the existing `SparseRbc` test, reusing the same fixture (`EpsNeighRbcTest`).

The idea is straightforward: call `eps_nn` twice with the same data and parameters. The first call (without dists) gives you the CSR structure (`adj_ia`/`adj_ja`) so you know how many total neighbors were found. Then allocate a `dists` vector of that size and make the second call with it.

Once you have the results, the verification is just brute-force: for each query, loop over its neighbors in adj_ja, compute the squared Euclidean distance by hand, and check two things:

  - The returned distance matches the hand-computed one (within floating-point tolerance).
  - Every returned distance is <= eps * eps (sanity check that no out-of-range neighbors slipped through).

This gives you both correctness (distances match brute-force) and the epsilon invariant (no neighbor is farther than eps), without needing any fancy infrastructure (which is particularly annoying to write in C++ tests). 

3. I think the main consumer of this, for now at least, will be cuML so benchmarking there first is not a bad idea, since it'll tell us the actual impact and perf that users will see.

File: cpp/src/neighbors/ball_cover/registers.cuh [hunk 3]
```diff
@@ -599,6 +600,7 @@ RAFT_KERNEL block_rbc_kernel_eps_csr_pass(const value_t* X_reordered,
     // we have no neighbors to fill for this query
     if (offset == adj_ia[query_id + 1]) return;
     adj_ja += offset;
+    if (dists != nullptr) { dists += offset; }
```
jinsolp, line 604: we could cache this condition since the check `dists != nullptr` is done multiple times in this kernel!

jinsolp (commented):
Thanks @jcrist ! I am not so familiar with this algo, so just a minor comment

PR #1816, opened 2026-02-18
Speed up recall calculation in cuVS Bench for large top-K

jamxia155:
Currently, recall calculation in cuVS Bench essentially runs an outer for loop over the `k` ground-truth vector IDs and an inner loop over the `k` ANN result vector IDs, incrementing a counter if the computed value matches the ground truth. This works well assuming `k` is small but the complexity is `O(k^2)`.

When benchmarking use cases involving large `k` values, the recall calculation becomes a bottleneck especially since a large `k` does not necessarily lead to much slower search times, so the recall calculation is performed about as many times as would be for a small `k`, leading to unacceptable (or at least humanly unbearable) run times.

This update speeds up the recall calculation in the following ways:
1. Eager hashing of vector IDs
- During the construction of the dataset, we populate for each query a `std::unordered_map` of {vector_id, neighbor_rank}. This step has complexity `O(k)` and the hash maps are cached for all benchmark cases.
- During search, we look up the hash of each search result in the ground truth map to determine whether it is a true result. This step has complexity `O(k)`.
2. Parallelizing hash map build and lookup
- We use basic threading to parallelize recall calculation at the query level (for ease of implementation and cache locality).
- Care is taken to avoid oversubscribing the CPU when benchmarking is run on multiple threads e.g. in throughput mode.
3. Capping the total number of queries for which recall is calculated to about 10,000
- This avoids unbounded recall calculations if using large sets of queries and ground truths while performing many iterations of the benchmark case.
- Underlying assumption is that the sample of queries used for recall calculation will be representative of the recall performance for the benchmark case tested.

Testing at k=15000, batch-size=500, iterations=20, cpu=AMD EPYC 7413 24 cores/48 threads:
- baseline wall time: 285 s
- improved wall time: 3.7 s
- Note that the wall time includes loading and running the benchmarks, which takes over 1 s for these settings.
- Also note that if the number of iterations is not specified, the benchmark would run for over 100 iterations which would make the baseline runtime much slower as recall calculation is performed for far more than 10,000 queries.
- At k=10, wall times are 1.369 s (PR) vs. 1.362 s (baseline)


achirkin:
Hi @jamxia155 , could you please report an example of dataset+json causing the bottleneck in the recall computation?

cpp/bench/ann/src/common/benchmark.hpp
```diff
@@ -185,6 +185,9 @@ void bench_search(::benchmark::State& state,
   cuvs::bench::benchmark_thread_id = state.thread_index();
   cuvs::bench::benchmark_n_threads = state.threads();
   std::size_t queries_processed    = 0;
+  // Maximum number of queries to process for recall calculation (the actual number of recall
+  // calculations may exceed this value by up to n_queries)
+  constexpr size_t max_queries_for_recall = 10'000;
```
achirkin: The current suggestion does all the GT bookkeeping within the benchmark search case. That means
1) We load and keep all GT in the memory even though we use only first 10K rows
2) We repeat the filtering/hashset bookkeaping and allocations for every benchmark case.

Since the goal of this PR is to improve the speed of evaluation, I suggest to move this logic to the dataset header for loading:
1) Define the `constexpr size_t kMaxQueriesForRecall = 10'000` in there (NB cling-tidy for naming conventions) and use this constant to limit loading of the GT from disk.
2) *During the loading* create the GT map (ids -> neighbor rank); apply filter while loading (it's global setting per dataset, right? if so, no need to filter GT during benchmark case). No need to keep the raw GT arrays at all.
3) Use the GT map during recall calculation; discard all GT values with rank `>= k`.

Note, I'm suggesting to use the map and not the set: this allows using the same GT map for any `k` smaller than the `max_k` (that is, do a parameter sweep over `k` without recalculating the GT).

jamxia155: Updated as suggested. Would you like to take another look please? Thanks!

cpp/bench/ann/src/common/dataset.hpp
```diff
@@ -38,14 +39,20 @@ struct dataset {
...
   blob<DataT> base_set_;
   blob<DataT> query_set_;
-  std::optional<blob<IdxT>> ground_truth_set_;
   std::optional<blob<bitset_carrier_type>> filter_bitset_;
+  // Hash maps of {id, neighbor_rank} for up to kMaxQueriesForRecall queries in the ground truth set
+  // e.g. gt_maps_[i][j] = k means that the i-th query in the ground truth set has k-th nearest
+  // neighbor with id j. Note that the nearest neighbor rank starts from 0.
+  std::vector<std::unordered_map<IdxT, IdxT>> gt_maps_;
+  uint32_t max_k_ = 0;  // number of nearest neighbors in the ground truth
+  std::vector<uint32_t> filter_pass_counts_;
```
achirkin: Could you please put this into a separate structure (e.g. `struct ground_truth`) and encapsulate GT construction inside it, so it is evaluated (and ground truth is read) on the first call to it?
I'm sorry I may have said "construct ground truth eagerly in the dataset" before, but I just meant to move the code out from the benchmark loop.

The problem with the current approach is that it computes the GT even if it's not needed (e.g. for build benchmarks) and probably even throws an exception if the GT is not available. Let's try to keep the lazy semantics in the sense that the data is only accessed when it is required. Even better: maybe you could also add an `evaluate(query_id, neighbor_indices) -> recall` function to the GT object to encapsulate the GT calculation and just call it during search?

achirkin: Update: sorry, just noticed you guard everything with `groundtruth_neighbors_file.has_value()`, so it's not evaluated when GT is not provided. Yet, I think putting this in a separate struct is a good idea: later I want to extend it to compute GT using specified ANN indices on the fly.

cpp/bench/ann/src/common/dataset.hpp
```diff
@@ -38,6 +38,15 @@ struct dataset {
...
   static inline constexpr size_t kBitsPerCarrierValue = sizeof(bitset_carrier_type) * 8;
+  static constexpr uint32_t kMaxQueriesForRecall      = 10'000;
+  struct gt_entry {
+    // Hash map of {id, neighbor_rank} for a particular query in the ground truth set
+    // e.g. gt_map[i] = j means that the neighbor with id i is the j-th nearest.
+    // Note that the nearest neighbor rank starts from 0.
+    std::unordered_map<IdxT, IdxT> gt_map;
+    uint32_t filter_pass_count = 0;
+    std::once_flag once_flag;
+  };
```
achirkin: Just to clarify on my previous request: could you please put all of this in a new top-level structure that inherits from (or simply contains) `blob<IdxT>` and does all GT processing internally? So that we could replace private variable `std::optional<blob<IdxT>> ground_truth_set_` with `std::optional<groundtruth_map<IdxT>> ground_truth_map_` below with no further modifications to the dataset class compared to the main branch.

jamxia155: Created top-level `struct ground_truth_map` that computes the maps on construction and used this to replace the `ground_truth_set_` member in `dataset`. Hopefully I'm back on the right track. Thanks.

jamxia155: BTW I chickened out of moving recall calculation itself to this new struct due to the amount of refactoring work involved. Let's keep the scope of this PR to the optimization itself and focus on the refactoring in the next round.

cpp/bench/ann/src/common/dataset.hpp
```diff
@@ -33,19 +34,97 @@ void generate_bernoulli(CarrierT* data, size_t words, double p)
...
 
+template <typename T>
+struct ground_truth_map {
+  using bitset_carrier_type = uint32_t;
+  // Hash maps of {id, neighbor_rank} for up to kMaxQueriesForRecall queries in the ground truth set
+  // e.g. gt_maps_[i][j] = k means that for the i-th query in the ground truth set, the neighbor
+  // with idx j is the k-th nearest. Note that the nearest neighbor rank starts from 0.
+  std::vector<std::unordered_map<T, T>> gt_maps_;
+  uint32_t max_k_ = 0;  // number of nearest neighbors in the ground truth
+  std::vector<uint32_t> filter_pass_counts_;
```
achirkin: It seems to me  `filter_pass_counts_` duplicates the `gt_maps[i].size()`, is that right? Let's get rid of it.
```suggestion
```

cpp/bench/ann/src/common/dataset.hpp
```diff
@@ -33,19 +34,97 @@ void generate_bernoulli(CarrierT* data, size_t words, double p)
...
+    less than `k`. In the extreme case of very high filtering rate, we may be bypassing
+    entire rows of results. However, this is still better than no recall estimate at all.
+
+    */
+    auto ground_truth_set = blob<T>(file_name);
+    max_k_                = ground_truth_set.n_cols();
+    auto filter           = [&](T i) -> bool {
+      if (!filter_bitset.has_value()) { return true; }
+      auto word = filter_bitset->data(MemoryType::kHostMmap)[i >> 5];
+      return word & (1 << (i & 31));
```
achirkin, line 67: I know it's carry-over from the older code, but I bend my head trying to understand what it does, hence a comment suggestion:
```suggestion
      // bitset is `32 = bitset_carrier_type * 8` times more dense than the data
      // use bitwise arithmetic to get the `row_id` and correct bit pos in the `word`
      auto word = filter_bitset->data(MemoryType::kHostMmap)[i >> 5];
      return word & (1 << (i & 31));
```

cpp/bench/ann/src/common/dataset.hpp
```diff
@@ -33,19 +34,97 @@ void generate_bernoulli(CarrierT* data, size_t words, double p)
...
+  }
+};
+
 template <typename DataT, typename IdxT = int32_t>
 struct dataset {
  public:
-  using bitset_carrier_type                           = uint32_t;
+  using bitset_carrier_type = typename ground_truth_map<IdxT>::bitset_carrier_type;
   static inline constexpr size_t kBitsPerCarrierValue = sizeof(bitset_carrier_type) * 8;
+  static constexpr uint32_t kMaxQueriesForRecall      = 10'000;
```
achirkin: I think the `kMaxQueriesForRecall` belongs to `ground_truth_map` and you can safely move it there and do the size clamping in the constructor.

cpp/bench/ann/src/common/dataset.hpp
```diff
@@ -195,6 +283,13 @@ struct dataset {
     }
     return nullptr;
   }
+
+  [[nodiscard]] auto filter_pass_counts() const -> const uint32_t*
+  {
+    std::lock_guard<std::mutex> lock(mutex_);
+    if (ground_truth_map_.has_value()) { return ground_truth_map_->filter_pass_counts_.data(); }
+    return nullptr;
+  }
```
achirkin: Please remove, let's access this via the `gt_maps` member directly and keep dataset type more decoupled from the ground_truth_map type.

cpp/bench/ann/src/common/benchmark.hpp
```diff
@@ -369,39 +364,54 @@ void bench_search(::benchmark::State& state,
...
             for (std::uint32_t j = 0; j < k; j++) {
               auto act_idx = static_cast<std::int32_t>(neighbors_host[i_out_idx * k + j]);
-              if (act_idx == exp_idx) {
-                match_count++;
-                break;
-              }
+              if (gt_maps[i_orig_idx].count(act_idx) &&
+                  static_cast<uint32_t>(gt_maps[i_orig_idx].at(act_idx)) < k)
+                local_match_count[tid]++;
             }
```
achirkin: Let's hide just this part inside the `ground_truth_map` type:
```suggestion
          if (i_orig_idx >= gt_maps[i_orig_idx].size()) { break; }
          auto* candidates = neighbors_host + (out_offset + i)*k;
          auto [matching, total] = gt_maps[i_orig_idx].count_matches(candidates, k);
          local_match_count[tid] += matching;
          local_total_count[tid] += total;
```
Then the implementation is something along these lines:
```C++
auto count_matches(const T* candidates, T k) -> std::pair<T, T> {
  T matching = 0;
  T total = std::min(k, ground_truth_map_.size());
  for (std::uint32_t i = 0; i < k; i++) {
    auto candidate = candidates[i];
    if (auto search = ground_truth_map_.find(candidate); search != ground_truth_map_.end()) {
      matching++;
    }
  }
  return [matching, total];
}
```

cpp/bench/ann/src/common/dataset.hpp
jamxia155: Updated as suggested. Thanks.

cpp/bench/ann/src/common/dataset.hpp
jamxia155: Added suggested comment.

cpp/bench/ann/src/common/dataset.hpp
jamxia155: Moved parameter.

cpp/bench/ann/src/common/dataset.hpp
jamxia155: Updated as suggested.

cpp/bench/ann/src/common/benchmark.hpp
jamxia155: Thanks for the suggestion. Updated.

cpp/bench/ann/src/common/dataset.hpp
```diff
@@ -33,19 +34,114 @@ void generate_bernoulli(CarrierT* data, size_t words, double p)
...
 };
 
+template <typename T>
+struct ground_truth_map {
+  using bitset_carrier_type = uint32_t;
+  // Hash maps of {id, neighbor_rank} for up to kMaxQueriesForRecall queries in the ground truth set
+  // e.g. gt_maps_[i][j] = k means that for the i-th query in the ground truth set, the neighbor
+  // with idx j is the k-th nearest. Note that the nearest neighbor rank starts from 0.
+  std::vector<std::unordered_map<T, T>> gt_maps_;
+  uint32_t max_k_ = 0;  // number of nearest neighbors in the ground truth
```
achirkin: Please make these two members private

cpp/bench/ann/src/common/benchmark.hpp
```diff
@@ -369,39 +363,50 @@ void bench_search(::benchmark::State& state,
...
-            if (!filter(exp_idx)) { continue; }
-            filter_pass_count++;
-            for (std::uint32_t j = 0; j < k; j++) {
-              auto act_idx = static_cast<std::int32_t>(neighbors_host[i_out_idx * k + j]);
-              if (act_idx == exp_idx) {
-                match_count++;
-                break;
-              }
-            }
+      std::vector<std::thread> recall_calculation_workers;
```
achirkin, line 373: We have couple problems here that I overlooked in my previous reviews. In the original code, we used gbench threads only for parallel recall computation. In latency mode, the recall was always computed in one thread; in the throughput mode it was computed in the same number of threads as the main benchmark.  If you look above in the code, we get the result buffer for each thread using function `get_result_buffer_from_global_pool`. There are two things to note:
 1. The full result set is not a single contiguous buffer, but a set of buffers per gbench thread
 2. The distribution of the queries in the buffer is dictaed by the gbench batching.

Now let's look at the new multi-threaded code here.
#### Latency mode
In this mode, there's only one result buffer. The work in this buffer is distributed across multiple threads in contiguous chunks. One small problem: we cap the maximum number of queries to `kMaxQueriesForRecall = 10'000`; if the total number of queries is larger, the work is not distributed evenly. In the extreme, if we have a `320000` queries and 32 worker threads, then one thread will do all 10k calculations and the rest will stay idle.
Solution: either remove the cap (which is worse for the total calculation time) or distribute the work in strides.

#### Throughput mode
In this mode, the results are split in `--threads` contiguous chunks. The number of queries to check is now split unevenly across these chunks due to kMaxQueriesForRecall. So the new code spawns `num_recall_calculation_worker_threads * gbench_threads` (which is horrible for performance) and then likely crashes due to OOB, because it assumes a single result buffer available to all threads.
One solution to this could be to copy-gather first `kMaxQueriesForRecall` from the results buffers in each gbench thread and then initiate your multi-threaded compute logic in just one thread. 
However, by this time I can't help but wonder: maybe we can just leave the computation single-threaded as before for now (and carefully honor the original batch boundaries), and investigate how to safely add parallelism in a follow-up PR?

achirkin (changes requested):
Thanks @jamxia155 , the new ground truth maps wrapper looks really nice! Unfortunately, I've noticed one more problem with the new concurrency vs existing gbench threads that I overlooked in prior discussions.

cpp/bench/ann/src/common/dataset.hpp
jamxia155: Updated.

cpp/bench/ann/src/common/benchmark.hpp
jamxia155: Thanks for the thorough review!

Discussed your concern re: CPU oversubscription privately: already pre-empted.

Re: uneven distribution of query IDs with populated ground truth maps, I've implemented your (also private) suggestion to scatter the `kMaxQueriesForRecall` queries evenly across the full range of query IDs.

Finally, I think your concern with OOB access of the result buffer is not warranted as I was mindful of the fact that there would be `gbench_threads` threads running the search during benchmarking, each getting a contiguous chunk of results. Therefore, each of these `gbench_threads` threads is responsible for spawning worker threads that only access the specific chunk of results that their parent thread has access to. Please let me know if I'm missing something.

PR #1822, opened 2026-02-19
Update cagra config yaml for cuvs_bench

tfeher:
This PR updates the CAGRA config options for the parameter sweep used by cuvs-bench.

- `intermediate_graph_degree`: in practice we often use `intermediate_graph_degree = graph_degree * 1.5`. We pick a fixed value that works works with all of the `graph_degree `options we want to test.
- `max_iterations`: the default 0 value auto selects the iterations based on itopk. I have added value `16` additionally, which is a low number of iterations, and it can be useful to reach large throughput in recall < 0.9 region.
- `search_width`: we often reach around 99% recall with width 8, therefore I propose to remove higher values from the base scan
- `graph_build_algo`: ideally we wound want the algorithm automatically selected. Currently the benchmark framework would select the iterative method by default, which does not work with memory mapped data. The NN descent method can run out of memory for 10M scale dataset. Therefore I propose to fix `IVF_PQ`, until we enable a reliable default algorithm selection.


python/cuvs_bench/cuvs_bench/config/algos/cuvs_cagra.yaml
```diff
@@ -5,17 +5,18 @@ constraints:
 groups:
   base:
     build:
-      graph_degree: [32, 64, 96, 128]
-      intermediate_graph_degree: [32, 64, 96, 128]
-      graph_build_algo: ["NN_DESCENT"]
+      graph_degree: [32, 64, 96]
+      intermediate_graph_degree: [128]
```
tarang-jain, line 9: Dont we want intermediate_graph_degree 64 when the graph degree is 32? @bkarsin had also raised an issue in the past wherein CAGRA can give poor recall when the intermediate graph degree is quite large in comparison to the graph degree.

tarang-jain (commented):
LGTM except one comment.

python/cuvs_bench/cuvs_bench/config/algos/cuvs_cagra.yaml
tfeher: Ideally yes, `intermediate_graph_degree = 2 * graph_degree` would be sufficient. Our benchmark framework does not adjust this automatically. I would not expect recall drop by using larger intermediate degree. Do we have any detail on that that issue?

PR #1825, opened 2026-02-19
Fix thrust header

aamijar:
Follow up to https://github.com/rapidsai/raft/pull/2958.

### DIFF
--- a/cpp/src/neighbors/detail/knn_brute_force.cuh
+++ b/cpp/src/neighbors/detail/knn_brute_force.cuh
@@ -42,7 +42,6 @@
 #include <rmm/device_uvector.hpp>
 #include <thrust/for_each.h>
 #include <thrust/iterator/counting_iterator.h>
-#include <thrust/iterator/transform_iterator.h>
 
 #include <cstdint>
 #include <iostream>


cpp/src/neighbors/detail/knn_brute_force.cuh
```diff
@@ -40,6 +40,7 @@
 #include <cuda_fp16.h>
 #include <rmm/cuda_device.hpp>
 #include <rmm/device_uvector.hpp>
+#include <thrust/iterator/counting_iterator.h>
```
bdice, line 44: This is missing the header for `thrust::for_each`. Please audit RAFT / cuVS and friends for this kind of issue.
```suggestion
#include <thrust/for_each.h>
#include <thrust/iterator/counting_iterator.h>
```

cpp/src/neighbors/detail/knn_brute_force.cuh
```diff
@@ -40,6 +40,7 @@
...
 #include <thrust/iterator/transform_iterator.h>
```
bdice: Where are transform iterators used in this file?

aamijar: Thanks, removed it in https://github.com/rapidsai/cuvs/pull/1825/commits/4b76753baa840f3c0ca3e39b519cbadfaf7b0332

mythrocks:
~~I'm curious whether this addresses the errors mentioned in #1826. It didn't seem to in my build, but that likely wasn't the intention either.~~

Per discussion with @aamijar, this was because the change to include `<thrust/iterator/counting_iterator.h>` wasn't included in #1771.  I've left #1826 up, in case the move to `cuda::make_counting_iterator` is of value.

aamijar:
Thanks @mythrocks! Happy to merge #1826, as a follow up too. Or we can use yours if it passes CI first 😄

mythrocks:
@aamijar:  I think yours is less intrusive, and less controversial.  I vote that #1825 ship first.  #1826 might need examination by @cjnolet / @achirkin.

PR #1827, opened 2026-02-19
Expose supported brute force metrics in `all_neighbors`

### LINKED ISSUE #1715: [FEA] Expose supported metrics for brute force in `all_neighbors`
Brute force supports multiple metrics, but only a few are exposed through the all-neighbors api.

https://github.com/rapidsai/cuvs/blob/0cae1a66e1d1c6537d56f3a4171ebb5317a1ceb5/cpp/src/neighbors/all_neighbors/all_neighbors.cuh#L32-L39

jinsolp:
Closes https://github.com/rapidsai/cuvs/issues/1715

Exposes supported brute force metrics in `all_neighbors` API.


cpp/src/neighbors/all_neighbors/all_neighbors.cuh
```diff
@@ -32,10 +32,20 @@ GRAPH_BUILD_ALGO check_params_validity(const all_neighbors_params& params,
...
-                   "L2SqrtExpanded, CosineExpanded, or InnerProduct.");
+                             params.metric == cuvs::distance::DistanceType::L1 ||
+                             params.metric == cuvs::distance::DistanceType::L2Unexpanded ||
+                             params.metric == cuvs::distance::DistanceType::L2SqrtUnexpanded ||
+                             params.metric == cuvs::distance::DistanceType::InnerProduct ||
+                             params.metric == cuvs::distance::DistanceType::Linf ||
+                             params.metric == cuvs::distance::DistanceType::Canberra ||
+                             params.metric == cuvs::distance::DistanceType::LpUnexpanded ||
+                             params.metric == cuvs::distance::DistanceType::CorrelationExpanded ||
+                             params.metric == cuvs::distance::DistanceType::JensenShannon;
```
tarang-jain: Optional micro-optimization: It would probably make sense to put these metrics into an unordered set or list and check for presence with O(1)

python/cuvs/cuvs/tests/test_all_neighbors.py
```diff
@@ -57,6 +82,21 @@ def test_all_neighbors_device_build_quality(algo, cluster, metric):
...
+            centers=10,
+            cluster_std=1.0,
+            center_box=(0.0, 10.0),  # Non-negative values only
+            random_state=42,
+        )
+        # Normalize each row to sum to 1 (probability distribution)
+        X = np.abs(X)  # Ensure non-negative
+        row_sums = X.sum(axis=1, keepdims=True)
+        row_sums[row_sums == 0] = 1  # Avoid division by zero
+        X = X / row_sums
```
tarang-jain, line 91: Nit: sklearn to the rescue?
```
from sklearn.preprocessing import normalize
X = normalize(np.abs(X), norm='l1', axis=1)
```

tarang-jain: I think this comment might be a bit too much nitpicking. Marking this as resolved.

python/cuvs/cuvs/tests/test_all_neighbors.py
```diff
@@ -36,15 +36,40 @@ def make_cosine(
...
+        )
+    elif algo == "nn_descent" and metric not in [
+        "sqeuclidean",
+        "l2",
+        "cosine",
+        "inner_product",
+    ]:
+        pytest.skip(
+            "Skipping NN-Descent for distance metrics other than sqeuclidean, l2, cosine, or inner_product"
+        )
```
viclafargue: Could we check that the errors are raised as expected with correct message?

cpp/src/neighbors/all_neighbors/all_neighbors.cuh
```diff
@@ -9,52 +9,60 @@
...
+                                                            DT::CosineExpanded,
+                                                            DT::L1,
+                                                            DT::L2Unexpanded,
+                                                            DT::L2SqrtUnexpanded,
+                                                            DT::InnerProduct,
+                                                            DT::Linf,
+                                                            DT::Canberra,
+                                                            DT::LpUnexpanded,
+                                                            DT::CorrelationExpanded,
+                                                            DT::JensenShannon};
```
viclafargue, line 38: Small detail, a `switch` statement might be slightly faster than building and searching a `std::unordered_set`, but the `std::unordered_set` probably offers cleaner code.

viclafargue (approved):
Thanks! Not familiar with which metric would be supported by the different algorithms. However, the code change in itself looks good to me.

cpp/src/neighbors/all_neighbors/all_neighbors.cuh
jinsolp: thanks for the suggestion, but for readability i prefer to stick to the unordered sets!

tarang-jain (approved):

viclafargue (approved):
Thanks Jinsol! The change looks good to me.

Just one question regarding the choice of allowed metrics : were `Hamming`, `Hellinger`, `KLDivergence`, and `RusselRao` metrics omitted deliberately?

jinsolp:
@viclafargue they were left out intentionally! The metrics exposed and added in this PR are the ones being tested in `brute_force.cu`.

https://github.com/rapidsai/cuvs/blob/90c18a80b4ed485e671910d935321628a7abe785/cpp/tests/neighbors/brute_force.cu#L495-L539

cjnolet (approved):

PR #1830, opened 2026-02-20
[REVIEW] Generalize and improve cagra::optimize

mfoerste4:
In preparation for large scale graph creation this PR adds several changes to cagra:optimize by:

* adding full device path for pruning and merging / discarding host fallback code
* fusing two pruning steps into one batch-able kernel, reducing memory requirements
* batched reverse graph creation
* batched merging

Due to the batching in all substeps the memory footprint could even be decreased while significantly improving computation time.

The optimize API now supports all variations of memory locations for _knn_graph_ and _cagra_graph_.
Internally, the data will be buffered in device memory for best performance. Directly accessing managed/pinned/HMM memory from the device showed severe performance degradation upon the first access (x86/H200 with HMM):

```
=== Benchmarks (256 MiB, 10 iterations) ===

  [malloc] 1. Copy to device:     133.783 ms total (10 iters) -> 18.69 GB/s
  [malloc] 2. Kernel direct read: 4648.468 ms total (10 iters) -> 0.54 GB/s
  [malloc] 3. Kernel subsequent read: 15.164 ms total (10 iters) -> 164.87 GB/s
  [cudaMalloc] 1. Copy to device:     1.294 ms total (10 iters) -> 1932.35 GB/s
  [cudaMalloc] 2. Kernel direct read: 14.945 ms total (10 iters) -> 167.28 GB/s
  [cudaMalloc] 3. Kernel subsequent read: 14.963 ms total (10 iters) -> 167.07 GB/s
  [cudaMallocHost] 1. Copy to device:     95.002 ms total (10 iters) -> 26.32 GB/s
  [cudaMallocHost] 2. Kernel direct read: 290.486 ms total (10 iters) -> 8.61 GB/s
  [cudaMallocHost] 3. Kernel subsequent read: 290.789 ms total (10 iters) -> 8.60 GB/s
  [cudaMallocManaged] 1. Copy to device:     136.737 ms total (10 iters) -> 18.28 GB/s
  [cudaMallocManaged] 2. Kernel direct read: 766.153 ms total (10 iters) -> 3.26 GB/s
  [cudaMallocManaged] 3. Kernel subsequent read: 15.002 ms total (10 iters) -> 166.65 GB/s
```

New kernels are based on experiments by @bpark-nvidia

CC @tfeher , @irina-resh-nvda


bpark-nvidia:
I have a one question about `batched_device_view_from_host` in `kern_fused_prune` function. The nodes that are accessed during each iteration is a set of two-hop neighbors of the nodes in the batch, but if I understand the code correctly, only nodes in the batch are prefetched. Could you explain how these neighbors are accessed?

cpp/src/neighbors/detail/cagra/utils.hpp
```diff
@@ -294,4 +316,395 @@ void copy_with_padding(
...
+ * Utility to create a batched device view from a host view
+ *
+ * This utility will create a batched device view from a host view and will handle the prefetch and
+ * writeback of the data Each batch can be referenced exactlyonce by calling the next_view()
+ * function
+ *
+ * Usage:
+ * ```
+ * batched_device_view_from_host<float, int32_t> view(res, host_view, batch_size, host_writeback,
+ * initialize); while (view.next_view().extent(0) > 0) { auto device_view = view.next_view();
```
irina-resh-nvda: Intuitively, vew.next_vew() should change the state of the object, so calling it twice (once in the loop condition, once in the body) feels like we are skipping a batch. If we are not, then some kind of state change like view = device_view is missing from the usage example

cpp/src/neighbors/detail/cagra/utils.hpp
```diff
@@ -294,4 +316,395 @@ void copy_with_padding(
...
+      batch_size_(batch_size),
+      offset_(0),
+      batch_id_(-2),
+      num_buffers_(2),
+      host_writeback_(host_writeback),
+      initialize_(initialize)
+  {
+    if (host_view.extent(0) == 0) {
+      mem_strategy_ = memory_strategy::device_only;
+      return;
```
irina-resh-nvda: Is there a reason not to raise an error with `RAFT_EXPECTS` in this case?

cpp/src/neighbors/detail/cagra/utils.hpp
```diff
@@ -294,4 +316,395 @@ void copy_with_padding(
...
+        }
+        if (host_writeback_ && initialize_ &&
+            batch_size * 2 < static_cast<uint64_t>(host_view.extent(0))) {
+          num_buffers_ = 3;
+          device_mem_[2].emplace(raft::make_device_mdarray<T, IdxT>(
+            res,
+            raft::resource::get_workspace_resource(res),
+            raft::make_extents<int64_t>(batch_size, host_view.extent(1))));
+          device_ptr[2] = device_mem_[2]->data_handle();
+        }
```
irina-resh-nvda: if both ifs are false, num_buffers_ should be set to 1

cpp/src/neighbors/detail/cagra/utils.hpp
```diff
@@ -294,4 +316,395 @@ void copy_with_padding(
...
+      int32_t prefetch_pos             = (batch_id_ + 1) % num_buffers_;
+      actual_batch_size_[prefetch_pos] = min(batch_size_, host_view_.extent(0) - offset_);
+
+      switch (mem_strategy_) {
+        case memory_strategy::managed_only:
+          if (!host_writeback_ && batch_id_ > 1) {
+            uint32_t discard_pos = (batch_id_ - 1) % num_buffers_;
+            size_t discard_size  = batch_size_ * host_view_.extent(1) * sizeof(T);
+            discard_managed_region(device_ptr[discard_pos], discard_size);
+          }
```
irina-resh-nvda: If I understand it correctly, in case of`mem_strategy_ == memory_strategy::managed_only` but `host_writeback_==true` we rely on implicit managed memory coherence for the writeback cycle.
This is potentially very slow.
Wouldn't `cudaMemPrefetchAsync ` with `location.type = cudaMemLocationTypeHost` help?
If not, maybe you can add a comment explaining why it doesn't make sense to initiate an explicit writeback-like cycle in this case

cpp/src/neighbors/detail/cagra/graph_core.cuh
```diff
@@ -165,87 +171,266 @@ __global__ void kern_sort(const DATA_T* const dataset,  // [dataset_chunk_size,
...
   uint64_t* const num_full   = stats + 1;
 
-  const uint64_t iA = blockIdx.x + (batch_size * batch_id);
-  if (iA >= graph_size) { return; }
-  for (uint32_t k = threadIdx.x; k < graph_degree; k += blockDim.x) {
+  const uint32_t maxval16 = 0x0000ffff;
+
+  const uint32_t nid_batch = blockIdx.x * num_warps + wid;
+  const uint64_t nid       = static_cast<uint64_t>(nid_batch) +
+                       (static_cast<uint64_t>(batch_size) * static_cast<uint64_t>(batch_id));
```
irina-resh-nvda, line 241: This indexing heavily relies on the user of the kernel to call it with the right batch_size and batch_id. Without any checks, there will be indexing overlap resulting in duplicated computations and maybe other issues down the line. For example, for some combination of input parameters nid will be exactly the same. For example, if batch_size = 16, then for blockIdx.x = 0 and wid =0 and batch_id = 1, nid will be the same as for blockIdx.x = 4, wid = 0 and batch id = 0.
It doesn't happen because you compute the grid size as ceildiv(16, 4), but the kernel itself doesn't validate this.
There is also an issue if batch_size isn't divisible by 4 (there will be an overlap of computations, too), which doesn't happen because the grid size is either a constant that is divisible by 4, or there is only one batch.
I think this nuance deserves a comment, and the functions itself should be clearly marked as internal only (maybe `_kern_fused_prune`). Or the launch requirements need to be enforced.

cpp/src/neighbors/detail/cagra/graph_core.cuh
```diff
@@ -1069,94 +1464,153 @@ void mst_optimization(raft::resources const& res,
...
+    /*host_writeback*/ true,
+    /*initialize*/ false);
+
+  auto d_invalid_neighbor_list = raft::make_device_scalar<uint32_t>(res, 0u);
+
+  for (uint32_t i_batch = 0; i_batch < num_batch; i_batch++) {
+    auto output_view         = d_output_graph.next_view();
+    const uint32_t num_warps = 4;
+    const dim3 threads_prune(raft::WarpSize * num_warps, 1, 1);
+    const dim3 blocks_prune(raft::ceildiv(batch_size, num_warps), 1, 1);
```
irina-resh-nvda: For `kern_fused_prune`, the requirement of `batch_size` being divisible by 4 is critical if we want to avoid computation overlap between different batch_ids. I think it makes sense to have a RAFT enforcement here

cpp/src/neighbors/detail/cagra/graph_core.cuh
```diff
@@ -1069,94 +1464,153 @@ void mst_optimization(raft::resources const& res,
...
+  const uint64_t graph_size          = output_graph.extent(0);
+  const uint64_t knn_graph_degree    = knn_graph.extent(1);
+  const uint64_t output_graph_degree = output_graph.extent(1);
+
+  raft::common::nvtx::range<cuvs::common::nvtx::domain::cuvs> block_scope(
+    "cagra::graph::optimize/prune");
+  auto default_ws_mr = raft::resource::get_workspace_resource(res);
+
+  uint32_t batch_size =
+    std::min(static_cast<uint32_t>(graph_size), static_cast<uint32_t>(256 * 1024));
```
irina-resh-nvda: `static_cast<uint32_t>(256 * 1024)` <- this constant needs to be divisible by 4 for `kern_fused_prune` internal indexing. This line needs a comment for maintainability

cpp/src/neighbors/detail/cagra/graph_core.cuh
```diff
@@ -165,87 +171,266 @@ __global__ void kern_sort(const DATA_T* const dataset,  // [dataset_chunk_size,
...
+  __syncwarp();
+
+  num_edges_no_detour = cg::reduce(warp, num_edges_no_detour, cg::plus<uint32_t>());
+  num_edges_no_detour = min(num_edges_no_detour, output_graph_degree);
+
+  if (lane_id == 0) {
     atomicAdd((unsigned long long int*)num_retain, (unsigned long long int)num_edges_no_detour);
-    if (num_edges_no_detour >= degree) { atomicAdd((unsigned long long int*)num_full, 1); }
+    if (num_edges_no_detour >= output_graph_degree) {
+      atomicAdd((unsigned long long int*)num_full, 1);
```
irina-resh-nvda, line 292: num_full is only used for debug output, if i understand it correctly.
An atomic add just for debug stats is not ideal. Maybe you could guard these with a compile-time debug flag?

cpp/src/neighbors/detail/cagra/graph_core.cuh
```diff
@@ -165,87 +171,266 @@ __global__ void kern_sort(const DATA_T* const dataset,  // [dataset_chunk_size,
...
+  while (kr) {
+    kr -= 1;
+    if (rev_graph(nid, kr) < graph_size) {
+      uint64_t pos =
+        warp_pos_in_array<IdxT>(rev_graph(nid, kr), smem_sorted_output_graph, output_graph_degree);
+      if (pos < num_protected_edges) { continue; }
+      uint64_t num_shift = pos - num_protected_edges;
+      if (pos >= output_graph_degree) { num_shift = output_graph_degree - num_protected_edges - 1; }
+      if (lane_id == 0) {
+        thread_shift_array<IdxT>(smem_sorted_output_graph + num_protected_edges, num_shift);
```
irina-resh-nvda: can this shift step not be parallelized using the rest of the warp?

cpp/src/neighbors/detail/cagra/graph_core.cuh
```diff
@@ -165,87 +171,266 @@ __global__ void kern_sort(const DATA_T* const dataset,  // [dataset_chunk_size,
...
+        smem_sorted_output_graph[num_protected_edges] = rev_graph(nid, kr);
```
irina-resh-nvda: rev_graph(nid, kr) is referenced 3 times in this loop, maybe it's worth it to assign the value to a variable

cpp/src/neighbors/detail/cagra/graph_core.cuh
```diff
@@ -482,23 +667,225 @@ __global__ void kern_mst_opt_postprocessing(IdxT* outgoing_num_edges,  // [graph
...
+  const uint64_t output_graph_degree = output_graph.extent(1);
+
+  raft::common::nvtx::range<cuvs::common::nvtx::domain::cuvs> block_scope(
+    "cagra::graph::optimize/combine");
+
+  auto default_ws_mr             = raft::resource::get_workspace_resource(res);
+  const double merge_graph_start = cur_time();
+
+  auto d_check_num_protected_edges = raft::make_device_scalar<bool>(res, true);
+  auto d_invalid_neighbor_list     = raft::make_device_scalar<uint32_t>(res, 0u);
```
irina-resh-nvda: This doesn't seem to be used

cpp/src/neighbors/detail/cagra/graph_core.cuh
```diff
@@ -1069,94 +1464,153 @@ void mst_optimization(raft::resources const& res,
...
-  const bool guarantee_connectivity = true,
-  const bool use_gpu                = true)
+}  // namespace
+
+template <typename IdxT = uint32_t, typename InputMatrixView, typename OutputMatrixView>
+void optimize(raft::resources const& res,
+              InputMatrixView knn_graph,
+              OutputMatrixView new_graph,
+              const bool guarantee_connectivity = true,
+              const bool use_gpu                = true)
```
irina-resh-nvda: use_gpu=false will still use gpu in all steps except for mst_optimization. Maybe it's better to rename the parameter, or at least have a comment

cpp/src/neighbors/detail/cagra/utils.hpp
mfoerste4: Yes, you're right. This is a bug in the example.

cpp/src/neighbors/detail/cagra/utils.hpp
mfoerste4: Yes, there are optional data structures (like the mst graph) which will have zero extents and a nullptr to the data.

cpp/src/neighbors/detail/cagra/utils.hpp
mfoerste4: Agreed - although this is would be somewhat strange use-case where we don't initialize and don't write back. Something like a temporary scratch would not need to iterate.

mfoerste4: Ah, i even had that checked already:
```
RAFT_EXPECTS(host_writeback_ || initialize_,
                 "At least one of host_writeback or initialize must be true");
```

cpp/src/neighbors/detail/cagra/utils.hpp
mfoerste4: I am not sure what the best approach is here. The purpose of this call is to free up device memory where we know we can. A async prefetch to host would not achieve this to my understanding.
Admittedly I would hope to get rid of the whole managed part as it is only used as a fallback I am not even sure we need to support.

cpp/src/neighbors/detail/cagra/graph_core.cuh
mfoerste4: I have added some checks to guard against that.

cpp/src/neighbors/detail/cagra/graph_core.cuh
mfoerste4: I added guards to protect here as well.

cpp/src/neighbors/detail/cagra/graph_core.cuh
mfoerste4: Yes, same with `num_retain `which is triggered even more often. Guarded for debug builds now.

cpp/src/neighbors/detail/cagra/graph_core.cuh
mfoerste4: Yep, good idea

cpp/src/neighbors/detail/cagra/cagra_build.cuh
```diff
@@ -833,27 +833,26 @@ inline std::pair<size_t, size_t> optimize_workspace_size(size_t n_rows,
 
   // Prune stage memory
   // We neglect 8 bytes (both on host and device) for stats
-  size_t prune_host = n_rows * intermediate_degree * sizeof(uint8_t);  // detour count
+  size_t batch_size = std::min(static_cast<size_t>(256 * 1024), n_rows);
```
achirkin, line 835: How is this constant selected? What bounds the batch size: is it the workspace memory or large workspace memory or the host memory or something else? The workspace memory would be special case, because for it you could query the limit/available memory and calculate the batch size dynamically. For all other types I support the idea of having a constant, but it would be nice to add a small comment explaining where the batch lives and why the size is selected this way (also would be nice to not have this a root cause for unexplained OOMs in the next few years :) .

cpp/src/neighbors/detail/cagra/utils.hpp
```diff
@@ -155,6 +161,22 @@ struct gen_index_msb_1_mask {
 };
 }  // namespace utils
 
+template <typename T>
+bool is_ptr_device_accessible(T* ptr)
```
achirkin: Is this used anywhere in the code?

cpp/src/neighbors/detail/cagra/graph_core.cuh
```diff
@@ -482,23 +688,226 @@ __global__ void kern_mst_opt_postprocessing(IdxT* outgoing_num_edges,  // [graph
...
+
+  //
+  // Make reverse graph
+  //
+  const double time_make_start = cur_time();
+
+  raft::matrix::fill(res, d_rev_graph, IdxT(-1));
+  raft::matrix::fill(res, d_rev_graph_count, uint32_t(0));
+
+  if (is_ptr_host_accessible(output_graph.data_handle())) {
```
achirkin: This looks a little bit puzzling to me. You seem to test the host-accessibility as a means of whether it is NOT accessible from the device?
Let's go through the possible `output_graph` pointer states and figure out what do we actually want here:
- Unregistered memory, not-device-accessible -> definitely need to copy
- Registered host, but not-host-device-unified -> accessible, but slow. Probably want to copy?
- Unified memory -> accessible from both host and device, no faster alternative -> don't want to copy
- device memory -> don't want to copy
- Mirgatable UVM (starts on host, too big to fit)?...

cpp/src/neighbors/detail/cagra/graph_core.cuh
```diff
@@ -482,23 +688,226 @@ __global__ void kern_mst_opt_postprocessing(IdxT* outgoing_num_edges,  // [graph
...
+  RAFT_EXPECTS(check_num_protected_edges,
+               "Failed to merge the MST, pruned, and reverse edge graphs. "
+               "Some nodes have too "
+               "many MST optimization edges.");
+
+  RAFT_LOG_DEBUG("# Time for merging graphs: %.1lf ms",
+                 (merge_graph_end - merge_graph_start) * 1000.0);
+}
+
+template <typename IdxT, typename OutputMatrixView>
```
achirkin: I like that the OutputMatrixView is a template here. We can encode the host-device accessibility via types then. This has an advantage over runtime checks: by encoding the accessibility in the types, we inform about the actual intention of the caller on whether they want to copy the data to GPU or use it as-is, rather than relying on runtime checks and heuristics over hardware peculiarities.

cpp/src/neighbors/detail/cagra/graph_core.cuh
```diff
@@ -848,9 +1254,20 @@ void mst_optimization(raft::resources const& res,
       }
     } else {
       // Copy rank-k edges from the input knn graph to 'candidate_edges'
+      if (is_ptr_host_accessible(input_graph.data_handle())) {
```
achirkin: Less relevant than `output_graph`, but still worth checking: maybe we could instead use the compile-time information from the InputMatrixView template?
Also maybe it's worth delegating this matter to `raft::copy(res, candidate_edges.view(), {submdspan_or_whatever}(input_graph))`?

cpp/src/neighbors/detail/cagra/utils.hpp
```diff
@@ -155,6 +161,22 @@ struct gen_index_msb_1_mask {
...
+  return attr.devicePointer != nullptr;
+}
+
+template <typename T>
+bool is_ptr_host_accessible(T* ptr)
+{
+  cudaPointerAttributes attr;
+  RAFT_CUDA_TRY(cudaPointerGetAttributes(&attr, ptr));
+  return attr.hostPointer != nullptr;
+}
```
achirkin: It's sad we have many ways to check residency/accessibility across raft and cuvs yet there are edge cases where more helpers like this are needed.

Though I'm not sure we do need these two helpers here. Can we leave on with only type-level checks? Or maybe the pointer residency helpers in raft would be acceptable in all places you use this?

cpp/src/neighbors/detail/cagra/graph_core.cuh
```diff
@@ -165,87 +171,287 @@ __global__ void kern_sort(const DATA_T* const dataset,  // [dataset_chunk_size,
...
+
+  const uint64_t graph_size          = rev_graph.extent(0);
+  const uint32_t rev_graph_degree    = rev_graph.extent(1);
+  const uint32_t output_graph_degree = output_graph.extent(1);
+
+  for (uint64_t src_id = tid; src_id < graph_size; src_id += tnum) {
+    IdxT dest_id = output_graph(src_id, k);
+    if (dest_id >= graph_size) continue;
+
+    const uint32_t pos = atomicAdd(&rev_graph_count(dest_id), 1);
```
achirkin, line 192: a nitpick: use fancy `cuda::std::atomic<T>::fetch_add`

cpp/src/neighbors/detail/cagra/utils.hpp
```diff
@@ -294,4 +316,398 @@ void copy_with_padding(
...
+   * @param batch_size The batch size
+   * @param host_writeback Whether to write back the data to the host (only for host memory)
+   * (default: false)
+   * @param initialize Whether to initialize the data (default: true)
+   */
+  batched_device_view_from_host(raft::resources const& res,
+                                raft::host_matrix_view<T, IdxT> host_view,
+                                uint64_t batch_size,
+                                bool host_writeback = false,
+                                bool initialize     = true)
```
achirkin: Could you please comment on why we cannot use [batch_load_iterator](https://github.com/rapidsai/cuvs/blob/c9992089470bd03879e9dfc5e618964330477815/cpp/src/neighbors/detail/ann_utils.cuh#L405) here or adapt it to be used in both places?

cpp/src/neighbors/detail/cagra/graph_core.cuh
```diff
@@ -482,23 +688,226 @@ __global__ void kern_mst_opt_postprocessing(IdxT* outgoing_num_edges,  // [graph
...
+    // output graph is fully device accessible, so we need no copy to device
+    dim3 threads(256, 1, 1);
+    dim3 blocks(1024, 1, 1);
+    for (uint64_t k = 0; k < output_graph_degree; k++) {
+      kern_make_rev_graph_k<<<blocks, threads, 0, raft::resource::get_cuda_stream(res)>>>(
+        output_graph, d_rev_graph, d_rev_graph_count, k);
+    }
+  }
+
+  raft::resource::sync_stream(res);
```
achirkin: I think it could make sense to move the sync out of the function definition and put it after the function at the place where it's called. This may help to better understand the device-host sync behavior on a higher level (or maybe remove the sync/combine it with other syncs later).

cpp/src/neighbors/detail/cagra/graph_core.cuh
```diff
@@ -482,23 +688,226 @@ __global__ void kern_mst_opt_postprocessing(IdxT* outgoing_num_edges,  // [graph
...
+                     raft::host_vector_view<uint32_t, int64_t> mst_graph_num_edges,
+                     bool guarantee_connectivity)
+{
+  const uint64_t graph_size          = output_graph.extent(0);
+  const uint64_t output_graph_degree = output_graph.extent(1);
+
+  raft::common::nvtx::range<cuvs::common::nvtx::domain::cuvs> block_scope(
+    "cagra::graph::optimize/combine");
+
+  auto default_ws_mr             = raft::resource::get_workspace_resource(res);
```
achirkin: Please use `raft::resource::get_workspace_resource_ref` here and in the other places, the non-refs will soon be removed https://github.com/rapidsai/cuvs/pull/1990.

achirkin (changes requested):
Thank you Malte for the PR, this refactoring of `optimize()` is long overdue in cuVS :)

cpp/src/neighbors/detail/cagra/cagra_build.cuh
mfoerste4: The batches are allocated from the default workspace memory and chosen to be reasonably small to not eat up too much memory but also sufficiently large to saturate all SMs. When choosing this dynamically at runtime - should we work with a minimum here in the approximation?

cpp/src/neighbors/detail/cagra/utils.hpp
mfoerste4: Not anymore. Removed it.

cpp/src/neighbors/detail/cagra/graph_core.cuh
mfoerste4: Could you explain how you would want to decide that? We would still want to be able to distinguish pinned/UMM/managed memory from plain device memory.

cpp/src/neighbors/detail/cagra/graph_core.cuh
mfoerste4: Regarding the runtime check - same question as before.

Agree with the second part - will try to change to submdspan.

mfoerste4: Actually, the raft::copy does not seem to support cuda::std::submdspan.

cpp/src/neighbors/detail/cagra/utils.hpp
mfoerste4: I did not see any similar helpers in raft -- could you point me to where they are located?

cpp/src/neighbors/detail/cagra/graph_core.cuh
mfoerste4: IIUC this would require the global array to be initialized with type cuda::std::atomic<uint32_t> and be kept this way all its lifetime? Once written its being used read-only. Please confirm - I have just not seen this being used with global device memory before.

cpp/src/neighbors/detail/cagra/utils.hpp
mfoerste4: The two mechanisms differ in a few attributes:
* support of write-back to source
* different heuristic on when to copy
* fallback support for managed memory
* support of reset / iterator API

If we drop the managed memory support it would be easier to merge.

cpp/src/neighbors/detail/cagra/graph_core.cuh
mfoerste4: I have tried several different allocation spaces (see top of PR) - to me all spaces that are not plain device memory perform better when copied to device manually compared to other mechanisms. 
So anything accessible from host `is_ptr_host_accessible(output_graph.data_handle())` is worth copying explicitly.

achirkin: That is likely true on the machine you tested :) Consider DGX Spark or Grace/Vera: all device memory is host accessible and will be copied unconditionally. Hence I think using the type-level enforcement is a better fit here: the user of the code declares the intent: the array is device accessible or not (independently of what kind of pointer it actually is).

cpp/src/neighbors/detail/cagra/graph_core.cuh
achirkin: You control the entry point here, right? Somewhere in the CAGRA build process; then you can just create a new mdspan wrapping your pointer with a proper accessor type (e.g. use make_device_mdspan).

cpp/src/neighbors/detail/cagra/utils.hpp
achirkin: The closest by the mechanism is [raft/core/memory_type.hpp](https://github.com/rapidsai/raft/blob/1a5f667dfdca2eba854443196dd193acaddf1891/cpp/include/raft/core/memory_type.hpp#L76).

cpp/src/neighbors/detail/cagra/graph_core.cuh
mfoerste4: I re-wrote and simplified the code to work with generalized mdspan and templatized accessor.

cpp/src/neighbors/detail/cagra/utils.hpp
mfoerste4: I re-wrote and simplified the utility to work with generalized mdspan and templatized accessor.

cpp/src/neighbors/detail/cagra/cagra_build.cuh
achirkin: Could you please write this  as a small note in the comments to the code?

cpp/src/neighbors/detail/cagra/graph_core.cuh
achirkin: You're right.. we could just add a cast to the atomic type, but that doesn't sound better.

cpp/src/neighbors/detail/cagra/graph_core.cuh
```diff
@@ -165,87 +171,302 @@ __global__ void kern_sort(const DATA_T* const dataset,  // [dataset_chunk_size,
...
-  num_edges_no_detour += __shfl_xor_sync(0xffffffff, num_edges_no_detour, 4);
-  num_edges_no_detour += __shfl_xor_sync(0xffffffff, num_edges_no_detour, 8);
-  num_edges_no_detour += __shfl_xor_sync(0xffffffff, num_edges_no_detour, 16);
-  num_edges_no_detour = min(num_edges_no_detour, degree);
 
-  if (threadIdx.x == 0) {
-    num_no_detour_edges[iA] = num_edges_no_detour;
+  __syncwarp();
+
+  num_edges_no_detour = cg::reduce(warp, num_edges_no_detour, cg::plus<uint32_t>());
```
achirkin, line 285: That is fancy! 👍
Shall we also replace `__syncwarp()` with `warp.sync()` then?

cpp/src/neighbors/detail/cagra/graph_core.cuh
```diff
@@ -165,87 +171,302 @@ __global__ void kern_sort(const DATA_T* const dataset,  // [dataset_chunk_size,
...
-    if (knn_graph[k + ((uint64_t)graph_degree * iA)] == iA) {
+    smem_indices[k]    = knn_graph(nid, k);
+    if (smem_indices[k] == nid) {
       // Lower the priority of self-edge
-      smem_num_detour[k] = graph_degree;
+      smem_num_detour[k] = knn_graph_degree;
     }
   }
-  __syncthreads();
+  __syncwarp();
```
achirkin: Is the warp-level sync in place of the previous block-level sync really sufficient here?

cpp/src/neighbors/detail/cagra/graph_core.cuh
```diff
@@ -165,87 +171,302 @@ __global__ void kern_sort(const DATA_T* const dataset,  // [dataset_chunk_size,
...
+    const int64_t chunk_start    = (chunk_start_lo > 1) ? chunk_start_lo : 1;
+    const int64_t k              = chunk_start + static_cast<int64_t>(lane_id);
+    T val{};
+    const bool active = (k <= chunk_end);
+    if (active) { val = array[k - 1]; }
+    __syncwarp();
+    if (active) { array[k] = val; }
+    __syncwarp();
+  }
+}
```
achirkin: Perhaps a nitpick, but I feel like this could be faster/better expressed with a single `array[k] = raft::shfl_up(array[k], 1)` in place of two syncwarps.

cpp/src/neighbors/detail/cagra/graph_core.cuh
```diff
@@ -165,87 +171,302 @@ __global__ void kern_sort(const DATA_T* const dataset,  // [dataset_chunk_size,
...
+      const auto v     = output_graph(nid_batch, pruned_j);
+      unsigned int dup = 0;
+      for (uint32_t m = lane_id; m < output_j; m += raft::WarpSize) {
+        if (v == smem_sorted_output_graph[m]) {
+          dup = 1;
+          break;
+        }
+      }
+
+      unsigned int warp_dup = __ballot_sync(0xffffffff, dup);
```
achirkin: Another nitpick: have a look at `raft/util/warp_primitives.cuh` for more neat helpers for your kernels
```suggestion
      auto warp_dup = raft::ballot(dup);
```

cpp/src/neighbors/detail/cagra/graph_core.cuh
```diff
@@ -327,7 +549,7 @@ __global__ void kern_mst_opt_update_graph(IdxT* mst_graph,  // [graph_size, grap
     // Check to avoid duplication
     for (uint64_t kl = 0; kl < graph_degree; kl++) {
       uint64_t m = mst_graph[(graph_degree * l) + kl];
-      if (m > graph_size) continue;
+      if (m >= graph_size) continue;
```
achirkin, line 552: Good catch!

cpp/src/neighbors/detail/cagra/graph_core.cuh
```diff
@@ -482,23 +706,227 @@ __global__ void kern_mst_opt_postprocessing(IdxT* outgoing_num_edges,  // [graph
   }
 }
 
-template <class T>
-uint64_t pos_in_array(T val, const T* array, uint64_t num)
+template <typename IdxT>
+void log_incoming_edges_histogram(const IdxT* output_graph_ptr,
+                                  uint64_t graph_size,
+                                  uint64_t output_graph_degree)
+{
```
achirkin: Do I understand it right, that the only effect of this function is the printed output (which is disabled unless we enable debug logging, so we just do some computation and completely discard it in the normal build)?

If so, I'd say deserves a special treatment: either make the whole function early-return if the logging level is lower than needed or add some parameter to enable it explicitly and make the logging lines at the end of the function to INFO level.

cpp/src/neighbors/detail/cagra/graph_core.cuh
```diff
@@ -482,23 +706,227 @@ __global__ void kern_mst_opt_postprocessing(IdxT* outgoing_num_edges,  // [graph
...
+    d_output_graph.prefetch_next();
+    d_mst_graph.prefetch_next();
+    d_mst_graph_num_edges.prefetch_next();
+  }
+
+  bool check_num_protected_edges = true;
+  raft::copy(&check_num_protected_edges,
+             d_check_num_protected_edges.data_handle(),
+             1,
+             raft::resource::get_cuda_stream(res));
```
achirkin: Please pass the check_num_protected_edges as an mdspan and use raft::copy(res, ...) overload if possible. I'm not sure if it supports bools though - in that case, I think just using uint32_t is fine here (I'm preparing the ground for the dry run compliance).

cpp/src/neighbors/detail/cagra/graph_core.cuh
```diff
@@ -482,23 +706,227 @@ __global__ void kern_mst_opt_postprocessing(IdxT* outgoing_num_edges,  // [graph
...
+  raft::common::nvtx::range<cuvs::common::nvtx::domain::cuvs> block_scope(
+    "cagra::graph::optimize/reverse");
+
+  //
+  // Make reverse graph
+  //
+  raft::matrix::fill(res, d_rev_graph, IdxT(-1));
+  raft::matrix::fill(res, d_rev_graph_count, uint32_t(0));
+
+  if constexpr (!AccessorOutputGraph::is_device_accessible) {
```
achirkin: Also a preference question, but maybe it makes sense to invert the condition and put the no-copy version at the top (without negation it would read cleaner)?

cpp/src/neighbors/detail/cagra/graph_core.cuh
```diff
@@ -1069,94 +1514,156 @@ void mst_optimization(raft::resources const& res,
...
+
+  uint32_t batch_size      = static_cast<uint32_t>(std::min<uint64_t>(graph_size, 256 * 1024));
+  const uint32_t num_batch = (graph_size + batch_size - 1) / batch_size;
+
+  RAFT_LOG_DEBUG("# Pruning kNN Graph on GPUs\r");
+
+  const double prune_start = cur_time();
+
+  uint64_t num_keep __attribute__((unused)) = 0;
+  uint64_t num_full __attribute__((unused)) = 0;
```
achirkin: C++17
```suggestion
  [[maybe_unused]] uint64_t num_keep = 0;
  [[maybe_unused]] uint64_t num_full = 0;
```

cpp/src/neighbors/detail/cagra/graph_core.cuh
```diff
@@ -1166,465 +1673,88 @@ void optimize(
...
+  int64_t mst_graph_size = guarantee_connectivity ? graph_size : 0;
+  auto mst_graph =
+    raft::make_host_matrix<IdxT, int64_t, raft::row_major>(mst_graph_size, output_graph_degree);
+  auto mst_graph_num_edges = raft::make_host_vector<uint32_t, int64_t>(mst_graph_size);
+
   if (guarantee_connectivity) {
+#pragma omp parallel for
+    for (uint64_t i = 0; i < graph_size; i++) {
+      mst_graph_num_edges(i) = 0;
+    }
```
achirkin: Would it make sense to move the initialization of `mst_graph_num_edges` to the the mst_optimization function? I guess, you don't need it for the GPU branch in that function, right?

cpp/src/neighbors/detail/cagra/graph_core.cuh
```diff
@@ -1166,465 +1673,88 @@ void optimize(
...
-    const double time_make_end = cur_time();
-    RAFT_LOG_DEBUG("# Making reverse graph time: %.1lf ms",
-                   (time_make_end - time_make_start) * 1000.0);
+  // reverse graph creation will always use the GPU
+  // using default workspace resource for random access
+  // otherwise will be managed memory which is slow upon first access
+  auto d_rev_graph = raft::make_device_mdarray<IdxT>(res, raft::make_extents<int64_t>(0, 0));
+  try {
+    d_rev_graph = raft::make_device_mdarray<IdxT>(
+      res, raft::make_extents<int64_t>(graph_size, output_graph_degree));
```
achirkin: You're NOT using the default workspace resource here. Is this intentional? Please either update the comment (you're using the default/current device resource) or the code. Since the d_rev_graph is O(n_rows*graph_degree) size, it only use the `worksace_resource` for very small problem sizes.

achirkin (changes requested):
Hi Malte, thanks for the updates! Here's another batch of comments.
I see you updated optimize to decide on host/device execution based on template parameters as per our offline discussion. Do you plan to update the call site (cagra_build.cuh) to change the default choices for input/output graphs?

cpp/src/neighbors/detail/cagra/utils.hpp
cjnolet: We have a utility in RAFT for doing this very thing- please use that utility instead of creating a new one.

cpp/src/neighbors/detail/cagra/utils.hpp
cjnolet: Can we add that support to the existing batch_load_iterator please? It would be good to consolidate these typs of utilities/APIs as much as we possibly can- otherwise they add to the ongoing maintenance burdens.

cpp/src/neighbors/detail/cagra/graph_core.cuh
mfoerste4: Yes, I think we can remove it at this point. The legacy code relied on it in case mst was deactivated.

cpp/src/neighbors/detail/cagra/graph_core.cuh
mfoerste4: I will activate the path whenever debug logging is enabled. In that case it will also run for device memory now.

cpp/src/neighbors/detail/cagra/graph_core.cuh
mfoerste4: I changed it to use the default resource and fall back to large resource if allocation is failing.

cpp/src/neighbors/detail/cagra/graph_core.cuh
mfoerste4: Good idea.

cpp/src/neighbors/detail/cagra/graph_core.cuh
mfoerste4: It should be -- every warp uses its own chunk of shared memory.

cpp/src/neighbors/detail/cagra/utils.hpp
mfoerste4: I have merged the features into the existing batch_load_iterator. In order to still support the legacy usage pattern I created a helper utility to evaluate the pointer location at runtime and dispatch accordingly.

mfoerste4:
@achirkin , thanks for the additional review pass. I pushed a new version with a larger refactor covering the merge of the two iterators and also covered your other comments.

cpp/src/neighbors/detail/cagra/utils.hpp
mfoerste4: Has been removed in favor of the existing host_device_accessor template wrapper.

cpp/src/neighbors/detail/cagra/graph_core.cuh
achirkin: Oh, so you could in theory even implement this using exclusively warp shuffle operations? Would it make sense for the perf?

cpp/src/neighbors/detail/cagra/graph_core.cuh
mfoerste4: This is outdated. Decision is made based on host_device_accessor properties.

cpp/src/neighbors/detail/cagra/utils.hpp
mfoerste4: This is outdated. Decision is made based on host_device_accessor properties. The extra handling via device API has been removed.

cpp/src/neighbors/detail/cagra/graph_core.cuh
```diff
@@ -1166,465 +1736,82 @@ void optimize(
...
-      }
-    }
-    RAFT_EXPECTS(
-      num_dup == 0, "%lu duplicated node(s) are found in the generated CAGRA graph", num_dup);
-    RAFT_EXPECTS(num_oor == 0,
-                 "%lu out-of-range index node(s) are found in the generated CAGRA graph",
-                 num_oor);
+  // These host-side checks are expensive (O(N*D^2)) and only used as debug
+  // diagnostics, so only run them when debug logging is active at runtime.
+  if (raft::default_logger().should_log(rapids_logger::level_enum::debug)) {
```
achirkin, line 1809: nice!

cpp/src/neighbors/detail/cagra/graph_core.cuh
```diff
@@ -1069,94 +1557,176 @@ void mst_optimization(raft::resources const& res,
...
+      (double)std::min<IdxT>((i_batch + 1) * batch_size, graph_size) / graph_size * 100);
   }
+  raft::resource::sync_stream(res);
+  RAFT_LOG_DEBUG("\n");
+
+  uint32_t invalid_neighbor_list = 0;
+  raft::copy(&invalid_neighbor_list,
+             d_invalid_neighbor_list.data_handle(),
+             1,
+             raft::resource::get_cuda_stream(res));
```
achirkin: I think you have already fixed a very similar case to use the `res`-accepting overload and make_scalar_view function, right?

achirkin (approved):
Hi @mfoerste4 , thanks for addressing my comments, LGTM! A small nitpick below

mfoerste4:
@achirkin thanks for the review!

KyleFromNVIDIA (approved):

PR #1832, opened 2026-02-20
Make JIT LTO compilation a configurable CMake option

bdice:
## Summary

- Adds `CUVS_JIT_LTO_COMPILATION` cache variable with three values: `AUTO` (default), `ON`, and `OFF`
- `AUTO` preserves existing behavior: JIT LTO is enabled for CUDA 13+ and disabled otherwise
- `ON`/`OFF` allow explicit override regardless of CUDA version


cpp/CMakeLists.txt
```diff
@@ -67,6 +67,10 @@ option(DISABLE_OPENMP "Disable OpenMP" OFF)
 option(CUVS_COMPILE_DYNAMIC_ONLY "Only build the shared library and skip the static library." OFF)
 option(CUVS_NVTX "Enable nvtx markers" OFF)
 option(CUVS_RAFT_CLONE_ON_PIN "Explicitly clone RAFT branch when pinned to non-feature branch" ON)
+set(CUVS_JIT_LTO_COMPILATION
+    "AUTO"
+    CACHE STRING "Enable JIT LTO compilation (ON, OFF, or AUTO to enable for CUDA 13+)"
+)
```
KyleFromNVIDIA, line 73: Non-blocking suggestion:

For bonus points, you can use the [`STRINGS`](https://cmake.org/cmake/help/latest/prop_cache/STRINGS.html) property to make this cache variable multiple-choice:

```suggestion
set(CUVS_JIT_LTO_COMPILATION
    "AUTO"
    CACHE STRING "Enable JIT LTO compilation (ON, OFF, or AUTO to enable for CUDA 13+)"
)
set_property(CACHE CUVS_JIT_LTO_COMPILATION PROPERTY STRINGS "AUTO;ON;OFF")
```

cpp/CMakeLists.txt
```diff
@@ -346,10 +350,20 @@ if(NOT BUILD_CPU_ONLY)
...
+      set(JIT_LTO_COMPILATION ON)
+    endif()
+  elseif(CUVS_JIT_LTO_COMPILATION)
     set(JIT_LTO_COMPILATION ON)
   endif()
+  if(JIT_LTO_COMPILATION)
+    if(CMAKE_CUDA_COMPILER_VERSION VERSION_GREATER_EQUAL 13.0)
+      set(JIT_LTO_TARGET_ARCHITECTURE "75-real")
+    else()
+      set(JIT_LTO_TARGET_ARCHITECTURE "70-real")
```
KyleFromNVIDIA, line 364: This wasn't being set before. Is there a reason you're setting it now?

KyleFromNVIDIA: Oh, this is in case we do JIT LTO with CUDA < 13.

divyegala (changes requested):
As we discussed offline, JIT-LOT is not going to be an optional feature in cuVS once we upgrade our drivers. As such, we should not merge this PR.

Issue #1818, opened 2026-02-18
[BUG] cuVS 25.12 failed to build index with ITERATIVE_CAGRA_SEARCH and small intermediate_graph_degree


lanstonchu:
**Describe the bug**
We have tried tup upgraded the cuVS version from v25.10 to v25.12, but we saw lots of segmentation fault when we tried to build index that we haven't seen before in v25.10. For example, if we used below build parameters, we saw an seg-fault error at the 4th iteration (see attached error message).

We also tried larger intermediate_graph_degree, saying 32 and 64, but still see error, and only won't see error when intermediate_graph_degree is increased to 128. We also see error in other algorithm, but we saw errors more often at ITERATIVE_CAGRA_SEARCH then other algorithms.

**Steps/Code to reproduce bug**
```
cuDA version: 13.1
cuVS version: 25.10
metric: IP
build_algo: ITERATIVE_CAGRA_SEARCH
intermediate_graph_degree: 16
graph_degree: 16
```

**Expected behavior**
- index build without error in cuVS v25.10, so we expect cuVS v25.12 to behave same

**Environment details (please complete the following information):**
- cuDA version: 13.1
- cuVS version: 25.10
- We are installing cuDA/cuVS using Conda;
- The implementation mainly relies on cuVS’s C API (e.g., cuvsCagraBuild, cuvsCagraSearch, cuvsResourcesCreate, etc.), using opaque handle types such as cuvsCagraIndex_t. However, it also bypasses the C API abstraction by casting the internal handle (index->addr) back to the underlying C++ type (cuvs::neighbors::cagra::index<float, uint32_t>) and invoking a C++ member function (size()). Because it uses C++-specific constructs such as reinterpret_cast, namespaces (cuvs::neighbors::cagra), templates, and auto, the file must be compiled with a C++ compiler, even though most of its cuVS interactions go through the C API.

**Additional context**
[error_message.txt](https://github.com/user-attachments/files/25394463/error_message.txt)

tfeher:
Thank you @lanstonchu for reporting! Could you share the dataset shape (rows, dims) and data type so that we can reproduce?

Tagging @irina-resh-nvda to fix this. The error is likely related to internal alignment requirement not satisfied by the dataset dimension. The search kernel would run into the error and the subsequent memcpy op will report the error. We should check for this and add padding if necessary. 

> We also see error in other algorithm

[...]

PR #1834, opened 2026-02-23
Graph degree equals intermediate graph degree bug fix

irina-resh-nvda:
This PR fixes a cudaErrorIllegalAddress error in the iterative CAGRA graph build caused by an out-of-bounds memory access. The issue occurred because the algorithm allocated memory for intermediate_graph_degree neighbors but attempted to write one additional neighbor during the search phase. The fix updates the topk allocation to ensure buffers are sized for intermediate_graph_degree + 1.
I also added a new test into  NEIGHBORS_ANN_CAGRA_TEST_BUGS to reproduce the failure. It confirms the fix works correctly for float, int8_t, and uint8_t data types.
Fixes  https://github.com/rapidsai/cuvs/issues/1818
#1818

### DIFF
--- a/cpp/src/neighbors/detail/cagra/cagra_build.cuh
+++ b/cpp/src/neighbors/detail/cagra/cagra_build.cuh
@@ -2021,7 +2021,9 @@ auto iterative_build_graph(
 
   // Allocate memory for search results.
   constexpr uint64_t max_chunk_size = 8192;
-  auto topk                         = intermediate_degree + 1;
+  // +1 because the search may return the query node itself as a neighbor;
+  // this is consistent with the per-iteration curr_topk = next_graph_degree + 1
+  auto topk          = intermediate_degree + 1;
   auto dev_neighbors = raft::make_device_matrix<IdxT, int64_t>(res, max_chunk_size, topk);
   auto dev_distances = raft::make_device_matrix<float, int64_t>(res, max_chunk_size, topk);
 
--- a/cpp/tests/neighbors/ann_cagra/bug_iterative_cagra_build.cu
+++ b/cpp/tests/neighbors/ann_cagra/bug_iterative_cagra_build.cu
@@ -26,6 +26,8 @@ class CagraIterativeBuildBugTest : public ::testing::Test {
   {
     // Set up iterative CAGRA graph building
     cagra::index_params index_params;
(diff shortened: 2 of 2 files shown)


cpp/src/neighbors/detail/cagra/cagra_build.cuh
```diff
@@ -2021,7 +2021,7 @@ auto iterative_build_graph(
 
   // Allocate memory for search results.
   constexpr uint64_t max_chunk_size = 8192;
-  auto topk                         = intermediate_degree;
+  auto topk                         = intermediate_degree + 1;
```
dantegd:
```suggestion
  // +1 because the search may return the query node itself as a neighbor;
  // this is consistent with the per-iteration curr_topk = next_graph_degree + 1
  auto topk = intermediate_degree + 1;
```

Good find and fix!  I'd suggest adding a brief comment at the  assignment explaining why the +1 is needed. Without this, a future refactor or change might "optimize" this by removing the seemingly redundant +1.

cpp/tests/neighbors/ann_cagra/bug_iterative_cagra_build.cu
```diff
@@ -0,0 +1,83 @@
...
+ public:
+  using data_type = DataT;
+
+ protected:
+  void run()
+  {
+    // Set up iterative CAGRA graph building
+    cagra::index_params index_params;
+    index_params.graph_degree              = 16;
+    index_params.intermediate_graph_degree = 16;
```
dantegd, line 32: Similar as above, I'd suggest a comment that to trigger/test this it's required for these to coincide (perhaps add link to the issue)

dantegd (approved):
bug fix looks good to me, just had some small suggestions

cpp/tests/neighbors/ann_cagra/bug_iterative_cagra_build.cu
irina-resh-nvda: Done! Thank you

cpp/src/neighbors/detail/cagra/cagra_build.cuh
irina-resh-nvda: Makes sense. Done

PR #1836, opened 2026-02-23
Move `test_compute_matrix_product.py` to `cpp/tests`

KyleFromNVIDIA:
Follow-up to https://github.com/rapidsai/cuvs/pull/1812#discussion_r2835055379

### DIFF
--- a/dependencies.yaml
+++ b/dependencies.yaml
@@ -382,7 +382,7 @@ dependencies:
         packages:
           - *cmake_ver
           - openblas
-          - pytest
+          - &pytest pytest<9.0.0
   docs:
     common:
       - output_types: [conda]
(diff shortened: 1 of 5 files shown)


dependencies.yaml
```diff
@@ -382,6 +382,7 @@ dependencies:
         packages:
           - *cmake_ver
           - openblas
+          - pytest
```
jameslamb:
```suggestion
          - &pytest pytest<9.0.0
```

Let's please re-use the same `pytest` pin everywhere in the project.

This suggestion + updating this to `*pytest`:

https://github.com/rapidsai/cuvs/blob/d0175025be4e8885f43d00172f26276040521b87/dependencies.yaml#L447

jameslamb (approved):
Thank you, just one small suggestion. I don't need to re-review.

PR #1839, opened 2026-02-23
fix(rust): change Index::search to take &self instead of self

### LINKED ISSUE #1838: Rust API: Index::search() takes ownership (self), preventing index reuse
## Summary

The cuVS Rust bindings' `Index::search()` method takes `self` by value, consuming the index and preventing reuse. This requires rebuilding the index after every search, making real-time/high-throughput search impractical.

## Affected Methods

- `cuvs::cagra::Index::search(self, ...)`
- `cuvs::ivf_pq::Index::search(self, ...)`
- `cuvs::ivf_flat::Index::search(self, ...)`

## Current Behavior

```rust
// Current API signature (simplified)
impl Index {
    pub fn search(self, res: &Resources, ...) -> Result<()>
    //           ^^^^
    //           Takes ownership, index is consumed
}
```

This means:
```rust
let index = Index::build(...)?;
index.search(...)?;  // First search - OK
index.search(...)?;  // COMPILE ERROR: use of moved value
```

## Expected Behavior

```rust
// Proposed API signature
impl Index {
    pub fn search(&self, res: &Resources, ...) -> Result<()>
    //           ^^^^^
    //           Borrows, index can be reused
}
```

This enables:
```rust
let index = Index::build(...)?;
index.search(...)?;  // First search - OK
index.search(...)?;  // Second search - OK
index.search(...)?;  // Third search - OK, no rebuild needed
```

## Impact

We're building a semantic cache for LLM applications using cuVS CAGRA. With the current API, every cache lookup requires a full index rebuild (~670ms), making L0 GPU cache slower than L1 Redis cache (~5ms).

The workaround is to maintain a pool of pre-built indexes, but this wastes GPU memory and adds complexity.

## Proposed Fix

Change the method signature from:
```rust
pub fn search(self, ...) -> Result<()>
```

To:
```rust
pub fn search(&self, ...) -> Result<()>
```

This may require internal changes to how the C FFI handles the index pointer, but would make the Rust API consistent with other ANN libraries (faiss-rs, hnsw_rs, annoy-rs).

## Related

- [cuVS Documentation](https://docs.rapids.ai/api/cuvs/stable/)
- [cuVS Rust API](https://docs.rapids.ai/api/cuvs/stable/_static/rust/cuvs/)
- [CAGRA Paper](https://arxiv.org/abs/2308.15136)

## Environment

- cuVS Rust crate version: 26.2
- CUDA version: 12.x
- OS: Linux (Ubuntu 22.04)

zbennett10:
## Summary

The `search` methods on CAGRA, IVF-PQ, IVF-Flat, and Brute Force indexes were taking `self` by value, consuming the index and preventing reuse. This required users to rebuild the index after every search operation, making real-time/high-throughput search impractical.

## Changes

Changed the method signature from `self` to `&self` in:
- `cuvs::cagra::Index::search`
- `cuvs::ivf_pq::Index::search`
- `cuvs::ivf_flat::Index::search`
- `cuvs::brute_force::Index::search`

## Before

```rust
let index = Index::build(...)?;
index.search(...)?;  // First search - OK
index.search(...)?;  // COMPILE ERROR: use of moved value
```

## After

```rust
let index = Index::build(...)?;
index.search(...)?;  // First search - OK
index.search(...)?;  // Second search - OK
index.search(...)?;  // Third search - OK
```

## Rationale

The underlying C FFI functions (`cuvsCagraSearch`, `cuvsIvfPqSearch`, etc.) do not mutate the index during search - they only read from it. Therefore, taking a shared reference (`&self`) is safe and more ergonomic.

This change makes the Rust API consistent with other ANN libraries:
- `faiss-rs`: `index.search(&query, k)`
- `hnsw_rs`: `hnsw.search(&query, k, ef)`
- `annoy-rs`: `index.get_nns_by_vector(&v, n)`

## Testing

The existing unit tests in the repository continue to pass with this change. The tests use the `search` method exactly once per index, so they work with both the old and new signature.

Fixes #1838

### DIFF
--- a/.github/workflows/build.yaml
+++ b/.github/workflows/build.yaml
@@ -35,7 +35,7 @@ concurrency:
 jobs:
   cpp-build:
     secrets: inherit
-    uses: rapidsai/shared-workflows/.github/workflows/conda-cpp-build.yaml@main
+    uses: rapidsai/shared-workflows/.github/workflows/conda-cpp-build.yaml@python-3.14
     with:
       build_type: ${{ inputs.build_type || 'branch' }}
       branch: ${{ inputs.branch }}
@@ -46,7 +46,7 @@ jobs:
 
   rocky8-clib-standalone-build:
     secrets: inherit
-    uses: rapidsai/shared-workflows/.github/workflows/custom-job.yaml@main
+    uses: rapidsai/shared-workflows/.github/workflows/custom-job.yaml@python-3.14
     strategy:
       fail-fast: false
       matrix:
@@ -71,7 +71,7 @@ jobs:
   rust-build:
     needs: cpp-build
     secrets: inherit
-    uses: rapidsai/shared-workflows/.github/workflows/custom-job.yaml@main
+    uses: rapidsai/shared-workflows/.github/workflows/custom-job.yaml@python-3.14
     # Artifacts are not published from these jobs, so it's safe to run for multiple CUDA versions.
     # If these jobs start producing artifacts, the names will have to differentiate between CUDA versions.
     strategy:
@@ -92,7 +92,7 @@ jobs:
   go-build:
     needs: cpp-build
     secrets: inherit
-    uses: rapidsai/shared-workflows/.github/workflows/custom-job.yaml@main
+    uses: rapidsai/shared-workflows/.github/workflows/custom-job.yaml@python-3.14
     # Artifacts are not published from these jobs, so it's safe to run for multiple CUDA versions.
     # If these jobs start producing artifacts, the names will have to differentiate between CUDA versions.
     strategy:
@@ -113,7 +113,7 @@ jobs:
   java-build:
     needs: cpp-build
     secrets: inherit
-    uses: rapidsai/shared-workflows/.github/workflows/custom-job.yaml@main
+    uses: rapidsai/shared-workflows/.github/workflows/custom-job.yaml@python-3.14
     # Artifacts are not published from these jobs, so it's safe to run for multiple CUDA versions.
     # If these jobs start producing artifacts, the names will have to differentiate between CUDA versions.
     strategy:
@@ -135,7 +135,7 @@ jobs:
   python-build:
     needs: [cpp-build]
     secrets: inherit
-    uses: rapidsai/shared-workflows/.github/workflows/conda-python-build.yaml@main
+    uses: rapidsai/shared-workflows/.github/workflows/conda-python-build.yaml@python-3.14
     with:
       build_type: ${{ inputs.build_type || 'branch' }}
       branch: ${{ inputs.branch }}
@@ -147,7 +147,7 @@ jobs:
   upload-conda:
     needs: [cpp-build, python-build]
     secrets: inherit
-    uses: rapidsai/shared-workflows/.github/workflows/conda-upload-packages.yaml@main
+    uses: rapidsai/shared-workflows/.github/workflows/conda-upload-packages.yaml@python-3.14
     with:
       build_type: ${{ inputs.build_type || 'branch' }}
       branch: ${{ inputs.branch }}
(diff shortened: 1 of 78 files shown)


zbennett10:
## Added Tests

I've added `test_*_multiple_searches` tests to all affected index types:
- `cagra::Index`
- `ivf_flat::Index`
- `ivf_pq::Index`
- `brute_force::Index`

### Why these tests provide real value

With the **old** `fn search(self, ...)` signature, these tests would **fail to compile** with:

```
error[E0382]: use of moved value: `index`
   |
   |     index.search(...);
   |           -------- `index` moved due to this method call
   |
   |     index.search(...);
   |     ^^^^^ value used here after move
```

With the **new** `fn search(&self, ...)` signature, the tests compile and pass because the index is borrowed rather than consumed.

This provides compile-time validation that the fix is correct - if someone accidentally reverted to `self`, these tests would immediately fail to compile.

aamijar:
Hi @zbennett10, thanks for the great work! The rust-build CI job seems to be failing, if you have a moment to take a look.

zbennett10:
> Hi @zbennett10, thanks for the great work! The rust-build CI job seems to be failing, if you have a moment to take a look.

Will get to it as soon as I can - thanks!

zbennett10:
@aamijar Ok should be good to go!

zbennett10:
How's this one looking? Anyone need anything else from me? :)

zbennett10:
Anyone available to take a look at this one? It's currently blocking us. 😄

rust/cuvs/src/cagra/index.rs
```diff
@@ -59,7 +59,7 @@ impl Index {
     /// * `neighbors` - Matrix in device memory that receives the indices of the nearest neighbors
     /// * `distances` - Matrix in device memory that receives the distances of the nearest neighbors
     pub fn search(
-        self,
+        &self,
```
benfred, line 62: good catch! thanks for fixing this

rust/cuvs/src/brute_force.rs
```diff
@@ -12,8 +12,18 @@ use crate::error::{check_cuvs, Result};
...
+/// the brute force index holds a non-owning view into the original dataset.
+/// The `_dataset` field keeps the device memory alive for the lifetime of
+/// the index.
 #[derive(Debug)]
-pub struct Index(ffi::cuvsBruteForceIndex_t);
+pub struct Index {
+    inner: ffi::cuvsBruteForceIndex_t,
+    // The C library stores a view into the dataset (not a copy).
+    // We must keep the ManagedTensor alive so the device memory is not freed.
+    _dataset: Option<ManagedTensor>,
```
benfred: A couple thoughts on this:

* The brute force index isn't the only index that requires the dataset to be kept alive (for instance, cagra sometimes will take a non-owning view of the data - depending on the dataset dimensionality, it creates a copy if the dataset isn't 16 byte aligned).

* The ManagedTensor can itself be a non-owning view of the data - for instance, when being passed a host memory `ndarray::ArrayBase`, this code will get executed https://github.com/rapidsai/cuvs/blob/02192e4683a41aae6be59de08834fae603a6bf6a/rust/cuvs/src/dlpack.rs#L92-L96.  I think this will still work here right now because the bfknn index will internally copy the host input data to device memory, but this is requiring implementation details of the C++ library - which could change in the future

I'm wondering if we can achieve the same result in a different way - maybe by using a lifetime to tie the lifetime of the dataset to the index?

rust/cuvs/src/cagra/index.rs
zbennett10: No problem!

rust/cuvs/src/brute_force.rs
zbennett10: Yeah that's a good point... I'll rework this to use a lifetime-based approach.  I'll prototype a couple options and push an update.

zbennett10:
@benfred @aamijar   updated this PR accordingly - thanks

PR #1840, opened 2026-02-23
feat(rust): add serialize/deserialize support for CAGRA index

### LINKED ISSUE #1479: [FEA] Add support for index serialization/deserialization in rust
I wish I could serialize and de-serialize indexes using the cuVS rust crate.

**Describe the solution you'd like**
Idiomatic and safe rust crate bindings to underlying `*Serialize`/`*Deserialize` functions.

zbennett10:
Closes #1479

## Description

Adds serialization and deserialization support to the Rust CAGRA index bindings. The underlying C API functions (`cuvsCagraSerialize`, `cuvsCagraDeserialize`, `cuvsCagraSerializeToHnswlib`) already exist but were not exposed through the Rust crate.

### Methods added to `cagra::Index`:

- **`serialize(&self, res, filename, include_dataset)`** — Save the CAGRA index to file, with option to include or exclude the dataset
- **`serialize_to_hnswlib(&self, res, filename)`** — Save the CAGRA index in hnswlib-compatible format
- **`deserialize(res, filename) -> Result<Index>`** — Load a CAGRA index from file

### Design decisions:

- Uses `&self` (not `self`) for serialize methods so the index remains usable after serialization. This differs from the Vamana pattern but is more ergonomic for the common case of serializing a hot index.
- Uses `CString` for safe C string interop with the FFI layer.
- Follows the existing error handling pattern via `check_cuvs`.

### Tests:

- **`test_cagra_serialize_deserialize`** — Round-trip test: build → serialize (with dataset) → deserialize → search → verify nearest neighbors match
- **`test_cagra_serialize_without_dataset`** — Verifies serialization works with `include_dataset=false`
- **`test_cagra_serialize_to_hnswlib`** — Verifies hnswlib format export produces a non-empty file

### Documentation:

- Added serialization example to `cagra` module docs in `mod.rs`

## Checklist

### DIFF
--- a/rust/cuvs/src/cagra/index.rs
+++ b/rust/cuvs/src/cagra/index.rs
@@ -3,16 +3,29 @@
  * SPDX-License-Identifier: Apache-2.0
  */
 
+use std::ffi::CString;
 use std::io::{stderr, Write};
+use std::path::Path;
 
 use crate::cagra::{IndexParams, SearchParams};
 use crate::dlpack::ManagedTensor;
-use crate::error::{check_cuvs, Result};
+use crate::error::{check_cuvs, Error, Result};
 use crate::resources::Resources;
 
 /// CAGRA ANN Index
 #[derive(Debug)]
 pub struct Index(ffi::cuvsCagraIndex_t);
+
+/// Convert a filesystem path into a `CString` suitable for the cuVS C API,
+/// returning `Error::InvalidArgument` instead of panicking for paths that are
+/// not valid UTF-8 or that contain an interior NUL byte.
+fn path_to_cstring(path: &Path) -> Result<CString> {
+    let path_str = path
+        .to_str()
+        .ok_or_else(|| Error::InvalidArgument(format!("path is not valid UTF-8: {path:?}")))?;
+    CString::new(path_str)
+        .map_err(|e| Error::InvalidArgument(format!("path contains an interior NUL byte: {e}")))
+}
 
 impl Index {
     /// Builds a new Index from the dataset for efficient search.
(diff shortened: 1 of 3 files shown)


rust/cuvs/src/cagra/index.rs
```diff
@@ -83,6 +84,70 @@ impl Index {
...
+    ///
+    /// Experimental, both the API and the serialization format are subject to change.
+    ///
+    /// # Arguments
+    ///
+    /// * `res` - Resources to use
+    /// * `filename` - The file name for saving the index
+    /// * `include_dataset` - Whether to write out the dataset to the file
+    pub fn serialize(&self, res: &Resources, filename: &str, include_dataset: bool) -> Result<()> {
+        let c_filename = CString::new(filename).expect("filename contains null byte");
```
yan-zaretskiy: 1. This is a fallible function, please don't call `expect` here. Same for other methods in this PR.
2. Please use `impl AsRef<Path>` instead of `&str` for `filename`, which is a common Rust pattern.

rust/cuvs/src/cagra/index.rs
```diff
@@ -167,4 +232,141 @@ mod tests {
...
+            ndarray::Array::<f32, _>::random((n_datapoints, n_features), Uniform::new(0., 1.0));
+
+        // Build the CAGRA index
+        let build_params = IndexParams::new().unwrap();
+        let index =
+            Index::build(&res, &build_params, &dataset).expect("failed to create cagra index");
+
+        // Serialize to a temp file (including dataset)
+        let dir = std::env::temp_dir();
+        let filepath = dir.join("test_cagra_index.bin");
```
yan-zaretskiy: Looks like all the tests have a common setup logic, see if you can factor it out.

yan-zaretskiy:
Looks good to me otherwise, thank you for your contribution!

rust/cuvs/src/cagra/index.rs
zbennett10: Addressed in latest push:

- All three methods (`serialize`, `serialize_to_hnswlib`, `deserialize`) now take `filename: impl AsRef<Path>` instead of `&str`.
- Replaced `CString::new(...).expect(...)` with a small `path_to_cstring` helper that returns a new `Error::InvalidArgument(String)` variant when the path is non-UTF-8 or contains an interior NUL byte, so the functions are no longer panic-on-bad-input.
- Added `test_cagra_serialize_rejects_interior_nul` to cover the error path.

rust/cuvs/src/cagra/index.rs
zbennett10: Factored the shared setup into two test helpers:

- `build_test_index(&res, &build_params) -> (Array2<f32>, Index)` builds a small random dataset + CAGRA index.
- `search_and_verify_self_neighbors(&res, &index, &dataset, n_queries, k)` runs the standard self-neighbor search and assertions.

zbennett10:
@yan-zaretskiy this one is ready for another pair of eyes when you're ready - thanks!

rust/cuvs/src/cagra/index.rs
```diff
@@ -83,6 +96,75 @@ impl Index {
...
+    /// Load a CAGRA index from file.
+    ///
+    /// Experimental, both the API and the serialization format are subject to change.
+    ///
+    /// # Arguments
+    ///
+    /// * `res` - Resources to use
+    /// * `filename` - The path of the file that stores the index
+    pub fn deserialize<P: AsRef<Path>>(res: &Resources, filename: P) -> Result<Index> {
+        let c_filename = path_to_cstring(filename.as_ref())?;
```
zbennett10, line 157: As primarily a C programmer, it always warms my heart to see any mention of C strings. :)

PR #1842, opened 2026-02-23
feat(c-api): expose attach_dataset_on_build and add cuvsCagraUpdateDataset

### LINKED ISSUE #1456: [BUG] Cagra graph build copies dataset multiple times
**Describe the bug**
When we supply to `cagraIndexBuild` (or the equivalent C++ functions) a dataset that resides in device memory (e.g. because it was already copied, e.g. to remove strides as a workaround for #1455, or is the result of other GPU computations like quantization), a new copy of the dataset is created in device memory. This means we need (at least) twice the dataset memory on the device, meaning `cagraIndexBuild` is likely to fail for bigger datasets.

On a X GB GPU, supply to `cagraIndexBuild` a DLManagedTensor with dataset type kDLCUDA (dataset address on device memory) and size X * 0.6 GB (a bit more than half the ram, e.g. a 13GB dataset on a 24GB GPU). `cagraIndexBuild`  fails with a memory error.

**Expected behavior**

`cagraIndexBuild` should provide help/guidance/a way to avoid that:
- a way of not copying accepting the performance hit could be fine
- instructions on how to shape the data (pre-stridden/padded data) so we can avoid the copy and take a "fast path"

Any way to avoid the issue and allow more data to be fitted and processed on the GPU

**Additional context**

Another issue that was recently brought to light is the fact that nn-descent wants to copy the whole dataset into float16. 
That is a separate issue above and beyond the strided/padded dataset issue, but it's another area of concern around usage of GPU memory/multiple dataset copies.

zbennett10:
Closes #1456

## Description

Exposes the existing C++ `attach_dataset_on_build` parameter through the C API, and adds a new `cuvsCagraUpdateDataset()` function. Together, these enable a memory-efficient build workflow that avoids duplicating the dataset in device memory.

### Problem

When calling `cuvsCagraBuild` with a device-resident dataset, the build process creates an aligned copy of the dataset in the index constructor via `make_aligned_dataset`. This means the GPU needs at least 2x the dataset memory, causing OOM failures for datasets larger than ~60% of GPU memory (e.g., a 13GB dataset on a 24GB GPU).

The C++ API already has `attach_dataset_on_build` to skip this copy, but it was not exposed in the C API.

### Solution

**C API struct change:**
- Added `bool attach_dataset_on_build` field to `cuvsCagraIndexParams` (default: `true`, preserving backward compatibility)

**New C API function:**
- `cuvsCagraUpdateDataset(res, dataset, index)` — Attaches a dataset to an index that was built with `attach_dataset_on_build = false`. When the dataset is already on device with proper 16-byte row alignment, this is a zero-copy operation.

**Memory-efficient workflow:**
```c
// 1. Build graph only (no dataset copy)
build_params->attach_dataset_on_build = false;
cuvsCagraBuild(res, build_params, &dataset_tensor, index);

// 2. Attach dataset (zero-copy if properly aligned on device)
cuvsCagraUpdateDataset(res, &dataset_tensor, index);

// 3. Search
cuvsCagraSearch(res, search_params, index, ...);
```

### Testing

- Added `CagraC.BuildNoDatasetThenUpdateAndSearch` test that validates the complete workflow: build without dataset → update dataset → search → verify correctness against known expected results
- Ran all existing CAGRA C tests (7 tests) — all pass
- Ran all other C test suites (BRUTEFORCE, IVF_FLAT, IVF_PQ, DISTANCE, HNSW) — all pass, no regressions
- Tested on NVIDIA Tesla T4 (g4dn.xlarge) via EKS

## Checklist


c/src/neighbors/cagra.cpp
```diff
@@ -129,6 +129,26 @@ void* _build(cuvsResources_t res, cuvsCagraIndexParams params, DLManagedTensor*
...
+
+  if (cuvs::core::is_dlpack_device_compatible(dataset)) {
+    using mdspan_type = raft::device_matrix_view<T const, int64_t, raft::row_major>;
+    auto mds          = cuvs::core::from_dlpack<mdspan_type>(dataset_tensor);
+    index_ptr->update_dataset(*res_ptr, mds);
+  } else if (cuvs::core::is_dlpack_host_compatible(dataset)) {
+    using mdspan_type = raft::host_matrix_view<T const, int64_t, raft::row_major>;
+    auto mds          = cuvs::core::from_dlpack<mdspan_type>(dataset_tensor);
+    index_ptr->update_dataset(*res_ptr, mds);
+  }
```
dantegd, line 150: There's no else branch here. If `dataset_tensor` has an unsupported memory type, the function returns successfully without attaching anything, which would make follow up searches fail (or produce garbage maybe, haven't thought fully about it). 

The `_extend` function handles this with:

```C++
  } else {
    RAFT_FAIL("Unsupported dataset DLtensor dtype: %d and bits: %d",
              dataset.dtype.code,
              dataset.dtype.bits);
  }
```

we should do the same here

c/tests/neighbors/ann_cagra_c.cu
```diff
@@ -337,6 +337,110 @@ TEST(CagraC, BuildExtendSearch)
...
+  // create cuvsResources_t
+  cuvsResources_t res;
+  cuvsResourcesCreate(&res);
+  cudaStream_t stream;
+  cuvsStreamGet(res, &stream);
+
+  // create dataset DLTensor
+  DLManagedTensor dataset_tensor;
+  dataset_tensor.dl_tensor.data               = dataset;
+  dataset_tensor.dl_tensor.device.device_type = kDLCPU;
```
dantegd, line 356: If I'm not mistaken, the motivating scenario (dataset in device, avoiding the redundant copy) is not tested. Adding a variant that creates a `kDLCUDA` dataset tensor for the `cuvsCagraUpdateDataset `call would exercise that code path.

dantegd (changes requested):
PR looks good to me overall, just have one request to handle a potential silent error and one question about the test.

c/src/neighbors/cagra.cpp
zbennett10: good catch - thanks!

c/tests/neighbors/ann_cagra_c.cu
zbennett10: You're right — added `BuildNoDatasetThenUpdateDeviceAndSearch` which creates the dataset via `rmm::device_uvector`, constructs a `kDLCUDA` `DLManagedTensor`, builds with
  `attach_dataset_on_build=false`, then calls `cuvsCagraUpdateDataset` to exercise the device code path. Verified passing on a T4. Thanks!

PR #1847, opened 2026-02-25
Use 1D grid calculations in `epsilon_neighborhood`

divyegala:
Closes https://github.com/rapidsai/cuml/issues/5393 closes https://github.com/rapidsai/cuml/issues/7676

This was necessary to bypass the limits of `grid.y < 65536`, as cuml passes `n_rows` to the `n_cols` dimension in the usage of this API.

### DIFF
--- a/cpp/tests/neighbors/epsilon_neighborhood.cu
+++ b/cpp/tests/neighbors/epsilon_neighborhood.cu
@@ -1,5 +1,5 @@
 /*
- * SPDX-FileCopyrightText: Copyright (c) 2025, NVIDIA CORPORATION.
+ * SPDX-FileCopyrightText: Copyright (c) 2025-2026, NVIDIA CORPORATION.
  * SPDX-License-Identifier: Apache-2.0
  */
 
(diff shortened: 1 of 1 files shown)


divyegala:
> Looks good to me, though before approving explicitly wanted to ask if you think it'd be worth adding a explicit test to exercise this?

@dantegd good idea. Will add 👍

cpp/src/neighbors/detail/epsilon_neighborhood.cuh
```diff
@@ -39,6 +43,9 @@ struct EpsUnexpL2SqNeighborhood : public BaseClass {
                               char* _smem)
     : BaseClass(_x, _y, _m, _n, _k, _smem), adj(_adj), eps(_eps), vd(_vd), smem(_smem)
   {
+    n_blocks_y = raft::ceildiv<size_t>(static_cast<size_t>(_n), static_cast<size_t>(P::Nblk));
```
tarang-jain: This quantity can be defined outside the kernel, right? So it is not computed by each thread.

tarang-jain (commented):
I see what is being done here. LGTM except possibly one minor optimization

tarang-jain (approved):

viclafargue (approved):
Thanks Divye! LGTM. An alternative solution would have been to swap the x and y dimensions. The number of features divided by `Policy::Mblk` should rarely ever go beyond the limit of 65535. This would avoid adding an argument to the kernel and also avoid the two small instructions to get `block_x` and `block_y` back. But, I agree that the current solution may be more robust.

divyegala:
@viclafargue That solution does not work because we cannot guarantee always that `m` will be passed as `n`. cuML itself reverses the values of `m` and `n` before calling this function, but cuVS does not expect the reversal. So if anyone else uses the API then you run into the same problem.

dantegd (approved):

PR #1853, opened 2026-02-25
[WIP] [FEA] Separate MiniBatchKMeans and Streaming Batch Kmeans

tarang-jain:


cpp/include/cuvs/cluster/kmeans.hpp
```diff
@@ -391,6 +438,81 @@ void fit(raft::resources const& handle,
...
+ * @param[in]     X             Training data (host memory, row-major).
+ *                              [dim = n_samples x n_features]
+ * @param[in]     sample_weight Optional weights for each observation in X.
+ *                              [len = n_samples]
+ * @param[inout]  centroids     [in] Initial centroids / [out] Fitted centroids (device memory).
+ *                              [dim = n_clusters x n_features]
+ * @param[out]    inertia       Sum of squared distances.
+ * @param[out]    n_iter        Number of steps run.
+ */
+void minibatch_fit(raft::resources const& handle,
```
cjnolet, line 498: Definitely don't want multiple completely different APIs for the various kmeans impls. The params objects should be different enough. 

Note that we can still have two completely different code paths for now underneath, but we should be following the same APIs.

tarang-jain: @cjnolet sounds good. Anyway this is turning out to be a really big refactor and I'll revisit this during the 26.06 development cycle. Closing this PR for the time being (#1691 is the one to focus on).

PR #1861, opened 2026-03-01
[Rust] Add validated IndexParamsBuilder / SearchParamsBuilder (fixes #1859)

### LINKED ISSUE #1859: [Rust] IndexParams/SearchParams setters silently accept invalid values — add validated builder
## Environment

- cuVS: 26.2 (crates.io)
- CUDA: 12.4
- GPU: T4 (sm_75)
- Rust: 1.82 (MSRV)

## Problem

`IndexParams` setter methods return `Self` unconditionally. Invalid parameter
values are accepted at configuration time and only surface as an opaque CUDA
assertion inside `Index::build()` — after GPU memory has already been allocated
and graph construction has started.

```rust
// This compiles. It runs. It says nothing.
// ~1.8 seconds later, during Index::build():
let params = IndexParams::new()?
    .set_graph_degree(0)               // graph_degree=0 is invalid for CAGRA
    .set_intermediate_graph_degree(0)  // must be >= graph_degree
    .set_nn_descent_niter(0);          // must be > 0 for convergence

// Error surfaces here:
// cuvsError { code: CUVS_ERROR, text: "raft::exception: assertion failed" }
// No field name. No valid range. No hint.
let _index = Index::build(&res, &params, &dataset)?;
```

The same problem affects `SearchParams` (`set_hashmap_max_fill_rate(1.5)`,
`set_team_size(7)`) and `IndexParams` for `ivf_pq`, `ivf_flat`, and `vamana`.

When CAGRA parameters are loaded from a config file at runtime, a
misconfiguration blocks the GPU thread for the full build duration before
producing an error that names neither the offending field nor its valid range.

## Root Cause

The C API (`cuvsCagraBuild`) validates parameters on entry and returns
`CUVS_ERROR` on failure. The Rust bindings expose setters that write directly
to the FFI struct without pre-validation, so constraints documented in the C
header are invisible to callers until `Index::build()` fires.

## Proposed Fix

Add `IndexParams::builder()` — a parallel, additive entry point that validates
all parameters in a single `build() -> Result<IndexParams>` call before any
FFI allocation occurs. Existing `IndexParams::new()?.set_graph_degree(...)` is
unchanged.

```rust
pub struct IndexParamsBuilder { /* private */ }

impl IndexParamsBuilder {
    pub fn graph_degree(mut self, v: usize) -> Self;
    pub fn intermediate_graph_degree(mut self, v: usize) -> Self;
    pub fn nn_descent_niter(mut self, v: usize) -> Self;
    // ... remaining fields

    /// Validate parameters in pure Rust (no GPU needed).
    pub fn validate(&self) -> Result<()>;

    /// Validate all parameters and allocate the FFI struct.
    ///
    /// Returns `Err` with a message naming the offending field and its valid
    /// range before any GPU work begins.
    pub fn build(self) -> Result<IndexParams>;
}

impl IndexParams {
    pub fn builder() -> IndexParamsBuilder;
}
```

Usage:

```rust
let params = IndexParams::builder()
    .graph_degree(32)
    .intermediate_graph_degree(64)
    .nn_descent_niter(20)
    .build()?;
// Err("graph_degree must be > 0; got 0") fires immediately,
// not 1.8s later inside Index::build()
```

**Validation rules** (derived from C header constraints):
- `graph_degree > 0`; multiple of 32 preferred for warp alignment
- `intermediate_graph_degree >= graph_degree`
- `nn_descent_niter > 0`
- `itopk_size` must be a power of 2
- `hashmap_max_fill_rate` ∈ (0.1, 0.9)
- `team_size` ∈ {0, 4, 8, 16, 32}

## Backward Compatibility

The existing `IndexParams::new()` API and all setter methods are **unchanged**.
The builder is purely additive. No call sites break.

## Scope

Rust bindings only. No C API changes required. The validation logic duplicates
constraints already enforced by the C layer — we are surfacing them earlier,
before the FFI call, with useful error messages.

A proof-of-concept implementation is available at:
https://github.com/zbennett10/cuvs/tree/feat/rust-validated-index-params-builder

## Files

- `rust/cuvs/src/error.rs` — add `impl From<String> for Error`
- `rust/cuvs/src/cagra/index_params.rs` — `IndexParamsBuilder`
- `rust/cuvs/src/cagra/search_params.rs` — `SearchParamsBuilder`
- `rust/cuvs/src/cagra/mod.rs` — re-export builder types
- Apply same pattern to `ivf_pq`, `ivf_flat`, `vamana` for consistency

zbennett10:
## Summary

Closes #1859.

Adds `IndexParams::builder()` / `SearchParams::builder()` entry points to the
CAGRA, IVF-PQ, IVF-Flat, and Vamana Rust bindings. Invalid parameter values
(e.g. `graph_degree=0`) now produce a clear Rust error naming the offending
field and its valid range, before any FFI allocation or GPU work begins.

## Problem

The existing setter chain is silent about invalid values:

```rust
// Compiles fine. 1.823s later, inside Index::build():
// cuvsError { code: CUVS_ERROR, text: "raft::exception: assertion failed" }
let params = IndexParams::new()?
    .set_graph_degree(0)
    .set_intermediate_graph_degree(0)
    .set_nn_descent_niter(0);
let _index = Index::build(&res, &params, &dataset)?; // ← error here, no field name
```

## Solution

```rust
// Error fires immediately, before any GPU work:
// Err("graph_degree must be > 0; got 0")
let params = IndexParams::builder()
    .graph_degree(32)
    .intermediate_graph_degree(64)
    .nn_descent_niter(20)
    .build()?;
```

The builder provides two methods:
- `validate() -> Result<()>` — pure-Rust constraint checks, no GPU needed
- `build() -> Result<IndexParams>` — validates then allocates the FFI struct

The existing `IndexParams::new()?.set_*()` API is **completely unchanged**.

## Changes

| File | Change |
|------|--------|
| `src/error.rs` | Add `impl From<String> for Error` for validation errors |
| `src/cagra/index_params.rs` | `IndexParamsBuilder` + `IndexParams::builder()` |
| `src/cagra/search_params.rs` | `SearchParamsBuilder` + `SearchParams::builder()` |
| `src/cagra/mod.rs` | Re-export builder types |
| `examples/cagra.rs` | Demonstrate builder API |
| `src/ivf_pq/index_params.rs` | `IndexParamsBuilder` |
| `src/ivf_flat/index_params.rs` | `IndexParamsBuilder` |
| `src/vamana/index_params.rs` | `IndexParamsBuilder` |
| `src/{ivf_pq,ivf_flat,vamana}/mod.rs` | Re-export `IndexParamsBuilder` |

## Validation rules enforced

**CAGRA `IndexParams`:**
- `graph_degree > 0`
- `intermediate_graph_degree >= graph_degree`
- `nn_descent_niter > 0`

**CAGRA `SearchParams`:**
- `itopk_size` is a power of 2 (or 0 for auto)
- `team_size` ∈ {0, 4, 8, 16, 32}
- `hashmap_max_fill_rate` ∈ (0.1, 0.9)

**IVF-PQ / IVF-Flat `IndexParams`:**
- `n_lists > 0`
- `kmeans_trainset_fraction` ∈ (0, 1]

**Vamana `IndexParams`:**
- `graph_degree > 0`
- `visited_size >= graph_degree`
- `alpha > 0`

## Tests

All 6 specified PR gate tests are implemented. Tests 1–4 use `validate()` only
(no GPU required). Tests 5–6 call `build()` and require a CUDA device (will
run on RAPIDS CI):

1. `builder_rejects_zero_graph_degree` — validate only
2. `builder_rejects_invalid_intermediate_degree` — validate only
3. `builder_rejects_zero_niter` — validate only
4. `builder_accepts_valid_params` — validate only
5. `builder_round_trips_to_ffi` — requires GPU; compares builder vs. setter chain at FFI level
6. `existing_setter_api_unchanged` — requires GPU; confirms no regression

All 32 validation-logic tests verified locally via a standalone mock-FFI
test harness (pure Rust, no CUDA needed).

## Checklist

- [x] Existing `IndexParams::new()` + setter API is unchanged
- [x] `validate()` tests pass without a GPU (pure Rust logic)
- [x] `build()` / FFI round-trip tests require CUDA (RAPIDS CI)
- [x] `cargo +nightly fmt --check` clean
- [x] No breaking changes


yan-zaretskiy:
Hey Zachary! Thanks for the thorough work here. I agree that the current bindings lack Rust-side validation that we can report as proper Rust errors before they cross the FFI boundary.

I'd like to steer the implementation is a slightly different direction. We've been prototyping the next iteration of these Rust bindings, and we've landed on a pattern that avoids some of the pitfalls common with the hand-rolled builders. I think we can get there without too much rework.

One of the problems with your proposal is the default duplication between Rust and C. Each build has a manual `Default` impl like so:
```rust
impl Default for IndexParamsBuilder {                                                                                                                
    fn default() -> Self {                                
        Self {
            graph_degree: 64,               // must match C
            intermediate_graph_degree: 128, // must match C                                                                                          
            ...
        }                                                                                                                                            
    }                                                     
}
```
To avoid this, we should allocate the C params handle *first*, which gives us all the defaults, and then override explicitly set fields.

To avoid writing builders manually, I suggest to bring in the `bon` crate, and use its `#[builder]` macro on the `new` function, like so:

```rust
use bon::bon;                                             

#[bon]                                                                                                                                               
impl IndexParams {
    #[builder]                                                                                                                                       
    pub fn new(                                           
        graph_degree: Option<usize>,
        intermediate_graph_degree: Option<usize>,
        nn_descent_niter: Option<usize>,                                                                                                             
        build_algo: Option<BuildAlgo>,
        compression: Option<CompressionParams>,                                                                                                      
    ) -> Result<Self> {                                   
        // 1. Validate (only the fields that were set)                                                                                               
        if let Some(gd) = graph_degree {                                                                                                             
            if gd == 0 {                                                                                                                             
                return Err(/* ... */);                                                                                                               
            }                                                                                                                                        
        }                                                 
        // Cross-field validation where both are set, etc.
                                                                                                                                                     
        // 2. Allocate C params (gets exact library defaults)                                                                                     
        let params = Self::allocate_ffi()?;                                                                                                          
                                                                                                                                                     
        // 3. Override only what the caller specified     
        unsafe {                                                                                                                                     
            if let Some(v) = graph_degree {               
                (*params.0).graph_degree = v;
            }                                                                                                                                        
            // ...
        }                                                                                                                                            
        Ok(params)                                        
    }
}
```

With this approach, calling `IndexParams::builder().graph_degree(32).build()` just works, with proper validation defined in one place.

Next, the power-of-2 check on `itopk_size` is too strict. The C++ logic in `search_plan.cuh` only requires:

- `>= k` (the number of neighbors requested)
- `<= 1024` for the SINGLE_CTA algorithm

And it rounds up the provided value to a multiple of 32.

And the final suggestion is about error handling. Rust-side validation errors should be distinguishable from C library errors. Right now the `From<String> for Error` impl wraps them as `CuvsError`. A dedicated variant (e.g. `Error::Validation(String)`) would be cleaner.

yan-zaretskiy:
Closing this, since the proposed changes are now fully superseded by the `26.08` release, based on the `bon` crate and its fallible validated builder pattern.

PR #1864, opened 2026-03-03
Fix setting CAGRA graph build algo to iterative search by default

### LINKED ISSUE #1849: [BUG] cuvs-bench CAGRA default build algo
When cuvs-bench is used to run CAGRA benchmars, the user can specify the graph build algorithm using `graph_build_algo` parameter. If it is not specified explicitly, then currently the benchmark selects the ITERATIVE build algorithm. The ITERATIVE method is still experimental and its performance is not yet competitive with other build algorithms. Therefore it should not be used as default. 

We actually have a placeholder value for auto selecting the algorithm, and we should use that setting when the user does not specify the graph build algo.

achirkin:
When there are no graph build parameters set up and no `graph_build_algo` set explicitly in the CAGRA build params, ANN_BENCH defaults to iterative search, effectively bypassing the algorithm selection logic. This PR fixes that.

Resolves #1849

### DIFF
--- a/cpp/bench/ann/src/cuvs/cuvs_ann_bench_param_parser.h
+++ b/cpp/bench/ann/src/cuvs/cuvs_ann_bench_param_parser.h
@@ -282,6 +282,11 @@ void parse_build_param(const nlohmann::json& conf, cuvs::neighbors::cagra::index
             params.graph_build_params)) {
         params.graph_build_params = cuvs::neighbors::graph_build_params::ace_params{};
       }
+    } else if (conf.at("graph_build_algo") == "ITERATIVE_SEARCH") {
+      if (!std::holds_alternative<cuvs::neighbors::graph_build_params::iterative_search_params>(
+            params.graph_build_params)) {
+        params.graph_build_params = cuvs::neighbors::graph_build_params::iterative_search_params{};
+      }
     }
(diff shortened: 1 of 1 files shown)


cpp/bench/ann/src/cuvs/cuvs_ann_bench_param_parser.h
```diff
@@ -291,16 +296,18 @@ void parse_build_param(const nlohmann::json& conf, cuvs::neighbors::cagra::index
   nlohmann::json nn_descent_conf    = collect_conf_with_prefix(conf, "nn_descent_");
   nlohmann::json ace_conf           = collect_conf_with_prefix(conf, "ace_");
```
tarang-jain, line 297: Should we also collect iterative search params here and then check for that collected conf below in the monostate if condition?

tarang-jain: I think this will also require changes to `collect_conf_with_prefix`

achirkin: Thanks for noticing! No, the iterative search params are an empty data type - no parameters to collect

Issue #106, opened 2024-05-08
[FEA] Scalar quantization API


cjnolet:
We need a scalar quantization API that is decoupled from the IVF APIs, but which can ultimately be composed into an IVF. 

Ideally, we would follow a UX similar to FAISS or to scikit-learn's transformers estimators.

viclafargue:
Follow-up from the IVF-SQ review: `preprocessing::quantize::scalar` currently trains one global range, while IVF-SQ needs per-dimension ranges over residuals. The local `fused_column_minmax_kernel` in IVF-SQ is essentially doing this per-column range computation.

For reuse, we should add an optimized per-column scalar range-training primitive to `preprocessing::quantize::scalar`, preserving the current optimization properties: fused min/max computation in one pass and vectorized loads where possible.

Once that exists, IVF-SQ build can call the preprocessing primitive instead of carrying its own range-training kernel.

PR #1865, opened 2026-03-03
IVF-SQ C++ API

### LINKED ISSUE #1291: [FEA] Support IVF-SQ index

viclafargue:
Closes #1291.

### Overview
IVF-SQ combines an inverted file (IVF) partitioning scheme with 8-bit scalar quantization (SQ8) of residuals. Each float32 dimension is compressed to a single uint8 code, giving a 4x memory reduction over IVF-Flat while retaining high recall. The index implements various metrics (L2, inner-product, and cosine distance), data type (float, half) and also filtering.

### Build
- **K-Means clustering** : A training subset (controlled by `kmeans_trainset_fraction`) is sampled from the dataset. Balanced K-Means is run on it to produce `n_lists` centroids that partition the vector space.
- **SQ parameter training** : The training vectors are assigned to their nearest centroid and residuals are computed. A custom CUDA kernel computes the per-dimension min/max of these residuals (fast CUB reduction in shared memory). The observed range is expanded by a 5% margin on each side to reduce clipping on unseen data. The delta for each dimension is computed as `sq_delta[d] = (range + 2*margin) / 255`. These two per-dimension parameters (`sq_vmin` the lower end of the range, `sq_delta` the scale or quantization step) are stored in the index and are all that is needed to encode/decode any vector.
- **Data insertion** : If `add_data_on_build` is true (the default), the full dataset is inserted via the extend path described below.

### Extend
Extend adds new vectors to an existing index without retraining centroids or SQ parameters and in a batched fashion:
- **Cluster assignment** : New vectors are assigned to their nearest centroid via K-Means predict. ~(adaptative centers : when enabled, centroids are incrementally updated as new data arrives, and center norms are recomputed)~.
- **List resizing** : Per-list sizes are histogrammed and the IVF lists are grown to accommodate the new vectors, respecting the interleaved group alignment (kIndexGroupSize = 32).
- **Residual computation + SQ encoding** : A first kernel computes residuals. A second kernel quantize each residual dimension to `uint8` and write the code into the interleaved list layout.

### Search
Search proceeds in three stages:
- **Coarse search (GEMM-based)** : All query-to-centroid distances are computed in a single batched GEMM (`queries x centers^T`), with metric-specific pre/post-processing. The top `n_probes` nearest clusters per query are selected via `select_k`.
- **Fine scan (ivf_sq_scan_kernel)** : This is the performance-critical kernel, templated on `<BlockDim, Capacity, Metric, IdxT, SampleFilter>`. It operates in two grid modes controlled by the compile-time Capacity parameter:
  - **Fused top-k path** (Capacity > 0): grid is `(grid_dim_x, n_queries)`, each block loops over its share of probes and maintains a block_sort (warpsort) priority queue, writing only the local top-k to global memory. grid_dim_x is chosen via an occupancy query (`cudaOccupancyMaxActiveBlocksPerMultiprocessor`) to saturate the GPU. Instantiated for Capacity ∈ {32, 64, 128, 256}, covering k ≤ 256.
  - **Materialized path** (Capacity == 0): grid is `(n_probes, n_queries)`, one block per (query, probe) pair writes every distance to a flat buffer. Used as a fallback when k > 256 or the fused kernel has zero occupancy.

Three dim-length shared-memory arrays are filled in two phases: Phase 1 (invariant across probes) loads sq_delta and query terms; Phase 2 (per-probe) folds in the centroid so the inner loop reduces to s_query_term − code × s_sq_scale (L2) or s_query_term × (s_aux + code × s_sq_scale) (IP/Cosine). Each warp processes one interleaved group of 32 vectors; each lane loads a uint4, giving a fully coalesced 512-byte read per dimension block. Distance accumulation is fused with decoding — no separate decompress pass.
- **Final top-k selection** : In the fused path, a second select_k merges the per-block top-k results across grid_dim_x blocks (skipped when grid_dim_x == 1). In the materialized path, a single select_k over the full buffer extracts the global top-k. Distances then receive metric-specific fixups (e.g. sqrt for L2Sqrt) and list-local positions are mapped back to original dataset IDs.

### Benchmarks on B200

<div align="center">
Build benchmark (in seconds) :
ivf_sq_build_bench
<div/>
<br>
<br>
<div align="center">
wiki_all_10M-bench
<div/>
<br>
<div align="center">
deep-100M-bench
<div/>


cpp/include/cuvs/neighbors/ivf_sq.hpp
```diff
@@ -0,0 +1,336 @@
...
+ * @}
+ */
+
+/**
+ * @defgroup ivf_sq_cpp_list_spec IVF-SQ list storage spec
+ * @{
+ */
+
+template <typename SizeT, typename IdxT, typename ExtT>
+struct list_spec {
```
tarang-jain, line 89: Can we directly use the one in ivf_flat here? It seems like it is exactly the same for uint8

cpp/include/cuvs/neighbors/ivf_sq.hpp
```diff
@@ -0,0 +1,336 @@
...
+
+constexpr static uint32_t kIndexGroupSize = 32;
+
+struct index_params : cuvs::neighbors::index_params {
+  uint32_t n_lists                    = 1024;
+  uint32_t kmeans_n_iters             = 20;
+  double kmeans_trainset_fraction     = 0.5;
+  bool adaptive_centers               = false;
+  bool conservative_memory_allocation = false;
+  bool add_data_on_build              = true;
```
jinsolp: Can we add some documentation here? I think a lot of it can be copy-paste from the ivfpq docs:
https://github.com/rapidsai/cuvs/blob/b5eebba47193badd228055ca1a0682a18837de1d/cpp/include/cuvs/neighbors/ivf_pq.hpp#L44-L55

cpp/include/cuvs/neighbors/ivf_sq.hpp
```diff
@@ -0,0 +1,336 @@
...
+};
+
+struct search_params : cuvs::neighbors::search_params {
+  uint32_t n_probes = 20;
```
jinsolp, line 74: Documentations here too!

cpp/include/cuvs/neighbors/ivf_sq.hpp
```diff
@@ -0,0 +1,336 @@
...
+
+void search(raft::resources const& handle,
+            const cuvs::neighbors::ivf_sq::search_params& params,
+            const cuvs::neighbors::ivf_sq::index<uint8_t>& index,
+            raft::device_matrix_view<const half, int64_t, raft::row_major> queries,
+            raft::device_matrix_view<int64_t, int64_t, raft::row_major> neighbors,
+            raft::device_matrix_view<float, int64_t, raft::row_major> distances,
+            const cuvs::neighbors::filtering::base_filter& sample_filter =
+              cuvs::neighbors::filtering::none_sample_filter{});
+
```
jinsolp, line 562: Can we also have some docs here about params [in] [out] as well?

cpp/src/neighbors/ivf_sq/ivf_sq_build.cuh
```diff
@@ -0,0 +1,664 @@
...
+  RAFT_EXPECTS(n_rows > 0 && dim > 0, "empty dataset");
+  RAFT_EXPECTS(n_rows >= params.n_lists, "number of rows can't be less than n_lists");
+  RAFT_EXPECTS(params.metric != cuvs::distance::DistanceType::CosineExpanded || dim > 1,
+               "Cosine metric requires more than one dim");
+
+  index<IdxT> idx(handle, params, dim);
+  utils::memzero(idx.accum_sorted_sizes().data_handle(), idx.accum_sorted_sizes().size(), stream);
+  utils::memzero(idx.list_sizes().data_handle(), idx.list_sizes().size(), stream);
+  utils::memzero(idx.data_ptrs().data_handle(), idx.data_ptrs().size(), stream);
+  utils::memzero(idx.inds_ptrs().data_handle(), idx.inds_ptrs().size(), stream);
```
jinsolp: can we do these initializations inside the index constructor?

cpp/src/neighbors/ivf_sq/ivf_sq_build.cuh
```diff
@@ -0,0 +1,664 @@
...
+    rmm::device_uvector<T> trainset(
+      n_rows_train * idx.dim(), stream, raft::resource::get_large_workspace_resource(handle));
+    RAFT_CUDA_TRY(cudaMemcpy2DAsync(trainset.data(),
+                                    sizeof(T) * idx.dim(),
+                                    dataset,
+                                    sizeof(T) * idx.dim() * trainset_ratio,
+                                    sizeof(T) * idx.dim(),
+                                    n_rows_train,
+                                    cudaMemcpyDefault,
+                                    stream));
```
jinsolp: Doesn't this take the first `n_rows_train` number of vectors? Can we use `sample_rows` like ivfpq?

https://github.com/rapidsai/cuvs/blob/b5eebba47193badd228055ca1a0682a18837de1d/cpp/src/neighbors/ivf_pq/ivf_pq_build.cuh#L1298

cpp/src/neighbors/ivf_sq/ivf_sq_build.cuh
```diff
@@ -0,0 +1,664 @@
...
+      dim3 threads(32, 8);
+      dim3 blocks(raft::ceildiv<int64_t>(n_rows_train, threads.x),
+                  raft::ceildiv<uint32_t>(dim, threads.y));
+      compute_residuals_kernel<T><<<blocks, threads, 0, stream>>>(trainset.data(),
+                                                                  idx.centers().data_handle(),
+                                                                  train_labels.data_handle(),
+                                                                  residuals.data(),
+                                                                  n_rows_train,
+                                                                  dim);
+      RAFT_CUDA_TRY(cudaPeekAtLastError());
```
jinsolp: Looks like we don't use the `trainset` device array after this point. Do you think we can do `compute_residuals_kernel` in-place without allocating a new `residuals` device array? Think this could save some space!

cpp/src/neighbors/ivf_sq/ivf_sq_build.cuh
```diff
@@ -0,0 +1,664 @@
...
+        handle, pred_params, trainset_const_view, centers_const_view, train_labels.view());
+      raft::resource::sync_stream(handle);
+    }
+
+    rmm::device_uvector<float> residuals(
+      n_rows_train * dim, stream, raft::resource::get_large_workspace_resource(handle));
+    {
+      dim3 threads(32, 8);
+      dim3 blocks(raft::ceildiv<int64_t>(n_rows_train, threads.x),
+                  raft::ceildiv<uint32_t>(dim, threads.y));
```
jinsolp: Asking some questions here: 
- What was the motivation behind choosing (32,8) for the threads? It looks like each warp will have less memory coalescing opportunities because this is a row major array and we only have coalescing of 8 elements with (32, 8). 
- Also, if the 32 rows all have different labels, this means we are loading 32 different centroid vectors for this block. For arrays with larger dimensions it might be better to launch 1 block per row of the `trainset`.

cpp/src/neighbors/ivf_sq/ivf_sq_build.cuh
```diff
@@ -0,0 +1,664 @@
...
+  auto new_labels =
+    raft::make_device_mdarray<LabelT>(handle,
+                                      raft::resource::get_large_workspace_resource(handle),
+                                      raft::make_extents<int64_t>(n_rows));
+  cuvs::cluster::kmeans::balanced_params kmeans_params;
+  kmeans_params.metric     = index->metric();
+  auto orig_centroids_view = raft::make_device_matrix_view<const float, int64_t>(
+    index->centers().data_handle(), n_lists, dim);
+
+  constexpr size_t kReasonableMaxBatchSize = 65536;
```
jinsolp, line 299: how is this determined? 👀

cpp/src/neighbors/ivf_sq/ivf_sq_build.cuh
```diff
@@ -0,0 +1,664 @@
...
+      new_labels.data_handle() + batch.offset(), batch.size());
+    cuvs::cluster::kmeans::predict(
+      handle, kmeans_params, batch_data_view, orig_centroids_view, batch_labels_view);
+    vec_batches.prefetch_next_batch();
+    raft::resource::sync_stream(handle);
+  }
+
+  auto* list_sizes_ptr    = index->list_sizes().data_handle();
+  auto old_list_sizes_dev = raft::make_device_mdarray<uint32_t>(
+    handle, raft::resource::get_workspace_resource(handle), raft::make_extents<int64_t>(n_lists));
```
jinsolp: Just a nitpick: is there a reason we can't just use `make_device_vector` here?

cpp/src/neighbors/ivf_sq/ivf_sq_build.cuh
```diff
@@ -0,0 +1,664 @@
...
+        handle, idx.sq_vmin(), [vmin_ptr, vmax_ptr, delta_ptr, kMargin] __device__(uint32_t j) {
+          float range  = vmax_ptr[j] - vmin_ptr[j];
+          float margin = range * kMargin;
+          delta_ptr[j] = (range > 0.0f) ? (range + 2.0f * margin) / 255.0f : 1.0f;
+          return vmin_ptr[j] - margin;
+        });
+    }
+  }
+
+  if (params.add_data_on_build) { detail::extend<T, IdxT>(handle, &idx, dataset, nullptr, n_rows); }
```
jinsolp: maybe check whether if we're adding data on build, we don't have `adaptive_centers=true`? Think this will take us through unnecessary code paths in the `extend` function.

cpp/src/neighbors/ivf_sq/ivf_sq_build.cuh
```diff
@@ -0,0 +1,664 @@
...
+  auto stream  = raft::resource::get_cuda_stream(handle);
+  auto n_lists = index->n_lists();
+  auto dim     = index->dim();
+  list_spec<uint32_t, IdxT, int64_t> list_device_spec{index->dim(),
+                                                      index->conservative_memory_allocation()};
+  cuvs::common::nvtx::range<cuvs::common::nvtx::domain::cuvs> fun_scope(
+    "ivf_sq::extend(%zu, %u)", size_t(n_rows), dim);
+
+  RAFT_EXPECTS(new_indices != nullptr || index->size() == 0,
+               "You must pass data indices when the index is non-empty.");
```
jinsolp, line 288: Can we also add a check to ensure somehow that the newly passed indices are not overlapping with any that we have in the index already? Or is this supposed to be up to the user? Asking this because right now we would end up with duplicate indices in the index if the user passes the wrong optional indices.

jinsolp (commented):
Thanks @viclafargue ! Sharing my first batch of comments for `build` and `extend`.
Aside from this, I also think `ivf_sq.hpp` overall needs some documentation!

cjnolet:
@viclafargue can you please share the build time speedups over Faiss IVFSQ on GPU and CPU? Those are going to be crucial as the major value prop of cuVS over Faiss (both cpu and gpu) is index build speedup.

cpp/include/cuvs/neighbors/ivf_sq.hpp
viclafargue: Answered in https://github.com/rapidsai/cuvs/pull/1865/changes/3a911d887a19471137085bcc19f6ab881af5f76f.

cpp/include/cuvs/neighbors/ivf_sq.hpp
viclafargue: Answered in https://github.com/rapidsai/cuvs/pull/1865/changes/3a911d887a19471137085bcc19f6ab881af5f76f.

cpp/include/cuvs/neighbors/ivf_sq.hpp
viclafargue: Answered in https://github.com/rapidsai/cuvs/pull/1865/changes/3a911d887a19471137085bcc19f6ab881af5f76f.

cpp/src/neighbors/ivf_sq/ivf_sq_build.cuh
viclafargue: Answered in https://github.com/rapidsai/cuvs/pull/1865/changes/b1246284aef228e2e5a255e862149523716e1238.

cpp/src/neighbors/ivf_sq/ivf_sq_build.cuh
viclafargue: Answered in https://github.com/rapidsai/cuvs/pull/1865/changes/641c6ca34babd5d2914d6c1b6c7bbe47aec92980.

cpp/src/neighbors/ivf_sq/ivf_sq_build.cuh
viclafargue: Answered in https://github.com/rapidsai/cuvs/pull/1865/commits/70ca00a331cd96138b0eab3a57bb135858882341.

cpp/src/neighbors/ivf_sq/ivf_sq_build.cuh
viclafargue: Answered in https://github.com/rapidsai/cuvs/pull/1865/commits/e7d660cddcc9e197c12dacae0db54a978d75fb1e.

cpp/src/neighbors/ivf_sq/ivf_sq_build.cuh
viclafargue: I actually decided to drop the `adaptative_centers` config parameter and feature altogether, because moving the centroids without updating the residual encoding lists is not a good idea and updating the list would be prohibitive in term of perf. Addressed in https://github.com/rapidsai/cuvs/pull/1865/changes/206cb2e612538afadbb3017d90c668f742b5f7e1.

cpp/src/neighbors/ivf_sq/ivf_sq_build.cuh
viclafargue: Addressed in https://github.com/rapidsai/cuvs/pull/1865/commits/96b28db42cbb0aae730fefffc13ba0da2c763f59.

cpp/src/neighbors/ivf_sq/ivf_sq_build.cuh
viclafargue: None of the other indexes check for duplicates and the performance cost of doing so would be prohibitive. It should be the responsibility of the user to not insert duplicates.

cpp/src/neighbors/ivf_sq/ivf_sq_build.cuh
viclafargue: This is an arbitrary value used in other index types. This probably was a safe value that ensured that we never ran out of memory in smaller systems. But, thanks for pointing this out as it might be interesting to see if we could improve the way this value is determined (dataset dimensions and available VRAM). cc @achirkin @tfeher

cpp/src/neighbors/ivf_sq/ivf_sq_search.cuh
```diff
@@ -0,0 +1,549 @@
...
+                                  float* distances,
+                                  IvfSampleFilterT sample_filter = IvfSampleFilterT())
+{
+  cuvs::common::nvtx::range<cuvs::common::nvtx::domain::cuvs> fun_scope(
+    "ivf_sq::search(k = %u, n_queries = %u, dim = %zu)", k, n_queries, index.dim());
+
+  RAFT_EXPECTS(params.n_probes > 0,
+               "n_probes (number of clusters to probe in the search) must be positive.");
+  auto n_probes = std::min<uint32_t>(params.n_probes, index.n_lists());
+
```
jinsolp, line 677: Can we add a log here if n_probes was reset to `index.n_lists()` to let the user know?

cpp/src/neighbors/ivf_sq/ivf_sq_serialize.cuh
```diff
@@ -0,0 +1,159 @@
...
+constexpr int serialization_version = 1;
+
+template <typename IdxT>
+void serialize(raft::resources const& handle, std::ostream& os, const index<IdxT>& index_)
+{
+  RAFT_LOG_DEBUG(
+    "Saving IVF-SQ index, size %zu, dim %u", static_cast<size_t>(index_.size()), index_.dim());
+
+  std::string dtype_string = raft::detail::numpy_serializer::get_numpy_dtype<IdxT>().to_string();
+  dtype_string.resize(4);
```
jinsolp, line 33: is there a reason this is chosen to be 4?

cpp/src/neighbors/ivf_sq_index.cpp
```diff
@@ -0,0 +1,235 @@
...
+    lists_{n_lists},
+    list_sizes_{raft::make_device_vector<uint32_t, uint32_t>(res, n_lists)},
+    centers_(raft::make_device_matrix<float, uint32_t>(res, n_lists, dim)),
+    center_norms_(std::nullopt),
+    sq_vmin_{raft::make_device_vector<float, uint32_t>(res, dim)},
+    sq_delta_{raft::make_device_vector<float, uint32_t>(res, dim)},
+    data_ptrs_{raft::make_device_vector<IdxT*, uint32_t>(res, n_lists)},
+    inds_ptrs_{raft::make_device_vector<int64_t*, uint32_t>(res, n_lists)},
+    accum_sorted_sizes_{raft::make_host_vector<int64_t, uint32_t>(n_lists + 1)}
+{
```
jinsolp, line 45: Maybe add checks like `RAFT_EXPECTS(n_lists > 0)`?

cpp/src/neighbors/ivf_sq_index.cpp
```diff
@@ -0,0 +1,235 @@
...
+
+template <typename IdxT>
+const std::vector<std::shared_ptr<list_data<IdxT, int64_t>>>& index<IdxT>::lists() const noexcept
+{
+  return lists_;
+}
+
+template <typename IdxT>
+void index<IdxT>::check_consistency()
+{
```
jinsolp, line 223: can we also check of `sq_vmin` and `sq_delta` matching dim here?

jinsolp (commented):
Second batch of comments! I looked through search and serialize this time, looks good!

cpp/src/neighbors/ivf_sq/ivf_sq_serialize.cuh
viclafargue: This is common to all index types. The [`to_string` function](https://github.com/rapidsai/raft/blob/5b8a7dd6d4ad2241195390117bd459bb7e77abdf/cpp/include/raft/core/detail/mdspan_numpy_serializer.hpp#L86) produces a Numpy dtype descriptor string with one char for the byte order (`<` or `>`), one chat for the kind and (`f`, `i`, `u`, ...), and one or two chars for the itemsize (itemsize > 0 would be encoded with two chars). Since the deserialization process does not know what to expect, we always serialize 4 characters (and possibly append a null char if it would otherwise hold on 3).

cpp/src/neighbors/ivf_sq/ivf_sq_build.cuh
```diff
@@ -0,0 +1,640 @@
...
+    uint32_t j                                              = d % veclen;
+    list_dat[l * kIndexGroupSize + ingroup_id * veclen + j] = out;
+  }
+}
+
+/**
+ * Compute residuals: residual[i] = cast<float>(x_i) - centers[labels[i]]
+ */
+template <typename T>
+RAFT_KERNEL compute_residuals_kernel(const T* dataset,
```
tarang-jain: Can we use raft::linalg::map for this instead of writing a kernel?

cpp/src/neighbors/ivf_sq/ivf_sq_build.cuh
```diff
@@ -0,0 +1,640 @@
...
+    // and derive per-dimension vmin/delta from them.
+    {
+      auto train_labels = raft::make_device_vector<uint32_t, int64_t>(handle, n_rows_train);
+      cuvs::cluster::kmeans::balanced_params pred_params;
+      pred_params.metric      = idx.metric();
+      auto centers_const_view = raft::make_device_matrix_view<const float, int64_t>(
+        idx.centers().data_handle(), idx.n_lists(), dim);
+      cuvs::cluster::kmeans::predict(
+        handle, pred_params, trainset_const_view, centers_const_view, train_labels.view());
+      raft::resource::sync_stream(handle);
```
tarang-jain: do we need the sync here?

cpp/src/neighbors/ivf_sq/ivf_sq_build.cuh
```diff
@@ -0,0 +1,640 @@
...
+    auto dataset_view = raft::make_device_matrix_view<const T, int64_t>(dataset, n_rows, idx.dim());
+    raft::matrix::sample_rows<T, int64_t>(handle, random_state, dataset_view, trainset.view());
+    auto trainset_const_view = raft::make_const_mdspan(trainset.view());
+    auto centers_view        = raft::make_device_matrix_view<float, int64_t>(
+      idx.centers().data_handle(), idx.n_lists(), idx.dim());
+    cuvs::cluster::kmeans::balanced_params kmeans_params;
+    kmeans_params.n_iters = params.kmeans_n_iters;
+    kmeans_params.metric  = idx.metric();
+    cuvs::cluster::kmeans::fit(handle, kmeans_params, trainset_const_view, centers_view);
+    raft::resource::sync_stream(handle);
```
tarang-jain: do we need the sync here?

cpp/src/neighbors/ivf_sq/ivf_sq_serialize.cuh
jinsolp: I see, thanks for clarifying!

cpp/include/cuvs/neighbors/ivf_sq.hpp
viclafargue: Not directly. `make_list_extents` differs. Here we pad the dimension to a multiple of 16 (`kVecLen`) for vectorized access to the quantized codes, while the IVF-Flat version uses the raw dimension as-is. The rest of the struct is identical.

cpp/src/neighbors/ivf_sq/ivf_sq_build.cuh
viclafargue: You're right, this is not needed.

cpp/src/neighbors/ivf_sq/ivf_sq_build.cuh
viclafargue: You're right, this is not needed.

cpp/src/neighbors/ivf_sq/ivf_sq_build.cuh
viclafargue: I replaced it with a `raft::linalg::map_offset` call.

cpp/tests/neighbors/ann_ivf_sq/test_half_int64_t.cu
```diff
@@ -0,0 +1,30 @@
+/*
```
jinsolp, line 1: I think the filename for this is a bit misleading! Following existing convention, i think it'll be better to separate the float and half `.cu` files

cpp/tests/neighbors/ann_ivf_sq.cuh
```diff
@@ -0,0 +1,518 @@
...
+ public:
+  AnnIVFSQTest()
+    : stream_(raft::resource::get_cuda_stream(handle_)),
+      ps(::testing::TestWithParam<AnnIvfSqInputs<IdxT>>::GetParam()),
+      database(0, stream_),
+      search_queries(0, stream_)
+  {
+  }
+
+  void testIVFSQ()
```
jinsolp: I think this test is doing too many things. Can we split this up to different tests that check
1. comparing search results against naive knn
2. serialize - deserialize round trip
3. comparing search results of `idx` and `idx_empty`

cpp/tests/neighbors/ann_ivf_sq.cuh
```diff
@@ -0,0 +1,518 @@
...
+
+ private:
+  raft::resources handle_;
+  rmm::cuda_stream_view stream_;
+  AnnIvfSqInputs<IdxT> ps;
+  rmm::device_uvector<DataT> database;
+  rmm::device_uvector<DataT> search_queries;
+};
+
+const std::vector<AnnIvfSqInputs<int64_t>> inputs = {
```
jinsolp, line 382: great coverage! 🙌

jinsolp (commented):
Finished sweeping all changes. I'll do one last sweep after this for a final check!
I'm not too familiar with cuvs bench, so didn't have feedback on that part

gforsyth (approved):

cpp/src/neighbors/ivf_sq/ivf_sq_build.cuh
achirkin: That is correct, and one more argument here is that this size is well beyond the point where the GPU is saturated:
```
{big enough grid size} < kReasonableMaxBatchSize < {small enough memory footprint}
```

cpp/include/cuvs/neighbors/ivf_sq.hpp
```diff
@@ -0,0 +1,856 @@
...
+ * distances between vectors with high fidelity, offering a good trade-off between index size,
+ * search speed, and recall compared to flat (uncompressed) and product-quantized (PQ)
+ * representations.
+ *
+ * @tparam IdxT  SQ code type. Only uint8_t (8-bit, codes in [0,255]) for now.
+ *
+ */
+template <typename IdxT>
+struct index : cuvs::neighbors::index {
+  static_assert(std::is_same_v<IdxT, uint8_t>, "IVF-SQ code type IdxT must be uint8_t for now.");
```
lowener: IdxT usually refers to the indexing type (`uint32_t` / `int64_t`). Here the template name should be changed (maybe to CodeT)

cpp/src/neighbors/ivf_sq/ivf_sq_build.cuh
```diff
@@ -0,0 +1,608 @@
...
+  ColMinMaxPair agg = {std::numeric_limits<float>::max(), std::numeric_limits<float>::lowest()};
+
+  const int64_t stride = static_cast<int64_t>(BlockSize);
+  int64_t row          = static_cast<int64_t>(threadIdx.x);
+
+  for (; row + 3 * stride < n_rows; row += 4 * stride) {
+    float v0    = float(data[row * dim + col]);
+    float v1    = float(data[(row + stride) * dim + col]);
+    float v2    = float(data[(row + 2 * stride) * dim + col]);
+    float v3    = float(data[(row + 3 * stride) * dim + col]);
```
lowener, line 89: Have you considered using vectorized loads of 2 floats to compute 2 columns and 4 rows in the same loop? I used that in PQ and saw a great speed-up there. But it also depends on the dimensionality and on which function is the current bottleneck in building an IVF-SQ index.

cpp/src/neighbors/ivf_sq/ivf_sq_build.cuh
```diff
@@ -0,0 +1,608 @@
...
+  uint32_t padded_dim       = ((dim + veclen - 1) / veclen) * veclen;
+  auto* list_dat   = list_data_ptrs[list_id] + static_cast<size_t>(group_offset) * padded_dim;
+  const float* src = residuals + row_id * dim;
+
+  for (uint32_t d = lane_id; d < padded_dim; d += kWarpSize) {
+    uint8_t out;
+    if (d < dim) {
+      float val  = src[d];
+      float dv   = delta[d];
+      float v    = vmin[d];
```
lowener, line 196: Same, can each lane use vectorized load of 2 or 4? Would that improve build time?

cpp/src/neighbors/ivf_sq/ivf_sq_build.cuh
```diff
@@ -0,0 +1,608 @@
...
+
+  for (const auto& batch : vec_batches) {
+    auto batch_data_view =
+      raft::make_device_matrix_view<const T, int64_t>(batch.data(), batch.size(), index->dim());
+    auto batch_labels_view = raft::make_device_vector_view<LabelT, int64_t>(
+      new_labels.data_handle() + batch.offset(), batch.size());
+    cuvs::cluster::kmeans::predict(
+      handle, kmeans_params, batch_data_view, orig_centroids_view, batch_labels_view);
+    vec_batches.prefetch_next_batch();
+    raft::resource::sync_stream(handle);
```
lowener, line 282: Sync stream only if prefetch happened: enable_prefetch is true and dataset is on host.
Otherwise it can execute without sync

cpp/src/neighbors/ivf_sq/ivf_sq_build.cuh
```diff
@@ -0,0 +1,608 @@
...
+                                             index->sq_vmin().data_handle(),
+                                             index->sq_delta().data_handle(),
+                                             bs,
+                                             dim,
+                                             batch.offset());
+      RAFT_CUDA_TRY(cudaPeekAtLastError());
+    }
+
+    vec_batches.prefetch_next_batch();
+    raft::resource::sync_stream(handle);
```
lowener, line 371: Same, the sync might be unecessary

cpp/src/neighbors/ivf_sq/ivf_sq_build.cuh
```diff
@@ -0,0 +1,608 @@
...
+    RAFT_CUDA_TRY(cudaPeekAtLastError());
```
lowener, line 372: This is unecessary, it was checked just before

cpp/src/neighbors/ivf_sq/ivf_sq_build.cuh
```diff
@@ -0,0 +1,608 @@
...
+    kmeans_params.n_iters = params.kmeans_n_iters;
+    kmeans_params.metric  = idx.metric();
+    cuvs::cluster::kmeans::fit(handle, kmeans_params, trainset_const_view, centers_view);
+
+    // Train SQ: predict labels for the training subset, compute residuals in-place,
+    // and derive per-dimension vmin/delta from them.
+    {
+      auto train_labels = raft::make_device_vector<uint32_t, int64_t>(handle, n_rows_train);
+      cuvs::cluster::kmeans::balanced_params pred_params;
+      pred_params.metric      = idx.metric();
```
lowener, line 461: Reuse the kmeans::balanced params declared just above

cpp/tests/neighbors/ann_ivf_sq/test_float_int64_t.cu
```diff
@@ -0,0 +1,20 @@
...
+
+#include "../ann_ivf_sq.cuh"
+
+namespace cuvs::neighbors::ivf_sq {
+
+typedef AnnIVFSQTest<float, float, int64_t> AnnIVFSQTestF_float;
+TEST_P(AnnIVFSQTestF_float, AnnIVFSQSearch) { this->testSearch(); }
+TEST_P(AnnIVFSQTestF_float, AnnIVFSQSerialize) { this->testSerialize(); }
+TEST_P(AnnIVFSQTestF_float, AnnIVFSQExtend) { this->testExtend(); }
+TEST_P(AnnIVFSQTestF_float, AnnIVFSQFilter) { this->testFilter(); }
```
lowener, line 16: Wont this will lead to building the index 4 time for each test input? Can we build once and then do all the tests?

lowener (changes requested):

cpp/src/neighbors/ivf_sq/ivf_sq_build.cuh
viclafargue: Switched to a vectorized load.

cpp/src/neighbors/ivf_sq/ivf_sq_build.cuh
viclafargue: The reads here are already warp-coalesced.

cpp/src/neighbors/ivf_sq/ivf_sq_build.cuh
```diff
@@ -0,0 +1,726 @@
...
+ * The primary benefit over scalar loads is halving (VecCols=2) or
+ * quartering (VecCols=4) the number of LDG instructions issued per warp,
+ * which is the dominant cost in the column-strided access pattern of
+ * fused_column_minmax_kernel. VecCols=1 is provided as the degenerate
+ * scalar fallback for odd `dim`.
+ *
+ * Requires `p` to be aligned to sizeof(T) * VecCols.
+ */
+template <typename T, int VecCols>
+struct vec_loader;
```
lowener: Can raft::TxN_t be re-used here?

viclafargue:
The wheel-build CI was failing because the IVF-SQ scan kernel pushed `libcuvs.so` over the pip wheel size limit. To unblock the PR, I ported the scan kernel to JIT-LTO (same model as IVF-Flat). The kernel body now lives in `cpp/src/neighbors/ivf_sq/detail/jit_lto_kernels/scan_impl.cuh` and is linked at runtime from generated fatbins instead of being baked into the library.

cpp/src/neighbors/ivf_sq/ivf_sq_build.cuh
viclafargue: Nice catch!

cpp/src/neighbors/ivf_sq/detail/jit_lto_kernels/scan_impl.cuh
```diff
@@ -0,0 +1,260 @@
...
+//     query - reconstructed = (query - vmin - centroid) - code*delta
+//                           = s_query_term - code*s_sq_scale.
+//
+//   IP/Cosine path:
+//     Phase 1: s_query_term[d] = query[d]                  (invariant)
+//     Phase 2: s_aux[d] = centroid[d] + vmin[d]            (per-probe)
+//     Reconstructed vector component: s_aux[d] + code*s_sq_scale[d].
+//
+//   After all probes are scanned, the smem is reused for block_sort merge.
+template <int Capacity, typename MetricTag>
```
divyegala: We should not structure the kernel in a way where we need `MetricTag`. Usually, that is a way to tell us that related functionality could be split out in another device function fragment. Look at @KyleFromNVIDIA's work with IVF PQ for examples.

Also Kyle's rule of thumb: if a template parameter does not have a function parameter argument, it is likely that code related to that template parameter can be pulled into another fragment.

divyegala (changes requested):

KyleFromNVIDIA (approved):
Approved CMake changes and JIT+LTO changes. Very nice.

cpp/src/neighbors/ivf_sq/detail/jit_lto_kernels/scan_impl.cuh
KyleFromNVIDIA: The rule of thumb that a template parameter missing from a signature means the templated code can be split out into another fragment only holds when said template parameter isn't _tightly_ integrated into the body of the function. This one is borderline. Each fatbin generated by this fragment is <20KB (for a total of ~300KB), and I don't know how much of a gain we'd get by splitting out the metric-specific parts into their own fragment (it would be ~200KB at most, and that's not including the additional metric fragment). I'm inclined to think it's not worth it. At the very least, we can certainly save it for a follow-up.

My approval stands.

divyegala: I think `kAscending` could be promoted to a template parameter of the main kernel as many metrics would share the same value (most would be `true`?), so you wouldn't have to instantiate this kernel for every metric and internally infer the same value.

The parts related to `kIsCosine` can be separated out to different fragments, a `cosine` specific one and a `no-op` one.

viclafargue: Thanks both for the careful reviews! Took some time to test different things. Conclusion : even though the metrics are tightly coupled with the main body it is still possible to recover some beyond the filter fragment.

Two changes:

1) **Move metric specialization out of the kernel template:** into four device-function fragments :
- `setup_invariant_smem`
- `setup_per_probe_smem`
- `accumulate_distance`
- `finalize_distance` (the cosine vs no-op one)

The entrypoint now needs `(Capacity, Ascending)` instead of `(Capacity, MetricTag)`.

2) **Apply the IVF-PQ inner-product trick:** noticed that IVF-PQ negates IP scores in the kernel so `warpsort::block_sort` always runs ascending, then flip them back in `postprocess_distances` via `account_for_max_close`. Ascending becomes hardcoded, so the kernel is now templated on Capacity alone.

**Net effect:**
- Before : 5 Capacity * 3 Metrics -> 15 main body instantiations -> 91MB `libcuvs.so`
- After : 5 Capacity -> 5 main body instantiations -> 89MB `libcuvs.so`
- No measurable performance impact

divyegala (approved):
JIT-LTO related CMake and C++ approval. Awesome work!

cpp/src/neighbors/ivf_sq/ivf_sq_serialize.cuh
```diff
@@ -0,0 +1,159 @@
...
+{
+  std::ofstream of(filename, std::ios::out | std::ios::binary);
+  if (!of) { RAFT_FAIL("Cannot open file %s", filename.c_str()); }
+  detail::serialize(handle, of, index_);
+  of.close();
+  if (!of) { RAFT_FAIL("Error writing output %s", filename.c_str()); }
+}
+
+template <typename CodeT>
+auto deserialize(raft::resources const& handle, std::istream& is) -> index<CodeT>
```
lowener, line 88: Implement checks present in https://github.com/rapidsai/cuvs/pull/2068 to protect against integers overflow in deserialization.

lowener (approved):
LGTM

cpp/include/cuvs/neighbors/ivf_sq.hpp
```diff
@@ -0,0 +1,860 @@
...
+  constexpr explicit list_spec(const list_spec<OtherSizeT, CodeT, IdxT>& other_spec)
+    : dim{other_spec.dim}, align_min{other_spec.align_min}, align_max{other_spec.align_max}
+  {
+  }
+
+  static constexpr uint32_t kVecLen = 16;
+
+  constexpr auto make_list_extents(SizeT n_rows) const -> list_extents
+  {
+    uint32_t padded = ((dim + kVecLen - 1) / kVecLen) * kVecLen;
```
achirkin: nitpick: use `raft::round_up_safe`

cpp/src/neighbors/ivf_sq/detail/jit_lto_kernels/scan_impl.cuh
```diff
@@ -0,0 +1,260 @@
...
+
+      float dist      = 0.0f;
+      float v_norm_sq = 0.0f;
+
+      const uint8_t* group_data = codes + size_t(group) * padded_dim;
+
+      for (uint32_t bl = 0; bl < n_dim_blocks; bl++) {
+        uint8_t codes_local[veclen];
+        *reinterpret_cast<uint4*>(codes_local) = *reinterpret_cast<const uint4*>(
+          group_data + bl * (veclen * kIndexGroupSize) + lane_id * veclen);
```
achirkin: Why do you cast this to uint4 specifically? Is it to enforce 16-byte-long load from global memory? Do you ever foresee `veclen` being different?
In any case, I'd suggest to use [raft::ldg](https://github.com/rapidsai/raft/blob/main/cpp/include/raft/util/device_loads_stores.cuh) or the [vectorized helper struct](https://github.com/rapidsai/raft/blob/main/cpp/include/raft/util/vectorized.cuh) whatever is more appropriate.

cpp/src/neighbors/ivf_sq/detail/jit_lto_kernels/scan_impl.cuh
```diff
@@ -0,0 +1,260 @@
...
+    __syncthreads();  // (b)
+
+    if (cluster_sz == 0) {
+      if constexpr (!kManageLocalTopK) break;
+      continue;
+    }
+
+    const uint8_t* codes   = data_ptrs[cluster_id];
+    uint32_t sample_offset = (probe_ix > 0) ? my_chunk[probe_ix - 1] : 0;
+    uint32_t padded_dim    = ((dim + veclen - 1) / veclen) * veclen;
```
achirkin: nitpick: `raft::Pow2<veclen>::roundUp(dim)` not extremely obvious, but at least contains the word "roundUp".

achirkin (commented):
A few small suggestions

cpp/include/cuvs/neighbors/ivf_sq.hpp
```diff
@@ -0,0 +1,860 @@
...
+
+constexpr static uint32_t kIndexGroupSize = 32;
+
+struct index_params : cuvs::neighbors::index_params {
+  /** The number of inverted lists (clusters) */
+  uint32_t n_lists = 1024;
+  /** The number of iterations searching for kmeans centers (index building). */
+  uint32_t kmeans_n_iters = 20;
+  /** The fraction of data to use during iterative kmeans building. */
+  double kmeans_trainset_fraction = 0.5;
```
tfeher: Would be great to replace `kmeans_trainset_fraction` with the following parameter
```suggestion
  /** The number of data vectors (per cluster) to use during iterative kmeans building. */
  uint32_t max_train_points_per_cluster = 256;
```

- This way it is easier for scale the number of clusters during benchmark, and the kmeans subsampling is adjusted accordingly.
- This is similar how we configure CAGRA-Q, and the `preprocessing::quantize::pq`. We plan to add the same param IVF-RaBitQ.
- This aligns with FAISS k-means training parameters

cpp/bench/ann/src/cuvs/cuvs_ann_bench_param_parser.h
```diff
@@ -86,6 +91,26 @@ void parse_search_param(const nlohmann::json& conf,
...
 #endif
 
+#ifdef CUVS_ANN_BENCH_USE_CUVS_IVF_SQ
+template <typename T>
+void parse_build_param(const nlohmann::json& conf,
+                       typename cuvs::bench::cuvs_ivf_sq<T>::build_param& param)
+{
+  param.n_lists = conf.at("nlist");
+  if (conf.contains("niter")) { param.kmeans_n_iters = conf.at("niter"); }
+  if (conf.contains("ratio")) {
```
tfeher, line 101: Please replace this with `max_train_points_per_cluster`

cpp/include/cuvs/neighbors/ivf_sq.hpp
```diff
@@ -0,0 +1,860 @@
...
+namespace cuvs::neighbors::ivf_sq {
+
+/**
+ * @defgroup ivf_sq_cpp_index_params IVF-SQ index build parameters
+ * @{
+ */
+
+constexpr static uint32_t kIndexGroupSize = 32;
+
+struct index_params : cuvs::neighbors::index_params {
```
tfeher, line 33: Please clarify in the docstring of `index_params` that we are using 8-bit scalar quantization.

cpp/include/cuvs/neighbors/ivf_sq.hpp
```diff
@@ -0,0 +1,860 @@
...
+ * @param[in] handle
+ * @param[in] index_params configure the index building
+ * @param[in] dataset raft::device_matrix_view to a row-major matrix [n_rows, dim]
+ * @param[out] idx reference to ivf_sq::index
+ *
+ */
+void build(raft::resources const& handle,
+           const cuvs::neighbors::ivf_sq::index_params& index_params,
+           raft::device_matrix_view<const float, int64_t, raft::row_major> dataset,
+           cuvs::neighbors::ivf_sq::index<uint8_t>& idx);
```
tfeher: Consider supporting only one overload, that returns the index. See offline discussion.

cpp/include/cuvs/neighbors/ivf_sq.hpp
```diff
@@ -0,0 +1,860 @@
...
+ * @param[in] handle
+ * @param[in] new_vectors a device matrix view to a row-major matrix [n_rows, idx.dim()]
+ * @param[in] new_indices a device vector view to a vector of indices [n_rows].
+ *    If the original index is empty (`idx.size() == 0`), you can pass `std::nullopt`
+ *    here to imply a continuous range `[0...n_rows)`.
+ * @param[in] orig_index the original index
+ *
+ * @return the constructed extended ivf-sq index
+ */
+auto extend(raft::resources const& handle,
```
tfeher: Let's also reconsider which overloads do we need for extend. This is the current status:


extend() | brute_force | IVF-Flat | IVF-PQ | CAGRA | IVF-SQ
-- | -- | -- | -- | -- | --
void extend(index*) |   | Y | Y |   | Y
void extend(index& ) |   |   |   | Y |  
auto extend(index&) -> index |   | Y | Y |   | Y

tfeher (changes requested):
Hi @viclafargue thanks for this work. Looks great overall, but I am still going through the details. Here are my first comments related to the public API.

cpp/src/neighbors/ivf_sq/ivf_sq_build.cuh
```diff
@@ -0,0 +1,615 @@
...
+    fused_column_minmax_kernel<BlockSize, 2, T>
+      <<<dim / 2, BlockSize, 0, stream>>>(data, col_min, col_max, n_rows, dim);
+  } else {
+    fused_column_minmax_kernel<BlockSize, 1, T>
+      <<<dim, BlockSize, 0, stream>>>(data, col_min, col_max, n_rows, dim);
+  }
+}
+
+template <typename CodeT>
+auto clone(const raft::resources& res, const index<CodeT>& source) -> index<CodeT>
```
tfeher: Is this used anywhere?

tfeher (commented):
Hi @viclafargue, the updates in the build / extend API look good to me.

I have one question of the implementation of `build()`. I would have hoped that `preprocessing::quantize::scalar` can be reused for the quantization, but `ivf_sq` has its own quantization routines. Is this mainly due to the missing features (different scaling per dim) or also a performance optimization?

Could you update https://github.com/rapidsai/cuvs/issues/106 about this, and highlight next steps that would be needed to have code reuse between `ivf_sq` and `preprocessing::quantize::scalar`?

I am still reviewing search implementation, will get back to  you later today.

viclafargue:
> Hi @viclafargue, the updates in the build / extend API look good to me.
> 
> I have one question of the implementation of `build()`. I would have hoped that `preprocessing::quantize::scalar` can be reused for the quantization, but `ivf_sq` has its own quantization routines. Is this mainly due to the missing features (different scaling per dim) or also a performance optimization?
> 
> Could you update #106 about this, and highlight next steps that would be needed to have code reuse between `ivf_sq` and `preprocessing::quantize::scalar`?
> 
> I am still reviewing search implementation, will get back to you later today.

Hi @tfeher, good point. For build-time SQ training, the main mismatch indeed is that `preprocessing::quantize::scalar` trains one global range, while IVF-SQ needs per-dimension ranges over residuals. The local `fused_column_minmax_kernel` is essentially doing that per-column range computation.

There is also an optimization aspect: IVF-SQ computes min and max in one fused pass and uses vectorized loads where possible.

I’ll update #106 to say the next step for reuse would be adding a per-column scalar quantizer/range-training primitive to preprocessing, then wiring IVF-SQ build to use it.

cpp/src/neighbors/ivf_sq/ivf_sq_build.cuh
viclafargue: Good catch. It is not used anywhere anymore and can be safely removed.

cpp/src/neighbors/ivf_sq/ivf_sq_search.cuh
```diff
@@ -0,0 +1,786 @@
...
+  // negated by finalize_distance and undone by postprocess_distances below),
+  // so all scan-related top-k selections use select_min = true. Only the
+  // coarse search, which runs on the raw GEMM output, follows the metric's
+  // native direction.
+  const bool coarse_select_min = cuvs::distance::is_min_close(index.metric());
+
+  std::size_t n_queries_probes = std::size_t(n_queries) * std::size_t(n_probes);
+
+  bool needs_query_norms = index.metric() != cuvs::distance::DistanceType::InnerProduct;
+  rmm::device_uvector<float> query_norm_dev(needs_query_norms ? n_queries : 0, stream, search_mr);
```
tfeher: Nitpick, but we should prefer `device_mdspan`

cpp/src/neighbors/ivf_sq/detail/jit_lto_kernels/scan_impl.cuh
```diff
@@ -0,0 +1,261 @@
...
+
+template <>
+struct sq_block_sort<0> {
+  using type = ivf::detail::dummy_block_sort_t<float, uint32_t, /*Ascending=*/true>;
+};
+
+template <int Capacity>
+using sq_block_sort_t = typename sq_block_sort<Capacity>::type;
+
+// IVF-SQ scan kernel body with fused in-kernel top-k.
```
tfeher, line 49: Great level of detail in these developer docs, thanks!

tfeher (approved):
Thanks @viclafargue for this PR, it looks great! The search implementation follows the same pattern as other IVF method in cuvs. We should think about unifying the common logic across all of our IVF methods. But that is outside of the scope of the current PR.

LGTM.

PR #1867, opened 2026-03-03
Remove pytest upper bound pin

vyasr:
## Summary

Refs https://github.com/rapidsai/build-planning/issues/230

Removes the pytest `<9` upper bound pin. The pin was added as a workaround for a compatibility issue with pytest 9, which was resolved at the time of pytest 9's release (see https://github.com/rapidsai/cudf/issues/20570#issuecomment-3514160274). Removing the pin allows pytest 9 to be used.

Changes:
- Update `dependencies.yaml`: remove the `<9.0.0` upper bound from `pytest`, leaving it as an unpinned dependency
- Regenerate all derived files via `rapids-dependency-file-generator`

### DIFF
--- a/cpp/src/neighbors/detail/epsilon_neighborhood.cuh
+++ b/cpp/src/neighbors/detail/epsilon_neighborhood.cuh
@@ -1,5 +1,5 @@
 /*
- * SPDX-FileCopyrightText: Copyright (c) 2025, NVIDIA CORPORATION.
+ * SPDX-FileCopyrightText: Copyright (c) 2025-2026, NVIDIA CORPORATION.
  * SPDX-License-Identifier: Apache-2.0
  */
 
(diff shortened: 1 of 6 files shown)


dependencies.yaml
```diff
@@ -382,7 +382,7 @@ dependencies:
         packages:
           - *cmake_ver
           - openblas
-          - &pytest pytest<9.0.0
+          - &pytest pytest
```
KyleFromNVIDIA, line 385: This was previously a reference to keep the other `pytest` below in sync. Since we are no longer pinning to a specific version, we could take the reference out, as it seems kind of redundant now.

jameslamb: I vote keep it. It's helpful for reducing duplication. We want all the `pytest` constraints to be the same within the project, having only 1 where it's spelled out and everything else be references ensures that.

PR #1869, opened 2026-03-03
feat(rust)!: compile-time dataset lifetime safety for all index types

zbennett10:
This is follow-on from https://github.com/rapidsai/cuvs/pull/1839 based on comments from @benfred 

Introduces compile-time dataset lifetime enforcement across all 4 Rust index types (`brute_force`, `cagra`, `ivf_flat`, `ivf_pq`)... assuming we want to do it for all of them! It addresses the use-after-free risk when `ManagedTensor::from(&ndarray)` creates a non-owning view and the C library stores a pointer to the dataset.. and who knows what shenanigans may go on "down there".. 😄  

### Main Changes

- **`DatasetOwnership<'a>` enum** in `dlpack.rs` — tracks whether an index borrows or owns its dataset
- **`Index<'a>` with dual constructors** for all 4 index types:
  - `build(&'a ManagedTensor)` — borrowed path, compiler enforces dataset outlives index
  - `build_owned(ManagedTensor)` — owned path (`'static`), index is self-contained
- **`create_handle()` helper** — avoids double-free by separating FFI handle creation from `Index` struct construction

### Design

```rust
pub struct Index<'a> { // Ties the lifetime of the index to it's underlying data
    inner: ffi::cuvsXxxIndex_t,
    _data: DatasetOwnership<'a>,
}

// Borrowed — compiler enforces arr outlives index
let tensor = ManagedTensor::from(&arr);
let index = Index::build(&res, &params, &tensor)?;

// Owned — 'static, self-contained
let device = ManagedTensor::from(&arr).to_device(&res)?;
let index = Index::build_owned(&res, &params, device)?;
drop(arr); // Fine — index owns the GPU copy
```

## Breaking Change

`Index::build()` now takes `&ManagedTensor` instead of `impl Into<ManagedTensor>`. Users should:
- Use `build(&tensor)` with an explicit `ManagedTensor` reference (borrowed)
- Use `build_owned(tensor)` for the old move semantics (owned)

### DIFF
--- a/rust/cuvs/src/cagra/index.rs
+++ b/rust/cuvs/src/cagra/index.rs
@@ -5,6 +5,7 @@
 
 use std::ffi::CString;
 use std::io::{stderr, Write};
+use std::marker::PhantomData;
 use std::path::Path;
 
 use crate::cagra::{IndexParams, SearchParams};
@@ -12,9 +13,37 @@ use crate::dlpack::ManagedTensor;
 use crate::error::{check_cuvs, Error, Result};
 use crate::resources::Resources;
 
-/// CAGRA ANN Index
+/// CAGRA ANN Index.
+///
+/// CAGRA's behavior is runtime-dependent: the underlying C++ helper
+/// `make_aligned_dataset()` stores a non-owning view of the dataset when the
+/// input is device-accessible, row-major and 16-byte aligned, and copies
+/// otherwise. Because the non-owning path can't be ruled out at compile time,
+/// the returned index carries the lifetime of the `ManagedTensor` used to
+/// build it, so the borrow checker prevents use-after-free regardless of which
+/// path the C++ side takes.
+///
+/// Indices produced by [`Index::deserialize`] own their data and carry a
+/// `'static` lifetime, since the serialized file is fully read into memory
+/// managed by the C++ library.
+///
+/// # Example
+///
+/// ```no_run
+/// # use cuvs::{ManagedTensor, Resources};
+/// # use cuvs::cagra::{Index, IndexParams};
+/// let res = Resources::new().unwrap();
+/// let arr = ndarray::Array::<f32, _>::zeros((256, 16));
+/// let params = IndexParams::new().unwrap();
+/// let tensor = ManagedTensor::from(&arr);
+/// let index = Index::build(&res, &params, &tensor).unwrap();
+/// // `arr` and `tensor` must outlive `index`.
+/// ```
 #[derive(Debug)]
-pub struct Index(ffi::cuvsCagraIndex_t);
+pub struct Index<'a> {
+    inner: ffi::cuvsCagraIndex_t,
+    _dataset: PhantomData<&'a ()>,
+}
 
 /// Convert a filesystem path into a `CString` suitable for the cuVS C API,
 /// returning `Error::InvalidArgument` instead of panicking for paths that are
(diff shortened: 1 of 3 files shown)


zbennett10:
@benfred  This *does* introduce a breaking change.. which I"m not super happy about it.. I mean we could do something like this where we have a backward-compatible approach.. basically keep `build()` as the owned path, add `build_borrowed()` as the new, safe path.. but idk maybe it's best to bite the bullet there.                                                                          
                                      
  ```rust                            
  // Keep the old build() signature — it becomes the "owned" path                                                                                                                               
  impl Index<'static> {                                                                                                                                                                         
      pub fn build<T: Into<ManagedTensor>>(
          res: &Resources,
          params: &IndexParams,
          dataset: T,
      ) -> Result<Index<'static>> {
          let dataset: ManagedTensor = dataset.into();
          let inner = Self::create_handle()?;
          unsafe { check_cuvs(ffi::cuvsCagraBuild(res.0, params.0, dataset.as_ptr(), inner))?; }
          Ok(Index { inner, _data: DatasetOwnership::Owned(dataset) })
      }
  }

  // New safe borrowed path alongside it
  impl<'a> Index<'a> {
      pub fn build_borrowed(
          res: &Resources,
          params: &IndexParams,
          dataset: &'a ManagedTensor,
      ) -> Result<Index<'a>> { ... }
  }
  ```

  Existing code like `Index::build(&res, &params, &dataset)` would still compile — the `ndarray` ref goes through `Into<ManagedTensor>` and gets owned. 

  But — the struct still has `<'a>`, so `fn get_index() -> Index` breaks. So for this to work we would need a type alias:
Something like `pub type OwnedIndex = Index<'static>;`

yan-zaretskiy:
Hey Zachary, thank you for your contribution — the current state of the Rust bindings is indeed unsound for certain index types and this is a very important area of improvement. I've spent some time investigating the C++ internals to understand exactly which index types need lifetime enforcement, and here's what I found:

- `BruteForce` always stores a non-owning view of the dataset
- CAGRA behavior is runtime-dependent. `make_aligned_dataset()` in `common.hpp` stores a `non_owning_dataset` (just an `mdspan` view) when the input is device-accessible, row-major, and 16-byte aligned; and it performs a copy otherwise. Hence at the type level we have no choice but to also enforce a lifetime.
- IVF-Flat & IVF-PQ always copy the dataset, they don't need to reference the original data.

Our Rust bindings should retain the same ownership semantics as the underlying library. For this reason I feel that I'd need to push back on the proposed design. 

Besides, the way it is implemented in this PR is too fragile. The `ManagedTensor` in itself does not indicate at the type level if it actually owns the underlying dataset, even if it has a non-null deleter. So the `Owned` variant of the `DatasetOwnership<'a>` does not express the true ownership.

Instead, I propose to simplify this PR as follows:

1. Revert IVF-Flat and IVF-PQ — these always copy
2. Remove the `build_owned` method altogether.
3. Use `PhantomData<&'a ()>` directly on the index structs and delete the `DatasetOwnership`. For the borrowed path, the caller keeps the tensor alive and the compiler enforces it via `'a`.
4. `deserialize` should return `Index<'static>` — deserialized indices own their data.

zbennett10:
Appreciate you @yan-zaretskiy — thanks a lot for the great feedback

The public API change is narrower now — just `cagra::Index::build` and `brute_force::Index::build` taking `&ManagedTensor` instead of `impl Into<ManagedTensor>`. IVF-Flat and IVF-PQ are completely untouched. Let me know what you think...

rust/cuvs/src/cagra/index.rs
```diff
@@ -4,52 +4,91 @@
...
-            check_cuvs(ffi::cuvsCagraIndexCreate(index.as_mut_ptr()))?;
-            Ok(Index(index.assume_init()))
-        }
+    /// Creates a new empty index.
+    pub fn new() -> Result<Index<'a>> {
+        Ok(Index {
+            inner: Self::create_handle()?,
+            _dataset: PhantomData,
+        })
     }
```
yan-zaretskiy: I'm not sure you need this method on the public API. It's an empty index, so there's no real lifetime associated with it.

yan-zaretskiy:
Looks good to me otherwise!

rust/cuvs/src/cagra/index.rs
zbennett10: Good catch

yan-zaretskiy:
@zbennett10 Could you resolve the branch conflict so that we can merge this PR?

zbennett10:
> @zbennett10 Could you resolve the branch conflict so that we can merge this PR?

@yan-zaretskiy  👍

yan-zaretskiy:
@zbennett10 Your build is broken, I am afraid. I also wanted to point out, that we're recently started an internal refactor of our public Rust API. It takes care of soundness issues, similar to your PR. We will be gradually unrolling it over the next few weeks.

yan-zaretskiy:
Closing this, since the proposed changes are now fully superseded by the `26.08` release.

PR #1876, opened 2026-03-04
Add developer and user guides for JIT

### LINKED ISSUE #1522: Document how to use JIT-LTO in the DEVELOPER_GUIDE

### LINKED ISSUE #1523: Document limitations of JIT-LTO such as warmup time

divyegala:
closes #1522 closes #1523


docs/source/advanced_topics.rst
```diff
@@ -0,0 +1,15 @@
...
+
+Just-in-Time Compilation
+------------------------
+cuVS uses the Just-in-Time (JIT) compilation technology to compile certain kernels. When a JIT compilation is triggered, cuVS will compile the kernel for your architecture and automatically cache it in-memory and on-disk. The validity of the cache is as follows:
+
+1. In-memory cache is valid for the lifetime of the process.
+2. On-disk cache is valid until a CUDA driver upgrade is performed.
+Thus, the JIT compilation is a one-time cost and you can expect no loss in real performance after the first compilation. We recommend that you run a "warmup" to trigger the JIT compilation before the actual usage.
+
+Currently, the following algorithms will trigger a JIT compilation:
```
cjnolet:
```suggestion
Currently, the following capabilities will trigger a JIT compilation:
```

docs/source/jit_lto_guide.md
```diff
@@ -0,0 +1,814 @@
...
+} // namespace example::detail
+```
+
+### Step 3: Create JSON Matrix Files
+
+JSON matrix files define all the parameter combinations that need to be compiled. The build system uses these to generate `.cu` files from `.cu.in` templates.
+
+**How JSON Cross-Product Works**:
+- The build system computes the **Cartesian product** (cross-product) of all parameter combinations
+- **Leaf nodes** are the actual values (strings, numbers, or objects with named properties)
```
KyleFromNVIDIA: Leaf nodes only includes strings, numbers, booleans, and `null`. Dictionaries are always fully searched, even if they only contain leaf nodes and no lists.

docs/source/jit_lto_guide.md
```diff
@@ -0,0 +1,814 @@
...
+- **Parameters with `_` prefix** (e.g., `_data_type`, `_index`) create **groups** that are expanded together
+- Parameters **without `_` prefix** are treated as simple arrays of values
```
KyleFromNVIDIA: It's not the naming convention that determines the behavior but the structure of the document. If a dictionary is nested within a dictionary, the outer dictionary's key to the nested dictionary is not used, and only the keys within the nested dictionary are used.

The `_` convention is a courtesy for the reader, but has no effect on the algorithm, other than to automatically warn if it is not followed.

docs/source/jit_lto_guide.md
```diff
@@ -0,0 +1,814 @@
...
+}
+```
+
+This generates 2 × 2 × 2 = 8 combinations:
+- `{data_type: "float", idx_type: "uint32_t", capacity: "1"}`
+- `{data_type: "float", idx_type: "uint32_t", capacity: "2"}`
+- `{data_type: "float", idx_type: "int64_t", capacity: "1"}`
+- ... and so on
+
+When a group contains nested arrays (like `veclen: ["1", "4"]`), those are also expanded within that group before the cross-product is computed.
```
KyleFromNVIDIA, line 213: It would be nice to see an example of this `veclen` behavior in the examples below.

docs/source/jit_lto_guide.md
```diff
@@ -0,0 +1,814 @@
...
+#include "@embedded_header_file@"
+
+using namespace example::detail;
+
+namespace {
+
+__attribute__((__constructor__)) void register_kernel()
+{
+  registerAlgorithm<tag_@idx_abbrev@>(
+    "@filter_name@_@idx_type@",
```
KyleFromNVIDIA: Please make a note somewhere that the fragment names should contain the parameter names as well as their values for the sake of readability. For example, this should look more like:

```
"@filter_name@_index_@index_type@",
```

docs/source/jit_lto_guide.md
```diff
@@ -0,0 +1,814 @@
...
+### Step 3: Create JSON Matrix Files
+
+JSON matrix files define all the parameter combinations that need to be compiled. The build system uses these to generate `.cu` files from `.cu.in` templates.
+
+**How JSON Cross-Product Works**:
+- The build system computes the **Cartesian product** (cross-product) of all parameter combinations
+- **Leaf nodes** are the actual values (strings, numbers, or objects with named properties)
+- **Parameters with `_` prefix** (e.g., `_data_type`, `_index`) create **groups** that are expanded together
+- Parameters **without `_` prefix** are treated as simple arrays of values
+- Each group expands to create multiple combinations, and all groups are cross-multiplied
```
KyleFromNVIDIA:
```suggestion
- The build system computes a modified **Cartesian product** (cross-product) of all parameter combinations.
- **Leaf nodes** are the actual values. These can be strings, numbers, booleans, or `null`, but only strings should be used, even for numbers, for example ``"1"``.
- Related values can be grouped together in a dictionary consisting of single values. Any dictionary key in such a dictionary's ancestry will not be used in the final product, and should be prefixed with `_` to indicate that it is used only for grouping.
- Keys containing only leaf nodes will be used in the final product, and should not be prefixed with `_`.
- The matrix product algorithm will automatically warn if the proper naming convention (`_` prefix or not) is not followed.
- Each group expands to create multiple combinations, and all groups are cross-multiplied.
```

docs/source/jit_lto_guide.md
KyleFromNVIDIA: See my suggestion at https://github.com/rapidsai/cuvs/pull/1876/changes#r2891012749.

docs/source/jit_lto_guide.md
KyleFromNVIDIA: See my suggestion at https://github.com/rapidsai/cuvs/pull/1876/changes#r2891012749.

docs/source/advanced_topics.rst
```diff
@@ -0,0 +1,21 @@
...
+- `Just-in-Time Compilation`_
+
+Just-in-Time Compilation
+------------------------
+cuVS uses the Just-in-Time (JIT) compilation technology to compile certain kernels. When a JIT compilation is triggered, cuVS will compile the kernel for your architecture and automatically cache it in-memory and on-disk. The validity of the cache is as follows:
+
+1. In-memory cache is valid for the lifetime of the process.
+2. On-disk cache is valid until a CUDA driver upgrade is performed. This is stored in the user's home directory under the path ``~/.nv/ComputeCache/``, and can be portably shared between machines in network or cloud storage.
+
+Thus, the JIT compilation is a one-time cost and you can expect no loss in real performance after the first compilation. We recommend that you run a "warmup" to trigger the JIT compilation before the actual usage.
```
betatim, line 14: Do you want to make it super mega obvious to people who deploy services based on cuvs in containers that "We really really strongly recommend you make sure the cache is stored in a persistent location so that containers don't have to warm up the cache after each restart"

Is it possible to include something that warms up the cache in my `Dockerfile`? So that the cache is built into the image?

I am not sure if I'd make the connection from reading the current docs, hence wondering if a really explicit "hit people over the head with it" call out would be useful.

divyegala: I think that's a great idea. Let me add some phrasing to convey that very clearly.

> Is it possible to include something that warms up the cache in my Dockerfile? So that the cache is built into the image?

You mean automatically?

divyegala: How does it read now?

docs/source/advanced_topics.rst
```diff
@@ -0,0 +1,21 @@
+Advanced Topics
+===============
+
+.. toctree::
+   :maxdepth: 2
+
+   jit_lto_guide
```
cjnolet: Just making sure this renders as expected. Will this create the expected TOC list out of both the jit lto guide and the just-in-time compilation link below?

docs/source/advanced_topics.rst
```diff
@@ -0,0 +1,21 @@
...
+
+- `Just-in-Time Compilation`_
+
+Just-in-Time Compilation
+------------------------
+cuVS uses the Just-in-Time (JIT) compilation technology to compile certain kernels. When a JIT compilation is triggered, cuVS will compile the kernel for your architecture and automatically cache it in-memory and on-disk. The validity of the cache is as follows:
```
cjnolet: >  Just-in-Time (JIT)

Anything we can lnk to here in the "official" cuda docs?

docs/source/advanced_topics.rst
```diff
@@ -0,0 +1,21 @@
...
+
+1. In-memory cache is valid for the lifetime of the process.
+2. On-disk cache is valid until a CUDA driver upgrade is performed. This is stored in the user's home directory under the path ``~/.nv/ComputeCache/``, and can be portably shared between machines in network or cloud storage. We strongly recommend that you store the cache in a persistent location.
```
cjnolet: Is this by default? How does a user change it? Do they have to do a symlink? Can they pass an env var to change it? Maybe link to the part of the cuda docs that dicusses this?

docs/source/advanced_topics.rst
cjnolet: We will definitely need to make sure this is kept up to date

docs/source/developer_guide.md
```diff
@@ -406,3 +406,13 @@ void foo(const raft::resources& res, ...)
     ...
 }
 ```
+
+## Using Just-in-Time Link-Time Optimization
+
+cuVS is moving to using link-time optimization for new kernels, and this requires some changes to the way kernels are written. Instead of compiling all kernel variants at build time (which leads to binary size explosion), JIT LTO compiles kernel fragments separately and links them together at runtime based on the specific configuration needed.
```
cjnolet, line 412: Can we link somewhere in the cuda docs in this paragraph? Maybe for "link time optimation"?

cjnolet: Also, can you provide an ever so brief summary of the perf implications? Maybe link to the cuda docs where appropriate for expectations?

docs/source/developer_guide.md
```diff
@@ -406,3 +406,13 @@ void foo(const raft::resources& res, ...)
...
+
+This approach enables:
```
cjnolet:
```suggestion
This approach ultimately enables:
```

docs/source/jit_lto_guide.md
```diff
@@ -0,0 +1,818 @@
+# JIT LTO (Just-In-Time Link-Time Optimization) Guide
+
+## Background
+
+### What is JIT LTO?
+
+JIT LTO (Just-In-Time Link-Time Optimization) is a CUDA compilation strategy that enables dynamic kernel compilation and linking at runtime. Instead of pre-compiling all possible kernel variants (which would result in an explosion of binary size), JIT LTO compiles kernel **fragments** separately and links them together on-demand when a specific kernel configuration is needed.
```
cjnolet: Can we provide a link somewhere to the cuda docs? This has to be documented somewhere in the cuda docs right?

cjnolet (approved):
Approving so I don't block this, but I think you should address the comments about the links. Those can help users help themselves by reading more in the official documentation, so that they can have a base level knowledge to follow your new uides.

docs/source/developer_guide.md
divyegala: First run perf implications are very kernel and hardware dependent.. the CUDA docs make no guarantees about that.

docs/source/advanced_topics.rst
betatim: Sounds good now. Let's see if people get it, if not can always tune this later.

Wasn't thinking of something automatic, more a command I can include in my `Dockerfile` as a `RUN` command

divyegala: If you see the link that I added now, you can control where the cache is written with an environment variable. I'm hoping docker savvy users can now figure out the volume mount and environment variable connection.

betatim: Works for me

PR #1878, opened 2026-03-04
Implemantion for two-level KMeans Trees in ScaNN

rmaschal:
This PR adds the option for a second level in the kmeans tree for the ScaNN index. The second level is built in a bottom-up fashion, e.g. second-level cluster centers are trained on leaf centers of the bottom level. AVQ and SOAR are also applied to second-level node centers.

One divergence from OSS ScaNN is that the second-level centers are trained on the normalized leaf centers, rather than directly on leaf centers. I've found this gives better recall than training directly on leaf centers, and often gives better recall compared to the OSS algo. 

I've also added a notebook giving example code for producing missing protobuf artifacts for using cuVS built ScaNN indices with OSS ScaNN search functionality.


cpp/src/neighbors/scann/detail/scann_build.cuh
```diff
@@ -36,81 +36,123 @@ using namespace cuvs::spatial::knn::detail;  // NOLINT
...
-    } catch (raft::logic_error& e) {
-      RAFT_LOG_ERROR(
-        "Insufficient device memory for kmeans training set allocation. Please "
-        "reduce kmeans_n_rows_train.");
-      throw;
-    }
-
-    RAFT_LOG_DEBUG("Sampling rows.\n");
+  RAFT_LOG_DEBUG("Sampling rows.\n");
+  if (n_rows_train == dataset.extent(0)) {
```
huuanhhuyn: Shouldn't it be `if (n_rows_train >= dataset.extent(0))` ?

To copy and train on the whole dataset if the dataset size is under the given `n_rows_train `

cpp/src/neighbors/scann/detail/scann_build.cuh
```diff
@@ -36,81 +36,123 @@ using namespace cuvs::spatial::knn::detail;  // NOLINT
...
+    raft::copy(trainset.data_handle(),
+               dataset.data_handle(),
+               dataset.size(),
+               raft::resource::get_cuda_stream(res));
+  } else {
     raft::matrix::sample_rows(res, random_state, dataset, trainset.view());
+  }
 
-    raft::resource::sync_stream(res);
+  raft::resource::sync_stream(res);
```
huuanhhuyn: I guess this is not needed because the later kmeans fitting operates on the same `res` handle, and therefore the same stream as the copying and sampling.

cpp/src/neighbors/scann/detail/scann_build.cuh
```diff
@@ -121,25 +163,120 @@ index<T, IdxT> build(
...
+                                                    coarse_labels);
+
+    compute_soar_labels<T, uint32_t>(res,
+                                     raft::make_const_mdspan(centers_view),
+                                     raft::make_const_mdspan(residuals.view()),
+                                     coarse_centers,
+                                     coarse_labels,
+                                     idx.coarse_soar_labels(),
+                                     params.soar_lambda);
+    raft::resource::sync_stream(res);
```
huuanhhuyn: Please double-check if this is needed.

cpp/src/neighbors/scann/detail/scann_build.cuh
```diff
@@ -121,25 +163,120 @@ index<T, IdxT> build(
...
+                 params.kmeans_n_iters,
+                 centers_view.extent(0),
+                 random_state,
+                 cuvs::cluster::kmeans::kmeans_type::KMeans);
+
+    predict_kmeans(res,
+                   raft::make_const_mdspan(centers_view),
+                   raft::make_const_mdspan(coarse_centers),
+                   coarse_labels,
+                   max_batch_size,
```
huuanhhuyn: I guess we don't want to use the same batch size here as we are training on the leaf centers instead of the whole dataset.

Shouldn't it be something like `std::min<size_t>(centers_view.extent(0), kReasonableMaxBatchSize)`

huuanhhuyn (changes requested):
Thank you for the PR! I would have some minor doubts.

Should we / do we already have an example script or unit test to demonstrate how 2-level kmeans behaves?

cpp/src/neighbors/scann/detail/scann_build.cuh
rmaschal: I simply removed this whole block, and rely on raft::matrix::sample_rows to handle this case.

cpp/src/neighbors/scann/detail/scann_build.cuh
rmaschal: Yes, you are right, allocated buffers are larger than necessary with this logic.

rmaschal:
@huuanhhuyn thanks for the comments. I resolved the issues and added an additional examples for scann, using the the two-level funcionality.

cjnolet (approved):

cpp/src/neighbors/scann/detail/scann_build.cuh
```diff
@@ -36,81 +36,113 @@ using namespace cuvs::spatial::knn::detail;  // NOLINT
...
-
-  RAFT_LOG_DEBUG("Creating empty index");
+  auto trainset = raft::make_device_matrix<T, int64_t>(res, 0, 0);
 
-  index<T, IdxT> idx(res, params, dataset.extent(0), dataset.extent(1));
-  raft::device_matrix_view<T, int64_t> centroids_view = idx.centers();
-  raft::random::RngState random_state{137};
   // sample/project
+  try {
+    trainset = raft::make_device_matrix<T, int64_t>(res, n_rows_train, dataset.extent(1));
```
achirkin: Since this is a large, temporary allocation, we should pass [raft::resource::get_large_workspace_resource_ref](https://github.com/NVIDIA/raft/blob/aa9a6d5083fb445cd1948e7e47053abb7736cb8f/cpp/include/raft/core/resource/device_memory_resource.hpp#L312) to the constructor to use the large workspace resource here.
One caveat is that the user has full control over the large memory resource and may set it up with the managed memory (we actually do that in the benchmarks). In that case, the `try` is much more likely to succeed and result in poor performance when the data doesn't fit the GPU.

cpp/src/neighbors/scann/detail/scann_build.cuh
```diff
@@ -36,81 +36,113 @@ using namespace cuvs::spatial::knn::detail;  // NOLINT
...
+    float inertia  = 0.0;
+    int64_t n_iter = 0;
+
+    cuvs::cluster::kmeans::fit(res,
+                               kmeans_params,
+                               raft::make_const_mdspan(trainset.view()),
+                               std::nullopt,
+                               centers,
+                               raft::make_host_scalar_view<float>(&inertia),
+                               raft::make_host_scalar_view<int64_t>(&n_iter));
```
achirkin: A nitpick: although this is not a bug, perhaps it's better to use raft's provided scalar constructors rather than manually wrapping stack allocations.
```C++
auto inertia = raft::make_host_scalar<float>(res, 0);
auto n_iter  = raft::make_host_scalar<int64_t>(res, 0);
...
cuvs::cluster::kmeans::fit(res, ..., inertia.view(), n_iter.view());
```

cpp/src/neighbors/scann/detail/scann_build.cuh
```diff
@@ -121,25 +153,121 @@ index<T, IdxT> build(
...
+  auto copy_stream      = raft::resource::get_cuda_stream(res);
+  bool enable_prefetch  = false;
+
+  if (res.has_resource_factory(raft::resource::resource_type::CUDA_STREAM_POOL)) {
+    if (raft::resource::get_stream_pool_size(res) >= 1) {
+      enable_prefetch = true;
+
+      copy_stream = raft::resource::get_stream_from_stream_pool(res);
+    }
+  }
```
achirkin, line 210: Raft should do the outer check for you automatically, so you can simplify it:
```suggestion
  if (raft::resource::get_stream_pool_size(res) >= 1) {
    enable_prefetch = true;

    copy_stream = raft::resource::get_stream_from_stream_pool(res);
  }
```
However, if it makes sense for performance, you can also initialize a new resource with the stream if necessary:
```C++
raft::resources tmp_res{res};
if (raft::resource::get_stream_pool_size(tmp_res) < 1) {
  raft::resource::set_cuda_stream_pool(tmp_res, std::make_shared<rmm::cuda_stream_pool>(1));
}
auto copy_stream = raft::resource::get_stream_from_stream_pool(tmp_res);
```
Constructing a new `raft::resources` handle is considered cheap at build/training time. All resources except the new pool resource are shared between `tmp_res` and `res`.

achirkin (commented):
Thanks for the contribution! A few minor comments on the usage of raft primitives below.

cpp/src/neighbors/scann/detail/scann_build.cuh
rmaschal: @achirkin do you have an example? I don't see a way to create a device_matrix using the large workspace resource like you mentioned.

achirkin: @rmaschal sorry for the late reply.
Unfortunately, we don't have `make_device_vector` and `make_device_matrix` that take the memory resource as an argument, but there are plenty examples of `raft::make_device_mdarray`:

cagra graph build: https://github.com/NVIDIA/cuvs/blob/4310a3f17a1084d944551b7d63834014cdaac7c9/cpp/src/neighbors/detail/cagra/graph_core.cuh#L885-L888
IVF-PQ (also decides between normal and large resource):  https://github.com/NVIDIA/cuvs/blob/4310a3f17a1084d944551b7d63834014cdaac7c9/cpp/src/neighbors/ivf_pq/ivf_pq_build.cuh#L1049-L1056

rmaschal: Thanks for the examples. Following IVF-PQ, I changed the code to allocate the trainset against device_memory when it can fit, and the large resource otherwise.

cpp/src/neighbors/scann/detail/scann_build.cuh
rmaschal: I'm not sure raft is indeed performing the outer check, as any tests which don't explicitly allocate a stream pool will fail after removing the outer check. 

There are perf implications for not having a separate copy_stream.  Is creating a copy of `raft::resources` and adding a stream pool the preferred way to create a new stream during `build(..)`? Either way I'd prefer to defer this to a later PR, as this block is simply refactoring existing code to support two-level trees.

huuanhhuyn (approved):
Thank you. LGTM.

msarahan (approved):

rmaschal:
/ok to test [70140cb](https://github.com/NVIDIA/cuvs/pull/1878/commits/70140cb87ae07a8eb704cd2b0b9c25c9584ee548)

Issue #1762, opened 2026-02-02
[BUG] cuVS hierarchical KMeans does not compute inertia (always zero)


btong04:
Hierarchical KMeans does not output a non-zero `inertia` value. 

```
import cuvs.cluster.kmeans as cuvs_kmeans

# Set KMeans parameters:
cuvs_kmeans_params = cuvs_kmeans.KMeansParams(n_clusters=10, max_iter=None, hierarchical=True)

# Train cuVS kmeans model:
centroids, inertia, n_iter = cuvs_kmeans.fit(cuvs_kmeans_params, X)
```

The value for `inertia` is currently always zero. Inertia can be computed separately using the centroids and labels, but is memory intensive if done all at once. 
```
from cuvs.cluster.kmeans import cluster_cost
kmeans_result['inertia'] = cluster_cost(X, centroids)
```

cjnolet:
Thanks for opening an issue about this @btong04. Can't make any promises yet, but we can attempt to fix it for 26.04 release.

PR #1880, opened 2026-03-04
[FEA] Inertia Computation for Balanced KMeans and Add Option for Weighted Inertia

tarang-jain:
Closes https://github.com/rapidsai/cuvs/issues/1762


cpp/src/cluster/kmeans_cluster_cost.cu
```diff
@@ -11,16 +11,36 @@ namespace cuvs::cluster::kmeans {
...
+  cuvs::cluster::kmeans::cluster_cost<float, int64_t>(handle, X, centroids, cost, sample_weight);
+}
+
+void cluster_cost(const raft::resources& handle,
+                  raft::device_matrix_view<const double, int64_t> X,
+                  raft::device_matrix_view<const double, int64_t> centroids,
+                  raft::host_scalar_view<double> cost,
+                  std::optional<raft::device_vector_view<const double, int64_t>> sample_weight)
+{
+  cuvs::cluster::kmeans::cluster_cost<double, int64_t>(handle, X, centroids, cost, sample_weight);
```
jinsolp, line 44: Just want to make sure we aren't adding too much binary by doing this!

jinsolp (approved):
Thanks @tarang-jain LGTM!

cpp/src/cluster/kmeans_cluster_cost.cu
tarang-jain: ~0.6 MB. Seems insignificant. And there is no way around it -- if we want this feature, we'd have to call the detail namespace template with int64_t which would compile it with int64_t any way. I have made sure to call the public API `cluster_cost` so we are not re-compiling.

cpp/include/cuvs/cluster/kmeans.hpp
```diff
@@ -1390,12 +1404,16 @@ void transform(raft::resources const& handle,
...
-void cluster_cost(const raft::resources& handle,
-                  raft::device_matrix_view<const float, int> X,
-                  raft::device_matrix_view<const float, int> centroids,
-                  raft::host_scalar_view<float> cost);
+void cluster_cost(
+  const raft::resources& handle,
+  raft::device_matrix_view<const float, int> X,
+  raft::device_matrix_view<const float, int> centroids,
+  raft::host_scalar_view<float> cost,
+  std::optional<raft::device_vector_view<const float, int>> sample_weight = std::nullopt);
```
lowener, line 1416: Should this new parameter get exposed on the C API?

c/src/cluster/kmeans.cpp
```diff
@@ -63,12 +63,10 @@ void _fit(cuvsResources_t res,
...
-        cuvs::cluster::kmeans::fit(*res_ptr,
-                                   kmeans_params,
-                                   cuvs::core::from_dlpack<const_mdspan_type>(X_tensor),
-                                   cuvs::core::from_dlpack<mdspan_type>(centroids_tensor));
-
-        *inertia = 0;
+        T inertia_temp;
+        auto inertia_view = raft::make_host_scalar_view<T>(&inertia_temp);
+        cuvs::cluster::kmeans::fit(*res_ptr, kmeans_params, cuvs::core::from_dlpack<const_mdspan_type>(X_tensor), cuvs::core::from_dlpack<mdspan_type>(centroids_tensor), std::make_optional(inertia_view));
+        *inertia = inertia_temp;
```
divyegala, line 69: Why is the temp object necessary?

```suggestion
        auto inertia_view = raft::make_host_scalar_view<T>(inertia);
        cuvs::cluster::kmeans::fit(*res_ptr, kmeans_params, cuvs::core::from_dlpack<const_mdspan_type>(X_tensor), cuvs::core::from_dlpack<mdspan_type>(centroids_tensor), std::make_optional(inertia_view));
```

cpp/include/cuvs/cluster/kmeans.hpp
divyegala: It will be an ABI break, no? Do you know how we are supposed to change ABI going forward? 

cc @cjnolet

cpp/src/cluster/detail/kmeans_balanced.cuh
```diff
@@ -1164,6 +1166,18 @@ void build_hierarchical(const raft::resources& handle,
                      MathT{0.2},
                      mapping_op,
                      device_memory);
+
+  // Compute inertia if requested (only supported when T == MathT)
+  if (inertia != nullptr) {
+    if constexpr (std::is_same_v<T, MathT>) {
```
divyegala, line 1171: What happens if the conditional is not satisfied? Should you throw an error or warning that `cluster_cost` is not being computed? Can `MathT` be unequal to `T` in the first place?

cpp/src/cluster/kmeans.cuh
```diff
@@ -440,6 +453,16 @@ void cluster_cost(raft::resources const& handle,
     n_clusters,
     workspace);
 
+  // Apply sample weights if provided
+  if (sample_weight.has_value()) {
+    raft::linalg::map(
+      handle,
+      min_cluster_distance.view(),
+      [] __device__(DataT d, DataT w) { return d * w; },
```
divyegala: `map` here gets instantiated `DataT * IndexT` times due to the outer function being templated on those two types, even though the lambda only relies on `DataT`. Can you pull out the lambda into a functor?

cpp/src/cluster/kmeans_balanced.cuh
```diff
@@ -78,14 +81,18 @@ void fit(const raft::resources& handle,
...
   cuvs::cluster::kmeans::detail::build_hierarchical(handle,
                                                     params,
                                                     X.extent(1),
                                                     X.data_handle(),
                                                     X.extent(0),
                                                     centroids.data_handle(),
                                                     centroids.extent(0),
-                                                    mapping_op);
+                                                    mapping_op,
+                                                    static_cast<const MathT*>(nullptr),
```
divyegala: If always passing `nullptr`, why have this in the API at all?

divyegala (dismissed):

cpp/src/cluster/kmeans_balanced.cuh
lowener: This parameter corresponds to potential pre-computed norms. I added it in the PR for Cosine support in IVF-Flat but ended up not needing it. So it can actually get removed since it is unused at the moment.

cpp/src/cluster/detail/kmeans_balanced.cuh
tarang-jain: yeah so our inertia computation function (cluster_cost) does not work when T and MathT are different (there is no template combination with separate types for these two).

tarang-jain: MathT and T can be unequal when for example MathT is float, but T is uint8_t

divyegala: So you should throw an error or warning right? If someone tries to compute inertia for T uint8, either error or warn them that it won't work.

c/src/cluster/kmeans.cpp
tarang-jain: So that is because inertia is double in the _fit API signature, but the C++ API requires it to be of type T (which can be float). The regular kmeans fit path in this file (not balanced kmeans) also does the same thing.

cpp/src/cluster/detail/kmeans_balanced.cuh
tarang-jain: Added a warning there (in the `else` path).

cpp/src/cluster/kmeans_balanced.cuh
tarang-jain: Ok so I have removed this dataset_norm ptr

cpp/src/cluster/kmeans.cuh
tarang-jain: Changed this to raft::mul_op{}.

cpp/include/cuvs/cluster/kmeans.hpp
tarang-jain: Yeah so I can create a tracker issue, so we can do this change to the C API whenever we decide to make the breaking change.

tarang-jain: I have created as https://github.com/rapidsai/cuvs/issues/1900 to track the C API change. Marking this thread as resolved.

PR #1886, opened 2026-03-06
[FEA] Add Batching to KMeans

tarang-jain:
Merge after #1880

This PR adds support for streaming out of core (dataset on host) kmeans clustering. The idea is simple:

Batched accumulation of centroid updates: Data is processed in batches and batch-wise means and cluster counts are accumulated until all the batches i.e., the full dataset pass has completed.
This PR just brings a batch-size parameter to load and compute cluster assignments and (weighted) centroid adjustments on batches of the dataset. The final centroid 'updates' i.e. a single kmeans iteration only completes when all these accumulated sums are averaged once the whole dataset pass has completed.

### DIFF
--- a/cpp/src/cluster/kmeans_predict_float.cu
+++ b/cpp/src/cluster/kmeans_predict_float.cu
@@ -1,5 +1,5 @@
 /*
- * SPDX-FileCopyrightText: Copyright (c) 2022-2026, NVIDIA CORPORATION.
+ * SPDX-FileCopyrightText: Copyright (c) 2022-2025, NVIDIA CORPORATION.
  * SPDX-License-Identifier: Apache-2.0
  */
 
@@ -52,5 +52,4 @@ void predict(raft::resources const& handle,
   cuvs::cluster::kmeans::predict<float, int64_t>(
     handle, params, X, sample_weight, centroids, labels, normalize_weight, inertia);
 }
-
 }  // namespace cuvs::cluster::kmeans


c/include/cuvs/cluster/kmeans.h
```diff
@@ -90,8 +91,15 @@ struct cuvsKMeansParams {
...
   int batch_centroids;
 
+  /** Check inertia during iterations for early convergence. */
   bool inertia_check;
 
+  /**
+   * Compute final inertia after fit_batched completes (requires extra data pass).
+   * Only used by fit_batched; regular fit always computes final inertia.
+   */
+  bool final_inertia_check;
```
cjnolet: I think having 2 different fit methods doesn't make a whole lot of sense. Feel like there just be a batch size option in the parameters to enable the batched version. I also don't know we would do this, especially if we return intertia.

tarang-jain: So we want to return inertia always? Well we'd still need host APIs for fit with just batch_size as a param (currently only device side fit APIs exist). I just need to rename fit_batched to fit and embed batch_size into the params struct.

c/include/cuvs/cluster/kmeans.h
```diff
@@ -212,6 +220,39 @@ cuvsError_t cuvsKMeansClusterCost(cuvsResources_t res,
...
+ *                              centroids as the initial cluster centers.
+ *                              [out] The generated centroids from the
+ *                              kmeans algorithm are stored at the address
+ *                              pointed by 'centroids'. Must be on DEVICE memory.
+ *                              [dim = n_clusters x n_features]
+ * @param[out]    inertia       Sum of squared distances of samples to their
+ *                              closest cluster center.
+ * @param[out]    n_iter        Number of iterations run.
+ */
+cuvsError_t cuvsKMeansFitBatched(cuvsResources_t res,
```
cjnolet: Yeah as mentioned above, ler's not have a separate functinof or this and just specify it as a parameter. I'd like to minimizde the amount of public APIs users need to worry about calling.

cjnolet: If we are adding to an existing params struct, we should be okay.

tarang-jain: yeah well the intent was to have a combined new API for this and `MiniBatchKmeans`. But for the time being (at least for the scope of this PR) we can get rid of this.

cpp/include/cuvs/cluster/kmeans.hpp
```diff
@@ -139,6 +160,237 @@ enum class kmeans_type { KMeans = 0, KMeansBalanced = 1 };
...
+                     std::optional<raft::host_vector_view<const float, int64_t>> sample_weight,
+                     raft::device_matrix_view<const float, int64_t> centroids,
+                     raft::host_vector_view<int64_t, int64_t> labels,
+                     bool normalize_weight,
+                     raft::host_scalar_view<float> inertia);
+
+/**
+ * @brief Predict cluster labels for host data using batched processing (double).
+ */
+void predict_batched(raft::resources const& handle,
```
cjnolet: We shuldn't need predict-batched becaus eusers acan just do that themselves> The more challenging piece here is traiing by batching the data because it's an inherently iterative algorithm. We should be able to remove these functions altogether.

cpp/include/cuvs/cluster/kmeans.hpp
```diff
@@ -104,9 +106,27 @@ struct params : base_params {
   /**
    * if 0 then batch_centroids = n_clusters
    */
-  int batch_centroids = 0;  //
+  int batch_centroids = 0;
```
jinsolp, line 113: It'd be nice to have some explanation about how `batched_centroids` and `batch_samples` is related (or not) to the `batched_params` below.

c/include/cuvs/cluster/kmeans.h
```diff
@@ -165,7 +179,7 @@ cuvsError_t cuvsKMeansFit(cuvsResources_t res,
                           DLManagedTensor* sample_weight,
                           DLManagedTensor* centroids,
                           double* inertia,
-                          int* n_iter);
+                          int64_t* n_iter);
```
benfred: this is an ABI breaking change - and if its crucial we need to a `cuvsKMeansFit_v2` function until the 26.08 release

cpp/src/cluster/detail/kmeans_batched.cuh
```diff
@@ -0,0 +1,616 @@
...
+
+  std::vector<T> host_sample(n_samples_out * n_features);
+#pragma omp parallel for
+  for (IdxT i = 0; i < static_cast<IdxT>(n_samples_out); i++) {
+    IdxT src_idx = indices[i];
+    std::memcpy(host_sample.data() + i * n_features,
+                X.data_handle() + src_idx * n_features,
+                n_features * sizeof(T));
+    if (copy_weights) { host_weights[i] = weight_in->data_handle()[src_idx]; }
+  }
```
jinsolp: I believe we have a function for host -> device gather with device side indices. You can refer to this: https://github.com/rapidsai/raft/blob/bf258f39b670d65c60d8a681c8ec45d0ade5690f/cpp/include/raft/matrix/detail/sample_rows.cuh#L21-L54

cpp/src/cluster/detail/kmeans_batched.cuh
```diff
@@ -0,0 +1,616 @@
...
+                   (unsigned long long)iter_params.rng_state.seed);
+
+    if (iter_params.init != cuvs::cluster::kmeans::params::InitMethod::Array) {
+      init_centroids_from_host_sample(handle, iter_params, batch_size, X, centroids, workspace);
+    }
+
+    // Reset per-iteration state
+    T priorClusteringCost = 0;
+
+    for (n_iter[0] = 1; n_iter[0] <= iter_params.max_iter; ++n_iter[0]) {
```
jinsolp, line 320: I think updating this in-place is confusing. Can we pull this out? e.g. 
`n_iter_value = n_iter[0]` and use `n_iter_value` in this for loop?

cpp/src/cluster/detail/kmeans_batched.cuh
```diff
@@ -0,0 +1,616 @@
...
+      RAFT_LOG_DEBUG("KMeans batched: Iteration %d", n_iter[0]);
+
+      raft::copy(new_centroids.data_handle(), centroids.data_handle(), centroids.size(), stream);
+
+      T total_cost = 0;
+
+      raft::matrix::fill(handle, centroid_sums.view(), T{0});
+      raft::matrix::fill(handle, cluster_counts.view(), T{0});
+
+      auto centroids_const = raft::make_device_matrix_view<const T, IdxT>(
```
jinsolp: use `raft::make_const_mdspan`

cpp/src/cluster/detail/kmeans_batched.cuh
tarang-jain: We do need to update the value of n_iter. So updating in place makes sense.

cpp/src/cluster/detail/kmeans_batched.cuh
tarang-jain: I have updated this to use the `sample_rows` helper from raft.

c/include/cuvs/cluster/kmeans.h
tarang-jain: Ok I have reverted this.

c/include/cuvs/cluster/kmeans.h
```diff
@@ -90,8 +91,15 @@ struct cuvsKMeansParams {
...
   int batch_centroids;
 
+  /** Check inertia during iterations for early convergence. */
   bool inertia_check;
 
+  /**
+   * Number of samples to process per GPU batch for the batched (host-data) API.
+   * When set to 0, defaults to n_samples (process all at once).
+   */
+  int64_t batch_size;
```
benfred: Can we move this new field to the end of the struct? We are allowed to append new members to structs, but can't change the layout of the existing members

cpp/include/cuvs/cluster/kmeans.hpp
tarang-jain: Added a line for this to the docs.

python/cuvs/cuvs/cluster/kmeans/kmeans.pyx
```diff
@@ -70,6 +71,19 @@ cdef class KMeansParams:
...
+        [batch_samples x n_clusters].
+    batch_centroids : int
+        Number of centroids to process in each batch. If 0, uses n_clusters.
+    inertia_check : bool
+        If True, check inertia during iterations for early convergence.
+    batch_size : int
+        Number of samples to process per GPU batch when fitting with host
+        (numpy) data. When set to 0, defaults to n_samples (process all
+        at once). Only used by the batched (host-data) code path.
+        Default: 0 (auto).
```
jinsolp: can we add more intuitive high-level intuition about this option? e.g. recommend increasing this with host side data to reduce GPU memory pressure

python/cuvs/cuvs/cluster/kmeans/kmeans.pyx
```diff
@@ -202,10 +254,36 @@ def fit(
...
+    >>> centroids, inertia, n_iter = fit(params, X_host)
     """
 
+    # ---- detect host vs device data for data-preparation ----
+    is_host = isinstance(X, np.ndarray) or (
+        hasattr(X, '__array_interface__') and
+        not hasattr(X, '__cuda_array_interface__')
+    )
+
+    if is_host:
```
jinsolp, line 289: can we also add a check to raise a warning if user sets `batch_size` to be larger than the dataset size?

cpp/src/cluster/detail/kmeans_batched.cuh
```diff
@@ -0,0 +1,565 @@
...
+
+      for (const auto& data_batch : data_batches) {
+        IdxT current_batch_size = static_cast<IdxT>(data_batch.size());
+
+        auto batch_data_view = raft::make_device_matrix_view<const T, IdxT>(
+          data_batch.data(), current_batch_size, n_features);
+
+        auto batch_weights_fill_view =
+          raft::make_device_vector_view<T, IdxT>(batch_weights.data_handle(), current_batch_size);
+        if (sample_weight) {
```
jinsolp: `sample_weight.has_value()`

cpp/src/cluster/detail/kmeans_batched.cuh
```diff
@@ -0,0 +1,565 @@
...
+          raft::copy(batch_weights.data_handle(),
+                     sample_weight->data_handle() + data_batch.offset(),
+                     current_batch_size,
+                     stream);
+        } else {
+          raft::matrix::fill(handle, batch_weights_fill_view, T{1});
```
jinsolp: Can we fill the `batch_weights` matrix once outside the for loop instead?

jinsolp (commented):
Thanks @tarang-jain ! Adding some minor comments, but aside from these, LGTM

c/src/cluster/kmeans.cpp
```diff
@@ -49,8 +51,53 @@ void _fit(cuvsResources_t res,
...
   auto res_ptr = reinterpret_cast<raft::resources*>(res);
+  bool is_host = (X.device.device_type == kDLCPU);
 
-  if (cuvs::core::is_dlpack_device_compatible(X)) {
+  if (is_host) {
+    auto n_samples  = static_cast<IdxT>(X.shape[0]);
+    auto n_features = static_cast<IdxT>(X.shape[1]);
+
+    if (params.hierarchical) {
+      RAFT_FAIL("hierarchical kmeans is not supported with host data");
```
lowener, line 61: Can you add an issue for that as it would be a nice feature addition, and link it here as a comment to track.

cpp/include/cuvs/cluster/kmeans.hpp
```diff
@@ -100,15 +100,37 @@ struct params : base_params {
...
+   */
   bool inertia_check = false;
+
+  /**
+   * @struct batched_params
+   * Parameters specific to batched k-means (host-data overloads of fit/predict).
+   * These parameters are only used when calling fit() or predict() with host_matrix_view data
+   * and are ignored by the device_matrix_view overloads.
+   */
+  struct batched_params {
```
lowener: Why is a struct necessary? Are there other parameters that are expected to be added? Or just for the separation?

cpp/include/cuvs/cluster/kmeans.hpp
```diff
@@ -141,6 +163,196 @@ enum class kmeans_type { KMeans = 0, KMeansBalanced = 1 };
  * @{
  */
 
+/**
+ * @defgroup kmeans_batched Batched k-means for out-of-core / host data
```
lowener: These new groups need to be added on doxygen side. Check how they render in the CI artifact files

cpp/src/cluster/detail/kmeans_batched.cuh
tarang-jain: Good catch! I'll do this.

cpp/src/cluster/detail/kmeans_common.cuh
```diff
@@ -469,4 +472,134 @@ void countSamplesInCluster(raft::resources const& handle,
...
+
+  // new_centroids = centroid_sums / cluster_counts (0 when count is 0)
+  raft::copy(
+    new_centroids.data_handle(), centroid_sums.data_handle(), centroid_sums.size(), stream);
+
+  raft::linalg::matrix_vector_op<raft::Apply::ALONG_COLUMNS>(handle,
+                                                             raft::make_const_mdspan(new_centroids),
+                                                             cluster_counts,
+                                                             new_centroids,
+                                                             raft::div_checkzero_op{});
```
lowener: Can you remove the copy and instead in `matrix_vector_op` use `centroid_sums` instead of `new_centroids` as input mdspan?

cpp/src/cluster/detail/kmeans_batched.cuh
```diff
@@ -0,0 +1,585 @@
...
+    }
+
+    T batch_cost;
+    cuvs::cluster::kmeans::cluster_cost(handle,
+                                        batch_view,
+                                        centroids,
+                                        raft::make_host_scalar_view<T>(&batch_cost),
+                                        batch_weight_view);
+
+    total_inertia += batch_cost;
```
lowener: This function will require sync on every iteration. Is there a possibility to keep the cost of each batch on device on each iteration and transfer to host + sync only at the end?

cpp/src/cluster/detail/kmeans_batched.cuh
```diff
@@ -0,0 +1,585 @@
...
+      static_cast<size_t>(n_samples),
+      static_cast<size_t>(n_samples));
+  }
+
+  RAFT_EXPECTS(n_clusters > 0, "n_clusters must be positive");
+  RAFT_EXPECTS(static_cast<IdxT>(centroids.extent(0)) == n_clusters,
+               "centroids.extent(0) must equal n_clusters");
+  RAFT_EXPECTS(centroids.extent(1) == n_features, "centroids.extent(1) must equal n_features");
+
+  raft::default_logger().set_level(params.verbosity);
```
lowener: Remove that before merge

cpp/src/cluster/detail/kmeans_batched.cuh
```diff
@@ -0,0 +1,585 @@
...
+      T total_cost = 0;
+
+      raft::matrix::fill(handle, centroid_sums.view(), T{0});
+      raft::matrix::fill(handle, cluster_counts.view(), T{0});
+
+      auto centroids_const = raft::make_const_mdspan(centroids);
+
+      using namespace cuvs::spatial::knn::detail::utils;
+      batch_load_iterator<T> data_batches(
+        X.data_handle(), n_samples, n_features, batch_size, stream);
```
lowener: Have you considered using pre-fetch? If a stream pool resource is present it would probably be beneficial to use it

python/cuvs/cuvs/tests/test_kmeans.py
```diff
@@ -69,3 +74,56 @@ def test_cluster_cost(n_rows, n_cols, n_clusters, dtype):
     # need reduced tolerance for float32
     tol = 1e-3 if dtype == np.float32 else 1e-6
     assert np.allclose(inertia, sum(cluster_distances), rtol=tol, atol=tol)
+
+
+@pytest.mark.parametrize("n_rows", [1000, 5000])
+@pytest.mark.parametrize("n_cols", [10, 100])
+@pytest.mark.parametrize("n_clusters", [8, 16])
+@pytest.mark.parametrize("batch_size", [100, 500])
```
lowener:
```suggestion
@pytest.mark.parametrize("batch_size", [0, 100, 500])
```
Test default param too

python/cuvs/cuvs/cluster/kmeans/kmeans.pyx
```diff
@@ -202,10 +259,36 @@ def fit(
...
+        if not isinstance(X, np.ndarray):
+            X = np.asarray(X)
+        if not X.flags['C_CONTIGUOUS']:
+            X = np.ascontiguousarray(X)
+        if sample_weights is not None:
+            if not isinstance(sample_weights, np.ndarray):
```
lowener, line 295: Add a check on python side that the residency of sample_weights is the same as X, that would be a common source of user-error

lowener (changes requested):

cpp/src/cluster/detail/kmeans_batched.cuh
tarang-jain: Good catch! I completely removed that function. Instead, we already have the minClusterAndDistance computed. We are just going to reuse it to compute the inertia in the following situations:
1. inertia_check is enabled for early stopping
2. It is the last iteration
And I also removed that host scalar. Instead we accumulate it all into a device scalar and copy to host at the very end.

cpp/include/cuvs/cluster/kmeans.hpp
tarang-jain: I had created a combined struct for other params for MiniBatchKmeans. I can get rid of it for the time being as the MiniBatchKmeans algo is still in the exploratory phase.

cpp/src/cluster/detail/kmeans_common.cuh
tarang-jain: Done!

python/cuvs/cuvs/tests/test_kmeans.py
tarang-jain: Done!

cpp/src/cluster/detail/kmeans_batched.cuh
tarang-jain: Done!

cpp/include/cuvs/cluster/kmeans.hpp
tarang-jain: I have removed those groups. Completely unnecessary (at least for now)

python/cuvs/cuvs/cluster/kmeans/kmeans.pyx
tarang-jain: Done!

c/src/cluster/kmeans.cpp
tarang-jain: https://github.com/rapidsai/cuvs/issues/1916

cpp/include/cuvs/cluster/kmeans.hpp
tarang-jain: I got rid of this struct.

cpp/src/cluster/detail/kmeans_batched.cuh
tarang-jain: For the time being I think we want to avoid prefetch to keep the device memory in check ( my understanding is that prefetch would require device space for two batches).

python/cuvs/cuvs/tests/test_kmeans.py
```diff
@@ -69,3 +74,56 @@ def test_cluster_cost(n_rows, n_cols, n_clusters, dtype):
     # need reduced tolerance for float32
     tol = 1e-3 if dtype == np.float32 else 1e-6
     assert np.allclose(inertia, sum(cluster_distances), rtol=tol, atol=tol)
+
+
+@pytest.mark.parametrize("n_rows", [1000, 5000])
+@pytest.mark.parametrize("n_cols", [10, 100])
+@pytest.mark.parametrize("n_clusters", [8, 16])
+@pytest.mark.parametrize("batch_size", [0, 100, 500])
```
jinsolp: Can we add a test that isn't a divisor of `n_rows`? Super random number like 237 or something maybe?

cpp/src/cluster/detail/kmeans_common.cuh
```diff
@@ -469,4 +472,130 @@ void countSamplesInCluster(raft::resources const& handle,
...
+template <typename DataT, typename IndexT, typename LabelsIterator>
+void compute_centroid_adjustments(
+  raft::resources const& handle,
+  raft::device_matrix_view<const DataT, IndexT, raft::row_major> X,
+  raft::device_vector_view<const DataT, IndexT> sample_weights,
+  LabelsIterator cluster_labels,
+  IndexT n_clusters,
+  raft::device_matrix_view<DataT, IndexT, raft::row_major> centroid_sums,
+  raft::device_vector_view<DataT, IndexT> weight_per_cluster,
+  rmm::device_uvector<char>& workspace)
```
aamijar, line 505: Can we use `raft::device_vector` instead of `rmm::device_uvector`?

cpp/tests/cluster/kmeans.cu
```diff
@@ -346,4 +346,377 @@ TEST_P(KmeansTestF, Result) { ASSERT_TRUE(score == 1.0); }
...
+
+  void SetUp() override { fitBatchedTest(); }
+
+ protected:
+  raft::resources handle;
+  KmeansBatchedInputs<T> testparams;
+  rmm::device_uvector<int> d_labels;
+  rmm::device_uvector<int> d_labels_ref;
+  rmm::device_uvector<T> d_centroids;
+  rmm::device_uvector<T> d_centroids_ref;
```
aamijar, line 536: Same for these.

python/cuvs/cuvs/cluster/kmeans/kmeans.pyx
```diff
@@ -202,10 +259,51 @@ def fit(
...
+                    "host" if is_host else "device",
+                    "host" if is_sample_weight_host else "device"
+                )
+            )
+
+    if is_host:
+        if not isinstance(X, np.ndarray):
+            X = np.asarray(X)
+        if not X.flags['C_CONTIGUOUS']:
+            X = np.ascontiguousarray(X)
```
aamijar: Should we inform the user that their dataset will be copied since it's not in a compatible format in both of these two if cases?

python/cuvs/cuvs/tests/test_kmeans.py
```diff
@@ -69,3 +74,56 @@ def test_cluster_cost(n_rows, n_cols, n_clusters, dtype):
...
+    centroids_batched, _, _ = fit(
+        params_host,
+        X_host,
+        centroids=device_ndarray(initial_centroids_host.copy()),
+    )
+    centroids_batched = centroids_batched.copy_to_host()
+
+    assert np.allclose(
+        centroids_regular, centroids_batched, rtol=1e-3, atol=1e-3
+    ), f"max diff: {np.max(np.abs(centroids_regular - centroids_batched))}"
```
aamijar, line 137: Are we checking ARI score in any of the existing pytests? Or maybe on the cuml side?

cpp/src/cluster/detail/kmeans_batched.cuh
```diff
@@ -0,0 +1,504 @@
...
+                                            weight_per_cluster.view());
+
+        if (params.inertia_check || n_iter[0] == iter_params.max_iter) {
+          // Compute cluster cost for this batch and accumulate
+          cuvs::cluster::kmeans::detail::computeClusterCost(handle,
+                                                            minClusterAndDistance.view(),
+                                                            workspace,
+                                                            clustering_cost.view(),
+                                                            raft::value_op{},
+                                                            raft::add_op{});
```
jinsolp: are the stale values in `minClusterAndDistance.view()` handled properly for the last batch?

cpp/src/cluster/detail/kmeans_batched.cuh
```diff
@@ -0,0 +1,504 @@
...
+    }
+  }
+  if (n_init > 1) {
+    inertia[0] = best_inertia;
+    n_iter[0]  = best_n_iter;
+    RAFT_LOG_DEBUG("KMeans batched: Best of %d runs: inertia=%f, n_iter=%d",
+                   n_init,
+                   static_cast<double>(best_inertia),
+                   best_n_iter);
+  }
```
jinsolp, line 507: don't we have to copy the best centroids to the centroids output here?

achirkin:
Thanks for working on this much-needed feature, @tarang-jain ! Could you please add a short paragraph to the PR description telling _how_ the new batching is implemented? This is extremely helpful not only for reviews, but also for future revisions using tools like `git blame`, because the PR description is copied to the commit message in the main branch history.

c/src/cluster/kmeans.cpp
```diff
@@ -50,7 +52,51 @@ void _fit(cuvsResources_t res,
...
+    auto centroids_view =
+      cuvs::core::from_dlpack<raft::device_matrix_view<T, IdxT, raft::row_major>>(
+        centroids_tensor);
+
+    std::optional<raft::host_vector_view<T const, IdxT>> sample_weight;
+    if (sample_weight_tensor != NULL) {
+      auto sw = sample_weight_tensor->dl_tensor;
+      if (cuvs::core::is_dlpack_host_compatible(sw)) {
+        RAFT_FAIL("sample_weight must be host accessible when X is on host");
+      }
```
viclafargue:
```suggestion
      if (!cuvs::core::is_dlpack_host_compatible(sw)) {
        RAFT_FAIL("sample_weight must be host accessible when X is on host");
      }
```

cpp/src/cluster/detail/kmeans_batched.cuh
```diff
@@ -0,0 +1,525 @@
...
+  auto n_samples      = X.extent(0);
+  auto n_features     = X.extent(1);
+  auto n_clusters     = params.n_clusters;
+  auto metric         = params.metric;
+
+  // Read batch_size from params; default to n_samples if 0 (auto)
+  IdxT batch_size = static_cast<IdxT>(params.batch_size);
+
+  if (params.batch_size <= 0) {
+    RAFT_LOG_WARN("batch_size must be > 0, using n_samples=%zu", static_cast<size_t>(n_samples));
```
viclafargue: Isn't the `batch_size` parameter supposed to be set to 0 by default (to mean to adjust batch size to the number of samples)? It might be confusing for the user to see a warning message.

cpp/src/cluster/detail/kmeans_batched.cuh
```diff
@@ -0,0 +1,525 @@
...
+
+  inertia[0] = 0;
+
+  for (IdxT batch_idx = 0; batch_idx < n_samples; batch_idx += batch_size) {
+    IdxT current_batch_size = std::min(batch_size, n_samples - batch_idx);
+
+    raft::copy(batch_data.data_handle(),
+               X.data_handle() + batch_idx * n_features,
+               current_batch_size * n_features,
+               stream);
```
viclafargue: Why not using the `batch_load_iterator` utility here like in the `fit` function?

cpp/src/cluster/detail/kmeans_batched.cuh
```diff
@@ -0,0 +1,525 @@
...
+
+      using namespace cuvs::spatial::knn::detail::utils;
+      batch_load_iterator<T> data_batches(
+        X.data_handle(), n_samples, n_features, batch_size, stream);
+
+      for (const auto& data_batch : data_batches) {
+        IdxT current_batch_size = static_cast<IdxT>(data_batch.size());
+
+        RAFT_LOG_INFO(
+          "KMeans batched: Processing batch %zu of %zu", data_batch.offset(), n_samples);
```
viclafargue: Printing an information message for every batch may be excessive. Shouldn't we print it as debug message?

cpp/tests/cluster/kmeans.cu
```diff
@@ -346,4 +346,383 @@ TEST_P(KmeansTestF, Result) { ASSERT_TRUE(score == 1.0); }
...
+
+    if (score < 1.0) {
+      //   std::stringstream ss;
+      //   ss << "Expected: " << raft::arr2Str(d_labels_ref.data(), 25, "d_labels_ref", stream);
+      //   std::cout << (ss.str().c_str()) << '\n';
+      //   ss.str(std::string());
+      //   ss << "Actual: " << raft::arr2Str(d_labels.data(), 25, "d_labels", stream);
+      //   std::cout << (ss.str().c_str()) << '\n';
+      //   std::cout << "Score = " << score << '\n';
+    }
```
viclafargue: What is this code for?

cpp/src/cluster/detail/kmeans_batched.cuh
```diff
@@ -0,0 +1,525 @@
...
+
+        if (params.inertia_check || n_iter[0] == iter_params.max_iter) {
+          // Compute cluster cost for this batch and accumulate
+          cuvs::cluster::kmeans::detail::computeClusterCost(
+            handle,
+            minClusterAndDistance_view,
+            workspace,
+            raft::make_device_scalar_view(batch_clustering_cost.data_handle()),
+            raft::value_op{},
+            raft::add_op{});
```
viclafargue: It looks like the computation of the inertia does not account for the sample weights.

cpp/src/cluster/detail/kmeans_batched.cuh
```diff
@@ -0,0 +1,525 @@
...
+  auto best_centroids = n_init > 1
+                          ? raft::make_device_matrix<T, IdxT>(handle, n_clusters, n_features)
+                          : raft::make_device_matrix<T, IdxT>(handle, 0, 0);
+  T best_inertia      = std::numeric_limits<T>::max();
+  IdxT best_n_iter    = 0;
+
+  std::mt19937 gen(params.rng_state.seed);
+
+  // ----- Allocate reusable work buffers (shared across n_init iterations) -----
+  auto batch_data    = raft::make_device_matrix<T, IdxT>(handle, batch_size, n_features);
```
viclafargue: It looks like this device matrix is never used.

cpp/src/cluster/detail/kmeans_batched.cuh
```diff
@@ -0,0 +1,525 @@
...
+
+  // Warn if user explicitly set batch_size larger than dataset size
+  if (params.batch_size > 0 && static_cast<IdxT>(params.batch_size) > n_samples) {
+    RAFT_LOG_WARN(
+      "batch_size (%zu) is larger than dataset size (%zu). "
+      "batch_size will be effectively clamped to %zu.",
+      static_cast<size_t>(params.batch_size),
+      static_cast<size_t>(n_samples),
+      static_cast<size_t>(n_samples));
+  }
```
viclafargue: Please clip the batch size, otherwise we may allocate unnecessarily large buffers.

```suggestion
  if (params.batch_size > 0 && static_cast<IdxT>(params.batch_size) > n_samples) {
    RAFT_LOG_WARN(
      "batch_size (%zu) is larger than dataset size (%zu). "
      "batch_size will be effectively clamped to %zu.",
      static_cast<size_t>(params.batch_size),
      static_cast<size_t>(n_samples),
      static_cast<size_t>(n_samples));
      batch_size = n_samples;
  }
```

cpp/src/cluster/detail/kmeans_batched.cuh
```diff
@@ -0,0 +1,525 @@
...
+{
+  cudaStream_t stream = raft::resource::get_cuda_stream(handle);
+  auto n_features     = batch_data.extent(1);
+  auto n_clusters     = centroid_sums.extent(0);
+
+  auto workspace = rmm::device_uvector<char>(
+    batch_data.extent(0), stream, raft::resource::get_workspace_resource(handle));
+
+  auto batch_sums   = raft::make_device_matrix<MathT, IdxT>(handle, n_clusters, n_features);
+  auto batch_counts = raft::make_device_vector<MathT, IdxT>(handle, n_clusters);
```
viclafargue: Since the sizes of these allocations are predictable, we could maybe pre-allocate the buffers in the `fit` function? This is just a small optimization, but it may add complexity to the code for minimal gain.

cpp/src/cluster/detail/kmeans_batched.cuh
```diff
@@ -0,0 +1,525 @@
...
+      raft::matrix::fill(handle, weight_per_cluster.view(), T{0});
+
+      raft::matrix::fill(handle, clustering_cost.view(), T{0});
+      raft::matrix::fill(handle, batch_clustering_cost.view(), T{0});
+
+      auto centroids_const = raft::make_const_mdspan(centroids);
+
+      using namespace cuvs::spatial::knn::detail::utils;
+      batch_load_iterator<T> data_batches(
+        X.data_handle(), n_samples, n_features, batch_size, stream);
```
viclafargue: The batch iterator could maybe be instantiated once outside of the loop and re-used?

cpp/src/cluster/detail/kmeans_batched.cuh
```diff
@@ -0,0 +1,525 @@
...
+
+      if (sqrdNormError < params.tol) done = true;
+
+      if (done || n_iter[0] == iter_params.max_iter) {
+        RAFT_LOG_DEBUG("KMeans batched: Converged after %d iterations", n_iter[0]);
+        // Inertia for the last iteration is always computed
+        if (!params.inertia_check) {
+          raft::copy(inertia.data_handle(), clustering_cost.data_handle(), 1, stream);
+          raft::resource::sync_stream(handle);
+        }
```
viclafargue: I think there is a bug here. If `inertia_check = false` and the convergence happen before the max iteration limit, `clustering_cost` will have its default value of 0.0 as its value is never updated. This may also break the `n_init > 1` case, as the `inertia` may be falsely set to 0.0.

tarang-jain:
@achirkin I updated the PR desc.

cpp/src/cluster/detail/kmeans_batched.cuh
tarang-jain: yeah I removed all of that. I was trying to debug / test something.

cpp/src/cluster/detail/kmeans_batched.cuh
tarang-jain: Right! I added a final inertia computation after the iterations. So if the convergence happened, the inertia is computed on the entire dataset with the NEW centroids. This step is also present in regular (on-device) kmeans.

cpp/src/cluster/detail/kmeans_batched.cuh
tarang-jain: Done!

cpp/src/cluster/detail/kmeans_batched.cuh
tarang-jain: Good catch. I have added these buffers into the main fit function and they are created before the iterations.

cpp/src/cluster/detail/kmeans_batched.cuh
tarang-jain: Removed!

cpp/tests/cluster/kmeans.cu
tarang-jain: Its a debug print. It is also present in the other kmeans and kmeans_balanced tests

python/cuvs/cuvs/cluster/kmeans/kmeans.pyx
tarang-jain: yeah this is a bit problematic. We dont want to create a copy of hundreds of GB of data. I'll just throw an error message instead if it is not C contiguous.

python/cuvs/cuvs/cluster/kmeans/kmeans.pyx
tarang-jain: we are doing that on the C++ side.

python/cuvs/cuvs/tests/test_kmeans.py
tarang-jain: We are checking the centroids for an exact match and the predict function is already checked. So ARI need not be checked.

python/cuvs/cuvs/tests/test_kmeans.py
tarang-jain: Done

cpp/src/cluster/detail/kmeans_batched.cuh
tarang-jain: yeah I took care of it. Let me know if you still notice something unusual.

cpp/src/cluster/detail/kmeans_batched.cuh
tarang-jain: Done!

cpp/src/cluster/detail/kmeans_batched.cuh
tarang-jain: Done!

cpp/src/cluster/detail/kmeans_batched.cuh
tarang-jain: Done! I have fixed the warning's conditions.

cpp/src/cluster/detail/kmeans_batched.cuh
tarang-jain: That is true. It is also the case for regular kmeans (not batched). I am tracking it in https://github.com/rapidsai/cuvs/issues/1940 and will fix it for both in another follow up PR.

tarang-jain:
I have renamed batch_size to streaming_batch_size to make its purpose clearer and differentiate it from the batch_samples and batch_centroids parameters.

cpp/src/cluster/detail/kmeans_batched.cuh
```diff
@@ -0,0 +1,634 @@
...
```
viclafargue:
```suggestion
        // Inertia for the last iteration is always computed after KMeans training
```

Do we truly need this copy here? This is probably dead code since we are computing the inertia unconditionally at the end.

python/cuvs/cuvs/tests/test_kmeans.py
```diff
@@ -69,3 +74,69 @@ def test_cluster_cost(n_rows, n_cols, n_clusters, dtype):
...
+        centroids=device_ndarray(initial_centroids_host.copy()),
+        sample_weights=sample_weights_host,
+    )
+    centroids_batched = centroids_batched.copy_to_host()
+
+    assert np.allclose(
+        centroids_regular, centroids_batched, rtol=1e-3, atol=1e-3
+    ), f"max diff: {np.max(np.abs(centroids_regular - centroids_batched))}"
+
+    print(inertia_regular, inertia_batched)
```
viclafargue: Do we want to keep the print here?

cpp/src/cluster/detail/kmeans_batched.cuh
```diff
@@ -0,0 +1,634 @@
...
+  raft::device_vector_view<const MathT, IdxT> sample_weights,
+  raft::device_matrix_view<MathT, IdxT> centroid_sums,
+  raft::device_vector_view<MathT, IdxT> cluster_counts,
+  raft::device_matrix_view<MathT, IdxT> batch_sums,
+  raft::device_vector_view<MathT, IdxT> batch_counts)
+{
+  cudaStream_t stream = raft::resource::get_cuda_stream(handle);
+
+  auto workspace = rmm::device_uvector<char>(
+    batch_data.extent(0), stream, raft::resource::get_workspace_resource(handle));
```
viclafargue, line 176: This is called once per batch. We could maybe move this outside of the loop in the caller function?

cpp/include/cuvs/cluster/kmeans.hpp
```diff
@@ -141,6 +157,178 @@ enum class kmeans_type { KMeans = 0, KMeansBalanced = 1 };
...
+ *                              closest cluster center.
+ * @param[out]    n_iter        Number of iterations run.
+ */
+void fit(raft::resources const& handle,
+         const cuvs::cluster::kmeans::params& params,
+         raft::host_matrix_view<const float, int> X,
+         std::optional<raft::host_vector_view<const float, int>> sample_weight,
+         raft::device_matrix_view<float, int> centroids,
+         raft::host_scalar_view<float> inertia,
+         raft::host_scalar_view<int> n_iter);
```
viclafargue: It looks like the host `fit` function are provided for both `<T, int>` and `<T, int64_t>`, but the `predict` and `fit_predict` are not (only `int64_t`). Is this something that is expected? Shouldn't we only expose `int64_t` even for device functions? Are there performance implications?

cpp/src/cluster/detail/kmeans_common.cuh
```diff
@@ -469,4 +472,130 @@ void countSamplesInCluster(raft::resources const& handle,
...
+{
+  cudaStream_t stream = raft::resource::get_cuda_stream(handle);
+  auto n_samples      = X.extent(0);
+
+  workspace.resize(n_samples, stream);
+
+  raft::linalg::reduce_rows_by_key(const_cast<DataT*>(X.data_handle()),
```
viclafargue:
```suggestion
  raft::linalg::reduce_rows_by_key(X.data_handle(),
```

The casting is unnecessary here as the first argument is const anyway.

cpp/include/cuvs/cluster/kmeans.hpp
```diff
@@ -141,6 +157,178 @@ enum class kmeans_type { KMeans = 0, KMeansBalanced = 1 };
...
+void fit(raft::resources const& handle,
+         const cuvs::cluster::kmeans::params& params,
+         raft::host_matrix_view<const double, int64_t> X,
+         std::optional<raft::host_vector_view<const double, int64_t>> sample_weight,
+         raft::device_matrix_view<double, int64_t> centroids,
+         raft::host_scalar_view<double> inertia,
+         raft::host_scalar_view<int64_t> n_iter);
+
+/**
+ * @defgroup predict_host K-Means Predict (host data)
```
lowener: Either close this group + add it to doxygen or remove this group. Right now it's breaking the documentation and no `predict` or `fit_predict` function are displayed. Have a look at the generated docs in the build artifact to check it out.

cpp/src/cluster/detail/kmeans_batched.cuh
tarang-jain: yes but its different for the last batch if n_samples is not divisible by the batch size. I did not want to allocate a larger workspace than is needed for each batch.

cpp/src/cluster/detail/kmeans_batched.cuh
tarang-jain: Yeah I have cleaned it up.

cpp/include/cuvs/cluster/kmeans.hpp
tarang-jain: I have removed the int ones for host.

cpp/include/cuvs/cluster/kmeans.hpp
```diff
@@ -141,6 +157,156 @@ enum class kmeans_type { KMeans = 0, KMeansBalanced = 1 };
...
+ * @param[in]     handle        The raft handle.
+ * @param[in]     params        Parameters for KMeans model.
+ * @param[in]     X             Training instances on HOST memory. [dim = n_samples x n_features]
+ * @param[in]     sample_weight Optional weights for each observation (on host).
+ * @param[inout]  centroids     Cluster centers on device. [dim = n_clusters x n_features]
+ * @param[out]    labels        Predicted cluster labels on HOST memory. [dim = n_samples]
+ * @param[out]    inertia       Sum of squared distances to nearest centroid.
+ * @param[out]    n_iter        Number of iterations run.
+ */
+void fit_predict(raft::resources const& handle,
```
lowener: Don't add those 2 fit_predict() functions, there is an upcoming PR https://github.com/rapidsai/cuvs/pull/1939#discussion_r2991440236 to remove fit_predict()

cpp/include/cuvs/cluster/kmeans.hpp
tarang-jain: Removed the predict and fit_predict functions altogether.

cpp/include/cuvs/cluster/kmeans.hpp
tarang-jain: Update: The host side API is changed to <T, int> in order to make it consistent with the other APIs. I created a tracker issue https://github.com/rapidsai/cuvs/issues/1961 to update it to `int64_t` later. That would be part of a broader plan for using int64_t extents.

tarang-jain:
Taking into consideration these comments: https://github.com/rapidsai/cuvs/pull/1886#discussion_r2998146882 and https://github.com/rapidsai/cuvs/pull/1886#discussion_r2893129394 I have removed the host side predict function and put them into #1962 with DO NOT MERGE status to track those changes.

python/cuvs/cuvs/tests/test_kmeans.py
```diff
@@ -69,3 +74,68 @@ def test_cluster_cost(n_rows, n_cols, n_clusters, dtype):
...
+    )
+    centroids_batched = centroids_batched.copy_to_host()
+
+    assert np.allclose(
+        centroids_regular, centroids_batched, rtol=1e-3, atol=1e-3
+    ), f"max diff: {np.max(np.abs(centroids_regular - centroids_batched))}"
+
+    assert np.allclose(
+        inertia_regular, inertia_batched, rtol=1e-3, atol=1e-3
+    ), f"max diff: {np.max(np.abs(inertia_regular - inertia_batched))}"
```
tarang-jain, line 141: Added this inertia check to the python side to evaluate that the inertia calculation is correct in the batched fit function.

cpp/src/cluster/detail/kmeans_batched.cuh
```diff
@@ -0,0 +1,607 @@
...
+template <typename T, typename IdxT>
+void predict(raft::resources const& handle,
+             const cuvs::cluster::kmeans::params& params,
+             raft::host_matrix_view<const T, IdxT> X,
+             std::optional<raft::host_vector_view<const T, IdxT>> sample_weight,
+             raft::device_matrix_view<const T, IdxT> centroids,
+             raft::host_vector_view<IdxT, IdxT> labels,
+             bool normalize_weight,
+             raft::host_scalar_view<T> inertia)
+{
```
tarang-jain: I have still left this function here in the detail namespace because I think there is some utility to it. Its not being used anywhere and tests and public API have been moved to #1962

cpp/include/cuvs/cluster/kmeans.hpp
tarang-jain: Another update on this: I had to revert this back to int64_t as the type for the extents because otherwise I was running into integer overflow problems with int32_t for datasets that wont fit on the device memory.

cpp/src/cluster/kmeans_predict_double.cu
```diff
@@ -1,8 +1,9 @@
 /*
- * SPDX-FileCopyrightText: Copyright (c) 2024-2025, NVIDIA CORPORATION.
+ * SPDX-FileCopyrightText: Copyright (c) 2024-2026, NVIDIA CORPORATION.
  * SPDX-License-Identifier: Apache-2.0
  */
 
+#include "detail/kmeans_batched.cuh"
```
lowener: Unused header, remove from fit_predict and predict instantiation file

tarang-jain: Done!

cpp/src/cluster/detail/kmeans_batched.cuh
tarang-jain: After a discussion with @lowener this function has ben removed. It lives in #1962 atm

benfred (approved):

PR #1889, opened 2026-03-06
Add compute-sanitizer CI workflows for memcheck, racecheck, and synccheck

bdice:
## Summary

- Adds GitHub Actions workflows to run `compute-sanitizer` with memcheck, racecheck, and synccheck on all libcuvs C++ test executables
- Adds `cuda-sanitizer-api` to test dependencies
- PR jobs temporarily replaced with sanitizer-only jobs for initial testing (to be reverted before merge)

## Details

**New files:**
- `ci/discover_libcuvs_tests.sh` — discovers `*_TEST` executables from the conda package, excludes multi-GPU tests, outputs JSON array
- `ci/run_compute_sanitizer_test.sh` — runs `compute-sanitizer` on a single test with `--force-blocking-launches`, kernel name exclusion opt-out (`_no_sanitize`, `_no_<tool>`), and `--track-stream-ordered-races`
- `.github/workflows/compute-sanitizer-run.yaml` — reusable workflow creating one matrix job per test per tool
- `.github/workflows/compute-sanitizer-trigger.yaml` — weekly (Saturday 10:00 UTC) + manual dispatch entry point for all three tools

**Modified files:**
- `dependencies.yaml` — added `cuda-sanitizer-api` to `test_libcuvs`
- `.github/workflows/pr.yaml` — temporarily runs only sanitizer jobs (other jobs commented out, to be restored before merge)

Modeled after https://github.com/rapidsai/cudf/pull/20542.

### DIFF
--- a/.github/workflows/compute-sanitizer-trigger.yaml
+++ b/.github/workflows/compute-sanitizer-trigger.yaml
@@ -15,7 +15,9 @@ jobs:
     container:
       image: rapidsai/ci-conda:26.04-latest
     outputs:
-      tests: ${{ steps.find-tests.outputs.tests }}
+      memcheck_tests: ${{ steps.find-tests.outputs.memcheck_tests }}
+      racecheck_tests: ${{ steps.find-tests.outputs.racecheck_tests }}
+      synccheck_tests: ${{ steps.find-tests.outputs.synccheck_tests }}
     steps:
       - name: Checkout code
         uses: actions/checkout@v5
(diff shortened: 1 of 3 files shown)


aamijar:
Thanks @bdice, do you think we should also be running compute-sanitizer on the pytests?

bdice:
> Thanks @bdice, do you think we should also be running compute-sanitizer on the pytests?

Probably not necessary. The C++ tests should cover it. I think there may be more false positives in the Python side anyway.

Issue #1875, opened 2026-03-04
[BUG] IVF-PQ support of IP distance assumes vectors have been normalized


cjnolet:
IVFPQ w/ IP distance assumes vectors have been normalized, but not all users would expect or think to do this. Essentially, kmeans works with L2 distance or with IP and l2 normalized vectors. Currently if IP is used with vectors that aren't normalized then it runs into degenerate cases where most vectors are placed in a single cluster, and that causes issues for downstream algorithms (like the IVF portions). 
 

One way to deal with this is to do normalization in-place or to normalize the input dataset then return it to its original state after the clustering is done.

tarang-jain:
If we do the normalization, it essentially becomes the cosine metric, which is already an available metric, right? I think the fact that with inner product, the vectors end up in the same cluster is the nature of the metric itself and possibly the incorrect metric choice, rather than a bug in the implementation.

cjnolet:
@tarang-jain we are being asked to support inner product, but the assumption for kmeans is that the data is normalized. We need to normalize the data only in kmeans to avoid these degenerate cases. 

That's the ask, and the solution here. If that means using the cosine solution just for the normalization in kmeans then that's fine too.

PR #1892, opened 2026-03-07
[WIP] IP Vector Normalization to avoid all vectors clumped into single cluster in IVF-PQ 

HowardHuang1:
Addresses #1875.

Current Implementation: Have Inner Product code path match Cosine code path.
Within kmeans, the E-step of cluster assignment will be operating on normalized vectors but the M-step recomputes centers based on raw vectors. Search still happens on raw vectors.

In each EM iteration, before predict() we L2 row-normalize the centroid matrix (cluster_centroids) for IP. Within the E-step's call to predict(), we compute the norm on all vectors in the dataset and dataset_norm is passed into predict_core() and used in fusedDistanceNNMinReduce<> kernel. This is a norm to norm comparison between normalized vectors and normalized centroids.

Then in the M-Step, calc_centers_and_sizes() recalculates the centroids based on the raw dataset and rewrites raw centroids into cluster_centers. 

For the code review, the bulk of the changes are in this file: [‎cpp/src/cluster/detail/kmeans_balanced.cuh](https://github.com/rapidsai/cuvs/pull/1892/changes#diff-8897f2ac3c256338c72604715c11a277d9f6b846a07d5942cf0a085d9e8f4958)
The rest have to do with linking and can be ignored. Those files will be removed prior to merging.

So far, we've attempted:

1. Normalization of whole IVF-PQ path by normalizing input dataset itself
2. Normalization of kmeans only by making a copy (2x memory spike) and recovering raw vectors for search
3. [**Current Implementation ] Setting IP to same code path as Cosine for normalization. 
4. [**In progress ] Normalization within kernels themselves.

Pareto plots for approach #3 (working):
IVF-PQ only:
<table>
<td>image</td>
<td>image</td>
</table>

<table>
<td>image</td>
<td>image</td>
</table>
 
Cagra end-to-end (Showing no harm in search performance):
Showing Fix for Oracle's Issue (in Cagra):
<table>
<td>Screenshot 2026-04-01 at 7 04 01 PM</td>
<td>Screenshot 2026-04-01 at 3 30 59 PM</td>
</table>

<table>
<td>image</td>
<td>image</td>
</table>


aamijar:
Hi @HowardHuang1, are we able to document an example in the description of the PR where vectors were clumped in a single cluster, and then verify that the fix (this PR) solves that case?

Also you can run `pre-commit` to fix the CI style check.

HowardHuang1:
Yes will do. I'll upload a before and after example that illustrates the fixed cluster distributions and timing improvements for search.

aamijar:
Hi @HowardHuang1, is this targeting release/26.04? If so, please retarget from main to release/26.04

jinsolp:
Hi @HowardHuang1 you'll have to rebase on `release/26.04` if you want to target that branch. 🙂

cpp/src/neighbors/ivf_pq/ivf_pq_build.cuh
```diff
@@ -1324,30 +1324,61 @@ auto build(raft::resources const& handle,
...
+      raft::linalg::row_normalize<raft::linalg::L2Norm>(
+        handle, raft::make_const_mdspan(trainset_kmeans_view), trainset_kmeans_view);
+      auto trainset_kmeans_const_view = raft::make_const_mdspan(trainset_kmeans.view());
+      cuvs::cluster::kmeans::fit(handle, kmeans_params, trainset_kmeans_const_view, centers_view);
       raft::linalg::row_normalize<raft::linalg::L2Norm>(handle, centers_const_view, centers_view);
+      cuvs::cluster::kmeans::predict(
+        handle, kmeans_params, trainset_kmeans_const_view, centers_const_view, labels_view);
+      // Recompute centers in original space (mean of unnormalized trainset per cluster), overwrites centers_view
+      rmm::device_uvector<uint32_t> cluster_sizes(impl->n_lists(), stream, device_memory);
+      cuvs::cluster::kmeans::detail::calc_centers_and_sizes<float, float, internal_extents_t, uint32_t, uint32_t>(
```
jinsolp: can we use the `calc_centers_and_sizes` in the public namespace if possible?

jinsolp (commented):
Thanks @HowardHuang1 ! I have a few questions as I try to understand the fix in this code.
1. So we normalize the data, and train kmeans on this normalized data. However, instead of saving the resulting centroids (which come from the normalized space), we re-compute the proper centroids on the raw vectors (what `calc_centers_and_sizes` do)?
2. what are the changes in the cuvs bench for? I'm not so familiar with this part of the code so this is out of curiosity.
3. If I'm reading things right in the screenshot, we get a lower recall with the normalized vectors compared to using the raw vectors? And the search also takes longer?

cpp/src/neighbors/ivf_pq/ivf_pq_build.cuh
HowardHuang1: Hey @jinsolp !
1.  Yes exactly. If we were to keep the normalized centroids that would cause the computation to no longer be Inner Product and instead it will degenerate to Cosine which is not what we want. To prevent this, we use `calc_centers_and_sizes` to re-compute the proper centroids on the raw vectors. In essence, we only want normalization to happen in the kmeans cluster assignment step rather than the whole pipeline (normalization of the whole pipeline changes metric to Cosine). 
2. The changes in cuvs bench are to resolve a linker issue. Will remove these when merging. They can be ignored for now.
3. The screenshot is a bit outdated, will replace soon with a new one. Recall should be same search should be faster.

cpp/src/neighbors/ivf_pq/ivf_pq_build.cuh
```diff
@@ -1324,30 +1324,61 @@ auto build(raft::resources const& handle,
...
+    auto labels_view =
+      raft::make_device_vector_view<uint32_t, internal_extents_t>(labels.data(), n_rows_train);
     auto centers_const_view = raft::make_device_matrix_view<const float, internal_extents_t>(
       cluster_centers, impl->n_lists(), impl->dim());
-    if (impl->metric() == distance::DistanceType::CosineExpanded) {
+
+    if (impl->metric() == distance::DistanceType::InnerProduct) {
+      // Normalization only for k-means: use a copy so trainset stays in original space; metric
+      // remains inner product for the rest of the pipeline.
+      auto trainset_kmeans = raft::make_device_mdarray<float>(
```
cjnolet: Wait- we seem to be doing a copy of the trainset? Why? Just do the normalization in the distance kernels.

cjnolet: This is not the way we do things- we don't do unecessary copies just to avoid changing kernels. This becomes frustrating for users, because they are almost always memory limited and every GB counts.

HowardHuang1: Fixed.

cjnolet:
Closing in favor of #2076

PR #1896, opened 2026-03-09
Make some minor fixes to JIT+LTO functionality

KyleFromNVIDIA:
* Make it so that if a matrix file changes, CMake re-configures when attempting to rebuild. This puts matrix files on par with CMake files in triggering a re-configure.
* Don't build JIT+LTO kernels with debug symbols, as nvcc does not currently support LTO and debug symbols simultaneously.


KyleFromNVIDIA:
These fixes were originally introduced in https://github.com/rapidsai/cuvs/pull/1807, but are now being moved into their own PR.

PR #1898, opened 2026-03-09
[REVIEW] Add L1 support to NN-Descent

yan-zaretskiy:
Use the SIMT local-join path for L1 distance and keep the existing WMMA path for dot-product metrics. This preserves existing behavior while extending NN-Descent to support L1.


cpp/src/neighbors/detail/nn_descent_gnnd.hpp
```diff
@@ -216,7 +216,7 @@ class GNND {
                          int2* list_sizes,
                          cudaStream_t stream = 0);
 
-  template <typename DistEpilogue_t = raft::identity_op>
+  template <typename DistEpilogue_t>
```
jinsolp, line 219: why remove the default `raft::identity_op` here?

cpp/src/neighbors/detail/nn_descent.cuh
```diff
@@ -635,48 +668,47 @@ __launch_bounds__(BLOCK_SIZE)
   int lane_id             = threadIdx.x % raft::warp_size();
   constexpr int num_warps = BLOCK_SIZE / raft::warp_size();
 
-  if (metric != cuvs::distance::DistanceType::BitwiseHamming) {
-    int tid = threadIdx.x;
-    for (int i = tid; i < MAX_NUM_BI_SAMPLES * SKEWED_MAX_NUM_BI_SAMPLES; i += blockDim.x)
-      s_distances[i] = 0.0f;
+  int tid = threadIdx.x;
```
jinsolp, line 676: why remove the `if (metric != cuvs::distance::DistanceType::BitwiseHamming)` check? We don't need to compute the dot product for bitwise hamming distance.

cpp/src/neighbors/detail/nn_descent.cuh
jinsolp, line 729: can we skip this `caltulate_metric` call for l1 distance since we don't do anything?

cpp/src/neighbors/detail/nn_descent.cuh
yan-zaretskiy: The `BitwiseHamming` calculation does not go through this SIMT path, it always uses the WMMA kernel, so this check is unnecessary.

cpp/src/neighbors/detail/nn_descent_gnnd.hpp
yan-zaretskiy: The `local_join` is always called with the second argument, and the `DistEpilogue_t` is inferred from it, so this default is never used.

cpp/src/neighbors/detail/nn_descent.cuh
yan-zaretskiy: I believe `caltulate_metric` still:

1. applies `dist_epilogue`
2. fills out-of-range cells with `std::numeric_limits<float>::max()`

So it's not a no-op for L1. Please correct me if I am wrong.

jinsolp: you're right! sorry for the confusion

cpp/src/neighbors/detail/nn_descent.cuh
```diff
@@ -541,7 +570,12 @@ __device__ __forceinline__ void calculate_metric(float* s_distances,
 // MAX_RESIDENT_THREAD_PER_SM = BLOCK_SIZE * BLOCKS_PER_SM = 2048
 // For architectures 750 and 860 (890), the values for MAX_RESIDENT_THREAD_PER_SM
 // is 1024 and 1536 respectively, which means the bounds don't work anymore
-template <typename Index_t, typename ID_t = InternalID_t<Index_t>, typename DistEpilogue_t>
+template <typename DistanceOp_t,
```
divyegala: Can you pass the DistanceOp_t as a runtime argument? We are trying to be cognizant of our binary sizes.

cpp/src/neighbors/detail/nn_descent.cuh
```diff
@@ -1380,22 +1410,72 @@ void GNND<Data_t, Index_t>::local_join(cudaStream_t stream, DistEpilogue_t dist_
...
-                                                        l2_norms_.data_handle(),
-                                                        build_config_.metric,
-                                                        dist_epilogue);
+    auto* data_ptr = d_data_float_->data_handle();
+    switch (build_config_.metric) {
+      case cuvs::distance::DistanceType::InnerProduct:
+      case cuvs::distance::DistanceType::CosineExpanded:
+      case cuvs::distance::DistanceType::L2Expanded:
+      case cuvs::distance::DistanceType::L2SqrtExpanded:
+        local_join_kernel<dot_dist_op>
```
divyegala: This is getting hard for me to follow. Can you rename the one that does not use `wmma` instructions to `local_join_kernel_no_wmma`?

cpp/src/neighbors/detail/nn_descent.cuh
yan-zaretskiy: I understand what you mean. But a runtime argument would imply some form of a virtual dispatch, while this is kernel code.

divyegala: You could just pass the distance enum, and if-else on that like the other kernels do.

yan-zaretskiy: Fair, not sure why I was overthinking it. Will update.

yan-zaretskiy:
> In general, I'm finding the code complicated to follow now. Please consider a way to clarify when the `wmma` kernel is used and when the `non_wmma` kernel is used (meaning for what types and distances).

I simplified the local_join. Now there are 3 calls:
- F32 -> SIMT branch
- else if L1 distance (with any dype) -> SIMT branch
- else WMMA

cjnolet:
Yan, can you please update the doxygen for the public CAGRA and nn-descent functions? Need to say that nn-descent supports L1 and CAGRA supports L1 when using the nn-descent graph_build_algo.

cpp/include/cuvs/neighbors/cagra.hpp
```diff
@@ -1047,6 +1051,9 @@ auto build(raft::resources const& res,
  * The following distance metrics are supported:
  * - L2
  * - CosineExpanded (dataset norms are computed as float regardless of input data type)
+ * - L1
```
jinsolp:
```suggestion
 * - L1 (currently only supported with NN-Descent and Iterative Search as the build algorithm)
```

cpp/src/neighbors/detail/nn_descent.cuh
```diff
@@ -535,13 +556,28 @@ __device__ __forceinline__ void calculate_metric(float* s_distances,
   }
 }
 
+struct DistAccumulator {
+  cuvs::distance::DistanceType metric;
+  __device__ __forceinline__ float operator()(float a, float b) const
+  {
+    if (metric == cuvs::distance::DistanceType::L1) { return raft::abs(a - b); }
+    return a * b;
```
divyegala, line 565: Can you add a comment for this fallback mentioning that it is a reusable calculation for IP, cosine, and L2?

cpp/src/neighbors/detail/nn_descent.cuh
```diff
@@ -1380,39 +1414,56 @@ void GNND<Data_t, Index_t>::local_join(cudaStream_t stream, DistEpilogue_t dist_
...
-                                                        d_data_float_.value().data_handle(),
-                                                        ndim_,
-                                                        graph_buffer_.data_handle(),
-                                                        dists_buffer_.data_handle(),
-                                                        DEGREE_ON_DEVICE,
-                                                        d_locks_.data_handle(),
-                                                        l2_norms_.data_handle(),
-                                                        build_config_.metric,
-                                                        dist_epilogue);
+    local_join_kernel_simt<<<nrow_, BLOCK_SIZE, 0, stream>>>(graph_.h_graph_new.data_handle(),
```
divyegala, line 1423: Also a comment here and in the else path detailing what conditions lead to what kernel to be launched - I know you responded in a GH comment, it would just be nice to have that context in the code :)

PR #1837, opened 2026-02-23
Modernize the uses of raft in cuVS

achirkin:
A non-breaking src-only changes to modernize the use of raft primitives across cuVS source code. The general rule applied here is to prefer raft helpers taking `raft::resources` as an argument over other raft helpers over third-party libraries.

- thrust::fill / thrust::fill_n → raft::matrix::fill
- thrust::transform → raft::linalg::map
- thrust::sequence / thrust::tabulate → raft::linalg::map_offset
- raft::linalg::unaryOp / raft::linalg::binaryOp → raft::linalg::map
- raft::linalg::add (pointer-based) → raft::linalg::add (mdspan-based)
- raft::copy (pointer-based) → raft::copy (mdspan-based)
- raft::update_device / raft::update_host → raft::copy (mdspan-based)
- raft::linalg::rowNorm → raft::linalg::norm
- raft::linalg::reduce (pointer-based) → raft::linalg::reduce (mdspan-based)
- cudaMemsetAsync → raft::matrix::fill

The purpose of this PR is to improve the consistency in using the library code (even though sometimes at the cost of a bit more auxiliary code).
This is also a prerequisite to achieving dry run compliance in cuVS if we choose to merge that in https://github.com/rapidsai/raft/pull/2961

PR #1899, opened 2026-03-09
ScaNN: Fix AVQ prefetch

rmaschal:
Switching to usage of modern RAFT in cuVS (https://github.com/rapidsai/cuvs/pull/1837) introduced a bug where the prefetched gather for AVQ is performed using the stream associated with raft::device_resources rather than the provided stream for copying. This led to two issues:

1) Elimination of the benefit for prefetching, as copies where scheduled on the same stream as other gpu work
2) Possible recall loss. Synchronization was still performed against the copy stream, potentially allowing host to proceed before the prefetch copy is complete. 

This PR sets the stream associated with the resource to the copy stream before prefetching, and back when done.

### DIFF
--- a/cpp/src/neighbors/scann/detail/scann_avq.cuh
+++ b/cpp/src/neighbors/scann/detail/scann_avq.cuh
@@ -513,14 +513,14 @@ class cluster_loader {
     if (needs_copy_) {
       // For prefetching to overlap with other gpu work
       // we need to schedule copies on the provided copy stream stream_
-      auto stream = raft::resource::get_cuda_stream(res);
-      raft::resource::set_cuda_stream(res, stream_);
+      auto copy_res = raft::resources(res);
+      raft::resource::set_cuda_stream(copy_res, stream_);
 
       // htod
       auto h_cluster_ids =
         raft::make_pinned_vector_view<LabelT, int64_t>(cluster_ids_buf_.data_handle(), size);
 
-      raft::copy(res, h_cluster_ids, cluster_ids);
+      raft::copy(copy_res, h_cluster_ids, cluster_ids);
       raft::resource::sync_stream(res, stream_);
 
       auto pinned_cluster = raft::make_pinned_matrix_view<T, int64_t>(
@@ -534,11 +534,8 @@ class cluster_loader {
                sizeof(T) * dim_);
       }
 
-      raft::copy(res, cluster_vectors, raft::make_const_mdspan(pinned_cluster));
+      raft::copy(copy_res, cluster_vectors, raft::make_const_mdspan(pinned_cluster));
       raft::resource::sync_stream(res, stream_);
-
-      // reset stream back to previous value
-      raft::resource::set_cuda_stream(res, stream);
     } else {
       // dtod
       auto dataset_view =


cpp/src/neighbors/scann/detail/scann_avq.cuh
```diff
@@ -532,6 +537,8 @@ class cluster_loader {
       raft::copy(res, cluster_vectors, raft::make_const_mdspan(pinned_cluster));
       raft::resource::sync_stream(res, stream_);
 
+      // reset stream back to previous value
+      raft::resource::set_cuda_stream(res, stream);
```
divyegala: You can create a new resources object for no cost:

```suggestion
      // For prefetching to overlap with other gpu work
      // we need to schedule copies on the provided copy stream stream_
      auto copy_res = raft::resources(stream_);

// htod
auto h_cluster_ids =
raft::make_pinned_vector_view<LabelT, int64_t>(cluster_ids_buf_.data_handle(), size);
@@ -532,6 +537,8 @@ class cluster_loader {
raft::copy(copy_res, cluster_vectors, raft::make_const_mdspan(pinned_cluster));
raft::resource::sync_stream(res, stream_);
```

divyegala (changes requested):

cpp/src/neighbors/scann/detail/scann_avq.cuh
achirkin: I vote on this too. Although not very relevant here, another benefit of creating a temporary resources handle is the automatic reset of the changes if an exception is raised. However, I'd like to warn that there are caveats in doing this:
1) I'd suggest to use a copy the resources handle rather than creating a new resources handle to inherit all other initialized resources from the current handle.
```C++
auto copy_res = raft::resources(res);
raft::resource::set_cuda_stream(res, stream);
```
2) Even when you do a copy: if you've used _any_ other not-yet-initialized resources from the `copy_res` handle, it would create those resources on every invocation of `prefetch_cluster`, which would incur significant overhead.

rmaschal: Thank you for the comments and info. I now copy the passed resources, set the stream, and use that for scheduling copies

divyegala (approved):

aamijar:
Hi @rmaschal, is this targeting release/26.04? If so, please retarget from main to release/26.04

rmaschal:
> Hi @rmaschal, is this targeting release/26.04? If so, please retarget from main to release/26.04

Hi @aamijar, just saw the comment, I assume it's too late for 26.04? In that case main is fine, then targeting 26.06

aamijar:
Hi @rmaschal, looks like we are good to merge this right, or are you seeing CI failures related this PR? No need to keep merging in main, as long as the Recently Updated check is passing. You can also rerun failed jobs if they are flaky.

rmaschal:
@aamijar yes should be good, some python CI failures but they should be unrelated.

aamijar:
Great! I will try to rebase to release/26.04. Would be nice to have in the release since it a bug fix/ perf regression fix.

Issue #1903, opened 2026-03-10
[FEA] Improve batched `all-neighbors` given CPU indices/distances


jinsolp:
Current implementation uses managed memory to control single reference of indices/distances.
Using managed memory for large indices/distances arrays may oversubscribe the GPU and lead to performance issues.

Explore performance implications with large indices/distances arrays and work on optimizations if needed.

PR #1905, opened 2026-03-10
API for CPU side output indices and distances of `all_neighbors`

### LINKED ISSUE #1902: [FEA] API for CPU side indices/distances in `all-neighbors`
Current all-neighbors only allows returning device side indices/distances arrays.
Allow returning CPU side indices/distances arrays at the API level.

jinsolp:
Closes https://github.com/rapidsai/cuvs/issues/1902

**This PR allows indices/distances result of the `all_neighbors` function to be on host memory. Focuses on exposing the API, and internally just does a copy.**

Internally, it still allocates GPU memory for the full indices/distances
- For `n_clusters=1` this is not a problem. If the user faces OOM issues they should use `n_clusters>1`.
- For `n_clusters>1` we are still using managed memory regardless of the memory location of the user given indices/distances. Will open a follow-up PR with optimizations (related issue: https://github.com/rapidsai/cuvs/issues/1903)


cpp/src/neighbors/all_neighbors/all_neighbors_batched.cuh
```diff
@@ -616,6 +622,8 @@ void batch_build(
       raft::make_device_vector_view(distances.value().data_handle(), num_rows * k),
       raft::make_device_vector_view<const T>(global_distances.data_handle(), num_rows * k));
   }
+
+  raft::resource::sync_stream(handle);
```
aamijar: Do we need a sync because we need to wait for GPU to finish some task?

aamijar (commented):
Thanks for the PR Jinsol! 
So currently, with the old code the API only returns distances and indices on device and would have to manually copy to host if needed? 
Now with this new change the copy step is hidden from the user?

cpp/src/neighbors/all_neighbors/all_neighbors_batched.cuh
jinsolp: this is removed!

jinsolp:
@aamijar yes, so eventually all-neighbors will be further optimized for host side distances/indices!
I didn't want to make the PR too big, so this PR only contains the API exposure.

c/include/cuvs/neighbors/all_neighbors.h
```diff
@@ -28,7 +28,8 @@ extern "C" {
  * provide the dataset on host.
  *
  * Notes:
- * - Outputs (indices, distances, core_distances) are expected to be on device memory.
+ * - Outputs (indices, distances) can be on host memory (numpy arrays)
```
dantegd, line 31: It's kinda odd to see Python terminology in the C header, I guess one of the main consumers are the python apis, but still... not necessarily suggesting to change it, but was wondering ig it might be better to keep the docstring to C relevant concepts (like dlpack)

c/include/cuvs/neighbors/all_neighbors.h
```diff
@@ -94,16 +95,16 @@ CUVS_EXPORT cuvsError_t cuvsAllNeighborsIndexParamsDestroy(cuvsAllNeighborsIndex
...
+ * @param[out] distances      Optional 2D tensor [num_rows x k] on host or device (float32); can be NULL
  * @param[out] core_distances Optional 1D tensor [num_rows] on device (float32); can be NULL
  * @param[in] alpha           Mutual-reachability scaling; used only when core_distances is provided
  *
  * The function automatically detects whether the dataset is host-resident or device-resident
  * and calls the appropriate implementation. For host datasets, it partitions data into
  * `n_clusters` clusters and assigns each row to `overlap_factor` nearest clusters. For device
  * datasets, `n_clusters` must be 1 (no batching); `overlap_factor` is ignored.
- * Outputs always reside in device memory.
+ * Outputs can be on host memory (numpy arrays) or device memory (CUDA arrays).
```
dantegd, line 107: Same thing, it sticks out to me. To be honest I didn't check if we used it in other places, but also the other thing is that numpy arrays are just one potential origin for host memory, it could just be C arrays, or say other host python arrays from the python layer, and this seems to imply that there's something that would make these functions need numpy arrays, so I'm leaning towards removing numpy from C docs.

cpp/src/neighbors/all_neighbors/all_neighbors.cuh
```diff
@@ -7,13 +7,44 @@
...
+{
+  size_t n_rows = in_out.extent(0);
+  size_t n_cols = in_out.extent(1);
+  RAFT_EXPECTS(n_cols > k, "Shift size k should be smaller than the number of columns in matrix.");
+
+  // Shift columns to the right by k
+#pragma omp parallel for
+  for (size_t i = 0; i < n_rows; i++) {
+    // Copy columns from right to left
+    for (size_t j = n_cols - 1; j >= k; j--) {
```
dantegd, line 33: Minor thing, but j is unsigned, so j >= 0 is always true. The current RAFT_EXPECTS(n_cols > k, …) doesn't catch k == 0 which would make this be an infinite loop.

Today everyone calls this with k>=1, so it's not exploitable, but since it's a generic-looking helper in detail::, mind adding a RAFT_EXPECTS(k > 0, …) so a future caller can't trip on it?

python/cuvs/cuvs/tests/test_all_neighbors.py
```diff
@@ -51,9 +51,19 @@ def make_cosine(
         "jensenshannon",
     ],
 )
-def test_all_neighbors_device_build_quality(algo, cluster, metric):
+@pytest.mark.parametrize(
+    "output_location",
+    ["host_arrays", "device_arrays", "return_on_host", "return_on_device"],
```
dantegd, line 56: Nice coverage of the placement combos. One gap is that there's no test that exercises core_distances  with host outputs. Might be a good idea tp parametrize output_location on the existing mutual-reach test, or maybe add a small dedicated test

c/src/neighbors/all_neighbors.cpp
```diff
@@ -138,26 +141,63 @@ void _build_host(cuvsResources_t res,
...
+      cuvs::core::is_dlpack_host_compatible(core_distances_tensor->dl_tensor);
+    RAFT_EXPECTS(core_distances_is_host == indices_is_host,
+                 "core_distances must be on the same memory location as indices and distances");
   }
 
-  cuvs::neighbors::all_neighbors::build(
-    cpp_res, cpp_params, dataset, indices, distances, core_distances, alpha);
+  auto dataset = cuvs::core::from_dlpack<dataset_mdspan_t>(dataset_tensor);
+
+  if (indices_is_host) {
```
dantegd, line 162: The if (indices_is_host) { … } else { … } blocks here are essentially the same 30-ish lines duplicated four times across _build_host and _build_device (host branch and device branch in each). Would you be open to extracting a small templated helper that takes the dataset mdspan type + a host/device tag and does the optional unpacking + dispatch? Should cut this down a lot, or is this cod that'll significantly change in teh follow up and not worth the change?

cpp/src/neighbors/all_neighbors/all_neighbors.cuh
```diff
@@ -140,52 +228,98 @@ void build(
...
+      if (distances.has_value()) {
+        host_shift_columns(distances.value(), 1, std::make_optional<T>(0.0));
+      }
+    } else {
+      auto indices_d = raft::make_device_matrix_view<IdxT, IdxT>(
+        const_cast<IdxT*>(indices.data_handle()), indices.extent(0), indices.extent(1));
+      raft::matrix::shift(handle, indices_d, 1);
+      if (distances.has_value()) {
+        auto distances_d =
+          raft::make_device_matrix_view<T, IdxT>(const_cast<T*>(distances.value().data_handle()),
```
dantegd, line 243: Tiny nit (applies to all 8 of these in this file): I think the const_cast<IdxT*> and const_cast<T*> here are no-ops? We're in the else branch, so outputs_are_host is false, which means IndicesMdspan/DistancesMdspan are the device instantiations from all_neighbors.cu: device_matrix_view<IdxT, IdxT> and device_matrix_view<T, IdxT>, and data_handle() already gives us IdxT*/T*. single_build does the same thing right above without a cast. Mind dropping these? My worry is mostly that if someone later adds a device_matrix_view<const IdxT, …> instantiation, the cast would silently strip const on an output buffer.

That said, if you think the follow up will change it singificantly regardless, then happy to drop the comment here.

dantegd (changes requested):
Just had a few questions and comments

jinsolp:
Closing because this feature was included in https://github.com/NVIDIA/cuvs/pull/2313

PR #1907, opened 2026-03-10
Add cuvs-bench-elastic: HTTP backend for Elasticsearch GPU vector search

afourniernv:
Introduce cuvs-bench-elastic as an optional plugin for cuvs-bench that provides an Elasticsearch backend. The backend communicates with Elasticsearch via HTTP and supports HNSW indexing with optional GPU acceleration when using the Elasticsearch GPU image (cuVS-accelerated vector search).

- Add cuvs_bench_elastic package with backend and config loader entry points
- Extend cuvs_bench registry and search spaces for pluggable backends
- Add elastic and integration optional dependencies to cuvs-bench
- Add modularization tests and integration test scaffolding (disabled until CI has ES GPU image, cuVS libs, and GPU runner)


python/cuvs_bench/cuvs_bench/backends/registry.py
```diff
@@ -375,10 +380,43 @@ def get_backend(name: str, config: Dict) -> BenchmarkBackend:
...
+        if hasattr(eps, "select"):  # Python 3.10+
+            eps = list(eps.select(name=name))
+        else:
+            eps = [e for e in eps if e.name == name]
+        for ep in eps:
+            try:
+                ep.load()()
+            except ImportError as e:
+                if "elasticsearch" in str(e).lower() or "elasticsearch" in str(e):
+                    raise ImportError(
```
jnke2016: Nice. This is what I discussed with @cjnolet pertaining to lazy import and used Milvus as an example

```
class MilvusBackend(BenchmarkBackend):
    def __init__(self, config: Dict[str, Any]):
        super().__init__(config)
        try:
            from pymilvus import connections, Collection
        except ImportError:
            raise ImportError(
                "pymilvus is required for MilvusBackend. "
                "Install with: pip install pymilvus"
            )
        connections.connect(host=config["host"], port=config["port"])
```

python/cuvs_bench/cuvs_bench/tests/integration/conftest.py
```diff
@@ -0,0 +1,73 @@
...
+def _elasticsearch_installed():
+    try:
+        import elasticsearch  # noqa: F401
+        return True
+    except ImportError:
+        return False
+
+
+@pytest.fixture(scope="module")
+def elasticsearch_container():
```
cjnolet: Why a container? We should just be able to assume an elasticsearch cluster is ready to accept requests, right? cuvs-bench doesn't need to be self-contained, just be able to send the proper requests to an existing elasticsearch cluster. Or am I missing a big detail here?

cjnolet: This just seems unnecessary.

python/cuvs_bench/cuvs_bench/backends/elasticsearch.py
```diff
@@ -0,0 +1,663 @@
...
+    "vector_field",
+)
+# Search params passed through from config (knn)
+_SEARCH_PARAM_KEYS = ("num_candidates", "vector_field")
+
+
+class ElasticBackend(BenchmarkBackend):
+    """Elasticsearch GPU backend for vector benchmarking."""
+
+    def __init__(self, config: Dict[str, Any]):
```
cjnolet, line 120: Oh I see- is the container just for the integration tests?

python/cuvs_bench_elastic/pyproject.toml
```diff
@@ -0,0 +1,50 @@
+# SPDX-FileCopyrightText: Copyright (c) 2024-2026, NVIDIA CORPORATION.
+# SPDX-License-Identifier: Apache-2.0
+
```
cjnolet: We can't have different packages deployed for every backend. This significantly adds to the maintenance burden as every new package now needs to be versioned, deloyed, and audited for dependency trails. Let's consolidate this into the existing cuvs-bench package. All dependencies on anything elasticsearch or cuvs-lucene should be soft dependendencies in Python (that is, they should test if they can import the package and if they can't, throw a warning and fail gracefully).

python/cuvs_bench/cuvs_bench/backends/elasticsearch.py
```diff
@@ -0,0 +1,772 @@
...
+from .registry import register_backend, register_config_loader
+from ..orchestrator.config_loaders import (
+    BenchmarkConfig,
+    ConfigLoader,
+    DatasetConfig,
+    IndexConfig,
+)
+
+
+def _load_fbin(path: Path) -> np.ndarray:
```
cjnolet, line 38: Can we centralize these funtions or used existing centralized functions? Really don't want the file formats we support to diverge across different backends, so we should have 1 way to load the datasets.

python/cuvs_bench/cuvs_bench/backends/elasticsearch.py
```diff
@@ -0,0 +1,772 @@
...
+        if self._client is None:
+            try:
+                from elasticsearch import Elasticsearch
+            except ImportError as e:
+                raise ImportError(
+                    "`elasticsearch` is required for the Elasticsearch backend. "
+                    "Install with: pip install cuvs-bench[elastic]"
+                ) from e
+            host = self.config.get("host", "localhost")
+            port = self.config.get("port", 9200)
```
cjnolet, line 105: Just want to confirm- are there any other protocols to interact with Elasticsearch or is HTTP/Rest pretty much it? Mostly asking to make sure it's the best (fastest) way to comunicate so we aren't adding additional overheads in the mix.

python/cuvs_bench/cuvs_bench/backends/elasticsearch.py
cjnolet: cc @jnke2016

python/cuvs_bench/cuvs_bench/backends/elasticsearch.py
```diff
@@ -0,0 +1,772 @@
...
+            vectors = _load_fbin(base_path)
+            n_vectors = len(vectors)
+            dims = vectors.shape[1]
+            client = self._get_client()
+
+            vector_field = build_params.get("vector_field", _DEFAULT_VECTOR_FIELD)
+            index_type = build_params.get("type", _DEFAULT_INDEX_TYPE)
+            m = build_params.get("m", _DEFAULT_M)
+            ef_construction = build_params.get(
+                "ef_construction", _DEFAULT_EF_CONSTRUCTION
```
cjnolet, line 240: We should decouple the index type from the backend type. User should be able to specify the backend and the algorithm they want to run in that backend (or maybe the two are coupled, but the backend should not assume the algorithm). 

So for example, what if the index type is not supported? What happens when a user wants to test diskbbq against hnsw/cagra? We should decouple these from the beginning, ideally with separate functions for each (for e.g. `_parse_cagra_params()`, `_parse_hnsw_params()`)

cjnolet: It's also important that we throw a proper (descriptive) error when the backend doesn't support an index type. For example, if someone passes in "SCaNN" to the Elasticsearch backend, they should get a message to the effect of "Received params for SCaNN index type, but Elasticsearch backend does not support this index type. Please check configuration."

python/cuvs_bench/cuvs_bench/backends/elasticsearch.py
```diff
@@ -0,0 +1,772 @@
...
+        batch_size: int = 10000,
+        mode: str = "latency",
+        force: bool = False,
+        search_threads: Optional[int] = None,
+        dry_run: bool = False,
+    ) -> SearchResult:
+        """Run kNN search over all search-param combinations and compute recall."""
+        if dry_run:
+            return SearchResult(
+                neighbors=np.zeros((0, k), dtype=np.int64),
```
cjnolet, line 333: Suggest using np.empty() for all of these.

python/cuvs_bench/cuvs_bench/tests/test_modularization.py
```diff
@@ -0,0 +1,659 @@
+#
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026, NVIDIA CORPORATION.
+# SPDX-License-Identifier: Apache-2.0
+#
+"""
+Smoke tests for cuvs-bench modularization (optional deps, entry points, lazy loading).
```
cjnolet, line 6: It's great to see improved test coverage!

cjnolet:
@afourniernv this is a really big PR so I'm reviewing in stages. Thank you for bearing with me. I finally had some time to look through the implementation for comprehensively. Mostly minor, but important things.

python/cuvs_bench/cuvs_bench/backends/elasticsearch.py
jnke2016: We definitely need to centralize this. I went ahead and compared both the OpenSearch and Elasticsearch backends side by side to identify which functions can be generalized into shared utilities. For vector loading specifically, OpenSearch has a single _load_vectors that handles all five binary formats (.fbin, .f16bin, .u8bin, .i8bin, .ibin) via a dtype lookup table and supports subset_size, while Elasticsearch has two separate functions (_load_fbin and _load_ibin) that only handle float32 and int32 respectively. These all parse the same underlying binary format (uint32 header followed by typed data) and should be a single shared utility so that format support stays consistent and any bug fix or new dtype only needs to happen in one place

cjnolet: We shoukd use the same data loading tools that the original scripts were using (before the new backends). We need to be able to load all data precisions (fo32, fp16, uint8, int8). We also need tools to be able to produce these files from a dataset in memory (ideally to be streamable for both input and output so that a user can read/write a batch of vectors at a time without having to flush and close the file after every batch.

jnke2016: Thanks for bringing this to my attention. There are indeed already existing data loading utilities in the codebase. `generate_groundtruth/utils.py` has `memmap_bin_file()` with `dtype_from_filename()` supporting all dtypes, memory-mapped read/write, and batch access via slicing. There is also duplicated I/O code scattered across multiple modules (`generate_groundtruth/utils.py, get_dataset/fbin_to_f16bin.py, get_dataset/hdf5_to_fbin.py`). Not sure if these should be consolidated into a single shared location in the future, but putting them in `backends/utils.py` wouldn't be the right place since `generate_groundtruth` and `get_dataset` are not backends and shouldn't need to import from the backends subpackage. Similarly, having backends import from `generate_groundtruth/utils.py` would be equally awkward. None of these subpackages currently import from each other. A top-level shared module like cuvs_bench/io.py could be the right place since it would break no existing dependencies and all three subpackages could import from it cleanly. Open to suggestions on how you'd like to handle this.

cjnolet: I'm not as concerned with where these common pieces live. But neither backends nor users should have to know that they exist- loading and saving files and configuration params is something the benchmarking framework should be doing, not the backends nor the users. It's the job of the framework to centralize these things.

python/cuvs_bench/cuvs_bench/backends/elasticsearch.py
```diff
@@ -0,0 +1,772 @@
...
+                f"Available: {list(groups.keys())}"
+            )
+        group_conf = groups[group_name]
+
+        build_params = group_conf.get(
+            "build", {"m": [16], "ef_construction": [100]}
+        )
+        search_params = group_conf.get("search", {"num_candidates": [100]})
+
+        # Ensure all param values are lists for itertools.product
```
jnke2016: **Cartesian product expansion**: I've added expand_param_grid() in `backends/utils.py` that handles the expansion via Cartesian product. The expansion block at lines 646-660 would be replaced with:

```
from .utils import expand_param_grid

build_combos = expand_param_grid(_to_list_values(build_params))
search_params_list = expand_param_grid(_to_list_values(search_params))
```

python/cuvs_bench/cuvs_bench/backends/elasticsearch.py
```diff
@@ -0,0 +1,772 @@
...
+                    latencies.append((time.perf_counter() - t_q) * 1000)
+                    hits = resp.get("hits", {}).get("hits", [])
+                    for j, hit in enumerate(hits[:k]):
+                        neighbors[i, j] = int(hit["_id"])
+                        distances[i, j] = float(hit["_score"])
+
+                elapsed_ms = (time.perf_counter() - t0) * 1000
+                qps = n_queries / (elapsed_ms / 1000) if elapsed_ms > 0 else 0.0
+
+                recall = 0.0
```
jnke2016: **Recall calculation**: I've added a shared compute_recall() in `backends/utils.py` so that any bug fix or improvement happens in one place and applies to all Python-native backends. The inline recall block at lines 444-451 would be replaced with:
```
from .utils import compute_recall

recall = 0.0
if groundtruth is not None:
    recall = compute_recall(neighbors, groundtruth, k)
```

python/cuvs_bench/cuvs_bench/backends/elasticsearch.py
jnke2016: This is now handled transparently by the framework. Backend authors no longer need to know about or call `expand_param_grid()` directly.

python/cuvs_bench/cuvs_bench/backends/elasticsearch.py
jnke2016: `recall` is now computed by the orchestrator after `backend.search()` returns. Backends no longer compute it themselves. The C++ backend is naturally skipped since it returns empty neighbor arrays (`recall` is already computed by the subprocess).

python/cuvs_bench/cuvs_bench/backends/registry.py
afourniernv: Aligned this with the in-tree optional-backend model now on `main`. The separate package was folded back into `cuvs_bench`, backend imports remain lazy at client construction time, and the registry now rewrites optional-backend `ImportError`s into backend-specific install hints (`cuvs-bench[elastic]` / `cuvs-bench[opensearch]`).

python/cuvs_bench/cuvs_bench/tests/integration/conftest.py
afourniernv: Yes. The backend now assumes an existing Elasticsearch endpoint and does not orchestrate containers. The Brev/docker helper scripts stayed out of the PR; the benchmark contract here is just host/port plus backend params.

python/cuvs_bench/cuvs_bench/backends/elasticsearch.py
afourniernv: Correct. The runtime backend path assumes an already-running cluster and only issues HTTP requests to it; no container lifecycle is part of the backend.

python/cuvs_bench_elastic/pyproject.toml
afourniernv: Addressed. The separate elastic package was folded back into `python/cuvs_bench`, and Elastic/OpenSearch are both treated as optional backends with lazy imports and backend-specific install hints.

python/cuvs_bench/cuvs_bench/backends/elasticsearch.py
afourniernv: Moved toward the framework-owned path: build/search now consume `Dataset.training_vectors`, `Dataset.query_vectors`, and `Dataset.groundtruth_neighbors`, and the elastic config loader now goes through the shared `ConfigLoader` flow. The remaining `_load_fbin/_load_ibin` wrappers just delegate to the shared loader for compatibility rather than owning parsing logic themselves.

python/cuvs_bench/cuvs_bench/backends/elasticsearch.py
afourniernv: Confirmed. For Elasticsearch the supported client path here is the official HTTP/REST interface; the backend does not add another protocol layer on top of that. We re-ran the backend live against the Brev deployment after the recent fixes. Leaving this open since it is more an interface tradeoff than a code bug.

python/cuvs_bench/cuvs_bench/backends/elasticsearch.py
afourniernv: Partially addressed. The backend now validates supported algorithm names, index types, and similarities up front with descriptive errors instead of failing later in Elasticsearch. It still models the benchmark family as `elastic_*` plus build-param `type`, so this is better than before but not a fully generic backend/index abstraction yet.

python/cuvs_bench/cuvs_bench/backends/elasticsearch.py
afourniernv: Updated the zero-length result buffers to use `np.empty(...)` in both Elastic and OpenSearch.

python/cuvs_bench/cuvs_bench/backends/elasticsearch.py
afourniernv: Handled by the shared config-loader expansion now; the elastic loader no longer does its own manual Cartesian-product expansion.

python/cuvs_bench/cuvs_bench/backends/elasticsearch.py
afourniernv: Handled now by the orchestrator. The backend marks aggregated recall as authoritative for multi-search-param runs, and the orchestrator skips recomputation in that case so the averaged recall is preserved.

python/cuvs_bench/cuvs_bench/backends/elasticsearch.py
```diff
@@ -0,0 +1,921 @@
...
+if TYPE_CHECKING:
+    from elasticsearch import Elasticsearch
+
+
+def _load_fbin(path: Path) -> np.ndarray:
+    """Load big-ann-bench fbin format via shared vector loader."""
+    return load_vectors(os.fspath(path))
+
+
+def _load_ibin(path: Path) -> np.ndarray:
```
jnke2016: One small cleanup: _load_fbin and _load_ibin are now thin wrappers around load_vectors and aren't called anywhere, so they can be removed.

python/cuvs_bench/cuvs_bench/backends/elasticsearch.py
```diff
@@ -0,0 +1,921 @@
...
+                index_path="",
+                build_time_seconds=0,
+                index_size_bytes=0,
+                algorithm=self.algo,
+                build_params={},
+                success=False,
+                error_message=str(e),
+            )
+
+        vectors = dataset.training_vectors
```
jnke2016: Vector loading looks good now and matches the centralized approach. build() reads dataset.training_vectors and search() reads dataset.query_vectors and dataset.groundtruth_neighbors, so the framework handles loading transparently and the backend no longer parses files itself. This also picks up the shared format support (uint64 headers, .f16bin/.u8bin/.i8bin, subset_size) along

python/cuvs_bench/cuvs_bench/backends/elasticsearch.py
```diff
@@ -0,0 +1,921 @@
...
+    """Config loader for Elasticsearch backend."""
+
+    def __init__(self, config_path: Optional[str] = None):
+        self.config_path = config_path or _get_cuvs_bench_config_path()
+
+    @property
+    def backend_type(self) -> str:
+        return "elastic"
+
+    def load(
```
jnke2016: `ElasticConfigLoader.load()` can be removed. It only forwards to `super().load()`, so it adds nothing, the base `ConfigLoader.load()` already runs all the shared steps and calls into the backend-specific hooks `_discover_algo_groups()` and `_build_benchmark_configs()`, which is where the elastic logic lives. `OpenSearch` relies on this directly: it defines no `load()`, inherits the base one, and implements only those two hooks. Dropping it keeps elastic consistent with `OpenSearch`

python/cuvs_bench/cuvs_bench/backends/elasticsearch.py
```diff
@@ -0,0 +1,921 @@
...
+        dataset: str = "",
+        dataset_path: str = "",
+        **kwargs,
+    ) -> Tuple[DatasetConfig, List[BenchmarkConfig]]:
+        """Load Elasticsearch benchmark configuration via shared ConfigLoader flow."""
+        return super().load(
+            dataset=dataset, dataset_path=dataset_path, **kwargs
+        )
+
+    def _discover_algo_groups(
```
jnke2016, line 571: Parameter expansion looks good. `_discover_algo_groups()` just returns the algo groups and `_build_benchmark_configs()` receives the already-expanded `build_combos` / `search_combos` from the base loader. The framework handles the expansion, same as `OpenSearch`.

python/cuvs_bench/cuvs_bench/backends/elasticsearch.py
jnke2016: I reviewed it and the recall isn't actually computed by the orchestrator. It's still computed inside `search()` (lines 506‑518), and the orchestrator only skips its own computation because the result is marked `recall_is_authoritative` (line 559). So the flow is backend-computed and orchestrator-preserved, not orchestrator-computed. The intent of centralizing was for the orchestrator to compute `recall` for every backend; this path keeps it in the backend and re-implements the recall logic that already lives in the shared `compute_recall()` util in `backends/_utils.py`.

jnke2016: In fact, after `search()` runs the queries it computes `recall` itself per search-param combo (lines 506‑518), averages it (534), and returns it with metadata={"recall_is_authoritative": True} (559). The orchestrator's `_should_compute_recall()` sees that flag and skips its own computation, so ES's value is kept instead of the centralized one.

The downside is that this reverses the centralization. The recall logic here duplicates the shared `compute_recall()` in `backends/_utils.py`, so it's no longer computed in one place. Any future change or bug fix to `recall` would have to be made in the `orchestrator/shared` util and re-applied in every backend that re-implemented it, exactly the per-backend divergence the centralization was meant to prevent.

jnke2016: I acknowledge that this is a limitation of the current orchestrator, not just an ES choice: search() currently returns one SearchResult, so when a backend sweeps multiple combos it can only hold a single neighbors array. ES (like OpenSearch) keeps the last combo's (last_neighbors). Because the orchestrator could then only score that last combo, ES computes recall for every combo itself and marks it authoritative so it isn't overwritten. Below is the current workflow as it stands
**Current workflow (backend owns recall):**

_**Backend (ES):**_
```
for sp in search_params_list:
    # run queries -> neighbors
    recall = compute_inline(neighbors, groundtruth, k)   # 506-518
    last_neighbors = neighbors                           # only last combo kept
return SearchResult(neighbors=last_neighbors, recall=avg_recall,
                    metadata={"recall_is_authoritative": True})   # 559
```
**_Orchestrator:_**
```
search_result = backend.search(...)
if _should_compute_recall(search_result):    # False for ES (flag set)
    search_result.recall = compute_recall(search_result.neighbors, gt, count)
results.append(search_result)
```

jnke2016: Below is what I'd suggest so recall computation stays centralized in the orchestrator:

**Proposed workflow (orchestrator owns recall)**

**_Backend:_** build one SearchResult per combo (recall=0.0, no flag) and return them as a list:

```
results = []
for sp in search_params_list:
    # run queries -> neighbors
    results.append(SearchResult(neighbors=neighbors, recall=0.0, search_params=[sp]))
return results
```
**_Orchestrator:_** iterate the results, compute recall per combo via the shared util:
```
search_results = backend.search(...)
if isinstance(search_results, SearchResult):   # back-compat: C++ returns one
    search_results = [search_results]
for r in search_results:
    if _should_compute_recall(r) and gt is not None:
        r.recall = compute_recall(r.neighbors, gt, count)
    results.append(r)
```

Every combo is scored by the shared `compute_recall()`, so recall stays centralized, no in-backend recall block and no recall_is_authoritative. `OpenSearch` also gets the same benefit

jnke2016: I like the `_should_compute_recall `abstraction, it makes the orchestrator's decision explicit and keeps it in one readable place, rather than relying on the implicit `recall == 0.0` convention to signal whether the orchestrator should compute recall

python/cuvs_bench/cuvs_bench/backends/elasticsearch.py
```diff
@@ -0,0 +1,921 @@
...
+
+        index_name = self.config.get("index_name", "cuvs_bench_vectors")
+        idx = indexes[0] if indexes else None
+        build_params = dict(idx.build_param or {}) if idx else {}
+        for k, v in self.config.items():
+            if k not in build_params and k in _BUILD_PARAM_KEYS:
+                build_params[k] = v
+
+        try:
+            client = self._get_client()
```
jnke2016, line 249: Is the reason for connecting lazily through `_get_client()` inside `build()/search()`, rather than in `initialize()`, to avoid reaching a live server on a dry_run? The orchestrator always calls `initialize()`, even for a dry run, while `build()/search()` return early on dry_run before touching the client, so the lazy path seems to keep dry runs from needing a server.

python/cuvs_bench/cuvs_bench/backends/elasticsearch.py
```diff
@@ -0,0 +1,921 @@
...
+                algorithm=self.algo,
+                build_params={},
+                success=False,
+                error_message=f"pre-flight check failed: {skip_reason}",
+            )
+
+        index_name = self.config.get("index_name", "cuvs_bench_vectors")
+        idx = indexes[0] if indexes else None
+        build_params = dict(idx.build_param or {}) if idx else {}
+        for k, v in self.config.items():
```
jnke2016: This loop is redundant in the normal flow. `build_params` already copies `idx.build_param` (line 223), and the loader spreads that same dict into backend_config via `**build_param` (line 818), so every `_BUILD_PARAM_KEY` in `self.config` is already in build_params. The remaining entries in self.config are connection keys (`host`, `port`, …) that the loop's `_BUILD_PARAM_KEYS` filter skips. So it never adds anything and can be dropped, leaving idx.build_param as the single source.

The only case where the loop does something is the unusual one where indexes is empty (idx is None, so `build_params` starts as {}) — but that doesn't happen in the orchestrator path, since each config carries exactly one `IndexConfig`.

python/cuvs_bench/cuvs_bench/backends/elasticsearch.py
```diff
@@ -0,0 +1,921 @@
...
+    if not reg.is_registered("elastic"):
+        register_backend("elastic", ElasticBackend)
+    if "elastic" not in _CONFIG_LOADER_REGISTRY:
+        register_config_loader("elastic", ElasticConfigLoader)
+
+
+# ── Convenience API ───────────────────────────────────────────────────────────
+
+
+def run_build(
```
jnke2016: The module-level `run_build` / `run_search` / `run_benchmark` helpers duplicate the entry point already provided by `BenchmarkOrchestrator` and the `CLI`, they just call `orchestrator.run_benchmark`(...) as below:

**run_build(...)**

> **_# Python API_**

```
BenchmarkOrchestrator(backend_type="elastic").run_benchmark(
    build=True, search=False,
    dataset=..., dataset_path=..., algorithms=..., host=..., port=..., force=force,
)
```

> **_# CLI_**

```
python -m cuvs_bench.run --build \
    --dataset <name> --dataset-path <path> --algorithms <group> \
    --backend-config <elastic-config>
```

**run_search(...)**

> **_# Python API_**

```
BenchmarkOrchestrator(backend_type="elastic").run_benchmark(
    build=False, search=True,
    dataset=..., dataset_path=..., algorithms=..., host=..., port=...,
)
```

> **_# CLI_**

```
python -m cuvs_bench.run --search \
    --dataset <name> --dataset-path <path> --algorithms <group> \
    --backend-config <elastic-config>
```

**run_benchmark(...)**

> **_# Python API_**

```
BenchmarkOrchestrator(backend_type="elastic").run_benchmark(
    build=build, search=search,
    dataset=..., dataset_path=..., algorithms=..., host=..., port=..., force=force,
)
```

> **_# CLI_**

```
python -m cuvs_bench.run --build --search \
    --dataset <name> --dataset-path <path> --algorithms <group> \
    --backend-config <elastic-config>
```

python/cuvs_bench/cuvs_bench/backends/elasticsearch.py
afourniernv: Addressed in a0a4b673b. The unused `_load_fbin` / `_load_ibin` wrappers and their now-unused shared-loader import are removed, so the backend just relies on the framework-owned dataset loading path.

python/cuvs_bench/cuvs_bench/backends/elasticsearch.py
afourniernv: Addressed in fe34fd465. `ElasticConfigLoader` now just inherits the base `ConfigLoader.load()` path directly, matching the OpenSearch loader and leaving the elastic-specific behavior in the hook methods only.

python/cuvs_bench/cuvs_bench/backends/elasticsearch.py
afourniernv: Addressed in 1b81f4469. `build()` now takes its build params directly from the `IndexConfig` it was given, which is the normal orchestrator path and matches the OpenSearch backend pattern instead of redundantly backfilling the same keys from `self.config`.

python/cuvs_bench/cuvs_bench/backends/elasticsearch.py
afourniernv: Addressed in fcc13f546. The elastic-only `run_build` / `run_search` / `run_benchmark` wrappers are removed so the backend relies on the same orchestrator and CLI entrypoints as the rest of cuvs-bench, matching the OpenSearch backend structure.

python/cuvs_bench/cuvs_bench/tests/conftest.py
```diff
@@ -0,0 +1,23 @@
...
+    try:
+        import elasticsearch  # noqa: F401
+    except ImportError:
+        return
+
+    try:
+        from cuvs_bench_elastic import register
+        register()
+    except ImportError:
+        pass
```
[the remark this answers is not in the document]
afourniernv: Addressed in cb5b0fda4. `pytest_configure()` now tries the legacy `cuvs_bench_elastic.register` path first and falls back to the in-tree `cuvs_bench.backends.elasticsearch.register`, without silently swallowing a remaining import failure.

dependencies.yaml
```diff
@@ -585,6 +593,11 @@ dependencies:
       - output_types: [requirements, pyproject]
         packages:
           - matplotlib>=3.9
+  bench_elastic:
+    common:
+      - output_types: [conda, pyproject, requirements]
+        packages:
+          - cuvs-bench-elastic>=26.4.0
```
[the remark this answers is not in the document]
afourniernv: Addressed in 8e0329173. `dependencies.yaml` is now the source of truth for the elastic extra again and matches `python/cuvs_bench/pyproject.toml` by depending on `elasticsearch>=8.0` directly, consistent with the in-tree optional-backend model.

python/cuvs_bench/cuvs_bench/backends/elasticsearch.py
```diff
@@ -0,0 +1,921 @@
...
+_DEFAULT_NUM_REPLICAS = 0
+
+_DEFAULT_VECTOR_FIELD = "embedding"
+_DEFAULT_NUM_CANDIDATES = 100
+
+_SUPPORTED_INDEX_TYPES = ("hnsw", "int8_hnsw", "int4_hnsw", "bbq_hnsw")
+_SUPPORTED_SIMILARITIES = ("l2_norm", "cosine", "max_inner_product")
+_SUPPORTED_ALGOS = tuple(
+    f"elastic_{index_type}" for index_type in _SUPPORTED_INDEX_TYPES
+)
```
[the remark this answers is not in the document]
afourniernv: Addressed in da8aafac5. The elastic loader now accepts the supported variant names (`elastic_int8_hnsw`, `elastic_int4_hnsw`, `elastic_bbq_hnsw`) by resolving them through the shared `elastic_hnsw` config and preserving the requested variant when building the `IndexConfig` / backend config, so `build_param["type"]` is set consistently without duplicating YAML entries. I also added a focused loader test for the `elastic_int8_hnsw` path. I could not run that pytest slice locally here because `pytest` is not installed in the available Python environment.

python/cuvs_bench/cuvs_bench/backends/registry.py
```diff
@@ -375,10 +412,41 @@ def get_backend(name: str, config: Dict) -> BenchmarkBackend:
...
+            eps = [e for e in eps if e.name == name]
+        for ep in eps:
+            try:
+                ep.load()()
+            except ImportError as e:
+                rewritten = _rewrite_optional_backend_import_error(name, e)
+                if rewritten is not None:
+                    raise rewritten from e
+                raise
+            return  # Plugin loaded successfully
```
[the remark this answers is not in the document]
afourniernv: Addressed in 3f80d6b3f. `_try_load_plugin()` no longer returns after the first successful entry-point load, so a same-named plugin can register pieces from both `cuvs_bench.backends` and `cuvs_bench.config_loaders` before lookup completes. I also added a regression test covering a split plugin where the backend and config loader are registered from different entry-point groups. I could not run the pytest slice locally here because `pytest` is not installed in the available Python environment.

python/cuvs_bench/cuvs_bench/backends/elasticsearch.py
afourniernv: Addressed in e30bdd158. Elastic no longer computes recall inside `search()` or marks it with `recall_is_authoritative`; it now returns the top-level neighbor matrix and lets the orchestrator compute recall via the shared `compute_recall()` path, matching the current OpenSearch ownership model. I kept this scoped to the Elastic backend. Full per-search-param recall would require a shared result-shape change because Elastic/OpenSearch currently both aggregate timing across search-param combinations while exposing top-level neighbors for the last combination.

python/cuvs_bench/cuvs_bench/backends/elasticsearch.py
afourniernv: Yes, that is the reason. `initialize()` intentionally does not create the Elasticsearch client because the orchestrator calls `initialize()` even for `dry_run`; keeping connection setup inside `_get_client()` lets `build()` / `search()` return early on dry runs without requiring a live cluster. Real build/search paths still call `_check_network_available()` and then create the client lazily before issuing requests.

python/cuvs_bench/cuvs_bench/backends/elasticsearch.py
afourniernv: Follow-up addressed in df0cb52ef. The stale tests left behind by removing `_load_fbin` / `_load_ibin` are cleaned up: the lazy `Dataset` loading tests now patch the shared `cuvs_bench.backends._utils.load_vectors` seam, and the obsolete Elastic-wrapper loader tests are removed. The shared binary loader coverage remains in `test_utils.py`. Validated no stale references remain and `py_compile` passes; the focused pytest slice could not run here because `pytest` is not installed in the available Python environment.

python/cuvs_bench/cuvs_bench/backends/elasticsearch.py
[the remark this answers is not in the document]
afourniernv: Follow-up in 30d85e694 after running pytest locally. The variant algorithm path now overrides the shared `elastic_hnsw` YAML `type` when the requested algorithm is a variant like `elastic_int8_hnsw`, so `IndexConfig.build_param["type"]` and `backend_config["type"]` both resolve to `int8_hnsw` as intended. `test_modularization.py` now passes locally.

python/cuvs_bench/cuvs_bench/backends/elasticsearch.py
```diff
@@ -0,0 +1,816 @@
...
+_BUILD_PARAM_KEYS = (
+    "type",
+    "m",
+    "ef_construction",
+    "similarity",
+    "number_of_shards",
+    "number_of_replicas",
+    "vector_field",
+)
+_SEARCH_PARAM_KEYS = ("num_candidates", "vector_field")
```
jnke2016: This is not referenced anywhere anymore I think so it can be removed

python/cuvs_bench/cuvs_bench/backends/elasticsearch.py
```diff
@@ -0,0 +1,816 @@
...
+
+    if cuvs_bench.__file__:
+        return os.path.join(os.path.dirname(cuvs_bench.__file__), "config")
+    raise RuntimeError(
+        "Cannot determine cuvs_bench config path. "
+        "Ensure cuvs_bench is properly installed or on PYTHONPATH."
+    )
+
+
+def _get_elastic_config_path() -> str:
```
jnke2016: This is not referenced anywhere I think so it can be removed

python/cuvs_bench/cuvs_bench/orchestrator/orchestrator.py
```diff
@@ -24,6 +24,15 @@
 from .config_loaders import DatasetConfig
 
 
+def _should_compute_recall(result: SearchResult) -> bool:
```
jnke2016, line 27: Now that recall is computed by the orchestrator for every backend, no backend sets recall_is_authoritative, so the flag branch in _should_compute_recall is only exercised by a unit test (`test_modularization.py`), not by any real backend. Now that recall is computed by the orchestrator for every backend, the flag no longer has a caller. You may want to simplify _should_compute_recall to success and neighbors.size > 0 (the empty-neighbors check already covers the C++ subprocess case) and drop the corresponding test.

jnke2016 (approved):
Thanks @afourniernv  for addressing the feedback, the PR looks good to me. Just a few minor cleanup requests

python/cuvs_bench/cuvs_bench/backends/elasticsearch.py
afourniernv: Addressed in 4df7c9c41. Removed the now-unused Elastic parameter-key constants; the backend now takes build/search params directly from the config-loader/index config flow.

python/cuvs_bench/cuvs_bench/backends/elasticsearch.py
afourniernv: Addressed in 4df7c9c41. Removed the unused `_get_elastic_config_path()` helper; `ElasticConfigLoader` uses `_get_cuvs_bench_config_path()` directly.

python/cuvs_bench/cuvs_bench/orchestrator/orchestrator.py
afourniernv: Addressed in 4df7c9c41. `_should_compute_recall()` now only checks `result.success` and `result.neighbors.size > 0`, and the obsolete authoritative-recall unit test was removed. No backend uses or sets `recall_is_authoritative` anymore.

afourniernv:
Hi @cjnolet, this PR has been caught up with latest `main` and the latest review comments have been addressed. Could you take a look and help with final review/merge. Requested per @jnke2016

python/cuvs_bench/cuvs_bench/backends/elasticsearch.py
```diff
@@ -0,0 +1,800 @@
...
+                success=False,
+                error_message=str(e),
+            )
+
+    def search(
+        self,
+        dataset: Dataset,
+        indexes: List[IndexConfig],
+        k: int = 10,
+        batch_size: int = 10000,
```
jrbourbeau, line 380: It looks like this search `batch_size` parameter isn't being used and always uses single vector search

python/cuvs_bench/cuvs_bench/backends/elasticsearch.py
```diff
@@ -0,0 +1,800 @@
...
+    from cuvs_bench.backends.registry import (
+        _CONFIG_LOADER_REGISTRY,
+        get_registry,
+    )
+
+    reg = get_registry()
+    if not reg.is_registered("elastic"):
+        register_backend("elastic", ElasticBackend)
+    if "elastic" not in _CONFIG_LOADER_REGISTRY:
+        register_config_loader("elastic", ElasticConfigLoader)
```
jrbourbeau, line 799: There's a lot of machinery around entrypoints/registering the elasticsearch backend. What benefit is there to this vs. just raising an informative error when we attempt to use the backend?

python/cuvs_bench/cuvs_bench/backends/registry.py
```diff
@@ -390,10 +470,13 @@ def get_backend_class(name: str) -> Type[BenchmarkBackend]:
         Backend class
     """
     registry = get_registry()
+    if name not in registry._backends:
+        _try_load_plugin(name)
```
jrbourbeau, line 471: Does this break the opensearch backend given it's not registered as an entrypoint?

jrbourbeau (commented):
Thanks @afourniernv

python/cuvs_bench/cuvs_bench/config/algos/elasticsearch.yml
```diff
@@ -0,0 +1,52 @@
...
+#
+# Elasticsearch GPU HNSW: index_options (build) and knn (search)
+# Per ES-GPU-API-REFERENCE.md and DOCKER-INSPECT-FINDINGS.md
+#
+# Build (index_options): type, m, ef_construction
+#   type: hnsw (float), int8_hnsw (4x quantized), int4_hnsw (8x), bbq_hnsw (32x)
+#   similarity: l2_norm, cosine, max_inner_product (overrides dataset distance)
+# Index settings: number_of_shards, number_of_replicas, use_gpu, vector_field
+# Search (knn): num_candidates, vector_field
+name: elastic_hnsw
```
cjnolet, line 12: Nitpick- can we please call this file elasticsearch.yml. It would be good to keep the expected yaml file the same name as the backend name (and the same as the backend name)

python/cuvs_bench/cuvs_bench/backends/elasticsearch.py
```diff
@@ -0,0 +1,800 @@
...
+_DEFAULT_NUM_CANDIDATES = 100
+
+_SUPPORTED_INDEX_TYPES = ("hnsw", "int8_hnsw", "int4_hnsw", "bbq_hnsw")
+_SUPPORTED_SIMILARITIES = ("l2_norm", "cosine", "max_inner_product")
+_SUPPORTED_ALGOS = tuple(
+    f"elastic_{index_type}" for index_type in _SUPPORTED_INDEX_TYPES
+)
+
+
+class ElasticBackend(BenchmarkBackend):
```
cjnolet: Can we call this "ElasticsearchBackend"? Elastic is a company. Just want to make things consistnet and clear.

python/cuvs_bench/cuvs_bench/config/algos/elasticsearch.yml
afourniernv: Addressed in 16060fc4c. Renamed the benchmark config file to python/cuvs_bench/cuvs_bench/config/algos/elasticsearch.yml.

python/cuvs_bench/cuvs_bench/backends/elasticsearch.py
afourniernv: Addressed in 9a065a5ee. Renamed ElasticBackend to ElasticsearchBackend and updated registration and test references.

KyleFromNVIDIA (approved):

afourniernv:
Caught up with latest `main` at `4e2a2f11c`; the full updated pre-commit suite passes locally. @cjnolet, could you please `/ok to test 4e2a2f1` and `/merge` once the replacement CI run is green? The prior CI failure was caused by a cancelled wheel-test job.

cjnolet (approved):

Issue #1767, opened 2026-02-03
[FEA] Enable code rabbit for AI code reviews


cjnolet:

cjnolet:
Linking @csadorf's PR for enabling code rabbit in cuML [here](https://github.com/rapidsai/cuml/pull/7725).

cjnolet:
Until the PR is merged to enable [C ABI breakage checks](https://github.com/rapidsai/cuvs/pull/1749), we should disable the code rabbit from checking anything in the C APIs.

PR #1908, opened 2026-03-10
Coderabbit integration

benfred:
Following on the work of https://github.com/rapidsai/rmm/pull/2244 and https://github.com/rapidsai/cuml/pull/7725, this enables general instructions and configuration to enable coderabbit codereviews on cuvs.

Closes https://github.com/rapidsai/cuvs/issues/1767


benfred:
fwiw, I found this paper about the effectiveness of AGENTS.md files somewhat interesting (and possibly relevant to this PR) https://arxiv.org/pdf/2602.11988

cpp/agents.md
```diff
@@ -0,0 +1,311 @@
...
+- Deadlock potential in resource acquisition
+- Thread-unsafe use of global/static variables
+- **Concurrent operations sharing streams incorrectly** (multi-GPU without proper isolation)
+- **Stream reuse across independent operations** (causing unwanted serialization or race conditions)
+
+### Design & Architecture
+- Hard-coded GPU device IDs or resource limits
+- Inappropriate use of exceptions in performance-critical paths
+- Significant code duplication (3+ occurrences). Including in kernel logic.
+- Reinventing functionality already available in RAFT, RMM,  libcudacxx, thrust, or CUB
```
bdice:
```suggestion
- Reinventing functionality already available in RAFT, RMM, libcudacxx, thrust, or CUB
```

cpp/agents.md
```diff
@@ -0,0 +1,311 @@
...
+Why: Can produce Inf/NaN values corrupting results
+Consider: Add epsilon threshold check or use safe division helper
+```
+
+**HIGH** (performance issue):
+```
+HIGH: Unnecessary synchronization in hot path
+
+Issue: cudaDeviceSynchronize() or raft::resource::sync_stream() inside iteration loop
+Why: Blocks GPU pipeline, 10x slowdown on benchmarks
```
bdice: Just as a heads-up: someday Coderabbit will quote "this is 10x slower because..." and you'll wonder where the value came from. I suggest removing hyperspecific or case-dependent details like this, they tend to be cited at unexpected times and may confuse the person reading the agent's output.

python/REVIEW_GUIDELINES.md
bdice, line 1: I named this file (and the C++ one) `{cpp,python}/REVIEW_GUIDELINES.md` and set up the CodeRabbit configuration to reference those files.

If you name it `AGENTS.md`, it will be automatically loaded by _all_ agents. I did not want these review rules to affect all agents, particularly if they aren't being used for code review. I think this is too much context for _generic_ agents and would recommend only including it for reviews.

Several folks on RMM/cuDF have been hesitant to use `AGENTS.md` because of its tendency to overload agent context without providing sufficient value. The alternative is to use skills like this one: https://github.com/rapidsai/cudf/blob/main/.agents/skills/review-cudf/SKILL.md Skills are loaded dynamically/contextually, only when deemed relevant by the agent. That'll save valuable context space.

cpp/agents.md
aamijar: Addressed in [4055e06](https://github.com/rapidsai/cuvs/pull/1908/commits/4055e067dd0964480e10a37e4ba76e3a3e8cc1f2)

python/REVIEW_GUIDELINES.md
aamijar: Addressed in [75f936a](https://github.com/rapidsai/cuvs/pull/1908/commits/75f936a316e61d8c487608a0e4f56ef20ff8c4fb)

PR #1910, opened 2026-03-11
IVF-SQ C API

viclafargue:
Depends on #1865. Do not merge.


tfeher (approved):
LGTM.

c/include/cuvs/neighbors/ivf_sq.h
```diff
@@ -0,0 +1,364 @@
...
+ *
+ * @param[in] res cuvsResources_t opaque C handle
+ * @param[in] search_params cuvsIvfSqSearchParams_t used to search IVF-SQ index
+ * @param[in] index ivfSqIndex which has been returned by `cuvsIvfSqBuild`
+ * @param[in] queries DLManagedTensor* queries dataset to search
+ * @param[out] neighbors DLManagedTensor* output `k` neighbors for queries
+ * @param[out] distances DLManagedTensor* output `k` distances for queries
+ * @param[in] filter cuvsFilter to filter neighbors based on the given bitset
+ */
+CUVS_EXPORT cuvsError_t cuvsIvfSqSearchWithFilter(cuvsResources_t res,
```
lowener: IVF-Flat and IVF-PQ only use `cuvsIvf*Search`, without a separate `*WithFilter` function, can the same be done here?

c/tests/neighbors/run_ivf_sq_c.c
```diff
@@ -0,0 +1,86 @@
...
+  DLManagedTensor queries_tensor;
+  queries_tensor.dl_tensor.data              = (void*)query_data;
+  queries_tensor.dl_tensor.device.device_type = kDLCUDA;
+  queries_tensor.dl_tensor.ndim              = 2;
+  queries_tensor.dl_tensor.dtype.code        = kDLFloat;
+  queries_tensor.dl_tensor.dtype.bits        = 32;
+  queries_tensor.dl_tensor.dtype.lanes       = 1;
+  int64_t queries_shape[2]                   = {n_queries, n_dim};
+  queries_tensor.dl_tensor.shape             = queries_shape;
+  queries_tensor.dl_tensor.strides           = NULL;
```
lowener: `#include "../../src/core/interop.hpp"` and use `cuvs::core::to_dlpack` to simplify the interaction with DLManagedTensor

c/tests/neighbors/run_ivf_sq_c.c
lowener, line 1: Merge run_ivf_sq_c and ann_ivf_sq_c into the same file.
The total line number won't be so big once `to_dlpack` is used

c/tests/neighbors/ann_ivf_sq_c.cu
```diff
@@ -0,0 +1,130 @@
...
+                           float* index_data,
+                           float* query_data,
+                           float* distances_data,
+                           int64_t* neighbors_data,
+                           cuvsDistanceType metric,
+                           size_t n_probes,
+                           size_t n_lists);
+
+template <typename T>
+void generate_random_data(T* devPtr, size_t size)
```
lowener: Take handle as param instead of building a new one

c/tests/neighbors/ann_ivf_sq_c.cu
```diff
@@ -0,0 +1,130 @@
...
+
+  rmm::device_uvector<float> index_data(n_rows * n_dim, stream);
+  rmm::device_uvector<float> query_data(n_queries * n_dim, stream);
+  rmm::device_uvector<int64_t> neighbors_data(n_queries * n_neighbors, stream);
+  rmm::device_uvector<float> distances_data(n_queries * n_neighbors, stream);
+
+  generate_random_data(index_data.data(), n_rows * n_dim);
+  generate_random_data(query_data.data(), n_queries * n_dim);
+
+  run_ivf_sq(n_rows,
```
lowener: Use the same handle in that function, otherwise without sync there are no guarantee that the device data was initialized

c/tests/neighbors/ann_ivf_sq_c.cu
```diff
@@ -0,0 +1,130 @@
...
+template <typename T>
+void generate_random_data(T* devPtr, size_t size)
+{
+  raft::handle_t handle;
+  raft::random::RngState r(1234ULL);
+  raft::random::uniform(handle, r, devPtr, size, T(0.1), T(2.0));
+};
+
+template <typename T, typename IdxT>
+void recall_eval(T* query_data,
```
lowener: Re-use the same handle

lowener (changes requested):

c/tests/neighbors/run_ivf_sq_c.c
viclafargue: I was about to make the change, but I realized that `cuvs::core::to_dlpack` is a C++ utility and part of the test is to ensure that everything build successfully with a pure C compiler. Even though this adds a lot of lines, it should probably still remain this way. At least other C tests are built this way.

c/tests/neighbors/run_ivf_sq_c.c
viclafargue: `run_ivf_pq_c.c` contains the `run_ivf_pq` launcher and should compile successfully in C. While `ann_ivf_sq_c.cu` is the Gtest orchestration. Other C tests follow this pattern.

c/tests/neighbors/ann_ivf_sq_c.cu
viclafargue: Since `run_ivf_sq` is a C function, I added the `res` argument to it. This should solve any syncing issue.

lowener (approved):

PR #1912, opened 2026-03-11
Add Elasticsearch GPU deployment example

afourniernv:
Add examples/elasticsearch/ with Makefile and README for deploying Elasticsearch GPU + cuVS on a Linux host with NVIDIA GPU (e.g. Brev). Uses basic license (no trial); patch-gpu-license bypasses Enterprise check for GPU indexing.


cjnolet:
I admit this is the first of its kind in the codebase and it's not immediately obvious to me where it belongs. Examples usually contains code examples, but in this case we're talking about a build example. Need to think about this one a little bit.

afourniernv:
> I admit this is the first of its kind in the codebase and it's not immediately obvious to me where it belongs. Examples usually contains code examples, but in this case we're talking about a build example. Need to think about this one a little bit.

@cjnolet  this feels awkward to me personally to slot into this library.. Mana strongly requested it so I slotted it in here... but it feels like it should be in our docs? Happy to close if you want.

PR #1913, opened 2026-03-12
IVF-SQ Python API

viclafargue:
Depends on #1910. Do not merge.


python/cuvs/cuvs/neighbors/ivf_sq/ivf_sq.pyx
```diff
@@ -0,0 +1,530 @@
...
+        return n_lists
+
+    @property
+    def dim(self):
+        """ dimensionality of the cluster centers """
+        cdef int64_t dim = 0
+        cuvsIvfSqIndexGetDim(self.index, &dim)
+        return dim
+
+    def __len__(self):
```
dantegd, line 184: Quick question: should n_lists, dim, and __len__ raise on an untrained index the same way centers does just below? Right now they go straight into the C wrapper, which does reinterpret_cast<index<uint8_t>*>(addr) with addr == 0 for a fresh Index(), so we'd segfault instead of getting a nice ValueError.

python/cuvs/cuvs/neighbors/ivf_sq/ivf_sq.pyx
```diff
@@ -0,0 +1,530 @@
...
+        return self.trained
+
+    def __repr__(self):
+        return "Index(type=IvfSq)"
+
+    @property
+    def n_lists(self):
+        """ The number of inverted lists (clusters) """
+        cdef int64_t n_lists = 0
+        cuvsIvfSqIndexGetNLists(self.index, &n_lists)
```
dantegd: Tiny consistency nit:  __len__ and centers wrap their C calls in check_cuvs(...), but n_lists here and dim below don't. Mind wrapping these two too so any failure surfaces as a Python exception instead of silently returning 0?

python/cuvs/cuvs/neighbors/ivf_sq/ivf_sq.pyx
```diff
@@ -0,0 +1,530 @@
...
+            neighbors_dlpack,
+            distances_dlpack,
+            filter.prefilter
+        ))
+
+    return (distances, neighbors)
+
+
+@auto_sync_resources
+def save(filename, Index index, bool include_dataset=True, resources=None):
```
dantegd: include_dataset is in the signature but never reaches cuvsIvfSqSerialize (and isn't in the docstring either), so include_dataset=False silently does the same thing as True. I know ivf_flat.save has the same issue, but since this is a brand-new file, we could either drop the kwarg or wire it through?  Even just adding it to the docstring with a "currently a no-op" note would be nice.

python/cuvs/cuvs/neighbors/ivf_sq/ivf_sq.pyx
```diff
@@ -0,0 +1,530 @@
...
+    uint8_t,
+    uint32_t,
+    uint64_t,
+    uintptr_t,
+)
+
+from cuvs.common.exceptions import check_cuvs
+
+
+cdef class IndexParams:
```
dantegd, line 41: none of the parameters here are actually scalar-quantization specific, the shape is identical to ivf_flat.IndexParams. Does the underlying C++ ivf_sq::index_params expose any SQ knobs (bit-width, quantizer range/scaling, residual vs. raw, etc.)? If so, surfacing them here gives users the actual SQ tuning surface; if not, a one-liner in the docstring saying so would be nice

python/cuvs/cuvs/neighbors/ivf_sq/ivf_sq.pyx
```diff
@@ -0,0 +1,530 @@
...
+    """
+
+    def __cinit__(self):
+        cuvsIvfSqIndexParamsCreate(&self.params)
+
+    def __dealloc__(self):
+        if self.params != NULL:
+            check_cuvs(cuvsIvfSqIndexParamsDestroy(self.params))
+
+    def __init__(self, *,
```
dantegd, line 94: Cross-IVF consistency thingL ivf_flat and ivf_pq both use kmeans_trainset_fraction (float in [0,1], default 0.5) for sizing the training set, whereas this uses max_train_points_per_cluster (int, default 256). Is that because the C++ ivf_sq::index_params only has the per-cluster knob, or could we also expose kmeans_trainset_fraction for parity?

python/cuvs/cuvs/neighbors/ivf_sq/ivf_sq.pyx
```diff
@@ -0,0 +1,530 @@
...
+                 n_lists=1024,
+                 metric="sqeuclidean",
+                 metric_arg=2.0,
```
dantegd, line 97: metric_arg is accepted here and exposed as a property, but doesn't appear in the docstring's Parameters table. Mind adding?

python/cuvs/cuvs/neighbors/ivf_sq/ivf_sq.pyx
```diff
@@ -0,0 +1,530 @@
...
+
+    kmeans_n_iters : int, default = 20
+        The number of iterations searching for kmeans centers during index
+        building.
+    max_train_points_per_cluster : int, default = 256
+        The number of data vectors per cluster to use during iterative
+        kmeans building. The index uses at most
+        n_lists * max_train_points_per_cluster rows for training.
+    add_data_on_build : bool, default = True
+        After training the coarse and fine quantizers, we will populate
```
dantegd: This sentence got carried over from ivf_flat/ivf_pq, but IVF-SQ doesn't really have "coarse and fine quantizers" in the IVF-PQ sense, right? It's the coarse quantizer plus a scalar quantizer on the residuals. Worth rewording to reflect what SQ actually does so the user-facing description is accurate?

python/cuvs/cuvs/neighbors/ivf_sq/ivf_sq.pyx
```diff
@@ -0,0 +1,530 @@
...
+
+@auto_sync_resources
+def build(IndexParams index_params, dataset, resources=None):
+    """
+    Build the IvfSq index from the dataset for efficient search.
+
+    IVF-SQ (Scalar Quantization) combines an IVF coarse quantizer with
+    per-dimension scalar quantization. Each vector's residual is encoded
+    as one byte per dimension, providing ~4x memory reduction vs IVF-Flat
+    with higher recall than IVF-PQ at similar memory budgets.
```
dantegd: isn't "higher recall than IVF-PQ at similar memory budgets" data and tuning-dependent?  Might be good in as "typically higher recall than IVF-PQ at similar memory budgets, depending on the dataset" or similar. Also, "~4x memory reduction vs IVF-Flat" is that for float32 inputs? That would make it ~2x for float16, and 8x for float64 (eventually), no?

python/cuvs/cuvs/neighbors/ivf_sq/ivf_sq.pxd
```diff
@@ -0,0 +1,101 @@
...
+
+    cuvsError_t cuvsIvfSqIndexGetCenters(cuvsIvfSqIndex_t index,
+                                         DLManagedTensor * centers)
+
+    cuvsError_t cuvsIvfSqBuild(cuvsResources_t res,
+                               cuvsIvfSqIndexParams* params,
+                               DLManagedTensor* dataset,
+                               cuvsIvfSqIndex_t index) except +
+
+    cuvsError_t cuvsIvfSqSearch(cuvsResources_t res,
```
dantegd, line 68: is this called in the python code? I only saw cuvsIvfSqSearchWithFilter, could we drop these two?

python/cuvs/cuvs/neighbors/ivf_sq/ivf_sq.pxd
```diff
@@ -0,0 +1,101 @@
...
+                                   cuvsIvfSqIndex_t index) except +
+
+    cuvsError_t cuvsIvfSqDeserialize(cuvsResources_t res,
+                                     const char * filename,
+                                     cuvsIvfSqIndex_t index) except +
+
+    cuvsError_t cuvsIvfSqExtend(cuvsResources_t res,
+                                DLManagedTensor* new_vectors,
+                                DLManagedTensor* new_indices,
+                                cuvsIvfSqIndex_t index)
```
dantegd:
```suggestion
                                cuvsIvfSqIndex_t index) except + 
```

python/cuvs/cuvs/tests/test_ivf_sq.py
```diff
@@ -0,0 +1,143 @@
...
+    recall = calc_recall(out_idx, skl_idx)
+    assert recall > 0.7
+
+    centers = index.centers
+    assert centers.shape[0] == build_params.n_lists
+    assert centers.shape[1] == n_cols
+
+
+@pytest.mark.parametrize("inplace", [True, False])
+@pytest.mark.parametrize("dtype", [np.float32])
```
dantegd: we should add np.flaot16 too here, no?

python/cuvs/cuvs/tests/test_ivf_sq.py
```diff
@@ -0,0 +1,143 @@
...
+    nn_skl = NearestNeighbors(
+        n_neighbors=k, algorithm="brute", metric=skl_metric
+    )
+    nn_skl.fit(dataset)
+    skl_idx = nn_skl.kneighbors(queries, return_distance=False)
+
+    recall = calc_recall(out_idx, skl_idx)
+    assert recall > 0.7
+
+    centers = index.centers
```
dantegd, line 131: We could add a few more cheap assertions here for some additional testing:

```python
assert len(index) == n_rows
assert index.dim == n_cols
assert index.n_lists == build_params.n_lists
```

python/cuvs/cuvs/tests/test_ivf_sq.py
```diff
@@ -0,0 +1,143 @@
...
+)
+def test_ivf_sq(inplace, dtype, metric):
+    run_ivf_sq_build_search_test(
+        dtype=dtype,
+        inplace=inplace,
+        metric=metric,
+    )
+
+
+@pytest.mark.parametrize("dtype", [np.float32, np.float16])
```
dantegd, line 149: Couple things/testing gaps: 

1. Save/load is currently only tested inside test_extend(serialize=True), which mixes two concerns. A small standalone test_serialization (build → save → load → assert n_lists/dim/centers match) would isolate the path and make failures easier to diagnose.
2. run_ivf_sq_build_search_test only exercises extend when add_data_on_build=False. The add_data_on_build=True + later extend() path hits a different branch in cuvsIvfSqExtend (where index->dtype is already set vs. the empty-index path), so worth covering too. 

What do you think?

dantegd (changes requested):

python/cuvs/cuvs/neighbors/ivf_sq/ivf_sq.pyx
viclafargue: Good catch. I dropped the argument.

python/cuvs/cuvs/neighbors/ivf_sq/ivf_sq.pyx
viclafargue: IVF-SQ only support 8-bit quantization and does not have any additional SQ-specific tuning knobs. I added a small note about this in the documentation.

python/cuvs/cuvs/neighbors/ivf_sq/ivf_sq.pyx
viclafargue: [@tfeher recommended replacing `kmeans_trainset_fraction` by `max_train_points_per_cluster`](https://github.com/rapidsai/cuvs/pull/1865#discussion_r3248076412) because it is easier for scaling the number of clusters during benchmark and because it aligns with other work.

python/cuvs/cuvs/neighbors/ivf_sq/ivf_sq.pyx
viclafargue: Thanks for spotting this. This should indeed be updated to better reflect what IVF-SQ is doing. I replaced it with "After training the coarse clustering model and residual scalar quantization parameters".

python/cuvs/cuvs/neighbors/ivf_sq/ivf_sq.pxd
viclafargue: Following the most recent review of the C API, I dropped the `cuvsIvfSqSearchWithFilter` function and added a filter parameter to the `cuvsIvfSqSearch` function. This aligns with other indexes.

python/cuvs/cuvs/tests/test_ivf_sq.py
viclafargue: You're right. Updated accordingly.

lowener (approved):
LGTM

dantegd (approved):

PR #1915, opened 2026-03-13
Add KDE kernel

Intron7:
@cjnolet moved the kernel for KDE from cuml to cuvs. This PR is needed for https://github.com/rapidsai/cuml/pull/7833


cpp/include/cuvs/distance/kde.hpp
```diff
@@ -0,0 +1,103 @@
...
+ * @param[in]  n_features  Dimensionality of the data
+ * @param[in]  bandwidth   Kernel bandwidth (must be > 0)
+ * @param[in]  sum_weights Sum of sample weights (or n_train if uniform)
+ * @param[in]  kernel      Density kernel function
+ * @param[in]  metric      Distance metric
+ * @param[in]  metric_arg  Metric parameter (e.g. p for Minkowski; ignored otherwise)
+ */
+template <typename T>
+void kde_score_samples(raft::resources const& handle,
+                       const T* query,
```
cjnolet: Can we accept mdspans here instead of raw pointers? That would also simplify the function signatures for the public APIs. Can still expose pointers at the cuml layer for now (for the python layer) if that's easier.

cpp/include/cuvs/distance/kde.hpp
```diff
@@ -0,0 +1,103 @@
...
+ * @param[in]  n_train     Number of training points
+ * @param[in]  n_features  Dimensionality of the data
+ * @param[in]  bandwidth   Kernel bandwidth (must be > 0)
+ * @param[in]  sum_weights Sum of sample weights (or n_train if uniform)
+ * @param[in]  kernel      Density kernel function
+ * @param[in]  metric      Distance metric
+ * @param[in]  metric_arg  Metric parameter (e.g. p for Minkowski; ignored otherwise)
+ */
+template <typename T>
+void kde_score_samples(raft::resources const& handle,
```
cjnolet: Would it make sense just call this "kde" or is there signficant enough post processing being done after calling this?

cpp/include/cuvs/distance/kde.hpp
```diff
@@ -0,0 +1,103 @@
...
+
+namespace cuvs::distance {
+
+/**
+ * @brief Density kernel type for Kernel Density Estimation.
+ *
+ * These are the smoothing kernels used in KDE — distinct from the dot-product
+ * kernels (RBF, Polynomial, etc.) in cuvs::distance::kernels used by SVMs.
+ */
+enum class DensityKernelType : int {
```
cjnolet: Can we at least consolidate this struct w/ the existing `KernelType` struct in distance.hpp?

aamijar:
@Intron7, thanks for the PR!
Please run `pre-commit` to enforce style checks.

cpp/src/distance/kde.cu
```diff
@@ -0,0 +1,675 @@
...
+         std::optional<raft::device_vector_view<const T, std::int64_t>> weights,
+         raft::device_vector_view<T, std::int64_t> output,
+         T bandwidth,
+         T sum_weights,
+         DensityKernelType kernel,
+         cuvs::distance::DistanceType metric,
+         T metric_arg)
+{
+  int n_query = static_cast<int>(query.extent(0));
+  int n_train = static_cast<int>(train.extent(0));
```
viclafargue, line 547: The public API uses `int64_t` extent, but the `n_query` and `n_train` are casted to `int`. Let's check that `n_query`, `n_train`, and `d` fit in `int32`.
```suggestion
  RAFT_EXPECTS(query.extent(0) <= std::numeric_limits<int>::max() &&
                             train.extent(0) <= std::numeric_limits<int>::max() &&
                             query.extent(1) <= std::numeric_limits<int>::max(),
                             "n_query, n_train, and n_features must fit in int32");
  int n_query = static_cast<int>(query.extent(0));
  int n_train = static_cast<int>(train.extent(0));
```

cpp/src/distance/kde.cu
```diff
@@ -0,0 +1,675 @@
...
+// ============================================================================
+
+template <typename T>
+void kde(raft::resources const& handle,
+         raft::device_matrix_view<const T, std::int64_t, raft::layout_c_contiguous> query,
+         raft::device_matrix_view<const T, std::int64_t, raft::layout_c_contiguous> train,
+         std::optional<raft::device_vector_view<const T, std::int64_t>> weights,
+         raft::device_vector_view<T, std::int64_t> output,
+         T bandwidth,
+         T sum_weights,
```
viclafargue, line 537: Is there a reason `sum_weights` is caller-provided rather than computed internally from weights? A mismatch silently produces incorrectly scaled output, and a device reduction would eliminate that risk.

cjnolet:
Tests look good, thanks so much @Intron7. It looks like CI just complaining on the branch being too old. If you can update the branch with upstream main, we can get this merged.

cpp/tests/distance/kde.cu
```diff
@@ -0,0 +1,731 @@
...
+    4.79228e-01,  3.85546e-01,  1.35226e-01};
+
+const double golden_weights[] = {
+    5.138e-01, 2.5387e+00, 2.2671e+00, 2.3225e+00,
+    2.4282e+00, 6.851e-01, 1.3962e+00, 7.897e-01};
+// clang-format on
+
+constexpr int N_QUERY  = 4;
+constexpr int N_TRAIN  = 8;
+constexpr int D        = 3;
```
viclafargue, line 76: Feature tiling loop (D <= FEAT_TILE), train tiling loop (N_TRAIN <= CELL_TILE) are untested.

viclafargue (approved):
Thanks! LGTM

PR #1923, opened 2026-03-16
Always build with JIT+LTO

KyleFromNVIDIA:
Since https://github.com/rapidsai/cuvs/pull/1909, we've been able to use older versions of the CUDA driver, since we no longer rely on `cudaLibraryEnumerateKernels()`. Since https://github.com/rapidsai/cuvs/pull/1918, we've been using static cudart, which allows us to run on platforms with versions of CUDA older than 12.8 installed, since the runtime library API is now bundled with cuvs. Always build with JIT+LTO so that we can get the full compile time and binary size benefits in CUDA 12 too.


cpp/CMakeLists.txt
```diff
@@ -355,12 +355,8 @@ if(NOT BUILD_CPU_ONLY)
...
   endif()
 
-  set(JIT_LTO_TARGET_ARCHITECTURE "")
-  set(JIT_LTO_COMPILATION OFF)
-  if(CMAKE_CUDA_COMPILER_VERSION VERSION_GREATER_EQUAL 13.0)
-    set(JIT_LTO_TARGET_ARCHITECTURE "75-real")
-    set(JIT_LTO_COMPILATION ON)
-  endif()
+  set(JIT_LTO_TARGET_ARCHITECTURE "75-real")
+  set(JIT_LTO_COMPILATION ON)
```
KyleFromNVIDIA: This is just a first pass to ensure that JIT+LTO works on CUDA 12. Once that's done, I will go back and remove the cruft that was needed to maintain the two different code paths.

KyleFromNVIDIA:
It seems that even after https://github.com/rapidsai/cuvs/pull/1918, we're still [not using](https://github.com/rapidsai/cuvs/actions/runs/23163936077/job/67309072078?pr=1923#step:12:1054) `cudart_static`, since rmm forces us to use the shared version:

```
./../../..//bin/gtests/libcuvs/STATS_TEST: symbol lookup error: /opt/conda/envs/test/bin/gtests/libcuvs/../../../lib/libcuvs.so: undefined symbol: cudaLibraryGetKernel, version libcudart.so.12
```

We have to get rmm on `cudart_static`.

KyleFromNVIDIA:
We've decided to switch to the driver API instead, since rmm is blocked on https://github.com/rapidsai/cudf/pull/20814, which in turn is also blocked.

cpp/src/detail/jit_lto/AlgorithmPlanner.cpp
```diff
@@ -106,12 +108,14 @@ std::shared_ptr<AlgorithmLauncher> AlgorithmPlanner::build()
...
   // cubin is linked, so now load it
-  cudaLibrary_t library;
-  RAFT_CUDA_TRY(
-    cudaLibraryLoadData(&library, cubin.get(), nullptr, nullptr, 0, nullptr, nullptr, 0));
+  CUlibrary library;
+  CU_TRY(cuLibraryLoadData(&library, cubin.get(), nullptr, nullptr, 0, nullptr, nullptr, 0));
 
-  cudaKernel_t kernel;
-  RAFT_CUDA_TRY(cudaLibraryGetKernel(&kernel, library, this->entrypoint.c_str()));
+  CUkernel kernel;
```
divyegala: Is `CUkernel` not launchable? I would assume you could use the cudart API everywhere else (like `AlgorithmLauncher`) because `CUkernel` and `CUlibrary` should be the same types as `cudaKernel_t` and `cudaLibrary_t`.

dependencies.yaml
```diff
@@ -326,16 +326,7 @@ dependencies:
...
-      - output_types: conda
-        matrices:
-          - matrix:
-              cuda: "13.*"
-            packages:
-              - libnvjitlink-dev
-          - matrix:
-              cuda: "12.*"
-            packages:
+          - libnvjitlink-dev
```
divyegala, line 376: Add stub package `cuda-driver-dev` and ignore its `run_export` please (we'll pick the system `cuda-driver`)

cpp/src/detail/jit_lto/AlgorithmPlanner.cpp
KyleFromNVIDIA: See https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__EXEC.html#group__CUDA__EXEC_1gb9c891eb6bb8f4089758e64c9c976db9. `cuLaunchKernelEx()` takes a `CUfunction`, not a `CUkernel`.

divyegala: You could continue using `cudaLaunchKernelEx` though, and just assign the `CUkernel` to `cudaKernel_t`.

KyleFromNVIDIA: Is mixing driver and runtime types standard practice? As a reviewer, it wouldn't be immediately clear to me that the types are interchangeable. Besides, this is more consistent.

divyegala: Fair enough. I'm not sure if it's standard practice, I just wanted to keep the diff small and use the driver API/types only when necessary. 

If you prefer to keep driver API, that's understandable for clarity. That said, `CUfunction` and `CUkernel` can be interchangeably used.

KyleFromNVIDIA: > That said, `CUfunction` and `CUkernel` can be interchangeably used.

I wouldn't bet on that:

https://docs.nvidia.com/cuda/cuda-driver-api/group__CUDA__LIBRARY.html#group__CUDA__LIBRARY_1ge4cf9abafaba338acb977585b0d7374a

If they were interchangeable, there would be no need for `cuKernelGetFunction()`. Plus, I got a compile error when I tried to mix them.

cpp/CMakeLists.txt
```diff
@@ -780,6 +776,7 @@ if(NOT BUILD_CPU_ONLY)
              $<BUILD_LOCAL_INTERFACE:$<TARGET_NAME_IF_EXISTS:NCCL::NCCL>>
              $<BUILD_LOCAL_INTERFACE:$<TARGET_NAME_IF_EXISTS:hnswlib::hnswlib>>
              $<$<BOOL:${CUVS_NVTX}>:CUDA::nvtx3>
+             CUDA::cuda_driver
       PRIVATE
         $<TARGET_NAME_IF_EXISTS:OpenMP::OpenMP_CXX>
         $<COMPILE_ONLY:nvidia::cutlass::cutlass>
```
divyegala: This should be in `PRIVATE`

```suggestion
PRIVATE
$<TARGET_NAME_IF_EXISTS:OpenMP::OpenMP_CXX>
$<COMPILE_ONLY:nvidia::cutlass::cutlass>
CUDA::cuda_driver
```

cpp/src/detail/jit_lto/AlgorithmPlanner.cpp
divyegala: You can C style cast it `(CUfunction) kernel`, but this is okay too.

c/tests/CMakeLists.txt
```diff
@@ -94,7 +94,7 @@ endif()
 
 ConfigureTest(NAME cuvs_c_headers PATH core/headers.c)
 ConfigureTest(NAME cuvs_c_test PATH core/c_api.c)
-target_link_libraries(cuvs_c_test PRIVATE CUDA::cudart)
+target_link_libraries(cuvs_c_test PRIVATE CUDA::cudart_static)
```
KyleFromNVIDIA, line 97: This and related changes are from https://github.com/rapidsai/cuvs/pull/1931 and will disappear once that is merged

cpp/src/detail/jit_lto/AlgorithmPlanner.cpp
```diff
@@ -4,6 +4,7 @@
  */
 
 #include <chrono>
+#include <iostream>
```
KyleFromNVIDIA:
```suggestion
```

cpp/src/detail/jit_lto/AlgorithmPlanner.cpp
```diff
@@ -65,6 +66,7 @@ std::shared_ptr<AlgorithmLauncher> AlgorithmPlanner::build()
   RAFT_CUDA_TRY(cudaDeviceGetAttribute(&minor, cudaDevAttrComputeCapabilityMinor, device));
 
   std::string archs = "-arch=sm_" + std::to_string((major * 10 + minor));
+  std::cerr << "Passing argument to nvJitLink: " << archs << "\n";
```
KyleFromNVIDIA:
```suggestion
```

conda/recipes/libcuvs/recipe.yaml
```diff
@@ -431,9 +409,7 @@ outputs:
         - libcurand
         - libcusolver
         - libcusparse
-        - if: cuda_major == "13"
-          then:
-            - libnvjitlink
+        - libnvjitlink
```
bdice: We need to:
- remove the `run:` on `libnvjitlink`
- remove the `ignore_run_exports:` on `libnvjitlink`

This will allow us to use the _default_ run exports of `libnvjitlink-dev`. I believe that's what we really want here, because it enforces that the runtime version of `libnvjitlink` is `>=` the build-time version. That's the compatibility model for `libnvjitlink` which we want to adhere to.

See https://github.com/rapidsai/cudf/pull/21671#discussion_r2955324667 for justification and further explanation.

KyleFromNVIDIA:
Blocked on https://github.com/rapidsai/cuvs/pull/1936

gforsyth:
Just merged #1936 in

PR #1924, opened 2026-03-16
Fix hanging pytests

aamijar:
This PR attempts to fix the hanging pytest issues we are seeing in CI.
From my investigations there are two issues.

One is the `test_cagra_ace_tiny_memory_limit_triggers_disk_mode` and `test_hnsw_ace_tiny_memory_limit_triggers_disk_mode` tests which I have seen locally to hang. ~~At the moment, I am skipping these pytests in CI.~~
**Update**: After testing here 20 times https://github.com/rapidsai/cuvs/actions/runs/23182916436?pr=1893, I don't think we need to skip these tests.

Second is a memory out of bounds error in `device_matrix_view_from_host` , first revealed by compute-sanitizer. This is likely a HMM issue. I have added more specific checks in the constructor so a copy is used when necessary.

See this CI run from my test PR where I run 10 times to ensure that CI works with these fixes.
https://github.com/rapidsai/cuvs/actions/runs/23131230353
I also reran CI on that test PR again. So all the pytests were run 20 times.

### DIFF
--- a/python/cuvs/cuvs/tests/test_cagra_ace.py
+++ b/python/cuvs/cuvs/tests/test_cagra_ace.py
@@ -190,7 +190,6 @@ def test_cagra_ace_hierarchy(hierarchy):
     )
 
 
-@pytest.mark.skip(reason="Disk mode can cause pytest to hang")
 def test_cagra_ace_tiny_memory_limit_triggers_disk_mode():
     """Test that setting tiny memory limits triggers disk mode automatically."""
     n_rows = 5000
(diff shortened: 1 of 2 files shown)


cpp/src/neighbors/detail/cagra/utils.hpp
```diff
@@ -174,8 +174,11 @@ class device_matrix_view_from_host {
   {
     cudaPointerAttributes attr;
     RAFT_CUDA_TRY(cudaPointerGetAttributes(&attr, host_view.data_handle()));
-    device_ptr = reinterpret_cast<T*>(attr.devicePointer);
-    if (device_ptr == NULL) {
+    device_ptr      = reinterpret_cast<T*>(attr.devicePointer);
+    device_ptr      = reinterpret_cast<T*>(attr.devicePointer);
```
tarang-jain: This assignment of device_ptr is happening twice?

aamijar: Oops removed the extra one now

PR #1927, opened 2026-03-17
Remove JIT+LTO fragment database

KyleFromNVIDIA:
Rather than register each fragment in a runtime class with a string key, "register" them with the linker using template specialization. This solves a number of problems:

1. It simplifies the code by removing the `FragmentDatabase` class.
2. It addresses https://github.com/rapidsai/cuvs/pull/1909#discussion_r2941678281 by bypassing the issue entirely. There is no longer a need to build the fragment name string at runtime.
3. For clients that use the `cuvs_static` static library, it allows the linker to pick and choose which fragment symbols it needs rather than including all of them with every client just in case any of them are needed.
4. Since there is no longer a need for `$<WHOLE_ARCHIVE:...>` linkage, there is no need for the `cuvs_jit_lto_kernels` target at all, thus simplifying the CMake code too.


KyleFromNVIDIA:
We might have to rework it a bit, but it should still be supported. The UDF library takes the user-provided code, compiles it into a fragment, and calls `AlgorithmPlanner::add_fragment(const FragmentEntry&)`. We can add a `UDFFatbinFragmentEntry` class to support this.

cpp/src/neighbors/ivf_flat/ivf_flat_interleaved_scan_jit.cuh
```diff
@@ -148,16 +118,17 @@ void launch_kernel(const index<T, IdxT>& index,
...
-                                               decltype(get_idx_type_tag<IdxT>())>(
-    Capacity, Veclen, Ascending, ComputeNorm);
-  kernel_planner.template add_metric_device_function<decltype(get_data_type_tag<T>()),
-                                                     decltype(get_acc_type_tag<AccT>())>(
-    get_metric_name<MetricTag, Veclen, T, AccT>(), Veclen);
-  kernel_planner.add_filter_device_function(get_filter_name<IvfSampleFilterTag>());
-  kernel_planner.add_post_lambda_device_function(get_post_lambda_name<PostLambdaTag>());
+  InterleavedScanPlanner kernel_planner;
+  kernel_planner
+    .add_entrypoint<DataTag, AccTag, IdxTag, Capacity, Veclen, Ascending, ComputeNorm>();
```
divyegala, line 128: How is the template keyword not needed here?

KyleFromNVIDIA: I don't think it was ever needed to begin with. https://en.cppreference.com/w/cpp/language/member_template.html lists the following example:

```c++
class Base
{
    virtual void f(int);
};
 
struct Derived : Base
{
    // this member template does not override Base::f
    template<class T> void f(T);
 
    // non-template member override can call the template:
    void f(int i) override
    {
         f<>(i);
    }
};
```

Notice that `void f(int i)` is calling `f<>(i)` without the `template` keyword.

PR #1928, opened 2026-03-18
Preserve input memory location for NN Descent

### LINKED ISSUE #1901: Compute distances in NN Descent kernels in native types
Currently, NN Descent copies input data to a `fp16` for the distance computation kernel. This increases memory pressure severely, either adds an extra 0.5x for `fp32` or 2x for `int8/uint8` which is very restrictive to scaling.

jinsolp:
Closes https://github.com/rapidsai/cuvs/issues/1901

Previous Code
- We almost always allocate device side fp16 arrays. This was for... 
  - allowing wmma usage
  - allowing data modification for `CosineExpanded` preprocessing

Current PR Changes
- This PR makes nn-descent reuse the user's device-side data buffer when input data is already on device, instead of always allocating and copying into a separate staging buffer. This roughly halves the peak device memory footprint for the common UMAP/HDBSCAN call path where the dataset is already on GPU.
- Remove preprocessing for `CosineExpanded` metric (because we don't want to allocate additional device side data arrays) and do the computation inside the `calculate_metric` function.
  
### Peak memory usage Changes
  - food data (5M x 384) = 7.25GiB
  - sports data (13M x 284) = 18.55GiB

- notice how for FP32->FP16 Device (meaning data is already on device), previous code allocates a new fp16 array, resulting in more gpu mem usage. This PR ensures that we convert to fp16 on-the-fly (resulting in the overhead in time) instead of allocating new fp16 memory for that.
performance_metrics

### Performance Changes
- Conversion Overhead: On-the-fly conversion introduces negligible overhead.
- Cosine Metric: Now reads l2 norms inside the `calculate_metric` function, aligning with access pattern used by the L2 distance metric. Adds minimal overhead (e.g. previously 18.2937s VS 18.7598s for 5Mx384 data)


c/include/cuvs/neighbors/nn_descent.h
```diff
@@ -62,21 +73,80 @@ struct cuvsNNDescentIndexParams {
...
+ */
+struct cuvsNNDescentIndexParams_v6 {
+  cuvsDistanceType metric;
+  float metric_arg;
+  size_t graph_degree;
+  size_t intermediate_graph_degree;
+  size_t max_iterations;
+  float termination_threshold;
+  bool return_distances;
+  bool compress_to_fp16;
```
cjnolet, line 106: I think we should remove this. If the user wants fp16, they should just do this themselves. The problem is that this flips the ownership model (albeit it's done only temporarily, it still leads to unexpected behavior when we have to copy the data in fp16 form). Better if the user just converts this themselves. The problem is that we could offer this for every index type, but it's not really necessary when the user could just convert the d-type and call the index building process w/ it. Then they wouldn't have to deal w/ the additional copy in device memory at all.

c/include/cuvs/neighbors/nn_descent.h
```diff
@@ -17,9 +17,14 @@ extern "C" {
...
  * @brief Dtype to use for distance computation
- * - `NND_DIST_COMP_AUTO`: Automatically determine the best dtype for distance computation based on the dataset dimensions.
- * - `NND_DIST_COMP_FP32`: Use fp32 distance computation for better precision at the cost of performance and memory usage.
+ * - `NND_DIST_COMP_AUTO`: Automatically determine the best dtype for distance computation based on
+ * the dataset dimensions.
+ * - `NND_DIST_COMP_FP32`: Use fp32 distance computation for better precision at the cost of
+ * performance and memory usage.
  * - `NND_DIST_COMP_FP16`: Use fp16 distance computation.
+ *
+ * @deprecated To be removed in 26.08. Use cuvsNNDescentIndexParams_v6 with compress_to_fp16
```
cjnolet, line 26: Is the `compress_to_fp16` the only thing that's different between the old API and the new one? If that's the case, I suggest we remove the `compress_to_fp16` option altogether and never copy the dataset. I think setting the distance type is useful, but I don't think copying the dataset is useful.

c/include/cuvs/neighbors/nn_descent.h
jinsolp: That could be an option, but if we do so, the downstream ML algos will experience a slowdown. 
For example, UMAP or HDBSCAN only supports fp32, and the users are forced to experience a 2x slowdown in the knn computation step because nn descent will always use the fp32 distance computation.
Are we okay with this?

c/include/cuvs/neighbors/nn_descent.h
jinsolp: Essentially the existing distance types (`NND_DIST_COMP_x`) and the new `compress_to_fp16` do the same thing. I think the name could be a bit misleading, but this means "compress fp32 to fp16 to use fp16 distance computation". I'll change the name to `use_fp16_dist_comp`.

The reason it's changed from having 3 different dist comp options is because now the default behavior would be to use the original dtype.
Previously with the three distance computation types:
- `NND_DIST_COMP_AUTO`: if fp32, dispatch to fp32 or fp16 computation depending on dim. no affect for other dtype inputs.
- `NND_DIST_COMP_FP32`: force fp32 input to fp32 distance computation
-  `NND_DIST_COMP_FP16`: force fp32 input to fp16 distance computation

Since now we want fp32 input to always compute distance in fp32, having the `AUTO` and the `FP32` options doesn't make sense. So I decided to use a single boolean instead to decide whether to use fp32 distance computation OR fp16 distance computation for fp32 input.

cjnolet: I'm trying to avoid overcomplicating this- if the user wants distanc to be computed in fp16, they should convert their vectors to fp16, that's the benefit to them doing this conversion. 

For example, we supply a "distance-dtype" option in IVFPQ to decouple the distance precision from the input precision, but that's also because the vectors are nontrivially quantized. 

If we are to keep a separate option for the user to select the dtype for the distance itself, we should at least be consistent and follow the naming conventions that we use elsewhere and not introduce completely different naming. I also suspect "compressed" is the incorrect term to use here- we aren't explicitly quantizing fp32 to fp16, right? We're just downcasting to a small datatye.

python/cuvs/cuvs/neighbors/nn_descent/nn_descent.pyx
```diff
@@ -102,15 +102,8 @@ cdef class IndexParams:
...
-        if dist_comp_dtype is "auto":
-            self.params.dist_comp_dtype = cuvsNNDescentDistCompDtype.NND_DIST_COMP_AUTO
-        elif dist_comp_dtype is "fp32":
-            self.params.dist_comp_dtype = cuvsNNDescentDistCompDtype.NND_DIST_COMP_FP32
-        elif dist_comp_dtype is "fp16":
-            self.params.dist_comp_dtype = cuvsNNDescentDistCompDtype.NND_DIST_COMP_FP16
-        else:
-            raise ValueError(f"Invalid dist_comp_dtype: {dist_comp_dtype}. Supported options are 'auto', 'fp32', and 'fp16'.")
+        if use_fp16_dist_comp is not None:
+            self.params.use_fp16_dist_comp = use_fp16_dist_comp
```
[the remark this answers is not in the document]
jinsolp: outdated

python/cuvs/cuvs/neighbors/nn_descent/nn_descent.pyx
```diff
@@ -63,22 +63,22 @@ cdef class IndexParams:
...
-        Supported dtypes are `auto`, `fp32`, and `fp16`
-        `auto` automatically determines the best dtype for distance computation based on the dataset dimensions.
-        `fp32` uses fp32 distance computation for better precision at the cost of performance and memory usage. This option is only valid when data type is fp32.
-        `fp16` uses fp16 distance computation for better performance and memory usage at the cost of precision.
+    use_fp16_dist_comp : bool, default = False
+        When True and the input data is fp32, distance computation is performed
+        in fp16 for better performance and lower memory usage at the cost of
+        precision. This requires copying the fp32 input to an internal fp16
+        buffer on the device. Has no effect on non-fp32 input types (fp16,
+        int8, uint8) which always use fp16 distance computation.
```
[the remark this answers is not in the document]
jinsolp: addressed by renaming

cpp/src/neighbors/detail/nn_descent.cuh
```diff
@@ -1414,62 +1430,60 @@ template <typename DistEpilogue_t>
...
-                                                             ndim_,
-                                                             graph_buffer_.data_handle(),
-                                                             dists_buffer_.data_handle(),
-                                                             DEGREE_ON_DEVICE,
-                                                             d_locks_.data_handle(),
-                                                             l2_norms_.data_handle(),
-                                                             build_config_.metric,
-                                                             dist_epilogue);
+    // Data stored as input_t: device data used directly, or host data copied as-is.
+    launch_kernel(static_cast<const input_t*>(d_data_ptr_));
```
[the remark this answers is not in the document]
jinsolp: not a proper use case

cpp/src/neighbors/detail/nn_descent.cuh
```diff
@@ -1414,62 +1430,64 @@ template <typename DistEpilogue_t>
...
+                                                               NUM_SAMPLES,
+                                                               typed_ptr,
+                                                               ndim_,
+                                                               graph_buffer_.data_handle(),
+                                                               dists_buffer_.data_handle(),
+                                                               DEGREE_ON_DEVICE,
+                                                               d_locks_.data_handle(),
+                                                               l2_norms_.data_handle(),
+                                                               build_config_.metric,
+                                                               dist_epilogue);
```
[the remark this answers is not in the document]
jinsolp: done

cpp/src/neighbors/detail/nn_descent_gnnd.hpp
```diff
@@ -299,15 +307,15 @@ inline BuildConfig get_build_config(raft::resources const& res,
...
-                           .dist_comp_dtype       = params.dist_comp_dtype};
+  BuildConfig build_config{.max_dataset_size        = num_rows,
+                           .dataset_dim             = num_cols,
+                           .node_degree             = extended_graph_degree,
+                           .internal_node_degree    = extended_intermediate_degree,
+                           .max_iterations          = params.max_iterations,
+                           .termination_threshold   = params.termination_threshold,
+                           .output_graph_degree     = params.graph_degree,
+                           .metric                  = params.metric,
+                           .internal_distance_dtype = params.internal_distance_dtype};
```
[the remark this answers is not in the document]
jinsolp: done

c/include/cuvs/neighbors/nn_descent.h
```diff
@@ -64,21 +75,83 @@ struct cuvsNNDescentIndexParams {
...
+
+/**
+ * @brief De-allocate NN-Descent Index params (v6)
+ *
+ * @since 26.06
+ *
+ * @param[in] index_params
+ * @return cuvsError_t
+ */
+cuvsError_t cuvsNNDescentIndexParamsDestroy_v6(cuvsNNDescentIndexParams_v6_t index_params);
```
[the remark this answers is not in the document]
jinsolp: done

c/include/cuvs/neighbors/nn_descent.h
```diff
@@ -19,9 +19,14 @@ extern "C" {
...
- * - `NND_DIST_COMP_AUTO`: Automatically determine the best dtype for distance computation based on the dataset dimensions.
- * - `NND_DIST_COMP_FP32`: Use fp32 distance computation for better precision at the cost of performance and memory usage.
+ * - `NND_DIST_COMP_AUTO`: Automatically determine the best dtype for distance computation based on
+ * the dataset dimensions.
+ * - `NND_DIST_COMP_FP32`: Use fp32 distance computation for better precision at the cost of
+ * performance and memory usage.
  * - `NND_DIST_COMP_FP16`: Use fp16 distance computation.
+ *
+ * @deprecated To be removed in 26.08. Use cuvsNNDescentIndexParams_v6 with internal_distance_dtype
+ * instead.
```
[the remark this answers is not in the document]
jinsolp: done

jinsolp:
@cjnolet Thanks Corey addressed your reviews, but still would like to allow the user to choose the distance compute type. Summarizing below:

1. Naming: renamed to`internal_distance_dtype` to match the IVFPQ convention. Dropped the "compressed" wording throughout. You're right, we aren't quantizing, just downcasting fp32 → fp16. 
2. Keeping the option (vs. requiring callers to convert to fp16 themselves): I'd like to keep this for downstream ML. cuML UMAP/HDBSCAN currently feed fp32 data into NN-Descent and would regress noticeably for high-dim fp32 inputs if we don't take the wmma path.
3. "Don't copy the dataset": clarifying this - we only ever have 1 copy of the dataset on device, in its native dtype. The only exception is when the input is fp32 on host and `internal_distance_dtype = CUDA_R_16F`, in which case the on-device copy is in fp16 instead of its native fp32 dtype. Even in that case it's still only 1 copy of the data on device (fp16 replacing what would otherwise be fp32, which halves device memory and WMMA read bandwidth).

python/cuvs/cuvs/neighbors/nn_descent/nn_descent.pyx
```diff
@@ -63,22 +71,25 @@ cdef class IndexParams:
         The delta at which nn-descent will terminate its iterations
     return_distances : bool
         Whether to return distances array
-    dist_comp_dtype : str, default = "auto"
-        Dtype to use for distance computation.
-        Supported dtypes are `auto`, `fp32`, and `fp16`
-        `auto` automatically determines the best dtype for distance computation based on the dataset dimensions.
-        `fp32` uses fp32 distance computation for better precision at the cost of performance and memory usage. This option is only valid when data type is fp32.
-        `fp16` uses fp16 distance computation for better performance and memory usage at the cost of precision.
+    internal_distance_dtype : numpy dtype, default = np.float32
```
jinsolp: having this as numpy types follows ivfpq's usage

cjnolet:
> Naming: renamed tointernal_distance_dtype to match the IVFPQ convention. Dropped the "compressed" wording throughout. You're right, we aren't quantizing, just downcasting fp32 → fp16. 

Yep, and downcasting should be fine in many cases. If it's not, the user can just revert to using fp32. We'll need to document this accordingly. 

> Keeping the option (vs. requiring callers to convert to fp16 themselves): I'd like to keep this for downstream ML. cuML UMAP/HDBSCAN currently feed fp32 data into NN-Descent and would regress noticeably for high-dim fp32 inputs if we don't take the wmma path.

Actually, I think HDBSCAN and UMAP should be accepting fp16 natively. Have you done any benchmarks to see the e2e perf impact through an algorithm like UMAP?

> "Don't copy the dataset": clarifying this - we only ever have 1 copy of the dataset on device, in its native dtype. The only exception is when the input is fp32 on host and internal_distance_dtype = CUDA_R_16F, in which case the on-device copy is in fp16 instead of its native fp32 dtype. Even in that case it's still only 1 copy of the data on device (fp16 replacing what would otherwise be fp32, which halves device memory and WMMA read bandwidth).

How about we do this: short-term, let's keep this only for host inputs today, but let's let the user decide when to use that (I think you've done this already from your summary above). In the meantime, let's create 2 Github issues: one to remove that functionality when cuML is ready, and one to migrate cuml so that it can natively accept fp16. Let's add the first to comments in the code so we don't lose track. 

I agree that if perf impact is 2x, we should not lose the general capability, but I get deeply concerned when we have algorithms doing things fundamentally differently from other algorithms- it creates unexpected surprises, and makes it more challenging for the users to reason about what's going on under the hood.

jinsolp:
> Actually, I think HDBSCAN and UMAP should be accepting fp16 natively. Have you done any benchmarks to see the e2e perf impact through an algorithm like UMAP?

Unfortunately UMAP and HDBSCAN doesn't support fp16 🥲

> In the meantime, let's create 2 Github issues: one to remove that functionality when cuML is ready, and one to migrate cuml so that it can natively accept fp16. Let's add the first to comments in the code so we don't lose track.

Sounds good. I'll add issues and link them here too

https://github.com/rapidsai/cuvs/issues/2079
https://github.com/rapidsai/cuml/issues/8102

cjnolet:
> Unfortunately UMAP and HDBSCAN doesn't support fp16 

I know. I'm saying we should eventually make them accept fp16.

jinsolp:
@cjnolet This has been changed to focus only on avoiding double-allocation of the data. No public user-facing changes.

Strictly keeping the input type and support for uint8/int8 distance computation will follow in another PR. https://github.com/rapidsai/cuvs/issues/2085

cpp/src/neighbors/detail/nn_descent.cuh
```diff
@@ -141,6 +141,22 @@ struct dtype_traits<__half> {
...
+};
+
+template <>
+struct dtype_traits<int8_t> {
+  static constexpr int APAD           = 4;
+  static constexpr int BPAD           = 4;
+  static constexpr int TILE_COL_WIDTH = 128;
+  static __device__ __forceinline__ float to_float(int8_t v) { return static_cast<float>(v); }
+};
+
```
divyegala: Minor stylistic choice, C++20 simplifies this:

```suggestion
template <typename T>
concept Byte = std::is_same_v<T, uint8_t> or std::is_same_v<T, int8_t>;

template <Byte T>
struct dtype_traits<T> {
  static constexpr int APAD           = 4;
  static constexpr int BPAD           = 4;
  static constexpr int TILE_COL_WIDTH = 128;
  static __device__ __forceinline__ float to_float(T v) { return static_cast<float>(v); }
};
```

cpp/src/neighbors/detail/nn_descent.cuh
```diff
@@ -244,69 +260,64 @@ __device__ __forceinline__ void load_vec(Data_t* vec_buffer,
...
+template <typename Data_t>
+  requires(!std::is_same_v<Data_t, __half>)
+__device__ __forceinline__ void load_vec(__half* vec_buffer,
+                                         const Data_t* d_vec,
+                                         const int load_dims,
+                                         const int padding_dims,
+                                         const int lane_id)
+{
+  constexpr int num_load_elems_per_warp = raft::warp_size();
+  __half half_0                         = __float2half(0.0f);
```
divyegala: Why run a mathematical expression here? Shouldn't this be possible?

```suggestion
  constexpr __half half_0                         = 0;
```

cpp/src/neighbors/detail/nn_descent.cuh
```diff
@@ -1415,61 +1422,68 @@ void GNND<Data_t, Index_t>::local_join(cudaStream_t stream, DistEpilogue_t dist_
...
+  //   fp32 dist (only fp32 input, dist_comp_dtype == FP32 or AUTO with dim <= 16) -> SIMT: scalar
+  //     element-wise distance computation in fp32.
+  //   fp16 dist (everything else: fp16/int8/uint8 input, or fp32 input with dist_comp_dtype ==
+  //     FP16 or AUTO with dim > 16) -> WMMA (tensor-core accelerated dot product). Non-fp16
+  //     dtypes are converted to fp16 on-the-fly while loading into shared memory; for fp32 host
+  //     input this conversion happens earlier at copy-in time (see d_data_half_).
+  //   L1 distance for any input -> SIMT (L1 needs element-wise ops, can't use tensor cores).
+  using DCT = cuvs::neighbors::nn_descent::DIST_COMP_DTYPE;
+  bool use_fp16_dist =
+    std::is_same_v<input_t, float> && (build_config_.dist_comp_dtype == DCT::FP16 ||
```
divyegala, line 1430: I'm a bit confused about the code-paths. When the input is in `int8` or `uint8`, what happens? Weren't we up-casting it before and wouldn't we wish to compute in native going forward?

divyegala: Oh I just saw this comment https://github.com/rapidsai/cuvs/pull/1928#issuecomment-4454799667

jinsolp: yes, currently we keep the upcasting behavior here. I'll make a follow up PR targeting next release!

cpp/src/neighbors/detail/nn_descent.cuh
jinsolp: It looks like the CUDA __half(int) constructor isn't constexpr, so `constexpr __half half_0 = 0` fails compilation

divyegala (approved):

PR #1931, opened 2026-03-18
Use static cudart

KyleFromNVIDIA:
Rehash of https://github.com/rapidsai/cuvs/pull/1627.

- Enable static linking of libcudart by default (`CUDA_STATIC_RUNTIME=ON`)
- Remove `cuda-cudart` from conda recipe run requirements (no longer needed when statically linked)

This is part of a RAPIDS-wide effort to switch to static CUDA runtime linking. See https://github.com/rapidsai/build-planning/issues/235 for tracking.

- `cpp/CMakeLists.txt`: Change `CUDA_STATIC_RUNTIME` default from OFF to ON
- `conda/recipes/cuvs/recipe.yaml`: Remove `cuda-cudart` from run deps
- `conda/recipes/libcuvs/recipe.yaml`: Remove `cuda-cudart` from run deps (4 outputs)

Note: Python builds already use `CUDA_STATIC_RUNTIME=ON` (set in `python/libcuvs/CMakeLists.txt`).


KyleFromNVIDIA:
Downloading and examining the Conda artifacts confirms that `libcuvs.so` has no link to `libcudart.so.1`. I think this one is the winner.

cpp/CMakeLists.txt
```diff
@@ -765,16 +763,20 @@ if(NOT BUILD_CPU_ONLY)
 
     target_link_libraries(
       cuvs
-      PUBLIC rmm::rmm
-             raft::raft
+      INTERFACE $<COMPILE_ONLY:rmm::rmm>
```
divyegala, line 766: I think we leak RMM symbols in our public header. With RMM now being a shared library, that means we pass on the dependency fully to our consumers right?

KyleFromNVIDIA: I'm aware that we leak RMM headers, but I'm not sure if we also leak shared library symbols.

I've made this `COMPILE_ONLY` because we don't want downstream consumers (tests, etc.) to get RMM's cudart dependency, and I'd rather not give them all the `NO_CUDART_DEP` target property.

KyleFromNVIDIA: Probably a better thing to do would be to stop leaking RMM symbols.

divyegala: I'm not aware of the boundary of header vs shared library symbols.

divyegala: So downstream consumers of cuVS will now need to add an `rmm::rmm` link right?

KyleFromNVIDIA: Only if they actually link against symbols from `librmm.so`. If not, they don't need to.

PR #1937, opened 2026-03-20
Improve the IVF-PQ Coarse Batch Size Workspace Estimation

julianmi:
OpenAI 5M dataset runs OOM when using quantized data (`float` to `int8`) as show in this [example](https://github.com/tfeher/cuvs_ace_test/blob/cagra_hnsw/src/hnsw_openai_int8.cu). The `kMinWorkspaceRatio` is too small for the `int8` data such that the IVF-PQ search `max_internal_batch_size` is not reduced. Doubling `kMinWorkspaceRatio` fixes this. There is no performance degradation for the OpenAI 5M `float` dataset with this change. Let me know if I should test more datasets.

CC @tfeher

### DIFF
--- a/cpp/src/neighbors/detail/cagra/cagra_build.cuh
+++ b/cpp/src/neighbors/detail/cagra/cagra_build.cuh
@@ -2080,7 +2080,7 @@ auto iterative_build_graph(
       curr_itopk_size = curr_topk + 32;
     }
 
-    RAFT_LOG_INFO(
+    RAFT_LOG_DEBUG(
       "# graph_size = %lu (%.3lf), graph_degree = %lu, query_size = %lu, itopk = %lu, topk = %lu",
       (uint64_t)cagra_graph.extent(0),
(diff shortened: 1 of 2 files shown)


aamijar (approved):
Thanks @julianmi! Is this targeted for release/26.04?

julianmi:
> Thanks @julianmi for the PR! I think the current fix most likely helps to avoid the OOM for all reasonable use cases. However, it doesn't address the problem: IVF-PQ search shouldn't ever fail with OOM - it does workspace calculation internally and should scale correctly. Looking at the IVF-PQ search code, I think we need to make the [coarse batch size estimate](https://github.com/rapidsai/cuvs/blob/3f77e4680c8a67df74ba337159397bdf07f75b53/cpp/src/neighbors/ivf_pq/ivf_p

Thank you, this is the much better fix. As discussed offline, we need to keep the `max_per_ws / 2` factor to keep space for the inner loop batching.

PR #1939, opened 2026-03-20
[FEA] Remove kmeans::fit_predict, optionally keep predicted labels from fit

lowener:
In this PR I remove the `kmeans::fit_predict()` function, to instead add an optional to keep the labels already computed in `fit()`.

Some unnecessary template instantiation are removed as well.


cpp/src/cluster/kmeans_balanced.cuh
```diff
@@ -91,7 +95,31 @@ void fit(const raft::resources& handle,
...
+                                                    inertia_ptr,
+                                                    labels_ptr);
+}
+
+template <typename DataT,
+          typename MathT,
+          typename IndexT,
+          typename LabelT,
+          typename MappingOpT = raft::identity_op>
+void fit_predict(const raft::resources& handle,
```
cjnolet: Question-  can we just remove fit_predict() and use fit() as the only mechanism for passing optional labels? Would be nice to clean this up and remove it altogether.

lowener: Yes it would be possible to only have fit().
Note: I created #1944 to add a C API to fit_predict(). If we want to go with only fit() then we should skip it before introducing it.

lowener:
Closing this PR:
After some benchmarking I found that returning the labels after a fit only lead to a slight improvements of the `fit_predict` of under 2% in average when n_init=1 (best case).

PR #1943, opened 2026-03-24
Vamana half type instantiation

### LINKED ISSUE #1753: [FEA] Add float16 Instantiations for Vamana index

viclafargue:
Closes #1753.


cpp/tests/neighbors/ann_vamana.cuh
```diff
@@ -251,7 +257,22 @@ class AnnVamanaTest : public ::testing::TestWithParam<AnnVamanaInputs> {
     database.resize(((size_t)ps.n_rows) * ps.dim, stream_);
     search_queries.resize(((size_t)ps.n_queries) * ps.dim, stream_);
     raft::random::RngState r(1234ULL);
-    if constexpr (std::is_same<DataT, float>{}) {
+    if constexpr (std::is_same_v<DataT, half>) {
+      rmm::device_uvector<float> database_f(ps.n_rows * ps.dim, stream_);
+      rmm::device_uvector<float> queries_f(ps.n_queries * ps.dim, stream_);
+      raft::random::normal(handle_, r, database_f.data(), ps.n_rows * ps.dim, 0.1f, 2.0f);
+      raft::random::normal(handle_, r, queries_f.data(), ps.n_queries * ps.dim, 0.1f, 2.0f);
+      thrust::transform(raft::resource::get_thrust_policy(handle_),
```
tarang-jain: Lets use raft::linalg with raft::cast_op instead of thrust here.

tarang-jain (commented):
Thanks for adding this instantiation. Have you compared with https://github.com/microsoft/DiskANN/blob/cpp_main/src/index.cpp? Serializing a half dataset might not work for DiskANN search (when `include_dataset` is set to True). Perhaps we must add a note, warning the user about this when they attempt to include the dataset while serializing the index.

viclafargue:
> Thanks for adding this instantiation. Have you compared with https://github.com/microsoft/DiskANN/blob/cpp_main/src/index.cpp? Serializing a half dataset might not work for DiskANN search (when `include_dataset` is set to True). Perhaps we must add a note, warning the user about this when they attempt to include the dataset while serializing the index.

Thanks @tarang-jain, good catch. I added a note in the doc and a warning in the serialization function.

cpp/src/neighbors/detail/vamana/vamana_serialize.cuh
```diff
@@ -310,6 +310,21 @@ void serialize(raft::resources const& res,
                bool include_dataset,
                bool sector_aligned)
 {
+  if constexpr (std::is_same_v<T, half>) {
+    if (include_dataset) {
+      RAFT_LOG_WARN(
+        "Serializing a half-precision (float16) dataset is not compatible with DiskANN search. "
+        "The serialized .data file uses raw half values with no type metadata, so DiskANN (which "
```
tarang-jain, line 323: Are you sure there is no type metadata? Does DiskANN break with a "half" `.data` file or does it give garbage results? Just want to make sure that the wording of the warning is accurate.

tarang-jain (approved):
I am ready to approve this PR barring my comment above. However, I am not sure about the utility of this feature yet (I am guessing once we have built the graph on half data which theoretically should be faster, we want to search using DiskANN with an fp32 dataset?). Can you also please report the binary size increase in libcuvs due to the new instantiation?

cjnolet:
Tagging @bkarsin for more insight- I know half type in vamana has come up a lot. Does DiskANN support half precision? Does building a graph in half precision make sense? Does it make sense to do mixed (build in half an dsearch in full)?

Issue #1944, opened 2026-03-24
[FEA] Add C and Python API for `kmeans::fit_predict`


lowener:
Currently implemented in C++ but absent from C and Python.

PR #1956, opened 2026-03-25
[REVIEW] Add ``KMeans.fit_predict`` to Python API

jrbourbeau:
This PR handles the Python portion of https://github.com/rapidsai/cuvs/issues/1944


python/cuvs/cuvs/cluster/kmeans/kmeans.pyx
```diff
@@ -239,6 +243,86 @@ def fit(
...
+    labels_out, inertia_pred = predict(
+        params,
+        X,
+        centroids_out,
+        sample_weights=sample_weights,
+        labels=labels,
+        normalize_weight=normalize_weight,
+        resources=resources,
+    )
+    return FitPredictOutput(labels_out, centroids_out, inertia_pred, n_iter)
```
lowener, line 323: This would need to call the missing C function "cuvsKMeansFitPredict" because #1939 is adding some improvements to the `fit_predict` function that make it run faster than calling independantly both functions.

lowener (changes requested):

python/cuvs/cuvs/cluster/kmeans/kmeans.pyx
jrbourbeau: Thanks for pointing me to https://github.com/rapidsai/cuvs/pull/1939 @lowener. Do the changes in that PR break the code here? I'd like to keep the scope of this PR to just adding the Python API. Seems like maybe this could be updated to a more optimized version after the C API is added

cjnolet: @jrbourbeau, given that we're likely to merge @lowener's PR first in sequence, could you base your branch on his and build from his changes? Then we can merge yours in shortly after.

lowener: FYI the latest update on #1939 is that the C++ fit_predict() function will be removed and the labels will be returned as part of the fit() function as an optionnal output

docs/source/python_api/cluster_kmeans.rst
```diff
@@ -16,6 +16,11 @@ K-Means Fit
 
 .. autofunction:: cuvs.cluster.kmeans.fit
```
cjnolet, line 18: This is going away in preference for the Fern-based docs, which we can generate with codex. Just FYI. The process is about to get a lot easier.

python/cuvs/cuvs/cluster/kmeans/kmeans.pyx
cjnolet: @lowener is this still the case after your benchmarking or are we planning to keep the explicit fit_predict?

cjnolet (commented):
Apologies- I had this feedback queued up for awhile and never explicitly posted it.

PR #1962, opened 2026-03-28
[DO NOT MERGE] [FEA] KMeans - Batched (Streaming) Predict for Host Matrices

tarang-jain:
The predict function was separated out from #1886 into this PR. This PR is exploratory and we must evaluate the need for a host side predict function because the streaming of batches to device and calling of device side predict can be done by the user with the device side predict API itself.


tarang-jain:
Closing this PR based on offline discussion. `predict` is trivially batch-able.

Issue #1513, opened 2025-11-07
Remove duplicate kernel instantiations across `libcuvs`


divyegala:
The rule of thumb is that any-time a kernel is launched from a TU (directly or indirectly) using the triple chevron syntax `kernel<<<...>>>`, a copy of the kernel will be included in that TU. This starts becoming a problem when the library has the same kernel instance being launched from multiple TUs.

The only solution is to make sure that the kernel is launched as part of a pre-compiled function, and other functions that need to launch the kernel only have visibility of the declaration of that pre-compiled function.

PR #1969, opened 2026-03-31
Expose `kmeans::helpers::build_clusters`

aamijar:
Expose `kmeans::helpers::build_clusters` instead of relying on the internal api such as:
```
#include "../../cluster/kmeans_balanced.cuh"
```


cpp/src/cluster/kmeans_balanced.cuh
```diff
@@ -209,26 +209,30 @@ void build_clusters(const raft::resources& handle,
...
+    X.data_handle(),
+    X.extent(0),
+    centroids.extent(0),
+    centroids.data_handle(),
+    labels.data_handle(),
+    cluster_sizes.data_handle(),
+    mapping_op,
+    raft::resource::get_workspace_resource(handle),
+    X_norm.has_value() ? X_norm.value().data_handle() : nullptr);
+}
```
aamijar, line 235: Are these changes okay to make given that this function should be called using the public api only now?
1. Move definition back into `kmeans_balanced.cuh`
2. Delete `kmeans_balanced_build_clusters_impl.cuh`

Reference:
> Call the function from the public API wherever possible

https://github.com/rapidsai/cuvs/issues/1513

cjnolet:
@aamijar cam you please explain more clearly why this is needed in the public APIs? Is this absolutely needed in the public APIs or is it just for a mild convenience?

aamijar:
> @aamijar cam you please explain more clearly why this is needed in the public APIs? Is this absolutely needed in the public APIs or is it just for a mild convenience?

I was trying to address this comment

https://github.com/rapidsai/cuvs/blob/99b38018e44fc9f305514ce854c9f1de7b9bf905/cpp/src/neighbors/ivf_pq/ivf_pq_build.cuh#L22-L23

aamijar:
Closing for now after offline discussion.

PR #1970, opened 2026-03-31
Refactor `StaticFatbinFragmentEntry` to use tags

KyleFromNVIDIA:
Instead of requiring fragments to extend this class, allow them to simply pass a tag of their choice as a template parameter. This allows us to stop duplicating the `data` and `length` fields everywhere.

Also run `clang-format` on CMake C++ placeholder template files and update the JIT+LTO development guide.

### DIFF
--- a/docs/source/jit_lto_guide.md
+++ b/docs/source/jit_lto_guide.md
@@ -637,7 +637,17 @@ constexpr auto get_out_type_tag() {
   if constexpr (std::is_same_v<OutType, double>) return tag_d{};
 }
 
-template <typename T, typename OutT, typename IdxT, typename DistanceTag, typename FilterTag, bool Optimized, int Veclen>
+template <DistanceType Metric>
+constexpr auto get_metric_tag() {
+  if constexpr (Metric == DistanceType::Euclidean) return tag_metric_euclidean{};
+}
+
+template <FilterType Filter>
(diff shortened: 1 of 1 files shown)


.pre-commit-config.yaml
```diff
@@ -91,6 +92,14 @@ repos:
                 pass_filenames: false
                 files: rust/.*
                 language: rust
+              - id: clang-format-with-cmake-placeholders
```
bdice, line 95: If we start using this in multiple RAPIDS repos we might centralize the hook definition in https://github.com/rapidsai/pre-commit-hooks/.

docs/source/jit_lto_guide.md
```diff
@@ -633,7 +631,13 @@ constexpr auto get_idx_type_tag() {
...
 }
 
-template <typename T, typename OutT, typename IdxT>
+template <typename OutType>
+constexpr auto get_out_type_tag() {
+  if constexpr (std::is_same_v<OutType, float>) return tag_f{};
+  if constexpr (std::is_same_v<OutType, double>) return tag_d{};
+}
+
+template <typename T, typename OutT, typename IdxT, typename DistanceTag, typename FilterTag, bool Optimized, int Veclen>
```
divyegala: At this function entry, we should be getting real types and not tags.

PR #1984, opened 2026-04-02
Refactor instantiation matrices to generate at build time

KyleFromNVIDIA:
Rather than generating these matrices of files with Python scripts at development time (an anti-pattern), generate them at build time in CMake script.

For each algorithm that has multiple instantiation TUs, we create a JSON matrix with all of the dimensions of the instantiations. Then we call `compute_matrix_product()` in CMake script to get the matrix product, and then call `configure_file()` to substitute the variables in the instantiation template files (`*.cu.in` and `*.cpp.in`). We add these files to a list variable which then gets passed to `add_library()`.


cpp/src/distance/detail/pairwise_matrix/dispatch_matrix.json
```diff
@@ -0,0 +1,128 @@
...
+        "acc_abbrev": "f",
+        "out_type": "float",
+        "out_abbrev": "f"
+      }
+    ],
+    "_op": [
+      {
+        "op_type": "cuvs::distance::detail::ops::canberra_distance_op",
+        "op_abbrev": "canberra",
+        "archs": "#include <distance/detail/pairwise_matrix/dispatch_sm60.cuh>"
```
dantegd: This is pretty clever and creative! But somewhat unusual. It's coupling the JSON data file with source-level implementation details, a comment or documentation somewhere explainin the patterns for it would be useful.

cpp/cmake/modules/generate_string_matrix.cmake
```diff
@@ -0,0 +1,66 @@
...
+
+function(process_string_matrix_entry string_var)
+  set(options)
+  set(one_value ITEM_FORMAT GLUE MATRIX_JSON_ENTRY)
+  set(multi_value)
+
+  cmake_parse_arguments(_GSM "${options}" "${one_value}" "${multi_value}" ${ARGN})
+
+  populate_matrix_variables("${_GSM_MATRIX_JSON_ENTRY}")
+  set(semicolon ";")
```
dantegd, line 22: lol gotta love that we need this workaround. Most people familiar with CMake will realize why this is needed, but a brief comment for other folks might be useful.

cpp/src/neighbors/ball_cover/detail/ball_cover/registers_00_generate.py
```diff
@@ -1,130 +0,0 @@
-# SPDX-FileCopyrightText: Copyright (c) 2025, NVIDIA CORPORATION.
-# SPDX-License-Identifier: Apache-2.0
-
-import datetime
```
dantegd, line 4: I might have missed in the long diff or be wrong, but seems that we are missing the JSON replacement for the RBC generator here?

cpp/CMakeLists.txt
```diff
@@ -222,100 +225,110 @@ if(NOT BUILD_CPU_ONLY)
...
+    ITEM_FORMAT
+    "extern template struct standard_descriptor_spec<DistanceType::@metric@, @team_size@, @dim@, @data_type@, @index_type@, @distance_type@>@semicolon@"
+    GLUE
+    "\n"
+    MATRIX_JSON_FILE
+    "${CMAKE_CURRENT_SOURCE_DIR}/src/neighbors/detail/cagra/compute_distance_standard_matrix.json"
+  )
+
+  generate_inst_matrix(
+    cagra_compute_distance_standard_inst_files
```
dantegd:
```suggestion
    cagra_compute_distance_vpq_inst_files
```

cpp/src/neighbors/ball_cover/detail/ball_cover/registers_00_generate.py
KyleFromNVIDIA: Indeed, I'll restore this file next time I push. RBC was a minor offender that I decided to leave alone for the time being.

cpp/src/distance/detail/pairwise_matrix/dispatch_matrix.json
KyleFromNVIDIA: It's an ugly workaround for the fact that we're doing simple string substitutions from the matrix product without being able to do any intelligent transformation on them. I hate that I had to do this. Some time in the future, I'd like to introduce some sort of "additional transformation lambda" functionality into the matrix product to make it easier to do more complex things like this, but we can save that for a follow-up.

PR #1988, opened 2026-04-02
Remove dangling pointers in JIT Fragments

divyegala:


cpp/include/cuvs/detail/jit_lto/AlgorithmPlanner.hpp
```diff
@@ -5,29 +5,33 @@
...
 
   std::string entrypoint;
-  std::vector<const FragmentEntry*> fragments;
+  std::vector<std::unique_ptr<FragmentEntry>> fragments;
 
-  void add_fragment(const FragmentEntry& fragment);
+  template <typename T, typename = std::enable_if_t<std::is_convertible_v<T*, FragmentEntry*>>>
+  void add_fragment(std::unique_ptr<T> fragment)
+  {
+    fragments.push_back(std::unique_ptr<FragmentEntry>(std::move(fragment)));
```
dantegd, line 28: More of a curiosity if I'm correct, but this can be simplified to 

```C++
fragments.push_back(std::move(fragment));
```

no? My reasoning is that `std::unique_ptr<Derived>` has an implicit converting move constructor to `std::unique_ptr<Base>` when the pointer types are convertible (which the SFINAE constraint guarantees)

divyegala: Yes, I think you are right because `push_back` does not take a reference. But I will address this in any of our other JIT PRs as CI has passed.

PR #1990, opened 2026-04-03
Migrate RMM usage to CCCL MR design

bdice:
## Summary
- Migrate all RMM usage to the new CCCL memory resource design (de-templated resources, `device_async_resource_ref` instead of `device_memory_resource*`, value semantics)
- Replace `get_workspace_resource()` / `get_large_workspace_resource()` with `_ref()` variants across 65 call sites
- Rewrite `cuda_huge_page_resource` to satisfy CCCL `resource` concept directly
- Remove `owning_wrapper` / `dynamic_cast` patterns in C API and benchmarks

Depends on https://github.com/rapidsai/rmm/pull/2361.
Depends on https://github.com/rapidsai/ucxx/pull/636.
Depends on https://github.com/rapidsai/raft/pull/2996.

## Changes
- **33 files changed** (~208 insertions, ~221 deletions)
- `device_memory_resource*` params → `device_async_resource_ref` (ivf_common, ivf_pq, naive_knn)
- `get_current_device_resource()` → `get_current_device_resource_ref()`
- `set_current_device_resource()` → `set_current_device_resource_ref()`
- De-templated `pool_memory_resource`, `failure_callback_resource_adaptor` in bench utils
- Removed `&resource` pointer patterns (resources are now copyable value types)
- Removed spurious `mr` arg from `select_k` calls (previously compiled due to implicit pointer→bool conversion)
- C API pool resource management rewritten without `owning_wrapper`


c/src/core/c_api.cpp
```diff
@@ -132,60 +135,47 @@ extern "C" cuvsError_t cuvsRMMAlloc(cuvsResources_t res, void** ptr, size_t byte
...
+    auto mr      = rmm::mr::get_current_device_resource_ref();
+    mr.deallocate(raft::resource::get_cuda_stream(*res_ptr), ptr, bytes);
   });
 }
 
-thread_local std::shared_ptr<
-  rmm::mr::owning_wrapper<rmm::mr::pool_memory_resource<rmm::mr::device_memory_resource>,
-                          rmm::mr::device_memory_resource>>
-  pool_mr;
+thread_local cuda::mr::any_resource<cuda::mr::device_accessible> pool_upstream;
```
bdice: We do not need to define `pool_upstream` or `pool_mr` here anymore if they're only being defined here for lifetime reasons. The `pool_mr` will own its upstream, and `rmm::mr::set_current_device_resource` will store to a map of owning `any_resource` types. I need to check if this is only used for lifetime reasons, and remove if so.

cpp/bench/ann/src/common/cuda_huge_page_resource.hpp
```diff
@@ -17,37 +19,25 @@
 
 namespace raft::mr {
```
bdice, line 20: Why is cuVS defining classes in the `raft::` namespace? (This is not in scope for this PR.)

cpp/bench/ann/src/common/cuda_huge_page_resource.hpp
```diff
@@ -60,36 +50,29 @@ class cuda_huge_page_resource final : public rmm::mr::device_memory_resource {
...
+                       std::size_t bytes,
+                       std::size_t alignment = rmm::CUDA_ALLOCATION_ALIGNMENT) noexcept
   {
-    return dynamic_cast<cuda_huge_page_resource const*>(&other) != nullptr;
+    deallocate(cuda::stream_ref{cudaStream_t{nullptr}}, ptr, bytes, alignment);
   }
+
+  bool operator==(cuda_huge_page_resource const&) const noexcept { return true; }
+
+  friend void get_property(cuda_huge_page_resource const&, cuda::mr::device_accessible) noexcept {}
```
bdice: Also need to add host_accessible here.

cpp/bench/ann/src/cuvs/cuvs_ann_bench_utils.h
```diff
@@ -97,13 +98,10 @@ class shared_raft_resources {
...
-  auto get_large_memory_resource() noexcept
-  {
-    return static_cast<rmm::mr::device_memory_resource*>(&large_mr_);
-  }
+  auto get_large_memory_resource() noexcept -> rmm::device_async_resource_ref { return large_mr_; }
 
  private:
-  rmm::mr::device_memory_resource* orig_resource_;
+  cuda::mr::any_resource<cuda::mr::device_accessible> orig_resource_;
   pool_mr_type pool_resource_;
```
bdice: We no longer need to store `pool_resource_` as a member if its only purpose is to preserve the lifetime of the upstream memory resource for `resource_`.

cpp/internal/cuvs_internal/neighbors/naive_knn.cuh
```diff
@@ -116,8 +116,7 @@ void naive_knn(raft::resources const& handle,
                                           static_cast<int>(k),
                                           dist_topk + offset * k,
                                           indices_topk + offset * k,
-                                          type != cuvs::distance::DistanceType::InnerProduct,
-                                          mr);
+                                          type != cuvs::distance::DistanceType::InnerProduct);
```
bdice, line 119: Need to check if this change is intentional or a mistake.

cpp/tests/neighbors/naive_knn.cuh
```diff
@@ -131,8 +132,7 @@ void naive_knn(raft::resources const& handle,
                                                 static_cast<int>(k),
                                                 dist_topk + offset * k,
                                                 indices_topk + offset * k,
-                                                cuvs::distance::is_min_close(type),
-                                                mr);
+                                                cuvs::distance::is_min_close(type));
```
bdice, line 135: Need to check if this API change was intentional.

cpp/internal/cuvs_internal/neighbors/naive_knn.cuh
bdice: Seems like this was changed in https://github.com/rapidsai/raft/pull/1786?

cpp/tests/neighbors/naive_knn.cuh
bdice: Seems like this was changed in https://github.com/rapidsai/raft/pull/1786?

c/src/core/c_api.cpp
bdice: Done, this has been removed.

cpp/bench/ann/src/common/cuda_huge_page_resource.hpp
bdice: Done.

cpp/bench/ann/src/cuvs/cuvs_ann_bench_utils.h
bdice: Done. It turns out `resource_` could also be removed, because this class is only functioning as an RAII manager for the current device resource. It doesn't expose `resource_` itself.

divyegala (approved):

cpp/internal/cuvs_internal/neighbors/naive_knn.cuh
bdice: Maybe this was a dead code path. I can't see how it would've been building successfully prior to this migration otherwise.

cpp/tests/neighbors/naive_knn.cuh
bdice: Maybe this was a dead code path. I can't see how it would've been building successfully prior to this migration otherwise.

bdice:
Requesting admin-merge. Most CI is passing, the one failure seems like a fluke. I am not going to wait for the slow L4 C++ job since all others have passed and this is blocking builds for cuML + cuGraph.

PR #1996, opened 2026-04-07
[FEA] Standardize KMeans Instantiations

tarang-jain:
Simplify and standardize the kmeans public API types:

- **Extents**: Remove all `int` (int32) extent overloads; only `int64_t` remains. This halves the number of template instantiations
- **Labels**: All `predict` and `fit_predict` outputs use `uint32_t` labels (cluster IDs always fit in 32 bits since `n_clusters` is `int`). Removes redundant `int` and `int64_t` label overloads from balanced kmeans.
- **n_iter**: Changed from `host_scalar_view<IndexT>` to `host_scalar_view<int>` throughout — iteration count is always small.

These changes are made in the detail templates themselves, so the `.cu` wrappers are clean pass-throughs with no temporary buffers or type conversions.


cjnolet:
@tarang-jain these changes will need to be verified through cuml and cugraph. Does Faiss need to be checked also?

tarang-jain:
Faiss wont be affected.
cuML will be affected yes.

PR #2001, opened 2026-04-08
Reuse minClusterAndDistance Helper for Balanced KMeans

tarang-jain:
- The norm computation + fused reduction is already present in the minClusterDistanceCompute function. We can reuse that for balanced kmeans.

- Furthermore, this PR updates the minClusterDistanceCompute function to also use the fused kernel for the cosine metric.

Binary size savings (conda-cpp-build check with CUDA 12.9.1 + x86_64):
main: 660.72 MB
This PR: 648.83 MB
(conda-cpp-build check with CUDA 13.1.1 + amd):
main: 305.70 B
This PR: 300.63 MB


aamijar (approved):
Thanks for the refactor Tarang! Could you add a description to the PR?

cpp/src/cluster/detail/minClusterDistanceCompute.cu
```diff
@@ -31,74 +31,79 @@ void minClusterAndDistanceCompute(
...
+
+    auto centroidsNorm =
+      raft::make_device_vector_view<DataT, IndexT>(L2NormBuf_OR_DistBuf.data(), n_clusters);
+
+    raft::KeyValuePair<IndexT, DataT> initial_value(0, std::numeric_limits<DataT>::max());
+    raft::matrix::fill(handle, minClusterAndDistance, initial_value);
+
+    workspace.resize((sizeof(int)) * n_samples, stream);
+
+    // todo(lsugy): remove cIdx
```
jinsolp:
```suggestion
```

I think this trickled in from the code above

cpp/src/cluster/detail/minClusterDistanceCompute.cu
```diff
@@ -31,74 +31,79 @@ void minClusterAndDistanceCompute(
...
+        raft::sqrt_op{});
+    } else {
+      raft::linalg::norm<raft::linalg::L2Norm, raft::Apply::ALONG_ROWS>(
+        handle,
+        centroids,
+        raft::make_device_vector_view<DataT, IndexT>(L2NormBuf_OR_DistBuf.data(), n_clusters));
+    }
+
+    auto centroidsNorm =
+      raft::make_device_vector_view<DataT, IndexT>(L2NormBuf_OR_DistBuf.data(), n_clusters);
```
jinsolp:
```suggestion
auto centroidsNorm =
      raft::make_device_vector_view<DataT, IndexT>(L2NormBuf_OR_DistBuf.data(), n_clusters); 
      
    if (metric == cuvs::distance::DistanceType::CosineExpanded) {
      raft::linalg::norm<raft::linalg::L2Norm, raft::Apply::ALONG_ROWS>(
        handle,
        centroids,
        centroidsNorm,
        raft::sqrt_op{});
    } else {
      raft::linalg::norm<raft::linalg::L2Norm, raft::Apply::ALONG_ROWS>(
        handle,
        centroids,
        centroidsNorm;
    }
```
nitpick

cpp/src/cluster/detail/minClusterDistanceCompute.cu
jinsolp:
```suggestion
```
I think this trickled in by mistake

cpp/src/cluster/detail/minClusterDistanceCompute.cu
```diff
@@ -31,74 +31,79 @@ void minClusterAndDistanceCompute(
   auto n_clusters     = centroids.extent(0);
   // todo(lsugy): change batch size computation when using fusedL2NN!
   bool is_fused = metric == cuvs::distance::DistanceType::L2Expanded ||
-                  metric == cuvs::distance::DistanceType::L2SqrtExpanded;
+                  metric == cuvs::distance::DistanceType::L2SqrtExpanded ||
+                  metric == cuvs::distance::DistanceType::CosineExpanded;
   auto dataBatchSize = is_fused ? (IndexT)n_samples : getDataBatchSize(batch_samples, n_samples);
   auto centroidsBatchSize = getCentroidsBatchSize(batch_centroids, n_clusters);
 
   if (is_fused) {
```
jinsolp, line 37: If my understanding is correct, we are no longer doing the batching ("tiling over input data") when `is_fused=True`? Is this intended? Please correct me if I'm wrong

jinsolp (changes requested):
Thanks @tarang-jain ! Suggesting small changes and adding a question:

cpp/src/cluster/detail/minClusterDistanceCompute.cu
tarang-jain: I had removed those, but then re-added them to not change comments that we already had.  I am aware that this comment does not seem to be in the right place though. Will inspect further.

cpp/src/cluster/detail/minClusterDistanceCompute.cu
tarang-jain: That is correct. And I can move the batch size computation to the "else" block. So we dont even compute the batch size for the not fused condition.

tarang-jain: I moved the entire "tiling" code and batch size computation to the else condition.

cpp/src/cluster/detail/minClusterDistanceCompute.cu
tarang-jain: I have removed this one. I think what the comment mentions is that they want to get rid of the looping for the centroids in the fused case, which is precisely what this PR does.

tarang-jain: Removed this as well.

cpp/src/cluster/detail/minClusterDistanceCompute.cu
```diff
@@ -181,86 +181,70 @@ void minClusterDistanceCompute(raft::resources const& handle,
...
     L2NormBuf_OR_DistBuf.resize(n_clusters, stream);
+    auto centroidsNorm =
+      raft::make_device_vector_view<DataT, IndexT>(L2NormBuf_OR_DistBuf.data(), n_clusters);
+
     raft::linalg::norm<raft::linalg::L2Norm, raft::Apply::ALONG_ROWS>(
       handle,
       raft::make_device_matrix_view<const DataT, IndexT>(
         centroids.data_handle(), centroids.extent(0), centroids.extent(1)),
```
jinsolp: should we be computing norms like this here too?
```
if (metric == cuvs::distance::DistanceType::CosineExpanded) {
      raft::linalg::norm<raft::linalg::L2Norm, raft::Apply::ALONG_ROWS>(
        handle, centroids, centroidsNorm, raft::sqrt_op{});
    } else {
      raft::linalg::norm<raft::linalg::L2Norm, raft::Apply::ALONG_ROWS>(
        handle, centroids, centroidsNorm);
    }
```

cpp/src/cluster/detail/minClusterDistanceCompute.cu
```diff
@@ -181,86 +181,70 @@ void minClusterDistanceCompute(raft::resources const& handle,
...
+      minClusterDistance.data_handle(),
+      X.data_handle(),
+      centroids.data_handle(),
+      L2NormX.data_handle(),
+      centroidsNorm.data_handle(),
+      n_samples,
+      n_clusters,
+      n_features,
+      (void*)workspace.data(),
+      metric != cuvs::distance::DistanceType::L2Expanded,
```
jinsolp, line 221: Looking at the deleted code in `kmeans_balanced.cuh`,  this used to be false for the `CosineExpended` metric. However, this condition passes true for the `CosineExpended` metric and sqrt-s the distances output.

tarang-jain: Yes that is because Cosine is now supported in our fused kernel.

cpp/src/cluster/detail/minClusterDistanceCompute.cu
tarang-jain: Good catch! I have fixed this.

cpp/src/cluster/detail/minClusterDistanceCompute.cu
tarang-jain: This wasn't being caught because we were lacking tests for the cosine metric. We always test L2 (default). I am trying to add cosine inputs to the tests next.

achirkin:
Could you please refresh my understanding on couple math questions?

1. By the time we call `minClusterDistanceCompute`, the data is normalized, so we can use the same L2 norm everywhere, right?
2. Why is sqrt needed for the cosine case?

tarang-jain:
> By the time we call minClusterDistanceCompute, the data is normalized, so we can use the same L2 norm everywhere, right?

Cosine is only supported in balanced kmeans. When we call minClusterAndDistance, the data is not normalized, but the centroids are.

> Why is sqrt needed for the cosine case?

sqrt is needed because the cosine distance op directly divides by the norms (it does not do the sqrt):
https://github.com/rapidsai/cuvs/blob/988ebf565925fd4af278cf0448d40db2b84f66de/cpp/src/distance/detail/distance_ops/cosine.cuh#L20

cpp/src/cluster/detail/minClusterDistanceCompute.cu
jinsolp: I don't think anybody will call into this directly, but can we add a note here saying we require normalized input when using the cosine metric? I'm worried that this might just silently fail without the user knowing it. Or mabe we can randomly grab a few rows and check whether they are normalized? If not throw an error and exit?

cpp/src/cluster/detail/minClusterDistanceCompute.cu
```diff
@@ -181,86 +181,75 @@ void minClusterDistanceCompute(raft::resources const& handle,
...
+      raft::matrix::fill(handle, centroidsNorm, DataT{1});
+    } else {
+      raft::linalg::norm<raft::linalg::L2Norm, raft::Apply::ALONG_ROWS>(
+        handle,
+        raft::make_device_matrix_view<const DataT, IndexT>(
+          centroids.data_handle(), centroids.extent(0), centroids.extent(1)),
+        centroidsNorm);
+    }
+
+    workspace.resize((sizeof(IndexT)) * n_samples, stream);
```
jinsolp: do we want this to be `sizeof(int)` to avoid over-allocating (not sure if `IndexT` is ever `int64` here but just to be safe)? It looks like `fusedDistanceNNMinReduce` uses the workspace as a mutex which is always int types.

tarang-jain: you're right! I am now using an int workspace.

cpp/src/cluster/detail/minClusterDistanceCompute.cu
tarang-jain: Added a note at the top of the functions

cpp/src/cluster/detail/minClusterDistanceCompute.cu
```diff
@@ -29,76 +32,75 @@ void minClusterAndDistanceCompute(
...
+    auto centroidsNorm =
+      raft::make_device_vector_view<DataT, IndexT>(L2NormBuf_OR_DistBuf.data(), n_clusters);
+
+    if (metric == cuvs::distance::DistanceType::CosineExpanded) {
+      // Centroids are L2-normalized for cosine metric
+      raft::matrix::fill(handle, centroidsNorm, DataT{1});
+    } else {
+      raft::linalg::norm<raft::linalg::L2Norm, raft::Apply::ALONG_ROWS>(
+        handle, centroids, centroidsNorm);
+    }
```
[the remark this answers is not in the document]
tarang-jain: Regular kmeans does not have cosine today.

jinsolp (approved):
Thanks @tarang-jain !

cpp/src/cluster/detail/minClusterDistanceCompute.cu
```diff
@@ -29,76 +32,75 @@ void minClusterAndDistanceCompute(
...
-      raft::make_device_vector_view<DataT, IndexT>(L2NormBuf_OR_DistBuf.data(), n_clusters));
+    auto centroidsNorm =
+      raft::make_device_vector_view<DataT, IndexT>(L2NormBuf_OR_DistBuf.data(), n_clusters);
+
+    if (metric == cuvs::distance::DistanceType::CosineExpanded) {
+      // Centroids are L2-normalized for cosine metric
+      raft::matrix::fill(handle, centroidsNorm, DataT{1});
+    } else {
+      raft::linalg::norm<raft::linalg::L2Norm, raft::Apply::ALONG_ROWS>(
+        handle, centroids, centroidsNorm);
```
dantegd, line 47: I think the coderabbit comment indeed was wrong,  non-balanced kmeans::fit isn't actually a cosine caller  and this PR doesn't change that. 

One thing I wasn't sure this resolves, though: what about the public kmeans_balanced::predict (kmeans_balanced.cuh:130)? If a user calls it with their own centroids under CosineExpanded, which seems reasonable since cosine is scale-invariant, would they now silently get wrong labels, given the fused helper divides by yn=1 instead of ‖c‖? I don't see the unit-norm precondition documented on the signature or in /include/.../kmeans.hpp, so I'm wondering whether it's worth either threading a centroids_are_unit_norm flag (default false, balanced inner loop passes true) or normalizing into a scratch buffer at the public entry. Would either fit here?

cpp/src/cluster/detail/minClusterDistanceCompute.cu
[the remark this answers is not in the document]
dantegd: One thing I think would be worth leaving for future us: when non-balanced kmeans grows proper cosine support, whoever does that work will need to know that minClusterAndDistanceCompute / minClusterDistanceCompute now quietly assume pre-normalized centroids on the cosine fused path. Without that context it'd be easy to wire cosine in, skip the centroid normalize (since cosine is "scale-invariant"), and ship subtly wrong distances.

Would it be worth a short TODO/comment near the raft::matrix::fill(handle, centroidsNorm, DataT{1}) block flagging that the unit-norm assumption would need to be revisited (either by an opt-in flag or by pre-normalizing at the caller) before any non-balanced cosine support could reuse this helper?

cpp/src/cluster/detail/minClusterDistanceCompute.cu
```diff
@@ -11,7 +12,9 @@ namespace cuvs::cluster::kmeans::detail {
 
 // Calculates a <key, value> pair for every sample in input 'X' where key is an
 // index to an sample in 'centroids' (index of the nearest centroid) and 'value'
-// is the distance between the sample and the 'centroid[key]'
+// is the distance between the sample and the 'centroids[key]'.
+//
+// NB: (CosineExpanded): `centroids` rows must be L2-normalized when the cosine metric is used.
```
dantegd: These comments are only visible when reading the implementation. Other translation units and Doxygen see only kmeans_common.cuh, no? Which has no mention of the cosine unit-norm precondition. It'd be a good idea duplicate the NB onto both declarations here so it is visible at the call site.

cpp/tests/cluster/kmeans_balanced.cu
```diff
@@ -223,6 +247,9 @@ KB_TEST((KmeansBalancedTest<float, float, int, int64_t, raft::identity_op, true>
 // KB_TEST((KmeansBalancedTest<float, float, int64_t, int64_t, raft::identity_op>),
 //         KmeansBalancedTestFFI64I64,
 //         inputsf_i64);
+KB_TEST((KmeansBalancedTest<float, float, uint32_t, int, raft::identity_op, false>),
+        KmeansBalancedTestCosineFFU32I32,
+        inputsf_cosine_i32);
```
dantegd, line 252: Quick one on the test coverage, the new cosine case uses SeparateFitPredict = false, so it only exercises fit_predict. That path keeps using the same centers fit just finished normalizing, so it won't catch it if the post-loop row_normalize in balancing_em_iters ever gets lost in a refactor (which is the whole invariant this PR is leaning on now that ivf_pq_build dropped its manual normalize). Would it be easy to also add a SeparateFitPredict = true variant? Something like 

```C++
KB_TEST((KmeansBalancedTest<float, float, uint32_t, int, raft::identity_op, true>),
        KmeansBalancedTestCosineSeparateFFU32I32,
        inputsf_cosine_i32);
```

cpp/src/neighbors/ivf_pq/ivf_pq_build.cuh
```diff
@@ -1329,9 +1329,6 @@ auto build(raft::resources const& handle,
     rmm::device_uvector<uint32_t> labels(n_rows_train, stream, big_memory_resource);
     auto centers_const_view = raft::make_device_matrix_view<const float, internal_extents_t>(
       cluster_centers, impl->n_lists(), impl->dim());
-    if (impl->metric() == distance::DistanceType::CosineExpanded) {
-      raft::linalg::row_normalize<raft::linalg::L2Norm>(handle, centers_const_view, centers_view);
-    }
```
dantegd, line 1334: Removing this normalize is correct today because kmeans_balanced::fit now guarantees unit-norm centers for CosineExpanded, right?. 

But there's no comment or assertion tying these two files together, if balancing_em_iters's post-loop normalize ever gets lost in a refactor, IVF-PQ silently regresses recall with no failing test anywhere. It would be a good idea to leave a short comment at both sites (here and in balancing_em_iters) noting the dependency

dantegd (changes requested):
This refactor is really nice, just had a few comments

cpp/src/cluster/detail/kmeans_balanced.cuh
```diff
@@ -5,7 +5,6 @@
```
viclafargue, line 5: Please edit the docstrings of `predict` overloads taking a `balanced_params` parameters in `cpp/include/cuvs/cluster/kmeans.hpp` to clearly state that the centroids should be normalized when using the cosine metric. A user may have trained centroids elsewhere and may be attempting to GPU-accelerate the prediction.

viclafargue (commented):
Thanks! one comment

tarang-jain:
@dantegd @viclafargue I got rid of the unit norm assumption and restored the previous version. We can revisit the unit norm assumption later, because this will make the user experience difficult.

viclafargue (changes requested):
The PR description should be updated.
Also every callers of `minClusterAndDistanceCompute` / `minClusterDistanceCompute` aside from the Balanced KMeans should be guarded against the use of the cosine metric (notably Batched KMeans `fit`, MG KMeans `fit` and regular KMeans `kmeans_fit` and `kmeans_predict`).

tarang-jain:
> Also every callers of minClusterAndDistanceCompute / minClusterDistanceCompute aside from the Balanced KMeans should be guarded against the use of the cosine metric (notably Batched KMeans fit, MG KMeans fit and regular KMeans kmeans_fit and kmeans_predict).

Done!

viclafargue (approved):
Thanks! LGTM

dantegd (approved):

cpp/src/cluster/detail/kmeans.cuh
```diff
@@ -729,6 +732,9 @@ void kmeans_fit(raft::resources const& handle,
                 raft::host_scalar_view<DataT> inertia,
                 raft::host_scalar_view<IndexT> n_iter)
 {
+  RAFT_EXPECTS(pams.metric == cuvs::distance::DistanceType::L2Expanded ||
+                 pams.metric == cuvs::distance::DistanceType::L2SqrtExpanded,
+               "kmeans only supports L2Expanded or L2SqrtExpanded distance metrics.");
```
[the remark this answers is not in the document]
tarang-jain: ivf-pq uses kmeans_balaced, which supports cosine.

PR #2005, opened 2026-04-09
Exposition of KMeans param object for PQ in C++

### LINKED ISSUE #1999: [FEA] Expose kmeans param object for PQ preprocessing
Currently the PQ preprocessing only expose the number of iterations of kmeans. A request was made to expose the init method of kmeans classic.
Instead of just exposing that parameter we can expose the whole kmeans param object using a std::variant for classic/balanced structs.

lowener:
Closes #1999.
Behavior change:
- The pq-kmeans params now follows the default kmeans params. For classical kmeans the init algorithm is now set to Kmeans++

Breaking change:
- pq::params now uses a constructor


cpp/src/preprocessing/quantize/detail/pq.cuh
```diff
@@ -344,6 +365,24 @@ void vpq_convert_math_type(const raft::resources& res,
...
+    in_params.pq_dim,
+    true,
+    true,
+    in_params.vq_n_centers,
+    in_params.kmeans_n_iters,
+    in_params.pq_kmeans_type,
+    std::min<uint32_t>(in_params.max_train_points_per_pq_code,
+                       n_rows * in_params.pq_kmeans_trainset_fraction / pq_n_centers),
+    std::min<uint32_t>(in_params.max_train_points_per_vq_cluster,
+                       n_rows * in_params.vq_kmeans_trainset_fraction / in_params.vq_n_centers)};
```
tarang-jain: Could vq_n_centers be set to zero. Perhaps we can just add a guard to avoid zero division.

tarang-jain (commented):
LGTM. Thanks for this! I just added a comment about one small thing. Apart from that, I have no issues with this PR.

cpp/src/preprocessing/quantize/detail/pq.cuh
lowener: The call just before to `fill_missing_params_heuristics` ensure that vq_n_centers is not zero. I still added a guard in case that function gets used elsewhere in the future.

cpp/src/neighbors/detail/vpq_dataset.cuh
```diff
@@ -74,50 +75,51 @@ namespace cuvs::neighbors::detail {
...
-      sub_labels_view,
-      pq_cluster_sizes_view,
-      cuvs::spatial::knn::detail::utils::mapping<MathT>{});
-  } else {
-    const auto pq_n_centers = pq_centers_view.extent(0);
-    cuvs::cluster::kmeans::params kmeans_params;
-    kmeans_params.n_clusters = pq_n_centers;
-    kmeans_params.max_iter   = params.kmeans_n_iters;
-    kmeans_params.metric     = cuvs::distance::DistanceType::L2Expanded;
-    kmeans_params.init       = cuvs::cluster::kmeans::params::InitMethod::Random;
```
viclafargue, line 108: The function used to set the initialization method to random, it now uses the field from the argument struct. This is an undocumented behavior change. Please add it to the PR description. Also, would you know why this was initially set this way?

c/src/preprocessing/quantize/pq.cpp
```diff
@@ -62,18 +62,10 @@ void* _build(cuvsResources_t res,
...
-    .vq_n_centers = params->vq_n_centers,
-    .kmeans_n_iters = params->kmeans_n_iters,
-    .pq_kmeans_type = static_cast<cuvs::cluster::kmeans::kmeans_type>(params->pq_kmeans_type),
-    .max_train_points_per_pq_code = params->max_train_points_per_pq_code,
-    .max_train_points_per_vq_cluster = params->max_train_points_per_vq_cluster
-  };
+  cuvs::preprocessing::quantize::pq::params quantizer_params(
+    params->pq_bits, params->pq_dim, params->use_subspaces, params->use_vq, params->vq_n_centers,
+    params->kmeans_n_iters,  static_cast<cuvs::cluster::kmeans::kmeans_type>(params->pq_kmeans_type), params->max_train_points_per_pq_code,
+     params->max_train_points_per_vq_cluster);
```
viclafargue, line 68: The struct constructor makes it impossible to initialize the struct as it could be done before :
```
auto quantizer_params = cuvs::preprocessing::quantize::pq::params{
.pq_bits = 8,
.use_vq = true
};
```

This may be considered a breaking change, since users may be using this to initialize the parameters.

cpp/include/cuvs/preprocessing/quantize/pq.hpp
```diff
@@ -5,22 +5,77 @@
...
+         cuvs::cluster::kmeans::kmeans_type pq_kmeans_type,
+         uint32_t max_train_points_per_pq_code,
+         uint32_t max_train_points_per_vq_cluster)
+  {
+    this->pq_bits       = pq_bits;
+    this->pq_dim        = pq_dim;
+    this->use_subspaces = use_subspaces;
+    this->use_vq        = use_vq;
+    this->vq_n_centers  = vq_n_centers;
+    this->kmeans_params =
```
viclafargue, line 49: Prefer member-initializer lists over `this->x = x` inside constructor body.

cpp/include/cuvs/preprocessing/quantize/pq.hpp
```diff
@@ -5,22 +5,77 @@
...
+  /**
+   * Simplified constructor that will build an appropriate kmeans params object.
+   */
+  params(uint32_t pq_bits,
+         uint32_t pq_dim,
+         bool use_subspaces,
+         bool use_vq,
+         uint32_t vq_n_centers,
+         uint32_t kmeans_n_iters,
+         cuvs::cluster::kmeans::kmeans_type pq_kmeans_type,
```
viclafargue, line 40:
```suggestion
         const cuvs::cluster::kmeans::kmeans_type pq_kmeans_type&,
```

The parameters could be a const-ref.

cpp/src/neighbors/detail/vpq_dataset.cuh
```diff
@@ -219,7 +221,7 @@ auto predict_vq(const raft::resources& res,
 
 template <typename MathT, typename DatasetT>
 auto train_pq(const raft::resources& res,
-              const vpq_params& params,
+              const cuvs::preprocessing::quantize::pq::params params,
```
viclafargue, line 224:
```suggestion
              const cuvs::preprocessing::quantize::pq::params& params,
```

cpp/src/neighbors/detail/vpq_dataset.cuh
```diff
@@ -74,50 +75,51 @@ namespace cuvs::neighbors::detail {
...
-                               pq_centers_view,
-                               raft::make_host_scalar_view<MathT>(&inertia),
-                               raft::make_host_scalar_view<IdxT>(&n_iter));
-  }
+  std::visit(
+    [&](auto const& base_kmeans_params) {
+      using KP = std::decay_t<decltype(base_kmeans_params)>;
+      if constexpr (std::is_same_v<KP, cuvs::cluster::kmeans::balanced_params>) {
+        auto bal_params   = base_kmeans_params;
+        bal_params.metric = cuvs::distance::DistanceType::L2Expanded;
```
viclafargue, line 89: Please add a `RAFT_EXPECTS` check to ensure that the struct was already using the `L2Expanded` metric. A user may attempt to use PQ with an incompatible dataset/metric.

Also, it could be interested to check that `n_clusters == 1 << pq_bits`.

cpp/src/neighbors/detail/vpq_dataset.cuh
```diff
@@ -74,50 +75,51 @@ namespace cuvs::neighbors::detail {
...
+          bal_params,
+          pq_trainset_view,
+          pq_centers_view,
+          sub_labels_view,
+          pq_cluster_sizes_view,
+          cuvs::spatial::knn::detail::utils::mapping<MathT>{});
+      } else {
+        auto classic_params       = base_kmeans_params;
+        classic_params.n_clusters = pq_centers_view.extent(0);
+        classic_params.metric     = cuvs::distance::DistanceType::L2Expanded;
```
viclafargue, line 107: Similar note. Please add a `RAFT_EXPECTS` check to ensure that the struct was already using the `L2Expanded` metric. A user may attempt to use PQ with an incompatible dataset/metric.

cpp/src/neighbors/detail/vpq_dataset.cuh
```diff
@@ -74,50 +75,51 @@ namespace cuvs::neighbors::detail {
...
+        RAFT_EXPECTS(classic_params.init != cuvs::cluster::kmeans::params::InitMethod::Array,
+                     "Array initialization is not supported for PQ training");
```
viclafargue, line 109: Shouldn't we move all the validation checks earlier in the call chain?

cpp/tests/preprocessing/product_quantization.cu
```diff
@@ -160,14 +160,21 @@ class ProductQuantizationTest : public ::testing::TestWithParam<ProductQuantizat
...
-    config_      = cuvs::preprocessing::quantize::pq::params(params_.pq_bits,
+    using LabelT                = uint32_t;
+    using kmeans_params_variant = cuvs::preprocessing::quantize::pq::kmeans_params_variant;
+    kmeans_params_variant kmeans_params =
+      (params_.pq_kmeans_type == cuvs::cluster::kmeans::kmeans_type::KMeansBalanced)
+        ? kmeans_params_variant(cuvs::cluster::kmeans::balanced_params{.n_iters = 25})
+        : kmeans_params_variant(
+            cuvs::cluster::kmeans::params{.n_clusters = 1 << params_.pq_bits,
+                                          .init = cuvs::cluster::kmeans::params::InitMethod::Random,
+                                          .max_iter = 25});
```
viclafargue, line 171: Could the test check that the newly available parameters fields (`init` method for instance) are actually honored?

viclafargue (commented):
Thanks! Here are some comments.

Also, shouldn't this feature be made available in C/Python?

cpp/include/cuvs/preprocessing/quantize/pq.hpp
lowener: It's an enum so an int. It's passed by copy so it doesn't change anything.

cpp/src/neighbors/detail/vpq_dataset.cuh
lowener: I added it to the PR description.
Because it's faster for build() and while testing I didn't notice major drop of clustering quality, but it was on synthetic dataset.
On some other datasets there is a drop of clustering quality so it will now be set to the kmeans default and available to modify

cpp/src/neighbors/detail/vpq_dataset.cuh
lowener: `n_clusters` is always infered from pq_bits. I think it's not even necessary to check it since there are no other value possible. It can just be set.

c/src/preprocessing/quantize/pq.cpp
lowener: Yes it is a breaking change and the PR is tagged `breaking`

cpp/src/preprocessing/quantize/detail/pq.cuh
```diff
@@ -344,6 +382,29 @@ void vpq_convert_math_type(const raft::resources& res,
...
+  return cuvs::preprocessing::quantize::pq::params{
+    in_params.pq_bits,
+    in_params.pq_dim,
+    true,
+    true,
+    in_params.vq_n_centers,
+    in_params.kmeans_n_iters,
+    in_params.pq_kmeans_type,
+    std::min<uint32_t>(in_params.max_train_points_per_pq_code,
+                       n_rows * in_params.pq_kmeans_trainset_fraction / pq_n_centers),
```
viclafargue, line 405: Just a detail, but integer truncation can make this calculation imprecise.

cpp/include/cuvs/preprocessing/quantize/pq.hpp
```diff
@@ -5,22 +5,78 @@
...
+    : pq_bits(pq_bits),
+      pq_dim(pq_dim),
+      use_subspaces(use_subspaces),
+      use_vq(use_vq),
+      vq_n_centers(vq_n_centers),
+      kmeans_params(
+        pq_kmeans_type == cuvs::cluster::kmeans::kmeans_type::KMeansBalanced
+          ? kmeans_params_variant{cuvs::cluster::kmeans::balanced_params{.n_iters = kmeans_n_iters}}
+          : kmeans_params_variant{cuvs::cluster::kmeans::params{.n_clusters = 1 << pq_bits,
+                                                                .max_iter = (int)kmeans_n_iters}}),
```
viclafargue:
```suggestion
 .max_iter = static_cast<int>(kmeans_n_iters)}}),
```

cpp/src/preprocessing/quantize/detail/pq.cuh
```diff
@@ -344,6 +382,29 @@ void vpq_convert_math_type(const raft::resources& res,
                     raft::make_const_mdspan(src.pq_code_book.view()));
 }
 
+inline auto from_vpq_params(const cuvs::neighbors::vpq_params& in_params, const uint64_t n_rows)
```
viclafargue: `make_pq_params_for_train_pq` would maybe be a more appropriate name for this function as it builds a `pq::params` suitable for the `train_pq` call.

viclafargue (approved):
Thanks! LGTM

Noting that the PR does not forward the feature downstream to the C and Python layer. The PR would need a rename to clearly state that it is C++ only. Also, it may be interesting to open an issue if we plan to implement the bindings at a later date.

cpp/tests/preprocessing/product_quantization.cu
```diff
@@ -246,6 +253,65 @@ class ProductQuantizationTest : public ::testing::TestWithParam<ProductQuantizat
   params config_;
 };
 
+TEST(ProductQuantizationTestF, Parameters)
```
tarang-jain, line 256: I have not seen tests that simply check if parameter are set. I think we should be mindful with CI resources already overloaded with cpp tests.

tarang-jain (approved):
Added one small comment. But otherwise approving!

PR #2011, opened 2026-04-10
Add option for inline generation matrix

KyleFromNVIDIA:
Rather than putting the generation matrix in its own file, allow developers to include it in the same file so that all of the placeholders are visible at once.


KyleFromNVIDIA:
Closing as we decided offline that putting JSON inside the files looks weird, and it would be best to keep it separate.

PR #2012, opened 2026-04-10
[REVIEW] Add OpenSearch backend to ``cuvs-bench``

jrbourbeau:
Would still like to refine a bit but pushing up for early feedback 

cc @singhmanas1 @janakivamaraju @afourniernv

xref https://github.com/rapidsai/cuvs/pull/1907 which does something similar but for Elastic

EDIT: Okay, this is now ready for review (cc @jnke2016)


python/cuvs_bench/cuvs_bench/orchestrator/orchestrator.py
```diff
@@ -222,43 +222,47 @@ def _run_sweep(
...
-                    force=force,
-                    dry_run=dry_run,
-                )
-                results.append(build_result)
-
-                if not build_result.success:
-                    print(
-                        f"Build failed for {config.index_name}: {build_result.error_message}"
+            try:
+                backend.initialize()
```
jrbourbeau, line 226: Note to reviewers: This is a separate, but related change to the orchestrator. I noticed the `initialize` and `cleanup` methods weren't actually being called like their docstrings said they were suppose to. This is a big diff but most of it is just intending over the existing code into a try/except block. The functional change is calling `initialize` + `cleanup`

EDIT: FWIW supporting a context manager for backend setup / teardown seems reasonable. But that's also not something I wanted to handle in this PR.

python/cuvs_bench/pyproject.toml
```diff
@@ -38,12 +38,20 @@ classifiers = [
...
+
 [project.urls]
 Homepage = "https://github.com/rapidsai/cuvs"
 
 [tool.setuptools.package-data]
 "*" = ["*.*", "VERSION"]
 
+[tool.pytest.ini_options]
+markers = [
+    "integration: tests that require a live OpenSearch node (run with '-m integration')",
```
jrbourbeau, line 52: TODO: Make this more generic language (not just opensearch)

python/cuvs_bench/cuvs_bench/orchestrator/orchestrator.py
jnke2016: Good catch. The `initialize()` and `cleanup()` hooks in BenchmarkBackend are defined in the abstract base class with docstrings describing their intended purpose, but neither method is actually called by the `BenchmarkOrchestrator` in `_run_sweep()` or `_run_trial()`. This was an oversight because the only backend implemented at the time was `CppGoogleBenchmarkBackend`, whose lifecycle is entirely self-contained within each subprocess call to `build()`/`search()`, making `initialize()`/`cleanup()` unnecessary for it.

jnke2016: > EDIT: FWIW supporting a context manager for backend setup / teardown seems reasonable. But that's also not something I wanted to handle in this PR.

I Agree that context manager support is reasonable. The purpose of the pluggable backend API is precisely to enable these kinds of extensions without modifying the core orchestrator. Since initialize() and cleanup() are already part of the BenchmarkBackend interface, wiring them up in the orchestrator (whether as explicit calls or a context manager) would ensure the plugin system works end-to-end for future backends without requiring further modifications to the core

jnke2016 (approved):
The PR looks good me.

python/cuvs_bench/cuvs_bench/backends/opensearch.py
```diff
@@ -0,0 +1,995 @@
...
+        ".fbin": np.float32,
+        ".f16bin": np.float16,
+        ".u8bin": np.uint8,
+        ".i8bin": np.int8,
+        ".ibin": np.int32,
+    }
+
+    ext = os.path.splitext(path)[1].lower()
+    dtype = _DTYPE_FOR_EXT.get(ext, np.float32)
+    with open(path, "rb") as f:
```
cjnolet, line 48: I mentioned this in the elasticsearch PR as well- we should standardize/centralize these calls so that we're not having to manually implement them for every backend. The benchmark should be opening the files and passing the vectors or the file handle) down to the actual backends. 

@jnke2016

python/cuvs_bench/cuvs_bench/backends/opensearch.py
```diff
@@ -0,0 +1,995 @@
...
+        n_rows = int(np.frombuffer(f.read(4), dtype=np.uint32)[0])
+        n_cols = int(np.frombuffer(f.read(4), dtype=np.uint32)[0])
+        if subset_size is not None:
+            n_rows = min(n_rows, subset_size)
+        raw = f.read(n_rows * n_cols * np.dtype(dtype).itemsize)
+        data = np.frombuffer(raw, dtype=dtype)
+    return data.reshape(n_rows, n_cols)
+
+
+def _load_yaml(path: str) -> Any:
```
cjnolet, line 58: This and the dataset config should also be getting loaded by the benchmark itself and not the backend. @jnke2016

python/cuvs_bench/cuvs_bench/backends/opensearch.py
```diff
@@ -0,0 +1,995 @@
...
+    for d in all_confs:
+        if dataset == d["name"]:
+            return d
+    raise ValueError(
+        f"Could not find a dataset configuration for {dataset!r} in datasets.yaml"
+    )
+
+
+def _gather_algo_configs(
+    config_path: str, algorithm_configuration: Optional[str]
```
cjnolet, line 73: We need to centralize this @jnke2016

python/cuvs_bench/cuvs_bench/backends/opensearch.py
```diff
@@ -0,0 +1,995 @@
...
+            "remote_build_s3_prefix",
+            "remote_build_s3_access_key",
+            "remote_build_s3_secret_key",
+            "remote_build_timeout",
+        )
+        conn_kwargs = {k: kwargs[k] for k in _conn_keys if k in kwargs}
+
+        benchmark_configs: List[BenchmarkConfig] = []
+
+        for algo_file in algo_files:
```
cjnolet, line 245: Yeah, this is all boilerplate- exactly why this should be centralized. The resulting config dicts should be loaded from the files by the benchmark and passed down to the individual backends. It's important that this is standardized across benchmarks so that we don't have specific backends doing things drastically different than others (that just makes it super complicated for everyone. But also, if we find a bug in a spcific loader or decide to change how we do things generally, every single backend then needs to be updated individually instead of just changing it one place ) @jnke2016

python/cuvs_bench/cuvs_bench/backends/opensearch.py
```diff
@@ -0,0 +1,995 @@
...
+            if faiss_method == "hnsw":
+                method_params = {
+                    "m": m,
+                    "ef_construction": ef_construction,
+                }
+            elif faiss_method == "ivf":
+                method_params = {"nlist": build_param.get("nlist", 4)}
+            else:
+                raise ValueError(
+                    f"Unsupported faiss_method {faiss_method!r}. "
```
cjnolet, line 483: Yes! This is important. I believe i suggested that this be done in the eladticserch PR as well. Need to fail gracefully when the user configures somethign that's not supported (or loads the wrong file).

cjnolet:
cc @navneet1v for thoughts / feedback on the opensearch backend for cuvs-bench.

python/cuvs_bench/cuvs_bench/backends/opensearch.py
jrbourbeau: Don't disagree with this comment (or the other ones about consolidating various utilities). IIUC it sounds like these will need deeper changes to the config loader / orchestrator specs. Is that something that's needed in this PR, or is a follow up to refactor fine?

python/cuvs_bench/cuvs_bench/orchestrator/orchestrator.py
cjnolet: Overall, I think we're leaning a little too strongly on the local (individual backend) vs global (benchmark orchestrator) control and we should probably prioritize global control for all boilerplate things so that we're not having to reimplement them in each backend. I'm okay doing this in follow-ups, but the way we typically handle these types of things is to create a github issue that clearly explains the intended changes for the follow-up (explain it in a way that assumes someone else might have to pick it up in the future, just in case) and then reference a link to the issue in a todo comment in the code where the change is intended to be made. This makes sure 1) people reading the code are aware of the intentions, and 2) the issue doesn't get orphaned and forgotten.

python/cuvs_bench/cuvs_bench/backends/opensearch.py
jnke2016: So the loading pipeline (loading dataset YAML, gathering algorithm files, filtering by algorithm/group names etc ..) is a shared logic that should be centralized in the base ConfigLoader instead of having each backend implementing this.

python/cuvs_bench/cuvs_bench/backends/opensearch.py
jnke2016: Since both backends(elastic search and OpenSearch) are separately implementing this, a follow up PR consolidating both might be easier especially since both PRs are still open. Otherwise I can push a PR consolidating those various utilities (general purpose) based on both PRs which they can pull after it is merged

jnke2016: @jrbourbeau 
Loading vectors in the orchestrator or config loader isn't the right approach because the C++ backend never loads vectors into Python. It only passes file paths to the C++ subprocess. Loading vectors at the orchestrator level would waste memory for the C++ backend and any future backend that operates the same way, so this doesn't require changes to the config loader or orchestrator specs.

There are two options to centralize this:

Option 1: Shared utility module. Create a backends/utils.py with a single load_vectors() function that handles all supported binary formats (.fbin, .f16bin, .u8bin, .i8bin, .ibin) and supports subset_size. Python-native backends like OpenSearch and Elasticsearch import and call it when they need vectors. The C++ backend, orchestrator, and config loader are completely untouched. This is the simpler approach with no core changes, though it relies on future backend authors knowing to use it.

Option 2: Lazy loading on Dataset. Make base_vectors, query_vectors, and groundtruth_neighbors lazy properties on the Dataset class that load from the corresponding file paths on first access. The C++ backend never accesses these properties (it uses file paths), so nothing loads. Python-native backends just access dataset.base_vectors and get loaded vectors automatically without needing to know about any utility or handle the empty-array fallback pattern. This prevents the duplication from recurring since future backend authors can't accidentally reimplement it, but it requires changing Dataset from a plain dataclass to a regular class.

I'd suggest we go with option 1 for now since it solves the problem without touching core types, and we can move to option 2 later if the pattern keeps getting reimplemented

python/cuvs_bench/cuvs_bench/backends/opensearch.py
jnke2016: Agreed. Functions like `_gather_algo_configs`, `_load_yaml`, and `_get_dataset_conf` are shared config-loading logic that every backend ends up reimplementing. @afourniernv - The Elasticsearch  PR has the same pattern. The C++ loader already has equivalent methods (load_yaml_file, get_dataset_configuration, gather_algorithm_configs) as instance methods. I'll promote those to the base ConfigLoader class so all config loaders inherit them automatically and developers don't need to reimplement them. The backend should only be responsible for its backend-specific concerns (connection parameters, index mappings, API calls), not for loading YAML files or finding datasets by name.

jnke2016: Though the load() method is abstract because each backend has different config-loading logic, helper methods like load_yaml_file and get_dataset_configuration are common operations that every config loader reimplements when implementing load(). gather_algorithm_configs is also shared between the C++ and OpenSearch loaders.

python/cuvs_bench/cuvs_bench/backends/opensearch.py
```diff
@@ -0,0 +1,1070 @@
...
+            if allowed_groups:
+                groups = {
+                    k: v for k, v in groups.items() if k in allowed_groups
+                }
+
+            for group_name, group_conf in groups.items():
+                build_spec: Dict[str, List] = group_conf.get("build", {})
+                search_spec: Dict[str, List] = group_conf.get("search", {})
+
+                build_keys = list(build_spec.keys())
```
jnke2016: **Cartesian product expansion**: Both backends (`Elasticsearch `and `OpenSearch`) implement the same param expansion pattern via Cartesian product). I've added expand_param_grid() in `backends/utils.py` that both can use instead of reimplementing. For OpenSearch, the expansion block at lines 261-276 would be replaced with:

```
from .utils import expand_param_grid

build_combos = expand_param_grid(build_spec)
search_params_list = expand_param_grid(search_spec)
```

python/cuvs_bench/cuvs_bench/backends/opensearch.py
```diff
@@ -0,0 +1,1070 @@
...
+                resp = self._client.search(index=index_name, body=body)
+                hits = resp["hits"]["hits"]
+                for j, hit in enumerate(hits[:k]):
+                    neighbors[i, j] = int(hit["_id"])
+                    distances[i, j] = float(hit["_score"])
+
+            elapsed = time.perf_counter() - t0
+            qps = n_queries / elapsed if elapsed > 0 else 0.0
+
+            recall = 0.0
```
jnke2016: **Recall calculation**: I've added a shared compute_recall() in `backends/utils.py` so that any bug fix or improvement happens in one place and applies to all Python-native backends. The inline recall block at lines 1027-1035 would be replaced with:
```
from .utils import compute_recall

recall = 0.0
if groundtruth is not None:
    recall = compute_recall(neighbors, groundtruth, k)
```

cjnolet: Running the parameters and calculating recall should be done by the orchestrator not the individual backends. The orchestrator should be controlling this end to end. Backends should not even have the option to do this because the orchestrator should be the one controlling it all.

Please look at how Ann-benchmarks is architected. That's what we want to do here (and that's what the original cuVS-bench scripts were doing.

jnke2016: The original cuVS-bench scripts had the framework control everything end-to-end, with recall computed by the C++ subprocess. This is still the case in the new pluggable benchmark API for the C++ backend. Python-native backends (OpenSearch, Elasticsearch) don't have this since they return raw neighbor IDs over HTTP. Moving recall computation to the orchestrator would require handling the case where the C++ backend already returns recall, to avoid overwriting it. One way to distinguish this is by checking neighbors.size > 0, but that relies on the C++ backend returning empty neighbor arrays rather than an explicit signal, and if any other backend also returns empty neighbors for a different reason, the orchestrator would wrongly skip recall computation. We could also hardcode an exception for the C++ backend in the orchestrator, but that goes against the pluggable design since any future backend that also computes recall internally would need another exception.

jnke2016: I looked at how `ann-benchmarks` handles this. In ann-benchmarks, every algorithm returns only raw neighbor indices. The framework stores them as-is, and recall is computed downstream during analysis, never by the algorithm itself as you pointed out. In our case, the C++ backend is the exception because its subprocess design doesn't return raw neighbors to Python. It computes recall internally and returns it in the JSON output. Following the ann-benchmarks pattern, the orchestrator would compute recall for any backend that returns actual neighbor arrays (neighbors.size > 0), and preserve the C++ backend's recall when it returns empty neighbors. This was the main reason I initially kept recall in the backends rather than the orchestrator. Relying on neighbors.size > 0 to skip the C++ backend is fragile if other backends also return empty neighbors for different reasons. I will have recall computed at the orchestrator level and skipping it for the C++ backend (through neighbors.size > 0) and I am open to suggestions.

python/cuvs_bench/cuvs_bench/backends/opensearch.py
jnke2016: Went with option 2. The Dataset class is now a regular class where `base_vectors`, `query_vectors`, and `groundtruth_neighbors` are loaded automatically from file paths on first access. Backends just use `dataset.base_vectors` and get a numpy array without needing to know about file loading. The C++ backend is unaffected since it only uses `dataset.base_file`.

python/cuvs_bench/cuvs_bench/backends/opensearch.py
jnke2016: `recall` is now computed by the orchestrator after `backend.search()` returns. Backends no longer compute it themselves. The C++ backend is naturally skipped since it returns empty neighbor arrays (`recall` is already computed by the subprocess).

jrbourbeau: Updated to have the orchestrator compute recall

jrbourbeau (commented):
Okay, I've updated the opensearch backend here to use the new centralized utilities / updated structure from https://github.com/rapidsai/cuvs/pull/2040 and https://github.com/rapidsai/cuvs/pull/2084

@jnke2016 @cjnolet could you take a look when you get a moment? 

cc @singhmanas1 for viz

python/cuvs_bench/cuvs_bench/backends/opensearch.py
```diff
@@ -0,0 +1,941 @@
...
+                    index_path=index_name,
+                    build_time_seconds=0.0,
+                    index_size_bytes=0,
+                    algorithm=self.algo,
+                    build_params=build_param,
+                    success=True,
+                )
+
+        # Dataset handles lazy vector loading from base_file when needed.
+        base_vectors = dataset.training_vectors
```
jnke2016, line 736: Just as intended: The backend just uses `dataset.training_vectors` without worrying about file formats, loading logic, or how the data got there. The framework handles all of that transparently.

python/cuvs_bench/cuvs_bench/backends/opensearch.py
```diff
@@ -0,0 +1,941 @@
...
+            print(
+                f"[dry_run] Would search OpenSearch index '{index_name}' with {len(search_params_list)} param set(s) (k={k})"
+            )
+
+            return SearchResult(
+                neighbors=np.zeros((0, k), dtype=np.int64),
+                distances=np.zeros((0, k), dtype=np.float32),
+                search_time_ms=0.0,
+                queries_per_second=0.0,
+                recall=0.0,
```
jnke2016, line 878: Good, the backend returns `recall=0.0` as a placeholder. The orchestrator then computes the actual recall from the returned neighbors and ground truth, so all backends are evaluated the same way without each one implementing its own recall logic

jrbourbeau (commented):
@cjnolet @jnke2016 can you give this a final review when you get a moment?

python/cuvs_bench/pyproject.toml
cjnolet: You can do this in a follow-up, but I think it would be a good idea to add a blurb to the cuvs-bench install documentation for running the integration tests. Maybe we could do that in the opensearch deployment PR.

cjnolet (approved):

python/cuvs_bench/pyproject.toml
jameslamb: I'd also say "integration" is a bit overly-general of a name if you intend to add more backends in the future. `-m opensearch` (or something similarly specific) would be a bit more future-proof.

python/cuvs_bench/pyproject.toml
```diff
@@ -38,12 +38,20 @@ classifiers = [
     "Programming Language :: Python :: 3.14",
 ]
 
+[project.optional-dependencies]
+opensearch = ["opensearch-py>=2.4.0"]
```
jameslamb: We generally prefer to have all dependencies managed via `dependencies.yaml` (the config file for https://github.com/rapidsai/dependency-file-generator), to improve consistency and for flexibility in generating dependency lists in different formats.

Could you please move this there?

Here's an example to follow:

https://github.com/rapidsai/cuvs/blob/4addf4e6c1da42743d5701e5ad10764333317732/dependencies.yaml#L185-L194

https://github.com/rapidsai/cuvs/blob/4addf4e6c1da42743d5701e5ad10764333317732/dependencies.yaml#L520-L535

If you get stuck and would prefer I just directly push to this branch, let me know.

jameslamb: Also is there a `conda` equivalent of this and if so do we want it as a dependency of the `cuvs-bench` package?

jameslamb (changes requested):
Left some small suggestions for your consideration.

python/cuvs_bench/pyproject.toml
jrbourbeau: Thanks for the feedback @jameslamb. I just pushed a commit with the changes that I _think_ you're asking for. But let me know if I missed something, or you're asking for something different -- it's my first time mucking around with the `dependencies.yaml` config file

jameslamb (approved):
Changes look great, thanks!

PR #2015, opened 2026-04-10
[Cleanup] Combine Batched and Regular KMeans Impl

tarang-jain:
## Combine batched and regular k-means implementations

- Unified the batched (host-data) and regular (device-data) k-means `fit` into a single `kmeans_fit` template that works with both host and device mdspans via `batch_load_iterator`
- Unified the device and host initialization paths in `init_centroids`
- Removed the `inertia_check` parameter — inertia-based convergence checking now always runs. Zero clustering cost (perfect fit) logs a warning instead of asserting. This is needed because spectral clustering can cause all points to converge on the cluster centroids itself.
- Added `init_size` parameter to control how many samples are drawn for KMeansPlusPlus initialization. Defaults to `n_samples` for device data, `(3 * n_clusters)` for host data
- Replaced per-iteration centroid `raft::copy` with `std::swap` of buffer pointers
- For streaming fit, precompute data norms once and cache them: host norms cached to a host buffer on the first iteration and copied back for subsequent iterations. `process_batch` no longer computes norms internally
- Replaced raw `cudaPointerGetAttributes` call with `raft::memory_type_from_pointer`
- Replaced `cub::DeviceReduce::Sum` calls with `raft::linalg::mapThenSumReduce`
- Guarded weight normalization against overflow: apply `(w / wt_sum) * n_samples` via a composed op instead of precomputing a scale, so very small `wt_sum` values don't produce inf
- Renamed `checkWeight` to `weightSum` and made it mdspan-based with an `Accessor` template: device reduce for device weights, host loop for host weights. Callers apply the scaling themselves
- Eliminated `batch_sums` / `batch_counts` scratch buffers by accumulating directly into `centroid_sums` / `weight_per_cluster` via `reset_sums=false` in `reduce_rows_by_key` / `reduce_cols_by_key`, removing two per-batch `raft::linalg::add` kernels
- Removed dead `update_centroids` helpers (both the `detail` and public template) — no remaining callers after the `fit_main` consolidation
- Perf: remove multiple `raft::sync_stream` calls and add a CUDA Event to record if the convergence criteria is met. Convergence check is now done on device. Average per-iteration time with mandatory inertia check now matches previous benchmarks even when previously inertia check was disabled.

## C Tests
This PR adds C tests for KMeans. These were missing. Here we test both -- the old version and the new (i.e. breaking change).

## Benchmarks:
With mandatory early stopping. Batch size is such that we fill up 90% of available GPU memory (95830MiB)
HW:
GPU:
`NVIDIA H100 NVL (CUDA 13.0)`
CPU:
```
Architecture:             x86_64
  CPU op-mode(s):         32-bit, 64-bit
  Address sizes:          52 bits physical, 57 bits virtual
  Byte Order:             Little Endian
CPU(s):                   256
  On-line CPU(s) list:    0-255
Vendor ID:                AuthenticAMD
  Model name:             AMD EPYC 9554 64-Core Processor 
```
```
================================================================================
 SUMMARY
================================================================================
  n_clusters     batch_size  fit_time(s)        inertia   n_iter
----------------------------------------------------------------
      10,000     29,120,352      1584.72     2.8677e+08       30
      20,000     29,120,352      2907.34     2.7368e+08       31
      30,000     29,101,305      4254.43     2.6617e+08       31
      40,000     29,092,704      5836.12     2.6086e+08       32
      50,000     29,083,488      7107.04     2.5680e+08       31
```

## Breaking Change
This PR is a breaking change of the C++ API because the `inertia_check` param is removed. The breaking changes to the C ABI will be applied in 26.08


c/include/cuvs/cluster/kmeans.h
```diff
@@ -91,7 +91,7 @@ struct cuvsKMeansParams {
    */
   int batch_centroids;
 
-  /** Check inertia during iterations for early convergence. */
+  /** Deprecated, ignored. Kept for ABI compatibility. */
```
tarang-jain, line 96: We probably shouldn't be modifying the wording here. And we probably want to use a different struct that breaks ABI, suffixed by the version.

cpp/src/cluster/detail/kmeans.cuh
```diff
@@ -791,70 +683,253 @@ void kmeans_fit(raft::resources const& handle,
...
 
+  IndexT centroid_buf_size = n_clusters * n_features;
+  rmm::device_uvector<DataT> centroid_buf_A(centroid_buf_size, stream);
+  rmm::device_uvector<DataT> centroid_buf_B(centroid_buf_size, stream);
+  DataT* cur_centroids_ptr = centroid_buf_A.data();
+  DataT* new_centroids_ptr = centroid_buf_B.data();
+
+  auto minClusterAndDistance = raft::make_device_vector<raft::KeyValuePair<IndexT, DataT>, IndexT>(
+    handle, streaming_batch_size);
+  auto L2NormBatch       = raft::make_device_vector<DataT, IndexT>(handle, streaming_batch_size);
```
viclafargue, line 694: `pams.streaming_batch_size = 0` by default in the data on device case, but nothing prevent a user from setting a value. This would allocate a smaller than `n_samples` `L2NormBatch` which would cause OOB writes (and later reads) during norm computation.

We should probably guard this with a check :
`RAFT_EXPECTS(streaming_batch_size == n_samples || !data_on_device, ...)`

cpp/src/cluster/detail/kmeans.cuh
```diff
@@ -697,90 +570,109 @@ void initScalableKMeansPlusPlus(raft::resources const& handle,
...
+    } else if (iter_params.init == cuvs::cluster::kmeans::params::InitMethod::KMeansPlusPlus) {
+      IndexT default_init_size =
+        data_on_device ? n_samples : std::min(static_cast<IndexT>(3 * n_clusters), n_samples);
+      IndexT init_sample_size = iter_params.init_size > 0
+                                  ? std::min(static_cast<IndexT>(iter_params.init_size), n_samples)
+                                  : default_init_size;
+
+      auto init_sample =
+        raft::make_device_matrix<DataT, IndexT>(handle, init_sample_size, n_features);
+      raft::matrix::sample_rows(handle, random_state, X, init_sample.view());
```
viclafargue, line 663:
```suggestion
if (init_sample_size == n_samples && data_on_device) {
  auto init_sample_const = raft::make_device_matrix_view<const DataT, IndexT>(X.data_handle(), n_samples, n_features);
  // pass directly to kmeansPlusPlus / initScalableKMeansPlusPlus
} else {
  auto init_sample = raft::make_device_matrix<DataT, IndexT>(handle, init_sample_size, n_features);
  raft::matrix::sample_rows(handle, random_state, X, init_sample.view());
  // pass init_sample to kmeansPlusPlus / initScalableKMeansPlusPlus
}
```

If `init_size = 0` in the data on device path, we basically double memory use by copying the dataset over. Let's skip this by creating a view on the dataset.

cpp/src/cluster/detail/kmeans_common.cuh
```diff
@@ -594,8 +604,174 @@ DataT compute_centroid_shift(raft::resources const& handle,
...
+                                              workspace,
+                                              centroid_norms);
+
+  KeyValueIndexOp<IndexT, DataT> conversion_op;
+  thrust::transform_iterator<KeyValueIndexOp<IndexT, DataT>,
+                             const raft::KeyValuePair<IndexT, DataT>*>
+    labels_itr(minClusterAndDistance.data_handle(), conversion_op);
+
+  auto batch_workspace = rmm::device_uvector<char>(
+    current_batch_sz, stream, raft::resource::get_workspace_resource(handle));
```
viclafargue, line 732: Every call to `process_batch` allocates both this workspace and the device scalar below. Both buffers could be instantiated out of the `process_batch` function.

cpp/src/cluster/detail/kmeans.cuh
```diff
@@ -697,90 +570,109 @@ void initScalableKMeansPlusPlus(raft::resources const& handle,
...
+      return;
+    }
+
+    raft::random::RngState random_state(iter_params.rng_state.seed);
+
+    if (iter_params.init == cuvs::cluster::kmeans::params::InitMethod::Random) {
+      raft::matrix::sample_rows(handle, random_state, X, centroidsRawData);
+    } else if (iter_params.init == cuvs::cluster::kmeans::params::InitMethod::KMeansPlusPlus) {
+      IndexT default_init_size =
+        data_on_device ? n_samples : std::min(static_cast<IndexT>(3 * n_clusters), n_samples);
```
viclafargue, line 656: Unlikely to be an actual issue, but `n_clusters` could be casted before the multiplication to avoid any risk of integer overflow.

cpp/src/cluster/detail/kmeans.cuh
```diff
@@ -791,70 +683,253 @@ void kmeans_fit(raft::resources const& handle,
...
+      std::swap(cur_centroids_ptr, new_centroids_ptr);
+
+      bool done = false;
+
+      DataT curClusteringCost = DataT{0};
+      raft::copy(&curClusteringCost, clustering_cost.data_handle(), 1, stream);
+      raft::resource::sync_stream(handle, stream);
+
+      if (curClusteringCost == DataT{0}) {
+        RAFT_LOG_WARN("Zero clustering cost detected: all points coincide with their centroids.");
```
viclafargue, line 881: Going from `ASSERT` to `RAFT_LOG_WARN` may indeed be useful for the spectral clustering case. However, removing the `inertia_check` option forces the sync at every iteration. Do we truly need to drop this option?

cpp/src/cluster/detail/kmeans_common.cuh
```diff
@@ -594,8 +604,174 @@ DataT compute_centroid_shift(raft::resources const& handle,
...
+    }
+  } else {
+    std::vector<DataT> h_weights(n_samples);
+    auto d_view = raft::make_device_vector_view<const DataT, IndexT>(weight_ptr, n_samples);
+    auto h_view = raft::make_host_vector_view<DataT, IndexT>(h_weights.data(), n_samples);
+    raft::copy(handle, h_view, d_view);
+    raft::resource::sync_stream(handle);
+    for (IndexT i = 0; i < n_samples; ++i) {
+      wt_sum += h_weights[i];
+    }
```
viclafargue, line 646: In the data on device case since the data is already on device it would be much faster to sumreduce thanks to `cub::DeviceReduce::Sum` or `raft::linalg::reduce`. The summation would also have better precision since it is done in a tree fashion O(log N).

python/cuvs/cuvs/cluster/kmeans/kmeans.pxd
```diff
@@ -33,9 +33,10 @@ cdef extern from "cuvs/cluster/kmeans.h" nogil:
         int batch_samples,
         int batch_centroids,
         bool inertia_check,
```
viclafargue, line 35: Maybe add a comment saying the field is present but deprecated.

cpp/src/cluster/detail/kmeans.cuh
```diff
@@ -791,70 +683,253 @@ void kmeans_fit(raft::resources const& handle,
...
+    DataT priorClusteringCost = 0;
+
+    for (n_current_iter = 1; n_current_iter <= iter_params.max_iter; ++n_current_iter) {
+      RAFT_LOG_DEBUG("KMeans.fit: Iteration-%d", n_current_iter);
+
+      raft::matrix::fill(handle, centroid_sums.view(), DataT{0});
+      raft::matrix::fill(handle, weight_per_cluster.view(), DataT{0});
+      raft::linalg::map(handle,
+                        raft::make_device_scalar_view(clustering_cost.data_handle()),
+                        raft::const_op<DataT>{DataT{0}});
```
viclafargue, line 789: Shouldn't we use `raft::matrix::fill` for a device scalar too?

viclafargue (commented):
Thanks! Here are some comments.

cpp/src/cluster/detail/kmeans.cuh
cjnolet: I'm not sure we need the log and an assert might be better?

Early stopping (aka skipping iterations) is ultimately going to be the best way to extract perf here. Whether it's by explicitly computing inertia or just looking at the residuals of the centroids from the prior iteration. 

Seems like inertia check / residuals could be done on gpu if we had to in order to avoid syncing so we would only need to sync in the final iteration, right?

tarang-jain: >Seems like inertia check / residuals could be done on gpu if we had to in order to avoid syncing so we would only need to sync in the final iteration, right?

Until the iteration has completed, the CPU should not start the next iteration. So all the operations on the GPU stream must complete to finish the iteration.

tarang-jain: >Seems like inertia check / residuals could be done on gpu if we had to in order to avoid syncing so we would only need to sync in the final iteration, right?

Yes, but this is throwing an error in the spectral clustering case wherein all the points converge on the centroids themselves. This is happening in one of the spectral tests and an assertion here is leading to an error, where instead it should simply return those centroids directly.

tarang-jain: >I'm not sure we need the log and an assert might be better?

Therefore, I had to change it to a warning instead of an assertion. Earlier those spectral tests were skipping the inertia check which was avoiding the assertion.

cpp/src/cluster/detail/kmeans_common.cuh
tarang-jain: When its device accessible, I have changed that to a `raft::linalg::mapThenSumReduce`. I have also removed this function and directly updated `checkWeight` (changed its name to `weightSum`). We do the scaling after the weight sum is computed.

cpp/src/cluster/detail/kmeans.cuh
tarang-jain: I have gotten rid of the batching for the device path. So when a user sets a batch size for device mdspan, we just set it to n_samples and warn the user. We should definitely not be creating a new buffer just for the init sample if we can accommodate the entire input matrix on device already.

python/cuvs/cuvs/cluster/kmeans/kmeans.pxd
tarang-jain: I dont think its necessary to add a comment here in the .pxd. The C header already has that information. And this file will be updated along with the C headers / src files.

cpp/src/cluster/detail/kmeans.cuh
tarang-jain: Done!

cpp/src/cluster/detail/kmeans_common.cuh
tarang-jain: moved the workspace buffer allocation outside the process_batch function.

cpp/src/cluster/detail/kmeans.cuh
tarang-jain: I completely skipped the sampling for the device path. That is how it was being done earlier. The init size is only used if the data is on host.

cpp/src/cluster/detail/kmeans.cuh
tarang-jain: I have updated this so that for device arrays, we simply ignore the streaming_batch_size and use the entire dataset always.

cpp/src/cluster/kmeans_impl.cuh
```diff
@@ -18,8 +18,12 @@ void fit_main(raft::resources const& handle,
               raft::host_scalar_view<IndexT> n_iter,
               rmm::device_uvector<char>& workspace)
```
viclafargue, line 19: The workspace argument should either be passed along or dropped.

cpp/src/cluster/detail/kmeans.cuh
```diff
@@ -791,70 +700,257 @@ void kmeans_fit(raft::resources const& handle,
...
+        raft::make_device_matrix_view<DataT, IndexT>(new_centroids_ptr, n_clusters, n_features);
+
+      std::optional<raft::device_vector_view<const DataT, IndexT>> centroid_norms_opt =
+        std::nullopt;
+      if (need_compute_norms) {
+        raft::linalg::norm<raft::linalg::L2Norm, raft::Apply::ALONG_ROWS>(
+          handle, centroids_const, centroid_norms_buf.view());
+        centroid_norms_opt = raft::make_device_vector_view<const DataT, IndexT>(
+          centroid_norms_buf.data_handle(), n_clusters);
+      }
```
viclafargue: Shouldn't the cosine metric use the `raft::sqrt_op` operator to compute the norm? I believe that `KMeans` does not handle the cosine metric at all anyway at the moment neither in `fusedDistanceNNMinReduce` nor in `pairwise_distance_kmeans`. We could maybe fix this in a follow-up PR.

cpp/src/cluster/detail/kmeans.cuh
```diff
@@ -697,90 +570,126 @@ void initScalableKMeansPlusPlus(raft::resources const& handle,
...
+        else
+          initScalableKMeansPlusPlus<DataT, IndexT>(
+            handle, iter_params, init_data, centroidsRawData, workspace);
+      };
+
+      if constexpr (data_on_device) {
+        run_kmeanspp(X);
+      } else {
+        IndexT default_init_size =
+          std::min(static_cast<IndexT>(std::int64_t{3} * n_clusters), n_samples);
```
viclafargue: We went from `3 * streaming_batch_size` to `3 * n_clusters`. This may lead to some quality regression. Is this intentional?

cpp/src/cluster/detail/kmeans_common.cuh
```diff
@@ -129,43 +133,43 @@ void countLabels(raft::resources const& handle,
...
       "[Warning!] KMeans: normalizing the user provided sample weight to "
-      "sum up to %d samples",
-      n_samples);
-
-    auto scale = static_cast<DataT>(n_samples) / wt_sum;
-    raft::linalg::map(
-      handle, weight, raft::mul_const_op<DataT>{scale}, raft::make_const_mdspan(weight));
+      "sum up to %zu samples",
+      static_cast<size_t>(n_samples));
   }
```
viclafargue: Please move the log where the division actually happens.

cpp/src/cluster/detail/kmeans.cuh
```diff
@@ -926,8 +1033,16 @@ void kmeans_predict(raft::resources const& handle,
   else
     raft::matrix::fill(handle, weight.view(), DataT(1));
 
-  // check if weights sum up to n_samples
-  if (normalize_weight) checkWeight(handle, weight.view(), workspace);
+  if (normalize_weight) {
+    DataT wt_sum = weightSum(handle, raft::make_const_mdspan(weight.view()));
+    if (wt_sum != static_cast<DataT>(n_samples)) {
```
viclafargue:
```suggestion
    const DataT rel_tol = n_samples * std::numeric_limits<DataT>::epsilon();
    if (std::abs(wt_sum - n_samples) <= rel_tol) {
```

We may have to account for the fact that the summation may be imprecise.

viclafargue (approved):
Thanks! LGTM. Just a few last things I noticed.

Giving approval, but noting that the changes would likely break the ABI.
The C++ field `inertia_check` in `cuvs::cluster::kmeans::params` disappeared leading to possible compilation error of code that worked earlier.
The C field `init_size` is appended to the `cuvsKMeansParams` struct changing allocation size (`sizeof(cuvsKMeansParams)`) and causing reading past the struct allocation if user binaries are not re-compiled with the new cuVS library.

cpp/src/cluster/detail/kmeans.cuh
tarang-jain: yes because 3 * streaming_batch_size would require 3 full batches to fit on the GPU. That can cause OOM.

cpp/src/cluster/detail/kmeans.cuh
tarang-jain: > I believe that KMeans does not handle the cosine metric at all anyway

That is right. Only L2 is supported. Cosine is supported in kmeans_balanced.

cpp/src/cluster/kmeans_impl.cuh
tarang-jain: Done! Now we use this workspace. So in this PR I got rid of the main loop function (`kmeans_fit_main`). Instead we directly use kmeans_fit.

cpp/src/cluster/detail/kmeans_common.cuh
tarang-jain: Done!

cpp/src/cluster/detail/kmeans.cuh
tarang-jain: So ideally we want the user to provide this. I think 3 * n_clusters would be a good heuristic as a bare minimum. That is, ~3 points per cluster for the initialization. IMO this might not be ideal for the quality of the initialization, but we want to choose a default that does not blow up GPU memory consumption. Another reasonable default to consider would be a sample of `streaming_batch_size` itself. So the GPU memory consumption is somewhat consistent with the actual clustering.

tarang-jain: Also, large scale kmeansplusplus (out of core) with this init sample is not evaluated in terms of how much does kmeansplusplus actually help the clustering if it is done out of core.

cpp/src/cluster/detail/kmeans.cuh
tarang-jain: So we dont do sqrt_op at all here (at least not until cosine is supported in regular kmeans)

cpp/src/cluster/kmeans_impl.cuh
tarang-jain: I have put the workspace reference in an optional for `kmeans_fit`. I think @viclafargue is any way refactoring the whole kmeans_mg file in a separate PR.

tarang-jain: If after #2017, kmeans_mg does not even end up using kmeans_impl.cuh, I am leaning towards completely getting rid of that file in the future. It just acts as a trivial wrapper in that case, and instead I suppose the instantiation files (kmeans_fit_float.cu etc) can directly use the detail headers.

tarang-jain:
@viclafargue ABI wont break. We are allowed to add new members to structs and enums. However I have added the `breaking` label to this PR because downstream users of the C++ API will be affected directly.

cpp/src/cluster/detail/kmeans.cuh
```diff
@@ -697,90 +528,132 @@ void initScalableKMeansPlusPlus(raft::resources const& handle,
...
-  // Display a message if batch_centroids is set and a fusedL2NN-compatible metric is used
-  if (pams.batch_centroids != 0 && pams.batch_centroids != pams.n_clusters &&
-      (pams.metric == cuvs::distance::DistanceType::L2Expanded ||
-       pams.metric == cuvs::distance::DistanceType::L2SqrtExpanded)) {
+
+  const DataT* weight_ptr =
+    sample_weight.has_value() ? sample_weight.value().data_handle() : nullptr;
+  DataT wt_sum = sample_weight.has_value() ? weightSum(handle, sample_weight.value())
+                                           : static_cast<DataT>(n_samples);
+  if (sample_weight.has_value() && wt_sum != static_cast<DataT>(n_samples)) {
```
dantegd: fit uses exact equality to decide whether to renormalize sample weights, but predict uses a relative tolerance in predict:

```c++
const DataT rel_tol = n_samples * std::numeric_limits<DataT>::epsilon();
if (std::abs(wt_sum - n_samples) > rel_tol) { /* normalize */ }
```

 With identical sample_weight, one call can renormalize while the other doesn't, producing inconsistent scaling between training and inference, we should be consistent about it.

cpp/src/cluster/detail/kmeans.cuh
```diff
@@ -791,70 +664,253 @@ void kmeans_fit(raft::resources const& handle,
...
+      DataT curClusteringCost = DataT{0};
+      raft::copy(&curClusteringCost, clustering_cost.data_handle(), 1, stream);
+      raft::resource::sync_stream(handle, stream);
+
+      if (curClusteringCost == DataT{0}) {
+        RAFT_LOG_WARN("Zero clustering cost detected: all points coincide with their centroids.");
+      } else if (n_current_iter > 1) {
+        DataT delta = curClusteringCost / priorClusteringCost;
+        if (delta > 1 - iter_params.tol) done = true;
+      }
```
dantegd: Seems that this block was gated by params.inertia_check which defaulted oto `false`, now it's run unconditionally which will force the device to host copy on every iteration, is this change intentional? I think we could avoid this copy, or make it conditional, or at least explicitly mention this change if it's intended.

cpp/src/cluster/detail/kmeans_mg.cuh
```diff
@@ -701,49 +697,45 @@ void fit(const raft::resources& handle,
...
+               raft::make_device_scalar_view(&(clusterCostD.data()->value)));
+
+    ASSERT(comm.sync_stream(stream) == raft::comms::status_t::SUCCESS,
+           "An error occurred in the distributed operation. This can result "
+           "from a failed rank");
+    if (curClusteringCost == (DataT)0.0) {
+      RAFT_LOG_WARN("Zero clustering cost detected: all points coincide with their centroids.");
+    } else if (n_iter[0] > 1) {
+      DataT delta = curClusteringCost / priorClusteringCost;
+      if (delta > 1 - params.tol) done = true;
```
dantegd, line 740: We have the same change of behavior here with this now running unconditionally, so we need to do the same thing we do in the single gpu change I mentioned before.

cpp/src/cluster/detail/kmeans.cuh
dantegd: The reasoning for the change makes sense to me, but we should add a log/warning to alert users of the behavior change that otherwise will go silent for users that don't set init_size, and add it do the param doc in kmeans.hpp

cpp/src/cluster/detail/kmeans.cuh
```diff
@@ -926,8 +993,20 @@ void kmeans_predict(raft::resources const& handle,
...
+    const DataT rel_tol = n_samples * std::numeric_limits<DataT>::epsilon();
+    if (std::abs(wt_sum - n_samples) > rel_tol) {
+      RAFT_LOG_DEBUG(
+        "[Warning!] KMeans: normalizing the user provided sample weight to sum up to %zu samples",
+        static_cast<size_t>(n_samples));
+      raft::linalg::map(handle,
+                        weight.view(),
+                        raft::compose_op(raft::mul_const_op<DataT>{static_cast<DataT>(n_samples)},
+                                         raft::div_const_op<DataT>{wt_sum}),
+                        raft::make_const_mdspan(weight.view()));
```
dantegd: If the user passes an all-zero sample_weight, wt_sum == 0 and div_const_op<DataT>{wt_sum} produces inf/NaN in every weight. It seems to me that the old checkWeight had the same problem, but it would be a good idea to add an explicit `RAFT_EXPECTS(wt_sum > 0, ...)`

cpp/src/cluster/detail/kmeans.cuh
```diff
@@ -791,70 +664,253 @@ void kmeans_fit(raft::resources const& handle,
...
+      raft::copy(batch_weights_buf.data_handle(), wt_batch.data(), cur_batch_size, stream);
+      if (wt_sum != static_cast<DataT>(n_samples)) {
+        auto bw = raft::make_device_vector_view<DataT, IndexT>(batch_weights_buf.data_handle(),
+                                                               cur_batch_size);
+        raft::linalg::map(handle,
+                          bw,
+                          raft::compose_op(raft::mul_const_op<DataT>{static_cast<DataT>(n_samples)},
+                                           raft::div_const_op<DataT>{wt_sum}),
+                          raft::make_const_mdspan(bw));
+      }
```
dantegd: Same wt_sum == 0 issue as in kmeans_predict: if the user passes an all-zero sample_weight, wt_sum is 0 and div_const_op<DataT>{wt_sum} produces inf/NaN for every batch weight, which then affects centroid_sums / weight_per_cluster and ultimately the fitted centroids. Worse, because wt_sum is computed once outside this lambda (at the top of kmeans_fit), the bad value is baked into every batch of every iteration of every n_init — there's no recovery path. The explicit guard would be good here like:

```C++
DataT wt_sum = sample_weight.has_value() ? weightSum(handle, sample_weight.value())
                                         : static_cast<DataT>(n_samples);
RAFT_EXPECTS(wt_sum > DataT{0},
             "KMeans: sample_weight must have a strictly positive sum");
```

We could also fall back to uniform weights with a warning I guess, not sure which one is better, I'd probably go with the guard.

c/include/cuvs/cluster/kmeans.h
```diff
@@ -108,7 +110,92 @@ struct cuvsKMeansParams {
...
    */
-   int64_t streaming_batch_size;
+  int64_t streaming_batch_size;
+
+  /**
+   * Number of samples to draw for KMeansPlusPlus initialization.
+   * When set to 0, uses heuristic min(3 * n_clusters, n_samples) for host data,
+   * or n_samples for device data.
+   */
+  int64_t init_size;
```
dantegd, line 120: Question: adding `init_size` at the tail changes `sizeof(cuvsKMeansParams` and therefore still breaks ABI for any caller that stack-allocates or sizeofs the struct, no?

 In practice cVS consumers go through `cuvsKMeansParamsCreate`, so this is a soft break and might be not much of an issue, but wanted to raise it to discuss.

cpp/src/cluster/detail/kmeans.cuh
```diff
@@ -791,70 +664,253 @@ void kmeans_fit(raft::resources const& handle,
...
+        DataT batch_cost = DataT{0};
+        cuvs::cluster::kmeans::cluster_cost(handle,
+                                            batch_data_view,
+                                            centroids_const,
+                                            raft::make_host_scalar_view(&batch_cost),
+                                            batch_sw);
+
+        iter_inertia += batch_cost;
+      }
+    }
```
dantegd, line 967: Now that clustering_cost is accumulated on device every iteration inside the Lloyd loop, the final accepted iteration's clustering_cost is exactly the inertia to report, no? 

This extra post-loop batched pass (one full read of X per n_init) duplicates work that was just done if I'm not mistaken. If so, consider copying the last clustering_cost scalar to iter_inertia

cpp/src/cluster/detail/kmeans.cuh
```diff
@@ -697,90 +528,132 @@ void initScalableKMeansPlusPlus(raft::resources const& handle,
...
+      } else {
+        IndexT default_init_size =
+          std::min(static_cast<IndexT>(std::int64_t{3} * n_clusters), n_samples);
+        IndexT init_sample_size =
+          iter_params.init_size > 0
+            ? std::min(static_cast<IndexT>(iter_params.init_size), n_samples)
+            : default_init_size;
+
+        auto init_sample =
+          raft::make_device_matrix<DataT, IndexT>(handle, init_sample_size, n_features);
```
dantegd: Small thing: This device matrix is allocated every call to init_centroids, once per seed_iter. Its shape is constant across the n_init loop, so we could do the allocation out of the lambda and reuse it

cpp/src/cluster/detail/kmeans_common.cuh
```diff
@@ -502,7 +505,8 @@ void compute_centroid_adjustments(
   IndexT n_clusters,
   raft::device_matrix_view<DataT, IndexT, raft::row_major> centroid_sums,
   raft::device_vector_view<DataT, IndexT> weight_per_cluster,
-  rmm::device_uvector<char>& workspace)
+  rmm::device_uvector<char>& workspace,
+  bool reset_sums = true)
```
dantegd, line 501: Docstring needs update

cpp/tests/cluster/kmeans.cu
```diff
@@ -433,9 +433,8 @@ class KmeansFitBatchedTest : public ::testing::TestWithParam<KmeansBatchedInputs
     auto d_centroids_ref_view =
       raft::make_device_matrix_view<T, int>(d_centroids_ref.data(), params.n_clusters, n_features);
 
-    params.init          = cuvs::cluster::kmeans::params::Array;
-    params.inertia_check = true;
-    params.max_iter      = 20;
+    params.init     = cuvs::cluster::kmeans::params::Array;
+    params.max_iter = 20;
```
dantegd, line 441: A few  new tests could be useful to consider:

- a test that runs fit with host input and device input on the same data/seed and asserts bit-identical centroids/inertia
- a test that exercises init_size (including 0 default and an explicit value)
- a test for the zero-cost warning path (e.g., n_samples == n_clusters with distinct points), to lock down the new behavior

python/cuvs/cuvs/cluster/kmeans/kmeans.pyx
```diff
@@ -134,8 +136,8 @@ cdef class KMeansParams:
             self.params.batch_samples = batch_samples
         if batch_centroids is not None:
             self.params.batch_centroids = batch_centroids
-        if inertia_check is not None:
-            self.params.inertia_check = inertia_check
+        if init_size is not None:
```
dantegd, line 151: Existing Python code that passes inertia_check=...  will now raise TypeError. We should consider accepting and doing a deprecation warning for one release, then removing. Same applies to the docstring update

dantegd (changes requested):
This PR improves things nicely, the code get significant improvement, love to see this!

cpp/src/cluster/detail/kmeans.cuh
tarang-jain: The final inertia was being computed even before this PR with a `cuvs::cluster::kmeans::cluster_cost` call. The thing is that the centroids are modified in the end after the inertia computation. So one extra pass is needed with the final centroids.

cpp/src/cluster/detail/kmeans.cuh
tarang-jain: This was intended. According to my discussion with @cjnolet we should always be early stopping. Yes this does mean that we now will always have to wait for one iteration to complete before the next one because the CPU should be made aware of whether or not it should proceed with the next iteration. Keeping this thread open to discussion if others have some inputs.

cpp/src/cluster/detail/kmeans_mg.cuh
tarang-jain: @viclafargue will be taking care of the iteration loop in kmeans_mg in #2017

c/include/cuvs/cluster/kmeans.h
tarang-jain: yes I was just following the instructions in the doc that we have as a guide for ABI stability. And we also have a CI check for the same, which will fail if there is a breaking change. I also added `streaming_batch_size` as a new param in the last release, which would have also changed the `sizeof(cuvsKMeansParams)`

cpp/src/cluster/detail/kmeans_common.cuh
tarang-jain: Done!

cpp/src/cluster/detail/kmeans.cuh
tarang-jain: Done! I have added the same check to fit and predict

python/cuvs/cuvs/cluster/kmeans/kmeans.pyx
tarang-jain: Because the C API still has it, this wont raise an error, but I added a deprecation warning.

cpp/src/cluster/detail/kmeans.cuh
tarang-jain: Done! We preallocate now.

cpp/src/cluster/detail/kmeans.cuh
tarang-jain: I added the check in weightSum itself. So that function always returns a sum that is greater than zero.

cpp/src/cluster/detail/kmeans.cuh
```diff
@@ -791,70 +681,253 @@ void kmeans_fit(raft::resources const& handle,
...
+  cuvs::spatial::knn::detail::utils::batch_load_iterator<DataT> weight_batches(
+    weight_ptr, n_samples, 1, streaming_batch_size, stream);
+
+  if (weight_ptr == nullptr) { raft::matrix::fill(handle, batch_weights_buf.view(), DataT{1}); }
+
+  auto prepare_batch_weights = [&](const auto& wt_batch, IndexT cur_batch_size) {
+    if (weight_ptr != nullptr) {
+      raft::copy(batch_weights_buf.data_handle(), wt_batch.data(), cur_batch_size, stream);
+      if (needs_wt_rescale) {
+        auto bw = raft::make_device_vector_view<DataT, IndexT>(batch_weights_buf.data_handle(),
```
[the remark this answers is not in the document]
tarang-jain: Took this coderabbit suggestion

cpp/tests/cluster/kmeans.cu
tarang-jain: The first one (bit-identical centroids and inertia with streaming host data and data on device) was already being tested.
I added tests for init_size and zero-cost paths as well as n_init > 0.

cpp/src/cluster/detail/kmeans.cuh
```diff
@@ -791,70 +683,277 @@ void kmeans_fit(raft::resources const& handle,
...
+    auto d_prior_cost = raft::make_device_scalar<DataT>(handle, DataT{0});
+    auto d_done_flag  = raft::make_device_scalar<int>(handle, 0);
+    auto h_done_flag  = raft::make_pinned_scalar<int>(handle, 0);
+
+    cudaEvent_t convergence_event;
+    RAFT_CUDA_TRY(cudaEventCreateWithFlags(&convergence_event, cudaEventDisableTiming));
+
+    for (n_current_iter = 1; n_current_iter <= iter_params.max_iter; ++n_current_iter) {
+      if (n_current_iter > 1) {
+        RAFT_CUDA_TRY(cudaEventSynchronize(convergence_event));
```
tarang-jain: The convergence event is recorded right after the inertia is computed, but we check it here before proceeding to the next iteration so that the operations after the convergence check are not blocked. The use of an event also means that only the operations up until the event need to be completed in order for it to be synchronized.

dantegd (approved):

c/include/cuvs/cluster/kmeans.h
```diff
@@ -108,7 +110,92 @@ struct cuvsKMeansParams {
...
+};
+
+/**
+ * @brief Hyper-parameters for the kmeans algorithm
+ */
+ struct cuvsKMeansParams_v1 {
```
cjnolet: This isn't how we version- we use the current minor version from calver (that would be 06). That makes it completely obvious exactly WHEN we made the change.

tarang-jain: So that is `cuvsKMeansParams_v08`?

cjnolet: Yes or just `cuvsKMeansParams_v6`. But also make sure you add comments appropriately (and add fixme or todo anywhere that uses it so that it's easier to spot when we go to change it in 26.08).

tarang-jain: Done!

cpp/include/cuvs/cluster/kmeans.hpp
```diff
@@ -113,9 +113,14 @@ struct params : base_params {
...
-   * If true, check inertia during iterations for early convergence.
+   * Number of samples to randomly draw for the KMeansPlusPlus initialization
+   * step. A random subset of this size is used for centroid seeding.
+   * When set to 0 the default depends on the data location:
+   *   - Device data: n_samples (use the full dataset).
+   *   - Host data:   min(3 * n_clusters, n_samples).
+   * Default: 0.
    */
-  bool inertia_check = false;
+  int64_t init_size = 0;
```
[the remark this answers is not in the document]
dantegd: I soflty agree with this rabbit comment, would be worth to add a not in the PR description and code, C++ callers using the C++ API directly will fail to compile, not warn. That's a real source-API break worth a release-notes line at least

dantegd (approved):
Just had a minor suggestion, not blocking

tarang-jain:
@dantegd I have updated the PR desc. Since this PR is marked as `breaking`, it will automatically be mentioned in CHANGELOG.md by the bot, right?

cpp/src/neighbors/ivf_pq/ivf_pq_build.cuh
```diff
@@ -1329,6 +1329,9 @@ auto build(raft::resources const& handle,
     rmm::device_uvector<uint32_t> labels(n_rows_train, stream, big_memory_resource);
     auto centers_const_view = raft::make_device_matrix_view<const float, internal_extents_t>(
       cluster_centers, impl->n_lists(), impl->dim());
+    if (impl->metric() == distance::DistanceType::CosineExpanded) {
+      raft::linalg::row_normalize<raft::linalg::L2Norm>(handle, centers_const_view, centers_view);
+    }
```
tarang-jain, line 1334: This normalization was lost in #2001. So this PR adds it again.

c/tests/cluster/run_kmeans_c.c
```diff
@@ -0,0 +1,200 @@
...
+ #include <stdbool.h>
+ #include <stddef.h>
+ #include <stdint.h>
+
+ static void fill_matrix_tensor(DLManagedTensor* t,
+                                void* data,
+                                int64_t* shape,
+                                DLDeviceType device_type,
+                                uint8_t code,
+                                uint8_t bits)
```
lowener: The common usage for C tests seems to be `#include "../../src/core/interop.hpp"` and use `cuvs::core::to_dlpack` to simplify the interaction with DLManagedTensor. Can it get used here too, instead of `fill_matrix_tensor` and `fill_vector_tensor`

c/tests/cluster/run_kmeans_c.c
lowener, line 1: Merge run_kmeans_c and kmeans_c into the same file.
The total line number won't be so big once the deprecated path are removed.

cpp/src/cluster/detail/kmeans.cuh
```diff
@@ -739,54 +572,103 @@ void kmeans_fit(raft::resources const& handle,
...
 
-  // check if weights sum up to n_samples
-  checkWeight<DataT>(handle, weight.view(), workspace);
+  auto init_centroids = [&](const cuvs::cluster::kmeans::params& iter_params,
+                            raft::device_matrix_view<DataT, IndexT> centroidsRawData) {
+    if (iter_params.init == cuvs::cluster::kmeans::params::InitMethod::Array) {
+      raft::copy(
+        handle,
+        raft::make_device_vector_view(centroidsRawData.data_handle(), n_clusters * n_features),
+        raft::make_device_vector_view(centroids.data_handle(), n_clusters * n_features));
```
lowener:
```suggestion
        centroidsRawData.view(),
        centroids.view());
```

cpp/src/cluster/detail/kmeans.cuh
```diff
@@ -797,70 +679,277 @@ void kmeans_fit(raft::resources const& handle,
     n_init = 1;
   }
 
+  IndexT centroid_buf_size = n_clusters * n_features;
+  rmm::device_uvector<DataT> centroid_buf_A(centroid_buf_size, stream);
+  rmm::device_uvector<DataT> centroid_buf_B(centroid_buf_size, stream);
+  DataT* cur_centroids_ptr = centroid_buf_A.data();
+  DataT* new_centroids_ptr = centroid_buf_B.data();
```
lowener: Keep the cur / new naming instead of A/B

cpp/src/cluster/detail/kmeans.cuh
```diff
@@ -797,70 +679,277 @@ void kmeans_fit(raft::resources const& handle,
...
+  auto L2NormBatch       = raft::make_device_vector<DataT, IndexT>(handle, streaming_batch_size);
+  auto batch_weights_buf = raft::make_device_vector<DataT, IndexT>(handle, streaming_batch_size);
+  rmm::device_uvector<DataT> L2NormBuf_OR_DistBuf(0, stream);
+
+  auto centroid_sums      = raft::make_device_matrix<DataT, IndexT>(handle, n_clusters, n_features);
+  auto weight_per_cluster = raft::make_device_vector<DataT, IndexT>(handle, n_clusters);
+  auto centroid_norms_buf = raft::make_device_vector<DataT, IndexT>(handle, n_clusters);
+  auto clustering_cost    = raft::make_device_scalar<DataT>(handle, DataT{0});
+
+  rmm::device_uvector<char> batch_workspace(streaming_batch_size, stream);
```
lowener, line 695: Can `ws` be reused? (given workspace or local workspace)

cpp/src/cluster/detail/minClusterDistanceCompute.cu
```diff
@@ -24,7 +26,8 @@ void minClusterAndDistanceCompute(
   cuvs::distance::DistanceType metric,
   int batch_samples,
   int batch_centroids,
-  rmm::device_uvector<char>& workspace)
+  rmm::device_uvector<char>& workspace,
+  std::optional<raft::device_vector_view<const DataT, IndexT>> precomputed_centroid_norms)
```
lowener: Use the precomputed norms in the function

dantegd (approved):
cmake review

cpp/src/cluster/detail/kmeans.cuh
```diff
@@ -797,70 +679,277 @@ void kmeans_fit(raft::resources const& handle,
...
+  RAFT_LOG_DEBUG(
+    "KMeans.fit: n_samples=%zu, n_features=%zu, n_clusters=%d, streaming_batch_size=%zu",
+    static_cast<size_t>(n_samples),
+    static_cast<size_t>(n_features),
+    n_clusters,
+    static_cast<size_t>(streaming_batch_size));
+
+  bool need_compute_norms = metric == cuvs::distance::DistanceType::L2Expanded ||
+                            metric == cuvs::distance::DistanceType::L2SqrtExpanded;
+  bool use_norm_cache = need_compute_norms && !data_on_device;
```
lowener: On the device path the norms will be recomputed max_iter * n_init. It might be faster as well to cache the norm there (on device preferably since the dataset already fit there).

cpp/src/cluster/detail/kmeans.cuh
```diff
@@ -797,70 +679,277 @@ void kmeans_fit(raft::resources const& handle,
...
+  IndexT centroid_buf_size = n_clusters * n_features;
+  rmm::device_uvector<DataT> centroid_buf_A(centroid_buf_size, stream);
+  rmm::device_uvector<DataT> centroid_buf_B(centroid_buf_size, stream);
+  DataT* cur_centroids_ptr = centroid_buf_A.data();
+  DataT* new_centroids_ptr = centroid_buf_B.data();
+
+  auto minClusterAndDistance = raft::make_device_vector<raft::KeyValuePair<IndexT, DataT>, IndexT>(
+    handle, streaming_batch_size);
+  auto L2NormBatch       = raft::make_device_vector<DataT, IndexT>(handle, streaming_batch_size);
+  auto batch_weights_buf = raft::make_device_vector<DataT, IndexT>(handle, streaming_batch_size);
```
lowener, line 688: Currently the weights are recomputed every iteration.
For device path this vector can contain pre-normalized weights, and weight-norm can be skipped

cpp/src/cluster/detail/kmeans.cuh
```diff
@@ -797,70 +679,277 @@ void kmeans_fit(raft::resources const& handle,
...
+
+  cuvs::spatial::knn::detail::utils::batch_load_iterator<DataT> data_batches(
+    X.data_handle(), n_samples, n_features, streaming_batch_size, stream);
+  // Only materialize weight batches when weights are provided; otherwise we
+  // never touch this iterator (and never dereference a null-pointer batch).
+  std::optional<cuvs::spatial::knn::detail::utils::batch_load_iterator<DataT>> weight_batches;
+  if (weight_ptr != nullptr) {
+    weight_batches.emplace(weight_ptr, n_samples, 1, streaming_batch_size, stream);
+  } else {
+    raft::matrix::fill(handle, batch_weights_buf.view(), DataT{1});
```
lowener, line 709: Fill with `1/n_samples` and skip weight normalization

cpp/src/cluster/detail/kmeans.cuh
lowener: I don't think that there would be any differences between the use of cudaEventSynchronize and a sync_stream at the end of the loop? The only operation that happens is the loop increment?

cpp/src/cluster/detail/kmeans.cuh
```diff
@@ -797,70 +679,277 @@ void kmeans_fit(raft::resources const& handle,
...
+
+        std::optional<raft::device_vector_view<const DataT, IndexT>> batch_sw = std::nullopt;
+        if (weight_ptr != nullptr) { batch_sw = prepare_batch_weights(wt_data, cur_batch_size); }
+
+        DataT batch_cost = DataT{0};
+        cuvs::cluster::kmeans::cluster_cost(handle,
+                                            batch_data_view,
+                                            centroids_const,
+                                            raft::make_host_scalar_view(&batch_cost),
+                                            batch_sw);
```
lowener, line 963: The cluster_cost could be modified to accept an optional `X_norm` to avoid recomputing it. That function also synchronize each time it is called, while it would be better to accumulate cluster_cost on device and transfer+sync once at the end.
Maybe it's not in the scope of that PR but wanted to point it out.

lowener (changes requested):

cpp/src/cluster/detail/kmeans.cuh
tarang-jain: there is no separate weight normalization step. The weights are directly used in computing the cost as is.

cpp/src/cluster/detail/kmeans.cuh
```diff
@@ -829,24 +862,27 @@ void kmeans_fit(
...
+                         L2NormBatch.data_handle(),
+                         cur_batch_size,
+                         stream);
+            }
+          } else if (use_norm_cache) {
+            raft::copy(L2NormBatch.data_handle(),
+                       h_norm_cache.data_handle() + data_batch.offset(),
                        cur_batch_size,
                        stream);
           }
```
lowener: Currently this branch will never be reached: when `use_norm_cache` is true, `need_compute_norms` is true too.
Bring back the `norms_cached` boolean, with it's flip after the first pass on all the data and switch to 
```cpp
if (need_compute_norms && !norms_cached)
```

tarang-jain: I brought back norms_cached and got rid of use_norms_cache (we are any way checking if the data is on device)

cpp/src/cluster/detail/kmeans.cuh
tarang-jain: tracking here https://github.com/rapidsai/cuvs/issues/2057

lowener (approved):
LGTM

PR #2016, opened 2026-04-13
Fix argmin/argmax based on the distance type

achirkin:
Fix the distance type not taken into account when generating ground truth on GPU.


tfeher (approved):
LGTM.

cjnolet (approved):

PR #2017, opened 2026-04-13
Multi-GPU Batched KMeans

### LINKED ISSUE #1989: [FEA] Multi-node Multi-GPU Kmeans (C++) to support new out-of-core batching
We recently added out-of-core batching to the single-gpu K-means, which allows it to accept a host matrix for training, along with a batch size so that it can break up the dataset to fit batches in device memory. 

We should extend the multi-node mutli-gpu C++ NCCL implementation to allow each rank to specify a host matrix. 

Since most distributed systems will already have partitioned their datasets, we should consider accepting the data as partitioned and setting the batch size automatically (note that the difference here is that each batch *could* potentially be a different size, so we might need a small update to the single-gpu version to also support an array of batch sizes).

viclafargue:
Closes #1989.

Adds multi-GPU support to KMeans fit for host-resident data, with two modes:
- **OpenMP (cuVS SNMG)**: A single process drives all local GPUs via OMP threads and raw NCCL. Activated automatically when the handle is a `device_resources_snmg`.
- **RAFT comms (Ray / Dask / MPI)**: Each rank is a separate process that calls fit with its own data shard and an initialized RAFT communicator. Coordination uses the RAFT comms.

Both modes share the same core Lloyd's loop, batched streaming of host data, NCCL/comms allreduce of centroid sums and counts, and synchronized convergence. Supports sample weights, n_init best-of-N restarts, KMeansPlusPlus initialization, and float/double. Falls back to single-GPU when neither multi-GPU resources nor comms are present.


viclafargue:
Here are some instructions to test the Multi-GPU Batched KMeans API with RAFT comms (to be used with Ray/Dask) :

<details>
<summary>RAFT comms (Ray/Dask) demo code</summary>

```
#include <cuvs/cluster/kmeans.hpp>

#include <raft/comms/std_comms.hpp>
#include <raft/core/device_mdarray.hpp>
#include <raft/core/host_mdspan.hpp>
#include <raft/core/resource/comms.hpp>
#include <raft/core/resource/cuda_stream.hpp>

#include <cuda_runtime.h>
#include <mpi.h>
#include <nccl.h>

#include <algorithm>
#include <cmath>
#include <cstdio>
#include <cstdlib>
#include <cstring>
#include <numeric>
#include <random>
#include <vector>

#define CHECK_CUDA(call)                                                 \
  do {                                                                   \
    cudaError_t e = (call);                                              \
    if (e != cudaSuccess) {                                              \
      std::fprintf(stderr, "CUDA error %s @ %s:%d\n",                   \
                   cudaGetErrorString(e), __FILE__, __LINE__);           \
      MPI_Abort(MPI_COMM_WORLD, 1);                                      \
    }                                                                    \
  } while (0)

#define CHECK_NCCL(call)                                                 \
  do {                                                                   \
    ncclResult_t r = (call);                                             \
    if (r != ncclSuccess) {                                              \
      std::fprintf(stderr, "NCCL error %s @ %s:%d\n",                   \
                   ncclGetErrorString(r), __FILE__, __LINE__);           \
      MPI_Abort(MPI_COMM_WORLD, 1);                                      \
    }                                                                    \
  } while (0)

int main(int argc, char** argv)
{
  MPI_Init(&argc, &argv);

  int rank, num_ranks;
  MPI_Comm_rank(MPI_COMM_WORLD, &rank);
  MPI_Comm_size(MPI_COMM_WORLD, &num_ranks);

  CHECK_CUDA(cudaSetDevice(rank));

  ncclUniqueId nccl_id;
  if (rank == 0) CHECK_NCCL(ncclGetUniqueId(&nccl_id));
  MPI_Bcast(&nccl_id, sizeof(nccl_id), MPI_BYTE, 0, MPI_COMM_WORLD);

  ncclComm_t nccl_comm;
  CHECK_NCCL(ncclCommInitRank(&nccl_comm, num_ranks, nccl_id, rank));

  raft::resources handle;
  raft::comms::build_comms_nccl_only(&handle, nccl_comm, num_ranks, rank);

  // --- Demo parameters ---
  constexpr int64_t n_samples       = 100'000;
  constexpr int64_t n_features      = 32;
  constexpr int     n_clusters      = 10;
  constexpr int64_t streaming_batch = 10'000;
  constexpr float   cluster_spread  = 1.0f;
  constexpr float   center_range    = 30.0f;

  if (rank == 0) {
    std::printf("=== Multi-GPU KMeans Demo (%d ranks) ===\n", num_ranks);
    std::printf("Samples: %ld | Features: %ld | k: %d | batch: %ld\n\n",
                long(n_samples), long(n_features), n_clusters, long(streaming_batch));
  }

  // Generate synthetic blobs with well-separated cluster centers
  std::vector<float> h_data(n_samples * n_features);
  std::vector<int>   h_true_labels(n_samples);
  std::vector<float> cluster_centers(n_clusters * n_features);
  {
    std::mt19937 gen(12345);
    std::uniform_real_distribution<float> center_dist(-center_range, center_range);
    std::normal_distribution<float> noise(0.0f, cluster_spread);

    for (int c = 0; c < n_clusters; ++c)
      for (int d = 0; d < n_features; ++d)
        cluster_centers[c * n_features + d] = center_dist(gen);

    for (int64_t i = 0; i < n_samples; ++i) {
      int label = static_cast<int>(i % n_clusters);
      h_true_labels[i] = label;
      for (int d = 0; d < n_features; ++d)
        h_data[i * n_features + d] = cluster_centers[label * n_features + d] + noise(gen);
    }

    // Shuffle so labels aren't just sequential runs
    std::vector<int64_t> perm(n_samples);
    std::iota(perm.begin(), perm.end(), 0);
    std::shuffle(perm.begin(), perm.end(), gen);

    std::vector<float> tmp_data(h_data);
    std::vector<int>   tmp_labels(h_true_labels);
    for (int64_t i = 0; i < n_samples; ++i) {
      std::memcpy(h_data.data() + i * n_features,
                  tmp_data.data() + perm[i] * n_features,
                  n_features * sizeof(float));
      h_true_labels[i] = tmp_labels[perm[i]];
    }
  }

  int64_t base    = n_samples / num_ranks;
  int64_t rem     = n_samples % num_ranks;
  int64_t offset  = rank * base + std::min<int64_t>(rank, rem);
  int64_t n_local = base + (rank < rem ? 1 : 0);

  std::printf("[rank %d / GPU %d]  rows [%ld .. %ld)  (%ld samples)\n",
              rank, rank, long(offset), long(offset + n_local), long(n_local));

  auto X_local = raft::make_host_matrix_view<const float, int64_t>(
    h_data.data() + offset * n_features, n_local, n_features);

  auto d_centroids = raft::make_device_matrix<float, int64_t>(handle, n_clusters, n_features);

  cuvs::cluster::kmeans::params params;
  params.n_clusters           = n_clusters;
  params.max_iter             = 50;
  params.tol                  = 1e-4;
  params.init                 = cuvs::cluster::kmeans::params::KMeansPlusPlus;
  params.rng_state.seed       = 42;
  params.inertia_check        = true;
  params.streaming_batch_size = streaming_batch;

  float   inertia = 0.0f;
  int64_t n_iter  = 0;

  cuvs::cluster::kmeans::fit(handle,
                             params,
                             X_local,
                             std::nullopt,
                             d_centroids.view(),
                             raft::make_host_scalar_view(&inertia),
                             raft::make_host_scalar_view(&n_iter));

  auto stream = raft::resource::get_cuda_stream(handle);
  CHECK_CUDA(cudaStreamSynchronize(stream));

  if (rank == 0) {
    // --- Predict labels on the full dataset (on rank 0) ---
    auto d_X = raft::make_device_matrix<float, int64_t>(handle, n_samples, n_features);
    CHECK_CUDA(cudaMemcpy(d_X.data_handle(), h_data.data(),
                          sizeof(float) * n_samples * n_features, cudaMemcpyHostToDevice));

    auto d_labels = raft::make_device_vector<int64_t, int64_t>(handle, n_samples);
    float predict_inertia = 0.0f;

    cuvs::cluster::kmeans::predict(
      handle, params,
      raft::make_device_matrix_view<const float, int64_t>(d_X.data_handle(), n_samples, n_features),
      std::nullopt,
      raft::make_device_matrix_view<const float, int64_t>(
        d_centroids.data_handle(), n_clusters, n_features),
      d_labels.view(),
      false,
      raft::make_host_scalar_view(&predict_inertia));
    CHECK_CUDA(cudaStreamSynchronize(stream));

    std::vector<int64_t> h_labels(n_samples);
    CHECK_CUDA(cudaMemcpy(h_labels.data(), d_labels.data_handle(),
                          sizeof(int64_t) * n_samples, cudaMemcpyDeviceToHost));

    // --- Quality: permutation-invariant accuracy via majority voting ---
    // For each predicted cluster, find which true label appears most often.
    std::vector<std::vector<int64_t>> confusion(n_clusters, std::vector<int64_t>(n_clusters, 0));
    for (int64_t i = 0; i < n_samples; ++i)
      confusion[h_labels[i]][h_true_labels[i]]++;

    // Greedy matching: assign each predicted cluster to its dominant true label
    std::vector<int> pred_to_true(n_clusters, -1);
    std::vector<bool> true_taken(n_clusters, false);
    for (int round = 0; round < n_clusters; ++round) {
      int64_t best_count = -1;
      int best_pred = -1, best_true = -1;
      for (int p = 0; p < n_clusters; ++p) {
        if (pred_to_true[p] >= 0) continue;
        for (int t = 0; t < n_clusters; ++t) {
          if (true_taken[t]) continue;
          if (confusion[p][t] > best_count) {
            best_count = confusion[p][t];
            best_pred = p;
            best_true = t;
          }
        }
      }
      pred_to_true[best_pred] = best_true;
      true_taken[best_true] = true;
    }

    int64_t correct = 0;
    std::vector<int64_t> cluster_sizes(n_clusters, 0);
    std::vector<int64_t> cluster_correct(n_clusters, 0);
    for (int64_t i = 0; i < n_samples; ++i) {
      int p = static_cast<int>(h_labels[i]);
      cluster_sizes[p]++;
      if (h_true_labels[i] == pred_to_true[p]) {
        ++correct;
        ++cluster_correct[p];
      }
    }
    double accuracy = 100.0 * correct / n_samples;

    // --- Compute centroid-to-true-center distances ---
    std::vector<float> h_centroids(n_clusters * n_features);
    CHECK_CUDA(cudaMemcpy(h_centroids.data(), d_centroids.data_handle(),
                          sizeof(float) * n_clusters * n_features, cudaMemcpyDeviceToHost));

    std::printf("\n============ Multi-GPU KMeans Results ============\n");
    std::printf("  Ranks             : %d\n", num_ranks);
    std::printf("  Total samples     : %ld\n", long(n_samples));
    std::printf("  Features          : %ld\n", long(n_features));
    std::printf("  Clusters (k)      : %d\n", n_clusters);
    std::printf("  Streaming batch   : %ld\n", long(streaming_batch));
    std::printf("  Lloyd iterations  : %ld\n", long(n_iter));
    std::printf("  Final inertia     : %.6f\n", double(inertia));
    std::printf("  Predict inertia   : %.6f\n", double(predict_inertia));
    std::printf("\n  --- Clustering Quality ---\n");
    std::printf("  Overall accuracy  : %.2f%% (%ld / %ld)\n",
                accuracy, long(correct), long(n_samples));

    std::printf("\n  Per-cluster breakdown:\n");
    std::printf("  %6s  %10s  %10s  %8s  %12s\n",
                "Pred", "TrueLabel", "Size", "Acc%", "CentroidErr");
    for (int p = 0; p < n_clusters; ++p) {
      int t = pred_to_true[p];
      double pct = cluster_sizes[p] > 0
                     ? 100.0 * cluster_correct[p] / cluster_sizes[p]
                     : 0.0;

      // L2 distance between learned centroid and ground truth center
      double dist2 = 0.0;
      for (int d = 0; d < n_features; ++d) {
        double diff = h_centroids[p * n_features + d] - cluster_centers[t * n_features + d];
        dist2 += diff * diff;
      }
      std::printf("  %6d  %10d  %10ld  %7.2f%%  %12.4f\n",
                  p, t, long(cluster_sizes[p]), pct, std::sqrt(dist2));
    }

    std::printf("\n  Expected accuracy for well-separated blobs: >99%%\n");
    if (accuracy >= 99.0)
      std::printf("  PASS: Clustering quality is high.\n");
    else if (accuracy >= 90.0)
      std::printf("  WARN: Clustering quality is acceptable but not ideal.\n");
    else
      std::printf("  FAIL: Clustering quality is poor!\n");

    std::printf("==================================================\n");
  }

  CHECK_NCCL(ncclCommDestroy(nccl_comm));
  MPI_Finalize();
  return 0;
}
```

</details>

<details>
<summary>Compilation command</summary>

```
nvcc -std=c++17 -x cu --extended-lambda -arch=native       \
 -I$CONDA_PREFIX/include/rapids                            \
 -I$CONDA_PREFIX/include                                   \
 demo_mg_kmeans_raft_comms.cu                              \
 -L$CONDA_PREFIX/lib -lcuvs -lnccl -lrmm -lmpi             \
 -lucxx -lucp -lucs                                       \
 -Xlinker=-rpath,$CONDA_PREFIX/lib                         \
 -o demo_mg_kmeans
```

</details>

<details>
<summary>Launch command</summary>

`mpirun -np 2 ./demo_mg_kmeans`

</details>

cpp/src/cluster/detail/kmeans_mg_batched.cuh
```diff
@@ -0,0 +1,582 @@
...
+      if (params.inertia_check) { raft::matrix::fill(dev_res, clustering_cost.view(), T{0}); }
+
+      auto rank_centroids_const = raft::make_device_matrix_view<const T, IdxT>(
+        rank_centroids.data_handle(), n_clusters, n_features);
+
+      // Phase 1: local batch accumulation (skip if no local data)
+      if (has_data) {
+        auto& data_batches = *data_batches_opt;
+        data_batches.reset();
+        data_batches.prefetch_next_batch();
```
tarang-jain, line 311: prefetch needs to be evaluated. If we have to reduce the streaming batch size in order to enable prefetch and fit two batches on the device, it might actually be slower. The lesser the number of batches, faster the iteration.

cpp/src/cluster/detail/kmeans_mg_batched.cuh
```diff
@@ -0,0 +1,582 @@
...
+        for (const auto& data_batch : data_batches) {
```
tarang-jain, line 312: I have separated out all the computations to be done for a single batch into a new function `process_batch` [here](https://github.com/rapidsai/cuvs/pull/2015). We should be able to reuse that entirely here.

cpp/src/cluster/detail/kmeans_mg_batched.cuh
```diff
@@ -0,0 +1,582 @@
...
+      SNMG_GROUP_END();
+      raft::resource::sync_stream(dev_res);
+
+      // Phase 3: finalize centroids
+      auto centroid_sums_const = raft::make_device_matrix_view<const T, IdxT>(
+        centroid_sums.data_handle(), n_clusters, n_features);
+      auto weight_per_cluster_const =
+        raft::make_device_vector_view<const T, IdxT>(weight_per_cluster.data_handle(), n_clusters);
+
+      cuvs::cluster::kmeans::detail::finalize_centroids<T, IdxT>(dev_res,
```
tarang-jain, line 390: Does this mean that every rank is performing the `finalize_centroids` call and computing the exact same centroids?

cpp/src/cluster/detail/kmeans_mg_batched.cuh
```diff
@@ -0,0 +1,582 @@
...
+template <typename T>
+void batched_fit_omp(const raft::resources& clique,
+                     const cuvs::cluster::kmeans::params& params,
+                     raft::host_matrix_view<const T, int64_t> X,
+                     std::optional<raft::host_vector_view<const T, int64_t>> sample_weight,
+                     raft::device_matrix_view<T, int64_t> centroids,
+                     raft::host_scalar_view<T> inertia,
+                     raft::host_scalar_view<int64_t> n_iter)
+{
+  using IdxT = int64_t;
```
tarang-jain: Nit: why dont we template on IdxT and simply instantiate only the main fit function with the correct types.

cpp/src/cluster/detail/kmeans_mg_batched.cuh
```diff
@@ -0,0 +1,582 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION.
+ * SPDX-License-Identifier: Apache-2.0
+ */
```
tarang-jain, line 4: Can we get rid of this file entirely and combine with regular mg kmeans (just as we are doing in PR #2015)? Is that possible?

tarang-jain: Also, MNMG should be able to reuse the `snmg_fit` function (for a single worker) as is, right? Except that the nccl reduce macro will be replaced by something like `comms.allreduce()`

cpp/src/cluster/detail/kmeans_mg_batched.cuh
viclafargue: Yes, exactly. We do the same in [the non-batched MG version of KMeans](https://github.com/rapidsai/cuvs/blob/988ebf565925fd4af278cf0448d40db2b84f66de/cpp/src/cluster/detail/kmeans_mg.cuh#L646-L694). The new centroids can only be computed after every worker has completed processing its batches and shared the local results via the `allreduce` operations. After the `allreduce`, every rank holds identical `centroid_sums` and `weight_per_cluster`, so each rank independently computes the same new centroids. Computing them on a single GPU and broadcasting the (possibly large) centroid array would not be any faster. The redundant local division is cheap compared to an extra communication round.

tarang-jain: Yeah makes sense. Thanks!

tarang-jain:
Referencing this comment here: https://github.com/rapidsai/cuvs/pull/2015#discussion_r3126928156
@viclafargue please ensure to incorporate the inertia related changes:
1. rel_tol with the batched weight addition
2. Removal of the inertia check parameter
3. wt_sum = 0 check if all-zero sample weights are passed. Check if you can reuse the weightSum helper from #2015.

cpp/src/cluster/detail/kmeans_mg_batched.cuh
viclafargue: I removed the next batch prefetching (now similar to the single GPU path).

cpp/src/cluster/detail/kmeans_mg_batched.cuh
viclafargue: I refactored the code to work flawlessly with the single GPU refactor. The `process_batch` function is now being used. The weight normalization is improved (`rel_tol` and zero/invalid check). The `inertia_check` field is not used anymore.

However, I do not feel confident implementing a massive unifying refactor in this PR that is originally dedicated to introducing MG Batched KMeans. Would it be fine to leave things as is for now? We could come back to it in a dedicated follow-up PR.

tarang-jain: If we can get this and the refactor into 26.06, its ok to break it down. We'll need to track the MG refactor in an issue and that would be a high priority.

tarang-jain:
Can you pull the latest version of my branch in #2015? That will automatically fix the merge conflicts.

cpp/src/cluster/detail/kmeans_common.cuh
```diff
@@ -129,43 +133,36 @@ void countLabels(raft::resources const& handle,
...
+  auto n_samples = weight.extent(0);
+
+  DataT wt_sum = DataT{0};
+  if constexpr (raft::is_device_mdspan_v<decltype(weight)>) {
+    auto stream   = raft::resource::get_cuda_stream(handle);
+    auto d_wt_sum = raft::make_device_scalar<DataT>(handle, DataT{0});
+    raft::linalg::mapThenSumReduce(
+      d_wt_sum.data_handle(), n_samples, raft::identity_op{}, stream, weight.data_handle());
+    raft::copy(&wt_sum, d_wt_sum.data_handle(), 1, stream);
+    raft::resource::sync_stream(handle);
```
tarang-jain: Again just pull in the latest version here. I have removed the sync stream

cpp/src/cluster/detail/kmeans_mg.cuh
```diff
@@ -484,16 +482,21 @@ void checkWeights(const raft::resources& handle,
                         stream);
   DataT wt_sum = wt_aggr.value(stream);
   raft::resource::sync_stream(handle, stream);
```
tarang-jain: If accumulating in a device scalar, we might be able to avoid syncing the stream

cpp/src/cluster/detail/kmeans_mg.cuh
```diff
@@ -557,9 +560,15 @@ void fit(const raft::resources& handle,
...
-  auto wtInCluster = raft::make_device_vector<DataT, IndexT>(handle, n_clusters);
+  // Running weighted sum of samples per cluster and sum of weights per cluster,
+  // accumulated by process_batch and allreduced each iteration.
+  auto centroid_sums      = raft::make_device_matrix<DataT, IndexT>(handle, n_clusters, n_features);
+  auto weight_per_cluster = raft::make_device_vector<DataT, IndexT>(handle, n_clusters);
+  auto clustering_cost    = raft::make_device_scalar<DataT>(handle, DataT{0});
+
+  // Separate workspace for compute_centroid_adjustments (distinct from the one
+  // used by minClusterAndDistanceCompute inside process_batch).
+  rmm::device_uvector<char> batch_workspace(0, stream);
```
tarang-jain: I am tempted to put the entire batching logic itself in a helper, so that both kmeans and kmeans_mg can use it identically without duplicating code at all. But that will be a significant amount of work in a follow up

cpp/src/cluster/detail/kmeans_mg_batched.cuh
```diff
@@ -0,0 +1,569 @@
...
+        data_batches.reset();
+        for (const auto& data_batch : data_batches) {
+          IdxT current_batch_size = static_cast<IdxT>(data_batch.size());
+
+          raft::matrix::fill(dev_res, batch_clustering_cost.view(), T{0});
+
+          auto batch_data_view = raft::make_device_matrix_view<const T, IdxT>(
+            data_batch.data(), current_batch_size, n_features);
+
+          cuvs::cluster::kmeans::detail::copy_and_scale_batch_weights<T, IdxT>(dev_res,
```
tarang-jain: this helper has been removed. We are now caching weights in pinned memory. Question: can we have a single CPU cache for the weights even in the multi-GPU case? If not, we can always recompute it for each batch.

cpp/src/cluster/detail/kmeans_mg_batched.cuh
```diff
@@ -0,0 +1,569 @@
...
+            L2NormBatch_const,
+            L2NormBuf_OR_DistBuf,
+            metric,
+            params.batch_samples,
+            params.batch_centroids,
+            workspace);
+
+          auto minClusterAndDistance_const = raft::make_const_mdspan(minClusterAndDistance_view);
+
+          cuvs::cluster::kmeans::detail::accumulate_batch_centroids<T, IdxT>(
```
tarang-jain: we no longer need this accumulation. It is taken case of because the atomic add now does not reset the sums.

cpp/src/cluster/detail/kmeans_mg_batched.cuh
```diff
@@ -0,0 +1,569 @@
...
+      }
+    } else {
+      if (rank == 0) {
+        raft::copy(
+          rank_centroids.data_handle(), centroids.data_handle(), n_clusters * n_features, stream);
+      }
+    }
+    raft::resource::sync_stream(dev_res);
+    SNMG_BCAST(rank_centroids.data_handle(), n_clusters * n_features, 0);
+    raft::resource::sync_stream(dev_res);
```
tarang-jain: we probably dont need to sync so much, right? The per device streams will order things automatically. So the next operation wont start until the broadcast is complete isnt it?

tarang-jain: Unless you are saying that this is needed for the OpenMP route, but not in the partition comms

cpp/src/cluster/detail/kmeans_mg_batched.cuh
```diff
@@ -0,0 +1,569 @@
...
+
+      // Phase 2: grouped allreduce
+      SNMG_GROUP_START();
+      SNMG_ALLREDUCE(
+        centroid_sums.data_handle(), centroid_sums.data_handle(), n_clusters * n_features);
+      SNMG_ALLREDUCE(
+        weight_per_cluster.data_handle(), weight_per_cluster.data_handle(), n_clusters);
+      SNMG_ALLREDUCE(clustering_cost.data_handle(), clustering_cost.data_handle(), 1);
+      SNMG_GROUP_END();
+      raft::resource::sync_stream(dev_res);
```
tarang-jain: again -- why sync each stream here? The next operations are ordered on the correct per-device stream , right? So that guarantees the ALLREDUCE is completed.

cpp/src/cluster/detail/kmeans_mg_batched.cuh
viclafargue: Yes, we can scale the weight once in pinned memory too. For the OpenMP case, each rank receives a cache and fills its own slice. For the multi-process case, each rank allocates and fills its own pinned memory cache.

cpp/src/cluster/detail/kmeans_mg_batched.cuh
```diff
@@ -0,0 +1,643 @@
...
+      // run `check_convergence` via `map_offset`, allreduce the flag,
+      // shadow into pinned host. Consumed at top of next iteration.
+      cuvs::cluster::kmeans::detail::compute_centroid_shift<T, IdxT>(
+        dev_res,
+        raft::make_const_mdspan(rank_centroids.view()),
+        raft::make_const_mdspan(new_centroids.view()),
+        sqrd_norm_error_dev.view());
+
+      raft::copy(
+        rank_centroids.data_handle(), new_centroids.data_handle(), n_clusters * n_features, stream);
```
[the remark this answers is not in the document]
tarang-jain: cosine is not supported for kmeans

cpp/src/cluster/kmeans_impl.cuh
```diff
@@ -18,8 +18,12 @@ void fit_main(raft::resources const& handle,
...
               rmm::device_uvector<char>& workspace)
 {
-  cuvs::cluster::kmeans::detail::kmeans_fit_main<DataT, IndexT>(
-    handle, params, X, sample_weights, centroids, inertia, n_iter, workspace);
+  cuvs::cluster::kmeans::params p = params;
+  p.init                          = kmeans::params::InitMethod::Array;
+  auto sw                         = std::make_optional(
+    raft::make_device_vector_view<const DataT, IndexT>(sample_weights.data_handle(), X.extent(0)));
+  cuvs::cluster::kmeans::detail::kmeans_fit(
+    handle, p, X, sw, centroids, inertia, n_iter, std::ref(workspace));
```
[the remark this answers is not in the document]
tarang-jain: @viclafargue I sort of agree with this. We should be comparing the extend of X and the sample weights if they are passed. Can we add an assertion for this if we already dont have on?

cpp/src/cluster/detail/kmeans_mg_batched.cuh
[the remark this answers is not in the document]
tarang-jain: But @viclafargue I do feel that we should be checking for the metric somewhere. So the metric needs to be checked in both kmeans_fit and predict. I think this was being done indirectly via minClusterDistanceCompute.cu earlier, but since we started using the fused path for cosine, cosine will pass through and a metric assertion might be missing.

tarang-jain:
The KMeans++ init on only rank zero can be quite problematic.
1. The shard might not be representative of the whole dataset
2. The shard's size might be smaller than the user passed init_size parameter.
I think the rabbit comment makes sense here.

cpp/src/cluster/detail/kmeans_mg_batched.cuh
```diff
@@ -0,0 +1,643 @@
...
+  rmm::device_uvector<char> batch_workspace(0, stream);
+
+  // Weight normalization: rescale so global weights sum to global_n
+  T weight_scale = T{1};
+  if (sample_weight.has_value()) {
+    auto d_n_local = raft::make_device_scalar<IdxT>(dev_res, static_cast<IdxT>(n_local));
+    SNMG_ALLREDUCE(d_n_local.data_handle(), d_n_local.data_handle(), 1);
+    IdxT global_n{};
+    raft::copy(&global_n, d_n_local.data_handle(), 1, stream);
+    raft::resource::sync_stream(dev_res);
```
tarang-jain: we dont need to sync after every scalar copy here. Instead if we really have to, we can sync after all the scalars have been copied, so after global_n and global_n_wt. Actually, why dont we have those as scalars on device? Then we can avoid syncing at all.

tarang-jain: Also check if you can use the device convergence checker from kmeans_common.cuh here.

cpp/src/cluster/detail/kmeans_mg_batched.cuh
```diff
@@ -0,0 +1,643 @@
...
+    raft::copy(
+      rank_centroids.data_handle(), best_centroids.data_handle(), n_clusters * n_features, stream);
+    local_inertia = best_inertia;
+    local_n_iter  = best_n_iter;
+  }
+
+  if (rank == 0) {
+    raft::copy(
+      centroids.data_handle(), rank_centroids.data_handle(), n_clusters * n_features, stream);
+    raft::resource::sync_stream(dev_res);
```
tarang-jain: do we need this sync? Its the same as `stream` isnt it?

zhilizju:
Hi @viclafargue @tarang-jain , thanks for the excellent work on this PR!

  We have a large-scale clustering pipeline that requires multi-GPU batched
   KMeans — we're running on hundreds of nodes and need this capability to
  handle our dataset scale. We're eagerly waiting to adopt this feature
  once it lands.

  Could you share a rough timeline for when this might be ready to merge?
  Do you think the remaining items (rebase on #2015, the KMeans++ init
  concern, etc.) could be wrapped up by this weekend, or would it need
  another week or so? We'd like to plan our integration work accordingly.

  We'd be happy to be among the first adopters — once this is in a testable
   state, we can run it against our production workload and provide detailed feedback on correctness,
  scalability, and API ergonomics. Let us know if there's a branch or
  pre-release build we can start testing against.

  Thanks again for pushing this forward!

cjnolet:
Hi @zhilizju,

This **should** be merged by early next week, but just know that we are in the process of doing large-scale benchmarks, so those will likely not be complete before merging (this mileage might vary).


This will be part of our 26.06 release, which will be officially released in early June. 

Will you be integrating this through c++ or another one of the language wrappers?

cjnolet:
> The shard might not be representative of the whole dataset

@tarang-jain it's usually safe to assume the data is randomly scattered. Kmeans++ is also expensive enough, especially with larger k, that I think doing the initialization on a single partition is going to be a good trade off most of the time. I'd even go as far as to say that we could offer an option for this in the mnmg version, and default it to preferring a single shard. Think of it like taking a simple random sample of a simple random sample (aka a random sample from the entire training set).

tarang-jain:
yeah I think that makes sense -- random sample of a random sample.

cpp/tests/cluster/kmeans_mg_batched.cu
```diff
@@ -0,0 +1,382 @@
...
+    int n_features = testparams_.n_col;
+    int n_clusters = testparams_.n_clusters;
+    int num_ranks  = raft::resource::get_num_ranks(clique_);
+
+    auto stream = raft::resource::get_cuda_stream(clique_);
+
+    auto X      = raft::make_device_matrix<T, int>(clique_, n_samples, n_features);
+    auto labels = raft::make_device_vector<int, int>(clique_, n_samples);
+
+    raft::random::make_blobs<T, int>(X.data_handle(),
```
tarang-jain: So I added a helper in `kmeans.cu` to prepare the blobs. Can we reuse it? I know that is not a header though, so including it in this TU might not be possible.

cpp/tests/cluster/kmeans_mg_batched.cu
```diff
@@ -0,0 +1,382 @@
...
+    // --- Evaluate: compare SNMG labels with reference (make_blobs) labels ---
+    ari_vs_ref_ = raft::stats::adjusted_rand_index(
+      d_labels_ref.data(), d_labels_snmg.data(), n_samples, sg_stream);
+
+    // ARI between SNMG and single-GPU results
+    ari_vs_sg_ = raft::stats::adjusted_rand_index(
+      d_labels_sg.data(), d_labels_snmg.data(), n_samples, sg_stream);
+
+    raft::resource::sync_stream(sg_handle, sg_stream);
+
```
tarang-jain, line 218: I like that you  are testing both -- ARI in the random init and bit matching centroids in the array init. That should give us a robust set of tests.

zhilizju:
> Hi @zhilizju,
> 
> This **should** be merged by early next week, but just know that we are in the process of doing large-scale benchmarks, so those will likely not be complete before merging (this mileage might vary).
> 
> This will be part of our 26.06 release, which will be officially released in early June.
> 
> Will you be integrating this through c++ or another one of the language wrappers?

Thanks @cjnolet! Great to hear it's targeting early next week.

  We're already running the RAFT comms + Ray path (via cuML 25.10 _fit(multigpu=True) with NCCL across hundreds of
  nodes), so we're very familiar with this architecture. Looking forward to the cuVS native implementation for better performance with host-resident batched streaming.

  We'll primarily use the Python wrapper. A couple of questions:

  1. Once merged, will the multi-GPU batched KMeans be accessible from Python with the RAFT comms path? We'd like
  to integrate it into our existing Ray + NCCL setup.
  2. Is there a nightly conda package that will pick this up automatically after merge, or would we need to build from
  source?

cpp/src/cluster/detail/kmeans_mg_batched.cuh
viclafargue: I placed the two reduce operations into a single NCCL group and also made sure to sync only once following `global_n` and `global_wt` copies. The `check_convergence` device function is already being used.

cpp/src/cluster/detail/kmeans_mg_batched.cuh
viclafargue: Nope, fixed that.

cpp/tests/cluster/kmeans_mg_batched.cu
viclafargue: Reusing `prepareBlobInputs` as a whole was not really possible, but I could create a `kmeans_test_blobs.cuh` file to make sure that we have some code re-use between the two tests.

viclafargue:
Hi @zhilizju,

> Is there a nightly conda package that will pick this up automatically after merge, or would we need to build from source?

The [cuVS nightly package](https://anaconda.org/channels/rapidsai-nightly/packages/cuvs/files) will be available following the merge. However, it will only offer a C++ interface.

> Once merged, will the multi-GPU batched KMeans be accessible from Python with the RAFT comms path? We'd like to integrate it into our existing Ray + NCCL setup.

The plan so far would be to have the Python wrapper be placed in cuML in similar fashion to the regular MG KMeans.

cpp/src/cluster/detail/kmeans_mg_batched.cuh
```diff
@@ -0,0 +1,666 @@
...
+                                                                 weight_per_cluster_const,
+                                                                 rank_centroids_const,
+                                                                 new_centroids.view());
+
+      // Phase 4: device-side convergence evaluation. Compute shift,
+      // run `check_convergence` via `map_offset`, allreduce the flag,
+      // shadow into pinned host. Consumed at top of next iteration.
+      cuvs::cluster::kmeans::detail::compute_centroid_shift<T, IdxT>(
+        dev_res,
+        raft::make_const_mdspan(rank_centroids.view()),
```
tarang-jain: why not use rank_centroids_const here. The const view already exists.

cpp/src/cluster/detail/kmeans_mg_batched.cuh
```diff
@@ -0,0 +1,666 @@
...
+  auto n_clusters = static_cast<IdxT>(params.n_clusters);
+  auto metric     = params.metric;
+
+  RAFT_EXPECTS(n_clusters > 0, "n_clusters must be positive");
+  RAFT_EXPECTS(static_cast<IdxT>(centroids.extent(0)) == n_clusters,
+               "centroids.extent(0) must equal n_clusters");
+  RAFT_EXPECTS(centroids.extent(1) == n_features, "centroids.extent(1) must equal n_features");
+  RAFT_EXPECTS(num_ranks > 0, "num_ranks must be positive");
+
+  RAFT_LOG_DEBUG("SNMG KMeans fit: rank=%d/%d, n_local=%zu, n_features=%zu, n_clusters=%d",
```
jinsolp: maybe change log names to MNMG Kmeans to match the function name?

cpp/src/cluster/detail/kmeans_mg_batched.cuh
```diff
@@ -0,0 +1,666 @@
...
+
+  for (int seed_iter = 0; seed_iter < n_init; ++seed_iter) {
+    cuvs::cluster::kmeans::params iter_params = params;
+    iter_params.rng_state.seed                = gen();
+
+    // Centroid init: rank 0 produces, then broadcast to all ranks
+    if (iter_params.init != cuvs::cluster::kmeans::params::InitMethod::Array) {
+      if (rank == 0) {
+        cuvs::cluster::kmeans::detail::init_centroids_from_host_sample<T, IdxT>(
+          dev_res, iter_params, streaming_batch_size, X_local, rank_centroids.view(), workspace);
```
jinsolp: Can we document that we require a well-shuffled dataset to run MG kmeans? Otherwise this will end up doing bad init because is only uses rank0 local data.

cpp/src/cluster/detail/kmeans_mg_batched.cuh
```diff
@@ -0,0 +1,666 @@
...
+    raft::copy(
+      rank_centroids.data_handle(), best_centroids.data_handle(), n_clusters * n_features, stream);
+    local_inertia = best_inertia;
+    local_n_iter  = best_n_iter;
+  }
+
+  bool write_outputs = !use_nccl || rank == 0;
+  if (write_outputs) {
+    raft::copy(
+      centroids.data_handle(), rank_centroids.data_handle(), n_clusters * n_features, stream);
```
jinsolp, line 504: should we sync before returning if the caller expects `centroids` to be ready?

jinsolp (commented):
Thanks @viclafargue ! No major concerns from my side

cpp/src/cluster/detail/kmeans_mg_batched.cuh
cjnolet: Definitely should go one step further here and add an option to switch between "use kmeans++ globally and use kmeans++ locally".

cpp/src/cluster/detail/kmeans_mg_batched.cuh
viclafargue: Yes, you're right, the user would most likely use the same stream, but this is much safer to simply sync before returning.

viclafargue:
The issue of the bias introduced by single rank sampling is now solved.
Here is how the current implementation works  :
- **Array :** simple broadcast
- **Random :** samples `n_clusters` unique rows globally
- **KMeansPlusPlus :** samples `init_sample_size` unique rows globally, root runs local `kmeansPlusPlus` or `initScalableKMeansPlusPlus` on this sample data.

This keeps the implementation simple and memory bounded: the full input remains on host, and only the initialization sample is materialized on device. After the global sample is built, there is no additional multi-rank communication during KMeans++ initialization.

Using true multi-GPU KMeans++ for very large `init_sample_size` is possible, but it would be a larger follow-up. It would likely require:

- An API/ABI-sensitive parameter or policy to opt into distributed initialization.
- Large rRefactoring of MG KMeans++ implementation to accommodate for different comm system, distributed host-row sampling, handling of empty-rank cases and variable-size candidate gathers.

Given the added review surface, I think the current sampled-root initialization is the right approach for this PR. True distributed KMeans++ initialization can be tracked separately if we see workloads where `init_sample_size` becomes a bottleneck.

cpp/src/cluster/detail/kmeans_mg_batched_init.cuh
```diff
@@ -0,0 +1,249 @@
...
+namespace cuvs::cluster::kmeans::mg::detail {
+
+template <typename IndexT>
+IndexT get_global_kmeanspp_init_sample_size(const cuvs::cluster::kmeans::params& params,
+                                            IndexT global_n,
+                                            IndexT n_clusters)
+{
+  IndexT default_init_size = std::min(static_cast<IndexT>(std::int64_t{3} * n_clusters), global_n);
+  IndexT init_sample_size  = params.init_size > 0
+                               ? std::min(static_cast<IndexT>(params.init_size), global_n)
```
tarang-jain, line 38: Question: the same `init_size` will be available to all the ranks, right?

cpp/src/cluster/detail/kmeans_mg_batched_init.cuh
```diff
@@ -0,0 +1,249 @@
...
+                                                 IndexT sample_size,
+                                                 std::uint64_t seed)
+{
+  std::mt19937 gen(seed);
+  std::vector<IndexT> sampled;
+  sampled.reserve(static_cast<std::size_t>(sample_size));
+  std::unordered_set<IndexT> selected;
+  selected.reserve(static_cast<std::size_t>(sample_size));
+
+  for (IndexT j = n_samples - sample_size; j < n_samples; ++j) {
```
tarang-jain, line 54: why not do this on device? I see below you are copying the sampled IDs to device anyway on rank 0.

tarang-jain (commented):
this idea of global samples seems sound to me. I think the underlying assumption of contiguous sample IDs across ranks holds. We should definitely benchmark such an initialization since the creation of the init sample is done on host? one sample at a time?

tarang-jain:
In the kmeanspp init, I think there is one more assumption -- which is that the init sample fits on rank 0. We should document this, otherwise user might get surprise OOM (for example an edge case can be when rank 0 has less available GPU RAM).

cpp/src/cluster/detail/kmeans_mg_batched.cuh
```diff
@@ -0,0 +1,647 @@
...
+    MNMG_GROUP_START();
+    MNMG_ALLREDUCE(d_n_local.data_handle(), d_n_local.data_handle(), 1);
+    MNMG_ALLREDUCE(d_wt.data_handle(), d_wt.data_handle(), 1);
+    MNMG_GROUP_END();
+
+    IdxT global_n{};
+    T global_wt{};
+    raft::copy(&global_n, d_n_local.data_handle(), 1, stream);
+    raft::copy(&global_wt, d_wt.data_handle(), 1, stream);
+    raft::resource::sync_stream(dev_res);
```
tarang-jain: this means all the ranks are syncing for the exact same copy. there should be a better way of doing this. Why do we need the sum to be on host? We can scale the weights of each batch on device, no?

cpp/src/cluster/detail/kmeans_mg_batched.cuh
```diff
@@ -0,0 +1,647 @@
...
+    data_batches_opt.emplace(X_local.data_handle(),
+                             n_local,
+                             n_features,
+                             streaming_batch_size,
+                             stream,
+                             rmm::mr::get_current_device_resource_ref(),
+                             true);
+  }
+
+  bool need_compute_norms = metric == cuvs::distance::DistanceType::L2Expanded ||
```
tarang-jain: we dont need this boolean. We already have an assertion to check that the metric is either L2Expanded or L2SqrtExpanded.

cpp/src/cluster/detail/kmeans_mg_batched_init.cuh
viclafargue: Yes, it comes from the param struct that is the same for all ranks.

cpp/src/cluster/detail/kmeans_mg_batched_init.cuh
viclafargue: The samples are copied to device to be broadcasted through NCCL. Finding unique sample IDs is probably better suited on CPU. For normal n_clusters / init sample sizes, host sampling + one small H2D copy is likely faster and simpler.

cpp/src/cluster/detail/kmeans_mg_batched.cuh
viclafargue: Yes, I updated the code so that the scaled weights stay on device.

cpp/src/cluster/detail/kmeans_mg_batched.cuh
```diff
@@ -0,0 +1,608 @@
...
+      raft::linalg::map_offset(
+        dev_res,
+        raft::make_device_vector_view<int64_t, int>(d_done_flag.data_handle(), 1),
+        [=] __device__(int) {
+          cuvs::cluster::kmeans::detail::check_convergence(
+            d_cost_view, d_prior_view, d_norm_view, tol, iter, d_done_view);
+          return *d_done_view.data_handle();
+        });
+
+      MNMG_ALLREDUCE(d_done_flag.data_handle(), d_done_flag.data_handle(), 1);
```
tarang-jain: Why are we not doing MNMG_GROUP_START and MNMG_GROUP_END here?

cpp/src/cluster/detail/kmeans_mg_batched.cuh
```diff
@@ -0,0 +1,608 @@
...
+      _mnmg_comm.allreduce(sendbuf, recvbuf, count, raft::comms::op_t::SUM, stream);      \
+    }                                                                                     \
+  } while (0)
+
+#define MNMG_BCAST(buf, count, root)                                                           \
+  do {                                                                                         \
+    using _mnmg_bcast_t = std::remove_pointer_t<decltype(buf)>;                                \
+    if (use_nccl) {                                                                            \
+      RAFT_NCCL_TRY(                                                                           \
+        ncclBroadcast(buf, buf, count, nccl_dtype<_mnmg_bcast_t>(), root, nccl_comm, stream)); \
```
tarang-jain: Question: can we reuse these from raft instead of creating new macros?

We have for example https://github.com/rapidsai/raft/blob/c38237f7b17bf9b4d52fb416d298573dada40bf8/cpp/include/raft/comms/detail/std_comms.hpp#L380.

Also, if you absolutely have to define new macros, can we avoid duplication in this header and kmeans_mg_batched_init.cuh?

tarang-jain (approved):
Approving apart from the small comments above.

cpp/src/cluster/detail/kmeans_mg_batched_init.cuh
tarang-jain: Unique IDs are not necessary, right? We can sample with repetition IIRC.

cpp/src/cluster/detail/kmeans_mg_batched.cuh
viclafargue: NCCL groups are only useful to queue a series of NCCL communication operation. There is no purpose in using a group for a single operation.

cpp/src/cluster/detail/kmeans_mg_batched.cuh
viclafargue: RAFT comms for multi-process and raw NCCL operations for multi-thread work differently and cannot be merged at the moment. I refactored the comms for MG Batched KMeans into a special class in `kmeans_mg_batched_comms.cuh` for now. This avoid instantiating and passing along lambda functions for comms. In the long term we would probably want to have some of this code to be moved into RAFT.

cpp/src/cluster/detail/kmeans_mg_batched.cuh
```diff
@@ -0,0 +1,522 @@
...
+
+    comms.group_start();
+    comms.allreduce(d_global_n->data_handle(), d_global_n->data_handle(), 1);
+    comms.allreduce(d_global_wt->data_handle(), d_global_wt->data_handle(), 1);
+    comms.group_end();
+
+    if (has_data) {
+      const IdxT* d_global_n_ptr = d_global_n->data_handle();
+      const T* d_global_wt_ptr   = d_global_wt->data_handle();
+      raft::linalg::map(
```
dantegd, line 199: We are saving a reduce here by not checking that wt_sum > 0, right? Checking that not checking the weights not all being zero is intentional, I wonder what perf impact would the allreduce sum to host memory have, the division by zero wouldn't be a nice failure mode, so wanted to quickly get your thoughts on that.

cpp/src/cluster/detail/kmeans_mg_batched.cuh
```diff
@@ -0,0 +1,522 @@
...
+
+    if (has_data && !sample_weight.has_value()) {
+      raft::matrix::fill(dev_res, batch_weights.view(), T{1});
+    }
+
+    // Reset per-pass convergence state to avoid leaking it across n_init.
+    raft::matrix::fill(dev_res, d_prior_cost.view(), T{0});
+    *h_done_flag.data_handle() = 0;
+
+    for (local_n_iter = 1; local_n_iter <= iter_params.max_iter; ++local_n_iter) {
```
dantegd, line 297: Curious about the reported n_iter on saturation, if the early-break path doesn't fire, local_n_iter falls out of the loop at max_iter + 1, and that's what gets written into n_iter[0]. The test ASSERT_LE(snmg_n_iter_, max_iter) only passes today because every test config converges in 1 iter; the max_iter=2 case happens to land in 1 iter on separated blobs. I think the single-GPU kmeans_fit has the same quirk, so this isn't unique to MG, but should we either clamp to max_iter at the bottom of the loop or add a saturation test that actually doesn't converge (e.g. random init at max_iter=2) to make sure we have the behavior we think we do?

cpp/include/cuvs/cluster/kmeans.hpp
```diff
@@ -125,6 +125,10 @@ struct params : base_params {
    * When set to 0 (default) with host data uses `min(3 * n_clusters, n_samples)`
    * as a default.
    *
+   * In Batched multi-GPU host-data fits, the effective KMeansPlusPlus initialization
+   * sample is materialized on device and seeding is run on rank 0. Rank 0
```
dantegd: Small doc correction: looking at sample_global_host_rows, every rank allocates the full sample_size × n_features device matrix and participates in the zero-masked allreduce, so the memory bound applies to all ranks, not just rank 0. Only the seeding workspace (the buffers inside kmeansPlusPlus / initScalableKMeansPlusPlus) is rank 0 only. Could we tweak the wording to "every rank must have enough GPU memory for the initialization sample, and rank 0 also needs the seeding workspace" or something like that? Otherwise users will plan capacity wrong.

cpp/src/cluster/detail/kmeans_mg_batched_init.cuh
```diff
@@ -0,0 +1,249 @@
...
+                                            IndexT global_n,
+                                            IndexT n_clusters)
+{
+  IndexT default_init_size = std::min(static_cast<IndexT>(std::int64_t{3} * n_clusters), global_n);
+  IndexT init_sample_size  = params.init_size > 0
+                               ? std::min(static_cast<IndexT>(params.init_size), global_n)
+                               : default_init_size;
+  if (params.streaming_batch_size > 0) {
+    init_sample_size = std::min(init_sample_size, static_cast<IndexT>(params.streaming_batch_size));
+  }
```
dantegd: The single-GPU host path lets params.init_size go up to n_samples and only logs about the default. Here, a user who explicitly sets params.init_size = 1_000_000, params.streaming_batch_size = 1_000 silently gets init_size = 1_000, which can really hurt KMeans++ seed quality and would be hard to discover. Two options that both seem fine to me:

- Drop the clamp and let the user own their memory budget (matches single-GPU).
- Keep the clamp but RAFT_LOG_WARN when it triggers, and call it out in the public doc.

What do you think?

cpp/src/cluster/detail/kmeans_mg_batched.cuh
```diff
@@ -0,0 +1,522 @@
...
+
+  raft::resource::get_nccl_comms(clique);
+  int num_ranks   = raft::resource::get_num_ranks(clique);
+  IdxT n_samples  = X.extent(0);
+  IdxT n_features = X.extent(1);
+
+  IdxT base = n_samples / num_ranks;
+  IdxT rem  = n_samples % num_ranks;
+
+  cuvs::core::omp::check_threads(num_ranks);
```
dantegd, line 538: check_threads just logs a warning if OMP_NUM_THREADS < num_ranks; it doesn't actually guarantee the parallel region gets all num_ranks threads, no?

 If the runtime returns fewer (nested OMP off, OMP_DYNAMIC=true, thread-pool exhaustion under weird env), some ranks never enter the region and the ones that do I think will end up in hang on the first ncclAllReduce waiting for the missing ranks. That's a really nasty failure mode for users to debug. Do you think adding a omp_get_num_threads() == num_ranks check at the top of the parallel region (done by one thread, then #pragma omp barrier) and throw if it doesn't match? A clean throw is so much friendlier than a deadlock. Of course I didn't test this, so maybe this is a non issue for something I haven't thought about.

cpp/src/cluster/detail/kmeans_mg_batched_comms.cuh
```diff
@@ -0,0 +1,104 @@
...
+      use_nccl_(use_nccl),
+      nccl_comm_(nccl_comm),
+      stream_(raft::resource::get_cuda_stream(dev_res_))
+  {
+  }
+
+  cudaStream_t stream() const { return stream_; }
+
+  template <typename T>
+  void allreduce(T* sendbuf, T* recvbuf, std::size_t count) const
```
dantegd: allreduce is hard-coded to ncclSum. Today that's exactly what we need, but if we ever want MAX (e.g. an OR-style convergence flag, see my other comment about the done-flag allreduce being defensive), we'd need to extend the wrapper. WDYT about taking an op parameter that defaults to SUM? Cheap to do now, awkward to retrofit later.

cpp/src/cluster/detail/kmeans_mg_batched_comms.cuh
```diff
@@ -0,0 +1,104 @@
...
+
+#include <nccl.h>
+
+#include <cstddef>
+#include <cstdint>
+
+namespace cuvs::cluster::kmeans::mg::detail {
+
+template <typename T>
+ncclDataType_t nccl_dtype();
```
dantegd: The nccl_dtype<> table only specializes float, double, int64_t. If someone later instantiates with IndexT = int (which legacy MG kmeans already supports), the primary template fails at instantiation time with a not-very-helpful error. Could we either add an int (to ncclInt32) specialization or replace the primary template with a static_assert ladder that fails with a useful message? Bonus: that might get rid of the int64_t workaround you noted for the convergence flag.

cpp/src/cluster/kmeans_fit_float.cu
```diff
@@ -71,8 +75,19 @@ void fit(raft::resources const& handle,
          raft::host_scalar_view<float> inertia,
          raft::host_scalar_view<int64_t> n_iter)
 {
-  cuvs::cluster::kmeans::fit<float, int64_t>(
-    handle, params, X, sample_weight, centroids, inertia, n_iter);
+#ifdef CUVS_BUILD_MG_ALGOS
```
dantegd, line 81: Small thing: when CUVS_BUILD_MG_ALGOS is undefined and someone passes an SNMG handle, we silently fall through to cuvs::cluster::kmeans::detail::fit and do single-GPU. I think a RAFT_LOG_WARN("MG handle detected but cuVS built without MG support; falling back to single-GPU.") could save someone a debugging session if they built the wrong thing.

cpp/src/cluster/detail/kmeans_mg_batched.cuh
```diff
@@ -0,0 +1,522 @@
...
+  RAFT_EXPECTS(num_ranks > 0, "num_ranks must be positive");
+
+  RAFT_LOG_DEBUG("MNMG KMeans fit: rank=%d/%d, n_local=%zu, n_features=%zu, n_clusters=%d",
+                 rank,
+                 num_ranks,
+                 static_cast<size_t>(n_local),
+                 static_cast<size_t>(n_features),
+                 static_cast<int>(n_clusters));
+
+  IdxT streaming_batch_size = static_cast<IdxT>(params.streaming_batch_size);
```
dantegd: Quick question on semantics: this code per-rank-clamps streaming_batch_size to n_local, so a user moving the same streaming_batch_size from single-GPU to MG effectively gets per-rank batches that, summed, are num_ranks × the single-GPU batch. Can we document in the public header that streaming_batch_size is per-rank in MG mode?

cpp/src/cluster/detail/kmeans_mg_batched.cuh
```diff
@@ -0,0 +1,522 @@
...
+      raft::linalg::map_offset(
+        dev_res,
+        raft::make_device_vector_view<int64_t, int>(d_done_flag.data_handle(), 1),
+        [=] __device__(int) {
+          cuvs::cluster::kmeans::detail::check_convergence(
+            d_cost_view, d_prior_view, d_norm_view, tol, iter, d_done_view);
+          return *d_done_view.data_handle();
+        });
+
+      comms.allreduce(d_done_flag.data_handle(), d_done_flag.data_handle(), 1);
```
dantegd: I think this allreduce might be unnecessary? After the centroid_sums / weight_per_cluster allreduce, all ranks have bit-identical inputs to finalize_centroids and compute_centroid_shift, and both of those are deterministic (matrix_vector_op + gather_if and CUB's mapThenSumReduce over a fixed-length array). So every rank should be computing the same done_flag independently. If I'm reading it right, you could drop this collective (saves one per iteration) and rely on the determinism you already have. If we want to keep it as a safety belt against future changes, could we update the comment from the FP-nondeterminism framing to something like "defensive in case future changes to finalize_centroids introduce atomics-based reductions"? The current phrasing makes it sound load-bearing, which sent me down a rabbit hole if I'm not wrong in this :P

cpp/src/cluster/detail/kmeans_mg_batched_init.cuh
```diff
@@ -0,0 +1,249 @@
...
+    if (rank == root) {
+      raft::copy(centroids.data_handle(),
+                 initial_centroids.data_handle(),
+                 static_cast<size_t>(n_clusters) * n_features,
+                 stream);
+    }
+    return;
+  }
+
+  auto rank_counts = get_rank_sample_counts<IndexT>(handle, n_local, num_ranks, comms);
```
dantegd: Small perf opportunity: init_centroids_for_mg_batched is called inside the n_init loop in mnmg_fit, but rank_counts and global_n are invariant across n_init iterations. With n_init = 10 we do 10 unnecessary allgathers. Worth pulling it out of the seed loop and passing in?

cpp/src/cluster/detail/kmeans_mg_batched.cuh
```diff
@@ -0,0 +1,522 @@
...
+
+      raft::copy(dev_res,
+                 raft::make_pinned_scalar_view(h_done_flag.data_handle()),
+                 raft::make_device_scalar_view<const int64_t>(d_done_flag.data_handle()));
+    }
+
+    // Recompute inertia against the converged centroids
```
dantegd, line 436: A couple of things make this recompute more expensive than it needs to be:

1. The h_norm_cache we built up in the Lloyd loop has exactly the L2 norms we need here, but the new cluster_cost device-scalar overload re-norms from scratch internally and re-allocates workspace + x_norms fresh on every batch call (see cpp/src/cluster/kmeans.cuh:394-422). For the sort of datasets that need this MG path, this could be quite significant
2. We could probably skip this entirely by doing one extra process_batch pass against the final centroids using the existing buffers + cache, then allreduce. The Lloyd-loop's clustering_cost lags by one iteration, which is the only reason we need the recompute at all.

Not blocking, but worth a follow-up. Maybe a recompute_cost_into_with_norms(...) helper in kmeans_common.cuh if useful.

cpp/src/cluster/detail/kmeans_mg_batched.cuh
```diff
@@ -0,0 +1,522 @@
...
+ * (one rank per process with RAFT comms). The active backend is selected via
+ * `raft::resource::is_multi_gpu(handle)`. Each rank streams its local shard
+ * through Lloyd iterations using batched device-side reductions, allreducing
+ * partial centroid sums, weights, and clustering cost at the end of each
+ * iteration. Best-of-`n_init` is tracked per rank. RAFT comms ranks each write
+ * their own caller-provided outputs; SNMG OMP threads share caller-provided
+ * outputs, so only rank 0 writes them.
+ *
+ * @tparam T    Data / weight type (float or double)
+ * @tparam IdxT Index type (int or int64_t)
```
dantegd: Should we use IdxT or IndexT for the template parameter?

cpp/tests/cluster/kmeans_mg_batched.cu
```diff
@@ -0,0 +1,383 @@
...
+  {1000, 32, 5, 0.0001f, kmeans_weight_mode::mild_nonuniform, 1000, 1},
+  {10000, 16, 10, 0.0001f, kmeans_weight_mode::mild_nonuniform, 2000, 1},
+  // Extreme non-uniform weights (100:1 ratio): stresses weight normalization
+  {1000, 32, 5, 0.0001f, kmeans_weight_mode::extreme_nonuniform, 1000, 1},
+  // Extreme batch size = 1: single-element work buffers, many batch iterations
+  {100, 8, 3, 0.001f, kmeans_weight_mode::none, 1, 1},
+  // Very small dataset: some ranks may get only 2-3 rows with 4+ GPUs
+  {10, 4, 3, 0.001f, kmeans_weight_mode::none, 10, 1},
+  // Fewer rows than GPUs: exercises empty-rank (has_data=false) partitions on 4+ GPU systems
+  {3, 4, 2, 0.001f, kmeans_weight_mode::none, 3, 1},
```
dantegd, line 336: This case is great for testing the empty-rank path, but ARI on 3 samples and 2 clusters across 4+ GPUs is going to be non deterministic (several ranks get 0 rows, one or two ranks get 1 row each). Today the assertion ASSERT_GE(ari_vs_ref_, 0.94) happens to pass because the algorithm is well-behaved here, but I'd be a little nervous about flakiness on different runners or systems. Would it be worth either special-casing this row in checkResult to assert "doesn't crash + finite inertia" instead of an ARI bound, or bumping the row count to something where ARI is meaningful (say 64 rows, 2 clusters)?

dantegd (changes requested):
Mostly small comments and suggestions from my part

cpp/src/cluster/detail/kmeans_mg_batched.cuh
viclafargue: We already do the reduction because it is required for weight normalization. The previous implementation was following the single-GPU device-weight path, where the reduced sum stays on device to avoid a device-to-host copy and stream sync.

That said, this setup happens outside the main iteration hot path, so avoiding that scalar sync is not worth it. I added an explicit `global_wt > 0` check before normalization.

cpp/src/cluster/detail/kmeans_mg_batched.cuh
viclafargue: Good catch. I agree `max_iter + 1` should not leak into `n_iter`. I also applied the same reporting fix to single-GPU KMeans for consistency.

cpp/src/cluster/detail/kmeans_mg_batched_init.cuh
viclafargue: Agreed. I dropped the `streaming_batch_size` clamp so explicit `params.init_size` now matches the single-GPU host behavior and is only bounded by `n_samples`. That keeps `streaming_batch_size` as a batching/memory-transfer parameter instead of silently changing KMeans++ seeding quality.

cpp/src/cluster/detail/kmeans_mg_batched.cuh
viclafargue: You’re right. I added a guard at the start of the parallel region: one thread records `omp_get_num_threads()`, all threads barrier, and the NCCL path is only entered if the actual team size matches `num_ranks`. If it doesn’t, we throw after leaving the OpenMP region, so users get a clean error instead of a likely NCCL hang.

cpp/src/cluster/detail/kmeans_mg_batched.cuh
viclafargue: You’re right. After the sums/weights/cost reductions, every rank has identical inputs and the current finalize/shift path is deterministic, so the flag `allreduce` is not necessary. I dropped that comm operation and updated the comments to reflect that convergence is evaluated locally on each rank.

cpp/src/cluster/detail/kmeans_mg_batched.cuh
viclafargue: I switched to `IndexT` for conformity.

cpp/src/cluster/detail/kmeans_mg_batched_comms.cuh
lowener, line 1: This file can live independantly from the cluster namespace.
Can it be moved to src/core/ or src/utils/, to promote re-use of it in other parts of the codebase when necessary?

viclafargue: Done.

cpp/src/cluster/detail/kmeans_mg_batched.cuh
viclafargue: @tarang-jain would need this PR merged ASAP, but we can make it a follow-up.

lowener (approved):
LGTM

cpp/src/cluster/detail/kmeans_mg_batched.cuh
dantegd: Sounds good by me, just create a gh issue to track

dantegd (approved):

Issue #1464, opened 2025-10-25
[FEA] support filtering in c/downstream APIs


jeremywgleeson:
**Is your feature request related to a problem? Please describe.**
I would like to perform filtered searches on IVF_PQ and CAGRA indexes from Rust. Currently the Rust API does not support this, but the underlying C++ implementation supports bitset/bitmap filters.

**Describe alternatives you've considered**
Directly using C++ apis via FFI (not ideal).

**Additional context**
I'd be happy to implement this!

PR #2019, opened 2026-04-13
feat(rust): add search_with_filter to CAGRA Index

jamie8johnson:
## Summary

Add `Index::search_with_filter()` to the CAGRA Rust bindings that accepts a bitset filter via DLPack `ManagedTensor`. The C API `cuvsCagraSearch()` already supports `cuvsFilter` with `BITSET` type, but the Rust bindings hardcoded `NO_FILTER`. This exposes the existing capability to Rust consumers.

## Details

The bitset is a 1-D `uint32` device tensor with `ceil(n_rows / 32)` elements. Bit = 1 includes the row, bit = 0 excludes it. Filtering happens during CAGRA graph traversal, not post-retrieval, giving better recall than over-fetch-and-filter approaches.

The existing `search()` method is unchanged (backward compatible). `search_with_filter()` is additive.

## Motivation

We use CAGRA via the Rust crate for a code search tool. Currently we post-filter search results by metadata (chunk type, language), which requires 3x over-fetching to compensate for filtered-out candidates. Native bitset filtering would eliminate the over-fetch, reduce GPU work, and improve recall for filtered queries.

## Test

`test_cagra_search_with_filter`: builds a 256-point index, creates a bitset that includes only even-indexed rows, searches with the filter, and verifies all returned neighbors are even-indexed.

## Changes

- `rust/cuvs/src/cagra/index.rs`: Added `search_with_filter()` method + test

### DIFF
--- a/rust/cuvs/src/cagra/index.rs
+++ b/rust/cuvs/src/cagra/index.rs
@@ -97,6 +97,48 @@ impl Index {
         }
     }
 
+    /// Perform a filtered Approximate Nearest Neighbors search on the Index
+    ///
+    /// Like [`search`](Self::search), but accepts a bitset filter to exclude
+    /// vectors during graph traversal. Filtered vectors are never visited,
+    /// giving better recall than post-filtering.
+    ///
+    /// # Arguments
+    ///
+    /// * `res` - Resources to use
+    /// * `params` - Parameters to use in searching the index
+    /// * `queries` - A matrix in device memory to query for
+    /// * `neighbors` - Matrix in device memory that receives the indices of the nearest neighbors
+    /// * `distances` - Matrix in device memory that receives the distances of the nearest neighbors
+    /// * `bitset` - A 1-D `uint32` device tensor with `ceil(n_rows / 32)` elements.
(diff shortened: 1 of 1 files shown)


jamie8johnson:
Related to #1464 — this exposes the existing C-level CAGRA bitset filter to the Rust bindings. The C API already supports it via cuvsFilter (added in #452); this PR just stops hardcoding NO_FILTER in the Rust wrapper.

jamie8johnson:
Labels needed: improvement + non-breaking (external contributor, cannot self-label).

jamie8johnson:
@aamijar thanks for re-triggering after the E2 — CI came back green across all 54 checks on `006aa84`. The diff is small (+112/-0, single file) and additive, just exposing the existing C-level cuvsFilter to the Rust bindings (the C API has supported this since #452).

Anything else needed from my end, or is this in the review queue?

yan-zaretskiy:
Looks good to me!

yan-zaretskiy (approved):

cjnolet (approved):

PR #2020, opened 2026-04-14
Fix MG kmeans intertia_check n_iters

aamijar:
We should only check the intertia_check convergence on n_iters > 1 (on the second iteration). Otherwise we don't have a previous iteration to compare against. The SG path is already like this.

MG n_iters > 0 (bug)
SG  n_iters > 1


tarang-jain (approved):
Thanks for fixing this.

PR #2021, opened 2026-04-14
Refactor JIT+LTO kernels

KyleFromNVIDIA:
Share the filter code between IVF-flat and IVF-PQ. Make IVF-flat more granular to further cut down on the number of fragment dimensions and thus binary size. Clean up IVF-flat to make naming and template parameter ordering more consistent.


KyleFromNVIDIA:
Benchmarks for IVF-flat on `main`:

search-deep-image-96-inner-k10-batch_size10000

This PR:

search-deep-image-96-inner-k10-batch_size10000

Most data points have no appreciable difference, with the exception of a few with a very high recall.

KyleFromNVIDIA:
Both `main` and this PR JITlink the kernel in about 1.5s on the same machine, no appreciable difference there.

cpp/src/neighbors/detail/jit_lto_kernels/filter_data.cuh
```diff
@@ -0,0 +1,28 @@
...
+#pragma once
+
+#include <cstddef>
+#include <cstdint>
+
+namespace cuvs::neighbors::detail {
+
+// Structure to hold bitset filter data
+// This is passed as void* to the extern sample_filter function
+// Used by both CAGRA and IVF Flat
```
divyegala:
```suggestion
// Used by both IVF PQ and IVF Flat
```

cpp/src/neighbors/detail/jit_lto_kernels/filter_impl.cuh
```diff
@@ -0,0 +1,80 @@
...
+namespace cuvs::neighbors::detail {
+
+template <typename BitsetT, typename IndexT>
+__device__ bool sample_filter_none(uint32_t query_id, IndexT node_id, void* filter_data)
+{
+  return true;
+}
+
+// Inline implementation of bitset_view::test() to avoid including bitset.cuh
+// which transitively includes Thrust
```
divyegala: We can skip this now, not only did thrust/cub fix the issue of injecting `EmptyKernel` but also we are not picking our kernel up by name from the final linked cubin. So you can directly include the header.

cpp/src/neighbors/detail/jit_lto_kernels/filter_data.cuh
KyleFromNVIDIA: Irrelevant as I have restored the old implementation and thus deleted this file

divyegala (approved):

KyleFromNVIDIA:
OK, I've re-run the benchmarks with the latest implementation. Benchmarks from `main`:

search-deep-image-96-inner-k10-batch_size10000

Benchmarks from this branch:

search-deep-image-96-inner-k10-batch_size10000

While trying to compute the variance, I discovered that, for reasons that aren't clear to me, this branch is getting more benchmark datapoints than `main`. In particular, the highest-recall datapoint for this branch (which has a latency of 32000ms) is not present at all on `main`. I've revised my variance script to account for the discrepancy and do an apples-to-apples comparison. The variance of latency between all the datapoints that `main` and this branch have in common is 0.0005390499983075876s, which is on par for random noise.

PR #2022, opened 2026-04-14
[REVIEW] Rewrite cuvs-sys build to discover pre-installed cuVS via cmake-package

yan-zaretskiy:
- Current build process works by manual CMake invocation, which is suboptimal. The cmake-package is an official Rust crate meant to be used in the build scripts for this purpose. The new scripts attempts the discovery in existing library path, via pip, or CMAKE_PREFIX_PATH
- Added dlpack to rust conda environment for bindgen
- Added pre-generated bindings (src/bindings.rs) plus an optional "generate-bindings" cargo feature for opt-in regeneration via bindgen. To support this build scenario, also added the generate-bindings.sh script for CI staleness verification. This would allow us to build cuvs on machines that have no CUDA or cmake, e.g. docs.rs
- Moved cuvs-sys dependency to [workspace.dependencies] to deduplicate version
- Updated Rust edition to 2024
- CI workflows no longer need `LIBCLANG_PATH`


rust/cuvs-sys/build.rs
```diff
@@ -3,91 +3,274 @@
...
+        ),
+    })
+}
+
+#[cfg(feature = "generate-bindings")]
+fn find_dlpack_package() -> Result<cmake_package::CMakePackage> {
+    find_package("dlpack").find().map_err(|e| match e {
+        CmakeError::CMakeNotFound | CmakeError::UnsupportedCMakeVersion => {
+            anyhow::anyhow!(
+                "CMake is not installed or too old (3.19+ required). Install CMake and try again."
```
robertmaynard, line 104: Minimum required CMake version is 3.30.4. You shouldn't hard-code the version here at all as it is subject to change.

rust/cuvs-sys/build.rs
```diff
@@ -3,91 +3,274 @@
...
+/// Locate cuVS: try standard CMake discovery first, then fall back to pip.
+fn locate_cuvs() -> Result<CuvsMetadata> {
+    let system_err = match try_find_cuvs_package(None) {
+        Ok(metadata) => return Ok(metadata),
+        Err(e) => e,
+    };
+
+    // Pip installs cmake configs under site-packages/libcuvs/lib64/cmake/cuvs/.
+    // If there's no pip cuVS, report the original system error.
+    let pip_cmake_dir = match pip_cuvs_cmake_dir() {
```
robertmaynard, line 208: All searches for non explicit paths should be opt-in. So the user should be required to ask for pip version 
instead of automatically searching for it

rust/cuvs-sys/build.rs
```diff
@@ -3,91 +3,274 @@
...
+    package_name: &str,
+    target_name: &str,
+) -> Result<cmake_package::CMakeTarget> {
+    package.target(target_name).with_context(|| {
+        format!("Found CMake package {package_name}, but target {target_name} was not exported.")
+    })
+}
+
+fn find_cuvs_package(prefix: Option<PathBuf>) -> Result<cmake_package::CMakePackage> {
+    let mut builder = find_package("cuvs").components([CUVS_COMPONENT.to_owned()]);
```
robertmaynard, line 54: We should explicitly specify the version of CUVS we want Cmake to find. Otherwise you could be finding CUVS 26.02 while using the rust bindings logic from 26.08+

robertmaynard (changes requested):

rust/cuvs-sys/src/lib.rs
```diff
@@ -1,43 +1,16 @@
...
-        unsafe {
-            let mut res: cuvsResources_t = 0;
-            assert_eq!(cuvsResourcesCreate(&mut res), cuvsError_t::CUVS_SUCCESS);
-            assert_eq!(cuvsResourcesDestroy(res), cuvsError_t::CUVS_SUCCESS);
-        }
-    }
-}
+// Bindings are pre-generated and checked in at src/bindings.rs.
+// Use `rust/scripts/generate-bindings.sh` to regenerate them.
+include!("bindings.rs");
```
benfred: Given that the bindings are checked in now - maybe we should go something like  `pub use bindings::*` instead of including via the `include!` macro ? (or just have the generate_bindings.sh script write out this entire file instead ? )

benfred (commented):
nice! 

I think that checking in the generated bindings is definitely the way to go here - since this removes the need for clang etc to build the rust bindings, which should make building the rust code easier for your users. As part of this, I'd at least consider removing the `clang` dependencies from `dependencies.yaml` though -

rust/cuvs-sys/build.rs
yan-zaretskiy: Updated `cuvs-sys/build.rs` to it doesn't hard-code a CMake version in the error message.

rust/cuvs-sys/build.rs
yan-zaretskiy: Changed this to opt-in. Python discovery only runs when the `LIBCUVS_USE_PYTHON=1` env var is set. Otherwise cuvs-sys uses the standard CMake discovery path. I also verified the Python discovery mechanism in a fresh venv with a pip-installed libcuvs wheel.

rust/cuvs-sys/build.rs
yan-zaretskiy: This has already been working as requested. I refactored the code to make this version checking more obvious when reading the code, it was buried too deep. The build script requires the underlying libcuvs_c's version to exactly match the crate version.

rust/cuvs-sys/src/lib.rs
yan-zaretskiy: Switched this to a normal module re-export in `rust/cuvs-sys/src/lib.rs`, now it uses `mod bindings; pub use bindings::*`. We don't want the script to write the `lib.rs` directly, because we can't get crate level docs this way.

rust/scripts/generate-bindings.sh
```diff
@@ -0,0 +1,146 @@
...
+    -path '*/cuvs-sys-*/out/cuvs_bindings.rs' \
+    -printf '%T@ %p\n' \
+  | sort -n \
+  | tail -n 1 \
+  | cut -d' ' -f2-
+)"
+
+if [[ -z "${generated_file}" || ! -f "${generated_file}" ]]; then
+  echo "Could not locate generated cuvs_bindings.rs in ${target_dir}/debug/build" >&2
+  exit 1
```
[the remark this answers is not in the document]
yan-zaretskiy: This library is Linux only, so these assumptions are safe.

yan-zaretskiy:
The bot made a comment to "constrain CMake discovery to the required cuVS version, not just post-validate it." This does not work (that's what I did in the beginning), because the cmake handles the version requirement as a lower bound. But we want the strict equality check, hence I had to enforce it in the build script itself.

robertmaynard:
> The bot made a comment to "constrain CMake discovery to the required cuVS version, not just post-validate it." This does not work (that's what I did in the beginning), because the cmake handles the version requirement as a lower bound. But we want the strict equality check, hence I had to enforce it in the build script itself.

That isn't correct at all. Version support is determined by how a project generates the `<project>-config-version.cmake`. In the case of cuVS the version file which clearly has constraints and will clamp to `SameMinorVersion` (  https://docs.rapids.ai/api/rapids-cmake/stable/command/rapids_export/ )

rust/cuvs-sys/src/bindings.rs
```diff
@@ -0,0 +1,3447 @@
+/* automatically generated by rust-bindgen 0.72.1 */
+
+#[repr(u32)]
+#[derive(Debug, Copy, Clone, Hash, PartialEq, Eq)]
+pub enum cudaError {
```
robertmaynard: I presume that the request to commit pre-computed bindings is the reason these exist.

The bindings to things like `cudaError` are tied to the CUDA major version ( the enum table can be extended in 13 compared to 12 ), therefore those kinds of bindings need to be generated at compile time and not pre-computed

robertmaynard:
> > The bot made a comment to "constrain CMake discovery to the required cuVS version, not just post-validate it." This does not work (that's what I did in the beginning), because the cmake handles the version requirement as a lower bound. But we want the strict equality check, hence I had to enforce it in the build script itself.
> 
> That isn't correct at all. Version support is determined by how a project generates the `<project>-config-version.cmake`. In the case of cuVS the version file which clearly has constraints and will clamp to `SameMinorVersion` ( https://docs.rapids.ai/api/rapids-cmake/stable/command/rapids_export/ )

Now this could be an issue of the rust cmake crate but this isn't a limitation of CMake proper

rust/cuvs-sys/build.rs
```diff
@@ -3,91 +3,308 @@
...
+) -> Result<()> {
+    let found = package
+        .version
+        .context("Found cuVS, but it did not report a parseable package version.")?;
+    anyhow::ensure!(
+        found == *required_version,
+        "Found cuVS {found}, but cuvs-sys requires exact version {PACKAGE_VERSION}."
+    );
+    Ok(())
+}
```
robertmaynard: See my other comment but if cuVS is found it should be a compatible version ( same Major and Minor )

rust/cuvs-sys/src/bindings.rs
```diff
@@ -0,0 +1,3447 @@
...
+    cudaErrorCapturedEvent = 907,
+    cudaErrorStreamCaptureWrongThread = 908,
+    cudaErrorTimeout = 909,
+    cudaErrorGraphExecUpdateFailure = 910,
+    cudaErrorExternalDevice = 911,
+    cudaErrorInvalidClusterSize = 912,
+    cudaErrorFunctionNotLoaded = 913,
+    cudaErrorInvalidResourceType = 914,
+    cudaErrorInvalidResourceConfiguration = 915,
+    cudaErrorStreamDetached = 917,
```
robertmaynard: Specifically `cudaErrorStreamDetached` is a CUDA 13+ only error code

rust/cuvs-sys/src/bindings.rs
```diff
@@ -0,0 +1,3447 @@
...
+pub const cudaDataType_t_CUDA_R_4F_E2M1: cudaDataType_t = 33;
+pub type cudaDataType_t = ::std::os::raw::c_uint;
+unsafe extern "C" {
+    pub fn cudaMemcpyAsync(
+        dst: *mut ::std::os::raw::c_void,
+        src: *const ::std::os::raw::c_void,
+        count: usize,
+        kind: cudaMemcpyKind,
+        stream: cudaStream_t,
+    ) -> cudaError_t;
```
robertmaynard: This is the CUDA 12 ABI signature, the signature has changed for CUDA 13 and will need a different binding signature

robertmaynard (changes requested):

rust/cuvs-sys/src/bindings.rs
yan-zaretskiy: I understand the overall point of CUDA bindings in here being undesirable. But this one specifically looks [correct](https://docs.nvidia.com/cuda/cuda-runtime-api/group__CUDART__MEMORY.html#group__CUDART__MEMORY_1g85073372f776b4c4d5f89f7124b7bf79), no? Here I just mean that these were all generated for CUDA 13.

robertmaynard: You are correct. I was thinking of cudaMemPrefetchAsync

yan-zaretskiy:
I implemented a workaround so that we can rely on CMake proper to discover the compatible cuVS version and then use the `cmake-package` crate subsequently on that specific library instance to extract the include/linker paths for bindgen. Incidentally, I also submitted a PR to `cmake-package` to fix this upstream. If it gets merged, we will be able to remove my workaround.

I also removed all CUDA symbols from the pre-generated bindings. We only use two types from CUDA headers, the `cudaStream_t` - but only as an opaque pointer, and `cudaDataType_t`, which is a uint-based C enum. We only use 4 distinct values from that enum. For now I created an ABI-compatible Rust-side newtype around `c_unit`, so that I can still call the C API functions that require `cudaDataType_t`.

robertmaynard (approved):

cjnolet (approved):

PR #2027, opened 2026-04-15
Fix cuvs_bench pytest pareto assert

aamijar:
It is possible that the `throughput.csv` and `latency.csv` files can have fewer rows than `raw.csv`. That is because when we create the pareto csv(s) we drop certain rows if the latency and recall values are "dominated" by another point. For example if the latency and recall is lower than another point, we do not include that in the pareto. So we should relax the assert.

```
        for rel_path, expectations in expected_files.items():
            file_path = temp_datasets_dir / rel_path
            assert file_path.exists(), f"Expected file {file_path} does not exist."
            assert file_path.stat().st_size > 0, (
                f"Expected file {file_path} is empty."
            )
    
            df = pd.read_csv(file_path)
    
            actual_header = list(df.columns)
            actual_rows = len(df)
    
            # breakpoint()
            assert actual_header == expectations["header"], (
                f"Wrong header produced in file f{rel_path}"
            )
>           assert actual_rows == expectations["rows"]
E           assert 1 == 2

tests/test_cli.py:442: AssertionError
```

This scenario is possible for certain hardware (gpu, cuda version) and configs. First datapoint is strictly worse than the second, so it is dropped in the pareto csv. This leads to the assertion error.
```
nprobe=1: Recall=0.1788, items_per_second=1.29669M/s, Latency=77.1218u
nprobe=5: Recall=0.3722, items_per_second=1.3073M/s,  Latency=76.4963u
```


python/cuvs_bench/cuvs_bench/tests/test_cli.py
```diff
@@ -439,7 +439,22 @@ def test_run_command_creates_results(temp_datasets_dir: Path):
         assert actual_header == expectations["header"], (
             f"Wrong header produced in file f{rel_path}"
         )
-        assert actual_rows == expectations["rows"]
+        is_frontier = rel_path.endswith(("latency.csv", "throughput.csv"))
+        if is_frontier:
+            # Frontier files may have fewer rows than the raw results
+            # because the Pareto frontier drops dominated points.
+            assert actual_rows >= 1, f"Frontier file {rel_path} is empty"
```
dantegd: Unlikely, but there _could_ eventually be a bug or something that could trigger there to be more rows than there should be, so a small improvement would be to make this assertion a bit tighter, so something like:


```suggestion
            assert 1 <= actual_rows <= expectations["rows"], (
                f"Frontier file {rel_path} has {actual_rows} row(s), "
                f"expected between 1 and {expectations['rows']}"
)
```

dantegd (approved):
Fix looks good, have two small suggestions but not blocking

Issue #1829, opened 2026-02-20
[BUG] Multi-GPU KMeans convergence degradation with default random generator


viclafargue:
with `raft::random::GeneratorType::GenPhilox` :

```
[mg::detail::fit] rank=1 iter=1 sqrdNormError=2.9123220703e+04 tol=1.0000000000e-04 inertia_check=0 done=0
[mg::detail::fit] rank=0 iter=1 sqrdNormError=2.9123222656e+04 tol=1.0000000000e-04 inertia_check=0 done=0
[mg::detail::fit] rank=1 iter=2 sqrdNormError=1.2270645833e-08 tol=1.0000000000e-04 inertia_check=0 done=1
[mg::detail::fit] rank=0 iter=2 sqrdNormError=1.2270649385e-08 tol=1.0000000000e-04 inertia_check=0 done=1
```

with `raft::random::GeneratorType::GenPC` (default) :

```
[mg::detail::fit] rank=1 iter=1 sqrdNormError=2.9187794922e+04 tol=1.0000000000e-04 inertia_check=0 done=0
[mg::detail::fit] rank=0 iter=1 sqrdNormError=2.9187775391e+04 tol=1.0000000000e-04 inertia_check=0 done=0
[mg::detail::fit] rank=0 iter=2 sqrdNormError=7.4286144227e-03 tol=1.0000000000e-04 inertia_check=0 done=0
[mg::detail::fit] rank=1 iter=2 sqrdNormError=7.4286144227e-03 tol=1.0000000000e-04 inertia_check=0 done=0
[mg::detail::fit] rank=1 iter=3 sqrdNormError=5.3626983427e-03 tol=1.0000000000e-04 inertia_check=0 done=0
[mg::detail::fit] rank=0 iter=3 sqrdNormError=5.3626988083e-03 tol=1.0000000000e-04 inertia_check=0 done=0
[mg::detail::fit] rank=1 iter=4 sqrdNormError=4.0814131498e-03 tol=1.0000000000e-04 inertia_check=0 done=0
[mg::detail::fit] rank=0 iter=4 sqrdNormError=4.0814126842e-03 tol=1.0000000000e-04 inertia_check=0 done=0
[mg::detail::fit] rank=0 iter=5 sqrdNormError=3.3664486837e-03 tol=1.0000000000e-04 inertia_check=0 done=0
[...]

PR #2029, opened 2026-04-15
 Fix RNG seeding bug in multi-GPU KMeans

vinaydes:
This fixes the issue #1829.

**Root cause**

In the multi-GPU (kmeans_mg) KMeans implementation, multiple distinct random sampling operations were all initialized with the same seed (`params.rng_state.seed`). Also all the threads were also initialized with the same seed. This caused each operation to generate identical random sequences, leading to biased or degenerate centroid initialization, particularly visible during KMeans++ and random init phases where independent draws are expected. A secondary seeding issue existed in single-GPU kmeansPlusPlus: the CPU and GPU RNGs were both seeded with the raw `params.rng_state.seed`, which is incorrect in my opinion.

User specified generator type was ignored at one place, which is also fixed by this PR

**Fix**

  - Multi-GPU (kmeans_mg.cuh): Each rank now initializes a single `std::mt19937_64` generator seeded with `params.rng_state.seed + (uint64_t(my_rank) << 32)`, ensuring per-rank uniqueness while remaining deterministic. This generator is passed through the init call chain (`initRandom`, `initKMeansPlusPlus`) and is used to derive independent seeds for each GPU RNG operation (`RngState`, `shuffle_and_gather`, `fit_main`, fallback random init), so no two phases share the same seed.
  - Single-GPU (kmeans.cuh): Replaced `std::mt19937` with `std::mt19937_64` for consistency, and derive the GPU RNG seed from it rather than reusing the raw seed directly.
  - The repeated instantiation of CPU generator is needed because we have kept the `params` argument read-only (and thus params.rng_state`). May be we should separate rng_state from `params`?

The API changes suggested in the fix are a bit clunky. Better way to do it would be welcome.


cpp/src/cluster/detail/kmeans_mg.cuh
viclafargue, line 55: Looks like `default_params` is now dead code and could be removed.

cpp/src/cluster/detail/kmeans_mg.cuh
```diff
@@ -436,10 +440,10 @@ void initKMeansPlusPlus(const raft::resources& handle,
 
     // generate `n_random_clusters` centroids
     cuvs::cluster::kmeans::params rand_params = params;
-    rand_params.rng_state                     = default_params.rng_state;
+    rand_params.rng_state.seed                     = gen_64();
```
viclafargue, line 443: Could be misleading to set this `rand_params.rng_state.seed` field here as `initRandom` does not use it anymore, but uses the `gen_64` argument instead.

cpp/src/cluster/detail/kmeans_mg.cuh
```diff
@@ -404,16 +407,17 @@ void initKMeansPlusPlus(const raft::resources& handle,
     // seed they should generate the same potentialCentroids
     auto const_centroids = raft::make_device_matrix_view<const DataT, IndexT>(
       potentialCentroids.data_handle(), potentialCentroids.extent(0), potentialCentroids.extent(1));
+    auto params_copy = params;
+    params_copy.rng_state.seed = gen_64();
     cuvs::cluster::kmeans::init_plus_plus(
-      handle, params, const_centroids, centroidsRawData, workspace);
+      handle, params_copy, const_centroids, centroidsRawData, workspace);
```
viclafargue, line 413: This is an issue. When `initKMeansPlusPlus` oversamples (`potentialCentroids.extent(0) > n_clusters`), a mini single-GPU KMeans is run locally on each rank to reduce the candidates down to `n_clusters`. Since `potentialCentroids` and `weight` are already identical across ranks (via prior `allgatherv/allreduce`), and no communication happens after this reduction, all ranks must use the same RNG seed to produce identical results. Currently, `init_plus_plus` is given a per-rank-divergent seed from `gen_64()`, which breaks this invariant.

viclafargue (changes requested):
Thanks for looking into this @vinaydes! 

Each worker should indeed be seeded differently. And it looks like the PR solves the issue. The reproducer now has roughly similar number of iterations with different random number generators (much more though on my machine).

One important, but solvable issue though, is that the recluster in `initKMeansPlusPlus` when `potentialCentroids.extent(0) > n_clusters` should absolutely be deterministic. The final centroids to be used for KMeans would otherwise diverge across ranks.

cpp/src/cluster/detail/kmeans_mg.cuh
vinaydes: We can generate a seed at rank 0 and broadcast it to other ranks? Would that be an acceptable solution? What implications would a sync here will have on load balancing?

viclafargue: > What implications would a sync here will have on load balancing?

At this step the workers just completed the allreduce operation. Broadcasting a seed should be relatively cheap to do.

> We can generate a seed at rank 0 and broadcast it to other ranks? Would that be an acceptable solution?

I think that this solution could work. That said since earlier steps already introduced some randomness, we could theoretically perform this recluster step with a constant seed.

Also it would be great if we could double check that everything works as expected before we merge the PR. Especially that the KMeans algorithm actually starts with similar centroids initialization on all workers whatever the init mode chosen and whether there is a recluster step or not.

PR #2030, opened 2026-04-16
Add UDF Usage and Developer docs

### LINKED ISSUE #1873: Document how to add new UDFs

divyegala:
Closes #1873


docs/source/udf_usage.rst
```diff
@@ -0,0 +1,71 @@
...
+
+* IVF-flat — :doc:`search <cpp_api/neighbors_ivf_flat>` (``search_params.metric_udf`` / ``CUVS_METRIC``).
+
+Requirements and tips
+-----------------------
+
+* Include ``<cuvs/neighbors/ivf_flat.hpp>`` and define a metric with ``CUVS_METRIC(MyName, { ... })``.
+  Set ``search_params.metric_udf`` to the string returned by ``MyName_udf()``.
+* Prefer the helpers documented next to the macro (``squared_diff``, ``abs_diff``, ``dot_product``,
+  ``point`` element access, and so on) so the same definition works across ``float``, ``int8_t`` /
```
dantegd, line 28: "and so on" kinda leaves the user guessing, almost like "left as an exercise to the reader".  A user writing a custom metric needs to know exactly what's available, a brief table listing all the helpers (name, signature, description) or at minimum a cross-reference to the header line numbers where they're defined would be helpful here.

docs/source/udf_usage.rst
```diff
@@ -0,0 +1,71 @@
...
+
+   #include <cuvs/neighbors/ivf_flat.hpp>
+
+   namespace ivf = cuvs::neighbors::ivf_flat;
+
+   // L∞ (Chebyshev): per dimension, acc = max(acc, |x - y|); acc starts at 0 in the scan kernel.
+   CUVS_METRIC(my_chebyshev, {
+     auto d = abs_diff(x, y);
+     acc    = (d > acc) ? d : acc;
+   })
```
dantegd, line 50: Small thing, but this differs from [the example in ivf_flat.hpp ](https://github.com/rapidsai/cuvs/blob/a626f60f7b1a264b6b9f841036842841ff44a277/cpp/include/cuvs/neighbors/ivf_flat.hpp#L3525)where we don't use the helper, would be good to make them match

docs/source/jit_lto_guide.md
```diff
@@ -701,6 +716,210 @@ void search_jit(
 } // namespace example::detail
 ```
 
+### Step 7b: Example — NVRTC UDFs for `compute_distance` and `apply_filter`
```
dantegd, line 719: This is clever, however, the explanatory text leading into the code snippets  is a really large wall of text that mixes:

- Architectural rationale (why NVRTC and not static fatbins)
- Implementation constraints (explicit instantiations, forwarding definitions)
- Usage instructions (which planner API to call)

Which makes it really hard to read and follow, particularly for someone that has never used NVRTC or familiar with the UDF infra at all. 

Consider restructuring into something like:

- What you're building (1-2 sentences)
- Architecture diagram or flow (entry kernel → forward declarations → NVRTC TU → explicit instantiations → planner registration)
- Code with inline comments
- Pitfalls/constraints as a separate callout

docs/source/jit_lto_guide.md
```diff
@@ -602,6 +604,16 @@ struct SearchPlanner : AlgorithmPlanner {
...
+
+  void add_metric_udf_fragment(std::unique_ptr<UDFFatbinFragment> fragment)
+  {
+    add_fragment(std::move(fragment));
+  }
+
+  void add_filter_udf_fragment(std::unique_ptr<UDFFatbinFragment> fragment)
+  {
+    add_fragment(std::move(fragment));
+  }
```
dantegd, line 616: These do the same and is just semantic sugar, no? Would be good to add a comment to state so for readability and documentation.

docs/source/jit_lto_guide.md
```diff
@@ -701,6 +716,210 @@ void search_jit(
...
+constexpr auto get_metric_tag() {
+  if constexpr (Metric == DistanceType::Euclidean) return tag_metric_euclidean{};
+  else if constexpr (Metric == DistanceType::MetricUdf) return tag_metric_custom_udf{};
+}
+
+template <FilterType Filter>
+constexpr auto get_filter_tag() {
+  if constexpr (Filter == FilterType::None) return tag_filter_none{};
+  else if constexpr (Filter == FilterType::FilterUdf) return tag_filter_custom_udf{};
+}
```
dantegd, line 895: I think this is similar to a comment I had in the UDF infra PR, if a new enum value is added later but these functions aren't updated, the compiler will give a confusing "no return" error rather than a clear message. Adding an else { static_assert(...); } fallback would be a good idea. Since this is documentation/example code, arguably it's even more important here to model best practices.

docs/source/jit_lto_guide.md
```diff
@@ -701,6 +716,210 @@ void search_jit(
...
+
+Call `instantiate_compute_distance_udf` / `instantiate_apply_filter_udf` to append forwarding templates and explicit instantiations, using the same `type_name<...>()` tokens the entry TU will use. Concatenate that glue onto each `*_udf()` string, then compile and register.
+
+**4. Extend `search_jit.cuh` (Step 7) for UDF vs static**
+
+Step 7 only registered static matrix fragments. To add NVRTC UDFs without breaking the Euclidean / no-filter path, add `#include <string>`, replace the single-value `DistanceType` / `FilterType` enums and the `get_metric_tag` / `get_filter_tag` templates with the extended versions below, add the empty UDF tag structs, and keep the UDF glue (`my_l2_udf`, `instantiate_*`, `type_name`, `nvrtc_compiler`) in the same translation unit. Then **replace** the two unconditional `add_compute_distance_device_function` / `add_filter_device_function` lines with the `if constexpr` planner block (second snippet).
+
+```cpp
+enum class DistanceType { Euclidean, MetricUdf };
+enum class FilterType { None, FilterUdf };
```
dantegd, line 880:
```suggestion
// Extend the DistanceType / FilterType enums from Step 7:
enum class DistanceType { Euclidean, MetricUdf };
enum class FilterType { None, FilterUdf };
```

Small comment that might help readability

dantegd (changes requested):
Nice PR overall, just had a few comments

dantegd (approved):

cjnolet:
Let's migrate this over to the new docs and not merge it as-is. The new docs change is grueling to keep fixing merge issues and having to re-migrate, whereas this is consolidated into a small set of files that should migrate fairly cleanly over to the new infra.

cpp/include/cuvs/neighbors/ivf_flat.hpp
```diff
@@ -3526,10 +3526,8 @@ inline std::string instantiate_udf(char const* data_type, char const* acc_type,
  *   })
  *
  *   CUVS_METRIC(my_chebyshev, {
- *       for (int i = 0; i < x.size(); ++i) {
- *           auto diff = (x[i] > y[i]) ? (x[i] - y[i]) : (y[i] - x[i]);
- *           if (diff > acc) acc = diff;
- *       }
+ *       auto d = abs_diff(x, y);
```
cjnolet, line 3529: Why is this change in here? Was this an accident?

divyegala: No, it was not exactly correct when the PR went in. Just fixing it with this PR as it is related to UDF usage.

fern/docs.yml
```diff
@@ -210,15 +217,10 @@ navigation:
...
-        path: "./pages/advanced_topics.md"
-        contents:
-          - page: "ABI Stability"
-            path: "./developer_guide/abi_stability.md"
-          - page: "JIT Compilation"
-            path: "./pages/jit_compilation.md"
-          - page: "Link-time Optimization"
-            path: "./pages/jit_lto_guide.md"
+      - page: "ABI Stability"
+        path: "./developer_guide/abi_stability.md"
```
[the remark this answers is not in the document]
divyegala: Finding is invalid — skipping.

Verification:

./developer_guide/abi_stability.md resolves to fern/developer_guide/abi_stability.md and exists (10 KB, 368 lines).
./pages/abi_stability.md does not exist, so the proposed change would introduce the very broken link it claims to fix.
There is a separate user-facing doc at ./pages/user_guide/abi_stability.md (6 KB), already referenced by the User Guide "Compatibility" entry. These are two intentionally distinct files — the developer-facing one (ABI internals, header layout, etc.) and the user-facing one (release compatibility windows).
The Fern dev server has been reloading cleanly with the current ./developer_guide/abi_stability.md path; no broken-links error from the build.
The path format is unusual (a sibling of pages/, not under it), but that's the actual layout of the repo, not a bug. Path strings in docs.yml are relative to the config file, not constrained to ./pages/. Renaming or relocating the source file would be a much bigger refactor with no functional benefit.

No changes made.

cjnolet (approved):

PR #2031, opened 2026-04-16
CAGRA: variable graph degree for CPU-based algorithm

achirkin:
Modify `optimize` routine of CAGRA build process to allow variable graph degree.

Introduce `variable_graph_degree_fraction` parameter (by default = 1.0 = normal / constant degree behavior). This parameter defines the minimum allowed graph degree for any graph node.


cpp/src/neighbors/detail/cagra/graph_core.cuh
```diff
@@ -1738,13 +1858,32 @@ void optimize(
...
+  const uint64_t target_pruned_degree = std::max<uint64_t>(
+    1, static_cast<uint64_t>(std::ceil(output_graph_degree * variable_graph_degree_fraction)));
+  const bool variable_graph_degree = (target_pruned_degree < output_graph_degree);
+
+  if (variable_graph_degree) {
+    RAFT_LOG_INFO("# Pruning kNN graph (size=%lu, degree=%lu, target_pruned_degree=%lu)",
+                  graph_size,
+                  knn_graph_degree,
+                  target_pruned_degree);
+  }
```
mfoerste4, line 1725: You might want to move down the legacy RAFT_LOG here as well and combine it.

cpp/src/neighbors/detail/cagra/graph_core.cuh
```diff
@@ -1791,17 +1942,52 @@ void optimize(
...
+      raft::div_const_op<double>{static_cast<double>(graph_size)}, raft::cast_op<double>{}};
+    raft::linalg::reduce<raft::Apply::ALONG_ROWS>(
+      res,
+      raft::make_device_matrix_view<const uint32_t, int64_t, raft::row_major>(
+        d_natural_degree.data_handle(), int64_t{1}, static_cast<int64_t>(graph_size)),
+      raft::make_device_vector_view<double, int64_t>(d_avg_natural.data_handle(), int64_t{1}),
+      0.0,
+      false,
+      normalize_mean);
+    raft::copy(res, raft::make_host_scalar_view(&avg_natural), d_avg_natural.view());
```
mfoerste4, line 1824: Why did you not choose map_reduce on the vector directly?

cpp/src/neighbors/detail/cagra/graph_core.cuh
```diff
@@ -314,6 +352,17 @@ __global__ void kern_fused_prune(KnnGraphView knn_graph,        // [graph_chunk_
...
+    if constexpr (VariableDegree) {
+      if (i + 1 == target_pruned_degree) {
+        // Freeze the detour level after we've placed exactly target_pruned_degree edges.
+        target_detour_level = warp_min_count;
+      } else if (i >= target_pruned_degree && warp_min_count > target_detour_level &&
+                 natural_degree == output_graph_degree) {
+        // The detour level just rose above the target band. Record the natural degree once.
+        natural_degree = i;
+      }
+    }
```
mfoerste4, line 243: So we just track the 'natural_degree' here but continue to fill up the output graph - is this required?

mfoerste4 (commented):
`graph_core.cuh` looks good to me.

cjnolet:
@achirkin can we get some benchmarks attached to this PR please demonstrating the improvements? 

It's a lot easier to point users to PRs to see benefits of changes than it is to have to look up slide decks or internal convos.

achirkin:
Here are results on couple datasets. We look at red line (CAGRA graph with default parameters) vs broun line (HNSW graph). I'm not yet happy with the performance and want to tweak the merging logic just a little bit more. Apart from this, the PR is ready for review.

search-wiki-split-10M-vardegree
search-openai-5M-vardegree

(the build time is virtually unchanged with/without variable degree option)

cpp/src/neighbors/detail/cagra/graph_core.cuh
achirkin: Yes, because we need the full graph in the other steps; we only use the natural degree later during the merging step.

achirkin:
Updated thinning logic:
image
image

cpp/src/neighbors/detail/cagra/graph_core.cuh
achirkin: I'm using the new raft::stats::mean that allows different I/O types now.

cpp/src/neighbors/detail/cagra/cagra_serialize.cuh
```diff
@@ -277,8 +277,14 @@ void serialize_to_hnswlib(
   size_t bytes_written = 0;
   float GiB            = 1 << 30;
   for (std::size_t i = 0; i < index_.size(); i++) {
-    auto graph_degree = static_cast<int>(index_.graph_degree());
-    os.write(reinterpret_cast<char*>(&graph_degree), sizeof(int));
+    int actual_degree = static_cast<int>(index_.graph_degree());
+    for (int j = 0; j < actual_degree; j++) {
+      if (host_graph(i, j) == static_cast<IdxT>(-1)) {
```
julianmi: Should this use `kInvalidNeighbor` as well?

cpp/src/neighbors/detail/cagra/cagra_serialize.cuh
```diff
@@ -277,8 +277,14 @@ void serialize_to_hnswlib(
   size_t bytes_written = 0;
   float GiB            = 1 << 30;
   for (std::size_t i = 0; i < index_.size(); i++) {
-    auto graph_degree = static_cast<int>(index_.graph_degree());
-    os.write(reinterpret_cast<char*>(&graph_degree), sizeof(int));
+    int actual_degree = static_cast<int>(index_.graph_degree());
+    for (int j = 0; j < actual_degree; j++) {
```
julianmi: Which values of `actual_degree` do we see for the two heuristics 0.35 and 0.7? I wonder if it make sense to revert this loop depending of how close the actual and fixed graph degrees are.

cpp/src/neighbors/detail/cagra/graph_core.cuh
```diff
@@ -322,20 +393,43 @@ __global__ void kern_merge_graph(
...
+    max(current_mst_graph_num_edges, min(effective_degree, output_graph_degree / 2));
 
-  if (num_protected_edges > output_graph_degree) {
-    check_num_protected_edges[0] = 0u;
+  if (num_protected_edges > effective_degree) {
+    check_num_protected_edges(0) = 0u;
     return;
   }
-  if (num_protected_edges == output_graph_degree) { return; }
+  // Variable-degree path always needs to write the -1 padding (and write-back natural_degree),
```
julianmi:
```suggestion
  // Variable-degree path always needs to write the `kInvalidNeighbor` padding (and write-back natural_degree),
```

cpp/src/neighbors/detail/cagra/cagra_serialize.cuh
julianmi: We could move `graph_row`'s definition above this inner loop and reuse it here to improve readability.

cpp/include/cuvs/neighbors/cagra.hpp
```diff
@@ -153,11 +153,37 @@ enum class hnsw_heuristic_type : uint32_t {
...
+   * threshold that covers at least ceil(graph_degree * fraction) edges per node,
+   * then lets reverse edges expand the degree further. Unused slots are filled
+   * with a sentinel value (`kInvalidNeighbor`).
+   *
+   * This is intended for the CAGRA-to-HNSW conversion pipeline: the resulting
+   * graph, when imported into hnswlib, produces variable-degree neighbor lists
+   * similar to natively-built HNSW graphs. Do not use this with CAGRA's native
+   * GPU search.
+   */
+  double variable_graph_degree_fraction = 1.0;
```
julianmi, line 190: Could we also update the `SIMILAR_SEARCH_PERFORMANCE` description above? It still says CAGRA produces a fixed-degree graph.

cpp/src/neighbors/detail/cagra/graph_core.cuh
```diff
@@ -1579,13 +1705,32 @@ void optimize(
   RAFT_EXPECTS(new_graph.extent(1) <= knn_graph.extent(1),
                "output graph cannot have more columns than input graph");
   // const uint64_t input_graph_degree  = knn_graph.extent(1);
-  const uint64_t knn_graph_degree    = knn_graph.extent(1);
-  const uint64_t output_graph_degree = new_graph.extent(1);
-  const uint64_t graph_size          = new_graph.extent(0);
+  const uint64_t knn_graph_degree     = knn_graph.extent(1);
+  const uint64_t output_graph_degree  = new_graph.extent(1);
+  const uint64_t graph_size           = new_graph.extent(0);
+  const uint64_t target_pruned_degree = std::max<uint64_t>(
```
julianmi, line 1714: I think we should check the range `(0, 1]` of `variable_graph_degree_fraction`. Not sure if this is the best place or in the public routines.

cpp/src/neighbors/detail/cagra/graph_core.cuh
```diff
@@ -345,11 +439,26 @@ __global__ void kern_merge_graph(
         lane_id, smem_sorted_output_graph + num_protected_edges, num_shift);
       if (lane_id == 0) { smem_sorted_output_graph[num_protected_edges] = rev_graph_value; }
       __syncwarp();
+      // A new rev-graph edge has been inserted at position `num_protected_edges`, growing the
+      // node's effective degree up to (but not beyond) the full output_graph_degree. In the
+      // constant-degree path effective_degree already equals output_graph_degree, so this is a
+      // no-op and needs no VariableDegree guard.
+      if (effective_degree < output_graph_degree) { effective_degree++; }
```
julianmi: Is it guaranteed that a new edge was inserted? Wouldn't we need to also check `pos >= effective_degree` to ensure that the selected position was not within the prefixed area? E.g., in case of reciprocal edges.

julianmi (commented):
Thanks, this feature looks very promising. Could you add some statistics on the resulting graph degree with the `SAME_GRAPH_FOOTPRINT` and `SIMILAR_SEARCH_PERFORMANCE` please? Please also extend the results to include `variable_graph_degree_fraction = 0.7` if this becomes the default `SIMILAR_SEARCH_PERFORMANCE` heuristic.

cpp/src/neighbors/detail/cagra/graph_core.cuh
achirkin: Thanks for spotting this! I'm re-doing the benchmarks to see how it changes the results, but so far the difference in average degree is very small.

cpp/src/neighbors/detail/cagra/graph_core.cuh
achirkin: Added the check [0,1] in graph_core so that it's checked for all consumers

achirkin:
Hi @julianmi many thanks for the thorough review! After more benchmarks, I decided to disable the variable degree feature completely for the SIMILAR_SEARCH_PERFORMANCE heuristic. In this mode, the maximum degree is already `M` and not `2*M`, and the heuristics was tailored to show good performance executed in either CAGRA or HNSW, so naturally the best results are observed at fraction = 1.0.
image

achirkin:
## Typical degree distribution
### CAGRA (fraction 0.35)
image

### HNSW
image

julianmi (approved):
LGTM. Thank you for addressing my comments and providing the data.

cpp/src/neighbors/detail/cagra/graph_core.cuh
```diff
@@ -1578,14 +1704,38 @@ void optimize(
...
+  }
 
   raft::common::nvtx::range<cuvs::common::nvtx::domain::cuvs> fun_scope(
     "cagra::graph::optimize(%zu, %zu, %u)", graph_size, knn_graph_degree, output_graph_degree);
 
+  // Per-node "natural" degree produced by the prune step. Only allocated when the variable-degree
+  // path is enabled. The prune kernel records here the count of edges with detour level <=
+  // detour level at position (target_pruned_degree - 1). The merge kernel then both reads and
+  // updates this array (after rev-graph back-fill the value becomes the final effective degree).
+  auto d_natural_degree = raft::make_device_mdarray<uint32_t>(
```
tfeher, line 1734: The [memory estimates for optimize](https://github.com/NVIDIA/cuvs/blob/399c9fedf8ffcf1e0639a6ea966b90808539baee/cpp/src/neighbors/detail/cagra/cagra_helpers.cpp#L34) should be updated to include the additional array.  Please create a tracking issue.

tfeher (changes requested):
Thanks Artem for this PR! Overall it looks good, and I am really looking forward to this feature. Do the existing test cover the new code path?

PR #2032, opened 2026-04-17
Add ACE CAGRA Graph Reordering Option

julianmi:
The CAGRA graph built by the disk-backed ACE algorithm partitions the dataset. Thus, the CAGRA graph uses the reordered index space. Building a HNSW index using `from_cagra` uses the reordered dataset and CAGRA graph. Downstream consumers building an HNSW index would therefore require the reordered dataset, which is typically large when requiring the disk-backed ACE algorithm. Thus, remapping the graph to the original index space can minimize the network transfers for downstream consumers if they have the original dataset locally available.

This PR proposes an option to remap the CAGRA graph to original index space using `remap_disk_graph_to_original_ids`. Passing the original dataset to `from_cagra` for a disk-backed index would first reorder the graph to original index space and then assemble the HNSW index from the remapped graph and the original dataset using `serialize_to_hnswlib_with_original_dataset`. See `examples/cpp/src/cagra_hnsw_ace_example.cu` using `REMAP_GRAPH_TO_ORIGINAL_IDS` for an example.

`hnsw::build()` is unchanged. Use `cagra::build` with ACE disk parameters followed by `from_cagra` passing the original dataset to use this option.

`cuvs_cagra_hnswlib` bench wrapper and JSON config gain a `use_original_id_graph` flag to benchmark these changes. The reordering is mostly I/O bound and scales roughly with the disk specifications and graph size. The overhead of remapping it is typically smaller than sending the reordered dataset over a slow network interface.

### DIFF
--- a/cpp/tests/neighbors/ann_hnsw_ace.cuh
+++ b/cpp/tests/neighbors/ann_hnsw_ace.cuh
@@ -11,8 +11,8 @@
 
 #include <raft/core/detail/mdspan_numpy_serializer.hpp>
 
-#include <cstring>
 #include <cstdio>
+#include <cstring>
 #include <filesystem>
 #include <fstream>
 
@@ -133,7 +133,8 @@ class AnnHnswAceTest : public ::testing::TestWithParam<AnnHnswAceInputs> {
     ASSERT_TRUE(std::filesystem::exists(original_graph_path));
 
     auto mapping = load_npy_1d(mapping_path);
-    auto [reordered_dataset, reordered_dataset_shape] = load_npy_2d_typed<DataT>(reordered_data_path);
+    auto [reordered_dataset, reordered_dataset_shape] =
+      load_npy_2d_typed<DataT>(reordered_data_path);
     auto [reordered_graph, reordered_shape] = load_npy_2d(reordered_graph_path);
     auto [original_graph, original_shape]   = load_npy_2d(original_graph_path);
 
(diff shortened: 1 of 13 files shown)


cpp/tests/neighbors/ann_hnsw_ace.cuh
```diff
@@ -7,9 +7,14 @@
 #include "ann_cagra.cuh"
 
 #include <cuvs/neighbors/hnsw.hpp>
+#include <cuvs/util/file_io.hpp>
+
+#include <raft/core/detail/mdspan_numpy_serializer.hpp>
```
cjnolet, line 12: Definitely should not be using detail APIs from RAFT. 

If this is really needed in cuVS, it should be properly exposed through a public API.

cjnolet (changes requested):
We are planning to start using our new "Dataset" abstraction to represent datasets, and will be deprecating the mdspan-based datasets. Let's please wait until that lands so we can adjust accordingly. 

For reordering, we will want to make a new dataset and update that accordingly.

cpp/tests/neighbors/ann_hnsw_ace.cuh
julianmi: Thanks, this makes sense to me. We use `parse_descr`, `read_header`, `get_numpy_dtype`, `write_header` from `raft::detail::numpy_serializer` throughout the codebase. These would be good candidates to expose.

cjnolet: Yeah, so my point is that we can't be using internal APIs. They will need to be exposed before they PR can be merged. But it's also important that public apis are also generally reusable and we aren't just exposing them to avoid exposing internal APIs.

examples/cpp/src/hnsw_ace_reorder_example.cu
```diff
@@ -0,0 +1,235 @@
...
+  // Disk-mode ACE writes intermediate artifacts (reordered dataset, dataset
+  // mapping, ACE-internal graph) under build_dir. `build_to_disk` reuses
+  // this same directory to land the final layered HNSW artifacts, so the
+  // whole pipeline produces a single self-describing output directory.
+  ace_params.use_disk            = true;
+  ace_params.build_dir           = build_dir;
+  hnsw_params.graph_build_params = ace_params;
+
+  std::cout << "  hnsw_build_to_disk: writing layered HNSW index to " << build_dir << std::endl;
+  hnsw::build_to_disk(dev_resources, hnsw_params, dataset);
```
tfeher, line 114: So this calls `cagra::build`, then HNSW conversion. Is it saving in HNSW format? Or it is individual files on the disk?

PR #2035, opened 2026-04-22
Multi partition cagra search

jamxia155:
This PR adds a CAGRA search API and associated implementation for batching workloads involving a single query and multiple indexes (e.g. index segments built from one dataset by cuvs-lucene). This promotes GPU utilization for non-batched queries by parallelizing work over multiple CTAs, while minimizing the number of concurrent kernels for potential host-side parallelism.

Addresses cuvs-lucene [issue 124](https://github.com/NVIDIA/cuvs-lucene/issues/124), [issue 158](https://github.com/NVIDIA/cuvs-lucene/issues/158), [issue 159](https://github.com/NVIDIA/cuvs-lucene/issues/159), and cuvs [issue 2166](https://github.com/NVIDIA/cuvs/issues/2166).

Associated cuvs-lucene PR: https://github.com/NVIDIA/cuvs-lucene/pull/133.

### DIFF
--- a/c/include/cuvs/core/c_api.h
+++ b/c/include/cuvs/core/c_api.h
@@ -159,6 +159,23 @@ CUVS_EXPORT cuvsError_t cuvsStreamSync(cuvsResources_t res);
 CUVS_EXPORT cuvsError_t cuvsDeviceIdGet(cuvsResources_t res, int* device_id);
 
 /**
+ * @brief Configure the temporary workspace on this resources object as an uncapped pool, backed
+ *        by the current device memory resource. After the initial reservation is allocated on
+ *        first use, subsequent calls to cuvsRMMAlloc / cuvsRMMFree on the same resources handle
+ *        hit the pool cache rather than calling cudaMallocAsync / cudaFreeAsync, reducing CUDA
+ *        context lock contention under concurrent query threads. The pool grows without shrinking:
+ *        freed allocations are returned to the pool rather than to the device, so the pool's
+ *        high-water mark only increases until the resources object is destroyed.
+ *
+ * @param[in] res                cuvsResources_t opaque C handle
+ * @param[in] initial_size_bytes initial pool reservation in bytes; size to cover the
+ *                               steady-state working set to avoid growth after warmup
+ * @return cuvsError_t
+ */
+CUVS_EXPORT cuvsError_t cuvsResourcesSetWorkspacePool(cuvsResources_t res,
+                                                      size_t initial_size_bytes);
+
+/**
  * @brief Create an Initialized opaque C handle for C++ type `raft::device_resources_snmg`
  *        for multi-GPU operations
  *
@@ -239,6 +256,19 @@ CUVS_EXPORT cuvsError_t cuvsRMMFree(cuvsResources_t res, void* ptr, size_t bytes
 CUVS_EXPORT cuvsError_t cuvsRMMPoolMemoryResourceEnable(int initial_pool_size_percent,
                                             int max_pool_size_percent,
                                             bool managed);
+/**
+ * @brief Switches the working memory resource to use stream-ordered asynchronous allocation
+ * (cudaMallocAsync / cudaFreeAsync). Unlike the pool resource, this resource returns memory to
+ * the stream immediately without blocking the CPU, eliminating device-wide synchronization on
+ * deallocation. This is especially beneficial when multiple CAGRA searches run concurrently on
+ * separate CUDA streams, because the internal workspace allocations no longer serialize kernel
+ * launches. Be aware that this function will change the memory resource for the whole process
+ * and the new memory resource will be used until explicitly changed.
+ *
+ * @return cuvsError_t
+ */
+CUVS_EXPORT cuvsError_t cuvsRMMAsyncMemoryResourceEnable();
+
 /**
  * @brief Resets the memory resource to use the default memory resource (cuda_memory_resource)
  * @return cuvsError_t
--- a/c/include/cuvs/neighbors/cagra.h
+++ b/c/include/cuvs/neighbors/cagra.h
@@ -736,6 +736,49 @@ CUVS_EXPORT cuvsError_t cuvsCagraSearch(cuvsResources_t res,
                             cuvsFilter filter);
 
 /**
+ * @brief Search multiple CAGRA index partitions concurrently and return the global top-k per
+ * query.
+ *
+ * For each query row, the function searches all partitions in parallel into an internal
+ * intermediate buffer, applies per-partition distance post-processing, runs a batched top-k
+ * merge across partitions, and writes the final outputs to the caller-supplied device tensors.
+ * All work is submitted to the CUDA stream associated with @p res; use @c cuvsStreamSync to
+ * wait for completion.
+ *
+ * The index element type may be float32, float16, int8, or uint8. All partitions must share the
+ * same element type, and the queries must use that same type.
+ *
+ * @param[in]  res            cuvsResources_t opaque C handle
+ * @param[in]  params         search parameters (shared across partitions)
+ * @param[in]  num_partitions number of index partitions
+ * @param[in]  indices        array of num_partitions cuvsCagraIndex_t pointers, all of the same
+ *                            element type
+ * @param[in]  queries        DLManagedTensor* (device, same dtype as the indices, [n_queries,
+ *                            dim]); the queries matrix is searched against every partition
+ * @param[out] partition_ids  DLManagedTensor* (device, uint32, [n_queries, k]); which partition
+ *                            each returned neighbor came from
+ * @param[out] neighbors      DLManagedTensor* (device, uint32 or int64, [n_queries, k]); ordinal
+ *                            in the corresponding partition's dataset
+ * @param[out] distances      DLManagedTensor* (device, float32, [n_queries, k]); post-processed
+ *                            distance for each (query, neighbor)
+ * @param[in]  filters        array of `num_partitions` filters, one per partition (or NULL for a
+ *                            fully unfiltered search). `filters[i]` applies to partition `i`: use
+ *                            {.type=NO_FILTER, .addr=0} for no filter on that partition, or
+ *                            {.type=BITSET, .addr=ptr} where ptr is a uintptr_t-cast
+ *                            DLManagedTensor* holding that partition's own bitset (one bit per
+ *                            vector in that partition; standard 32-bit packing).
+ */
+CUVS_EXPORT cuvsError_t cuvsCagraSearchMultiPartition(cuvsResources_t res,
+                                                      cuvsCagraSearchParams_t params,
+                                                      uint32_t num_partitions,
+                                                      cuvsCagraIndex_t* indices,
+                                                      DLManagedTensor* queries,
+                                                      DLManagedTensor* partition_ids,
+                                                      DLManagedTensor* neighbors,
+                                                      DLManagedTensor* distances,
+                                                      cuvsFilter* filters);
+
+/**
  * @}
  */
 
--- a/c/include/cuvs/selection/select_k.h
+++ b/c/include/cuvs/selection/select_k.h
@@ -1,0 +1,37 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2025-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+#pragma once
+
+#include <cuvs/core/c_api.h>
+#include <dlpack/dlpack.h>
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+/**
+ * @brief Select the k smallest values from a flat device array of n candidates.
+ *
+ * Treats `in_val` as a matrix of shape [1, n] and selects the `k` smallest
+ * float values.  `out_idx` receives the int64 column positions of the selected
+ * values in [0, n), so the caller can recover per-segment identity as:
+ *
+ *   segment_index        = out_idx[j] / segment_k
+ *   position_in_segment  = out_idx[j] % segment_k
+ *
+ * @param[in]  res      cuvsResources_t handle
+ * @param[in]  in_val   DLManagedTensor* shape [1, n], float32, device memory
+ * @param[out] out_val  DLManagedTensor* shape [1, k], float32, device memory
+ * @param[out] out_idx  DLManagedTensor* shape [1, k], int64,   device memory
+ * @return cuvsError_t
+ */
+CUVS_EXPORT cuvsError_t cuvsSelectK(cuvsResources_t res,
+                                    DLManagedTensor* in_val,
+                                    DLManagedTensor* out_val,
+                                    DLManagedTensor* out_idx);
+
+#ifdef __cplusplus
+}
+#endif
--- a/c/src/neighbors/cagra.cpp
+++ b/c/src/neighbors/cagra.cpp
@@ -263,6 +263,96 @@ void _search(cuvsResources_t res,
   }
 }
 
+template <typename T, typename OutIdxT>
+void _search_multi_partition(cuvsResources_t res,
+                             cuvsCagraSearchParams params,
+                             uint32_t num_partitions,
+                             cuvsCagraIndex_t* indices,
+                             DLManagedTensor* queries,
+                             DLManagedTensor* partition_ids,
+                             DLManagedTensor* neighbors,
+                             DLManagedTensor* distances,
+                             cuvsFilter* filters)
+{
+  using IdxT      = uint32_t;
+  using DistanceT = float;
+  using IndexT    = cuvs::neighbors::cagra::index<T, IdxT>;
+
+  auto res_ptr       = reinterpret_cast<raft::resources*>(res);
+  auto search_params = cuvs::neighbors::cagra::search_params();
+  convert_c_search_params(params, &search_params);
+
+  std::vector<const IndexT*> idx_vec(num_partitions);
+  for (uint32_t i = 0; i < num_partitions; i++) {
+    idx_vec[i] = reinterpret_cast<const IndexT*>(indices[i]->addr);
+  }
+
+  using queries_view_t = raft::device_matrix_view<const T, int64_t, raft::row_major>;
+  using pid_view_t     = raft::device_matrix_view<uint32_t, int64_t, raft::row_major>;
+  using nbrs_view_t    = raft::device_matrix_view<OutIdxT, int64_t, raft::row_major>;
+  using dist_view_t    = raft::device_matrix_view<DistanceT, int64_t, raft::row_major>;
+
+  auto queries_view       = cuvs::core::from_dlpack<queries_view_t>(queries);
+  auto partition_ids_view = cuvs::core::from_dlpack<pid_view_t>(partition_ids);
+  auto neighbors_view     = cuvs::core::from_dlpack<nbrs_view_t>(neighbors);
+  auto distances_view     = cuvs::core::from_dlpack<dist_view_t>(distances);
+
+  // One bitset view per partition; an empty view (null ptr) means "no filter for that partition".
+  using bitset_view_t   = cuvs::core::bitset_view<std::uint32_t, int64_t>;
+  using bitset_mdspan_t = raft::device_vector_view<std::uint32_t, int64_t, raft::row_major>;
+  std::vector<bitset_view_t> partition_bitsets(
+    num_partitions, bitset_view_t(static_cast<std::uint32_t*>(nullptr), static_cast<int64_t>(0)));
+  if (filters != nullptr) {
+    for (uint32_t i = 0; i < num_partitions; i++) {
+      if (filters[i].type == NO_FILTER) {
+        continue;  // leave the empty (accept-all) view for this partition
+      } else if (filters[i].type == BITSET) {
+        auto* bitset_tensor = reinterpret_cast<DLManagedTensor*>(filters[i].addr);
+        RAFT_EXPECTS(
+          bitset_tensor != nullptr, "BITSET filter addr must be non-null (partition %u)", i);
+        auto bitset_mds = cuvs::core::from_dlpack<bitset_mdspan_t>(bitset_tensor);
+        partition_bitsets[i] =
+          bitset_view_t(bitset_mds, static_cast<int64_t>(bitset_mds.size()) * 32);
+      } else {
+        RAFT_FAIL("Unsupported filter type for multi-partition search (partition %u): %d",
+                  i,
+                  (int)filters[i].type);
+      }
+    }
+  }
+
+  cuvs::neighbors::cagra::search(*res_ptr,
+                                 search_params,
+                                 idx_vec,
+                                 queries_view,
+                                 partition_ids_view,
+                                 neighbors_view,
+                                 distances_view,
+                                 partition_bitsets);
+}
+
+template <typename T>
+void _search_multi_partition(cuvsResources_t res,
+                             cuvsCagraSearchParams params,
+                             uint32_t num_partitions,
+                             cuvsCagraIndex_t* indices,
+                             DLManagedTensor* queries,
+                             DLManagedTensor* partition_ids,
+                             DLManagedTensor* neighbors,
+                             DLManagedTensor* distances,
+                             cuvsFilter* filters)
+{
+  if (neighbors->dl_tensor.dtype.code == kDLUInt && neighbors->dl_tensor.dtype.bits == 32) {
+    _search_multi_partition<T, uint32_t>(
+      res, params, num_partitions, indices, queries, partition_ids, neighbors, distances, filters);
+  } else if (neighbors->dl_tensor.dtype.code == kDLInt && neighbors->dl_tensor.dtype.bits == 64) {
+    _search_multi_partition<T, int64_t>(
+      res, params, num_partitions, indices, queries, partition_ids, neighbors, distances, filters);
+  } else {
+    RAFT_FAIL("neighbors should be of type uint32_t or int64_t");
+  }
+}
+
 template <typename T>
 void _serialize(cuvsResources_t res,
                 const char* filename,
@@ -686,6 +776,54 @@ extern "C" cuvsError_t cuvsCagraSearch(cuvsResources_t res,
       RAFT_FAIL("Unsupported queries DLtensor dtype: %d and bits: %d",
                 queries.dtype.code,
                 queries.dtype.bits);
+    }
+  });
+}
+
+extern "C" cuvsError_t cuvsCagraSearchMultiPartition(cuvsResources_t res,
+                                                     cuvsCagraSearchParams_t params,
+                                                     uint32_t num_partitions,
+                                                     cuvsCagraIndex_t* indices,
+                                                     DLManagedTensor* queries,
+                                                     DLManagedTensor* partition_ids,
+                                                     DLManagedTensor* neighbors,
+                                                     DLManagedTensor* distances,
+                                                     cuvsFilter* filters)
+{
+  return cuvs::core::translate_exceptions([=] {
+    RAFT_EXPECTS(num_partitions > 0, "num_partitions must be > 0");
+    RAFT_EXPECTS(indices != nullptr && queries != nullptr && partition_ids != nullptr &&
+                   neighbors != nullptr && distances != nullptr,
+                 "All pointer arguments must be non-null");
+
+    // Every partition index must be present, built, and share one dtype; the search dispatches on
+    // that common dtype. The queries dtype is validated against T inside from_dlpack.
+    RAFT_EXPECTS(indices[0] != nullptr, "Index at position 0 is null");
+    auto index_dtype = indices[0]->dtype;
+    for (uint32_t i = 0; i < num_partitions; i++) {
+      RAFT_EXPECTS(indices[i] != nullptr && indices[i]->addr != 0,
+                   "Index at position %u is null or not built", i);
+      RAFT_EXPECTS(
+        indices[i]->dtype.code == index_dtype.code && indices[i]->dtype.bits == index_dtype.bits,
+        "All partition indices must share the same dtype; position %u differs from position 0", i);
+    }
+
+    if (index_dtype.code == kDLFloat && index_dtype.bits == 32) {
+      _search_multi_partition<float>(
+        res, *params, num_partitions, indices, queries, partition_ids, neighbors, distances, filters);
+    } else if (index_dtype.code == kDLFloat && index_dtype.bits == 16) {
+      _search_multi_partition<half>(
+        res, *params, num_partitions, indices, queries, partition_ids, neighbors, distances, filters);
+    } else if (index_dtype.code == kDLInt && index_dtype.bits == 8) {
+      _search_multi_partition<int8_t>(
+        res, *params, num_partitions, indices, queries, partition_ids, neighbors, distances, filters);
+    } else if (index_dtype.code == kDLUInt && index_dtype.bits == 8) {
+      _search_multi_partition<uint8_t>(
+        res, *params, num_partitions, indices, queries, partition_ids, neighbors, distances, filters);
+    } else {
+      RAFT_FAIL("Unsupported multi-partition index DLtensor dtype: %d and bits: %d",
+                index_dtype.code,
+                index_dtype.bits);
     }
   });
 }
--- a/c/src/selection/select_k.cpp
+++ b/c/src/selection/select_k.cpp
@@ -1,0 +1,56 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2025-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include <cuvs/core/c_api.h>
+#include "../core/exceptions.hpp"
+#include "../core/interop.hpp"
+#include <cuvs/selection/select_k.h>   // C declaration with CUVS_EXPORT (default visibility)
+#include <cuvs/selection/select_k.hpp>
+#include <dlpack/dlpack.h>
+
+#include <raft/core/device_mdspan.hpp>
+#include <raft/core/error.hpp>
+#include <raft/core/resources.hpp>
+
+extern "C" cuvsError_t cuvsSelectK(cuvsResources_t res,
+                                   DLManagedTensor* in_val,
+                                   DLManagedTensor* out_val,
+                                   DLManagedTensor* out_idx)
+{
+  return cuvs::core::translate_exceptions([=] {
+    auto* res_ptr = reinterpret_cast<raft::resources*>(res);
+
+    RAFT_EXPECTS(in_val != nullptr && out_val != nullptr && out_idx != nullptr,
+                 "Tensors passed to cuvsSelectK must not be null");
+    RAFT_EXPECTS(in_val->dl_tensor.byte_offset == 0 && out_val->dl_tensor.byte_offset == 0 &&
+                   out_idx->dl_tensor.byte_offset == 0,
+                 "Tensors passed to cuvsSelectK must have a zero byte_offset");
+
+    // from_dlpack validates ndim, dtype, device, and row-major contiguity against each mdspan type.
+    using in_view_type  = raft::device_matrix_view<const float, int64_t, raft::row_major>;
+    using val_view_type = raft::device_matrix_view<float, int64_t, raft::row_major>;
+    using idx_view_type = raft::device_matrix_view<int64_t, int64_t, raft::row_major>;
+
+    auto in_view      = cuvs::core::from_dlpack<in_view_type>(in_val);
+    auto out_val_view = cuvs::core::from_dlpack<val_view_type>(out_val);
+    auto out_idx_view = cuvs::core::from_dlpack<idx_view_type>(out_idx);
+
+    RAFT_EXPECTS(in_view.extent(0) == 1 && out_val_view.extent(0) == 1 &&
+                   out_idx_view.extent(0) == 1,
+                 "cuvsSelectK only supports a single row (batch_size == 1)");
+    RAFT_EXPECTS(out_val_view.extent(1) == out_idx_view.extent(1),
+                 "out_val and out_idx passed to cuvsSelectK must have the same length");
+    RAFT_EXPECTS(out_val_view.extent(1) <= in_view.extent(1),
+                 "k passed to cuvsSelectK must not exceed the input length");
+
+    cuvs::selection::select_k(
+      *res_ptr,
+      in_view,
+      std::nullopt,  // implicit positions [0, n) as in_idx
+      out_val_view,
+      out_idx_view,
+      true);  // select_min = true (smallest distance = nearest neighbor)
+  });
+}
--- a/cpp/include/cuvs/neighbors/cagra.hpp
+++ b/cpp/include/cuvs/neighbors/cagra.hpp
@@ -6,6 +6,7 @@
 #pragma once
 
 #include "common.hpp"
+#include <cuvs/core/bitset.hpp>
 #include <cuvs/distance/distance.hpp>
 #include <cuvs/neighbors/common.hpp>
 #include <cuvs/neighbors/ivf_pq.hpp>
@@ -31,6 +32,7 @@
 #include <optional>
 #include <string>
 #include <variant>
+#include <vector>
 
 namespace CUVS_EXPORT cuvs {
 namespace neighbors {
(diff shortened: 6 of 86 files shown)


jamxia155:
Pareto frontiers from initial benchmarking:
image
image

Hardware: 32/64-core/thread CPU, L40S 48GB
Dataset: 50M x 64 dim x 16-bit
Top-K: 100
CAGRA_HNSW (CAGRA index converted to HNSW, search on CPU) parameter sweep ranges:

- "efSearch": [100, 150, 200]
- "queryThreads": [1, 8, 64]
- "cagraGraphDegree": [16, 32, 64, 128]
- "cagraIntermediateGraphDegree": [32, 64, 128, 256, 512]
- "cagraHnswLayers": [1, 2, 4, 6]

CAGRA_SEARCH (CAGRA index used directly for search on GPU) parameter sweep ranges:
- "efSearch": [100, 150]
- "queryThreads": [1, 8, 64]
- "cagraGraphDegree": [16, 32, 64, 128]
- "cagraIntermediateGraphDegree": [32, 64, 128, 256, 512]
- "cagraSearchWidth": [4, 16, 128, 256, 1024]

c/include/cuvs/neighbors/cagra.h
```diff
@@ -712,6 +712,34 @@ cuvsError_t cuvsCagraSearch(cuvsResources_t res,
...
+ *
+ * @param[in]  res          cuvsResources_t opaque C handle
+ * @param[in]  params       search parameters
+ * @param[in]  num_segments number of index segments
+ * @param[in]  indices      array of num_segments cuvsCagraIndex_t pointers
+ * @param[in]  queries      array of num_segments DLManagedTensor* (device, float32, [nq, dim])
+ * @param[out] neighbors    array of num_segments DLManagedTensor* (device, uint32, [nq, topk])
+ * @param[out] distances    array of num_segments DLManagedTensor* (device, float32, [nq, topk])
+ */
+cuvsError_t cuvsCagraSearchMultiSegment(cuvsResources_t res,
```
cjnolet: Just to align on nomenclature a bit, I wonder if we can think of a more general name. Maybe "Partition"? Segment is pretty closely coupled to databases, and more specifically to LSM-based databases, but cuVS the library is more general that that. cuVS is at the level of "hash partitioning" or "blind sharding" (those are the terms we tend to use in this context). I think "MultiPartition" would be a more fitting name.

c/include/cuvs/selection/select_k.h
```diff
@@ -0,0 +1,37 @@
...
+ *   segment_index        = out_idx[j] / segment_k
+ *   position_in_segment  = out_idx[j] % segment_k
+ *
+ * @param[in]  res      cuvsResources_t handle
+ * @param[in]  in_val   DLManagedTensor* shape [1, n], float32, device memory
+ * @param[out] out_val  DLManagedTensor* shape [1, k], float32, device memory
+ * @param[out] out_idx  DLManagedTensor* shape [1, k], int64,   device memory
+ * @return cuvsError_t
+ */
+cuvsError_t cuvsSelectK(cuvsResources_t res,
```
cjnolet: Oh this is great. I was just working on code examples for the new docs and realized we only have a C++ API for select_k. It'll be great to get the C APis, and later on the Python and other language wrappers for select-k.

cpp/include/cuvs/neighbors/cagra.hpp
```diff
@@ -1723,6 +1724,87 @@ void search(raft::resources const& res,
...
+  const std::vector<raft::device_matrix_view<const float, int64_t, raft::row_major>>& queries,
+  const std::vector<raft::device_matrix_view<uint32_t, int64_t, raft::row_major>>& neighbors,
+  const std::vector<raft::device_matrix_view<float, int64_t, raft::row_major>>& distances);
+
+void search_multi_segment(
+  raft::resources const& res,
+  cuvs::neighbors::cagra::search_params const& params,
+  const std::vector<const cuvs::neighbors::cagra::index<float, uint32_t>*>& indices,
+  const std::vector<raft::device_matrix_view<const float, int64_t, raft::row_major>>& queries,
+  const std::vector<raft::device_matrix_view<int64_t, int64_t, raft::row_major>>& neighbors,
```
cjnolet: It seems a bit odd to return a vector of outputs instead of a single output. Can't we perform the reduction for the final single neighborhood outputs?

c/include/cuvs/neighbors/cagra.h
jamxia155: Aligning to "partition" for now. FYI, also considered: "MultiShard", "MultiIndex", "Federated", but these might come with unintended connotations.

cpp/include/cuvs/neighbors/cagra.hpp
jamxia155: Refactored this API to return the global top-k results (i.e. after `select_k`) instead of returning per-partition top-k results which then have to be reduced via `select_k`.

c/include/cuvs/selection/select_k.h
jamxia155: With the refactoring prompted by your [other comment](https://github.com/rapidsai/cuvs/pull/2035#discussion_r3215111111), select-k is no longer needed for this work. Leaving the C API intact in case it might be useful to others.

jamxia155:
Pareto frontiers from initial benchmarking with prefiltering:

image
image

Hardware: 32/64-core/thread CPU, L40S 48GB
Dataset: 50M x 64 dim x 16-bit
Top-K: 512
Prefilter reject rate: 84%

CAGRA_HNSW (CAGRA index converted to HNSW, search on CPU) parameter sweep ranges:

"efSearch": [512, 1024]
"queryThreads": [8, 16, 64]
"cagraGraphDegree": [16, 32, 64, 128]
"cagraIntermediateGraphDegree": [32, 64, 128, 256]
"cagraHnswLayers": [1, 2, 4, 6]

CAGRA_SEARCH (CAGRA index used directly for search on GPU) parameter sweep ranges:

"efSearch": [512]
"queryThreads": [8, 16, 64]
"cagraGraphDegree": [16, 32, 64, 128]
"cagraIntermediateGraphDegree": [32, 64, 128, 256]
"cagraSearchWidth": [4, 16, 128, 256]
"cagraITopK": [32, 64, 128, 256, 512]

jamxia155:
Pareto frontiers unfiltered search with top-k = 1500:

image
image


Hardware: 32/64-core/thread CPU, L40S 48GB
Dataset: 50M x 64 dim x 16-bit
Top-K: 1500

CAGRA_HNSW (CAGRA index converted to HNSW, search on CPU) parameter sweep ranges:

"efSearch": [1500, 3000]
"queryThreads": [8, 16, 64]
"cagraGraphDegree": [16, 32, 64, 128]
"cagraIntermediateGraphDegree": [32, 64, 128, 256]
"cagraHnswLayers": [1, 2, 4, 6]

CAGRA_SEARCH (CAGRA index used directly for search on GPU) parameter sweep ranges:

"efSearch": 1500
"queryThreads": [8, 16, 64]
"cagraGraphDegree": [16, 32, 64, 128]
"cagraIntermediateGraphDegree": [16, 32, 64, 256]
"cagraSearchWidth": [1, 4, 8, 16, 32]
"cagraITopK": [1500, 2000, 3000]

cpp/include/cuvs/neighbors/cagra.hpp
```diff
@@ -1723,6 +1724,115 @@ void search(raft::resources const& res,
...
+  raft::resources const& res,
+  cuvs::neighbors::cagra::search_params const& params,
+  const std::vector<const cuvs::neighbors::cagra::index<uint8_t, uint32_t>*>& indices,
+  raft::device_matrix_view<const uint8_t, int64_t, raft::row_major> queries,
+  raft::device_matrix_view<uint32_t, int64_t, raft::row_major> partition_ids,
+  raft::device_matrix_view<int64_t, int64_t, raft::row_major> neighbors,
+  raft::device_matrix_view<float, int64_t, raft::row_major> distances,
+  const cuvs::neighbors::filtering::base_filter& sample_filter =
+    cuvs::neighbors::filtering::none_sample_filter{});
+
```
achirkin: Rather than creating a new set of overloads specific for CAGRA, maybe we'd better introduce a new index type, e.g. `multi_partition_cagra`? This whole thing looks like it can be represented as a "virtual merge" index we've already introduced once.
This has couple benefits:
 - Public API stays the same homogeneous - no new search functions for a user to discover - just a new index type and its constructor. 
 - the new index search would be composable with dynamic batching  (which takes some index and calls search on it).

achirkin (dismissed):

cpp/include/cuvs/neighbors/cagra.hpp
jamxia155: Thanks, @achirkin, for the insightful comment. Assuming that `composite_index` is the "virtual merge" index you mentioned, it does look very attractive as an existing infrastructure for multi-partition search. There are two wrinkles that prevent immediate adoption:

1. `composite_index` implements the [per-index search in a for loop](https://github.com/rapidsai/cuvs/blob/main/cpp/src/neighbors/composite/index.cu#L61-L96), which does not expose partition-level parallelism needed to maximize device utilization when the batch size is 1 as is always the case in cuVS-Lucene. For the current work, we actually had to create multi-partition versions of the entire call stack down to the kernels and device functions (e.g. [search_multi_cta_mp_jit](https://github.com/jamxia155/cuvs/blob/54a55f5002705f72d0fbe1e9968ebfeaf4e7828d/cpp/src/neighbors/detail/cagra/jit_lto_kernels/search_multi_cta_jit.cuh#L372)). Now, the multi-partition code can in theory entirely subsume the original single-partition code by essentially setting `num_partition=1` as needed but I don't think we should make such a sweeping change in this PR.
2. For cuVS-Lucene, we need to produce the partition-local ordinals of the neighbors instead of the global indices. However, `composite_index` internally [consumes the ordinals to produce only the global indices](https://github.com/rapidsai/cuvs/blob/main/cpp/src/neighbors/composite/index.cu#L82-L88).

Both issues above can possibly be resolved with additional configuration parameters and corresponding code paths in `composite_index`.

On the other hand, making `multi_partition_cagra` its own header probably involves easier changes. Do you think we should make a push for extending/customizing `composite_index` at this time?

cjnolet: > Rather than creating a new set of overloads specific for CAGRA, maybe we'd better introduce a new index type, e.g. multi_partition_cagra? This whole thing looks like it can be represented as a "virtual merge" index we've already introduced once.

I don't know that I agree with this. This needs to be a phyiscal merge, not a logical merge. This is critical for LSM-based databases and it'll be widely used outside of just Lucene. I'm not saying we can't (and shouldn't) improve the API experience wherever possible, but I'm not convinced the "logical merge" (or composite index) is not the way to do that.

achirkin: Hi Corey, I totally agree with you comment. I meant to say structurally the code is very similar and I think it is a good pattern to use for this case too.

cjnolet: I like your thinking here for sure- if there's a good reusable abstraction for this that can consolidate APIs (ideally without introducing a complex class hierarchy on the cuvs impl side) then I'm all for it. I say "cuvs impl side" because abstractions like `device_resources`, while they have established a hierarchy on the user side, are just meant to be simple user convenience and we use raft::resources container on the implementation side).

jamxia155:
Pareto frontiers 84% filtered search with top-k = 1500:

image
image

Hardware: 32/64-core/thread CPU, L40S 48GB
Dataset: 50M x 64 dim x 16-bit
Top-K: 1500

CAGRA_HNSW (CAGRA index converted to HNSW, search on CPU) parameter sweep ranges:

"efSearch": [1500, 2000, 3000]
"queryThreads": [8, 16, 64]
"cagraGraphDegree": [16, 32, 64, 128]
"cagraIntermediateGraphDegree": [16, 32, 64, 128, 256]
"cagraHnswLayers": [1, 2]

CAGRA_SEARCH (CAGRA index used directly for search on GPU) parameter sweep ranges:

"efSearch": 1500
"queryThreads": [8, 16, 64]
"cagraGraphDegree": [16, 32, 64, 128]
"cagraIntermediateGraphDegree": [16, 32, 64, 128, 256]
"cagraSearchWidth": [1, 4, 8, 16, 32]
"cagraITopK": [1500, 2000, 3000]

cpp/include/cuvs/neighbors/cagra.hpp
jamxia155: Following this discussion, I'm planning to refactor this work into a `multi_partition_cagra` index type that is constructed from a vector of CAGRA indices (like `composite_index`) and provides a `search` method that implements the current optimizations (unlike `composite_index` which doesn't really implement the search at a low-level). Does that sound about right? Does it defeat the purpose somewhat given that the signature of this `search` method can't conform to the signature common to other index types as we need an output array to indicate the id of the CAGRA index that a result comes from? Thanks for your thoughts.

cjnolet: Sorry. I would not suggest this design @jamxia155. Let's keep this simple. Index shoukd also not have search methods (we use free functions in cuVS not methods on objects).

Again to reiterate- composite index is going away. We should not be using that as a guide for designs.

jamxia155: After further discussions with @cjnolet, the plan is to keep things simple and essentially rename `search_multi_partition` as `search` for consistency, while allowing for necessary parameters in the signature even if they are not common to other `search` functions.

achirkin: I'd still suggest to go one step further and make this a new index as I proposed before. If the `composite_index` goes away, this could be something new like this:
```C++
template <typename BaseIndex>
struct multi_segment_index {
  std::vector<BaseIndex> segments;
}
```
The new necessary search parameters can be passed via a new parameter struct that envelopes member index parameters.
This is essentially the same as your latest version that searches the vector of indices, but with the added benefit that it is compatible with the dynamic batching index (by wrapping the new index type in the dynamic batching index).

cjnolet: We are going to stick with what we have for right now so that we don't rush in a new abstraction. This feature is needed today direclty by a customer, so we want to enable them. 

Let's continue the abstraction conversation in the meantime. This is a fairly low-risk change because at thie moment it's almost exclusively focused on a specific Lucene optimization. My hope is that this design will translate well to other databases, but I'd like us to really think through the abstraction layer.

jamxia155: Created cuVS issue https://github.com/NVIDIA/cuvs/issues/2281 and added a TODO comment referencing it.

jamxia155:
/ok to test https://github.com/NVIDIA/cuvs/pull/2035/commits/dbcab41804b06b2e65a299bd9687dbbb9403862f

jamxia155:
/ok to test [d2a9583](https://github.com/NVIDIA/cuvs/pull/2035/commits/d2a9583b662ec98e1e0810ab58928ce5b8521e29)

jamxia155:
/ok to test https://github.com/NVIDIA/cuvs/pull/2035/commits/643bd3f487aafb923818554eecc7873128b54cdf

anaruse:
@jamxia155, this is a great extension. Multi-partition CAGRA search is a useful capability, and the implementation work here is substantial.

Currently the single-index and multi-partition search paths are exposed as separate APIs: callers pass either a single `cagra::index` or a `std::vector<const cagra::index*>`, and must choose which path to use. That also means `num_partitions == 1` through the multi-partition API follows a different execution path from the regular single-index API.

Is the plan to keep this public API shape long term, or is this intended as an interim step until a multi-partition/partitioned index abstraction is introduced? For example, an index object carrying partition metadata could let search dispatch internally based on the index representation and avoid making callers choose between APIs.

If the multi-partition API remains separate, should `num_partitions == 1` be treated explicitly? For example, it could either be rejected with a clear error, or at least warn/document that this path is not equivalent to the regular single-index search path and may have different overhead.

cpp/src/neighbors/detail/cagra/cagra_search.cuh
```diff
@@ -252,4 +256,386 @@ void search_main(raft::resources const& res,
...
+  for (uint32_t i = 0; i < num_partitions; i++) {
+    RAFT_EXPECTS(!indices[i]->dataset_fd().has_value(),
+                 "Disk-based datasets are not supported for multi-partition search");
+    max_graph_degree = std::max(max_graph_degree, indices[i]->graph().extent(1));
+    max_dataset_size = std::max(max_dataset_size, indices[i]->data().n_rows());
+  }
+
+  if (params.max_queries == 0) {
+    cudaDeviceProp deviceProp = raft::resource::get_device_properties(res);
+    params.max_queries =
```
anaruse, line 335: The regular single-index path chunks the query set by `max_queries` before invoking the search plan. The multi-partition path appears to use the full query count directly for workspace allocation and kernel launch. Is that intentional? Should `max_queries` also bound the per-launch query count here, given that some workspaces scale with `num_queries * num_partitions`?

cpp/src/neighbors/detail/cagra/cagra_search.cuh
```diff
@@ -252,4 +256,386 @@ void search_main(raft::resources const& res,
...
+  RAFT_EXPECTS(neighbors.extent(0) == static_cast<int64_t>(n_queries),
+               "neighbors and queries must have the same number of rows");
+  RAFT_EXPECTS(distances.extent(0) == static_cast<int64_t>(n_queries) &&
+                 distances.extent(1) == static_cast<int64_t>(topk),
+               "distances shape must be [n_queries, k]");
+
+  // Find the max graph_degree across all partitions (needed for the shared kernel plan).
+  int64_t max_graph_degree = 0;
+  int64_t max_dataset_size = 0;
+  for (uint32_t i = 0; i < num_partitions; i++) {
```
anaruse, line 320: Should we validate that all partitions have the same metric here? The final `select_k` direction is chosen from `indices[0]->metric()`, so mixed metrics would not be ordered correctly.

cpp/src/neighbors/detail/cagra/cagra_search.cuh
```diff
@@ -252,4 +256,386 @@ void search_main(raft::resources const& res,
...
+    RAFT_EXPECTS(!indices[i]->dataset_fd().has_value(),
+                 "Disk-based datasets are not supported for multi-partition search");
+    max_graph_degree = std::max(max_graph_degree, indices[i]->graph().extent(1));
```
anaruse: The implementation appears to support different graph degrees across partitions by sizing the plan with `max_graph_degree` and passing each partition's actual `graph_degree` to the kernel. Do we have a test covering mixed graph degrees? The current multi-partition tests seem to build all partitions with the same graph degree.

Given that shared-memory/layout sizing is driven by the maximum graph degree, mixed graph degrees may also be suboptimal for performance. If this is not a required use case, should we reject mixed graph degrees explicitly and document that all partitions should use the same degree?

jamxia155:
> Currently the single-index and multi-partition search paths are exposed as separate APIs: callers pass either a single `cagra::index` or a `std::vector<const cagra::index*>`, and must choose which path to use. That also means `num_partitions == 1` through the multi-partition API follows a different execution path from the regular single-index API.
> 
> Is the plan to keep this public API shape long term, or is this intended as an interim step until a multi-partition/partitioned index abstraction is introduced? For example, an index object carrying partition metadata could let search dispatch internally based on the index representation and avoid making callers choose between APIs.
> 
> If the multi-partition API remains separate, should `num_partitions == 1` be treated explicitly? For example, it could either be rejected with a clear error, or at least warn/document that this path is not equivalent to the regular single-index search path and may have different overhead.

Hi @anaruse, thank you for your comments. There has been a discussion on API design [here](https://github.com/NVIDIA/cuvs/pull/2035#discussion_r3371586736) but it's probably been buried by all the updates since. The current consensus is to provide this feature quickly to satisfy customer needs while allowing ourselves some more time to think through the appropriate long-term abstraction to unify these search functions.

Re: `num_partitions == 1`, it is technically possible for the multi-partition execution path to subsume the single-partition one as a base case. For this PR, I deliberately tried (where possible) to keep the changes separate and additive so as to avoid regression in the pre-existing CAGRA search. As the multi-partition code path matures, it may become advisable to combine them. Currently, the initial user of this API is cuVS-Lucene and there is value in the simplicity of sticking to the multi-partition API even when `num_partitions == 1` (e.g. single path of failure). To document the potentially different behavior in the single-partition case, I would like to go with a comment in the cuVS interface rather than a loud warning that a cuVS-Lucene end-user would see without knowing what to do about.

Will address your other comments and update.

jamxia155:
/ok to test https://github.com/NVIDIA/cuvs/pull/2035/commits/6b2b0b321258a03735da1787a543b412703acf54

c/include/cuvs/core/c_api.h
```diff
@@ -210,6 +226,19 @@ cuvsError_t cuvsRMMFree(cuvsResources_t res, void* ptr, size_t bytes);
...
+ * (cudaMallocAsync / cudaFreeAsync). Unlike the pool resource, this resource returns memory to
+ * the stream immediately without blocking the CPU, eliminating device-wide synchronization on
+ * deallocation. This is especially beneficial when multiple CAGRA searches run concurrently on
+ * separate CUDA streams, because the internal workspace allocations no longer serialize kernel
+ * launches. Be aware that this function will change the memory resource for the whole process
+ * and the new memory resource will be used until explicitly changed.
+ *
+ * @return cuvsError_t
+ */
+cuvsError_t cuvsRMMAsyncMemoryResourceEnable();
```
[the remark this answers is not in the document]
jamxia155: Removed redundant thread_local variable.

c/src/neighbors/cagra.cpp
```diff
@@ -689,6 +689,54 @@ extern "C" cuvsError_t cuvsCagraSearch(cuvsResources_t res,
...
+    std::vector<raft::device_matrix_view<DistanceT, int64_t, raft::row_major>> d_vec(num_segments);
+
+    for (uint32_t i = 0; i < num_segments; i++) {
+      RAFT_EXPECTS(indices[i] != nullptr && indices[i]->addr != 0,
+                   "Index at position %u is null or not built", i);
+      idx_vec[i] = reinterpret_cast<const IndexT*>(indices[i]->addr);
+      q_vec[i]   = cuvs::core::from_dlpack<std::remove_reference_t<decltype(q_vec[i])>>(queries[i]);
+      n_vec[i]   = cuvs::core::from_dlpack<std::remove_reference_t<decltype(n_vec[i])>>(neighbors[i]);
+      d_vec[i]   = cuvs::core::from_dlpack<std::remove_reference_t<decltype(d_vec[i])>>(distances[i]);
+    }
```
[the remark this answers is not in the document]
jamxia155: Added the validation.

c/src/selection/select_k.cpp
```diff
@@ -0,0 +1,42 @@
...
+    auto out_idx_view = raft::make_device_matrix_view<int64_t, int64_t, raft::row_major>(
+      static_cast<int64_t*>(out_idx->dl_tensor.data), 1, k);
+
+    cuvs::selection::select_k(
+      *res_ptr,
+      in_view,
+      std::nullopt,  // implicit positions [0, n) as in_idx
+      out_val_view,
+      out_idx_view,
+      true);  // select_min = true (smallest distance = nearest neighbor)
```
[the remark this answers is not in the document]
jamxia155: Added validation via existing helper `cuvs::core::from_dlpack`.

cpp/include/cuvs/neighbors/cagra.hpp
```diff
@@ -1723,6 +1724,87 @@ void search(raft::resources const& res,
...
+  const std::vector<raft::device_matrix_view<uint32_t, int64_t, raft::row_major>>& neighbors,
+  const std::vector<raft::device_matrix_view<float, int64_t, raft::row_major>>& distances);
+
+void search_multi_segment(
+  raft::resources const& res,
+  cuvs::neighbors::cagra::search_params const& params,
+  const std::vector<const cuvs::neighbors::cagra::index<uint8_t, uint32_t>*>& indices,
+  const std::vector<raft::device_matrix_view<const uint8_t, int64_t, raft::row_major>>& queries,
+  const std::vector<raft::device_matrix_view<int64_t, int64_t, raft::row_major>>& neighbors,
+  const std::vector<raft::device_matrix_view<float, int64_t, raft::row_major>>& distances);
```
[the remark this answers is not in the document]
jamxia155: Added Doxygen.

jamxia155:
/ok to test https://github.com/NVIDIA/cuvs/pull/2035/commits/7a7c3e69871ef7b07b881fb9da8fd10de9f9a0e2

java/cuvs-java/src/main/java/com/nvidia/cuvs/CuVSResources.java
```diff
@@ -57,6 +57,24 @@ interface ScopedAccess extends AutoCloseable {
...
+   * high-water mark only increases until the resources object is closed.
+   *
+   * <p>The pool is per-resources-handle (i.e. per query thread when resources are thread-local),
+   * so there is no cross-thread pool mutex contention. Call this once after creating the resources
+   * object; calling it again replaces the pool.
+   *
+   * @param initialSizeBytes initial pool reservation in bytes; size {@code initialSizeBytes} to
+   *                         cover the steady-state working set to avoid growth after warmup
+   */
+  void setWorkspacePool(long initialSizeBytes);
```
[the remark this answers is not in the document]
jamxia155: Added validation.

java/cuvs-java/src/main/java/com/nvidia/cuvs/SynchronizedCuVSResources.java
```diff
@@ -40,6 +40,11 @@ public void close() {
     inner.close();
   }
 
+  @Override
+  public void setWorkspacePool(long sizeBytes) {
+    inner.setWorkspacePool(sizeBytes);
+  }
```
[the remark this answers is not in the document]
jamxia155: Added locking.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/CagraIndexImpl.java
```diff
@@ -338,6 +338,123 @@ public SearchResults search(CagraQuery query) throws Throwable {
...
+    final long prefilterLen = hasPreFilter ? (prefilterDataLength + 31) / 32 : 0;
+    final long prefilterBytes = C_INT_BYTE_SIZE * prefilterLen;
+
+    // Pointers into the global buffer at this segment's slice.
+    long neighborByteOffset = (long) segmentIdx * topK * C_INT_BYTE_SIZE;
+    long distanceByteOffset = (long) segmentIdx * topK * Float.BYTES;
+    MemorySegment neighborSlice =
+        MemorySegment.ofAddress(globalNeighborsDP.address() + neighborByteOffset);
+    MemorySegment distanceSlice =
+        MemorySegment.ofAddress(globalDistancesDP.address() + distanceByteOffset);
```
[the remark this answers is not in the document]
jamxia155: This section of the code was for global top-k merge which has since been moved inside the C++ layer.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/CudaStreamPool.java
```diff
@@ -0,0 +1,133 @@
...
+        checkCuVSError(cuvsStreamSet(resources[i], streams[i]), "cuvsStreamSet");
+
+        // Pre-allocate one reusable event per slot (disable timing to avoid overhead).
+        var pEvent = arena.allocate(cudaEvent_t);
+        checkCudaError(
+            cudaEventCreateWithFlags(pEvent, cudaEventDisableTiming()), "cudaEventCreate");
+        events[i] = pEvent.get(cudaEvent_t, 0);
+      }
+    }
+  }
```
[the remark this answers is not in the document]
jamxia155: `CudaStreamPool` has since been removed from this PR.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/CudaStreamPool.java
```diff
@@ -0,0 +1,133 @@
...
+  // Lifecycle
+  // -------------------------------------------------------------------------
+
+  @Override
+  public void close() {
+    for (int i = 0; i < size; i++) {
+      checkCudaError(cudaEventDestroy(events[i]), "cudaEventDestroy");
+      cuvsResourcesDestroy(resources[i]);
+      checkCudaError(cudaStreamDestroy(streams[i]), "cudaStreamDestroy");
+    }
```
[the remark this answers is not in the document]
jamxia155: `CudaStreamPool` has since been removed from this PR.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/CuVSResourcesImpl.java
```diff
@@ -28,6 +28,9 @@ public class CuVSResourcesImpl implements CuVSResources {
   private final int deviceId;
 
   private final PinnedMemoryBuffer hostBuffer = new PinnedMemoryBuffer();
+  private final CudaStreamPool streamPool =
+      new CudaStreamPool(
+          Integer.getInteger(CudaStreamPool.SIZE_PROPERTY, CudaStreamPool.DEFAULT_SIZE));
```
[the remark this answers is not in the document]
jamxia155: `CudaStreamPool` has since been removed from this PR.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/MultiSegmentCagraSearch.java
```diff
@@ -0,0 +1,220 @@
...
+      try (var globalNeighborsDP = allocateRMMSegment(cuvsRes, neighborsBytes);
+          var globalDistancesDP = allocateRMMSegment(cuvsRes, distancesBytes);
+          var outIdxDP = allocateRMMSegment(cuvsRes, outIdxBytes);
+          var outValDP = allocateRMMSegment(cuvsRes, outValBytes)) {
+
+        // --- Phase 1: call cuvsCagraSearchMultiSegment ---
+        // Single kernel launch covers all segments; results land in globalNeighborsDP /
+        // globalDistancesDP on the same CUDA stream, so SelectK below sees them via ordering.
+        try (var arena = Arena.ofConfined()) {
+          MemorySegment sp = CuVSParamsHelper.buildCagraSearchParams(arena, searchParameters);
```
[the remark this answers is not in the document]
jamxia155: `MultiPartitionSearchResults` now only accepts one `CagraQuery` as input.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/MultiSegmentCagraSearch.java
```diff
@@ -0,0 +1,220 @@
...
+
+            // Distances DLTensor — slice of global buffer
+            long dByteOffset = (long) i * k * Float.BYTES;
+            MemorySegment dSlice =
+                MemorySegment.ofAddress(globalDistancesDP.handle().address() + dByteOffset);
+            distancesArray.setAtIndex(
+                ValueLayout.ADDRESS,
+                i,
+                prepareTensor(arena, dSlice, segShape, kDLFloat(), 32, kDLCUDA()));
+          }
```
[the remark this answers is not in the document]
jamxia155: `MultiPartitionSearchResults` now only accepts one `CagraQuery` as input.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/CudaStreamPool.java
```diff
@@ -0,0 +1,133 @@
...
+
+  /**
+   * Advances the round-robin counter by {@code count} and returns the starting slot index for
+   * this call. Slot indices are wrapped modulo {@link #size()}.
+   */
+  public int nextSlot(int count) {
+    int start = slotCounter;
+    slotCounter += count;
+    return start;
+  }
```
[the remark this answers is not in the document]
jamxia155: `CudaStreamPool` has since been removed from this PR.

cpp/src/neighbors/cagra.cuh
```diff
@@ -406,6 +406,19 @@ index<T, IdxT> merge(raft::resources const& handle,
...
+  raft::resources const& res,
+  search_params const& params,
+  const std::vector<const index<T, IdxT>*>& indices,
+  const std::vector<raft::device_matrix_view<const T, int64_t, raft::row_major>>& queries,
+  const std::vector<raft::device_matrix_view<OutputIdxT, int64_t, raft::row_major>>& neighbors,
+  const std::vector<raft::device_matrix_view<float, int64_t, raft::row_major>>& distances)
+{
+  cagra::detail::search_multi_segment<T, OutputIdxT, IdxT>(
+    res, params, indices, queries, neighbors, distances);
+}
```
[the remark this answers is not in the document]
jamxia155: These checks already existed in the detail layer; now moved upfront.

java/cuvs-java/src/main/java/com/nvidia/cuvs/MultiSegmentSearchResults.java
```diff
@@ -0,0 +1,55 @@
...
+ * segment-specific {@code ordToDoc} function and adding {@code docBase}.
+ *
+ * @since 25.10
+ */
+public class MultiSegmentSearchResults {
+
+  private final int count;
+  private final int[] segmentIndices;
+  private final int[] ordinals;
+  private final float[] distances;
```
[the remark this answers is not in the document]
jamxia155: Kept int representation but added conditional handling for negative ordinal: silently drop 0xFFFFFFFF (the unfilled slot sentinel); throw otherwise.

jamxia155:
/ok to test [1770971](https://github.com/NVIDIA/cuvs/pull/2035/commits/1770971a86ad3d65af43a7b0a06d6bfa40486399)

c/src/neighbors/cagra.cpp
```diff
@@ -689,6 +689,54 @@ extern "C" cuvsError_t cuvsCagraSearch(cuvsResources_t res,
...
+
+    cuvs::neighbors::cagra::search_multi_segment(
+      *res_ptr, search_params, idx_vec, q_vec, n_vec, d_vec);
```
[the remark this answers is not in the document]
jamxia155: Added requirement for uniform distance metric.

cpp/src/neighbors/detail/cagra/cagra_search.cuh
jamxia155: This was an oversight. Added the chunking mechanism to align with existing practice. Thanks.

cpp/src/neighbors/detail/cagra/cagra_search.cuh
jamxia155: Added validation for uniform distance metrics.

cpp/src/neighbors/detail/cagra/cagra_search.cuh
jamxia155: Added validation for uniform graph degree. Also added a note to the public header stating the requirements:
```
All index partitions must use the same distance metric and graph degree; partition sizes may differ.
```

anaruse:
Thanks @jamxia155.
The added max_queries chunking, the metric/graph-degree validation, and the API notes address my concerns.
LGTM.

jamxia155:
/ok to test [d4c5ff7](https://github.com/NVIDIA/cuvs/pull/2035/commits/d4c5ff7ce14b978ead688ef5bd4f6290b6c7b279)

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/FilterBitsetHandleImpl.java
```diff
@@ -0,0 +1,112 @@
...
+  /**
+   * Returns the shared device allocation for this filter, uploading on first call (lazy,
+   * thread-safe via double-checked locking).
+   *
+   * @param cuvsRes the native cuvsResources handle for the calling thread
+   */
+  DeviceData getOrUpload(long cuvsRes) {
+    if (closed) throw new IllegalStateException("FilterBitsetHandle has been closed");
+    DeviceData data = sharedDeviceData;
+    if (data != null) return data;
```
imotov, line 174: This is a slightly different variation of the cache issue in cuvs-lucene I mentioned yesterday, but we may still encounter a race condition where an acquired resource can be closed while it is still in use. I think both cases require a different approach. Perhaps we should switch to a reference-counting pattern with explicit release for managing these resources.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/FilterBitsetHandleImpl.java
```diff
@@ -0,0 +1,112 @@
...
+    // Host arena must outlive the stream sync that confirms the H2D copy.
+    try (var arena = Arena.ofConfined()) {
+      MemorySegment hostBitset = arena.allocate(combinedBitsetBytes, Long.BYTES);
+      MemorySegment.copy(
+          combinedLongs, 0, hostBitset, ValueLayout.JAVA_LONG, 0, combinedLongs.length);
+      cudaMemcpyAsync(
+          combinedBitsetDP.handle(), hostBitset, combinedBitsetBytes, HOST_TO_DEVICE, stream);
+
+      checkCuVSError(cuvsStreamSync(cuvsRes), "cuvsStreamSync in FilterBitsetHandle.upload");
+    }
```
imotov: If any of these lines throw an exception, it appears that we may leak an unclosed `CloseableRMMAllocation`.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/MultiPartitionCagraSearchImpl.java
```diff
@@ -0,0 +1,204 @@
...
+              prepareTensor(arena, partitionIdsDP.handle(), outShape, kDLUInt(), 32, kDLCUDA());
+          MemorySegment neighborsTensor =
+              prepareTensor(arena, neighborsDP.handle(), outShape, kDLUInt(), 32, kDLCUDA());
+          MemorySegment distancesTensor =
+              prepareTensor(arena, distancesDP.handle(), outShape, kDLFloat(), 32, kDLCUDA());
+
+          MemorySegment filterSeg = cuvsFilter.allocate(arena);
+          if (filter != null) {
+            FilterBitsetHandleImpl.DeviceData dev =
+                ((FilterBitsetHandleImpl) filter).getOrUpload(cuvsRes);
```
imotov: The API suggests that any implementation of `FilterBitsetHandle` can be used as a filter parameter. However, the current implementation assumes the object is a `FilterBitsetHandleImpl` and will throw a `ClassCastException` otherwise.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/CuVSParamsHelper.java
```diff
@@ -25,6 +27,27 @@ public final class CuVSParamsHelper {
...
+  /**
+   * Allocates and populates a {@code cuvsCagraSearchParams} struct into {@code arena}.
+   * The returned segment is valid for the lifetime of {@code arena}.
+   */
+  public static MemorySegment buildCagraSearchParams(Arena arena, CagraSearchParams params) {
+    MemorySegment seg = cuvsCagraSearchParams.allocate(arena);
+    cuvsCagraSearchParams.max_queries(seg, params.getMaxQueries());
+    cuvsCagraSearchParams.itopk_size(seg, params.getITopKSize());
+    cuvsCagraSearchParams.max_iterations(seg, params.getMaxIterations());
+    cuvsCagraSearchParams.algo(seg, params.getCagraSearchAlgo().value);
```
imotov: Since `CagraSearchParams.Builder` doesn't enforce non-null values, we shouldn't assume that `params.getCagraSearchAlgo()` is never null. Same with hashMapMode below.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/MultiPartitionCagraSearchImpl.java
```diff
@@ -0,0 +1,204 @@
...
+      throws Throwable {
+    int numPartitions = indices.size();
+    if (numPartitions == 0) {
+      return new MultiPartitionSearchResults(0, new int[0], new int[0], new float[0]);
+    }
+
+    CagraIndexImpl[] buffered = new CagraIndexImpl[numPartitions];
+    for (int i = 0; i < numPartitions; i++) {
+      CagraIndex idx = indices.get(i);
+      if (!(idx instanceof CagraIndexImpl)) {
```
imotov, line 92: This seems somewhat fragile, though it's not unique to this change since we already rely on a similar pattern in other places. We should probably take another look at the CagraIndex abstraction at some point.

imotov (commented):
I reviewed the java portion and left a few comments.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/MultiPartitionCagraSearchImpl.java
cjnolet: Agree about this- ideally we'd allow the polymorphism to work it's magic here instead of having to do explicit type checking. I agree this is fairly brittle. Hopefully we can work towards improving this.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/FilterBitsetHandleImpl.java
jamxia155: Implemented reference counting. Thanks.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/FilterBitsetHandleImpl.java
jamxia155: Strengthened exception handling logic. Thanks.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/MultiPartitionCagraSearchImpl.java
jamxia155: Added a runtime check with a clearer error message, and documentation stating the requirement that the handle must come from `FilterBitsetHandle.create()`.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/CuVSParamsHelper.java
jamxia155: Mapped null values to `AUTO`/`AUTO_HASH`.

jamxia155:
/ok to test [518dfcc](https://github.com/NVIDIA/cuvs/pull/2035/commits/518dfcce42d7f28fef374faa7a269d04b59c6bfa)

jamxia155:
/ok to test [f775d81](https://github.com/NVIDIA/cuvs/pull/2035/commits/f775d81ac745cfa7530610c474f1a1f42326fec5)

jamxia155:
/ok to test [483348d](https://github.com/NVIDIA/cuvs/pull/2035/commits/483348d8f2888e60c2cb0bf2f1585a5001afd62f)

jamxia155:
/ok to test [d85a21f](https://github.com/NVIDIA/cuvs/pull/2035/commits/d85a21f4c5e505d938c2f6612bffa08ca12f3aa2)

cpp/src/neighbors/detail/cagra/cagra_search.cuh
```diff
@@ -262,4 +267,464 @@ void search_main(raft::resources const& res,
...
+    const size_t chunk_partition_stride = static_cast<size_t>(chunk_queries) * per_partition_topk;
+
+    // Query norms, needed only by CosineExpanded. Each chunk covers a disjoint set of queries, so
+    // its norms are computed over just this chunk's rows (each query's norm computed once overall).
+    auto chunk_query_norms =
+      raft::make_device_vector_view<DistanceT, int64_t>(query_norms.data_handle(), chunk_queries);
+    if (needs_query_norms) {
+      auto scaled_sq_op = raft::compose_op(raft::sq_op{},
+                                           raft::div_const_op<DistanceT>{DistanceT(kScale)},
+                                           raft::cast_op<DistanceT>());
```
achirkin: Perhaps it would be cleaner to construct the mapping object directly rather than manually dividing the scales:
```suggestion
      auto scaled_sq_op = raft::compose_op(raft::sq_op{},
                                           cuvs::spatial::knn::detail::utils::mapping<DistanceT>{});
```

cpp/src/neighbors/detail/cagra/cagra_search.cuh
```diff
@@ -262,4 +267,464 @@ void search_main(raft::resources const& res,
...
+  uint32_t per_partition_topk = 0;
+
+  // Workspaces reused across chunks; sized to one max_queries-sized chunk once per_partition_topk
+  // is known. The intermediate buffers are laid out [num_partitions, chunk_queries,
+  // per_partition_topk]; transposed_distances and positions_buf hold one chunk's merge scratch.
+  lightweight_uvector<graph_idx_type> intermediate_neighbors(res);
+  lightweight_uvector<DistanceT> intermediate_distances(res);
+  lightweight_uvector<DistanceT> transposed_distances(res);
+  lightweight_uvector<uint32_t> positions_buf(res);
+  auto query_norms = raft::make_device_vector<DistanceT, int64_t>(res, max_queries);
```
achirkin: Performance danger zone: you're allocating during search. At the very least, use the workspace resource for allocating the vector, so we can rely on the user initializing it with a pool.
Then, I'd suggest to consider whether to use `lightweight_uvector` (if you know it's only ever used in one thread and one stream), and whether it makes sense to clump all these five allocations in one "workspace" allocation and manually map the pointers.

cpp/src/neighbors/detail/cagra/cagra_search.cuh
```diff
@@ -262,4 +267,464 @@ void search_main(raft::resources const& res,
...
+    // `per_partition_topk`. The two differ when the kernel emits more than `topk` candidates per
+    // partition (e.g. MULTI_CTA mp).
+    auto partition_ids_slice = raft::make_device_matrix_view<uint32_t, int64_t, raft::row_major>(
+      partition_ids.data_handle() + static_cast<size_t>(qid) * topk, chunk_queries, topk);
+    {
+      const uint32_t per_partition_topk_u32 = per_partition_topk;
+      raft::linalg::map(
+        res,
+        partition_ids_slice,
+        [per_partition_topk_u32] __device__(uint32_t pos) { return pos / per_partition_topk_u32; },
```
achirkin: You've used so nicely the raft op structs above, maybe you can use here too to avoid a lambda? :)

cpp/src/neighbors/detail/cagra/cagra_search.cuh
```diff
@@ -262,4 +267,464 @@ void search_main(raft::resources const& res,
...
+    intermediate_neighbors.resize(chunk_capacity, stream);
+    intermediate_distances.resize(chunk_capacity, stream);
+    transposed_distances.resize(chunk_capacity, stream);
+    positions_buf.resize(static_cast<size_t>(max_queries) * topk, stream);
+
+    using part_desc_t = multi_cta_search::multi_partition_desc_t<T, graph_idx_type, DistanceT>;
+    std::vector<part_desc_t> host_part_descs(num_partitions);
+
+    std::vector<dataset_descriptor_host<T, graph_idx_type, DistanceT>> part_dataset_descs;
+    part_dataset_descs.reserve(num_partitions);
```
achirkin: Would a `raft::make_host_vector<...>(res, num_partitions)` work here? It would help to account for the allocation with memory tracking resources and the dry run mode in future.

cpp/src/neighbors/detail/cagra/cagra_search.cuh
jamxia155: Updated as suggested. Thanks!

cpp/src/neighbors/detail/cagra/cagra_search.cuh
jamxia155: Consolidated the 5 internal allocations into one `lightweight_uvector`. Thanks!

cpp/src/neighbors/detail/cagra/cagra_search.cuh
jamxia155: Updated. Thanks for the spot.

cpp/src/neighbors/detail/cagra/cagra_search.cuh
jamxia155: Updated `host_part_descs` to use RAFT host vector. However, `dataset_descriptor_host` is a non-trivial type and cannot be used in `host_vector`. Having said that, `part_dataset_descs` is only 48 bytes * num_partitions so hopefully it's a rounding error for memory tracking?

jamxia155:
/ok to test [de32dde](https://github.com/NVIDIA/cuvs/pull/2035/commits/de32dde278dfdf15733431a64f06e0ab6c58f565)

jamxia155:
/ok to test [d99349e](https://github.com/NVIDIA/cuvs/pull/2035/commits/d99349ec45828520a94d8d042aa7377b0f01734f)

cjnolet (approved):

jamxia155:
/ok to test [4a60735](https://github.com/NVIDIA/cuvs/pull/2035/commits/4a60735c94187b9c8f00c1046f2514b5ce5b3829)

jamxia155:
/ok to test [33da211](https://github.com/NVIDIA/cuvs/commit/33da211ef0e83ebbc325dce3406750350fc83b79)

KyleFromNVIDIA (approved):

PR #2040, opened 2026-04-25
Centralize shared utilities across benchmark backends

jnke2016:
### Summary

This PR centralizes duplicated logic identified during review of the OpenSearch and Elasticsearch backend PRs. Both backends independently reimplemented common operations (vector file loading, YAML parsing, dataset lookup, DatasetConfig construction, parameter expansion, recall computation) that should be shared across all backends. This PR moves that shared logic into the framework so that backend authors focus only on backend-specific concerns. It also routes the CLI through the new pluggable benchmark API and restores a fully CLI-based workflow.

### Motivation

With the pluggable backend architecture, each new Python-native backend needs to load binary vector files, parse dataset YAML configs, look up datasets by name, expand parameter combinations, and compute recall. Without centralized utilities, each backend reimplements these independently, leading to:

- Format drift (e.g., OpenSearch supports 5 binary formats while Elasticsearch only supports 2)
- If a bug is found or a general change is needed, every backend needs to be updated individually instead of fixing it in one place
- Unnecessary code for new backend authors to write

### Changes

**1. Shared vector loading utility (`backends/utils.py`)**

Added `load_vectors()` and `dtype_from_filename()` (originating from `generate_groundtruth/utils.py`) for reading binary vector files. Supports all formats (`.fbin`, `.f16bin`, `.u8bin`, `.i8bin`, `.ibin`) with `subset_size` support and input validation (unsupported extensions, truncated headers/payloads). The C++ backend is unaffected since it passes file paths to the subprocess.

**2. Promoted config-loading methods to base `ConfigLoader` class**

Moved `load_yaml_file`, `get_dataset_configuration`, and `gather_algorithm_configs` from `CppGBenchConfigLoader` to the base `ConfigLoader` class. These were being reimplemented by both the OpenSearch and Elasticsearch config loaders. They are now inherited automatically. Also added a warning for invalid `algorithm_configuration` paths (previously silently ignored). Switched `gather_algorithm_configs` from exclusion-based filtering to `.yaml`/`.yml` inclusion.

**3. Dataset refactored from dataclass to regular class with transparent vector loading**

The `Dataset` class now transparently loads vectors from file paths on first access via properties. Python-native backends just access `dataset.base_vectors` and get a numpy array regardless of whether vectors were provided directly or loaded from disk. The C++ backend continues to use `dataset.base_file` only, never triggering any loading.

**4. ConfigLoader template method**

`ConfigLoader.load()` is now a concrete method that handles the shared steps (loading dataset YAML, finding the dataset by name, constructing `DatasetConfig` with path resolution via `build_dataset_config()`) then delegates to the abstract `_build_benchmark_configs()` for backend-specific logic. Backend authors implement `_build_benchmark_configs()` and receive the `DatasetConfig` already built. They cannot skip or reimplement the shared steps. The orchestrator is unchanged.

**5. Centralized parameter expansion (`backends/utils.py`)**

Added `expand_param_grid()` for expanding YAML parameter specs (dict-of-lists to list-of-dicts via Cartesian product). All backends use it transparently. The C++ backend uses it in `prepare_indexes()` (replacing inline `itertools.product`) and then applies its own constraint validation on the expanded combinations. Python-native backends use it directly without filtering.

**6. Recall computation moved to orchestrator**

Added `compute_recall()` to `backends/utils.py`. The orchestrator now computes recall transparently after `backend.search()` returns, for any backend that returns actual neighbor arrays. The C++ backend computes recall in the subprocess and returns empty neighbors, so this is naturally skipped for it. Backends no longer need to compute recall themselves.

**7. CLI routed through orchestrator**

Updated `run/__main__.py` to call `BenchmarkOrchestrator` directly instead of the legacy `run.py`. Added new CLI flags: `--mode` (sweep/tune), `--constraints` (tune mode targets), `--n-trials` (Optuna trials), and `--backend-config` (YAML config for non-C++ backends). The deprecation warning has been removed. The legacy `run.py` and `runners.py` are kept as deprecated for now.

**8. Restored CLI-based docs**

Updated `docs/source/cuvs_bench/index.rst` to use CLI commands for step 2 (build and search) instead of Python `BenchmarkOrchestrator` code blocks. All four steps (prepare, build/search, export, plot) are now fully CLI-based.

**9. New and updated tests**

- `tests/test_utils.py` (45 tests): covers `dtype_from_filename`, `load_vectors`, Dataset transparent loading, base ConfigLoader methods, `expand_param_grid`, and `compute_recall`
- `tests/test_cli.py` (5 new tests): covers `--mode`, `--backend-config`, `--n-trials`, `--constraints` flags and invalid backend config handling

### Remaining work

- Consider consolidating pre-existing I/O duplicates across `generate_groundtruth/utils.py`, `get_dataset/fbin_to_f16bin.py`, and `get_dataset/hdf5_to_fbin.py` into a top-level `cuvs_bench/io.py` module
- Streaming write capability (incremental batches without knowing total size upfront) 
- Remove `run.py` and `runners.py` once deprecated period is over


python/cuvs_bench/cuvs_bench/backends/base.py
```diff
@@ -38,25 +48,76 @@ class Dataset:
...
+        # Vectors are stored privately to support lazy loading.
+        # If the caller provides vectors directly, they are used as-is.
+        # If empty and a file path is provided, the corresponding property
+        # loads vectors from the file on first access. This keeps the
+        # loading logic invisible to backends: they just access
+        # dataset.base_vectors and get a numpy array regardless of whether
+        # it was passed in or loaded from disk.
+        self._base_vectors = base_vectors if base_vectors is not None else np.empty((0, 0))
+        self._query_vectors = query_vectors if query_vectors is not None else np.empty((0, 0))
+        self._groundtruth_neighbors = groundtruth_neighbors
```
jrbourbeau, line 92: Why is `_groundtruth_neighbors` treated differently here than  `_base_vectors` and `_query_vectors`?

python/cuvs_bench/cuvs_bench/orchestrator/config_loaders.py
```diff
@@ -152,6 +228,138 @@ def backend_type(self) -> str:
...
+            A list of paths to the algorithm configuration files.
+        """
+        algos_conf_fs = os.listdir(os.path.join(config_path, "algos"))
+        algos_conf_fs = [
+            os.path.join(config_path, "algos", f)
+            for f in algos_conf_fs
+            if ".json" not in f
+            and "constraint" not in f
+            and ".py" not in f
+            and "__pycache__" not in f
```
jrbourbeau: Are these supposed to be yaml files? If so, checking for that extension (`.yaml` + `.yml`) might be better. Maybe the situation is more complex though

jrbourbeau (approved):
Thanks @jnke2016 for moving these to a shared place. Overall LGTM -- just left a couple minor (nonblocking) comments

jrbourbeau (commented):
@jnke2016 just checking in here. I see this is marked as a draft -- are there other things you'd still like to include in this PR?

python/cuvs_bench/cuvs_bench/backends/base.py
jnke2016: `base_vectors` and `query_vectors `default to None in the constructor but are converted to np.empty((0, 0)) internally so that the dims, n_base, and n_queries properties can safely call .shape and .size without None guards. `groundtruth_neighbors `stays as None because it's optional (not every dataset has ground truth), and those properties don't depend on it.

jnke2016: However, I can refactor both to match `groundtruth_neighbors` by defaulting to None instead and introducing a None guard in the dims, n_base, and n_queries properties. This would make all three fields consistent

python/cuvs_bench/cuvs_bench/orchestrator/config_loaders.py
jnke2016: Yes those are supposed to be `yaml` file. The exclusion filter was part of the legacy benchmark scripts (run.py) in `gather_algorithm_configs()`, which lists all files in the `config/algos/` directory and filters out non-config files (`__init__.py`, `__pycache__/`, `constraints/`, `.json `files) to keep only the .yaml algorithm configuration files. It was ported as-is to the base ConfigLoader in the new API. Your suggestion to check for `.yaml/.yml` inclusion instead is better since it's more explicit and won't accidentally pick up new non-YAML file types that may be added to the directory in the future. I'll update it.

jnke2016:
@jrbourbeau Yes, there is one remaining change (minor): making the parameter expansion transparent so that the orchestrator handles it and no backend needs to implement. This should be pushed later today

python/cuvs_bench/cuvs_bench/orchestrator/config_loaders.py
```diff
@@ -127,23 +128,99 @@ class ConfigLoader(ABC):
...
+
+    @abstractmethod
+    def _build_benchmark_configs(
+        self,
+        dataset_config: DatasetConfig,
+        dataset_conf: dict,
+        dataset: str,
+        dataset_path: str,
+        **kwargs,
+    ) -> List[BenchmarkConfig]:
```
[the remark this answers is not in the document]
jnke2016: Nothing in the codebase imports it. The CLI and BenchmarkOrchestrator class cover all use cases.

python/cuvs_bench/cuvs_bench/backends/base.py
```diff
@@ -38,25 +48,108 @@ class Dataset:
...
+        self.groundtruth_distances = groundtruth_distances
+        self.distance_metric = distance_metric
+        self.base_file = base_file
+        self.query_file = query_file
+        self.groundtruth_neighbors_file = groundtruth_neighbors_file
+        self.metadata = metadata or {}
+
+    @property
+    def base_vectors(self) -> np.ndarray:
```
cjnolet: Maybe instead of base vectors, can we call this training_vectors for clarity?  It's a good contrast to `query_vectors`.

python/cuvs_bench/cuvs_bench/backends/base.py
```diff
@@ -38,25 +48,108 @@ class Dataset:
...
+            self._query_vectors = load_vectors(self.query_file)
+        return self._query_vectors
+
+    @query_vectors.setter
+    def query_vectors(self, value: Optional[np.ndarray]) -> None:
+        """Set query vectors directly."""
+        self._query_vectors = value if value is not None else np.empty((0, 0))
+
+    @property
+    def groundtruth_neighbors(self) -> Optional[np.ndarray]:
```
cjnolet, line 140: We should definitely store the neighbors AND the distances. At the very least for debugging purposes, we should have access to the distances so that users can print them out if needed.

python/cuvs_bench/cuvs_bench/backends/_utils.py
```diff
@@ -0,0 +1,202 @@
...
+            raise ValueError(
+                f"File is truncated: expected {expected_bytes} bytes of data "
+                f"({n_rows} rows x {n_cols} cols x {np.dtype(dtype).itemsize} bytes), "
+                f"got {len(raw)}: {path}"
+            )
+        data = np.frombuffer(raw, dtype=dtype)
+    return data.reshape(n_rows, n_cols)
+
+
+def expand_param_grid(
```
cjnolet, line 123: These functions seem like they should be core to the orchestrator / some other abstraction and shouldn't be merely utilities. Are these intended to be used directly by the backends? I think some sort of abstraction is needed to help force the backends to use them in some standard way. Otherwise, we're going to end up with users creating custom backends that don't follow our methodology and claiming they are getting some crazy different perf profile as a result (when in reality it's because they aren't using our tool in the intended way).

cjnolet: For example, I think there should be a class that centralizes all of the dataset stuff (liek what you have above) but also provides a standardized way to get these params (maybe this new abstraction uses the utilities here, but I think these utilities should be private (_utils.py) and we should favor that standard abstraction layer. Maybe in addition to Dateset class, we have a "BenchmarkRunner" class? Or a "ParamRunner" class that generates the params and evaluates the recall? What we should aim for is for the documentation for adding new backends to have a clean, cohesive, concise, and consistent few lines of code that each backend can execute in order to properly run benchmarks. This will keep everyone doing the same thing as much as possible so we're always comparing apples to apples.

python/cuvs_bench/cuvs_bench/orchestrator/orchestrator.py
```diff
@@ -253,6 +254,19 @@ def _run_sweep(
                     search_threads=search_threads,
                     dry_run=dry_run,
                 )
+
+                # Compute recall for backends that return actual neighbors.
+                # The C++ backend computes recall in the subprocess and returns
+                # empty neighbors, so this is skipped for it.
+                # Note: The recall computation relies on empty neighbors to
+                # distinguish backends that already computed recall.
+                if search_result.success and search_result.neighbors.size > 0:
```
cjnolet, line 263: This is a good centralization. Maybe this satisfies my concern above, at least about having a standardized way to compute recall without having to rely on the user to invoke a utility function in each backend.

python/cuvs_bench/cuvs_bench/tests/test_utils.py
```diff
@@ -0,0 +1,423 @@
...
+from cuvs_bench.backends.utils import (
+    compute_recall,
+    dtype_from_filename,
+    expand_param_grid,
+    load_vectors,
+)
+from cuvs_bench.orchestrator.config_loaders import CppGBenchConfigLoader
+
+
+def _write_test_bin(path, data):
```
cjnolet, line 23: Great to see all of these things being tested!

python/cuvs_bench/cuvs_bench/backends/base.py
jnke2016: Got it. I addressed this and it will be included in my next commit

python/cuvs_bench/cuvs_bench/backends/base.py
jnke2016: Got it

python/cuvs_bench/cuvs_bench/backends/_utils.py
jnke2016: > These functions seem like they should be core to the orchestrator / some other abstraction and shouldn't be merely utilities. Are these intended to be used directly by the backends? I think some sort of abstraction is needed to help force the backends to use them in some standard way. Otherwise, we're going to end up with users creating custom backends that don't follow our methodology and claiming they are getting some crazy different perf profile as a result (when in reality it's because they aren't using our tool in the intended way).

These functions are not intended to be called directly by backends. They are called transparently by the framework: `expand_param_grid` is called by the `config loader`, `compute_recall` is called by the `orchestrator`, and `load_vectors` is called by the `Dataset class`. Backends only implement `build()` and `search()`. The `utils.py` file is just a centralized place where these functions live, but the framework is what calls them, not the backends.

jnke2016: > For example, I think there should be a class that centralizes all of the dataset stuff (liek what you have above) but also provides a standardized way to get these params (maybe this new abstraction uses the utilities here ... Maybe in addition to Dateset class, we have a "BenchmarkRunner" class? Or a "ParamRunner" class that generates the params and evaluates the recall?

This is largely what the current architecture provides. The Dataset class centralizes all dataset handling with transparent vector loading. The base `ConfigLoader` standardizes parameter expansion via `expand_param_grid` internally. The BenchmarkOrchestrator covers the role of both a `BenchmarkRunner` and `ParamRunner`:
- Generates params: calls `config_loader.load()` which internally uses expand_param_grid
- Evaluates recall: calls `compute_recall` after `backend.search()` returns

Backend authors only implement `build()` and `search()` for a single configuration. Agreed on making utils.py private (_utils.py) to signal these are internal to the framework and not meant to be called by backends directly.

docs/source/cuvs_bench/index.rst
```diff
@@ -113,22 +113,8 @@ The steps below demonstrate how to download, install, and run benchmarks on a su
     # (1) Prepare dataset.
     python -m cuvs_bench.get_dataset --dataset deep-image-96-angular --normalize
 
-.. code-block:: python
-
     # (2) Build and search index.
-    from cuvs_bench.orchestrator import BenchmarkOrchestrator
```
cjnolet, line 119: It's wonderful that we have these ways of running the orchestrator through Python. Is it possible for us to use both approaches? Are there downsides to having one over the other? Can we keep both in the docs? 

For example, especially in the "tuning" mode, I think users are going to want to run that inside existing Python code more frequently than from the command line.

python/cuvs_bench/cuvs_bench/backends/_utils.py
cjnolet: These changes sound great, @jnke2016. Thank you for this! I think it's normal that we're seeing the design and scope evolve as we continue to add new backends.

python/cuvs_bench/cuvs_bench/backends/base.py
cjnolet: (I agree, we shouldn't necessarily make them required, but if a dataset has them, we should use them, I think).

cjnolet (approved):

python/cuvs_bench/cuvs_bench/backends/_utils.py
jrbourbeau: I'm working to rebase the opensearch PR on top of the changes here

> The base ConfigLoader standardizes parameter expansion via expand_param_grid internally.

@jnke2016 I just want to make sure I'm not missing something. From what I can tell it looks like the `CppGBenchConfigLoader` is using `expand_param_grid`, not the `ConfigLoader` base class. Am I understanding correctly that the opensearch config loader should also use `expand_param_grid` similar to the C++ config loader?

PR #2044, opened 2026-04-29
[REVIEW] Drop extra copy in `get_last_error_text`

jakirkham:
In `get_last_error_text`, the assignment of `const char*` to `bytes` results in a copy as `bytes` must own its memory. After that the `bytes` object is `decode`d to a Python `str`. However it is possible to `decode` directly from `const char*` to `str` avoiding the copy to the temporary `bytes` object. Hence this code makes that change.

Further Python and Cython `def` functions always `return None` if nothing else is `return`ed. Hence the logic can be simplified here to only focus on `return`ing the error text.

### DIFF
--- a/python/cuvs/cuvs/common/exceptions.pyx
+++ b/python/cuvs/cuvs/common/exceptions.pyx
@@ -1,5 +1,5 @@
 #
-# SPDX-FileCopyrightText: Copyright (c) 2024, NVIDIA CORPORATION.
+# SPDX-FileCopyrightText: Copyright (c) 2024-2026, NVIDIA CORPORATION.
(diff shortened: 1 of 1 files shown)


jakirkham:
Some CI jobs were seeing GitHub API timeouts (for [example]( https://github.com/rapidsai/cuvs/actions/runs/25135304259/job/73674905032#step:11:35 )). So needed to restart CI

aamijar (approved):
LGTM!

cjnolet (approved):

jakirkham:
Thank you Corey and Anupam! 🙏

PR #2066, opened 2026-05-07
[FEA] Multi-node Out of Core Streaming KMeans API

tarang-jain:
Merge after #2015 and #2017

Allows a stream of input matrices per worker, that are further batched using the `streaming_batch_size` parameter. Reasoning: We should be able to supply dask partitions (on host) directly without having to concatenate them into one consolidated matrix.

As a part of this PR, we also unify the multi-GPU implementations into one (earlier the out of core implementation was separate).
Tests: We get rid of the separate out of core test file. The single MG testing unit is taking care of both out of core and on device matrices.


cpp/src/cluster/detail/kmeans_mg_batched.cuh
```diff
@@ -0,0 +1,575 @@
...
+  auto minClusterAndDistance =
+    raft::make_device_vector<raft::KeyValuePair<IdxT, T>, IdxT>(dev_res, alloc_batch_size);
+  auto L2NormBatch = raft::make_device_vector<T, IdxT>(dev_res, alloc_batch_size);
+  rmm::device_uvector<T> L2NormBuf_OR_DistBuf(0, stream);
+  rmm::device_uvector<char> workspace(0, stream);
+  rmm::device_uvector<char> batch_workspace(0, stream);
+
+  // --- Weight normalization via allreduce (only when sample weights are provided) ---
+  T weight_scale = T{1};
+  if (sample_weight.has_value()) {
```
tarang-jain: here we should be checking for the size mismatch:
RAFT_EXPECTS(sample_weights.extent(0) == X.extent(0), ...)

cpp/src/cluster/detail/kmeans_mg_batched.cuh
```diff
@@ -0,0 +1,575 @@
...
+  }
+
+  const raft::resources& dev_res =
+    use_nccl ? raft::resource::set_current_device_to_rank(handle, rank) : handle;
+
+  auto stream     = raft::resource::get_cuda_stream(dev_res);
+  auto n_local    = X_local.extent(0);
+  auto n_features = X_local.extent(1);
+  auto n_clusters = static_cast<IdxT>(params.n_clusters);
+  auto metric     = params.metric;
```
tarang-jain: metric assertions might be missing.
RAFT_EXPECTS guard to mnmg_fit. Right now MG accepts params.metric into the loop, but the shared distance path only supports L2Expanded/L2SqrtExpanded.

cpp/src/cluster/detail/kmeans_mg_batched.cuh
```diff
@@ -0,0 +1,666 @@
...
+      raft::linalg::map_offset(
+        dev_res,
+        raft::make_device_vector_view<int64_t, int>(d_done_flag.data_handle(), 1),
+        [=] __device__(int) {
+          cuvs::cluster::kmeans::detail::check_convergence(
+            d_cost_view, d_prior_view, d_norm_view, tol, iter, d_done_view);
+          return *d_done_view.data_handle();
+        });
+
+      SNMG_ALLREDUCE(d_done_flag.data_handle(), d_done_flag.data_handle(), 1);
```
tarang-jain: why do we need this allreduce? Each rank computes its own done flag, right? And it would be based on the same centroids and inertia values, so it would be identical.

cpp/include/cuvs/cluster/kmeans.hpp
```diff
@@ -179,6 +179,13 @@ enum class kmeans_type { KMeans = 0, KMeansBalanced = 1 };
  * on the host. Data is processed in GPU-sized batches, streaming from host to device.
  * The batch size is controlled by params.streaming_batch_size.
  *
+ * Multi-GPU dispatch is selected automatically based on the handle state:
+ *   - If `raft::resource::is_multi_gpu(handle)` (cuVS SNMG): the full dataset X
```
cjnolet, line 205: Great!

cpp/src/cluster/kmeans_mg.hpp
```diff
@@ -1,64 +1,99 @@
...
-         const cuvs::cluster::kmeans::params& params,
-         raft::device_matrix_view<const float, int> X,
-         std::optional<raft::device_vector_view<const float, int>> sample_weight,
-         raft::device_matrix_view<float, int> centroids,
-         raft::host_scalar_view<float> inertia,
-         raft::host_scalar_view<int> n_iter);
+void fit(
+  raft::resources const& handle,
+  const cuvs::cluster::kmeans::params& params,
+  const std::vector<raft::device_matrix_view<const float, int>>& X_parts,
```
viclafargue: Please expose the new API in `cpp/include/cuvs/cluster/kmeans.hpp`.

cpp/tests/CMakeLists.txt
```diff
@@ -327,8 +327,8 @@ if(BUILD_MG_ALGOS)
   )
```
viclafargue, line 346: Commenting in this file, but could you edit `cpp/CMakeLists.txt` to make `src/cluster/kmeans_fit_mg_float.cu` and `src/cluster/kmeans_fit_mg_double.cu` compilation depend on `BUILD_MG_ALGOS` being `ON`?

cpp/src/cluster/detail/kmeans_mg.cuh
```diff
@@ -2,751 +2,618 @@
...
+    auto d_global_n = raft::make_device_scalar<IndexT>(dev_res, n_local);
+    comms.allreduce(d_global_n.data_handle(), d_global_n.data_handle(), 1);
+
+    auto d_wt = raft::make_device_scalar<DataT>(dev_res, DataT{0});
+    for (auto const& weights : sample_weight_parts.value()) {
+      auto n_weights = static_cast<IndexT>(weights.extent(0));
+      if (n_weights == 0) { continue; }
+
+      auto d_part_wt = raft::make_device_scalar<DataT>(dev_res, DataT{0});
+      cuvs::cluster::kmeans::detail::weightSum(dev_res, weights, d_part_wt.view());
```
viclafargue: Partition weights are validated per partition, not globally. `weightSum` throws an error when the partial sum is <= 0, but we should instead fail when the global sum (after reduce operation) is <= 0.

cpp/src/cluster/detail/kmeans_mg.cuh
```diff
@@ -2,751 +2,618 @@
...
-  const auto& comm    = raft::resource::get_comms(handle);
-  cudaStream_t stream = raft::resource::get_cuda_stream(handle);
-  const int my_rank   = comm.get_rank();
-  const int n_rank    = comm.get_size();
-
-  auto n_samples  = X.extent(0);
-  auto n_features = X.extent(1);
-  auto n_clusters = params.n_clusters;
+  auto stream     = comms.stream();
+  auto n_features = X_parts[0].extent(1);
```
viclafargue: The documentation says that the functions handles empty partitions, but it looks like if the first partition is empty it would cause a crash. `n_features` can be derived from `centroids`.

viclafargue (commented):
Thanks for working on this @tarang-jain. Unifying the implementations is a good step for maintainability, but I’d like to raise a few concerns around initialization.

This PR changes regular MG KMeans initialization semantics in ways that should be discussed explicitly:
- The truly distributed KMeans++ initialization is dropped. This limits the scaling of the initialization of centroids to what the root rank allows. This may be an acceptable change as there may be less communication steps with this simpler init, be it would need to be discussed.
- `init_size` is documented as a host-only parameter, but regular KMeans now uses it or defaults to `min(3 * n_clusters, global_n)`.
- Some NCCL communication patterns (left from the multi-GPU Batched KMeans PR) should be revisited :
     - **Array init** : Redundant broadcast from root. All ranks should normally already be correctly initialized with the user-defined centroids.
     - **Random init** : Points are randomly sampled on ranks, and then merged with an allreduce operation, finally they are broadcasted again from root. We should remove the unnecessary broadcast.
     - **KMeans++** : If we keep the current root-rank implementation, the communication pattern should be adjusted. Only the root rank needs the full initialization sample, so gather-to-root would be a better fit than dense allreduce over the full sampled matrix on every rank. The resulting centroids can then be broadcast.

cpp/src/cluster/kmeans_mg.hpp
tarang-jain: Yeah we need to find a good way to do that. The problem is that it would look identical to the single matrix, single GPU API. If a user passes an SG handle to this API, it would go through this MG impl

cpp/src/cluster/detail/kmeans_mg.cuh
tarang-jain: I intentionally removed the global assertion. Earlier we were doing a stream sync to bring the sum to host and validate on host. I dont think syncing the whole stream just for an assertion is worth the perf impact.

cpp/tests/CMakeLists.txt
tarang-jain: Done!

cpp/src/cluster/detail/kmeans_mg.cuh
tarang-jain: Done!

cpp/src/cluster/kmeans_mg.hpp
tarang-jain: I added them in the ::mg namespace.

cpp/src/cluster/kmeans_impl.cuh
```diff
@@ -37,7 +40,14 @@ void fit(raft::resources const& handle,
...
   if (raft::resource::comms_initialized(handle)) {
-    cuvs::cluster::kmeans::mg::fit(handle, params, X, sample_weight, centroids, inertia, n_iter);
+    // Wrap the single per-rank mdspan into a one-element vector so we can
+    // dispatch through the unified vector-of-partitions MG API.
+    std::vector<raft::device_matrix_view<const DataT, IndexT>> X_parts{X};
+    std::optional<std::vector<raft::device_vector_view<const DataT, IndexT>>> sw_parts;
+    if (sample_weight) {
+      sw_parts = std::vector<raft::device_vector_view<const DataT, IndexT>>{*sample_weight};
+    }
+    cuvs::cluster::kmeans::mg::fit(handle, params, X_parts, sw_parts, centroids, inertia, n_iter);
```
viclafargue: Could we condition the call to `kmeans::mg::fit` to `BUILD_MG_ALGOS` being true?

cpp/src/cluster/detail/kmeans_mg.cuh
viclafargue: Yes, but this is not the same check. Here if one partition has its weights set to 0 (unlikely but still valid input), it will fail. Since weights checks / normalization only happen once, the optimization may be unnecessary. We should still check the global sum and should not check the local sum.

cpp/src/cluster/detail/kmeans_mg.cuh
```diff
@@ -2,751 +2,622 @@
...
+        auto const& X_part = X_parts[part_idx];
+        auto part_rows     = static_cast<IndexT>(X_part.extent(0));
+        if (part_rows == 0) { continue; }
+
+        data_batch_iterator_t data_batches(dev_res,
+                                           X_part,
+                                           static_cast<size_t>(streaming_batch_size),
+                                           stream,
+                                           rmm::mr::get_current_device_resource_ref(),
+                                           true);
```
viclafargue, line 410: Could we instantiate the batch iterator out of the loop?

cpp/src/cluster/detail/kmeans_mg.cuh
```diff
@@ -2,751 +2,622 @@
...
+      auto const& X_part = X_parts[part_idx];
+      auto part_rows     = static_cast<IndexT>(X_part.extent(0));
+      if (part_rows == 0) { continue; }
+
+      data_batch_iterator_t data_batches(dev_res,
+                                         X_part,
+                                         static_cast<size_t>(streaming_batch_size),
+                                         stream,
+                                         rmm::mr::get_current_device_resource_ref(),
+                                         true);
```
viclafargue, line 538: Could we similarly instantiate the batch iterator out of the loop?

cpp/src/cluster/detail/kmeans_mg_batched_init.cuh
```diff
@@ -149,78 +166,97 @@ std::vector<IndexT> broadcast_sampled_global_indices(raft::resources const& hand
...
+    if constexpr (
+      raft::is_device_mdspan_v<
+        raft::mdspan<const DataT, raft::matrix_extent<IndexT>, raft::row_major, Accessor>>) {
+      raft::copy(handle, dst, raft::make_device_vector_view<const DataT, IndexT>(src, n_features));
+    } else {
+      raft::copy(handle, dst, raft::make_host_vector_view<const DataT, IndexT>(src, n_features));
+    }
   }
 
   comms.allreduce(sampled_rows.data_handle(), sampled_rows.data_handle(), sampled_rows.size());
```
viclafargue: Could we `gather` to root instead of doing an `allreduce`? It looks like this may be a very large data exchange.

viclafargue (commented):
Good progress. We're planning to bring back the old truly distributed `kmeansplusplus` initialization for data on device, correct?

cpp/src/cluster/detail/kmeans_mg.cuh
tarang-jain: It takes the mdspan as an argument (X_part)

cpp/src/cluster/detail/kmeans_mg.cuh
tarang-jain: It takes the mdspan as an argument (X_part)

cpp/src/cluster/detail/kmeans_mg_batched_init.cuh
tarang-jain: `gather` would first need a local gather, no? Each rank would have to gather the samples it owns into a local buffer and then gather it to root, so its two operations. I dont think there is a single operation to send only specific rows from each rank to root.
Currently I have changed it to `reduce` instead of `allreduce`.

cpp/src/cluster/kmeans_impl.cuh
tarang-jain: i have actually gotten rid of the mg dispatch from that header. It is confusing when we have two different multi-GPU signatures for the same function. Now the multi-gpu implementation lives specifically in the `mg` namespace and in a separate TU.

cpp/src/cluster/detail/kmeans_mg_batched_init.cuh
viclafargue: Going from `allreduce` to `reduce` is a great add for KMeans++ since it drastically reduces unnecessary comms, but still sends tons of zeroes back to the root rank. Ideally we would use `ncclGather`, but since all ranks do no necessarily sample a similar number of rows we could instead send on non-root, have a receiving loop on root.

This improved comms change is only valid for the sampling in preparation of KMeans++. Random sampling initialization on the other end is different because the samples then serve as the initial centroids so they have to be present on all ranks after init hence we should probably keep using allreduce for this.

tarang-jain:
> We're planning to bring back the old truly distributed kmeansplusplus initialization for data on device, correct?

Yes I just added that. So for data that is on device, we use the truly distributed algorithm and for data on host we use sampling.

cpp/src/cluster/detail/kmeans_mg_batched_init.cuh
tarang-jain: Ok so I added a parameter to `sample_global_rows` which optionally does reduce and allreduce --> we do reduce for kmeanspp and allreduce for random init.

tarang-jain:
/ok to https://github.com/rapidsai/cuvs/commit/100f7d6b8ee30b190a95cd146ea3408a50d05e2b

cpp/src/cluster/kmeans_impl.cuh
```diff
@@ -37,11 +21,12 @@ void fit(raft::resources const& handle,
...
 {
   if (raft::resource::comms_initialized(handle)) {
-    cuvs::cluster::kmeans::mg::fit(handle, params, X, sample_weight, centroids, inertia, n_iter);
-  } else {
-    cuvs::cluster::kmeans::detail::kmeans_fit<DataT, IndexT>(
-      handle, params, X, sample_weight, centroids, inertia, n_iter);
+    RAFT_LOG_WARN(
+      "Multi-GPU handle detected on single-GPU kmeans::fit() entry; "
+      "falling back to single-GPU. Use cuvs::cluster::kmeans::mg::fit(...) for multi-GPU.");
   }
```
viclafargue: Could you remind me why we need a `mg` namespace? Can't we dispatch according to the comms on the handle as we used to do? It looks like this may be a breaking change.

cpp/src/cluster/detail/kmeans_mg_distributed_init.cuh
```diff
@@ -0,0 +1,433 @@
...
+
+  // Step 1.2 - the source rank picks one of its local rows uniformly.
+  // Matches the original: re-seed mt19937(seed) so cIdx is the first draw of a
+  // fresh generator (not the second draw of the same one used for rp).
+  IndexT chosen_local_idx = -1;
+  auto initialCentroid    = raft::make_device_matrix<DataT, IndexT>(handle, 1, n_features);
+  if (rank == rp) {
+    RAFT_EXPECTS(n_local > 0,
+                 "selected source rank %d has no local rows; cannot pick an initial centroid",
+                 rp);
```
viclafargue, line 240: The `rp` rank is chosen at random. We expect it to contain at least one row, but the doc says that MG Batched Kmeans supports empty partitions/ranks. This is a bit of an edge case, but we should only sample on a rank that contains rows.

cpp/src/cluster/detail/kmeans_mg_distributed_init.cuh
```diff
@@ -0,0 +1,433 @@
...
+      std::ref(workspace));
+
+  } else if (static_cast<IndexT>(potentialCentroids.extent(0)) < n_clusters) {
+    // Fewer candidates than requested centroids; supplement with random rows
+    // drawn from the rank's local data (matches the original
+    // initRandom-on-X call). Replicates shuffleAndGather without
+    // concatenating the local parts: produce the same device permutation
+    // over [0, n_local), then map each of the first n_random local indices
+    // back to (part, row_in_part) and copy that single row into
+    // centroidsRawData.
```
viclafargue: The permutation may be the same, but the data being copied on each rank is different. We end up with different centroid candidates. If we have dissimilar candidates, the following KMeans++ init ran on these candidates on each rank is not deterministic anymore.

I think that sampling the missing candidates really comes back to a random initialization (sampling the missing candidates on all ranks). Basically, each rank should not sample the missing candidate from its data. Instead we should perform a global sampling.

cpp/tests/cluster/kmeans_mg.cu
```diff
@@ -1,189 +1,747 @@
...
+    snmg_params.rng_state.seed       = 42;
+    snmg_params.init                 = testparams_.init;
+    snmg_params.streaming_batch_size = testparams_.streaming_batch_size;
+
+    T snmg_inertia      = T{0};
+    int64_t snmg_n_iter = 0;
+
+    cuvs::cluster::kmeans::mg::fit(clique_,
+                                   snmg_params,
+                                   h_X_view,
```
viclafargue: The multi-partition per rank API (`std::vector`) is normally targeted for Dask, but we should be able to test it within a cuVS OpenMP test. Easy to mess up an offset calculation.

viclafargue (commented):
Thanks @tarang-jain! Good progress. Here are some comments.

cpp/src/cluster/kmeans_impl.cuh
tarang-jain: I want to keep the distinction so that the multi-gpu API is explicit even for single mdspan per rank. Rather than selectively compiling code blocks with the `MG` macro, we just selectively compile the entire multi-gpu TU. Secondly, in this PR we expose the std::vector<mdspan<...>> API to the public header. So a selective dispatch here means we are essentially having two public API signatures for the same functionality, which is confusing.

cpp/src/cluster/detail/kmeans_mg_distributed_init.cuh
tarang-jain: Actually that doc is not making a ton of sense. Why would a user even run a process on a GPU if there is no data on that device? Maybe I should remove that doc instead?

tarang-jain:
>Yes, but this is not the same check. Here if one partition has its weights set to 0 (unlikely but still valid input), it will fail. Since weights checks / normalization only happen once, the optimization may be unnecessary. We should still check the global sum and should not check the local sum.

@viclafargue I have re-added that host assertion.

cpp/src/cluster/detail/kmeans_mg_distributed_init.cuh
viclafargue: Dask users do not necessarily have any control on where their data is located. Because of this there may be edge cases with empty partitions/ranks, but this should be relatively rare. Like launching KMeans with less rows than ranks or something like this. I think we would be fine dropping this requirement in practice.

cpp/src/cluster/detail/kmeans_mg_distributed_init.cuh
tarang-jain: Yikes the global sampling got lost in the refactor too. I have updated this to reuse your `sample_global_rows` function to generate a global sample.

cpp/src/cluster/detail/kmeans_mg_distributed_init.cuh
tarang-jain: Ok I have updated the doc. For allowing completely empty partitions on a device, we might need to do checks in a lot of other places (such as `sample_global_rows`), which I dont think we currently have.

cpp/tests/cluster/kmeans_mg.cu
tarang-jain: I think what you are suggesting is we simulate dask arrays by making non-contiguous device chunks and putting them in a vector one each thread. Yes this can certainly be done.

tarang-jain: Done!

cpp/src/cluster/kmeans_impl.cuh
dantegd: What do you think about making this compatibility break more explicit? Before this PR, callers using `kmeans::fit` with initialized MG resources were routed to the distributed implementation; after this change they get a warning and a local single-GPU fit. I agree the explicit `mg::fit` API is cleaner, but this seems easy to miss and can silently change results in existing code. Could we either keep a hard error here for MG handles, or document the migration prominently in the API docs/release notes?

cpp/tests/CMakeLists.txt
```diff
@@ -327,8 +327,8 @@ if(BUILD_MG_ALGOS)
...
 
   ConfigureTest(
-    NAME CLUSTER_KMEANS_MG_BATCHED_TEST
-    PATH cluster/kmeans_mg_batched.cu
+    NAME CLUSTER_KMEANS_MG_TEST
+    PATH cluster/kmeans_mg.cu
```
dantegd, line 349: Since this test is registered with `GPUS 1`, local `ctest` runs will exercise the unified MG suite as a single-rank clique even on multi-GPU machines, no? That leaves the new multi-rank collectives and host-partition streaming path dependent on manually invoking the binary outside CTest. What do you think about adding a second, optional/local-only CTest registration with `GPUS 2+`, or documenting the local command used to validate the multi-GPU cases?

cpp/include/cuvs/cluster/kmeans.hpp
```diff
@@ -1607,6 +1608,196 @@ void cluster_cost(
  * @}
  */
 
+namespace mg {
```
dantegd: One compatibility question: the public header now declares cuvs::cluster::kmeans::mg::fit unconditionally, while cpp/CMakeLists.txt only compiles kmeans_fit_mg_float.cu and kmeans_fit_mg_double.cu under BUILD_MG_ALGOS. If a non-MG build installs this header, callers can compile against mg::fit but fail at link time. Would it be better to guard those declarations or provide stubs with a clear error?

cpp/src/cluster/detail/kmeans_mg_distributed_init.cuh
```diff
@@ -0,0 +1,422 @@
...
+
+  raft::random::RngState rng(params.rng_state.seed, raft::random::GeneratorType::GenPhilox);
+
+  auto d_rp = raft::make_device_scalar<int>(handle, 0);
+  int rp    = 0;
+
+  // Step 1.1 - choose the source rank deterministically (same seed -> same rp on all ranks).
+  if (rank == KMEANS_COMM_ROOT) {
+    std::mt19937 gen(params.rng_state.seed);
+    std::uniform_int_distribution<int> rank_dist(0, num_ranks - 1);
```
dantegd: To support empty ranks (fewer rows than GPUs, uneven partitions), `rp` can land on a rank with `n_local == 0`. That rank then hits the `RAFT_EXPECTS(n_local > 0, ...)` below and throws, while the other ranks go on to `comms.bcast(initialCentroid, ..., rp)` and block on a collective that never completes. Could we restrict this draw to ranks where `rank_counts[r] > 0` so the source is guaranteed to have data? What do you think?

cpp/src/cluster/detail/kmeans_common.cuh
```diff
@@ -180,7 +180,6 @@ void weightSum(
     for (IndexT i = 0; i < n_samples; ++i) {
       wt_sum += weight(i);
     }
-    RAFT_EXPECTS(wt_sum > DataT{0}, "invalid parameter (sum of sample weights must be positive)");
```
dantegd, line 183: Removing this makes sense for the MG case. But weightSum is also used by the single-GPU host path, which leaned on this to reject all-zero/negative weights.  Without it that path will divide by the sum and emit NaN/Inf centroids rather than a clear error. Could we move the positivity guard up into the single GPU caller so we keep the validation there? What do you think?

cpp/src/cluster/detail/kmeans_mg.cuh
```diff
@@ -2,751 +2,633 @@
...
+  auto d_prior_cost = raft::make_device_scalar<DataT>(dev_res, DataT{0});
+  auto d_done_flag  = raft::make_device_scalar<int>(dev_res, 0);
+  auto h_done_flag  = raft::make_host_scalar<int>(0);
+  auto h_norm_cache = raft::make_host_vector<DataT, IndexT>(!data_on_device ? n_local : IndexT{0});
+  auto d_norms =
+    raft::make_device_vector<DataT, IndexT>(dev_res, data_on_device ? n_local : IndexT{0});
+  bool norms_cached = false;
+
+  const DataT* d_weight_scale_ptr = d_weight_scale.data_handle();
+  auto prepare_batch_weights      = [&](size_t part_idx, IndexT batch_offset, IndexT cur_batch_size)
```
dantegd: `prepare_batch_weights` re-reads the source weights and re-applies weight_scale on every Lloyd iteration and every batch. For host streaming that's unavoidable, but for device partitions (where `streaming_batch_size` is forced to `max_part_rows` and the data is resident) the old kmeans_mg_batched.cuh precomputed scaled_weights once. Not a correctness issue, just redundant work each iteration, might be worth scaling device weights once up front. Not a blocker but would be a nice improvement.

cpp/include/cuvs/cluster/kmeans.hpp
```diff
@@ -1607,6 +1608,196 @@ void cluster_cost(
...
+ * @{
+ */
+
+/**
+ * @brief Multi-GPU k-means fit with one or more local data
+ *        partitions per rank.
+ *
+ * Each rank supplies its local training data as a vector of partitions. The
+ * implementation streams every partition through Lloyd iterations using
+ * `params.streaming_batch_size`.
```
dantegd: For device accessors streaming_batch_size is intentionally ignored (clamped to max_part_rows with a warning, right?). Worth a sentence in the public mg::fit docs clarifying that streaming_batch_size only applies to host partitions.

dantegd (changes requested):
Just had minor comments and suggestions, otherwise this looks great

cpp/include/cuvs/cluster/kmeans.hpp
tarang-jain: Fair enough! I'll add the guarding macro in the header itself.

tarang-jain: I am also thinking about removing the ::mg namespace and rather naming the API `cuvs::cluster::kmeans_multi_gpu` or renaming that namespace to `multi_gpu`. That makes it more explicit. I think abbreviations like MG, SNMG and MNMG are more internal, rather than user facing.

cpp/src/cluster/kmeans_impl.cuh
tarang-jain: I think a hard error would be extreme because a multi-gpu resource is also a valid single-gpu resource and running single GPU KMeans which such a handle is perfectly valid.
On the other hand, the old method of assuming the user wants to run on multiple GPUs just because they have comms initialized on the handle does not seem ideal. Having a clear new API and adding the migration to the API docs / release notes seems like the best route.

cpp/src/cluster/detail/kmeans_mg_distributed_init.cuh
tarang-jain: yes that makes perfect sense. I discussed this in the comments below with @viclafargue and updated the docs to mention that zero vectors on a rank are not allowed. But now I think if dask puts zero rows on a particular rank, it is beyond the user's control and perfectly valid use of the API. We should not be throwing in such a case.

cpp/src/cluster/detail/kmeans_common.cuh
tarang-jain: Ok I have added this. Added a parameter to that function, which conditionally does the check. We do the check for single gpu but avoid it for multi-gpu.

cpp/src/cluster/detail/kmeans_mg_distributed_init.cuh
tarang-jain: I have reverted the doc change and instead added a check to sample rp from only those ranks which have at least one row.

cpp/tests/CMakeLists.txt
tarang-jain: yes that makes sense. The only problem is that if the clique has exactly one GPU, the test will error out? If I change that line to `GPUS 2`, not sure if we want to allow that failure?

cpp/include/cuvs/cluster/kmeans.hpp
tarang-jain: I have updated the docs for streaming_batch_size in the kmeans params and also added a note in the multi-GPU kmeans API docs as well.

cpp/src/cluster/detail/kmeans_mg.cuh
tarang-jain: Done!

dantegd (approved):

cpp/src/cluster/detail/kmeans_mg.cuh
```diff
@@ -2,751 +2,666 @@
...
-  centroidsBuf.resize(initialCentroid.size(), stream);
-  raft::copy(handle,
-             raft::make_device_vector_view(centroidsBuf.begin(), initialCentroid.size()),
-             raft::make_device_vector_view(initialCentroid.data_handle(), initialCentroid.size()));
+  if (use_nccl && params.init == cuvs::cluster::kmeans::params::InitMethod::Array) {
+    raft::copy(rank_centroids.data_handle(),
+               centroids.data_handle(),
+               static_cast<size_t>(n_clusters) * n_features,
+               stream);
+  }
```
viclafargue, line 298: Sorry for changing my mind, but it looks like we may have over-optimized a bit. This looks like in the SNMG case array init performs a peer copy to each rank. Unfortunately this may not be as performant and reliable as a simple NCCL broadcast. This may possibly even not work on some machines.

Maybe we could expect Dask/Ray orchestration to provide similar centroids to each rank, and have the OMP version work with a NCCL broadcast?

cpp/src/cluster/detail/kmeans_mg_batched_init.cuh
```diff
@@ -149,86 +180,146 @@ std::vector<IndexT> broadcast_sampled_global_indices(raft::resources const& hand
...
+                                                    params,
+                                                    device_parts,
+                                                    n_features,
+                                                    centroids,
+                                                    workspace,
+                                                    rank_counts,
+                                                    global_n,
+                                                    rank,
+                                                    num_ranks,
+                                                    comms);
```
viclafargue, line 303: Does `initKMeansPlusPlus_distributed` handles `oversampling_factor == 0` correctly?

viclafargue (approved):
Great work! Pre-approving, just have two points that may need some attention.

cpp/src/cluster/detail/kmeans_mg_batched_init.cuh
tarang-jain: sampling probabilities effectively become zero during the iterations. As a result, it falls back to a truly random init. At first this makes sense because of the following reason:
The host side batched init runs regular kmeans++ on the init sample, but there is no sampling in the multi-GPU setup. If we were to do a truly sequential kmeans++ in this case, each sample would have to be taken from the global data across all GPUs one at a time. This is not sustainable, which is why the distributed algo was written in the first place. This was also not being taken care of before this PR.

Now one potential other solution is to clamp it to 1.0 in such a case. This is because, the probabilities effectively resolve to the regular (sequential) kmeans++ by removing the effect of the oversampling factor. We could potentially argue that the minimum oversampling factor should be 1.0 and not 0.0 for multi-GPU. This way there could still be a good amount of randomness because we are just relying on the distances for the sampling, so we still might not get a good amount of candidates. But this is still a better (i.e. more informed) initialization than completely random.

cpp/src/cluster/detail/kmeans_mg.cuh
tarang-jain: I added a new bcast overload that allows distinct send and receive buffers since both -- ncclBroadcast and raft comms support such a behavior. Now there are two places where we avoid copying to intermediate buffers on the sending node and instead broadcast directly.

cpp/src/cluster/detail/kmeans_mg_batched_init.cuh
tarang-jain: I have clamped all oversampling_factor < 1.0 cases to 1.0. Added a note in the code. Let me know what you think @viclafargue

cpp/src/cluster/detail/kmeans_mg_distributed_init.cuh
```diff
@@ -0,0 +1,426 @@
...
+    RAFT_EXPECTS(n_local > 0,
+                 "selected source rank %d has no local rows; cannot pick an initial centroid",
+                 rp);
+    std::mt19937 row_gen(params.rng_state.seed);
+    std::uniform_int_distribution<IndexT> row_dist(IndexT{0}, n_local - 1);
+    chosen_local_idx             = row_dist(row_gen);
+    auto [part_idx, row_in_part] = locate_local_row(part_offsets, chosen_local_idx);
+    auto const* src_row          = X_parts[part_idx].data_handle() + row_in_part * n_features;
+    // Step 1.3 - broadcast the chosen initial centroid to all ranks.
+    comms.bcast(src_row, centroidsBuf.data(), static_cast<size_t>(n_features), rp);
```
viclafargue: Broadcast is collective, so every rank must call it. It cannot be inside inside `if (rank == rp)`.

cpp/src/cluster/detail/kmeans_mg_batched_init.cuh
viclafargue: Sound good for my side. Even though I would be interested in knowing what @dantegd thinks about this.

cpp/src/cluster/detail/kmeans_mg_distributed_init.cuh
tarang-jain: Good catch! Fixed it.

cpp/src/cluster/detail/kmeans_mg_batched_init.cuh
dantegd: This makes sense to me , since `oversampling_factor == 0` cannot mean exact sequential k-means++ in this distributed KMeans|| path, and clamping to avoids silently degenerating into random fill-in, which is a very nice improvement, great to see!

T he main remaining semantic difference is that values in `(0, 1)` are also promoted to `1.0`, but I think that's fine as well, so the `1.0` is the practical lower bound for distributed initialization quality. The code comment makes that distinction clear enough from the single-GPU path. Is there a comment also in docstrings/public docs? 

A small test with `oversampling_factor=0` on the distributed path could be useful to lock in this behavior.

viclafargue (approved):

cpp/src/cluster/detail/kmeans_mg_batched_init.cuh
tarang-jain: Added the docstrings and a cpp test!

cpp/src/cluster/detail/kmeans_mg_batched_init.cuh
```diff
@@ -19,13 +22,42 @@
...
+    bool isRoot = true;                                      \
+    if (raft::resource::comms_initialized(handle)) {         \
+      const auto& comm  = raft::resource::get_comms(handle); \
+      const int my_rank = comm.get_rank();                   \
+      isRoot            = my_rank == 0;                      \
+    }                                                        \
+    if (isRoot) { RAFT_LOG_DEBUG(fmt, ##__VA_ARGS__); }      \
+  } while (0)
+
+#define KMEANS_COMM_ROOT 0
```
lowener:
```suggestion
#define CUVS_KMEANS_COMM_ROOT 0
```

cpp/src/cluster/detail/kmeans_mg_batched_init.cuh
```diff
@@ -149,86 +180,146 @@ std::vector<IndexT> broadcast_sampled_global_indices(raft::resources const& hand
...
 }
 
-template <typename DataT, typename IndexT>
-raft::device_matrix<DataT, IndexT> sample_global_host_rows(
+template <typename IndexT>
+std::pair<std::size_t, IndexT> locate_local_row(const std::vector<IndexT>& part_offsets,
+                                                IndexT local_idx)
+{
+  auto it       = std::upper_bound(part_offsets.begin(), part_offsets.end(), local_idx);
+  auto part_idx = static_cast<std::size_t>(std::distance(part_offsets.begin(), it) - 1);
```
lowener, line 188: Isn't there a risk of overflow if `local_idx==0`? Then the cast to size_t becomes max_uint64? Or is that a situation that would never happen?

cpp/src/cluster/detail/kmeans_mg_distributed_init.cuh
```diff
@@ -0,0 +1,428 @@
...
+  }
+
+  raft::random::RngState rng(params.rng_state.seed, raft::random::GeneratorType::GenPhilox);
+
+  auto d_rp = raft::make_device_scalar<int>(handle, 0);
+  int rp    = 0;
+
+  // Step 1.1 - choose the source rank deterministically (same seed -> same rp on all ranks).
+  if (rank == KMEANS_COMM_ROOT) {
+    // Only draw from ranks that actually own at least one row to avoid an
```
lowener: Unfinished comment

cpp/src/cluster/detail/kmeans_mg_distributed_init.cuh
```diff
@@ -0,0 +1,428 @@
...
+    auto norm_slice = raft::make_device_vector_view<DataT, IndexT>(
+      L2NormX.data_handle() + part_offsets[p], part_rows);
+    raft::linalg::norm<raft::linalg::L2Norm, raft::Apply::ALONG_ROWS>(handle, x_slice, norm_slice);
+  }
+
+  // Computes the global cluster cost (psi) by iterating over every local part,
+  // computing the per-row min cluster distance against the current candidate
+  // set, reducing locally and then allreducing across ranks.
+  auto d_partial           = raft::make_device_scalar<DataT>(handle, DataT{0});
+  auto compute_global_cost = [&]() -> DataT {
```
lowener: This lambda is quite long, I think it deserves its own function

cpp/src/cluster/detail/kmeans_mg_distributed_init.cuh
```diff
@@ -0,0 +1,428 @@
...
+  // Step 2: psi <- phi_X(C).
+  DataT psi = compute_global_cost();
+
+  const int niter = std::min(8, static_cast<int>(std::ceil(std::log(psi))));
+  RAFT_LOG_DEBUG(
+    "Distributed KMeans||: rank=%d, psi=%g, niter=%d", rank, static_cast<double>(psi), niter);
+
+  // Steps 3-6: sample candidates `O(log psi)` times, gathering across ranks.
+  for (int iter = 0; iter < niter; ++iter) {
+    psi = compute_global_cost();
```
lowener: Skip it on first iteration, it was computed just before

cpp/CMakeLists.txt
```diff
@@ -1118,8 +1118,6 @@ if(NOT BUILD_CPU_ONLY)
     src/cluster/detail/minClusterDistanceCompute.cu
     src/cluster/agglomerative.cu
     src/cluster/kmeans_cluster_cost.cu
-    src/cluster/kmeans_fit_mg_float.cu
-    src/cluster/kmeans_fit_mg_double.cu
```
lowener, line 1298: I would be in favor of keeping them here with all the other source file, but switching to 
`$<$<BOOL:${BUILD_MG_ALGOS}>:src/cluster/kmeans_fit_mg_float.cu>`
`$<$<BOOL:${BUILD_MG_ALGOS}>:src/cluster/kmeans_fit_mg_double.cu>`

cpp/src/cluster/detail/kmeans_mg_batched_init.cuh
tarang-jain: If `local_idx` is zero, The cast becomes 0, right? part_offsets will always start with 1. The `std::distance` would be 1.

lowener (approved):
LGTM

dantegd (approved):
Recent changes look good to me. Should we open an issue to track whether to check the CTEST registration of other MG algos?

conda/recipes/libcuvs/recipe.yaml
```diff
@@ -293,6 +293,9 @@ outputs:
         - librmm =${{ minor_version }}
         - libraft-headers =${{ minor_version }}
         - nccl ${{ nccl_version }}
+        # ucx/ucxx are linked by the multi-GPU (MG) tests
+        - libucxx ${{ ucxx_version }}
+        - ucx
```
bdice: These are needed at build time, right? We also need to list these under the cached host env, under line 74.

bdice (approved):

conda/recipes/libcuvs/recipe.yaml
tarang-jain: > We also need to list these under the cached host env, under line 74.

@divyegala this is the place from where I removed it, right? Should I add it again?

bdice (changes requested):
Ah, this is an entirely new dependency on ucxx. We have several other tasks to handle.

https://github.com/NVIDIA/cuvs/blob/main/ci/release/update-version.sh needs changes [like raft has](https://github.com/NVIDIA/raft/blob/main/ci/release/update-version.sh).

Also reference how RAFT uses `UCXX_VERSION` and `UCXX_BRANCH` throughout the repo, we'll need to replicate all that infrastructure.

tarang-jain:
/ok to test https://github.com/NVIDIA/cuvs/commit/ccd5e432deab509cb76b9905298001eb1e889953

cpp/cmake/thirdparty/get_raft.cmake
```diff
@@ -62,7 +62,7 @@ endfunction()
 find_and_configure_raft(VERSION  ${RAFT_VERSION}.00
         FORK                     ${RAFT_FORK}
         PINNED_TAG               ${RAFT_PINNED_TAG}
-        ENABLE_MNMG_DEPENDENCIES OFF
+	    ENABLE_MNMG_DEPENDENCIES ${BUILD_MG_ALGOS}
```
KyleFromNVIDIA:
```suggestion
        ENABLE_MNMG_DEPENDENCIES ${BUILD_MG_ALGOS}
```

KyleFromNVIDIA (changes requested):

cpp/tests/CMakeLists.txt
```diff
@@ -332,6 +332,14 @@ if(BUILD_CAGRA_HNSWLIB)
 endif()
 
 if(BUILD_MG_ALGOS)
+  find_package(Python COMPONENTS Interpreter)
+  if(Python_FOUND)
+    include(${rapids-cmake-dir}/cython-core/find_prefix_paths.cmake)
+    rapids_cython_find_prefix_paths("${Python_EXECUTABLE}" _cuvs_wheel_prefix_paths)
+    set_property(GLOBAL PROPERTY FIND_LIBRARY_USE_LIB64_PATHS TRUE)
```
bdice: I don’t fully understand what this code path is aiming to do, but conda doesn’t use lib64 paths iirc. Normally I would assume this is used in wheels, but we don’t build C++ tests for wheels. What’s going on here?

cpp/tests/CMakeLists.txt
bdice: Please mark all multi-GPU tests as SERIAL. You probably don’t want to run two 2-GPU tests concurrently on a 4-GPU system. It just gets messy. We recently discussed this topic for cudf and that was the right choice. https://github.com/rapidsai/cudf/pull/23188

cpp/tests/CMakeLists.txt
tarang-jain: I had done that to ensure that if those dependencies are coming from pip, they are found. Because find_package was failing if I get pylibraft from pip. I agree that we no longer do this in our conda builds, so it is not needed

cpp/tests/CMakeLists.txt
bdice: You might remove `GPUS 1` to reduce confusion, since this is multi-GPU. Marking it serial is enough.

bdice (approved):

viclafargue (approved):

viclafargue (changes requested):
The SNMG batched KMeans C/Python API was merged before and unfortunately now [calls the single-GPU path](https://github.com/NVIDIA/cuvs/blob/fd96a5d6ce7d9a97fb44e63e1e712199355781fc/c/src/cluster/mg_kmeans.cpp#L155) after this PR. We should fix this.

viclafargue (approved):
Thanks a lot!

lowener (approved):

cpp/src/cluster/detail/kmeans_mg_distributed_init.cuh
```diff
@@ -0,0 +1,490 @@
...
+    auto n_iter_out  = raft::make_host_scalar<IndexT>(0);
+
+    cuvs::cluster::kmeans::params default_params;
+    cuvs::cluster::kmeans::params recluster_params = params;
+    recluster_params.rng_state                     = default_params.rng_state;
+    recluster_params.init   = cuvs::cluster::kmeans::params::InitMethod::Array;
+    recluster_params.n_init = 1;
+
+    auto weight_opt = std::make_optional(raft::make_const_mdspan(weight.view()));
+    cuvs::cluster::kmeans::fit(handle,
```
viclafargue, line 444:
```suggestion
    cuvs::cluster::kmeans::detail::kmeans_fit(handle,
```

Worker needlessly emits the following warning :
```
Multi-GPU handle detected on single-GPU kmeans::fit() entry; falling back to single-GPU. Use cuvs::cluster::kmeans::mg::fit(...) for multi-GPU.
```

We need to fix this.

viclafargue (changes requested):

KyleFromNVIDIA (approved):

cpp/src/cluster/detail/kmeans_mg_distributed_init.cuh
tarang-jain: I don't think there is anything wrong with that warning. Yes it is triggered unnecessarily here by our own initialization implementation, but I am not in the favor of calling the detail namespace API. We should stick to calling public functions wherever possible.

viclafargue: From DM discussions, it looks like calling `cuvs::cluster::kmeans::detail::kmeans_fit` directly may unnecessarily increase binary size. We should maybe come back at this later since the warning is a bit odd. But I will approve the PR for now to get the PR unstucked.

viclafargue (approved):

cpp/include/cuvs/cluster/kmeans.hpp
cjnolet: Why are we declaring a separate public namespace for mg at all? The whole idea behind consolidating namespaces was for the user to have an easier experience by configuring the appropriate resources type to determine if they want mg operation. Please don't create separate namespaces for mg in the public APIs.

Victor did thag originally and I asked him to remove it.

cjnolet (changes requested):

cpp/include/cuvs/cluster/kmeans.hpp
tarang-jain: https://github.com/NVIDIA/cuvs/pull/2066#discussion_r3389536811 was the discussion about this.

tarang-jain: Since the multi GPU implementation brings in a new API which takes a vector of mdspans and we dont want that API for a single GPU resource.

cpp/include/cuvs/cluster/kmeans.hpp
tarang-jain: @viclafargue I brought back this dispatch. So https://github.com/NVIDIA/cuvs/issues/2319 is not needed anymore.

cjnolet (approved):

PR #2067, opened 2026-05-07
[Docs] Convert Sphinx docs to Fern

cjnolet:
This PR converts the current Sphinx docs to Fern in preparation for the move to docs.nvidia.com. Instead of manually composing the API reference docs, this PR also generates API reference docs for all supported languages directly from the code (as is standard in Fern). 

There's a lot of files in this PR, and most of the markdown files are either copied over to the Fern directory format from the old Sphinx docs, or they've been auto-generated using the new API reference docs generation scripts (`generate_api_reference.py` in the changes). When reviewing this PR, it's probably better to start with the non-markdown files, then build the docs and run them locally. 

The docs can be built in the usual way with `./build.sh docs`. You can run them locally using the following command:

```
fern/build_docs.sh dev --port 3000 --backend-port 3001
```


ci/build_docs.sh
```diff
@@ -35,27 +26,5 @@ set -eu
...
-pushd docs
-make dirhtml
-mv ../rust/target/doc ./build/dirhtml/_static/rust
-mkdir -p "${RAPIDS_DOCS_DIR}/cuvs/"html
-mv build/dirhtml/* "${RAPIDS_DOCS_DIR}/cuvs/html"
-popd
-
-RAPIDS_VERSION_NUMBER="${RAPIDS_VERSION_MAJOR_MINOR}" rapids-upload-docs
+rapids-logger "Validate Fern docs"
+fern/build_docs.sh "${FERN_DOCS_MODE:-check}"
```
bdice: Does `rapids-upload-docs` not work anymore? That was used for uploading HTML docs previews. We should restore that behavior.

bdice (approved):
Approving to unblock (since we're near burndown). My one request is that we should try to upload the generated HTML to our preview server as before.

ci/build_docs.sh
bdice: I'm trying to propose a fix here... Does fern not work offline at all?

cjnolet: Discussed offline, but I think developers can use `fern dev --preview` locally and CI can use `fern generate --docs --preview --id <PR #>`. I think the latter will allow CI to stage every PR for us- which I think is awesome, because we don't even need to be on the VPN to view them.

fern/pages/index.md
```diff
@@ -1,36 +1,30 @@
...
 
-
-
-What is cuVS?
-#############
+## What is cuVS?
 
 cuVS contains state-of-the-art implementations of several algorithms for running approximate and exact nearest neighbors and clustering on the GPU. It can be used directly or through the various databases and other libraries that have integrated it. The primary goal of cuVS is to simplify the use of GPUs for vector similarity search and clustering.
 
 Vector search is an information retrieval method that has been growing in popularity over the past few  years, partly because of the rising importance of multimedia embeddings created from unstructured data and the need to perform semantic search on the embeddings to find items which are semantically similar to each other.
```
mgrauer: I prefer evergreen content in docs, so you don't have to trim it out later. 

Specifically the "growing in popularity" part here

fern/pages/index.md
```diff
@@ -65,28 +56,19 @@ There are several benefits to using cuVS and GPUs for vector search, including
...
 
-.. image:: ../../img/tech_stack.png
-  :width: 600
-  :alt: cuVS is built on top of low-level CUDA libraries and provides many important routines that enable vector search and clustering on the GPU
-
-
-
-Contents
-########
+<img alt="cuVS is built on top of low-level CUDA libraries and provides many important routines that enable vector search and clustering on the GPU" src="/assets/images/tech_stack.png" />
```
mgrauer: Consider aligning the content in the alt text with the text itself

text
>cuVS is built on top of the RAPIDS RAFT library of high performance machine learning primitives and provides all the necessary routines for vector search and clustering on the GPU.

alt text
>cuVS is built on top of low-level CUDA libraries and provides many important routines that enable vector search and clustering on the GPU

fern/pages/index.md
```diff
@@ -40,18 +34,15 @@ Semantic search
 - Molecular search
 - Model training
```
mgrauer: Model training feels out of place here. Is there a way to make this more specific?

fern/pages/index.md
mgrauer: Parameter tuning feels out of place here. Not clear to me how this is a benefit of cuvs + GPU, compared to the others in this list. Can this be rephrased to show a benefit or else removed?

fern/pages/index.md
```diff
@@ -65,28 +56,19 @@ There are several benefits to using cuVS and GPUs for vector search, including
 
 In addition to the items above, cuVS shoulders the responsibility of keeping non-trivial accelerated code up to date as new NVIDIA architectures and CUDA versions are released. This provides a delightful development experience, guaranteeing that any libraries, databases, or applications built on top of it will always be receiving the best performance and scale.
```
mgrauer: This is a big deal and well stated! I wouldn't have realized how big a deal this was a few months ago. 

I know this isn't actionable, but it stuck out to be as super important.

mgrauer (commented):
review of index page

fern/pages/index.md
cjnolet: This is intentionally vage, but i can make it more specific- semantic deduplication during data curation is often done prior to model training, for example. Another example is the use of ANN fused within the actual layers of an LLM or transformer. Nearest neighbors is used all over the place, in part because we most often care about the closest similarities/furthest distances, and everything else becomes noise. 

I can make the list more explicit- just trying to avoid having a long laundry list so I opted to summarize. Perhaps I'll link to the use-cases guide here.

fern/pages/index.md
cjnolet: This comment doesn't seem to be attached to anything, but I'll try to respond anyways (and I can link to our tuning docs). Parameter tuning for vector search is multi-objective, which is a bit different than traditional ML. Users care about not only the quality, but also the speed (both in terms of latency/throughput and build time) and sometimes even the memory footprint. HPO techniques on the GPU can help significantly reduce the amount of time spent tuning. We offer a "tune mode" in cuvs-bench, for example, which underneath uses optuna. Most end-users of databases don't want to have to think about tuning parameters, so database developers prefer to make capaiblities like parameter tuning as simple as possible the user (for e.g. specify latency target and target recall level and have the GPU learn the params).

fern/pages/choosing_and_configuring_indexes.md
```diff
@@ -0,0 +1,94 @@
...
+
+Vector search indexes trade build time, search speed, memory use, and recall. Exact indexes compare every vector and return the true nearest neighbors, but they become expensive as datasets grow. Approximate indexes search a smaller candidate set, which is faster, but may miss some exact neighbors.
+
+In vector search, recall is the main quality metric. It measures how many of the exact nearest neighbors were returned by the approximate search. Higher recall usually costs more build time, more search time, more memory, or some combination of all three.
+
+This page introduces the main index families, how quantization and reranking change the tradeoffs, and how IVF-style partitioning differs from graph-based search. By the end, you should have a practical starting point for choosing an index type and knowing which parameters matter first.
+
+## Start with the workload
+
+The best index depends mostly on dataset size, vector dimensionality, recall target, and whether build time or query performance matters more.
```
mgrauer: This table is great, but is there an unstated assumption, something like GPU for index and CPU for search?

Specifically it relates to Vamana
>when the full dataset is too large to keep comfortably in memory.

is this GPU memory or host memory? It got me thinking that if we assume a certain usage pattern as part of this guidance we should be explicit

fern/pages/choosing_and_configuring_indexes.md
```diff
@@ -0,0 +1,94 @@
...
+
+| Workload | Good starting point |
+| --- | --- |
+| Tiny datasets, under 100K vectors | Use [brute-force](/user-guide/api-guides/indexing-guide/brute-force) or CPU [HNSW](/user-guide/api-guides/indexing-guide/cagra#interoperability-with-hnsw). A GPU index may not provide enough benefit to justify the extra complexity. |
```
mgrauer: For the two rows where we call out vector counts (100K, 1M) - does the guidance change depending on how large the vectors are?

fern/pages/choosing_and_configuring_indexes.md
```diff
@@ -0,0 +1,94 @@
...
+| Large datasets with fast ingest needs | Use [IVF-Flat](/user-guide/api-guides/indexing-guide/ivf-flat), [IVF-SQ](/user-guide/api-guides/indexing-guide/ivf-flat#ivf-sq-and-scalar-quantization), or [IVF-PQ](/user-guide/api-guides/indexing-guide/ivf-pq). These indexes partition the data and let you tune recall by searching more or fewer partitions. |
+| Large datasets with high recall needs | Use [CAGRA](/user-guide/api-guides/indexing-guide/cagra) on GPU or [HNSW](/user-guide/api-guides/indexing-guide/cagra#interoperability-with-hnsw) on CPU. Graph indexes usually provide strong search quality, but take longer to build than IVF indexes. |
+| Very large or disk-backed datasets | Use [Vamana/DiskANN](/user-guide/api-guides/indexing-guide/vamana) when the full dataset is too large to keep comfortably in memory. |
+
+## Common index choices
+
+| Algorithm | Build complexity | Search performance | What makes it unique |
```
mgrauer: Why use "Build complexity"? Is that like big O complexity (in which case the actual values feel inexact and I'd think "speed" would be better), or complexity of implementation (and if so, why would the user care, that complexity is borne by the library maintainers), or complexity of usage?

Seems like an odd measure to me.



After some time away, I'm wondering if the "what makes it unique" is helpful here. It some implementation detail and the start of this page is "what is vector search". So a user new to vector search could really get bogged down in that without it helping them.

I really like the "when to use each" section at the bottom, but I think it should be hoisted much higher in the page. I'd suggest replacing the "what makes it unique" column here with that "when to use it" content. 

I'm not sure yet when I'd actually want to know the "what makes it unique" content.

fern/pages/choosing_and_configuring_indexes.md
```diff
@@ -0,0 +1,94 @@
+# What is Vector Search?
+
+Vector search indexes trade build time, search speed, memory use, and recall. Exact indexes compare every vector and return the true nearest neighbors, but they become expensive as datasets grow. Approximate indexes search a smaller candidate set, which is faster, but may miss some exact neighbors.
+
+In vector search, recall is the main quality metric. It measures how many of the exact nearest neighbors were returned by the approximate search. Higher recall usually costs more build time, more search time, more memory, or some combination of all three.
+
+This page introduces the main index families, how quantization and reranking change the tradeoffs, and how IVF-style partitioning differs from graph-based search. By the end, you should have a practical starting point for choosing an index type and knowing which parameters matter first.
```
mgrauer: which parameters matter first => matter most

or maybe you are trying to say something else?

fern/pages/choosing_and_configuring_indexes.md
```diff
@@ -0,0 +1,94 @@
...
+| [IVF-PQ](/user-guide/api-guides/indexing-guide/ivf-pq) | Medium-high | Very memory-efficient; search speed depends on compression, probing, and refinement | IVF with product quantization. Splits vectors into subvectors and stores compact codes, enabling much smaller indexes with a larger accuracy tradeoff than IVF-SQ. |
+| [ScaNN](/user-guide/api-guides/indexing-guide/sca-nn) | Medium-high | Strong recall and speed tradeoff | Combines partitioning, quantization, and reranking to prune the search space while preserving result quality. |
+| [Vamana/DiskANN](/user-guide/api-guides/indexing-guide/vamana) | High | Excellent at very large scale, including SSD-backed search | Builds a graph designed for large or disk-backed indexes, especially when the full dataset cannot fit comfortably in memory. |
+
+## Quantization
+
+Quantization compresses vectors so they use less memory and require less bandwidth during search. It can be used with many index types, not just IVF. The main tradeoff is that compression usually improves speed and memory efficiency at the cost of some recall.
```
mgrauer: We've mentioned IVF multiple times already via variants (IVF-PQ, -SQ, etc) and this is the first mention on the page of IVF itself, but we haven't actually defined IVF anywhere.

fern/pages/choosing_and_configuring_indexes.md
```diff
@@ -0,0 +1,94 @@
...
+
+Scalar quantization, used by IVF-SQ, compresses each vector value into a smaller representation. Product quantization, used by IVF-PQ, splits each vector into smaller subvectors and represents each subvector with a compact code. Product quantization usually provides stronger compression than scalar quantization, but distance estimates become more approximate.
```
mgrauer: >Product quantization usually provides stronger compression than scalar quantization, but distance estimates become more approximate.

can you add one more phrase/sentence that connects this explanation into terms in the user domain (search) like recall/latency?

fern/pages/choosing_and_configuring_indexes.md
mgrauer: aha - this is the first mention of IVF, but still doesn't explain what it is

fern/pages/choosing_and_configuring_indexes.md
mgrauer: note there's actually a mention above, but not yet a definition

fern/pages/choosing_and_configuring_indexes.md
```diff
@@ -0,0 +1,94 @@
...
+
+Graph-based indexes can often be built directly over quantized vectors. In that case, the compressed representation is part of graph construction and search.
```
mgrauer: I'm not following exactly what the second sentence is trying to say, that is different from the first sentence.

mgrauer: hmmm, having read the next part about IVF indexes, and how they can use local compression, is the second sentence here trying to provide a contrast? like do compressed representations in graphs not include any local compression?

fern/pages/choosing_and_configuring_indexes.md
```diff
@@ -0,0 +1,94 @@
...
+
+IVF indexes usually apply quantization inside the partitioned structure. Because IVF first divides the dataset into coarse partitions, each partition can be compressed locally. This can improve the quality of the compressed representation while still reducing memory traffic during search.
```
mgrauer: Quite clearly explained!

fern/pages/choosing_and_configuring_indexes.md
```diff
@@ -0,0 +1,94 @@
...
+
+## Refinement and reranking
+
+Refinement, often called reranking, is a way to recover accuracy after an approximate or compressed search. The index first uses a fast representation, such as quantized vectors, to find a candidate set. Then it recomputes distances for those candidates using a more accurate representation, often the original full-precision vectors.
```
mgrauer: all makes sense except the exact relationship b/w an approximate and compressed search here.

are approximate/compressed the same? i think by definition search over compressed vectors is approximate, so compressed is a subset of approximate

maybe "is a way to recover accuracy after an approximate search over compressed vectors." ?

fern/pages/choosing_and_configuring_indexes.md
```diff
@@ -0,0 +1,94 @@
...
+
+This is especially useful with quantized indexes. Quantization makes search faster and smaller, but compressed distances are approximate. Reranking lets the system use quantized vectors to quickly narrow the search, then use full-precision vectors to choose the final nearest neighbors.
```
mgrauer: is it correct to say the "compressed distances" are approximate or "distances between compressed vectors" ?

fern/pages/choosing_and_configuring_indexes.md
```diff
@@ -0,0 +1,94 @@
...
+
+For example, an [IVF-SQ](/user-guide/api-guides/indexing-guide/ivf-flat#ivf-sq-and-scalar-quantization) or [IVF-PQ](/user-guide/api-guides/indexing-guide/ivf-pq) index may use compressed vectors to scan selected partitions quickly. After it finds the best candidate IDs, a refinement step can load the original vectors for those candidates and compute exact distances. This usually improves recall while keeping most of the speed and memory benefits of quantized search.
```
mgrauer: the curious reader in me wants to know how much i lose by searching on compressed vectors and how much i recover through refinement, but that is probably too in the weeds for this page.

fern/pages/choosing_and_configuring_indexes.md
mgrauer: i will note, if you do trim out the "what makes it unique" then that might lose the only place where ivf-flat is defined, and possibly other unique pieces of info on this page.

fern/pages/choosing_and_configuring_indexes.md
```diff
@@ -0,0 +1,94 @@
...
+
+[ScaNN](/user-guide/api-guides/indexing-guide/sca-nn) also relies heavily on this idea: it prunes the dataset quickly using partitioning and quantization, then reranks a smaller candidate set to improve final result quality.
+
+## IVF vs. graph-based indexes
```
mgrauer: i'd argue this section should come much higher in the page, as this this knowledge is a helpful scaffold for most other info on the page.

This is the first time that IVF comes close to a definition on the page, though I'd still like to know what the letters mean and why it is called that. 

this section also (other than the last paragraph) doesn't get too in to the weeds too soon.

fern/pages/choosing_and_configuring_indexes.md
```diff
@@ -0,0 +1,94 @@
...
+
+<img alt="A simple comparison of IVF-Flat and graph-based vector indexes, showing how each index is built and searched." src="/assets/images/ivf_vs_graph_index.png" />
+
+Graph-based indexes scale by connecting nearby vectors into a navigation structure. A query moves through the graph toward better candidates rather than scanning partitions. [HNSW](/user-guide/api-guides/indexing-guide/cagra#interoperability-with-hnsw) is a strong CPU graph index, [CAGRA](/user-guide/api-guides/indexing-guide/cagra) is optimized for GPU graph search and fast GPU graph construction, and [Vamana/DiskANN](/user-guide/api-guides/indexing-guide/vamana) is designed for very large or disk-backed graph search.
+
+The practical difference is that IVF narrows search by choosing partitions, while graph methods narrow search by walking neighbor links. IVF tends to have simpler scaling behavior and can be easier to distribute by partitions. Graph indexes often provide excellent recall and latency tradeoffs, but their build process and memory layout can be more complex.
```
mgrauer: simpler scaling behavior
build process and memory layout can be more complex

similar to my comment above in the table about complexity.

what does simpler scaling behavior mean and is that relevant to the user? perhaps b/c the scaling behavior is easier to understand?

does the user care that the build process and memory layout can be more complex for some reason? could be, e.g. maybe they are harder to tune or debug? but if this is primarily an implementor's concern, consider whether this is the right place for it.

probably good to consider overall for this page - who is the audience? i'll repeat this comment at the top

fern/pages/choosing_and_configuring_indexes.md
```diff
@@ -0,0 +1,94 @@
+# What is Vector Search?
```
mgrauer: probably good to consider overall for this page - who is the specific audience?

fern/pages/choosing_and_configuring_indexes.md
```diff
@@ -0,0 +1,94 @@
...
+
+The practical difference is that IVF narrows search by choosing partitions, while graph methods narrow search by walking neighbor links. IVF tends to have simpler scaling behavior and can be easier to distribute by partitions. Graph indexes often provide excellent recall and latency tradeoffs, but their build process and memory layout can be more complex.
+
+IVF and graph-based methods can also be combined. A hierarchical or graph-based structure can help assign vectors to IVF clusters faster during indexing, graph indexes can be stored inside individual IVF partitions to speed up search within each partition, and a tree-like or graph-like centroid structure can help keep partitions balanced over time as data drifts without rebuilding the entire index from scratch.
+
+## Tuning
```
mgrauer: this section seems like a jump in level of detail and discontinuous with the rest of the page in terms of  "about content" vs "how to" content. Might be intended for different audiences too.

consider putting this on some other page.

i also don't know that i have enough background after reading this page to know what to do with the info in this section. so it feels simultaneously like too much detail compared to the page and also doesn't have enough context for me to know how to engage.

e.g.  start with `n_lists = sqrt(n_vectors)` and sweep `n_probes`

is something i don't yet have context for

mgrauer: might be better in a tuning guide. seems like there could be a lot of content for just tuning.

fern/pages/choosing_and_configuring_indexes.md
```diff
@@ -0,0 +1,94 @@
...
+
+For IVF indexes, start with `n_lists = sqrt(n_vectors)` and sweep `n_probes` values such as 1%, 2%, 4%, 8%, and 16% of `n_lists`. Choose the smallest probe count that reaches the recall target.
+
+For quantized indexes, tune compression together with refinement. Stronger compression can reduce memory and bandwidth, but may need more probes, a larger candidate set, or reranking with original vectors to recover recall.
+
+For graph-based indexes, tune search breadth before increasing build complexity. If search-time tuning cannot reach the recall target within the latency budget, then increase graph quality, graph degree, or construction breadth.
+
+For multi-stage indexes such as [ScaNN](/user-guide/api-guides/indexing-guide/sca-nn), tune candidate count and reranking together. The goal is to keep the first stage broad enough to preserve recall and the reranking stage small enough to stay fast.
+
+## When to use each
```
mgrauer: good info, but i'd prefer it earlier on the page

fern/pages/choosing_and_configuring_indexes.md
```diff
@@ -0,0 +1,94 @@
...
+
+Use [brute-force](/user-guide/api-guides/indexing-guide/brute-force) for exact baselines, small datasets, or validation.
+
+Use [IVF-Flat](/user-guide/api-guides/indexing-guide/ivf-flat) when you want partition-based scaling while keeping full-precision vectors.
```
mgrauer: all the other entries in this section speak to the users language around search, but this one "when you want partition-based scaling" is almost tautological and also doesn't speak to users concerns. can you replace that with why someone would want partition based scaling?

mgrauer (commented):
choosing_and_configuring_indexes.md

it's going to take me forever to get through all the pages, so i wouldn't block this PR on me.

there are a lot of comments, structural and detailed both. i'd recommend skimming through all of them to get a sense of the biggest issues first, rather than trying to whack-a-mole.

fern/pages/index.md
mgrauer: it was in the section

> There are several benefits to using cuVS and GPUs for vector search, including
> 
> 1. Fast index build
> 2. Latency critical and high throughput search
> 3. Parameter tuning

fern/pages/choosing_and_configuring_indexes.md
cjnolet: > which parameters matter first => matter most

Thanks for pointing this out. Actually, `parameters` is being overloaded in this sentence. This guide isn't about parameters in the context of ML hyperparameters, it's about parameters in the context of benefits and tradeoffs. I'll reword this accordingly. The API guides are about the parameters.

fern/pages/choosing_and_configuring_indexes.md
cjnolet: Yeah, that's a good point. I've actually added a separate section dedicated to IVF. It's good we revisited this guide- I had originally had graph-based and IVF-based index designs separated, then I decided to combine them, but I think only comparing and contrasing is not doing a great job at conveying the main point. I've redone a lot of this page.

fern/pages/choosing_and_configuring_indexes.md
cjnolet: This page needs to provide the fundamental background and it should be comprehensive enough that they leave understanding the framework that vector search works inside of. I don't think enough was being presented in this page and there were some big gaps, as you've pointed out. 

The alternative scaling mechanism to IVF is actually hash-partitiond indexes. Graph-based indexes end up spilling to disk with DiskANN, but still end up scaling with hash-partitioning, ivf, or even both. It's important to drive that point home in this background doc.

fern/pages/choosing_and_configuring_indexes.md
cjnolet: > This table is great, but is there an unstated assumption, something like GPU for index and CPU for search?

The table is really not meant to imply anything about cuVS APIs or supported features outside of introducing that these index types exist. There are other pages intended to talk about design specifics and usage in the context of CPU and GPU. IN general, cuvs is not intended to be a CPU/GPU interop library. The fact that we support HNSWlib was a mistake- we put that out as an early experimental feature so early adopters could test it (the initial version required a small change to hnswlib, which hnswlib author was not going to accept). Then users started building on top of that and we couldn't easily remove it.

fern/pages/choosing_and_configuring_indexes.md
cjnolet: Correct, graph algorithms by themselves do not tend to provide local compression. 

Zoom out for a moment- IVF indexes are really a type of tree-based index, and you can essentially partition the vectors by their leaves in an IVF. You can partition a graph as well (using an IVF) or you can put a separate graph in an IVF partition. But each individual graph will not be able to benefit from localized compressed. This is why we will usually compress vectors and then build a graph on the comprsssed vectors directly, using refinement as needed.

The moral- graphs don't scale without some type of partitioning.

fern/pages/choosing_and_configuring_indexes.md
cjnolet: Well, so graphs don't partition, and compressed IVFs don't keep the raw vectors. Graphs could scale to disk without partitioning, for example.

I've redone this table to make it more clear.

fern/pages/choosing_and_configuring_indexes.md
cjnolet: We will have benchmarks in the getting started guide and will link them here.

robertmaynard (changes requested):
This needs to be rebased on 26.06 and not main if we are trying to make code freeze.

Currently it is bring in 26.08 containers into 26.06

cjnolet:
> This needs to be rebased on 26.06 and not main if we are trying to make code freeze. Currently it is bring in 26.08 containers into 26.06

@robertmaynard  Thank you for pointing this out. I just fixed it.

divyegala (approved):

dependencies.yaml
```diff
@@ -453,17 +453,7 @@ dependencies:
...
-          - graphviz
-          - ipython
-          - numpydoc
-          - recommonmark
-          - sphinx>=8.0.0
-          - sphinx-copybutton
-          - sphinx-markdown-tables
-          - pip:
-              - nvidia-sphinx-theme
+          - nodejs
```
robertmaynard: No minimum required version of nodejs?  I assume we want some minimum bound to avoid known CVE's in node

cjnolet: Ah thanks for noticing this. Fixed.

dependencies.yaml
```diff
@@ -453,7 +453,7 @@ dependencies:
     common:
       - output_types: [conda]
         packages:
-          - nodejs
+          - nodejs>=18
```
robertmaynard: We need nodejs of 22 or higher ( https://buildwithfern.com/learn/cli-api-reference/cli-reference/overview ).

cjnolet:
Just a heads up, CI is failing because 26.08 is mentioned in one of the commits while I was trying to revert changes from `main` back for 26.06. Going to need to rebase and force-push.

robertmaynard (approved):

PR #2073, opened 2026-05-12
Expose raft::memory_tracking_resources to C/Python/Java/Rust

### LINKED ISSUE #2982: [FEA] Python API for memory tracking resources
PR #2973 introduces a new `raft::memory_tracking_resources` that samples all resource-based memory usage:
```C++
  // optionally pass an existing resource handle
  raft::resources res;

  // The tracking handle is a child of resource handle; it wraps all memory resources with statistics adaptors
  raft::memory_tracking_resources tracked(res, "allocations.csv", std::chrono::milliseconds(1));

  // All allocations are logged to a .csv as long as `tracked` is alive
  cuvs::neighbors::cagra::build(tracked, ...);
```
It would be nice to have this feature exposed in the python API as well. The memory tracking resource handle can be exposed exactly the same way as the normal resource handle, so the implementation shouldn't be too complicated.

achirkin:
Expose the new `raft::memory_tracking_resources` in the public API of all supported languages and add documentation.

See [Raft #2973](https://github.com/rapidsai/raft/pull/2973) for more information about memory tracking resources.

Resolves https://github.com/rapidsai/raft/issues/2982


python/cuvs/cuvs/common/resources.pyx
```diff
@@ -50,10 +64,25 @@ cdef class Resources:
...
+                  memory_tracking_sample_interval_ms=10):
+        cdef bytes csv_bytes
+        if memory_tracking_csv_path:
+            csv_bytes = str(memory_tracking_csv_path).encode("utf-8")
+            check_cuvs(cuvsResourcesCreateWithMemoryTracking(
+                &self.c_obj,
+                csv_bytes,
+                <int64_t>memory_tracking_sample_interval_ms))
+        else:
+            check_cuvs(cuvsResourcesCreate(&self.c_obj))
```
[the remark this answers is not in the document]
achirkin: debatable

rust/cuvs/src/resources.rs
```diff
@@ -22,6 +25,40 @@ impl Resources {
...
+        sample_interval: Option<Duration>,
+    ) -> Result<Resources> {
+        let path_str = csv_path.as_ref().to_str().ok_or_else(|| {
+            Error::InvalidArgument(format!("csv_path is not valid UTF-8: {:?}", csv_path.as_ref()))
+        })?;
+        let c_path = CString::new(path_str).map_err(|e| {
+            Error::InvalidArgument(format!("csv_path contains an interior NUL byte: {}", e))
+        })?;
+        let sample_interval_ms =
+            sample_interval.unwrap_or(Duration::from_millis(10)).as_millis() as i64;
```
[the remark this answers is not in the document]
achirkin: I'm fine being limited by a few million years of sampling interval

docs/source/api_basics.rst
```diff
@@ -3,6 +3,7 @@ cuVS API Basics
 
 - `Memory management`_
 - `Resource management`_
+- `Memory tracking`_
```
cjnolet: The docs are about to change quite a bit. So that we don't hold up this, can you please break the docs portion into a separate PR?

rust/cuvs/src/resources.rs
```diff
@@ -22,6 +25,40 @@ impl Resources {
...
+    ///
+    /// The handle wraps all reachable memory resources (host, pinned, managed,
+    /// device, workspace, large_workspace) with allocation-tracking adaptors
+    /// and replaces the global host and device memory resources for the
+    /// lifetime of the handle. The CSV reporter is stopped and the global
+    /// memory resources are restored when the handle is dropped.
+    ///
+    /// `sample_interval` controls the minimum time between successive CSV
+    /// samples; when `None`, the C++ default of 10 ms is used.
+    pub fn with_memory_tracking<P: AsRef<Path>>(
```
cjnolet: @yan-zaretskiy can we get your review on these parts?

docs/source/api_basics.rst
achirkin: Ok, reverted the docs changes here

tfeher (approved):
C and Python changes LGTM

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/CuVSResourcesImpl.java
```diff
@@ -46,6 +49,45 @@ public CuVSResourcesImpl(Path tempDirectory) {
     }
   }
 
+  /**
+   * Constructor that allocates a tracking resources handle. All memory
+   * allocations made through this handle are written as CSV samples to
```
mythrocks, line 54: Nit:  Would it be worth clarifying here that this does not include the `cudaMallocHost` we do at line 87 here?
`hostBuffer.address()` calls `PinnedMemoryBuffer::createPinnedBuffer()`, which does a raw `cudaMallocHost` for 8MB.  This doesn't go through the memory-tracking infrastructure currently.

mythrocks (approved):
One minor documentation nit.  But otherwise, this should be good to ship.

rust/cuvs/src/resources.rs
```diff
@@ -22,6 +25,40 @@ impl Resources {
...
+        csv_path: P,
```
yan-zaretskiy: Since we don't refer to `P` in the function body, I slightly prefer this style:
```rust
pub fn with_memory_tracking(  
    csv_path: impl AsRef<Path>,
```

rust/cuvs/src/resources.rs
```diff
@@ -22,6 +25,40 @@ impl Resources {
...
+        sample_interval: Option<Duration>,
+    ) -> Result<Resources> {
+        let path_str = csv_path.as_ref().to_str().ok_or_else(|| {
+            Error::InvalidArgument(format!("csv_path is not valid UTF-8: {:?}", csv_path.as_ref()))
+        })?;
+        let c_path = CString::new(path_str).map_err(|e| {
+            Error::InvalidArgument(format!("csv_path contains an interior NUL byte: {}", e))
+        })?;
```
yan-zaretskiy: You can simplify this to
```rust
let c_path = CString::new(csv_path.as_ref().as_os_str().as_encoded_bytes())
    .map_err(|e| Error::InvalidArgument(format!("csv_path contains an interior NUL byte: {}", e)))?;
```

yan-zaretskiy (approved):
Minor suggestions only

achirkin:
Moving this 26.08 to let it merge after https://github.com/rapidsai/raft/pull/3004 , which now breaks the API that this PR uses

python/cuvs/cuvs/common/resources.pyx
```diff
@@ -1,16 +1,18 @@
...
 
 import functools
 
 from cuda.bindings.cyruntime cimport cudaStream_t
+from libc.stdint cimport int64_t
 
 from cuvs.common.c_api cimport (
     cuvsResources_t,
     cuvsResourcesCreate,
+    cuvsResourcesCreateWithMemoryTracking,
```
cjnolet, line 15: Being a little forward thinking here- what happens when we have a bunch of different resources types that we want to use together? If this was a constructor flag (or just an option flag) then it would hide the impl details from underneath.

achirkin: The C function `cuvsResourcesCreateWithMemoryTracking` produces the same pointer type as `cuvsResourcesCreate` (we expose all `raft::resources` sub-classes via the base class). So we can freely add more such factory functions (e.g. `cuvsResourcesCreateWithDryRun`) and use the produced pointers interchangeably.
I think this also helps to keep the C API/ABI stable, because we just add new free functions.

cjnolet: I'm tihnking about users that are going to want to use different types of resources together. These resources *should* all be types of "raft::resources" in the end, so that's why I'd like to see the different options exposed as flags instead of as explicit "make me this explicit resource type".

cjnolet: Ideally, a user should be able to mix and match the resouece types. The *only* reason we have convenience subclasses like "device_resources" and "device_resources_snmg" is for user convenience, but since they are still just raft::resources in the end, the user can still populate them using the lower-level resources functions.

achirkin: These *are* all types of `raft::resources` and *are* interchangeable, we just use different function names to create them. Honestly, I don't see how adding flags to differentiate between the functionality is better for a user. Memory-tracking resources is not something a user should regularly use in the production environment - it's an analysis tool. I'm afraid making this a function argument on the C side would only raise confusion, because on the C++ side the same object is created using a different class/constructor name rather than an argument.

cjnolet (approved):

PR #2076, opened 2026-05-12
Use L2 for IVF-PQ coarse clustering when the main metric is IP

yan-zaretskiy:
I ran 4 datasets with `cuvs-bench` to compare this change with the baseline. I pinned these parameters as well:

```
CAGRA graph_degree              = 64
CAGRA intermediate_graph_degree = 128
IVF-PQ internalDistanceDtype    = float
IVF-PQ smemLutDtype             = float
IVF-PQ coarse_search_dtype      = float
IVF-PQ refine_ratio             = 1
```

  | Dataset             | Vanilla IVF-PQ build | Vanilla recall@64 | L2-coarse IVF-PQ build | L2-coarse recall@64 | NN-Descent build | NN-Descent recall@64 |
  | ------------------- | -------------------: | ----------------: | ---------------------: | ------------------: | ---------------: | -------------------: |
  | gist-1m-ip          |               failed |                 - |               11.085 s |            0.965313 |         18.099 s |             0.507984 |
  | sift-1m-ip          |              6.539 s |          0.977297 |                5.911 s |            0.970359 |         11.462 s |             0.984250 |
  | deep-9990k-ip       |             54.887 s |          0.964438 |               38.429 s |            0.959828 |        103.170 s |             0.980578 |
  | text2image-10m-ip   |             62.840 s |          0.935516 |               45.933 s |            0.934091 |        115.616 s |             0.946979 |

The reported latency numbers are also either the same or slightly better. But I didn't run the full Pareto analysis.


cpp/src/neighbors/ivf_pq/ivf_pq_build.cuh
```diff
@@ -67,6 +67,14 @@ using namespace cuvs::spatial::knn::detail;  // NOLINT
 
 using internal_extents_t = int64_t;  // The default mdspan extent type used internally.
 
+inline cuvs::distance::DistanceType coarse_clustering_metric(
```
cjnolet, line 70: We shouldn't need to inline this right? It's not a templated function.

cjnolet (changes requested):

cpp/src/neighbors/ivf_pq/ivf_pq_build.cuh
yan-zaretskiy: We do need to `inline` it, cause it's a non-template namespace-scope function defined in a `.cuh` file and included by multiple TUs. Without it, it becomes an ODR violation and the linker will reject it.

cjnolet (approved):

yan-zaretskiy:
@cjnolet Just in case, I don't have the ability to merge.

PR #2080, opened 2026-05-14
doc: fix better the python api example

yihong0618:
This is a small change for python example that as other language examples
the `cagra.build`  first arg is `index_params` is better here for users can easily 
understand as  `build_params` is not init here


cjnolet (approved):

cjnolet:
Hi @yihong0618 unless the "recently updated" check is failing, please do not merge or hit the "Update Branch" button. That will cause CI to rerun unecessarily.

PR #2088, opened 2026-05-14
Fix segfault cuvs bench

### LINKED ISSUE #2087: [CI] Segfault in ann cuvs_bench binaries
This has been a longstanding flaky issue in CI

# Types of failures

1. https://github.com/rapidsai/cuvs/actions/runs/25753163294/job/75640409100#step:12:1526
rtx6000 pro, 13.2, **CUVS_IVF_FLAT_ANN_BENCH**
```
=========================== short test summary info ============================
FAILED tests/test_cli.py::test_run_command_creates_results - AssertionError: Run command failed with output:
  -- Using cuVS bench found in conda environment.
  -- Using cuVS bench found in conda environment.
  -- Using cuVS bench found in conda environment.
  -- Using cuVS bench found in conda environment.
  -- Using cuVS bench found in conda environment.
  Error occurred running benchmark: Command '['/opt/conda/envs/test/bin/ann/CUVS_IVF_FLAT_ANN_BENCH', '--build', '--data_prefix=/tmp/pytest-of-root/pytest-1/datasets0', '--benchmark_out_format=json', '--benchmark_counters_tabular=true', '--benchmark_out=/tmp/pytest-of-root/pytest-1/datasets0/test-data/result/build/cuvs_ivf_flat,test.json.lock', '--force', 'test-data_cuvs_ivf_flat,test,k10,bs100_2ce422d4-4e34-11f1-bc0f-6ef612336931.json']' died with <Signals.SIGSEGV: 11>.
  
assert 1 == 0
 +  where 1 = <Result FileNotFoundError(2, 'No such file or directory')>.exit_code
FAILED tests/test_cli.py::test_plot_command_creates_png_files - AssertionError: Plot command failed with output:
  
assert 1 == 0
 +  where 1 = <Result FileNotFoundError('No CSV result files found in /tmp/pytest-of-root/pytest-1/datasets0/test-data/result/search')>.exit_code
================== 2 failed, 47 passed, 3 warnings in 21.74s ===================
```

2. https://github.com/rapidsai/cuvs/actions/runs/25738253183/job/75590490314?pr=2072
A100, 12.2, **FAISS_GPU_IVF_FLAT_ANN_BENCH**

```
=========================== short test summary info ============================
FAILED tests/test_cli.py::test_run_command_creates_results - AssertionError: Run command failed with output:
  -- Using cuVS bench found in conda environment.
  -- Using cuVS bench found in conda environment.
  -- Using cuVS bench found in conda environment.
  -- Using cuVS bench found in conda environment.
  -- Using cuVS bench found in conda environment.
  Error occurred running benchmark: Command '['/opt/conda/envs/test/bin/ann/FAISS_GPU_IVF_FLAT_ANN_BENCH', '--build', '--data_prefix=/tmp/pytest-of-root/pytest-1/datasets0', '--benchmark_out_format=json', '--benchmark_counters_tabular=true', '--benchmark_out=/tmp/pytest-of-root/pytest-1/datasets0/test-data/result/build/faiss_gpu_ivf_flat,test.json.lock', '--force', 'test-data_faiss_gpu_ivf_flat,test,k10,bs100_214c59e6-4e11-11f1-acda-2e16edddf692.json']' died with <Signals.SIGSEGV: 11>.
  
assert 1 == 0
 +  where 1 = <Result FileNotFoundError(2, 'No such file or directory')>.exit_code
FAILED tests/test_cli.py::test_plot_command_creates_png_files - AssertionError: Plot command failed with output:
  
assert 1 == 0
 +  where 1 = <Result FileNotFoundError('No CSV result files found in /tmp/pytest-of-root/pytest-1/datasets0/test-data/result/search')>.exit_code
============= 2 failed, 47 passed, 3 warnings in 79.76s (0:01:19) ==============
```

##  [Disclaimer] Below is cursor's summary: Describe the bug

`cuvs::bench::blob_mmap<T>::handle()` in `cpp/bench/ann/src/common/blob.hpp` performs lazy `mmap()` initialization with no synchronization:

```cpp
[[nodiscard]] auto handle() const -> const std::tuple<mmap_owner, ptrdiff_t>&
{
  if (!handle_.has_value()) {                  // unsynchronized
    // ... mmap(...) ...
    handle_.emplace(std::move(owner), offset); // multiple threads can reach this
  }
  return *handle_;
}
```

The `ground_truth_map<T>` constructor (`dataset.hpp`) spawns `hardware_concurrency() - 1` workers that all call `data()` concurrently. When ≥ 2 threads race through the check, each calls `mmap()`, and each `emplace(...)` **destroys the previous `mmap_owner` → calls `munmap()` on it**. Threads still reading through the prior mapping crash with SIGSEGV.

Captured core dumps confirm exactly this: the crashing thread is inside the `ground_truth_map<int>` lambda while two siblings are in `mmap64()` and two more in `munmap()` on the same file.

Closes #1782 (closed prematurely in Feb 2026 without root-cause fix).

## Steps to reproduce

Most reliably reproduced via parallel hammering. Inside a `rapidsai/ci-conda:26.06-cuda13.2.0-ubuntu24.04-py3.13` container with `--cpus=16 --memory=62g`, run four parallel `CUVS_IVF_FLAT_ANN_BENCH --build` loops on the same dataset; expect `rc=139` (SIGSEGV) within ~150 cumulative attempts.

## Environment

- Docker: `rapidsai/ci-conda:26.06-cuda13.2.0-ubuntu24.04-py3.13`
- Hardware: NVIDIA RTX PRO 6000 + 16-CPU cgroup (CI nightly runner)
- Affected CI test: `python/cuvs_bench/cuvs_bench/tests/test_cli.py::test_run_command_creates_results`

## Validation evidence

`std::call_once` fix applied to `blob_mmap<T>::handle()` was validated under identical parallel hammer conditions:

| Binary | Attempts | SIGSEGV failures |
|--------|---------:|-----------------:|
| Unpatched | ~150 | 1 |
| Patched | 800 | 0 |

aamijar:
Resolves https://github.com/rapidsai/cuvs/issues/2087

# What does this PR do?

There is a race condition on the std::optional `!handle_.has_value()`.
While thread 1 works on assigning handle with mmap, thread 2 can race through and try to assign it as well. However,   during reassignment it causes destructor of thread 1 to be called which calls unmap, leading to a segfault.

To solve this we use `std::once` so that other threads can't enter the critical section. 
`std::once` is non reentrant, so we need to also refactor from using recursion to a while loop.
`std::once` is not movable, so we allocate it on the heap and use a unique_pointer to keep track of it instead.
We need to ensure member variables of blob are movable since blob is used in a std::variant

### DIFF
--- a/cpp/bench/ann/src/common/blob.hpp
+++ b/cpp/bench/ann/src/common/blob.hpp
@@ -430,8 +430,9 @@ struct blob_mmap {
   // Heap-allocate the once_flag so that blob_mmap remains movable (std::once_flag
   // itself is neither copyable nor movable). Multiple threads can race on the
   // first call to handle() via blob<T>::data(); without this serialization each
-  // racing thread would mmap the file and the losing emplace would munmap the
-  // winner's mapping out from under it -- a flaky SIGSEGV in ann benchmarks.
+  // racing thread would mmap the file and the losing emplace would munmap
+  // the winner's mapping AND invalidate the reference the winner had already
+  // returned to its caller -- a flaky SIGSEGV in ann benchmarks.
   mutable std::unique_ptr<std::once_flag> handle_once_ = std::make_unique<std::once_flag>();
 
   [[nodiscard]] auto handle() const -> const std::tuple<mmap_owner, ptrdiff_t>&
--- a/cpp/bench/ann/src/common/blob.hpp
+++ b/cpp/bench/ann/src/common/blob.hpp
@@ -1,5 +1,5 @@
 /*
- * SPDX-FileCopyrightText: Copyright (c) 2025, NVIDIA CORPORATION.
+ * SPDX-FileCopyrightText: Copyright (c) 2025-2026, NVIDIA CORPORATION.
  * SPDX-License-Identifier: Apache-2.0
  */
 #pragma once
@@ -16,6 +16,8 @@
 #include <cassert>
 #include <cstdint>
 #include <cstdio>
+#include <memory>
+#include <mutex>
 #include <optional>
 #include <stdexcept>
 #include <string>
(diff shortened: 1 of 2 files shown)


cpp/bench/ann/src/common/blob.hpp
```diff
@@ -425,66 +427,79 @@ struct blob_mmap {
   mutable bool hugepages_2mb_actual_;
 
   mutable std::optional<std::tuple<mmap_owner, ptrdiff_t>> handle_;
+  // Heap-allocate the once_flag so that blob_mmap remains movable (std::once_flag
+  // itself is neither copyable nor movable). Multiple threads can race on the
+  // first call to handle() via blob<T>::data(); without this serialization each
+  // racing thread would mmap the file and the losing emplace would munmap the
+  // winner's mapping out from under it -- a flaky SIGSEGV in ann benchmarks.
```
dantegd: Tiny suggestion — the comment explains the destructor-during-emplace half of the bug really well, but the other half is that the reference handle() already returned to the winning thread also gets invalidated by the second emplace. Maybe worth tweaking the last line to something like:

// racing thread would mmap the file and the losing emplace would both munmap // the winner's mapping AND invalidate the reference the winner had already // returned to its caller -- a flaky SIGSEGV in ann benchmarks.

Totally optional, the existing comment is already good.

dantegd (approved):
Nice catch on this one!

One thing I noticed while reading: ground_truth_map's ctor in dataset.hpp is the only place that goes parallel directly over a blob (everywhere else in dataset serializes via mutex_). With this PR it's safe, but if you ever want to make it future proof against new lazy paths in blob.hpp, prewarming ground_truth_set.data() and filter_bitset->data(...) on the main thread before spawning workers would do it. Not blocking, just a thought.

cpp/bench/ann/src/common/blob.hpp
aamijar: Thanks for the review! I addressed the future proofing and comments update here [39f5d9a](https://github.com/rapidsai/cuvs/pull/2088/commits/39f5d9af8973b42030e3434193201abdde423c40)

dantegd: I can't see [39f5d9a](https://github.com/rapidsai/cuvs/pull/2088/commits/39f5d9af8973b42030e3434193201abdde423c40), you meant [fa616f0](https://github.com/rapidsai/cuvs/pull/2088/commits/fa616f006b6806ae61e1db0d1796ada20c2c63d3) right?

aamijar: Yes, it got switched because of the rebase

PR #2091, opened 2026-05-15
Persistent CAGRA: benchmark group and bad config warnings

achirkin:
Improve user experience with CAGRA search persistent kernel mode:

- make sure SINGLE_CTA algo is selected in persistent mode if it is not passed explicitly
- warn about incompatible configuration
- add a cuvs benchmarking group for reference


tfeher (approved):
LGTM, thanks Artem!

cjnolet (approved):

PR #2101, opened 2026-05-16
fix(visibility): add symbol visibility controls to cuvs

vyasr:
Add proper symbol visibility controls to libcuvs.so, compatible with the raft visibility restrictions in rapidsai/raft#3019.

### Motivation

With raft restricting exports to only `core/` and `raft_runtime/`, cuvs needs its own visibility controls to:
1. Prevent symbol interposition between libcuml.so and libcuvs.so
2. Export only the intended public API from libcuvs.so
3. Keep internal implementation details hidden

### Changes

- **Add `CUDA_VISIBILITY_PRESET hidden` to `cuvs_objs`** — all symbols hidden by default
- **Add `CUVS_EXPORT` to public API symbols** — mark symbols that must be visible in libcuvs.so
- **Fix NVCC compatibility** — use `struct CUVS_EXPORT` on definitions (not instantiations) for NVCC
- **Fix `__half` type visibility** — add cuda_fp16.hpp wrapper to handle type visibility across TUs
- **Fix brute_force, cagra, resize_list exports** — ensure all needed symbols are exported
- **Fix remaining linker errors** — 4 additional hidden symbols that needed export
- **Fix CI rattler-build** — restore `rapids-rattler-channel-string` in both build_cpp.sh and build_python.sh

### CI note

The `test(ci): build against vyasr/raft#3019` commit that pins raft to the visibility branch will be reverted before merge (once raft#3019 is merged to release/26.06).

Depends on https://github.com/rapidsai/raft/pull/3019

Contributes to https://github.com/rapidsai/build-infra/issues/53

### DIFF
--- a/cpp/include/cuvs/neighbors/common.hpp
+++ b/cpp/include/cuvs/neighbors/common.hpp
@@ -19,6 +19,7 @@
 
 #include <cuvs/core/bitmap.hpp>
 #include <cuvs/core/bitset.hpp>
+#include <cuvs/core/export.hpp>
 #include <raft/core/detail/macros.hpp>
 
 #include <memory>
@@ -27,7 +28,6 @@
 
 #ifdef __cpp_lib_bitops
 #include <bit>
-#include <cuvs/core/export.hpp>
 #endif
 
 namespace CUVS_EXPORT cuvs {
--- a/cpp/include/cuvs/neighbors/common.hpp
+++ b/cpp/include/cuvs/neighbors/common.hpp
@@ -811,11 +811,11 @@ using enable_if_valid_list_t = typename enable_if_valid_list<ListT, T>::type;
  *       `cuvs::neighbors::ivf_pq::helpers::resize_list` which handle type casting internally.
  */
 template <typename ListT>
-void resize_list(raft::resources const& res,
-                 std::shared_ptr<ListT>& orig_list,  // NOLINT
-                 const typename ListT::spec_type& spec,
-                 typename ListT::size_type new_used_size,
-                 typename ListT::size_type old_used_size);
+CUVS_EXPORT void resize_list(raft::resources const& res,
+                             std::shared_ptr<ListT>& orig_list,  // NOLINT
+                             const typename ListT::spec_type& spec,
+                             typename ListT::size_type new_used_size,
+                             typename ListT::size_type old_used_size);
 
 /**
  * Serialize a list to an output stream.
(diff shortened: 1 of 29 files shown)


vyasr:
https://github.com/rapidsai/cuvs/actions/runs/26193227824?pr=2101 shows the important jobs passing when built against https://github.com/rapidsai/raft/pull/3019.

cpp/include/cuvs/neighbors/common.hpp
```diff
@@ -811,11 +811,11 @@ using enable_if_valid_list_t = typename enable_if_valid_list<ListT, T>::type;
  *       `cuvs::neighbors::ivf_pq::helpers::resize_list` which handle type casting internally.
  */
 template <typename ListT>
-void resize_list(raft::resources const& res,
-                 std::shared_ptr<ListT>& orig_list,  // NOLINT
-                 const typename ListT::spec_type& spec,
-                 typename ListT::size_type new_used_size,
-                 typename ListT::size_type old_used_size);
+CUVS_EXPORT void resize_list(raft::resources const& res,
```
divyegala, line 814: Do we have to manually mark templated functions for export?

cpp/src/neighbors/ivf_flat/ivf_flat_helpers.cu
```diff
@@ -206,3 +207,15 @@ void recompute_internal_state(const raft::resources& res, index<uint8_t, int64_t
 }
 
 }  // namespace cuvs::neighbors::ivf_flat::helpers
+
+namespace cuvs::neighbors::ivf {
+
+template CUVS_EXPORT void
+resize_list<list<cuvs::neighbors::ivf_flat::list_spec, uint32_t, half, int64_t>>(
```
divyegala, line 214: Was this a missing instantiation?

cpp/include/cuvs/neighbors/common.hpp
vyasr: I think this was basically an nvcc bug/quirk where it wouldn't accept the attribute in the same places that gcc would. Specifically I believe it was when there was a templated function that we wanted to export, the markup on the namespace wasn't affecting the symbol so we had to put it on individual symbols. I'd say more quirk than bug since I don't believe there's any standard requiring the attribute on the namespace to work, gcc just happens to support it.

cpp/src/neighbors/ivf_flat/ivf_flat_helpers.cu
vyasr: This is related to your other question. nvcc + template specialization + explicit instantiation + explicit symbol visibility markup hit a number of edge cases and we had to find the magic incantation that would both be supported by nvcc and result in the right output. I won't pretend that I was particularly principled in my choice, I just tried the combinations that were closest to standards-compliant in reverse order until one produced the outcome I wanted.

divyegala (approved):

PR #2103, opened 2026-05-18
Fix cagra::optimize modifying the state of raft::resources

achirkin:
The new, improved version of CAGRA optimize() function makes use of raft CUDA stream pool resource.
Setting the resource (`raft::resource::set_cuda_stream_pool`) affects the state of `raft::resources` and may lead to unintended consequences for a user.
Using non-const `raft::resources` will be enforced at compile time in https://github.com/rapidsai/raft/pull/3005 , so this PR fixes the use case ahead of the upstream change by copying the resources handle.

### DIFF
--- a/cpp/src/neighbors/detail/cagra/graph_core.cuh
+++ b/cpp/src/neighbors/detail/cagra/graph_core.cuh
@@ -1706,7 +1706,7 @@ template <typename IdxT = uint32_t,
           typename AccessorOutputGraph =
             raft::host_device_accessor<cuda::std::default_accessor<IdxT>, raft::memory_type::host>>
 void optimize(
-  raft::resources const& res,
+  raft::resources const& res_const,
   raft::mdspan<IdxT, raft::matrix_extent<int64_t>, raft::row_major, AccessorKnnGraph> knn_graph,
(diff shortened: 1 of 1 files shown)


mfoerste4 (approved):
Thanks @achirkin for fixing this!

tfeher (approved):
LGTM.

cpp/src/neighbors/detail/cagra/graph_core.cuh
```diff
@@ -1715,6 +1715,12 @@ void optimize(
   RAFT_LOG_DEBUG(
     "# Pruning kNN graph (size=%lu, degree=%lu)\n", knn_graph.extent(0), knn_graph.extent(1));
 
+  // TODO(achirkin): come up with a reasonable API to initialize a non-empty stream pool.
+  // raft::resource::set_cuda_stream_pool below modifies the resource, so it cannot be const.
+  // The optimize() is a heavy function, so copying the resource and creating a private stream pool
+  // is not a big overhead.
+  raft::resources res{res_const};
```
cjnolet, line 1722: Please add a GitHub issue for this and link it in the code so we don't lose sight of this.

achirkin: https://github.com/rapidsai/raft/issues/3022

PR #2107, opened 2026-05-19
Fix check for PQ vectorized load

lowener:
16‑byte load needs a 16‑byte aligned address, and 8‑byte load needs 8‑byte alignment.
Instead of checking for multiple the code was checking for powers, this led to `pq_len` values that are multiples of 2 but not power of 2 to not go through the vectorized load.
After some quick benchmarking, the pq transform operations with (`n_vecs=1M`, `dim=768`, `pq_dim=64`, `pq_bits=8`, `pq_len=6`) went from ~110ms to ~54ms.
With (`n_vecs=5M`, `dim=228`, `pq_dim=38`, `pq_bits=8`, `pq_len=6`) it went from ~148ms to ~112ms.

### DIFF
--- a/cpp/src/neighbors/detail/vpq_dataset.cuh
+++ b/cpp/src/neighbors/detail/vpq_dataset.cuh
@@ -583,10 +583,10 @@ __device__ __forceinline__ void process_4centers_vec(MathT& d0,
                                                      GetXFunc get_x_func)
 {
   uint32_t k = 0;
-  // If pq_len is a power of 2, we can use vectorized loads and stores
-  // Otherwise, we fall back to scalar loads and stores to avoid misaligned accesses
-  bool pq_len_is_pow2 = raft::is_pow2(pq_len);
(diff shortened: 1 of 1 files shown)


achirkin (approved):
Looks like a good contained improvement.
One question though: it looks like you only read the centers vectorized, right? And it's runtime decision anyway (pq_len is not constexpr)? Then maybe it would be easier to just check if centers_ptr is aligned and issue vectorized loads even if pq_len is not multiple of two or four? Like raft::linalg::map does (but this is probably too much hassle for a small device function).

lowener:
Yes only the loads of pq_centers are vectorized and yes it is only at runtime.
I don't think that the `raft::linalg::map` way is applicable here because map treats multiple elements with stride one, here the stride depends on pq_len,
I think a path forward that would improve PQ transform would be to pad the pq_centers to the nearest multiple of 2 or 4s to always use vectorized loads.

PR #2114, opened 2026-05-20
Accommodate rmm breaking change

tarang-jain:
https://github.com/rapidsai/rmm/pull/2387 removed the `_ref` suffix from memroy resources. This PR fixes the resultant CI break


aamijar (approved):
Thanks @tarang-jain!

PR #2119, opened 2026-05-22
contributing.md link to docs

aamijar:
Symlink `CONTRIBUTING.md` to fern/pages/contributing.md so that it appears as a separate tab next to readme on github.


cjnolet:
@aamijar I don't believe we can move it like this. Instead, let's just link to the contributing guide in the docs from a different contributing.md. 

I'm honestly not concerned with what cuopt is doing. We dont need to repeat others.

aamijar:
Hi @cjnolet, it worked fine for me when I built the fern docs and viewed them locally. All the links resolve fine.
And the contributing.md will show up like this on the github page: https://github.com/aamijar/cuvs/tree/contributing-guide-move?tab=contributing-ov-file#contributing

cjnolet:
@aamijar I hate to break it to you but you can't link to pages outside of fern/pages. It won't deploy that way. 

Let's move the contributor guide back into the docs and create a new file in the root that just links to  corresponding page in the docs. We don't need this duplicated in multiple places.

cjnolet (approved):

gforsyth (approved):

PR #2125, opened 2026-05-26
Native row-major PCA

### LINKED ISSUE #2115: [FEA] row-major PCA
Users are passing in row-major data. We want to natively support row-major in PCA without copying the data to be in col-major format.
This will require changes in raft since all the current primitives for PCA are in col-major. Some primitives may support row-major others may need to be refactored.

aamijar:
Currently PCA only supports col-major in the cpp layer. If a consumer's input dataset is in row-major it must be copied as col-major. With the introduction of RAFT native row-major support in PCA https://github.com/rapidsai/raft/pull/3036, users can seamlessly pass in row-major datasets.
This PR exposes the cpp PCA row-major APIs and updates the C and Python APIs accordingly.

Resolves https://github.com/rapidsai/cuvs/issues/2115. Depends on https://github.com/rapidsai/raft/pull/3036.


cpp/cmake/thirdparty/get_raft.cmake
```diff
@@ -60,8 +60,8 @@ endfunction()
 # To use a different RAFT locally, set the CMake variable
 # CPM_raft_SOURCE=/path/to/local/raft
 find_and_configure_raft(VERSION  ${RAFT_VERSION}.00
-        FORK                     ${RAFT_FORK}
-        PINNED_TAG               ${RAFT_PINNED_TAG}
+        FORK                     aamijar
+        PINNED_TAG               pca-row-major
```
aamijar: Revert later

cpp/tests/preprocessing/pca.cu
```diff
@@ -263,4 +262,139 @@ TEST_P(PcaTestF, Result) { this->testPca(); }
...
+
+ protected:
+  // Convert col-major data of shape (n_rows, n_cols) into row-major in dst.
+  // Both buffers live on device.
+  void to_row_major(const T* col_major_src, T* row_major_dst, int n_rows, int n_cols)
+  {
+    std::vector<T> host_col(n_rows * n_cols);
+    std::vector<T> host_row(n_rows * n_cols);
+    raft::update_host(host_col.data(), col_major_src, n_rows * n_cols, stream);
+    RAFT_CUDA_TRY(cudaStreamSynchronize(stream));
```
cjnolet: Please accept the handle here and use the stream sync from the handle. I know other tests still use `cudaStreamSync` directly, but we should use handles going forward (eventually we will clean up the other tests)

cjnolet (approved):

cpp/tests/preprocessing/pca.cu
aamijar: Addressed in https://github.com/NVIDIA/cuvs/pull/2125/changes/3dac191ced8d920661667e1943b9399b9c6c5576

PR #2130, opened 2026-05-26
`uint64` header support for `cuvs_bench`

### LINKED ISSUE #2129: [FEA] uint64 type header support for cuvs bench
Current max support is `uint32`, allowing datasets upto 4.2B.

We want to go higher with the new billion-scale gen.

jinsolp:
Closes https://github.com/rapidsai/cuvs/issues/2129

This PR adds a extended uint64 header support for `bin` data files. It automatically detects based on the filesize, and therefore does not break support for existing files.
This is useful for datasets that exceed `uint32` max number of rows (4.2B).


python/cuvs_bench/cuvs_bench/generate_groundtruth/utils.py
```diff
@@ -42,27 +44,60 @@ def suffix_from_dtype(dtype):
...
+    itemsize = np.dtype(dtype).itemsize
+
+    if shape is not None and len(shape) != 2:
+        raise ValueError(
+            f"shape must have exactly 2 dimensions (n_rows, n_cols), got {shape!r}"
+        )
 
     if mode[0] == "r":
-        a = np.memmap(bin_file, mode=mode, dtype=size_dtype, shape=(2,))
+        n_rows, n_cols, header_bytes = read_bin_header(bin_file, itemsize)
```
dantegd, line 137: What do you think about adding direct tests for memmap_bin_file with both legacy and extended headers? This is one of the important large-file paths, and this PR changes the read offset detection as well as the write path, so a small round-trip test here would make breakages easy to spot

python/cuvs_bench/cuvs_bench/tests/test_utils.py
```diff
@@ -170,14 +181,173 @@ def test_truncated_data(self, tmp_path):
...
+        """A hand-crafted extended-header file with >UINT32_MAX rows reads correctly.
+
+        We can't materialize the data section (>16 GB just for the dummy
+        bytes), so write only the header and pad with the exact number of
+        zero bytes ``read_bin_header`` expects to balance the size equation
+        -- using ``n_cols=0`` so the data section is empty.
+        """
+        path = tmp_path / "huge.fbin"
+        n_rows = UINT32_MAX + 17
+        n_cols = 0
```
dantegd: Could we make this use a positive column count and create the expected file size with truncate instead of using n_cols = 0? The zero-column case proves the parser can see a large uint64 value, but it doesn’t really exercise the large positive-shape case this feature is meant to support.

python/cuvs_bench/cuvs_bench/tests/test_utils.py
```diff
@@ -170,14 +181,173 @@ def test_truncated_data(self, tmp_path):
...
+        _write_test_bin(path, data, size_dtype=np.uint64)
+
+        loaded = load_vectors(path, subset_size=12)
+        assert loaded.shape == (12, 8)
+        np.testing.assert_array_equal(loaded, data[:12])
+
+    @pytest.mark.parametrize(
+        "ext, dtype, size_dtype",
+        [
+            (".fbin", np.float32, np.uint32),
```
dantegd, line 231: Could we include the float16 extension in this parametrized coverage too? load_vectors supports .f16bin, np? So it would be nice to cover that path for both header layouts as well.

dantegd (changes requested):
To make this end-to-end for the beyond uint32 rows goal, I think we may need to audit a few non-helper paths too. For example, cpp/bench/ann/src/common/blob.hpp still appears to treat .bin files as two uint32_t header values and uses an 8-byte data offset, so C++ benchmark consumers may not read these extended files correctly yet. Also, python/cuvs_bench/cuvs_bench/generate_groundtruth/__main__.py still writes ground-truth neighbor IDs as uint32, which may not be enough once base-row IDs can exceed 4B rows. Would it make sense to either update those paths here or call them out as explicit follow-up scope?

jinsolp:
Hi @dantegd I've addressed the test-related feedback.

And thanks for mentioning GT! The default for groundtruth `ibin`. Now it also writes groundtruth files as `u64bin` if needed.

I'll open another PR for the cpp side!: https://github.com/rapidsai/cuvs/issues/2219

python/cuvs_bench/cuvs_bench/generate_groundtruth/utils.py
```diff
@@ -39,30 +45,102 @@ def suffix_from_dtype(dtype):
...
+def offset_neighbor_indices(indices, batch_offset: int, n_base: int):
+    """Shift local neighbor IDs by a batch offset without integer overflow."""
+    dtype = neighbor_index_accumulator_dtype(n_base)
+    return indices.astype(dtype) + batch_offset
+
+
+def write_groundtruth_neighbors(path, indices, n_base: int):
+    """Write a ground-truth neighbor matrix using the correct on-disk dtype."""
+    storage_dtype = neighbor_index_dtype(n_base)
+    data = np.asarray(indices, dtype=storage_dtype)
```
dantegd: Could we make this handle CuPy arrays explicitly? In the GPU path, calc_truth can return indices as an xp=CuPy array, and np.asarray(cupy_array, ...) will fail (unless that's changed recently?) Maybe this should check for .get() or use an explicit CuPy-to-NumPy conversion before casting to the storage dtype.

dantegd (changes requested):
One small comment, besides that it looks great

dantegd (approved):

PR #2132, opened 2026-05-27
Add JIT-LTO based filter UDF support for CAGRA

dantegd:
This PR adds a first version of low-level JIT-LTO filter UDF support for CAGRA search in cuVS C++.

Users can provide CUDA device source for a predicate with this ABI:

```cpp
__device__ bool cuvs_filter_udf(uint32_t query_id,
                                source_index_t source_id,
                                void* filter_data);
```

The predicate returns `true` to allow a candidate and `false` to reject it. `filter_data` is not result data; it is an optional opaque device-accessible context pointer that lets the predicate read user metadata such as tenant ids, timestamps, language masks, ACL bitmaps, or query-specific thresholds. Results are still written to the normal CAGRA `neighbors` and `distances` outputs.

This gives CAGRA a path to support runtime metadata predicates without ahead-of-time template explosion, while keeping existing `none` and `bitset` filters on their static JIT-LTO fragment paths.

## What Changed

- Added `cuvs::neighbors::filtering::udf_filter`.
- Added `FilterType::UDF`.
- Added CAGRA dispatch for `udf_filter`, including `filtering_rate` fallback behavior.
- Added a dynamic JIT-LTO sample-filter fragment generator for UDF source.
- Reworked CAGRA JIT sample-filter payload plumbing from bitset-only payloads to a generic payload that supports:
  - no filter
  - bitset filter
  - UDF filter context pointer
  - query id offsets for batched/multi-kernel paths
- Updated single-CTA, multi-CTA, and multi-kernel CAGRA JIT paths to call the linked sample filter uniformly.
- Added docs for UDF behavior, caveats, and non-goals.
- Added a standalone C++ example showing three metadata-style UDFs.

## About `filter_data`

`filter_data` is the mechanism for passing runtime metadata into the device predicate. It is optional: simple predicates can ignore it or use `nullptr`.

For example, this UDF needs no context:

```cpp
__device__ bool cuvs_filter_udf(uint32_t, source_index_t source_id, void*)
{
  return source_id >= 704;
}
```

For metadata filters, callers pass a device pointer to a user-defined context struct:

```cpp
struct tenant_filter_context {
  const uint32_t* row_tenant_ids;
  const uint32_t* query_tenant_ids;
};
```

Then the UDF casts `filter_data` back to that type:

```cpp
__device__ bool tenant_filter(uint32_t query_id,
                              source_index_t source_id,
                              void* filter_data)
{
  auto* ctx = static_cast<const tenant_filter_context*>(filter_data);
  return ctx->row_tenant_ids[source_id] == ctx->query_tenant_ids[query_id];
}
```

The pointer and anything it points to must be device-accessible and remain valid for the duration of the search. Future typed wrappers or expression builders could hide this cast, but the internal ABI can stay stable as:

```cpp
bool predicate(uint32_t query_id, source_index_t source_id, void* filter_data);
```

## Example

The standalone example builds one CAGRA index and runs three different UDF predicates over the same metadata context.

First, the caller defines a host-side context type. This same layout is repeated in the UDF source string so device code knows how to interpret `filter_data`:

```cpp
struct metadata_filter_context {
  const uint32_t* row_tenant_ids;
  const int64_t* row_timestamps;
  const uint32_t* row_language_ids;
  const uint64_t* row_acl_masks;

  const uint32_t* query_tenant_ids;
  const int64_t* query_min_timestamps;
  const uint64_t* query_allowed_language_masks;
  const uint64_t* query_permission_masks;
};
```

The UDF source contains the device predicates. Each predicate receives the same `filter_data` pointer and casts it to `metadata_filter_context`:

```cpp
std::string source = R"cpp(
struct metadata_filter_context {
  const uint32_t* row_tenant_ids;
  const int64_t* row_timestamps;
  const uint32_t* row_language_ids;
  const uint64_t* row_acl_masks;

  const uint32_t* query_tenant_ids;
  const int64_t* query_min_timestamps;
  const uint64_t* query_allowed_language_masks;
  const uint64_t* query_permission_masks;
};

__device__ bool tenant_filter(uint32_t query_id,
                              source_index_t source_id,
                              void* filter_data)
{
  auto* ctx = static_cast<const metadata_filter_context*>(filter_data);
  return ctx->row_tenant_ids[source_id] == ctx->query_tenant_ids[query_id];
}

__device__ bool timestamp_filter(uint32_t query_id,
                                 source_index_t source_id,
                                 void* filter_data)
{
  auto* ctx = static_cast<const metadata_filter_context*>(filter_data);
  return ctx->row_timestamps[source_id] >= ctx->query_min_timestamps[query_id];
}

__device__ bool language_acl_filter(uint32_t query_id,
                                    source_index_t source_id,
                                    void* filter_data)
{
  auto* ctx = static_cast<const metadata_filter_context*>(filter_data);
  const auto language_bit = uint64_t{1} << ctx->row_language_ids[source_id];
  const bool language_ok =
    (ctx->query_allowed_language_masks[query_id] & language_bit) != 0;
  const bool acl_ok =
    (ctx->row_acl_masks[source_id] & ctx->query_permission_masks[query_id]) != 0;
  return language_ok && acl_ok;
}
)cpp";
```

The caller owns the metadata arrays. They must be copied to device memory before search:

```cpp
auto row_tenant_ids_device = raft::make_device_vector<uint32_t, int64_t>(res, n_rows);
auto row_timestamps_device = raft::make_device_vector<int64_t, int64_t>(res, n_rows);
auto row_language_ids_device = raft::make_device_vector<uint32_t, int64_t>(res, n_rows);
auto row_acl_masks_device = raft::make_device_vector<uint64_t, int64_t>(res, n_rows);

auto query_tenant_ids_device = raft::make_device_vector<uint32_t, int64_t>(res, n_queries);
auto query_min_timestamps_device = raft::make_device_vector<int64_t, int64_t>(res, n_queries);
auto query_allowed_language_masks_device =
  raft::make_device_vector<uint64_t, int64_t>(res, n_queries);
auto query_permission_masks_device =
  raft::make_device_vector<uint64_t, int64_t>(res, n_queries);

// Copy host metadata into these device arrays before launching search.
```

Then the caller builds one device-resident context struct whose fields point at those device arrays:

```cpp
auto context_device =
  raft::make_device_vector<metadata_filter_context, int64_t>(res, 1);

metadata_filter_context host_context{
  row_tenant_ids_device.data_handle(),
  row_timestamps_device.data_handle(),
  row_language_ids_device.data_handle(),
  row_acl_masks_device.data_handle(),
  query_tenant_ids_device.data_handle(),
  query_min_timestamps_device.data_handle(),
  query_allowed_language_masks_device.data_handle(),
  query_permission_masks_device.data_handle()
};

// Copy the struct itself to device memory. The pointer to this device struct
// is what we pass as filter_data.
raft::copy(context_device.data_handle(),
           &host_context,
           1,
           raft::resource::get_cuda_stream(res));
raft::resource::sync_stream(res);
```

Finally, each `udf_filter` selects which device function to link by name. All three filters reuse the same source and the same `filter_data` context:

```cpp
auto tenant_filter = cuvs::neighbors::filtering::udf_filter(
  source,
  context_device.data_handle(),  // filter_data
  0.75f,                         // estimated rejected fraction
  "tenant-filter-v1",            // cache key
  "tenant_filter");              // device function name

auto timestamp_filter = cuvs::neighbors::filtering::udf_filter(
  source,
  context_device.data_handle(),
  0.50f,
  "timestamp-filter-v1",
  "timestamp_filter");

auto language_acl_filter = cuvs::neighbors::filtering::udf_filter(
  source,
  context_device.data_handle(),
  0.875f,
  "language-acl-filter-v1",
  "language_acl_filter");
```

Example output:

```text
Building CAGRA index
tenant_filter first query neighbors: 3364 3488 260 772 3292 3344 2052 660
timestamp_filter first query neighbors: 4070 3942 3364 3629 3797 3488 3292 3495
language_acl_filter first query neighbors: 3488 3344 2480 2520 2304 2176 1360 400
All CAGRA filter UDF examples produced valid filtered neighbors.
```

## Validation

Focused CAGRA UDF test passes across all CAGRA search algorithms:

```text
NEIGHBORS_ANN_CAGRA_FILTER_UDF_TEST
```

Coverage includes:

- accept-all UDF matches no-filter results
- reject-all UDF returns no valid dataset row ids
- high `filtering_rate` UDF returns only accepted rows
- invalid source throws during compile/search setup
- repeated same-cache-key searches match
- UDF threshold predicate matches equivalent bitset filter
- query-specific tenant metadata works across SINGLE_CTA, MULTI_CTA, and MULTI_KERNEL

Broader CAGRA regression set also passed:

```text
NEIGHBORS_ANN_CAGRA_(FILTER_UDF_TEST|FLOAT_UINT32|INT8_UINT32|UINT8_UINT32|HALF_UINT32|TEST_BUGS)
```

## Notes

This PR intentionally keeps the v1 API low-level: users provide CUDA source plus an optional `void*` device context. Typed wrappers, expression builders, or write-once host/device ergonomics can be layered on later without changing the internal ABI.

Filter UDFs are candidate-validity predicates only. They do not control CAGRA graph traversal, distance computation, PQ/VPQ internals, or result selection.

## Benchmarks

Performance benchmarking is still TODO for this PR.

The expected regression risk for existing functionality is low because `none` and `bitset` filters still use static JIT-LTO fragments, and the UDF dynamic fragment path is only used for `udf_filter`. However, the CAGRA JIT kernel payload/signature plumbing changed from bitset-only payloads to a generic sample-filter payload, so we should still validate no-filter and bitset search performance against `main`.

Proposed planned benchmark coverage:

- no-filter CAGRA search vs `main`
- bitset-filtered CAGRA search vs `main`
- equivalent bitset predicate vs UDF predicate
- first-call UDF compile/link latency
- warm-cache repeated UDF search latency
- representative CAGRA algorithms/configs, especially SINGLE_CTA, MULTI_CTA, and MULTI_KERNEL where applicable

This should give us both existing-functionality regression coverage and a baseline for the new UDF path.


cpp/include/cuvs/neighbors/common.hpp
```diff
@@ -615,6 +617,50 @@ struct bitset_filter : public base_filter {
...
+  /** CUDA C++ source containing the device predicate. */
+  std::string source;
+  /** Opaque device-accessible pointer passed to the predicate. */
+  void* filter_data = nullptr;
+  /** Estimated fraction of rows rejected by the predicate, or negative if unknown. */
+  float filtering_rate = -1.0f;
+  /** Optional stable cache key for equivalent generated source. */
+  std::string cache_key;
+  /** Device function name to call from the generated CAGRA sample filter. */
+  std::string function_name = "cuvs_filter_udf";
```
divyegala: Why expose these 2? We can just hide it within implementation details. As you noted previously in IVF Flat, the cache key can just be the string itself

cpp/src/neighbors/detail/cagra/jit_lto_kernels/cagra_bitset.cuh
```diff
@@ -15,10 +15,14 @@ namespace cuvs::neighbors::cagra::detail {
 template <typename SourceIndexT>
 using cagra_bitset = cuvs::neighbors::detail::bitset_filter_data_t<SourceIndexT>;
 
-/// Host: bitset payload for kernels plus query offset for wrapped CAGRA filters.
+enum class cagra_filter_kind : std::uint32_t { none = 0, bitset = 1, udf = 2 };
+
+/// Host/device payload for linked CAGRA sample filters plus query offset for wrapped filters.
 template <typename SourceIndexT>
 struct cagra_sample_filter {
   cagra_bitset<SourceIndexT> bitset{};
```
divyegala: Can we hide this behind the opaque `filter_data`? `bitset` is an implementation detail of the `bitset_filter`

cpp/src/neighbors/detail/cagra/jit_lto_kernels/cagra_bitset.cuh
```diff
@@ -29,6 +33,28 @@ template <typename bitset_t, typename index_t>
...
+
+template <typename SourceIndexT, typename FilterT>
+void fill_cagra_sample_filter(cagra_sample_filter<SourceIndexT>& out, const FilterT& filter)
+{
+  using DecayedFilter = std::decay_t<FilterT>;
+  if constexpr (is_bitset_filter<DecayedFilter>::value) {
+    const auto bitset_view    = filter.view();
+    out.bitset.bitset_ptr     = const_cast<std::uint32_t*>(bitset_view.data());
+    out.bitset.bitset_len     = static_cast<SourceIndexT>(bitset_view.size());
+    out.bitset.original_nbits = static_cast<SourceIndexT>(bitset_view.get_original_nbits());
```
divyegala: So going by my above comment, you would assign `out.filter_data = cagra_bitset<SourceIndexT>{bitset_ptr, bitset_len, original_nbits}` and then reinterpret this in the fragment

cpp/src/neighbors/detail/cagra/jit_lto_kernels/cagra_bitset.cuh
```diff
@@ -15,10 +15,14 @@ namespace cuvs::neighbors::cagra::detail {
...
 
-/// Host: bitset payload for kernels plus query offset for wrapped CAGRA filters.
+enum class cagra_filter_kind : std::uint32_t { none = 0, bitset = 1, udf = 2 };
+
+/// Host/device payload for linked CAGRA sample filters plus query offset for wrapped filters.
 template <typename SourceIndexT>
 struct cagra_sample_filter {
   cagra_bitset<SourceIndexT> bitset{};
+  void* filter_data{nullptr};
+  cagra_filter_kind filter_kind{cagra_filter_kind::none};
```
divyegala: I'm not totally sure we need this

cpp/src/neighbors/detail/cagra/jit_lto_kernels/cagra_bitset.cuh
```diff
@@ -39,15 +65,34 @@ cagra_sample_filter<SourceIndexT> extract_cagra_sample_filter(const SampleFilter
...
+}
+
+template <typename SourceIndexT>
+__device__ __forceinline__ void* get_cagra_sample_filter_data(
+  cagra_sample_filter<SourceIndexT>& payload)
+{
+  if (payload.filter_kind == cagra_filter_kind::udf) { return payload.filter_data; }
+  if (payload.filter_kind == cagra_filter_kind::bitset) { return &payload.bitset; }
+  return nullptr;
+}
```
divyegala: If you make `bitset` opaque then you don't need this function

cpp/src/neighbors/detail/cagra/jit_lto_kernels/cagra_filter_payload.cuh
```diff
@@ -39,15 +65,34 @@ cagra_sample_filter<SourceIndexT> extract_cagra_sample_filter(const SampleFilter
...
 
+template <typename SampleFilterT>
+const cuvs::neighbors::filtering::udf_filter* get_cagra_udf_filter(
+  const SampleFilterT& sample_filter)
+{
+  using DecayedFilter = std::decay_t<SampleFilterT>;
+  if constexpr (is_udf_filter<DecayedFilter>::value) {
+    return &sample_filter;
+  } else if constexpr (requires { sample_filter.filter; }) {
+    return get_cagra_udf_filter(sample_filter.filter);
```
divyegala: Missing `offset`?

cpp/src/neighbors/detail/cagra/jit_lto_kernels/search_multi_jit.cuh
```diff
@@ -115,13 +115,14 @@ __device__ void compute_distance_to_child_nodes_kernel_jit(
 
   const auto team_size      = 1u << team_size_bits;
   const uint32_t ldb        = hashmap::get_size(hash_bitlen);
-  const auto tid            = threadIdx.x + blockDim.x * blockIdx.x;
-  const auto global_team_id = tid >> team_size_bits;
-  const auto query_id       = blockIdx.y;
+  const auto tid             = threadIdx.x + blockDim.x * blockIdx.x;
+  const auto global_team_id  = tid >> team_size_bits;
+  const auto local_query_id  = blockIdx.y;
+  const auto filter_query_id = filter_payload.query_id_offset + local_query_id;
```
divyegala, line 121: Was this a pre-existing bug?

cpp/include/cuvs/neighbors/common.hpp
dantegd: Agree on cache_key; it is an implementation detail and can be derived internally from the source/generated code.

For function_name, the intent was to support one source string with several predicates over the same metadata context, e.g. tenant_filter, timestamp_filter, and acl_filter, and let the user select which one to link without regenerating wrapper source each time. What do you think?

divyegala: Oh that is pretty smart, so you just pay compilation costs once.

cpp/src/neighbors/detail/cagra/jit_lto_kernels/search_multi_jit.cuh
dantegd: Yes, this was a latent multi-kernel bug for query-dependent filters. Existing bitset filtering ignored query_id, so it was not observable. UDFs can use per-query metadata, so the mid-search filter path must pass the global query id, including the batch offset.

cpp/src/neighbors/detail/cagra/jit_lto_kernels/cagra_bitset.cuh
dantegd: refactored this into a generic cagra_filter_payload.cuh and renamed the embedded storage to filter_data_storage in https://github.com/rapidsai/cuvs/pull/2132/commits/296d44c68654ba43e9e0ef64834979f13f144039

cpp/include/cuvs/neighbors/common.hpp
dantegd: cache_key addressed in https://github.com/rapidsai/cuvs/pull/2132/commits/292f7e857bd795d6341084714dffba87f72602bc

cpp/src/neighbors/detail/cagra/jit_lto_kernels/cagra_filter_payload.cuh
dantegd: Offset is not needed in get_cagra_udf_filter() because that helper only finds UDF compile/link metadata. Runtime query offset is carried through extract_cagra_sample_filter() and applied before calling the linked sample_filter. I added a comment to make this explicit https://github.com/rapidsai/cuvs/pull/2132/commits/51716a6b375c00423a59dc01086ea7f21d6b7389

cpp/src/neighbors/detail/cagra/jit_lto_kernels/cagra_filter_payload.cuh
```diff
@@ -0,0 +1,243 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2025-2026, NVIDIA CORPORATION & AFFILIATES. All rights
+ * reserved. SPDX-License-Identifier: Apache-2.0
+ */
+#pragma once
+
+#include "../../../sample_filter.cuh"  // public filter types
+#include "../../sample_filter_data.cuh"
+
+#if !defined(__CUDACC_RTC__)
```
divyegala: Are you including this header in runtime compilation? How are you doing so?

cpp/src/neighbors/detail/cagra/jit_lto_kernels/cagra_filter_payload.cuh
```diff
@@ -0,0 +1,243 @@
...
+#include <memory>
+#include <mutex>
+#include <unordered_map>
+#include <vector>
+#endif
+
+namespace cuvs::neighbors::cagra::detail {
+
+template <typename SourceIndexT>
+using cagra_filter_data_storage = ::cuvs::neighbors::detail::bitset_filter_data_t<SourceIndexT>;
```
divyegala: Unused type?

cpp/src/neighbors/detail/cagra/jit_lto_kernels/cagra_filter_payload.cuh
```diff
@@ -0,0 +1,243 @@
...
+    return state_->dev_ptr(stream);
+  }
+
+  PayloadT const* host_payload() const
+  {
+    return state_ == nullptr ? nullptr : &state_->host_payload;
+  }
+
+ private:
+  mutable std::shared_ptr<state> state_;
```
divyegala: Just wanted to point out that by using a `shared_ptr` that is initialized on construction, it becomes a requirement that `state` be initialized before threads branch off. Would that be the case if the user were handling threading and calling `search`? If not, would it make sense to make this `static` instead?

cpp/src/neighbors/detail/cagra/jit_lto_kernels/cagra_filter_payload.cuh
```diff
@@ -0,0 +1,243 @@
...
+};
+
+template <typename SourceIndexT>
+struct cagra_sample_filter_payload {
+  cagra_sample_filter<SourceIndexT> payload{};
+  cagra_device_payload_owner<cagra_filter_data_storage<SourceIndexT>> storage_owner{};
```
divyegala: `cagra_filter_data_storage` is still of type `bitset`. So we will have to add an `std::variant` here to support pre-compiled filter types? In which case, why even have the owner, we could just go back to storing it by value for pre-compiled types.

cpp/src/neighbors/detail/cagra/jit_lto_kernels/cagra_filter_payload.cuh
```diff
@@ -0,0 +1,243 @@
...
+  } else {
+    fill_cagra_sample_filter(out, sample_filter);
+  }
+  return out;
+}
+
+/// Host: find UDF compile/link metadata only. Query offsets stay in the runtime payload produced
+/// by @ref extract_cagra_sample_filter and are applied before calling the linked sample_filter.
+template <typename SampleFilterT>
+const ::cuvs::neighbors::filtering::udf_filter* get_cagra_udf_filter(
```
divyegala: This function is a bit ugly. Is it possible to enforce that any and all CAGRA filters are wrapped by the outer SampleFilter type?

cpp/src/neighbors/detail/cagra/search_multi_cta_kernel_launcher_jit.cuh
```diff
@@ -56,15 +56,17 @@ void select_and_run(const dataset_descriptor_host<DataT, IndexT, DistanceT>& dat
                     SampleFilterT sample_filter,
                     cudaStream_t stream)
 {
-  const auto bf                  = extract_cagra_sample_filter<SourceIndexT>(sample_filter);
-  const uint32_t query_id_offset = bf.query_id_offset;
+  const auto filter_payload_owner = extract_cagra_sample_filter<SourceIndexT>(sample_filter);
```
divyegala: Okay yep, so this is what I was talking about. `filter_payload_owner` is allocated on the stack here, which means it will be de-allocated at the end of the function call. For externally managed threading, this means that the device filter payload will be allocated and deallocated on every search call, instead of it being a no-op after the first one.

cpp/src/neighbors/detail/cagra/jit_lto_kernels/cagra_filter_payload.cuh
```diff
@@ -0,0 +1,243 @@
...
+    out.payload.filter_data = filter.filter_data;
+  }
+}
+
+/// Host: fill @ref cagra_sample_filter_payload from a CAGRA filter object.
+template <typename SourceIndexT, typename SampleFilterT>
+cagra_sample_filter_payload<SourceIndexT> extract_cagra_sample_filter(
+  const SampleFilterT& sample_filter)
+{
+  cagra_sample_filter_payload<SourceIndexT> out;
```
divyegala: I think this should be declared `static` before returning

cpp/src/neighbors/detail/cagra/jit_lto_kernels/cagra_filter_payload.cuh
dantegd: Yes, indirectly. The generated JIT kernel sources include kernel_def.hpp / the JIT kernel headers, and those need the cagra_sample_filter<SourceIndexT> parameter type in the kernel signatures. That’s why the small jit_lto_kernels/cagra_filter_payload.cuh header remains in the runtime-compiled/JIT side.

I moved the host-only ownership/cache/extraction logic out of this header into cagra_filter_payload.hpp, so the runtime-compiled side only sees the minimal kernel ABI struct.

cpp/src/neighbors/detail/cagra/cagra_filter_payload.hpp
```diff
@@ -0,0 +1,256 @@
...
+        entries.push_back(selected_state);
+      }
+    }
+
+    return selected_state->dev_ptr(stream);
+  }
+
+ private:
+  mutable std::mutex cache_mutex_;
+  mutable std::unordered_map<cache_key, std::vector<std::shared_ptr<state>>, cache_key_hash> cache_;
```
divyegala: minor nit: `state` does not need to be a `shared_ptr` anymore as the owner itself is `static`

cpp/src/neighbors/detail/cagra/cagra_filter_payload.hpp
```diff
@@ -0,0 +1,256 @@
...
+  : std::true_type {};
+
+template <typename T>
+struct is_udf_filter : std::false_type {};
+
+template <>
+struct is_udf_filter<::cuvs::neighbors::filtering::udf_filter> : std::true_type {};
+
+template <typename SourceIndexT, typename FilterT>
+cagra_filter_data_storage<SourceIndexT> make_cagra_filter_data_storage(const FilterT& filter)
```
divyegala: Ok to address later: name this `make_cagra_bitset_filter_storage` so we can add factories for other pre-compiled types and don't need the specific alias anymore

cpp/src/neighbors/detail/cagra/cagra_filter_payload.hpp
```diff
@@ -0,0 +1,256 @@
...
+    static_cast<SourceIndexT>(bitset_view.size()),
+    static_cast<SourceIndexT>(bitset_view.get_original_nbits())};
+}
+
+template <typename PayloadT>
+void* get_cagra_device_payload(PayloadT payload, cudaStream_t stream)
+{
+  static cagra_device_payload_owner<PayloadT> owner;
+  return owner.dev_ptr(payload, stream);
+}
```
divyegala, line 164: Call this in the factory? So the factory returns a perfectly valid device payload directly

cpp/src/neighbors/detail/cagra/cagra_filter_payload.hpp
```diff
@@ -0,0 +1,256 @@
...
+
+template <typename SourceIndexT, typename FilterT>
+void fill_cagra_sample_filter(cagra_sample_filter<SourceIndexT>& out,
+                              const FilterT& filter,
+                              cudaStream_t stream)
+{
+  using DecayedFilter = std::decay_t<FilterT>;
+  if constexpr (is_bitset_filter<DecayedFilter>::value) {
+    out.filter_data = get_cagra_device_payload(make_cagra_filter_data_storage<SourceIndexT>(filter),
```
divyegala: Would not need to call here then, going by above comment

divyegala (approved):
Minor comments, fantastic PR!

PR #2134, opened 2026-05-27
Upgrade faiss to 1.14.3

### LINKED ISSUE #1848: Apply patches in FAISS repo once they update to cuvs to 25.12+
Once FAISS updates cuvs to 25.12, 26.02, 26.04, apply the corresponding patch diffs in the FAISS repo:
faiss-1.13-cuvs-25.12.diff
faiss-1.13-cuvs-26.02.diff
faiss-1.13-cuvs-26.04.diff
cc @tarang-jain

aamijar:
Faiss has released [v1.14.3](https://github.com/facebookresearch/faiss/releases/tag/v1.14.3) which includes an upgrade to cuVS 26.06. We can now remove the corresponding patches from cuvs and bump the faiss version that we are building against.

Resolves https://github.com/NVIDIA/cuvs/issues/1848


cpp/cmake/patches/faiss-1.14-cuvs-25.12.diff
```diff
@@ -1,182 +0,0 @@
-diff --git a/faiss/gpu/GpuDistance.cu b/faiss/gpu/GpuDistance.cu
-index c82c73e7d..b9100c272 100644
---- a/faiss/gpu/GpuDistance.cu
```
cjnolet, line 3: Why are we maintaining so many diffs? Can't we just use the most recent version?

tarang-jain: yes that is what I was thinking as well. @aamijar lets just maintain a single patch per release. The pinned faiss version is always fixed and the patch should be only tied to that and the current cuvs build.

cpp/cmake/patches/faiss-1.14-cuvs-26.04.diff
```diff
@@ -1,12 +1,25 @@
-diff --git a/faiss/gpu/utils/CuvsUtils.cu b/faiss/gpu/utils/CuvsUtils.cu
```
tarang-jain, line 1: we'd probably want to rename the suffix to `-cuvs-26.08.diff` here for the current release.

aamijar:
I've updated this PR so that we are upgrading to faiss 1.14.3 since that was released recently. All the previous patches have been deleted and now we are maintaining a single `-cuvs-26.08.diff`.

divyegala (approved):

PR #2135, opened 2026-05-27
Fix workspace usage

mfoerste4:
This PR fixes the usage of the workspace resource in the cagra pipeline to increase stability. 
* cagra::optimize - usage of workspace resource for large data where the large workspace resource should be used instead
* ivf-pq - fix of inner/outer batchsize estimate that were unaware of each other and potentially oversubscribing available memory
* kmeans_balanced - limit workspace usage by available memory instead of hard-coded value

CC @tfeher


tfeher (changes requested):
Thanks @mfoerste4 for the PR, overall looks good, but please fix the truncation issues reported by coderabbit.

tfeher (approved):
LGTM, thanks Malte!

PR #2148, opened 2026-06-01
Add HNSW Layered Index Support

julianmi:
The CAGRA graph built by the disk-backed ACE algorithm partitions the dataset. Thus, the CAGRA graph uses the reordered index space. Building a HNSW index using `hnsw::from_cagra` uses the reordered dataset and CAGRA graph. Downstream consumers building an HNSW index would therefore require the reordered dataset, which is typically large when requiring the disk-backed ACE algorithm. Thus, building only the layers of the HNSW index without the dataset and moving this to the search node can minimize the network transfers for downstream consumers if they have the original dataset locally available. The `hnsw::deserialize` step then takes the layered index and combines it with the local dataset to form a hnswlib compatible search index.

## Artifact Layout

```text
hnsw_index.cuvs
  fixed_header
    magic = CUVS_HNSW_LAYERED
    version
    metadata_offset
    metadata_size

  metadata_binary
    dataset shape, dtype, metric
    hnsw parameters
    section sizes
    upper-layer descriptors

  levels
    uint8[n_rows]
    indexed by original dataset row ID

  base_nodes
    uint32[n_rows]
    maps each base topology row to original row ID

  base_links
    n_rows fixed-size hnswlib-ready rows
    [count:uint32][neighbors:uint32[maxM0]]
    neighbors are original row IDs

  upper_nodes
    concatenated uint32 original row IDs for layers 1..maxlevel

  upper_links
    fixed-size hnswlib-ready rows
    [count:uint32][neighbors:uint32[maxM]]
    neighbors are original row IDs
```

## Layered HNSW Serialization

The layered serializer creates `hnsw_index.cuvs` from the disk-backed ACE graph.

1. Create the `.cuvs` file and write the fixed header and metadata.
2. Generate HNSW levels in original ID space.
3. Write `levels` sequentially.
4. Read `dataset_mapping.npy` sequentially into `reordered_to_original`.
5. Read `cagra_graph.npy` source-sequentially in ACE reordered row order.
6. For each ACE graph row:
   - write `base_nodes[row] = reordered_to_original[ace_reordered_row]`
   - convert each neighbor from ACE reordered ID to original ID
   - write a padded hnswlib-ready row to `base_links[row]`
7. Gather promoted vectors from the original dataset.
8. Build upper-layer graphs using temporary HNSW promoted order.
9. Write `upper_nodes` and `upper_links` with node IDs and neighbor IDs converted back to original IDs.

This keeps remapping, link padding, and upper-layer KNN work on the build node.

## Deserialization

The search node reads:

- `hnsw_index.cuvs`
- the external original-order dataset from `index_params.dataset_path`

The loader:

1. Reads the fixed header and metadata.
2. Validates artifact shape, section sizes, and dataset shape.
3. Reads `levels` sequentially.
4. Allocates hnswlib storage.
5. Reads the external dataset sequentially in original row order.
6. Initializes hnswlib with:
   - internal ID = original ID
   - label = original ID
   - level = `levels[original_id]`
7. Reads `base_nodes` and `base_links` sequentially.
8. Copies each base link row into `get_linklist0(base_node_id)`.
9. Reads `upper_nodes` and `upper_links` sequentially by layer.
10. Copies each upper link row into `get_linklist(node_id, level)`.

The search node does no graph remapping, no level generation, no link padding, and no KNN work.

## Disk Access Patterns

Build node:

- Sequential scan of the original dataset for ACE partitioning.
- Buffered partition writes for reordered and augmented datasets.
- Contiguous per-partition reads from `reordered_dataset.npy` and `augmented_dataset.npy`.
- Source-sequential reads from `cagra_graph.npy` when creating the final layered artifact.
- Sequential writes to `hnsw_index.cuvs`.

Search node:

- Sequential reads from `hnsw_index.cuvs`.
- Sequential reads from the external original-order dataset.
- Scatter writes only into in-memory hnswlib link storage by original ID.

## Runtime Requirements

Only `hnsw_index.cuvs` is copied to the search node. ACE temporary files remain build-node-only.

The search node must have the original dataset in original row order and must provide that path through `index_params.dataset_path`.

## Misc

Unifies the logging format of the ACE algorithm.


cpp/bench/ann/src/cuvs/cuvs_cagra_hnswlib.cu
```diff
@@ -27,6 +28,9 @@ auto parse_build_param(const nlohmann::json& conf) ->
       hnsw_params.hierarchy = cuvs::neighbors::hnsw::HnswHierarchy::CPU;
     } else if (conf.at("hierarchy") == "gpu") {
       hnsw_params.hierarchy = cuvs::neighbors::hnsw::HnswHierarchy::GPU;
+    } else if (conf.at("hierarchy") == "gpu_layered_on_disk" ||
+               conf.at("hierarchy") == "gpu_layered" || conf.at("hierarchy") == "layered") {
+      hnsw_params.hierarchy = cuvs::neighbors::hnsw::HnswHierarchy::GPU_LAYERED_ON_DISK;
```
mfoerste4: Are these all just synonyms for `GPU_LAYERED_ON_DISK`?

cpp/bench/ann/src/cuvs/cuvs_cagra_hnswlib_wrapper.h
```diff
@@ -130,6 +146,18 @@ void cuvs_cagra_hnswlib<T, IdxT>::set_search_param(const search_param_base& para
...
 {
+  if (build_param_.hnsw_index_params.hierarchy ==
+      cuvs::neighbors::hnsw::HnswHierarchy::GPU_LAYERED_ON_DISK) {
+    const auto src_artifact = std::filesystem::path(hnsw_index_->file_path());
+    RAFT_EXPECTS(!src_artifact.empty(), "Layered HNSW artifact path is not available.");
+    RAFT_EXPECTS(std::filesystem::exists(src_artifact),
+                 "Layered HNSW artifact '%s' does not exist.",
+                 src_artifact.c_str());
+
+    copy_file_overwrite(src_artifact, std::filesystem::path(file));
```
mfoerste4: Why do we copy here instead of moving (like in the cagra_ace block below?)

cpp/src/neighbors/detail/hnsw.hpp
```diff
@@ -341,6 +357,829 @@ void all_neighbors_graph(raft::resources const& res,
...
+  auto end   = json.find_first_not_of("0123456789-.eE+", begin);
+  RAFT_EXPECTS(begin != std::string::npos, "Malformed double JSON value for key '%s'", key.c_str());
+  return std::stod(json.substr(begin, end - begin));
+}
+
+inline auto json_parse_layer_field(const std::string& object, const std::string& key) -> size_t
+{
+  return json_parse_size(object, key);
+}
+
```
mfoerste4: If we use JSON here - why don't we use a common json utility for parsing? This whole section feels very verbose for a standard. Is there a cleaner alternative?

cpp/src/neighbors/detail/hnsw.hpp
```diff
@@ -341,6 +357,829 @@ void all_neighbors_graph(raft::resources const& res,
...
+    const auto current_node_bytes = current_batch_size * sizeof(IdxT);
+    const auto current_link_bytes = current_batch_size * base_link_row_bytes;
+    cuvs::util::write_large_file(output_fd,
+                                 base_node_buffer.data(),
+                                 current_node_bytes,
+                                 base_nodes_offset + source_start * sizeof(IdxT));
+    cuvs::util::write_large_file(output_fd,
+                                 base_link_buffer.data(),
+                                 current_link_bytes,
+                                 base_links_offset + source_start * base_link_row_bytes);
```
mfoerste4, line 1041: I see that the graph connections / links are shifted to reflect the original ordering, but I don't see the rows/nodes themself being reordered. Is this intended?

cpp/src/neighbors/detail/hnsw.hpp
```diff
@@ -341,6 +357,829 @@ void all_neighbors_graph(raft::resources const& res,
...
+}
+
+template <typename T, typename IdxT, typename Callback>
+void build_hnsw_upper_layer_graphs(
+  raft::resources const& res,
+  raft::host_matrix_view<const T, int64_t, raft::row_major> promoted_dataset,
+  const hnsw_level_plan& plan,
+  size_t M,
+  cuvs::distance::DistanceType metric,
+  Callback&& callback)
```
mfoerste4, line 654: Good to have this extracted as utility.

cpp/src/neighbors/detail/hnsw.hpp
```diff
@@ -341,6 +357,829 @@ void all_neighbors_graph(raft::resources const& res,
...
+              static_cast<hnswlib::linklistsizeint>(layer.degree);
+            std::memcpy(link_row, &list_count, sizeof(list_count));
+            auto* dst = reinterpret_cast<IdxT*>(link_row + sizeof(hnswlib::linklistsizeint));
+            if (layer.degree > 0) {
+              auto* src = host_neighbors.data_handle() + row * layer.degree;
+              for (size_t j = 0; j < layer.degree; ++j) {
+                dst[j] = static_cast<IdxT>(hierarchy.order[src[j] + start_idx]);
+              }
+            }
+          }
```
mfoerste4, line 1283: Ok, here we seem to write in in both the modified data order (not original) and by level connections instead of by row. I assume this gets re-ordered upon deserialize.

cpp/src/neighbors/detail/hnsw.hpp
```diff
@@ -1241,6 +2064,355 @@ void serialize(raft::resources const& res, const std::string& filename, const in
...
+    for (int64_t batch_idx = 0; batch_idx < static_cast<int64_t>(current_batch_size); ++batch_idx) {
+      const auto node_id = static_cast<size_t>(base_node_buffer[batch_idx]);
+      if (node_id >= metadata.n_rows) {
+        invalid_node_id = true;
+        continue;
+      }
+      auto ll0 = appr_algo->get_linklist0(node_id);
+      memcpy(ll0,
+             base_link_buffer.data() + batch_idx * metadata.base_link_row_bytes,
+             metadata.base_link_row_bytes);
```
mfoerste4, line 2801: Ah ok - this is where we re-order upon insertion.

cpp/src/neighbors/detail/hnsw.hpp
```diff
@@ -1241,6 +2064,355 @@ void serialize(raft::resources const& res, const std::string& filename, const in
...
+        }
+        if (layer.level > static_cast<size_t>(levels_u8[node_id])) {
+          invalid_node_level = true;
+          continue;
+        }
+        auto ll = appr_algo->get_linklist(node_id, layer.level);
+        memcpy(ll,
+               link_buffer.data() + batch_idx * metadata.upper_link_row_bytes,
+               metadata.upper_link_row_bytes);
+      }
```
mfoerste4, line 2864: I see, this way we can re-arrange the levels upon load as well. I guess this works as long as we can provide our own deserialize-loader and don't have to mimic the serialized file structure of the original format.

But is this really the use-case? Do we eventually need to support a disk-layered+dataset->disk conversion?

mfoerste4 (approved):
Thanks @julianmi for the PR. The structure looks good to me. Only a few comments that are not necessarily actionable.

cpp/bench/ann/src/cuvs/cuvs_cagra_hnswlib.cu
julianmi: Yes, I've removed them to better align with the other formats.

cpp/bench/ann/src/cuvs/cuvs_cagra_hnswlib_wrapper.h
julianmi: Good catch, thanks. I've changed this and reuse the helper in the `cagra_ace` block.

cpp/src/neighbors/detail/hnsw.hpp
julianmi: This would pull the JSON dependency (`nlohmann/json`) into `libcuvs`. It is currently only used in the benchmarking. I've changed to a binary header format instead given the simplicity. Let me know what you think.

cpp/src/neighbors/detail/hnsw.hpp
julianmi: Yes. The idea is to write sequentially and read scatter into memory. I've added a comment.

cpp/src/neighbors/detail/hnsw.hpp
julianmi: Yes. I've added a comment.

cpp/src/neighbors/detail/hnsw.hpp
julianmi: Yes, this minimizes the I/O for reading and constructing the HNSW index in memory. However, the same approach can be used for the memory bound case where we need to support the disk-layered+dataset->disk conversion you've mentioned. I'm working on a follow-up PR that enables it. The idea is to reorder the base links to original ID space and interleave the dataset vectors by either:

1. Simple scatter if we have enough host memory.
2. File-baked mmap: probably very slow.
3. Write `[id][link row]` records to temporary files in buckets. Replay each bucket and store in small per-bucket scatter buffer. Thus, the base section needs to be written and re-read but disk I/O is sequential. This seems to be the most promising approach when memory constrained. Let me know if you have other ideas please.

julianmi: I've added https://github.com/rapidsai/cuvs/pull/2241 on top of this that implements strategies (1) and (3).

mfoerste4 (approved):
Still looking good to me. @julianmi , thanks for the updates!

cpp/include/cuvs/neighbors/hnsw.hpp
```diff
@@ -35,11 +36,16 @@ namespace graph_build_params = cuvs::neighbors::graph_build_params;
...
+ * which must be set for that path.
  */
 enum class HnswHierarchy {
-  NONE,  // base-layer-only index
-  CPU,   // full index with CPU-built hierarchy
-  GPU    // full index with GPU-built hierarchy
+  NONE,                // base-layer-only index
+  CPU,                 // full index with CPU-built hierarchy
+  GPU,                 // full index with GPU-built hierarchy
+  GPU_LAYERED_ON_DISK  // GPU-built hierarchy stored as layered on-disk topology
```
tfeher: I am not sure if this is the right place for this arg. The previous options define whether / where we calculate the hiararchy. The new option defines the serialization format.

cpp/include/cuvs/neighbors/hnsw.hpp
```diff
@@ -60,6 +66,16 @@ struct index_params : cuvs::neighbors::index_params {
...
 
+  /** Local dataset path used by layered HNSW deserialization.
+   *
+   * When `hierarchy == HnswHierarchy::GPU_LAYERED_ON_DISK`, the index artifact stores graph
+   * topology only. `deserialize` loads vectors from this local dataset path to reconstruct an
+   * in-memory HNSW index.
+   * Currently supported local dataset formats are `.npy` and ANN benchmark `*.bin` files with a
+   * `[uint32 rows, uint32 cols]` header.
+   */
+  std::string dataset_path;
```
tfeher: Why is this part of index_params? Normally index_params stores parameter's for index construction. The dataset is an additional input argument for the build / serialize methods.

examples/cpp/src/hnsw_ace_layered_example.cu
```diff
@@ -0,0 +1,278 @@
...
+  auto hnsw_params  = make_hnsw_ace_params(kBuildDir, dataset_path);
+
+  std::cout << "[stage 2] Build layered HNSW index with ACE" << std::endl;
+  auto artifact_path = hnsw_build<int8_t>(dev_resources, hnsw_params, dataset_i8_view);
+
+  std::cout << "[stage 3] Deserialize layered HNSW index" << std::endl;
+  auto hnsw_index = hnsw_deserialize<int8_t>(dev_resources, hnsw_params, artifact_path, n_dim);
+
+  std::cout << "[stage 4] Search HNSW index" << std::endl;
+  hnsw_search<int8_t>(dev_resources, *hnsw_index, queries_i8_view);
```
tfeher: I think it would be more interesting to demonstrate that after we have built the graph using the quantized data, we can build the HNSW index using the original full precision dataset.

cpp/src/neighbors/detail/hnsw.hpp
```diff
@@ -465,6 +495,801 @@ void all_neighbors_graph(raft::resources const& res,
...
 }
 
+struct hnsw_level_plan {
+  size_t n_rows = 0;
+  std::vector<size_t> hist;
+  // Bucket-end offsets after order construction. Level L starts at offsets[L - 1].
+  std::vector<size_t> offsets;
+  std::vector<size_t> order;
+  std::vector<size_t> order_bw;
+  std::vector<uint8_t> levels;
```
tfeher: Please add docstring. It is not clear to me what is `order`.

tfeher: We would want to include these in the memory estimates. Recently, [raft dry-run](https://github.com/NVIDIA/raft/pull/2961) was merged, that would allow tracking these allocations. @achirkin Do we need to use a specific host allocator so that these are captured during dry-run?

tfeher (commented):
Thanks Julian for this PR! This is an important feature that will reduce data transfer times while reading ACE results. I am still reviewing, but sharing the first few comments.

I think we need more clarity in how we name this option, and when it is used. Even if we build in memory (without ACE), it would be useful to save only the graph (and hierarchy) to file.

cpp/src/neighbors/detail/hnsw.hpp
achirkin: std::vector is not captured by default. We have to change all these to `raft::host_vector<..>` to properly count them.
An alternative workaround on the user side is to set the raft memory resource globally in the application:
```C++
struct pmr_proxy : public std::pmr::memory_resource
{
    void* do_allocate(std::size_t bytes, std::size_t alignment) override
    {
        return raft::get_default_host_resource().allocate_sync(bytes, alignment);
    }

    void do_deallocate(void* p, std::size_t bytes, std::size_t alignment) override
    {
        return raft::get_default_host_resource().deallocate_sync(p, bytes, alignment);
    }
};

pmr_proxy global_pmr{};

int main() {
  std::pmr::set_default_resource(*global_pmr);
}
```

cpp/include/cuvs/neighbors/hnsw.hpp
julianmi: I've removed this option in favor of a new `HnswOutputFormat` parameter as discussed offline. The default format is `HNSWLIB`. The proposed layered index would now use:
```
hnsw_params.hierarchy = hnsw::HnswHierarchy::GPU;
hnsw_params.output_format = hnsw::HnswOutputFormat::CUVS_LAYERED_TOPOLOGY;
```

cpp/include/cuvs/neighbors/hnsw.hpp
julianmi: This was a workaround to not have to change the `hnsw::deserialize` interface. I agree that this is confusing and changed it to a new interface:
```
void deserialize(raft::resources const& res,
                 const index_params& params,
                 const std::string& filename,
                 int dim,
                 index<T>** index);
```

examples/cpp/src/hnsw_ace_layered_example.cu
julianmi: I've changed the hardcoded data type in the format. I'll run some experiments to verify that this works. I might need to extend the materialization API proposed in the follow-up PR https://github.com/NVIDIA/cuvs/pull/2241 as well.

cpp/src/neighbors/detail/hnsw.hpp
julianmi: I've added the docstring and renamed to variables. Let me know if this is still unclear.

I've also changed the `std::vector` to `raft::host_vector`.

cpp/include/cuvs/neighbors/hnsw.hpp
julianmi: I've removed this option in favor of a new HnswOutputFormat parameter as discussed offline. The default format is HNSWLIB. The proposed layered index would now use:
```
hnsw_params.hierarchy = hnsw::HnswHierarchy::GPU;
hnsw_params.output_format = hnsw::HnswOutputFormat::CUVS_LAYERED_TOPOLOGY;
```

cpp/include/cuvs/neighbors/hnsw.hpp
julianmi: This was a workaround to not have to change the hnsw::deserialize interface. I agree that this is confusing and changed it to a new interface:
```
void deserialize(raft::resources const& res,
                 const index_params& params,
                 const std::string& filename,
                 int dim,
                 index<T>** index);
```

cpp/src/neighbors/detail/hnsw.hpp
julianmi: I've added the docstring and renamed to variables. Let me know if this is still unclear.
I've also changed the `std::vector` to `raft::host_vector`.

examples/cpp/src/hnsw_ace_layered_example.cu
julianmi: I've changed the hardcoded data type in the format. I'll run some experiments to verify that this works. I might need to extend the materialization API proposed in the follow-up PR [#2241](https://github.com/NVIDIA/cuvs/pull/2241) as well.

cpp/include/cuvs/neighbors/hnsw.hpp
```diff
@@ -1331,6 +1404,53 @@ void deserialize(raft::resources const& res,
...
+ *
+ * The topology artifact supplies the index dimensions, metric, and construction metadata. The
+ * attached dataset must have the recorded shape, but its element type may differ from the type used
+ * to construct the graph. The output pointer selects the attached dataset type. The loader accepts
+ * row-major `.npy` files and ANN benchmark binary files with a `[uint32 rows, uint32 cols]` header.
+ * Binary extensions must match the output index type: `.fbin` for `float`, `.f16bin` or
+ * `.fp16.fbin` for `half`, `.u8bin` for `uint8_t`, and `.i8bin` for `int8_t`.
+ *
+ * @param[in] res raft resources
+ * @param[in] topology_filename path to the layered HNSW topology artifact
```
tfeher: Can we just call it `graph_filename`? Ok, it is more than  a graph, but conceptually it would be simpler to understand.

cpp/include/cuvs/neighbors/hnsw.hpp
```diff
@@ -37,20 +37,36 @@ namespace graph_build_params = cuvs::neighbors::graph_build_params;
...
+ * @brief Output artifact format for an HNSW index
+ *
+ * `HNSWLIB` produces the standard hnswlib index format. `CUVS_LAYERED_TOPOLOGY` stores graph
+ * topology separately from vectors. The current implementation requires `HnswHierarchy::GPU` and
+ * disk-backed ACE. Load a layered topology with the two-filename `deserialize` overload, which
+ * reads the vectors from a separate local dataset.
+ */
+enum class HnswOutputFormat {
+  HNSWLIB,
+  CUVS_LAYERED_TOPOLOGY,
```
tfeher: Can we keep it short and simple?
```suggestion
GRAPH_ONLY
```

tfeher (changes requested):
Thanks Julian, the implementation looks good now, but I think we can still improve parameter naming to make the API easier to understand for new users.

cpp/include/cuvs/neighbors/hnsw.hpp
julianmi: Renamed to `graph_filename`.

cpp/include/cuvs/neighbors/hnsw.hpp
julianmi: Renamed to `GRAPH_ONLY`.

julianmi:
> Thanks Julian, the implementation looks good now, but I think we can still improve parameter naming to make the API easier to understand for new users.

Thanks, @tfeher. I like these simplifications. I think I've addressed all comments. Please double-check and approve/merge this PR if you're happy with it.

Please note that I had to rebase and force-push since the 26.12 tag was reachable before due to the merged main branch. This failed the `pr/checks/other-checks` test.

tfeher (approved):
Thanks Julian for the updates, LGTM!

PR #2151, opened 2026-06-01
fix(ci): fix configuration for breaking change notification workflow

jameslamb:
Contributes to https://github.com/rapidsai/shared-workflows/issues/560

Fixes misconfigurations in the GitHub Actions workflow that generates Notifications when PRs are labeled `breaking`.

Any other changes come from making the configuration for this workflow identical across all RAPIDS repos.

### DIFF
--- a/.github/workflows/trigger-breaking-change-alert.yaml
+++ b/.github/workflows/trigger-breaking-change-alert.yaml
@@ -11,6 +11,8 @@ on: # zizmor: ignore[dangerous-triggers]
       - labeled
       - unlabeled
 
+permissions: {}
+
 jobs:
   trigger-notifier:
     if: contains(github.event.pull_request.labels.*.name, 'breaking')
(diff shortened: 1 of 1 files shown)


.github/workflows/trigger-breaking-change-alert.yaml
```diff
@@ -11,14 +11,16 @@ on: # zizmor: ignore[dangerous-triggers]
       - labeled
       - unlabeled
 
+permissions: {}
```
jakirkham, line 14: Looks like some other repos (like RMM and cuDF) don't have this. Should we add this throughout?

jakirkham (approved):
Thanks James! 🙏

.github/workflows/trigger-breaking-change-alert.yaml
jameslamb: (repeating what I said offline for the history here)

Only `rmm` and `cudf` specifically were missing this. I've added it in both.

* https://github.com/rapidsai/rmm/pull/2438
* https://github.com/rapidsai/cudf/pull/22732

PR #2217, opened 2026-06-04
Informative warnings for CAGRA construction with low intermediate degree

### LINKED ISSUE #2033: [BUG] Better warning / assert and handling of cases where CAGRA graph degree is not strictly larger than intermediate graph degree.
Currently, CAGRA does an assertion that graph degree is not less than intermediate graph degree, but even if graph degree is too close to intermediate graph degree, it's possible there's not enough neighborhood information encoded in the intermediate knn graph for the pruning to have a connected graph. 

When this case happens, the user gets an error that's nebulus and doesn't provide enough information for the user to 

1) know what happened, and 
2) fix the issue. 

In general, users aren't going to know the complex relationships between parameters like this, but we often make a suggestion that intermediate graph degree by 1.5 * graph degree.

landrumb:
Resolves #2033 with more informative warnings about the intermediate graph degree parameter.


divyegala:
Could you make the PR title more specific and descriptive please? As we use squash commits, the PR title acts as the commit message after the PR is merged.

divyegala (approved):

cpp/src/neighbors/detail/cagra/graph_core.cuh
```diff
@@ -1679,11 +1679,18 @@ void prune_graph_gpu(
   raft::copy(res, host_stats.view(), raft::make_const_mdspan(dev_stats.view()));
   raft::resource::sync_stream(res);
 
+  const size_t recommended_intermediate = output_graph_degree + ((output_graph_degree + 1) / 2);
```
cjnolet: Rather than duplicating this, it would be better to put this in a shared location and also keep it out of the immediate usage. If this is defined somewhere else already, we should use that instead of redefining. 

For example, if someone ever has to change this value, they are likely not going to know that it exists in multiple locations. Business logic like this should be centralized as much as possible.

landrumb: moved to a constexpr in `cagra::utils`

dantegd:
/ok to test [0625256](https://github.com/rapidsai/cuvs/pull/2217/commits/06252569e7236294c27778fb80e6eef54fbd93a1)

PR #2222, opened 2026-06-09
Improve balanced k-means rebalancing

anaruse:
## Motivation

The current balanced k-means rebalancing heuristic is not always effective at reducing oversized partitions.

This PR improves the rebalancing heuristic, introduces separate lower and upper balance tolerances, and adds a C++ example for evaluating partition balance.

### Balanced k-means parameters

Added separate lower and upper balance tolerance parameters:

- `balance_lower_tolerance`
- `balance_upper_tolerance`

The thresholds are computed as:

```text
lower_threshold = average_partition_size * balance_lower_tolerance
upper_threshold = average_partition_size * balance_upper_tolerance
```

This allows underflow and overflow handling to be tuned independently.

### Rebalancing improvements

Improved the rebalancing heuristic by explicitly pairing underfull and overfull partitions.

New centroids are created near oversized partitions, resulting in a more balanced final partition size distribution.

### Example updates

Added a C++ example for evaluating balanced k-means partition balance. The example:

- accepts multiple lower and upper tolerance values in a single run
- runs regular k-means once as a baseline
- reports partition size statistics for regular and balanced k-means
- reports underflow and overflow partition counts
- prints histograms using a shared range for easier comparison

### Documentation

Updated the C++ API documentation and example documentation for the new balanced k-means parameters and defaults.

## Testing

- Built libcuvs successfully
- Built the balanced k-means C++ example successfully
- Evaluated SIFT-1M, GloVe, Wiki, and DEEP datasets
- Compared partition size statistics between regular and balanced k-means

No issue is closed by this PR.


anaruse:
I evaluated the updated balanced k-means implementation on the Wiki dataset (1M vectors, 1000 partitions).

The previous implementation (roughly corresponding to balance_lower_tolerance=0.25 and balance_upper_tolerance=4.0) already improves partition balance significantly, reducing the maximum partition size from 15.4× the average to 3.75× and lowering the standard deviation from 1039 to 374.

The new default configuration (balance_lower_tolerance=0.333, balance_upper_tolerance=3.0) further improves the partition size distribution, reducing the maximum partition size to 2.73× the average and lowering the standard deviation to 334.

For applications that require tighter balance, more aggressive settings (balance_lower_tolerance=0.5, balance_upper_tolerance=2.0) can further reduce the maximum partition size to 2.34× the average and lower the standard deviation to 268.

These results show that the updated rebalancing heuristic produces a more balanced partition size distribution and that the newly exposed lower and upper balance tolerances provide a useful mechanism for trading off balance strictness against flexibility.

```
# dtype: float
# partitions: 1000
# iterations: 20
# balance_lower_tolerances: 0.25 0.333 0.5
# balance_upper_tolerances: 4 3 2
# centroid_offset: 0.01
Partitioning 1000000 vectors with 768 dimensions into 1000 balanced partitions

Regular k-means partition size statistics: min=1, max=15367, median=786.5, mean=1000, standard deviation=1039.03, min/mean=0.001, max/mean=15.367
Regular k-means partition size histogram:
  [       1,      155]   12 | ##
  [     154,      309]   37 | ########
  [     308,      463]  115 | ###########################
  [     462,      617]  152 | ####################################
  [     616,      771]  166 | ########################################
  [     770,      925]  140 | #################################
  [     924,     1078]  101 | ########################
  [    1077,     1232]   77 | ##################
  [    1231,     1386]   51 | ############
  [    1385,     1540]   38 | #########
  [    1539,     1694]   20 | ####
  [    1693,     1848]   19 | ####
  [    1847,     2002]   17 | ####
  [    2001,     2155]   11 | ##
  [    2154,     2309]    4 | #
  [    2308,     2463]    1 | #
  [    2462,     2617]    6 | #
  [    2616,     2771]    0 | 
  [    2770,     2925]    6 | #
  [    2924,     3079]    4 | #
  (    3079,      inf]   23 | #####

# balance_lower_tolerance: 0.25
# balance_upper_tolerance: 4
Balanced k-means partition size statistics: min=228, max=3751, median=945, mean=1000, standard deviation=373.842, min/mean=0.228, max/mean=3.751, underflow=1 (< 250), overflow=0 (> 4000)
Balanced k-means partition size histogram:
  [       1,      155]    0 | 
  [     154,      309]    4 | #
  [     308,      463]   32 | #####
  [     462,      617]   73 | #############
  [     616,      771]  150 | ###########################
  [     770,      925]  220 | ########################################
  [     924,     1078]  183 | #################################
  [    1077,     1232]  129 | #######################
  [    1231,     1386]   86 | ###############
  [    1385,     1540]   51 | #########
  [    1539,     1694]   26 | ####
  [    1693,     1848]   16 | ##
  [    1847,     2002]   11 | ##
  [    2001,     2155]    9 | #
  [    2154,     2309]    2 | #
  [    2308,     2463]    1 | #
  [    2462,     2617]    2 | #
  [    2616,     2771]    1 | #
  [    2770,     2925]    2 | #
  [    2924,     3079]    0 | 
  (    3079,      inf]    2 | #
Balance improvement: max/mean 15.367 -> 3.751, standard deviation 1039.03 -> 373.842

# balance_lower_tolerance: 0.333
# balance_upper_tolerance: 3
Balanced k-means partition size statistics: min=183, max=2730, median=975.5, mean=1000, standard deviation=333.736, min/mean=0.183, max/mean=2.73, underflow=1 (< 333), overflow=0 (> 3000)
Balanced k-means partition size histogram:
  [       1,      155]    0 | 
  [     154,      309]    1 | #
  [     308,      463]   26 | #####
  [     462,      617]   84 | ################
  [     616,      771]  142 | ############################
  [     770,      925]  179 | ###################################
  [     924,     1078]  201 | ########################################
  [    1077,     1232]  156 | ###############################
  [    1231,     1386]  104 | ####################
  [    1385,     1540]   46 | #########
  [    1539,     1694]   29 | #####
  [    1693,     1848]   14 | ##
  [    1847,     2002]    6 | #
  [    2001,     2155]    6 | #
  [    2154,     2309]    2 | #
  [    2308,     2463]    3 | #
  [    2462,     2617]    0 | 
  [    2616,     2771]    1 | #
  [    2770,     2925]    0 | 
  [    2924,     3079]    0 | 
  (    3079,      inf]    0 | 
Balance improvement: max/mean 15.367 -> 2.73, standard deviation 1039.03 -> 333.736

# balance_lower_tolerance: 0.5
# balance_upper_tolerance: 2
Balanced k-means partition size statistics: min=357, max=2336, median=983.5, mean=1000, standard deviation=267.705, min/mean=0.357, max/mean=2.336, underflow=8 (< 500), overflow=1 (> 2000)
Balanced k-means partition size histogram:
  [       1,      155]    0 | 
  [     154,      309]    0 | 
  [     308,      463]    3 | #
  [     462,      617]   61 | ##########
  [     616,      771]  140 | #######################
  [     770,      925]  201 | #################################
  [     924,     1078]  241 | ########################################
  [    1077,     1232]  168 | ###########################
  [    1231,     1386]  107 | #################
  [    1385,     1540]   43 | #######
  [    1539,     1694]   22 | ###
  [    1693,     1848]   10 | #
  [    1847,     2002]    3 | #
  [    2001,     2155]    0 | 
  [    2154,     2309]    0 | 
  [    2308,     2463]    1 | #
  [    2462,     2617]    0 | 
  [    2616,     2771]    0 | 
  [    2770,     2925]    0 | 
  [    2924,     3079]    0 | 
  (    3079,      inf]    0 | 
Balance improvement: max/mean 15.367 -> 2.336, standard deviation 1039.03 -> 267.705

File: cpp/src/cluster/detail/kmeans_balanced.cuh [hunk 1]
```diff
@@ -567,36 +571,70 @@ auto adjust_centers(MathT* centers,
                                       601,  659,  733,  809,  863,  941,  1013, 1069, 1151, 1223,
                                       1291, 1373, 1451, 1511, 1583, 1657, 1733, 1811, 1889, 1987,
                                       2053, 2129, 2213, 2287, 2357, 2423, 2531, 2617, 2687, 2741};
-  static IdxT i        = 0;
   static IdxT i_primes = 0;
 
-  bool adjusted = false;
-  IdxT average  = n_rows / n_clusters;
+  IdxT average         = n_rows / n_clusters;
```
dantegd: Shouldn't this be computed this as a fp average? The public docs/example define the thresholds as average_partition_size * tolerance, but this integer division rounds down before applying the new tolerances. That makes the knobs behave a bit differently from what users set, and could affect smaller or uneven fine-cluster training sets, no?

examples/cpp/src/balanced_kmeans_example.cu
```diff
@@ -0,0 +1,573 @@
...
+  }
+
+  if (desc.file_format == BIGANN) {
+    std::fprintf(stderr, "# BIGANN type file (%s)\n", file_path.c_str());
+    desc.size = tmp_val[0];
+    desc.dim  = tmp_val[1];
+  } else {
+    std::fprintf(stderr, "# Xvec type file (%s)\n", file_path.c_str());
+    desc.dim  = tmp_val[0];
+    desc.size = (file_size_in_byte - sizeof(std::uint32_t)) / desc.dim / sizeof(DataT) - 1;
```
dantegd: xvec files have a uint32_t dim prefix on every row, right? So the row size is sizeof(uint32_t) + dim * sizeof(DataT). This formula ignores most of those per-row headers and will overestimate desc.size, no?  This would mean the example can read past EOF and report balance stats on invalid data.

examples/cpp/src/balanced_kmeans_example.cu
```diff
@@ -0,0 +1,573 @@
...
+  }
+
+  auto const array_size = sizeof(DataT) * desc.dim * desc.size;
+  desc.data             = std::make_unique<DataT[]>(desc.dim * desc.size);
+
+  if (desc.file_format == BIGANN) {
+    ifs.seekg(sizeof(std::uint32_t) * 2, std::ios::beg);
+    ifs.read(reinterpret_cast<char*>(desc.data.get()), array_size);
+  } else {
+    ifs.seekg(sizeof(std::uint32_t), std::ios::beg);
```
dantegd: I _think_ the xvec read path is misaligned. We seek past the first dimension word before the loop, then skip another uint32_t at the start of the first iteration, so the first vector starts 4 bytes late and all following rows are framed incorrectly. What do you think about starting at the beginning of the file and, for each row, reading/skipping the row dimension immediately before reading that row’s vector?

examples/cpp/src/balanced_kmeans_example.cu
```diff
@@ -0,0 +1,573 @@
...
+    resources, dataset_view, n_partitions, n_iters, regular_labels.view());
+  std::optional<partition_size_stats> regular_reference_stats;
+  if (has_regular_stats) {
+    regular_reference_stats =
+      compute_partition_size_stats(resources,
+                                   n_partitions,
+                                   raft::make_const_mdspan(regular_labels.view()),
+                                   balance_lower_tolerances.front(),
+                                   balance_upper_tolerances.front());
+    print_partition_size_summary("Regular k-means", regular_reference_stats.value());
```
dantegd, line 443: Small thing: the regular baseline only prints the summary once, while balanced stats are evaluated per tolerance pair, would it be useful to report regular k-means underflow/overflow counts for each lower/upper tolerance combination too?

cpp/src/cluster/detail/kmeans_balanced.cuh
```diff
@@ -567,36 +571,70 @@ auto adjust_centers(MathT* centers,
...
+  auto lower_threshold = static_cast<MathT>(average) * balance_lower_tolerance;
+  auto upper_threshold = static_cast<MathT>(average) * balance_upper_tolerance;
+  std::vector<CounterT> host_cluster_sizes(n_clusters);
+  raft::update_host(host_cluster_sizes.data(), cluster_sizes, n_clusters, stream);
+  raft::resource::sync_stream(handle, stream);
```
dantegd, line 646: Does the donor/receiver pairing have to happen on the host? Sorry if I’m missing something obvious, but since this is inside the balancing loop, it looks like we pay a device/host round trip plus a CPU sort on each rebalancing pass. 

Since this happens inside the balancing loop, it adds a CPU sort plus a device/host sync on each rebalancing pass, and this path is also used by IVF-PQ/IVF-SQ index builds.

cpp/src/cluster/detail/kmeans_balanced.cuh
```diff
@@ -567,36 +571,70 @@ auto adjust_centers(MathT* centers,
...
+                                                                        donor_clusters.data(),
+                                                                        centroid_offset,
                                                                         ofst,
                                                                         update_count.data(),
                                                                         mapping_op);
-  adjusted = update_count.value(stream) > 0;  // NB: rmm scalar performs the sync
-
-  return adjusted;
+  auto n_updates = update_count.value(stream);  // NB: rmm scalar performs the sync
+  RAFT_EXPECTS(n_updates == n_pairs, "Balanced k-means failed to update all adjusted centers");
```
dantegd, line 723: This can throw an exception somce the pairing loop never checks donor size, so if more that half the clusters are empty we pair against an empty donor, the kernel finds no point, and this fires. It's hittable from build_fine_clusters. Could we just guard or break when the donor is empty?

cpp/src/cluster/detail/kmeans_balanced.cuh
```diff
@@ -459,61 +460,57 @@ template <uint32_t BlockDimY,
...
-    } while (static_cast<IdxT>(cluster_sizes[labels[i]]) < average);
+  IdxT pair_id = threadIdx.y + BlockDimY * static_cast<IdxT>(blockIdx.x);
+  if (pair_id >= n_pairs) return;
+
+  auto receiver_cluster = receiver_clusters[pair_id];
+  auto donor_cluster    = donor_clusters[pair_id];
+  IdxT i                = n_rows;
+  IdxT j                = raft::laneId();
+  for (IdxT attempt = 0; attempt < n_rows; attempt += raft::WarpSize) {
+    auto candidate = (seed * (attempt + j + 1) + pair_id) % n_rows;
```
dantegd: If IdxT = int, this can overflow, no? With seed=2741 for example,  once the dataset is past ~780k rows, 2741 * n_rows ≈ 2.7e9. Can't remember, do we always use int64 Idxt?

dantegd (changes requested):
The improvement approach looks solid to me, really nice! Just had some questions and smaller comments around the example and code

cpp/src/cluster/detail/kmeans_balanced.cuh
anaruse: Good point. The threshold should be computed from a floating-point average rather than from an `IdxT` value, since the public parameter semantics are based on `average_partition_size * tolerance`.

I updated this to compute `average` as `MathT` before applying the lower and upper tolerances.

cpp/src/cluster/detail/kmeans_balanced.cuh
anaruse: Thanks for catching this. You are right: if the larger side of the pair is empty, there is no valid donor cluster and the kernel cannot find a point to reseed from.

I updated the pairing loop to break when the donor-side cluster size is 0.

cpp/src/cluster/detail/kmeans_balanced.cuh
anaruse: Thanks for catching this. You are right that the intermediate expression can overflow when `IdxT` is a 32-bit integer.

I updated the candidate index calculation so the intermediate arithmetic is performed with `int64_t` before converting the result back to `IdxT`.

examples/cpp/src/balanced_kmeans_example.cu
anaruse: That is a good point. Regular k-means itself does not use these tolerance values, but I originally reported the regular baseline underflow/overflow counts for each balanced k-means tolerance pair to make the comparison direct.

In practice, that made the output quite verbose when evaluating multiple tolerance combinations, so I changed the example to print the regular k-means summary once as a baseline and then print the balanced k-means stats for each tolerance pair.

cpp/src/cluster/detail/kmeans_balanced.cuh
anaruse: Good question. The pairing is currently done on the host intentionally to keep the implementation simple. We only sort `n_clusters` cluster-size entries, not the input vectors, and the main cost of balanced k-means is still expected to be the assignment/prediction work over the dataset.

A device-side sort would avoid the host round trip, but it would also add extra temporary storage and more complex device-side pairing logic. I think the host-side pairing is a reasonable first implementation, but I agree this is worth profiling, especially because this path is used by IVF-PQ/IVF-SQ builds. If it shows up as a bottleneck, we can move the pairing to the device in a follow-up.

examples/cpp/src/balanced_kmeans_example.cu
anaruse: Thanks for pointing this out. You were right: the xvec handling had a few issues.

I updated the size calculation to account for the per-row dimension header, and changed the read path to read each row from the beginning as `dim` followed by the vector payload. I also added checks for inconsistent row dimensions and read failures.

examples/cpp/src/balanced_kmeans_example.cu
anaruse: Fixed together with the xvec read path issue below. The size calculation now uses the full row size, including the per-row dimension header.

anaruse:
Thanks for the review! I addressed the main feedback in the latest commit and replied to the regular baseline reporting comment. Could you take another look when you have a chance?

dantegd (approved):

cpp/src/cluster/detail/kmeans_balanced.cuh
```diff
@@ -549,13 +553,16 @@ template <typename T,
           typename CounterT,
           typename MappingOpT>
 auto adjust_centers(MathT* centers,
+                    const raft::resources& handle,
```
tarang-jain: Can we avoid having both -- the handle and stream as arguments? Also, lets make the handle the first argument of the function.

anaruse: Thanks, I updated this accordingly. The handle is now the first argument, and the separate stream argument has been removed.

examples/cpp/src/balanced_kmeans_example.cu
```diff
@@ -0,0 +1,582 @@
...
+          underflow_count,
+          overflow_count,
+          median,
+          mean,
+          std::sqrt(variance),
+          lower_threshold,
+          upper_threshold};
+}
+
+void print_partition_size_stats(std::string const& label, partition_size_stats const& stats)
```
huuanhhuyn, line 257: Should we measure and evaluate inertia change after rebalancing? I guess it also helps us select/tune the optimal upper and lower tolerances.

anaruse: Thanks for the suggestion! I updated the example to report the inertia after rebalancing as well. This should help us evaluate the trade-off between partition balance and clustering quality, and also make it easier to tune the lower/upper thresholds.

cjnolet (approved):

huuanhhuyn (approved):

cjnolet:
@anaruse, I'd love to get these changes merged. It looks like there's some failures in the C++ tests (potentially related to these changes).

anaruse:
@cjnolet, Thanks for pointing this out. I’ve started investigating the C++ test failures.

This PR improves cluster-size balance at the cost of slightly worse inertia. The failures seem to be caused by a small recall regression in IVF-PQ, which uses balanced k-means for codebook construction. Since inertia is likely more important than cluster-size balance for codebook quality, this trade-off may be the cause. I’m currently verifying this hypothesis.

anaruse:
I found that the donor-selection strategy used in the existing balanced k-means implementation works better for PQ codebook training. Therefore, I restored the implementation currently used on the main branch and changed the PQ codebook generation path to use that version of balanced k-means. With this change, the IVF-PQ CI failures no longer reproduce in my local environment.

anaruse:
It looks like `CLUSTER_TEST` failed in CI, but I was unable to reproduce the failure in my local environment. Could you please rerun the CI?

PR #2229, opened 2026-06-09
feat(rust): add IVF-SQ bindings

jamie8johnson:
Adds Rust bindings for the IVF-SQ index, wrapping the C API introduced in v26.06 ("IVF-SQ C API"). Follow-up to the CAGRA serialize/deserialize and `search_with_filter` wrappers (#2019 lineage).

## What

- New `rust/cuvs/src/ivf_sq/` module mirroring the `cagra` module structure: `IndexParams` / `SearchParams` builders, `Index` with `build`, `search`, `search_with_filter` (`cuvsFilter` bitset), `extend`, `serialize` / `deserialize`, and the `n_lists` / `dim` / `size` getters.
- `rust/cuvs-sys/src/bindings.rs` regenerated via `rust/scripts/generate-bindings.sh` — purely additive (`cuvsIvfSq*` block).

## Notes for reviewers

- `cuvsIvfSqIndexGetCenters` is deliberately not wrapped, matching the cagra and ivf_pq modules — it fills a caller-allocated `DLManagedTensor`, which has no ergonomic safe-Rust shape yet (commented in `index.rs`).
- IVF-SQ exposes getters + `extend` where the ivf_pq Rust module currently doesn't; the C API supports them for both, so this module simply wraps the full practical surface. Happy to trim for symmetry if preferred.

## Testing

9 GPU tests (single-process): params builders, build + self-neighbor search, repeated search, getters, extend-from-empty (NULL indices path), bitset-filtered search asserting exclusion, serialize/deserialize round-trip re-verifying search on the loaded index, interior-NUL path rejection. `cargo fmt` / clippy clean. Validated against both the v26.06.00 tag (conda libcuvs 26.06) and current `main` — the `ivf_sq.h` header is byte-identical between the two.

### DIFF
--- a/rust/cuvs-sys/src/bindings.rs
+++ b/rust/cuvs-sys/src/bindings.rs
@@ -1765,6 +1765,182 @@ unsafe extern "C" {
         index: cuvsIvfFlatIndex_t,
     ) -> cuvsError_t;
 }
+#[doc = " @defgroup ivf_sq_c_index_params IVF-SQ index build parameters\n @{\n/\n/**\n @brief Supplemental parameters to build IVF-SQ Index\n"]
+#[repr(C)]
+#[derive(Debug, Copy, Clone)]
+pub struct cuvsIvfSqIndexParams {
+    #[doc = " Distance type."]
+    pub metric: cuvsDistanceType,
+    #[doc = " The argument used by some distance metrics."]
+    pub metric_arg: f32,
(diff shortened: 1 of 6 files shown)


jamie8johnson:
Rebased all five Rust binding PRs — #2229 (IVF-SQ), #2230 (refine), #2231 (brute-force serialize), #2234 (scalar quantizer), #2235 (tiered index) — onto current `main` and ported each to the new `AsDlTensor`/`AsDlTensorMut` lifetime-tracked tensor API from #2245, which they'd gone stale/conflicting against. Each is now a single squashed commit scoped to just its feature and compiles clean (`cargo check`/`clippy`/`fmt`); all five are back to MERGEABLE.

They're independent but related bindings, so they can be reviewed together whenever you have a chance. If it helps triage: label-wise they should all be `feature request` + `non-breaking` + `Rust` (matching #2245's convention) — happy to adjust anything on review. Thanks!

PR #2234, opened 2026-06-10
feat(rust): add scalar quantizer bindings

jamie8johnson:
# Add Rust bindings for the scalar quantizer

## What

This PR adds idiomatic Rust bindings for the scalar quantizer preprocessing
API (`cuvsScalarQuantizer*`) exposed by
`c/include/cuvs/preprocessing/quantize/scalar.h`.

It introduces a new `preprocessing` module tree under `rust/cuvs/src/` and wraps
the full scalar quantizer lifecycle:

- `ScalarQuantizerParams` — RAII wrapper around `cuvsScalarQuantizerParams_t`
  with a `set_quantile` builder method and a balanced `Drop`
  (`cuvsScalarQuantizerParamsCreate` / `Destroy`).
- `Quantizer` — RAII wrapper around `cuvsScalarQuantizer_t` with a balanced
  `Drop` (`cuvsScalarQuantizerCreate` / `Destroy`) and the three operations:
  - `Quantizer::train` — trains the quantizer on an `f32` dataset
    (`cuvsScalarQuantizerTrain`).
  - `Quantizer::transform` — quantizes an `f32` matrix into `i8`
    (`cuvsScalarQuantizerTransform`).
  - `Quantizer::inverse_transform` — reconstructs an approximate `f32` matrix
    from `i8` quantized data (`cuvsScalarQuantizerInverseTransform`).

A supporting `IntoDtype for i8` implementation was added to `dlpack.rs` so that
`i8` quantized tensors can be passed through `ManagedTensor`.

The bindings follow the existing sibling module conventions exactly:
`MaybeUninit` + `assume_init` only after a successful `check_cuvs`, no
`Copy`/`Clone` on the handle wrappers, balanced create/destroy in `Drop`, and a
custom `Debug` impl for the params type. A runnable doc example lives in
`preprocessing/quantize/mod.rs`.

## Notes for reviewers

- **Establishes the `preprocessing` module tree.** This PR creates
  `preprocessing/mod.rs` and `preprocessing/quantize/mod.rs` with only the
  `scalar` path wired up. The binary and product-quantizer (PQ) C APIs
  (`quantize/binary.h`, `quantize/pq.h`) are intentionally **out of scope** for
  this PR and are left for follow-up contributions; the module tree is laid out
  so they slot in additively next to `scalar`.

- **`bindings.rs` is additive / already covers scalar.** The
  `cuvsScalarQuantizer*` symbols are already present in
  `rust/cuvs-sys/src/bindings.rs` (they are reachable from `core/all.h`, which
  already `#include`s `cuvs/preprocessing/quantize/scalar.h`), so no change to
  `all.h` or the bindgen wrapper was required and no symbols were removed. The
  Rust wrapper in this PR consumes the existing generated FFI declarations.
  Regenerating the bindings against an unrelated/newer local `dlpack.h` would
  introduce unrelated DLPack ABI doc/struct churn, so the checked-in bindings
  were deliberately left untouched.

- **Sibling-PR conflict note.** This PR only adds new files plus three small
  additive lines (`pub mod preprocessing;` in `lib.rs` and the `i8` dtype impl
  in `dlpack.rs`). Other in-flight binding PRs that add new `pub mod` lines to
  `lib.rs` or new `IntoDtype` impls to `dlpack.rs` may touch the same hunks; if
  conflicts arise they are trivial textual conflicts (independent additions),
  resolvable by keeping both sides.

## Testing

Built and tested against conda `libcuvs` with `CUDA_VISIBLE_DEVICES=1` and
`--test-threads=1`. New tests in `preprocessing/quantize/scalar.rs`:

- `test_scalar_quantizer_params` — verifies the `set_quantile` builder updates
  the underlying C struct.
- `test_scalar_quantizer_roundtrip` — trains on a random `f32` dataset
  (1024×16, values in `[0, 10)`), quantizes to `i8`, then
  inverse-transforms back to `f32` and asserts the reconstruction error is
  within quantization tolerance. The quantized values are also asserted to span
  most of the `i8` range (confirming the transform is non-trivial). Observed max
  absolute reconstruction error: **~0.0196** on a data range of **10.0**
  (≈0.2% of range, ≈half a quantization step of `10/256 ≈ 0.039`), well below
  the loose epsilon (`data_range / 50 = 0.2`) and far below `5% × data_range`.
- `test_train_unsupported_dtype_errors` — exercises the C API error path by
  training on an integer dataset, which the C API rejects. (A freshly created,
  untrained quantizer has `min_ == max_ == 0`, which produces degenerate output
  but is *not* surfaced as an error by the C API, so the dtype guard is used to
  cover the error path instead.)

The `preprocessing/quantize/mod.rs` doctest also runs end-to-end (train →
transform → inverse_transform).

`cargo fmt` is clean and `cargo clippy` is clean for all new code. (Two
pre-existing `clippy::not_unsafe_ptr_arg_deref` findings in `resources.rs` are
surfaced only by newer local toolchains and are unrelated to this change.)

### DIFF
--- a/rust/cuvs/src/lib.rs
+++ b/rust/cuvs/src/lib.rs
@@ -18,6 +18,7 @@ pub mod dlpack;
 mod error;
 pub mod ivf_flat;
 pub mod ivf_pq;
+pub mod preprocessing;
 mod resources;
 #[cfg(test)]
 pub(crate) mod test_utils;
--- a/rust/cuvs/src/preprocessing/mod.rs
+++ b/rust/cuvs/src/preprocessing/mod.rs
@@ -1,0 +1,11 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+//! Preprocessing utilities for cuVS datasets.
+//!
+//! Currently this exposes the [`quantize`] module, which provides quantizers
+//! that compress floating-point datasets into more compact representations.
+
+pub mod quantize;
--- a/rust/cuvs/src/preprocessing/quantize/mod.rs
+++ b/rust/cuvs/src/preprocessing/quantize/mod.rs
@@ -1,0 +1,25 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+//! Dataset quantizers.
+//!
+//! Quantizers compress a floating-point dataset into a more compact
+//! representation. The [`scalar`] quantizer maps an interval of the input
+//! float range onto the full range of an 8-bit integer.
+//!
+//! Train a [`scalar::Quantizer`] on a dataset, then
+//! [`transform`](scalar::Quantizer::transform) a float matrix into int8 and
+//! [`inverse_transform`](scalar::Quantizer::inverse_transform) it back to an
+//! approximation of the original. Tensors are passed through the
+//! [`AsDlTensor`](crate::AsDlTensor) /
+//! [`AsDlTensorMut`](crate::AsDlTensorMut) traits; see the
+//! [`dlpack`](crate::dlpack) module for the tensor model and `examples/cagra.rs`
+//! for a device-tensor adapter.
+//!
+//! The binary and product (PQ) quantizers exposed by the cuVS C API are not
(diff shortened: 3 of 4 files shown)


jamie8johnson:
On the quantile-range nitpick (`set_quantile`, range `(0, 1]`): leaving it unvalidated deliberately — the C API doesn't enforce the range either, and the crate's builder setters (cagra, ivf_*) uniformly pass values through without validation, so a lone guard here would diverge from the established pattern. The docstring states the valid range. Happy to add a check (or a crate-wide setter-validation sweep as a follow-up) if maintainers prefer.

PR #2235, opened 2026-06-10
feat(rust): add tiered index bindings

jamie8johnson:
# Add Rust bindings for the tiered index

## What

Adds a safe Rust wrapper for the tiered index C API
(`c/include/cuvs/neighbors/tiered_index.h`) under
`rust/cuvs/src/tiered_index/`.

The tiered index couples a brute-force tier that absorbs incremental inserts
with an ANN tier (CAGRA by default; IVF-Flat and IVF-PQ also selectable). The
distinguishing property is that vectors added via `extend` are immediately
searchable, even before the ANN tier is (re)built.

New public API (`cuvs::tiered_index`):

- `IndexParams` — wraps `cuvsTieredIndexParams`. Setters for `metric`, `algo`
  (`AnnAlgo` = `cuvsTieredIndexANNAlgo`: CAGRA / IVF-Flat / IVF-PQ),
  `min_ann_rows`, and `create_ann_index_on_extend`. The embedded per-algorithm
  ANN params (`cagra_params`, `ivf_flat_params`, `ivf_pq_params`) are supplied
  via `set_cagra_params` / `set_ivf_flat_params` / `set_ivf_pq_params`, which
  move ownership of the sub-params wrapper into `IndexParams` so the underlying
  C struct outlives the raw pointer stored in the tiered params struct (same
  pattern as `cagra::IndexParams::set_compression`).
- `Index` — wraps `cuvsTieredIndex`. `build`, `extend`, `search`, and
  `search_with_filter` (the C `cuvsTieredIndexSearch` always takes a
  `cuvsFilter` prefilter; `search` passes a `NO_FILTER` sentinel, mirroring the
  existing IVF-Flat wrapper).

All handles use balanced `Create`/`Destroy` with `Drop`, and
`MaybeUninit::assume_init` is only reached after `check_cuvs` succeeds,
consistent with the other neighbor modules.

The `cuvsTieredIndex*` symbols are already produced by the checked-in
`rust/cuvs-sys/src/bindings.rs` (the bindgen wrapper includes
`cuvs/core/all.h`, which transitively includes `tiered_index.h`, and the
`cuvs.*` allowlist captures the symbols), so no bindings change is required.
`rust/scripts/generate-bindings.sh` reproduces the same output.

## Reviewer notes

- **Search params are opaque (`void*`).** `cuvsTieredIndexSearch` takes the ANN
  backend's search params as `void*`. This wrapper passes
  `cagra::SearchParams` (the default backend). If you'd prefer an enum that
  dispatches over the three backends' search params, I'm happy to extend it —
  the wrapping point is `Index::search_with_filter`.
- **No serialize/deserialize.** The tiered C API exposes no serialize/compact
  entry points (confirmed via `nm -D libcuvs_c.so` — only
  `Create/Destroy/ParamsCreate/ParamsDestroy/Build/Search/Extend/Merge`
  exist), so this PR intentionally omits persistence and takes no filesystem
  paths.
- **`Merge` not wrapped yet.** The C API offers `cuvsTieredIndexMerge`; it is
  left out of this first pass to keep the surface focused on
  build/extend/search. Easy to add in a follow-up (or this PR) if desired.
- **Local-testing caveat (important).** The Rust workspace is versioned
  `26.8.0`, but the newest libcuvs available in this contributor's environment
  is the conda **26.06** build. The `cuvs-sys` cmake gate requires a libcuvs of
  the same major/minor and not older than the crate version, so against 26.06
  the crate will not configure as-is. The tiered C **symbols are present in
  26.06** (verified with `nm -D`), so to exercise the tests I temporarily
  pinned the workspace version to `26.6.0`, ran the suite green on an
  `RTX 4000` (GPU 1), then reverted the version. CI on a matching 26.08 build
  should run the tests unmodified. Tested locally against libcuvs 26.06 with the workspace version temporarily pinned; CI against the current dev version should run unmodified.

## Testing

Added unit tests in `tiered_index/index_params.rs` and
`tiered_index/index.rs`, plus a runnable doc example in `mod.rs`:

- `test_index_params` / `test_embedded_cagra_params_retained` — setters mutate
  the C struct; embedded CAGRA params pointer stays live (not dangling).
- `test_tiered_build_and_search` — build on an initial dataset; each query
  finds itself.
- `test_tiered_extend_visibility` — **the key test**: vectors added via
  `extend` after build are immediately findable (each new vector is its own
  rank-0 neighbor with ~0 self-distance).
- `test_tiered_repeated_extends` — three successive `extend` rounds; each
  batch is immediately findable (asserted as top-k membership + ~0
  self-distance, robust to the ANN tier's approximate recall after an
  on-extend rebuild).
- `test_tiered_filtered_search` — a bitset prefilter excluding id 0 keeps id 0
  out of the result set.

Result (single-threaded, `CUDA_VISIBLE_DEVICES=1`, conda libcuvs 26.06 with
the workspace temporarily pinned to 26.6.0 for linking):

```
running 6 tests
test tiered_index::index::tests::test_tiered_build_and_search ... ok
test tiered_index::index::tests::test_tiered_extend_visibility ... ok
test tiered_index::index::tests::test_tiered_filtered_search ... ok
test tiered_index::index::tests::test_tiered_repeated_extends ... ok
test tiered_index::index_params::tests::test_embedded_cagra_params_retained ... ok
test tiered_index::index_params::tests::test_index_params ... ok
test result: ok. 6 passed; 0 failed
```

Doc test for `mod.rs` also passes. `cargo fmt --check` is clean and
`cargo clippy` reports no findings in `tiered_index/` (the pre-existing
`clippy::not_unsafe_ptr_arg_deref` note on `resources.rs` is unrelated to this
change and surfaces only under a newer local clippy).

## Sibling-PR conflict note

This PR adds `pub mod tiered_index;` to `rust/cuvs/src/lib.rs`. Other queued
Rust-binding PRs (brute-force serialize, scalar quantization, refine) also add
a `pub mod` line to the same `lib.rs` module list. These are adjacent one-line
additions and will conflict trivially on the `mod` list; resolve by keeping all
the new `pub mod` lines. No other files overlap.

### DIFF
--- a/rust/cuvs/src/lib.rs
+++ b/rust/cuvs/src/lib.rs
@@ -21,6 +21,7 @@ pub mod ivf_pq;
 mod resources;
 #[cfg(test)]
 pub(crate) mod test_utils;
+pub mod tiered_index;
 pub mod vamana;
 
 pub use dlpack::{AsDlTensor, AsDlTensorMut, DLPackError, DLTensorView, DLTensorViewMut, DType};
--- a/rust/cuvs/src/tiered_index/index.rs
+++ b/rust/cuvs/src/tiered_index/index.rs
@@ -1,0 +1,538 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+use std::io::{Write, stderr};
+
+use crate::dlpack::{AsDlTensor, AsDlTensorMut, DLTensorView};
+use crate::error::{Error, Result, check_cuvs};
+use crate::resources::Resources;
+use crate::tiered_index::{AnnAlgo, IndexParams, SearchParams};
+
+/// Tiered ANN Index.
+///
+/// A tiered index couples a brute-force tier that absorbs incremental inserts
+/// with an ANN tier (CAGRA by default). Vectors added via [`Index::extend`]
+/// land in the brute-force tier and are immediately searchable, even before the
+/// ANN tier has been (re)built — this is the defining behavior of the tiered
+/// index.
+///
+/// The C API offers no serialize/deserialize for the tiered index, so this
+/// wrapper does not expose persistence (and therefore takes no filesystem
+/// paths).
+#[derive(Debug)]
+pub struct Index {
+    handle: ffi::cuvsTieredIndex_t,
+    /// The ANN backend the index was built with, used to validate that search
+    /// params match the backend (the C search API reinterprets an opaque
+    /// `void*` per this algo).
+    algo: AnnAlgo,
+}
+
+impl Index {
+    /// Builds a new tiered Index over `dataset` for efficient search.
+    ///
+    /// `dataset` is a row-major matrix on the host or device implementing
+    /// [`AsDlTensor`]. It is copied into the index, so the caller may free it
+    /// once this call returns (hence `Index` carries no lifetime).
+    ///
+    /// # Arguments
+    ///
+    /// * `res` - Resources to use
+    /// * `params` - Parameters for building the index (backend + ANN params)
+    /// * `dataset` - The dataset to index
+    pub fn build<T>(res: &Resources, params: &IndexParams, dataset: &T) -> Result<Index>
+    where
+        T: AsDlTensor + ?Sized,
+    {
+        let dataset = dataset.as_dl_tensor()?;
+        let handle = Index::create_handle()?;
+        unsafe {
+            check_cuvs(ffi::cuvsTieredIndexBuild(
+                res.0,
+                params.as_ptr(),
+                dataset.to_c().as_mut_ptr(),
+                handle,
+            ))?;
+        }
+        // Capture the backend so search() can reject mismatched params.
+        Ok(Index { handle, algo: params.algo() })
+    }
+
+    /// Creates a new empty index handle.
+    ///
+    /// Private: [`Index::build`] is the only public constructor so that every
+    /// `Index` carries the backend algo captured from its build params.
+    fn create_handle() -> Result<ffi::cuvsTieredIndex_t> {
+        unsafe {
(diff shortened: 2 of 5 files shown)


jamie8johnson:
Both findings addressed in e2811ed4 (follow-up commit, no history rewrite):

- **Backend-aware search params** (Major): `tiered_index::SearchParams` is now an enum over `Cagra`/`IvfFlat`/`IvfPq` backend params. The index captures its `algo` at build time and `search`/`search_with_filter` validate the variant against it, returning `InvalidArgument` on mismatch instead of letting the C API reinterpret the wrong struct. New `test_search_params_backend_mismatch` covers it.
- **Raw `cuvsFilter` in a safe fn** (Critical): `search_with_filter` now takes the bitset as a `&ManagedTensor` and builds the filter internally, matching the cagra wrapper's shape — no arbitrary addresses through the safe API.

7/7 tests + doctest pass on GPU against libcuvs 26.06.

jamie8johnson:
Good catch — added test_tiered_ivf_flat_backend_search: builds an IVF-Flat-backed tiered index and verifies self-neighbor search end-to-end, so the algo/params mapping for a non-CAGRA backend is now positively exercised, not just the mismatch rejection. 8/8 tests green on GPU.

jamie8johnson:
Downstream data point for prioritization: we are now consuming this branch in production (pinned via a fork patch in [cqs](https://github.com/jamie8johnson/cqs), a code-search engine with a file-watcher daemon). The tiered index is what lets us eliminate periodic full CAGRA/HNSW rebuilds — incremental adds land in the brute-force tier and compact into CAGRA, so the watcher's rebuild thread (and its crash-recovery machinery) goes away entirely.

Practical notes from integration so far: the 26.06 conda `libcuvs` already exports the tiered C API this wraps, so the bindings work against the current release; and the backend-aware `SearchParams` + safe bitset filter shape from the review rounds is exactly what a real consumer needs (we pass filtered searches from live queries). Happy to report back with production numbers or adapt to any API changes review requires — we are watching this PR and will track rebases.

PR #2236, opened 2026-06-10
CAGRA Bloom Filter

divyegala:


examples/cpp/plot_filter_benchmark.py
```diff
@@ -0,0 +1,457 @@
+#!/usr/bin/env python3
```
cjnolet: Can we move this into a "scripts/" directory? It can be under `examples/cpp/scripts`. but just clean it up so it's clear where the c++ and python source code live.

divyegala: Oh sure, it was just a portable script for me to be able to benchmark bloom filter on multiple systems. I don't think it'll make it to the final PR

cpp/src/core/bloom_filter.cu
```diff
@@ -0,0 +1,193 @@
...
+  {
+    filter = cuco_filter_type(num_blocks, {}, {}, {}, stream);
+  }
+
+  void configure_or_validate_dataset_rows(raft::device_vector_view<const key_type, int64_t> keys,
+                                          cudaStream_t stream)
+  {
+    if (keys.extent(0) == 0) { return; }
+
+    auto inferred_dataset_rows = static_cast<std::size_t>(keys.extent(0));
```
dantegd: Should we separate index cardinality from insertion-batch size here? For example, if a 1,000-row index has 500 valid IDs and the caller uses the documented filtering_rate = 0.5, this treats the dataset size as 500 and allocates for only 250 insertions, even though all 500 keys are added. That invalidates the requested FPR and can reduce CAGRA recall, right? What do you think about accepting the index row count or expected insertion count explicitly?

cpp/src/core/bloom_filter.cu
```diff
@@ -0,0 +1,193 @@
...
+      configured_dataset_rows  = inferred_dataset_rows;
+      auto required_num_blocks = compute_num_blocks_from_rates(
+        inferred_dataset_rows, filtering_rate, target_false_positive_rate);
+      auto target_num_blocks = std::max<std::size_t>(filter.block_extent(), required_num_blocks);
+      if (target_num_blocks != filter.block_extent()) { rebuild_filter(target_num_blocks, stream); }
+      return;
+    }
+
+    RAFT_EXPECTS(
+      inferred_dataset_rows == *configured_dataset_rows,
```
dantegd: Is requiring every add() batch to have exactly the same length intentional? This prevents normal chunked insertion and also prevents clearing and repopulating with a different cardinality. Could sizing capacity be fixed independently while allowing arbitrary add-batch sizes?

cpp/src/neighbors/cagra.cuh
```diff
@@ -385,6 +385,28 @@ void search(raft::resources const& res,
   } catch (const std::bad_cast&) {
   }
 
+  try {
+    auto& sample_filter =
+      dynamic_cast<const cuvs::neighbors::filtering::bloom_filter&>(sample_filter_ref);
+    search_params params_copy = params;
+    if (params.filtering_rate <= 0.0f) {
```
dantegd: Could this use < 0.0f, matching the bitset and UDF paths? search_params documents negative values as automatic estimation, while 0.0 is a valid explicit setting. As written, Bloom unexpectedly overrides that setting and may change itopk_size.

cpp/src/core/bloom_filter.cu
```diff
@@ -0,0 +1,193 @@
...
+                                            std::size_t dataset_rows) const
+{
+  if (dataset_rows == 0) { return 0.0f; }
+  auto stream = raft::resource::get_cuda_stream(res);
+  auto policy = raft::resource::get_thrust_policy(res);
+
+  rmm::device_uvector<std::uint8_t> hits(dataset_rows, stream);
+
+  auto first_id = thrust::counting_iterator<key_type>(0);
+  impl_->filter.contains_async(first_id, first_id + dataset_rows, hits.data(), stream);
```
dantegd: The default CAGRA path calls this estimate on every search, so each search allocates an N-byte buffer, probes every index row, and synchronizes before ANN search begins, that could be costly, no? Could we cache an estimate invalidated by add/clear, retain insertion metadata, or use bounded sampling instead?

cpp/tests/neighbors/ann_cagra.cuh
```diff
@@ -946,6 +948,77 @@ class AnnCagraFilterTest : public ::testing::TestWithParam<AnnCagraInputs> {
...
+                        0.003);
+          RAFT_LOG_INFO("Bloom filter recall = %f (%zu/%zu), target_false_positive_rate = %f",
+                        bloom_recall,
+                        bloom_match_count,
+                        bloom_total_count,
+                        target_false_positive_rate);
+          RAFT_LOG_INFO("Bloom filter index recall = %f", bloom_index_recall);
+          bloom_recalls.push_back(bloom_recall);
+        }
+        if (bloom_recalls.size() > 1) {
```
dantegd: Should we make every Bloom run assert that each returned ID is accepted by global_bloom_filter.contains, and enforce an absolute recall floor against the exact filtered ground truth? 

Currently the normal one-FPR path makes no assertions, and the 4096-block floor means the FPR sweep does not exercise different filter sizes.

cpp/include/cuvs/neighbors/common.hpp
```diff
@@ -497,7 +500,7 @@ namespace filtering {
  * @{
  */
 
-enum class FilterType { None, Bitmap, Bitset, UDF };
+enum class FilterType { None, Bitmap, Bitset, Bloom, UDF };
```
dantegd: Would it make sense to append Bloom after UDF, or assign explicit values, so the existing UDF ordinal does not change? The C++ API is not the stable ABI boundary, so this is not necessarily blocking, but preserving the value is inexpensive.

cpp/src/core/bloom_filter.cu
```diff
@@ -0,0 +1,193 @@
...
+  auto stream = raft::resource::get_cuda_stream(res);
+  impl_->configure_or_validate_dataset_rows(keys, stream);
+  impl_->filter.add_async(keys.data_handle(), keys.data_handle() + keys.extent(0), stream);
+}
+
+void bloom_filter::contains(raft::resources const& res,
+                            raft::device_vector_view<const key_type, int64_t> keys,
+                            raft::device_vector_view<std::uint8_t, int64_t> output) const
+{
+  impl_->filter.contains(keys.data_handle(),
```
dantegd, line 188: Would it be worth validating that output.extent(0) >= keys.extent(0) in both contains variants? If I'm not mistaken, cuco writes one result per key, so a mismatched view otherwise permits an out-of-bounds device write.

examples/cpp/src/cagra_bloom_filter_example.cu
```diff
@@ -0,0 +1,160 @@
...
+    if (source_id >= static_cast<key_type>(n_rows)) {
+      std::cerr << "bloom_filter produced out-of-range source_id=" << source_id << std::endl;
+      return 1;
+    }
+    if (bloom_hits_host[i] == 0) {
+      std::cerr << "bloom_filter rejected source_id=" << source_id
+                << " but global bloom filter bulk contains says absent" << std::endl;
+      return 1;
+    }
+    if (!is_valid_row(source_id)) {
```
dantegd: Bloom false positives are expected, no? So this check can fail even when the implementation is behaving correctly. Could the example verify only that every returned ID is accepted by allowed_rows.contains, or apply an exact post-filter before asserting the even ID predicate?

cpp/include/cuvs/core/bloom_filter.hpp
```diff
@@ -0,0 +1,90 @@
...
+                raft::device_vector_view<const key_type, int64_t> keys,
+                raft::device_vector_view<std::uint8_t, int64_t> output) const;
+  void contains_async(raft::resources const& res,
+                      raft::device_vector_view<const key_type, int64_t> keys,
+                      raft::device_vector_view<std::uint8_t, int64_t> output) const;
+
+  [[nodiscard]] std::size_t num_blocks() const noexcept;
+  [[nodiscard]] float estimate_filtering_rate(raft::resources const& res,
+                                              std::size_t dataset_rows) const;
+  [[nodiscard]] impl const& get_impl() const noexcept;
```
dantegd: Could the internal factory be made a friend instead? Exposing impl through the public API weakens the intended opacity and couples the core wrapper to neighbors internals.

cpp/include/cuvs/neighbors/common.hpp
```diff
@@ -617,6 +620,31 @@ struct bitset_filter : public base_filter {
...
+ * Bloom filters have no false negatives: if a row was inserted, @c contains returns @c true. False
+ * positives are possible, so highly selective predicates may still need a bitset or UDF for exact
+ * filtering.
+ */
+struct bloom_filter : public base_filter {
+  void* filter_data{nullptr};
+
+  bloom_filter() = default;
+
+  explicit bloom_filter(const cuvs::core::bloom_filter& bloom_filter)
```
dantegd, line 643: The referenced core filter must outlive the adapter and all searches, and must not be moved or mutated concurrently, right? The stored pointer otherwise becomes invalid or refers to a moved-from object without a diagnostic, so we should document this if I'm right.

dantegd (changes requested):
One thing I noticed: cagra::merge() accepts a base_filter, but its implementation only handles Bitset. A Bloom filter would fall through to the unfiltered branch and keep every row. Should we reject Bloom there for now, like Bitmap, unless probabilistic merge filtering is meant to be supported here?

cpp/include/cuvs/core/bloom_filter.hpp
divyegala: I went down the friend class route, and did not like the idea of forward declaring an internal type in a public header to then mark it as `friend`. Good idea to just declare a friend function instead, I did not think of that.

cpp/src/core/bloom_filter.cu
divyegala: Hmm, good catch. I think accepting the index row count is a far cleaner design and tightens the contract by not having to infer the size of the bloom filter.

cpp/src/core/bloom_filter.cu
divyegala: Fixed as a consequence of above comment

cpp/src/core/bloom_filter.cu
divyegala: Also fixed by bounding to size ahead of time in the constructor. Very, very good catch.

cpp/include/cuvs/neighbors/common.hpp
divyegala: Ah, let me bound this by giving them actual values also then to tighten this for the future.

cpp/tests/neighbors/ann_cagra.cuh
divyegala: This brings me to another question: should we just remove `num_blocks` from the constructor? I think so. We can always just infer it for the user and make the API less confusing to use.

cpp/include/cuvs/neighbors/common.hpp
divyegala: Yes, I'll add the note but it is a problem with all our filters as we separate the implementation from the interface.

divyegala:
> Should we reject Bloom there for now, like Bitmap, unless probabilistic merge filtering is meant to be supported here?

@dantegd good question! Perhaps it is for yourself and Ben (I can't find his @) to answer? I'll explicitly reject it for now.

cpp/include/cuvs/neighbors/common.hpp
```diff
@@ -497,7 +500,7 @@ namespace filtering {
  * @{
  */
 
-enum class FilterType { None, Bitmap, Bitset, UDF };
+enum class FilterType : int { None = 0, Bitmap = 1, Bitset = 2, Bloom = 3, UDF = 100 };
```
dantegd, line 503: Quick question, why assign the new value to udf instead of to bloom?

cpp/src/core/bloom_filter.cu
```diff
@@ -0,0 +1,184 @@
...
+               "target_false_positive_rate must be in (0, 1).");
+
+  // Bloom sizing: m = -n * ln(p) / (ln(2)^2), then blocks = ceil(m / 256 bits-per-block).
+  constexpr double kBitsPerBlock = 256.0;
+  constexpr double kLn2          = 0.6931471805599453;
+  constexpr double kLn2Sq        = kLn2 * kLn2;
+
+  auto expected_insertions = std::max<std::size_t>(
+    1, static_cast<std::size_t>(std::ceil(static_cast<double>(dataset_rows) * filtering_rate)));
+  auto required_bits = -static_cast<double>(expected_insertions) *
```
dantegd: Could we account for the cuco policy’s fixed number of pattern bits when sizing this? This formula assumes the optimal hash count k = (m/n) ln(2), but the default cuco policy used below has a fixed pattern-bit count. 

What do you think about selecting/configuring the policy from the requested FPR, or using sizing math for the actual policy and adding an observed-FPR test?

cpp/src/neighbors/cagra.cuh
```diff
@@ -385,6 +385,28 @@ void search(raft::resources const& res,
...
+      dynamic_cast<const cuvs::neighbors::filtering::bloom_filter&>(sample_filter_ref);
+    search_params params_copy = params;
+    if (params.filtering_rate < 0.0f) {
+      const float min_filtering_rate = 0.0f;
+      const float max_filtering_rate = 0.999f;
+      auto const* bloom_filter_obj =
+        static_cast<const cuvs::core::bloom_filter*>(sample_filter.filter_data);
+      RAFT_EXPECTS(bloom_filter_obj != nullptr,
+                   "bloom_filter must carry a valid cuvs::core::bloom_filter handle.");
+      params_copy.filtering_rate = bloom_filter_obj->estimate_filtering_rate(
```
dantegd: Could we avoid running the full-domain estimator on every default search? search_params::filtering_rate defaults to negative, and this call allocates one byte per index row, probes every row ID, and synchronizes the stream.  Could the filter cache/update an estimate across add/clear, or derive one from its configured insertion rate and policy? What do you think?

dantegd (changes requested):
Two quick comments, the last one could be a quick optimization follow up.

cpp/include/cuvs/neighbors/common.hpp
divyegala: It follows the metric enum convention to separate pre-compiled types vs UDF unknown type as a sentinel value. New types can fill up between 3 and 100 going forward.

cpp/src/neighbors/cagra.cuh
divyegala: Hmm, yes. I thought it was a one-time run but I forgot that CAGRA will keep calling the estimate API due to search_params::filtering_rate. It should be easy to derive a scalar.

cpp/src/core/bloom_filter.cu
divyegala: That's a really good change request!

dantegd (approved):

PR #2240, opened 2026-06-12
[FEA] Roaring-bitmap prefilters for brute-force search (shared and per-query)

### LINKED ISSUE #1972: [FEA] Roaring Bitmap support
Filtered vector search optimization

**Describe the solution you'd like**
Adding Roaring Bitmap Support directly to cuVS.

**Describe alternatives you've considered**
Using the existing bitset implementation

**Additional context**
Original Roaring Bitmap paper: https://arxiv.org/abs/1402.6407
Follow on paper that introduced run encoding: https://arxiv.org/abs/1603.06549

  Summary

  I'd like to propose contributing https://roaringbitmap.org/ support on Exa's behalf as an alternative filter representation for cuVS's prefiltered search APIs (for CAGRA, IVF-Flat, brute-force, etc).

  Motivation

  cuVS currently supports bitset_filter for prefiltered search, backed by a flat bit array. This works, but has two
  limitations:

  1. Memory scaling. A flat bitset for N vectors requires N/8 bytes regardless of how many bits are set. At 10M vectors
  this is 1.2 MB; at 100M it's 12 MB; at 1B it's 120 MB. For sparse filters (e.g., 1% pass rate at 100M scale), a
  compressed representation could reduce this by 10-100x or even more.
  2. Cache pressure. During CAGRA graph traversal and IVF-Flat list scanning, filter checks access random positions
  in the bitset. When the bitset exceeds L1 cache (128 KB per SM on Blackwell), every check incurs an L2 round-trip
  (~200 cycles vs ~28 cycles for L1). Roaring's two-level structure (small key index + 8 KB containers) keeps the
  working set in L1.
  3. Interoperability. Roaring Bitmaps are the standard compressed bitmap format across the database and search
  ecosystem (Lucene, Elasticsearch, Apache Spark, ClickHouse, Redis, etc.). Accepting roaring filters directly avoids a decompress-to-bitset step on the ingestion path and allows for the copying of a tiny roaring bitmap filter to the GPU, potentially one per query vector.

I have already prototype some of the parts of this and it performs extremely well in specific cases, sparse data, clustered or multitenant indexes. It can also work very well in cases where we can presort our data so common filters are contigous so Roaring Bitmap containers store them using run encoding

I spoke with @cjnolet, Nathan Stephens, and Manas Singh earlier today. Filing this PR to kick off the discussion and will share more details next week

maxwbuckley:
# [FEA] Roaring-bitmap prefilters for brute-force search (shared and per-query)

Closes rapidsai/cuvs#1972.

## Problem

cuVS encodes search prefilters as flat bit arrays: `bitset_filter`
(`n_rows` bits shared by all queries) and `bitmap_filter`
(`n_queries x n_rows` bits, one row per query). Below ~40-50%
selectivity — the dominant production regime (tag predicates, ACLs,
recency cutoffs) — this leaves performance and memory on the table:

1. The filter costs `n_rows/8` bytes regardless of how few bits are set
   (`bitmap_filter`: times `n_queries`; 125 GB at `n_queries=1000`,
   `n_rows=1e9`).
2. The sparse pipeline pays a `count()` reduction + host sync per search
   to pick a branch, then converts bitmap→CSR **twice** (once in cuVS,
   once again inside `raft::sparse::linalg::masked_matmul`).
3. There is no path that computes only the selected rows densely — the
   mid-selectivity regime (1-30%) runs either a full-dataset GEMM + mask
   or a structure-blind SDDMM, both far from optimal.

## What this PR adds

A compressed Roaring-bitmap filter representation plus a
selectivity-driven three-regime dispatch for brute-force search.

### API (mirrors the existing filter API)

```cpp
#include <cuvs/core/roaring.hpp>
#include <cuvs/neighbors/roaring_filter.hpp>

// build once per filter; cardinality + container metadata stay host-side
auto bm = cuvs::core::from_sorted_ids(res, ids, n_ids, n_rows);

// one filter shared by all queries (counterpart of bitset_filter)
auto f = cuvs::neighbors::filtering::roaring_filter(bm);
cuvs::neighbors::brute_force::search(res, params, index, queries,
                                     neighbors, distances, f);

// one filter per query (counterpart of bitmap_filter)
std::vector<const cuvs::core::gpu_roaring*> per_query = {...};
auto mf = cuvs::neighbors::filtering::roaring_matrix_filter(
  per_query.data(), n_queries);
cuvs::neighbors::brute_force::search(res, params, index, queries,
                                     neighbors, distances, mf);
```

`cuvs::core::gpu_roaring` is RAII (rmm-backed), movable, and also
provides `set_and/set_or/multi_and/multi_or` for composing predicate
bitmaps on-GPU and `to_bitset()` for interop with the existing filters.

### Dispatch (no count kernels — cardinalities are known on the host)

| selectivity s | path | why |
|---|---|---|
| s ≤ t_sparse(dim) | container ids → CSR (indptr free from host cardinalities, one emission kernel) → `sddmm` → sparse `select_k` | nnz-proportional work |
| t_sparse < s < 0.45 (shared filter) | ids → chunked row **gather → dense GEMM → select_k** → id remap | computes `|filter|` columns instead of `n_rows`; GEMM-class efficiency |
| s ≥ 0.45 | decompress to bitset / bit-matrix, delegate to the existing dense masked pipeline | already optimal when most rows pass |

`t_sparse` is dimension-dependent because cusparse SDDMM degrades with
`dim` while dense GEMM does not: measured crossovers are ~3% at d=128
and ~0.1% at d=512 (RTX 5090); encoded as `dim >= 256 ? 0.001 : 0.03`.

### Measured (RTX 5090, fp32 IP, k=10, recall 1.0 in every cell)

Shared filter, 10M x 512d, 64 queries, vs `bitset_filter` today:

| s | uniform | clustered |
|---|---|---|
| 0.01% | 4.5x | 5.2x |
| 0.1% | 11.0x | 11.3x |
| 1% | 18.7x | 18.7x |
| 10% | 18.2x | 18.4x |
| 30% | 1.7x | 1.6x |
| ≥50% | 1.0x (delegates) | 1.0x |

Per-query filters, 10M x 128d, 256 queries, vs `bitmap_filter` today:
**2.8-6.6x** at s ≤ 1% (and 1.4-3.1x across 1-6.25%), with the dense
`[n_queries, n_rows]` bitmap never materialized below the dense regime.

Filter memory at 10M rows: bitset fixed 1.25 MB; roaring 3 KB at 0.01%
uniform, 2-64 KB clustered (the common production shape), parity only
when genuinely dense.

## Tests

`cpp/tests/neighbors/brute_force_roaring.cu` (added to NEIGHBORS_TEST):
20 parameterized cases cross-validating `roaring_filter` against
`bitset_filter` and `roaring_matrix_filter` against `bitmap_filter`
(tie-tolerant distance comparison + filter-membership assertions) across
InnerProduct / L2Expanded / CosineExpanded, d ∈ {64, 512}, and
selectivities exercising all three dispatch regimes. All pass.

## Limitations / follow-ups

- float32 datasets only (matches the existing filtered-path half
  limitation); half/int8 follow-up.
- `from_sorted_ids` builds containers host-side (construction is
  one-time per filter; a device-side builder is a follow-up).
- Negated (complement) bitmaps are rejected by the filters.
- CRoaring portable-serialization import (ecosystem interop) is a
  natural follow-up — the format is container-compatible.
- RUN containers are supported by the read paths but not yet produced
  by construction.
- IVF-Flat per-candidate roaring filtering (measured +12-22% at 5-10%
  selectivity) is a separate follow-up PR.

## Files

New:
- `cpp/include/cuvs/core/roaring.hpp` — `gpu_roaring`, `roaring_view`
  (device `test()`), construction, set ops, decompression, CSR emission
- `cpp/src/core/roaring/roaring.cu` — implementation
- `cpp/include/cuvs/neighbors/roaring_filter.hpp` — the two filters
- `cpp/src/neighbors/detail/knn_brute_force_roaring.cuh` — dispatch +
  the three pipelines
- `cpp/tests/neighbors/brute_force_roaring.cu`

Modified:
- `cpp/include/cuvs/neighbors/common.hpp` — `FilterType::{Roaring,
  RoaringMatrix}`
- `cpp/src/neighbors/detail/knn_brute_force.cuh` — include + two
  dynamic_cast dispatch cases in `detail::search`; default argument of
  `brute_force_search_filtered` moved to its first declaration
- `cpp/CMakeLists.txt`, `cpp/tests/CMakeLists.txt` — source/test
  registration

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -1142,6 +1142,7 @@ if(NOT BUILD_CPU_ONLY)
     src/cluster/single_linkage_float.cu
     src/cluster/spectral.cu
     src/core/bitset.cu
+    src/core/roaring/roaring.cu
     src/core/omp_wrapper.cpp
     src/util/file_io.cpp
     src/util/host_memory.cpp
--- a/cpp/include/cuvs/core/roaring.hpp
+++ b/cpp/include/cuvs/core/roaring.hpp
@@ -1,0 +1,373 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2025, NVIDIA CORPORATION.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#pragma once
+
+#include <cuvs/core/bitset.hpp>
+
+#include <raft/core/device_mdarray.hpp>
+#include <raft/core/resource/cuda_stream.hpp>
+#include <raft/core/resources.hpp>
+
+#include <rmm/device_uvector.hpp>
+
+#include <cstdint>
+#include <memory>
+#include <vector>
+
+namespace cuvs::core {
+
+/**
+ * @defgroup roaring GPU Roaring Bitmap
+ *
+ * GPU-native Roaring Bitmap for compressed prefiltering in vector search.
+ * Stores attribute bitmaps in compressed Roaring format in GPU memory,
+ * supports bulk set operations (AND/OR/ANDNOT/XOR), and decompresses
+ * to flat bitsets compatible with cuvs::core::bitset for search.
+ *
+ * Typical usage:
+ * @code{.cpp}
+ *   // Build filter bitmaps on CPU, upload to GPU
+ *   auto category = gpu_roaring::from_sorted_ids(res, cat_ids, universe);
+ *   auto price    = gpu_roaring::from_sorted_ids(res, price_ids, universe);
+ *
+ *   // Combine on GPU
+ *   auto combined = gpu_roaring::set_and(res, category, price);
+ *
+ *   // Decompress to bitset for search
+ *   auto bitset = gpu_roaring::to_bitset(res, combined);
+ *   auto filter = cuvs::neighbors::filtering::bitset_filter<uint32_t, int64_t>(bitset.view());
+ *   cuvs::neighbors::cagra::search(res, params, index, queries, neighbors, distances, filter);
+ * @endcode
+ * @{
+ */
+
+/** Container type tags for Roaring containers */
+enum class roaring_container_type : uint8_t { ARRAY = 0, BITMAP = 1, RUN = 2 };
+
+/**
+ * @brief Non-owning device view of a @ref gpu_roaring, usable inside
+ * kernels for per-sample membership tests (mirrors
+ * raft::core::bitset_view::test()).
+ */
+struct roaring_view {
+  const uint16_t* keys                = nullptr;
+  const roaring_container_type* types = nullptr;
+  const uint32_t* offsets             = nullptr;
+  const uint16_t* cardinalities       = nullptr;
+  const uint64_t* bitmap_data         = nullptr;
+  const uint16_t* array_data          = nullptr;
+  const uint16_t* run_data            = nullptr;
+  const uint16_t* key_index           = nullptr;  // optional O(1) key lookup
+  uint32_t n_containers               = 0;
+  uint32_t max_key                    = 0;
+  uint32_t n_rows                     = 0;
+
+#if defined(__CUDACC__)
+  /** Whether sample index `id` is a member of the filter. */
+  __device__ __forceinline__ bool test(uint32_t id) const
+  {
+    uint16_t key = static_cast<uint16_t>(id >> 16);
+    uint16_t low = static_cast<uint16_t>(id & 0xFFFFu);
+
+    int cid = -1;
+    if (key_index != nullptr) {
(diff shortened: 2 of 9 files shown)


maxwbuckley:
Note, part of why the speedups look quite so sweet is noted previously https://github.com/issues/created?issue=NVIDIA%7Ccuvs%7C1972 the 10% threshold is badly calibrated

PR #2245, opened 2026-06-15
Track tensor and index lifetimes in the Rust DLPack bindings

yan-zaretskiy:
The previous `ManagedTensor` type was a non-owning wrapper over the C's FFI `DLManagedTensor` type with no lifetime attached. Hence there was no mechanism to tie the tensor data and shape/stride metadata it referenced to their owners. Indexes that keep a non-owning view of their dataset (CAGRA, brute-force) could outlive that data. Here we replace it with lifetime-parameterized `DLTensorView` and `DLTensorViewMut` views. They are produced by the public `IntoDlTensor` and `IntoDlTensorMut` traits. Users are now expected to implement these traits on their tensor types, so that our API can accept them as input/output arguments.


yan-zaretskiy:
Since this is a fairly large change, I would like to explain my thought process to justify the proposed design. First of all, let's identify what we're trying to fix. 

The original `cuvs` type for tensor data is simply `pub struct ManagedTensor(ffi::DLManagedTensor)`. It is a thin wrapper around the C `DLManagedTensor` - the struct that holds the raw `data`, `shape`, and `strides` pointers. When it's built from a borrowed tensor (e.g. via `From<&ndarray>`) it holds those pointers with no lifetime attached, so nothing ties the `ManagedTensor` to whatever owns the data, which can be freed out from under it. Therefore, using this type is effectively unsafe, though it is not marked as such.

The second problem is in the index types. CAGRA and brute-force don't copy the dataset at build time, because the underlying C++ implementation keeps a non-owning view and reads the original vectors during search, so the dataset has to outlive the index. The old API doesn't model that.

The purpose of this PR is therefore to enable tracking two things: the lifetime of the tensor data, and the lifetime of the indexes that borrow it.

To fix the first issue, we add lifetimes to the tensor view types. Moreover, we make a distinction between mutable and immutable views: `DLTensorView<'a>` for read-only inputs like datasets and queries, and `DLTensorViewMut<'a>` for outputs the C API writes, like neighbors and distances.

Users convert their own tensor types into these views by implementing a pair of new public traits, `IntoDlTensor<'a>` and `IntoDlTensorMut<'a>`. Every entry point (`build`, `search`, etc) takes `impl IntoDlTensor<'a>` or `impl IntoDlTensorMut<'a>` rather than a concrete type, so `cuvs` ships no tensor type of its own and stays agnostic about where your data lives.

The path from `DLTensorView<'a>` into C requires some clarification. Here's how we define this type:

```rust
pub struct DLTensorView<'a> {
    data: *mut std::ffi::c_void,
    device: ffi::DLDevice,
    dtype: ffi::DLDataType,
    shape: TensorDims,
    strides: Option<TensorDims>,
    _marker: PhantomData<&'a ()>,
}
```

where `TensorDims` is a vector with SBO, and in practice is always stack allocated. This is done to avoid heap allocations when making FFI calls from Rust. The reason `DLTensorView<'a>` owns `shape` and `strides` is because the C type eventually reads them as `*mut i64`, so *something* must own the data that this `*mut i64` points to. One problem is that this something cannot be the original tensor type, because there's no guarantee it uses `i64` for shape/stride values. So we choose to store it inside `DLTensorView<'a>`. The second design decision is how we hand a `DLTensorView<'a>` to the C function, which ultimately wants a `*mut DLManagedTensor` whose `shape`/`strides` are `*mut i64` pointing at those arrays.

The obvious choice is to store a `DLManagedTensor` right inside the view, with its `shape`/`strides` pointing at our own `shape`/`strides` fields, and hand out a pointer to it. But that makes `DLTensorView` a self-referential struct, since the embedded `DLManagedTensor` would point back into the same value. And because `TensorDims` keeps its data inline, moving the view moves that inline array, while the `DLManagedTensor`'s `*mut i64` keeps pointing at the old location. Hence this choice is unsound.

There is a known way to make the embedding sound: store `shape`/`strides` as `Box<[i64]>` instead. Then the arrays live on the heap, and moving the view only moves the `Box` handle. But this forces a heap allocation per view.

So rather than embed the C struct, we build it only at a point we're about to call into C:
```rust
pub(crate) struct ManagedTensorRef<'a> {
    inner: ffi::DLManagedTensor,
    _borrow: PhantomData<&'a ()>,
}
impl<'a> DLTensorView<'a> {
    // Fills `inner` with `shape: self.shape.as_ptr() as *mut _`, etc.
    pub(crate) fn to_c(&self) -> ManagedTensorRef<'_> { /* ... */ }
}
```

Here `to_c` takes `&self`, and the returned `ManagedTensorRef<'_>` borrows it, so its lifetime is tied to the view. The compiler now guarantees the view, with the inline shape/strides its pointers point into, cannot move or be dropped while the `ManagedTensorRef` is alive. Finally, we take the raw `*mut DLManagedTensor` out of the `ManagedTensorRef` for the FFI call.

As for the second fix, the index types that borrow their dataset now also get a lifetime. CAGRA and brute-force now return `Index<'d>`, where `'d` is the dataset's lifetime, so the compiler forbids the index from outliving the data it reads. IVF-Flat, IVF-PQ and Vamana copy the dataset at build time, so they carry no lifetime.

One special case here is indexes loaded from disk. An index produced by `deserialize` doesn't borrow any caller-supplied dataset, hence there is nothing for `'d` to tie to. We model that by returning `Index<'static>`. The `'static` says the index borrows nothing, so it isn't constrained by any dataset's lifetime and can live as long as you like.

rust/cuvs/src/dlpack.rs
```diff
@@ -3,172 +3,415 @@
...
+///                 DLDevice { device_type: DLDeviceType::kDLCUDA, device_id: 0 },
+///                 &shape,
+///                 None,
+///                 T::dl_dtype(),
+///             )
+///         }
+///     }
+/// }
+/// ```
+pub trait IntoDlTensor<'a> {
```
tgockel: This shape is a little dangerous due to `into_dl_tensor` taking `self` and not something lifetime-bound. While your comment says to do it correctly, there's nothing that prevents someone from writing this dangerous code:

```rust
impl<'a> IntoDlTensor<'a> for GpuMatrix {
  fn into_dl_tensor(self) -> Result<DlTensorView<'a>, DLPackError> { ... }
}
```

A slightly safer approach would be:

```rust
pub trait AsDlTensor {
  fn as_dl_tensor(&self) -> Result<DLTensorView<'_>, DLPackError>;
}

pub trait AsDlTensorMut {
  fn as_dl_tensor_mut(&mut self) -> Result<DLTensorViewMut<'_>, DLPackError>;
}
```

You'd have to change the [`Index::build`](https://github.com/NVIDIA/cuvs/pull/2245/changes#diff-e7de240d1757052fb1d408c79a2cd3b524c37c85e07c829e6d2ce75d14d56645R39) function to be borrow-based over `dataset`:

```rust
pub fn build<T: AsDlTensor + ?Sized>(..., dataset: &'d T) -> Result<Index<'d>> { ... }
```

I _think_ that is a viable approach in all the cases here.

tgockel: Basically, the trait as-written imposes safety obligations on implementers of the trait. The solution to this is to mark the trait as `unsafe` if you want to keep the `IntoDlTensor` approach. But if you switch the API to take `&self`, then the compiler ties the lifetime for you and you don't need `unsafe`.

cjnolet (approved):

PR #2246, opened 2026-06-16
Predict IVF-PQ FP16 overflow and auto-switch to FP32

huuanhhuyn:
Large-magnitude unnormalized datasets (e.g. SIFT_1M) contains vectors whose norm overflowed the FP16 internal search types of IVF-PQ.

This PR detects it by doing probe ivf-pq search on a small query set and check the result distances whether values are nan or inf.

### DIFF
--- a/cpp/src/neighbors/detail/cagra/cagra_build.cuh
+++ b/cpp/src/neighbors/detail/cagra/cagra_build.cuh
@@ -1670,47 +1670,17 @@ void build_knn_graph(
   // This observes the actual computation, so it is agnostic of the selected distance type.
   if (pq.search_params.internal_distance_dtype == CUDA_R_16F ||
       pq.search_params.coarse_search_dtype == CUDA_R_16F) {
-    // --- Benchmark fp16 overflow detection: probe search with 128 queries ---
     {
-      raft::resource::sync_stream(res);
-      const auto t0  = std::chrono::steady_clock::now();
       const bool fp16_overflow = cuvs::neighbors::ivf_pq::helpers::detect_fp16_overflow(
         res, index, pq.search_params, dataset);
-      const double ms =
-        std::chrono::duration<double, std::milli>(std::chrono::steady_clock::now() - t0)
-          .count();
-      RAFT_LOG_INFO("detect_fp16_overflow (IVF-PQ search) took %.3f ms (overflow=%d)", ms, int(fp16_overflow));
+      if (fp16_overflow) {
+        RAFT_LOG_WARN(
+          "IVF-PQ FP16 distance produced non-finite results on a probe search for this dataset -> "
+          "switching 'internal_distance_dtype' and 'coarse_search_dtype' to FP32");
+        pq.search_params.internal_distance_dtype = CUDA_R_32F;
+        pq.search_params.coarse_search_dtype     = CUDA_R_32F;
+      }
     }
-    // --- Benchmark fp16 overflow detection: estimate max norm on the first 20k rows ---
-    {
-      raft::resource::sync_stream(res);
-      const auto t0 = std::chrono::steady_clock::now();
-      const bool fp16_overflow = cuvs::neighbors::ivf_pq::helpers::estimate_fp16_overflow(
-        res, dataset, pq.build_params.metric);
-      const double ms =
-        std::chrono::duration<double, std::milli>(std::chrono::steady_clock::now() - t0).count();
-      RAFT_LOG_INFO("estimate max norm on the first 20k rows took %.3f ms (overflow=%d)", ms, int(fp16_overflow));
-    }
-    // --- Benchmark fp16 overflow detection: estimate max norm on the IVF-PQ cluster centers ---
-    {
-      raft::resource::sync_stream(res);
-      const auto t0 = std::chrono::steady_clock::now();
-      const bool fp16_overflow = cuvs::neighbors::ivf_pq::helpers::estimate_fp16_overflow_centers(
-        res, index, pq.build_params.metric);
-      const double ms =
-        std::chrono::duration<double, std::milli>(std::chrono::steady_clock::now() - t0).count();
-      RAFT_LOG_INFO("estimate max norm on the IVF-PQ cluster centers took %.3f ms (overflow=%d)", ms, int(fp16_overflow));
-    }
-    // if (fp16_overflow) {
-    //   RAFT_LOG_WARN(
-    //     "IVF-PQ FP16 distance produced non-finite results on a probe search for this dataset -> "
-    //     "switching 'internal_distance_dtype' and 'coarse_search_dtype' to FP32");
-    //   pq.search_params.internal_distance_dtype = CUDA_R_32F;
-    //   pq.search_params.coarse_search_dtype     = CUDA_R_32F;
-    // }
-    // TODO delete this after benchmarking 3 approaches
-    //pq.search_params.internal_distance_dtype = CUDA_R_32F;
-    //pq.search_params.coarse_search_dtype     = CUDA_R_32F;
   }
 
   //
@@ -2267,28 +2237,6 @@ index<T, IdxT> build(
     }
   }
 
-  // Predict potential FP16 distance overflow for large-magnitude (e.g. unnormalized) datasets
-  // -> fall back to FP32.
-  if (auto* pq = std::get_if<cagra::graph_build_params::ivf_pq_params>(&knn_build_params)) {
-    const bool using_fp16_distance = pq->search_params.internal_distance_dtype == CUDA_R_16F ||
-                                     pq->search_params.coarse_search_dtype == CUDA_R_16F;
-    raft::resource::sync_stream(res);
-    const auto detect_start  = std::chrono::steady_clock::now();
-    const bool first_20k_rows_check = ivf_pq::helpers::estimate_fp16_overflow(res, dataset, params.metric);
-    const double detect_ms =
-      std::chrono::duration<double, std::milli>(std::chrono::steady_clock::now() - detect_start)
-        .count();
-    RAFT_LOG_INFO("1st attempt: estimate_fp16_overflow on first 20k rows took %.3f ms (overflow=%d)", detect_ms, int(first_20k_rows_check));
-    if (using_fp16_distance &&
-        ivf_pq::helpers::estimate_fp16_overflow(res, dataset, params.metric)) {
-      RAFT_LOG_WARN(
-        "IVF-PQ internal type of FP16 is insufficient for this dataset to avoid overflow in "
-        "distance computations -> "
-        "Switching 'internal_distance_dtype' and 'coarse_search_dtype' to FP32");
-      //pq->search_params.internal_distance_dtype = CUDA_R_32F;
-      //pq->search_params.coarse_search_dtype     = CUDA_R_32F;
-    }
-  }
   RAFT_EXPECTS(
     params.metric != cuvs::distance::DistanceType::BitwiseHamming ||
       std::holds_alternative<cagra::graph_build_params::iterative_search_params>(
--- a/cpp/src/neighbors/detail/cagra/cagra_build.cuh
+++ b/cpp/src/neighbors/detail/cagra/cagra_build.cuh
@@ -1,6 +1,6 @@
 /*
  * SPDX-FileCopyrightText: Copyright (c) 2023-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
- * SPDX-License-Identifier: Apache-2.0
+ * reserved. SPDX-License-Identifier: Apache-2.0
  */
 #pragma once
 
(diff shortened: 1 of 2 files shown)


cpp/src/neighbors/ivf_pq/ivf_pq_fp16_overflow.cuh
```diff
@@ -0,0 +1,137 @@
...
+  const int64_t dim    = dataset.extent(1);
+
+  // Determine sample size based on fraction, with bounds
+  int64_t n_sample = static_cast<int64_t>(sample_fraction * static_cast<double>(n_rows));
+  if (n_rows <= 1000) {
+    n_sample = n_rows;  // for small datasets, just use them all and skip the sampling overhead
+  } else if (n_rows > 100000) {
+    // cap the sample size to 100k for speed and keep memory use within the limited workspace resource
+    n_sample = 100000;
+  }
```
mfoerste4: Ensure clamping on the actual samples.

cpp/src/neighbors/ivf_pq/ivf_pq_fp16_overflow.cuh
```diff
@@ -0,0 +1,137 @@
...
+    raft::make_device_mdarray<DataT>(res, mr, raft::make_extents<int64_t>(n_sample, dim));
+  raft::matrix::sample_rows<DataT, int64_t>(
+    res, raft::random::RngState{137}, dataset, sample.view());
+
+  auto d_max_sq            = raft::make_device_scalar<float>(res, 0.0f);
+  constexpr int block_size = 256;
+  const int grid_size      = static_cast<int>((n_sample + block_size - 1) / block_size);
+  kern_max_squared_norm<DataT><<<grid_size, block_size, 0, stream>>>(
+    sample.data_handle(), n_sample, dim, d_max_sq.data_handle());
+  RAFT_CUDA_TRY(cudaPeekAtLastError());
```
mfoerste4: Please evaluate whether you can use an implementation from raft, e.g. `rafdt::linalg::norm`.

mfoerste4 (changes requested):
Thanks @huuanhhuyn for the PR. Please utilize raft for the norm computation if possible.

cpp/src/neighbors/detail/cagra/cagra_build.cuh
```diff
@@ -1621,6 +1622,21 @@ void build_knn_graph(
...
+  // Guard against potential FP16 distance overflow for large-magnitude datasets -> back to FP32.
+  const bool using_fp16_distance = pq.search_params.internal_distance_dtype == CUDA_R_16F ||
+                                   pq.search_params.coarse_search_dtype == CUDA_R_16F;
+  if (using_fp16_distance &&
+      ivf_pq::helpers::estimate_fp16_overflow(res, dataset, pq.build_params.metric)) {
+    RAFT_LOG_WARN(
+      "IVF-PQ internal type of FP16 is likely insufficient for this dataset to avoid overflow in "
+      "distance computations -> "
+      "Switching 'internal_distance_dtype' and 'coarse_search_dtype' to FP32");
+    pq.search_params.internal_distance_dtype = CUDA_R_32F;
```
huuanhhuyn: Move to params initialization phase before heuristic estimation to not break the heuristic correctness

cpp/src/neighbors/ivf_pq/ivf_pq_fp16_overflow.cuh
huuanhhuyn: `raft::linalg::norm` won't fit completely because we want to map (`detail::utils::mapping`) each data elements before reducing. Maybe I could use `coalesced_reduction` to reduce to a vector containing all squared norm and then `map_reduce` to find the max squared norm. I will try that

huuanhhuyn: I used raft's `reduce` and `map_reduce` in replacement for the manual kernel.

cpp/src/neighbors/ivf_pq/ivf_pq_fp16_overflow.cuh
huuanhhuyn: I changed the clamping approach to a smooth saturation equation (Please see the PR's attached image). It gives a smooth transition along the dataset size, where we sample most of the dataset for low number of rows and up to 100k samples when the number of rows grows large. The choice of 100k is heuristic.

cpp/src/neighbors/ivf_pq/ivf_pq_fp16_overflow.cuh
```diff
@@ -0,0 +1,137 @@
...
+  // - n_sample increases fast for small n_rows and slow to saturation for large n_rows
+  // Idea: we sample most of the dataset when it is small-sized, and only a small fraction
+  // (up to a maximum/saturation number) when the dataset size grows large.
+  // kSaturation and kDelay are selected as a compromise between runtime and outlier recall.
+  constexpr int64_t kSaturation = 20000;
+  constexpr int64_t kDelay      = kSaturation * 10;
+  RAFT_EXPECTS(kDelay >= kSaturation,
+               "kDelay must not be smaller than kSaturation so that n_sample is always less than "
+               "or equal to n_rows");
+  int64_t n_sample = (n_rows * kSaturation + (n_rows + kDelay - 1)) / (n_rows + kDelay);
```
mfoerste4: Please use `raft::ceildiv` here.

mfoerste4 (approved):
LGTM, thanks @huuanhhuyn for the PR!

cpp/src/neighbors/ivf_pq/ivf_pq_fp16_overflow.cuh
huuanhhuyn: I added it.

cpp/src/neighbors/ivf_pq/ivf_pq_fp16_overflow.cuh
```diff
@@ -0,0 +1,129 @@
...
+{
+  if (dataset.extent(0) == 0) { return false; }
+
+  // Cosine similarity scores does normalization itself, so overflow won't happen
+  if (metric == cuvs::distance::DistanceType::CosineExpanded) { return false; }
+
+  // FP16 largest finite value, with a defensive margin to also avoid precision loss near the limit.
+  constexpr float kFp16Max              = 65504.0f;
+  constexpr float kFp16DefensiveMargin  = 0.25f;
+  const float overflow_detect_threshold = kFp16DefensiveMargin * kFp16Max;
```
tfeher: While the intention is good, I am not sure if we ned so large `kFp16DefensiveMargin`.  E.g. someone could have normalized 4096 dim embeddings, where the current setting would trigger fallback to fp32 (in case of L2 norm). Admittedly, such large embeddings are rare, but I would prefer to only fall back to fp32 when the distance can overflow, and not when we have precision loss due to the limited range of the numeric type. @achirkin what do you think?

tfeher (changes requested):
Thanks Huy for the PR! It looks good, but I have one concern regarding the "defensive margin". Let's see what others think about this questien.

cpp/src/neighbors/ivf_pq/ivf_pq_fp16_overflow.cuh
huuanhhuyn: Thank you for checking it.

I have some arguments for keeping it this way:
- the precision loss would break the ivf-pq search when neighbor distances become indifferentiable. This will fall again into the "low self-included ratio" error message.
- The magnitude looks large but the bitwise distances between 65k and 16k are just 2 bits.

>  someone could have normalized 4096 dim embeddings, where the current setting would trigger fallback to fp32

Do you have the dataset? I could run a test on it

tfeher:
/ok to test [7f41b35](https://github.com/NVIDIA/cuvs/pull/2246/commits/7f41b35cb4ecc0891950a18b70e50819b0595624)

cpp/src/neighbors/ivf_pq/ivf_pq_fp16_overflow.cuh
```diff
@@ -0,0 +1,129 @@
...
+  // - n_sample increases fast for small n_rows and slow to saturation for large n_rows
+  // Idea: we examine most of the dataset when it is small-sized, and only a small fraction
+  // (up to a maximum/saturation number) when the dataset size grows large.
+  // kSaturation and kDelay are selected as a compromise between runtime and outlier recall.
+  constexpr int64_t kSaturation = 20000;
+  constexpr int64_t kDelay      = kSaturation * 10;
+  RAFT_EXPECTS(kDelay >= kSaturation,
+               "kDelay must not be smaller than kSaturation so that n_sample is always less than "
+               "or equal to n_rows");
+  int64_t n_sample = raft::ceildiv(n_rows * kSaturation, n_rows + kDelay);
```
achirkin: This looks to overcomplicate the problem a bit. Why not just take a fixed sample size?
```suggestion
  int64_t n_sample = std::min<int64_t>(n_rows, 20000);
```

cpp/src/neighbors/ivf_pq/ivf_pq_fp16_overflow.cuh
```diff
@@ -0,0 +1,129 @@
...
+
+namespace cuvs::neighbors::ivf_pq::detail {
+
+/**
+ * Estimate max_i ||x_i||^2 over the dataset.
+ */
+template <typename DataT, typename Accessor>
+float estimate_max_squared_norm(
+  raft::resources const& handle,
+  raft::mdspan<const DataT, raft::matrix_extent<int64_t>, raft::row_major, Accessor> dataset)
```
achirkin: What about the other distance types? Would it make sense to generalize it to pass the whatever distance type is requested by the constructed index?

cpp/src/neighbors/ivf_pq/ivf_pq_fp16_overflow.cuh
achirkin: I agree with @tfeher that this looks a little too restrictive. We also have to keep in mind that we search only the `n_probe` nearest clusters for every query; even if we get a lot of +infs in the coarse search, top-k algo should recover well from it, as long as there exist `n_probe` clusters with finite distances.

achirkin (changes requested):
Hi @huuanhhuyn, thanks for tackling this long-standing problem!
I have some doubts about the approach though. Have you considered using the IVF-PQ search or index itself to decide if FP16 is permissible? Examples:
1) Call IVF-PQ search on a small subset of queries and detect if any of the results are infinite/invalid; if so, adjust the search params to increase one or both internal dtypes. This is agnostic of the selected distance type.
2) Calculate the distance maximum like you do here, but use the cluster centers of the IVF-PQ index rather than the source data. This gives a small number of points that cover the dataset well.
3) A combination of 1+2: call IVF-PQ search using its own cluster centers as queries.
4) Calculate per-component min/max of the cluster centers to get a bounding box; use the diagonal of the bounding box as a proxy for the spread of points in the dataset.

cpp/src/neighbors/ivf_pq/ivf_pq_fp16_overflow.cuh
huuanhhuyn: I removed the margin

cpp/src/neighbors/ivf_pq/ivf_pq_fp16_overflow.cuh
huuanhhuyn: This catches the overflow on IVF-PQ search only. `build_knn_graph` by IVF_PQ only allows 3 distance types [cagra_build.cuh#L1635](https://github.com/huuanhhuyn/cuvs/blob/677f67ef80d6ac293a12ddafc8a7d085b38add60/cpp/src/neighbors/detail/cagra/cagra_build.cuh#L1635) (L2Expanded, CosineExpanded, InnerProduct).

HNSW excludes further the use of Cosine distance [hnsw.hpp#L123](https://github.com/huuanhhuyn/cuvs/blob/677f67ef80d6ac293a12ddafc8a7d085b38add60/cpp/src/neighbors/detail/hnsw.hpp#L123)

I have refactored the code for explicitness with a `switch` around line 104 of the same file.

Tested: overflow detected with the remaining distance types - L2Expanded and InnerProduct

huuanhhuyn: Probe search approach would be distance type agnostic. So I resolve this issue.

tfeher (approved):
Thanks Huy for the update, LGTM!

cpp/src/neighbors/ivf_pq/ivf_pq_fp16_overflow.cuh
```diff
@@ -0,0 +1,87 @@
...
+ * Runs a small probe search against an already-built IVF-PQ index with current distance types,
+ * and reports whether any returned distance is non-finite (inf/NaN).
+ */
+template <typename DataT, typename Accessor>
+bool detect_fp16_overflow(
+  raft::resources const& handle,
+  const cuvs::neighbors::ivf_pq::index<int64_t>& index,
+  cuvs::neighbors::ivf_pq::search_params search_params,
+  raft::mdspan<const DataT, raft::matrix_extent<int64_t>, raft::row_major, Accessor> dataset,
+  uint32_t node_degree)
```
achirkin:
```suggestion
  uint32_t k)
```

cpp/src/neighbors/ivf_pq/ivf_pq_fp16_overflow.cuh
```diff
@@ -0,0 +1,87 @@
...
+  raft::common::nvtx::range<cuvs::common::nvtx::domain::cuvs> fun_scope("fp16_ovfl_detect");
+  const int64_t n_rows = dataset.extent(0);
+  if (n_rows == 0) { return false; }
+
+  auto stream       = raft::resource::get_cuda_stream(handle);
+  const int64_t dim = dataset.extent(1);
+
+  constexpr int64_t kMaxSampleQueries = 128;
+  const int64_t n_sample              = std::min<int64_t>(n_rows, kMaxSampleQueries);
+  const uint32_t top_k = std::min<uint32_t>(static_cast<uint32_t>(n_rows), node_degree + 1);
```
achirkin:
```suggestion
  const uint32_t top_k = std::min<uint32_t>(static_cast<uint32_t>(n_rows), k);
```

cpp/src/neighbors/detail/cagra/cagra_build.cuh
```diff
@@ -1664,6 +1665,24 @@ void build_knn_graph(
...
   auto index = cuvs::neighbors::ivf_pq::build(res, pq.build_params, dataset);
 
+  // Empirically detect FP16 distance overflow on the just-built index: run a small FP16 probe
+  // search and downgrade the internal/coarse dtypes to FP32 if any distance comes back non-finite.
+  // This observes the actual computation, so it is agnostic of the selected distance type.
+  if (pq.search_params.internal_distance_dtype == CUDA_R_16F ||
+      pq.search_params.coarse_search_dtype == CUDA_R_16F) {
+    {
+      const bool fp16_overflow = cuvs::neighbors::ivf_pq::helpers::detect_fp16_overflow(
+        res, index, pq.search_params, dataset, node_degree);
```
achirkin:
```suggestion
        res, index, pq.search_params, dataset, node_degree + 1);
```

achirkin (approved):
Pre-approving to save time, because overall it looks good. A small nitpick: please rename the argument of `detect_fp16_overflow` from `node_degree` to something else, e.g. `k`. This function is part of IVF-PQ, and hence is not aware of node degrees or graphs in general. The `k = node_degree + 1` logic can live outside it in `cagra_build.cuh`.

PR #2248, opened 2026-06-17
Add Gaussian Mixture Models (GMM)

Intron7:
This adds GMM to cuvs. This a pure fp32/fp64 cuda-core version. It already implements some of the memory advance from Flash-GMM.


cpp/include/cuvs/cluster/gmm.hpp
```diff
@@ -0,0 +1,335 @@
...
+ * @{
+ */
+
+/** Covariance parameterization of the mixture components. */
+enum class covariance_type { FULL = 0, TIED = 1, DIAG = 2, SPHERICAL = 3 };
+
+/** Strategy used to initialize the responsibilities before EM. */
+enum class init_method {
+  /** Run k-means (itself seeded with k-means++) and use the hard labels. */
+  KMeans = 0,
```
cjnolet, line 39: Wow, even implemented the k-means initialization strategy. This is great!

cpp/include/cuvs/cluster/gmm.hpp
```diff
@@ -0,0 +1,335 @@
...
+  /** Random per-sample-normalized responsibilities. */
+  Random = 2,
+  /** Pick n_components samples at random as one-hot responsibilities. */
+  RandomFromData = 3,
+};
+
+/** Hyper-parameters for the Gaussian mixture EM solver. */
+struct params {
+  /** The number of mixture components (at most 65535). Default: 1. */
+  int n_components = 1;
```
cjnolet, line 56: Why 1?

cpp/include/cuvs/cluster/gmm.hpp
```diff
@@ -0,0 +1,335 @@
...
+namespace cluster {
+namespace gmm {
+
+/**
+ * @defgroup gmm_params Gaussian mixture hyperparameters
+ * @{
+ */
+
+/** Covariance parameterization of the mixture components. */
+enum class covariance_type { FULL = 0, TIED = 1, DIAG = 2, SPHERICAL = 3 };
```
cjnolet: Document what each of these mean, please.

cpp/include/cuvs/cluster/gmm.hpp
```diff
@@ -0,0 +1,335 @@
...
+  /** Seed to the random number generator. Default: 0. */
+  uint64_t seed = 0;
+};
+
+/**
+ * @}
+ */
+
+/**
+ * @defgroup gmm Gaussian mixture model APIs
```
cjnolet, line 82: How do these get parsed in the docs? Is this showing up in a special section?

cpp/include/cuvs/cluster/gmm.hpp
```diff
@@ -0,0 +1,335 @@
...
+ *
+ * The covariance-shaped buffers (``covariances``, ``precisions_chol``,
+ * ``precisions``) are passed as flat device vectors because their logical
+ * shape depends on ``params::cov_type``. With ``K = n_components`` and
+ * ``d = n_features`` the expected lengths are (row-major):
+ *
+ *   - ``FULL``:      K * d * d   (logically (K, d, d))
+ *   - ``TIED``:      d * d       (logically (d, d))
+ *   - ``DIAG``:      K * d       (logically (K, d))
+ *   - ``SPHERICAL``: K           (logically (K,))
```
cjnolet, line 93: Empirically, how does this impact the quality vs perf. It seems like the best perf would be K, right? But does that come at a significant hit to quality?

cpp/src/cluster/gmm_impl.cuh
```diff
@@ -0,0 +1,1613 @@
...
+inline const char* precision_error_message()
+{
+  return "Fitting the mixture model failed because some components have ill-defined empirical "
+         "covariance (for instance caused by singleton or collapsed samples). Try to decrease the "
+         "number of components, increase reg_covar, or scale the input data.";
+}
+
+// ----- cuBLAS gemm / gemv wrappers -----
+template <typename T>
+cublasStatus_t cublas_gemm(cublasHandle_t h,
```
cjnolet: Can weuse the raft cublas/cusolver wrappers for this? They protect us against future updates to these functions, which have happend in the past (rather than having to update everywhere that called these libraries, we can just update the central wrapper). I believe there should already be wrappers for all of these functions.

cpp/include/cuvs/cluster/gmm.hpp
Intron7: Yes — spherical is the best perf (0.176s vs 1.31s for full at n=2M, d=64, K=32, fp32 on an RTX PRO 6000; tied/diag/spherical are all within a few % of each other since those fits are input-bandwidth-bound). Whether that costs quality depends entirely on the data, same trade-off as sklearn's covariance_type: with separated or isotropic clusters, spherical matches full's ARI exactly and only gives up held-out density; with overlapping anisotropic clusters, ARI collapses from 0.90 (full) to 0.03 (spherical). That's why we expose the same knob with the same default (full) as scikit-learn and document the trade-off on the docs page.

cpp/src/cluster/gmm_impl.cuh
Intron7: done

cpp/include/cuvs/cluster/gmm.hpp
Intron7: added some more

cpp/include/cuvs/cluster/gmm.hpp
Intron7: did add some comments

cpp/include/cuvs/cluster/gmm.hpp
Intron7: its what sklearn does as their default

cpp/include/cuvs/cluster/gmm.hpp
Intron7: thank you

cpp/src/cluster/gmm_impl.cuh
```diff
@@ -0,0 +1,1445 @@
...
+         "covariance (for instance caused by singleton or collapsed samples). Try to decrease the "
+         "number of components, increase reg_covar, or scale the input data.";
+}
+
+// ----- batched cuSOLVER potrf / cuBLAS trsm wrappers -----
+// RAFT's central cuBLAS/cuSOLVER wrappers (raft/linalg/detail/*_wrappers.hpp) cover
+// every non-batched call used here; they have no potrfBatched / trsmBatched
+// equivalents, so those two are wrapped locally in the same style.
+template <typename T>
+cusolverStatus_t potrf_batched(cusolverDnHandle_t h,
```
cjnolet, line 72: Sorry to be nitpicky here, but all calls to math libs should be through the cublas, cusolver, cusparse wrappers in raft. If any wrappers are missing functions, we should add them. This centralization has proven to be pretty key in the past for making quick fixes in the event th math libs change APIs suddenly (or if we find bugs or perf improvements for certian calls).

cpp/src/cluster/gmm_impl.cuh
```diff
@@ -0,0 +1,1445 @@
...
+  } else {  // TIED, d <= 128: shared Cholesky U -> transform X̃ = X·U, μ̃ = means·U,
+            // then ‖Uᵀ(x-μ_k)‖² is Euclidean ‖x̃-μ̃_k‖² -> same register-tiled kernel.
+    cublasHandle_t cublas = raft::resource::get_cublas_handle(handle);
+    rmm::device_uvector<T> xt((size_t)n * d, stream);
+    rmm::device_uvector<T> mut((size_t)K * d, stream);
+    rmm::device_uvector<T> ones_pc(K, stream);
+    rmm::device_uvector<T> const_k(K, stream);
+    thrust::fill(thrust::cuda::par.on(stream), ones_pc.data(), ones_pc.data() + K, T(1));
+    T one = T(1), zero = T(0);
+    RAFT_CUBLAS_TRY(raft::linalg::detail::cublasgemm(cublas,
```
cjnolet: Are we able to use the publicly exposed gemm from raft?

cpp/src/cluster/gmm_impl.cuh
```diff
@@ -0,0 +1,1445 @@
...
+                  T beta)
+{
+  cudaStream_t stream   = raft::resource::get_cuda_stream(handle);
+  cublasHandle_t cublas = raft::resource::get_cublas_handle(handle);
+  covariance_type ct    = params.cov_type;
+  T one                 = T(1);
+
+  RAFT_CUBLAS_TRY(raft::linalg::detail::cublasgemv(
+    cublas, CUBLAS_OP_N, K, n, &one, resp, K, ws.ones.data(), 1, &beta, ws.N_k.data(), 1, stream));
+  RAFT_CUBLAS_TRY(raft::linalg::detail::cublasgemm(cublas,
```
cjnolet: @divyegala at some point we should consider exposing public versions. I forgot all these wrappers are behind the detail APIs. 

Ideally, we'd expose first-class primitive wrappers instead of the math libs wrappers directly. Like- we have eigensolver wrappers already.

cpp/src/cluster/gmm_impl.cuh
```diff
@@ -0,0 +1,1445 @@
...
+    <<<dim3(K), dim3(256), 0, stream>>>(ws.N_k.data(), ws.num.data(), weights, means, eps, n, d, K);
+  RAFT_CUDA_TRY(cudaPeekAtLastError());
+
+  // FULL: covariance is finalized after its separate centered pass; nothing here.
+  if (ct == covariance_type::TIED) {
+    rmm::device_uvector<T> weighted_outer((size_t)d * d, stream);
+    detail::scale_rows_by_kernel<T>
+      <<<dim3(K), dim3(256), 0, stream>>>(means, ws.N_k.data(), eps, d, K, ws.scaled_means.data());
+    RAFT_CUDA_TRY(cudaPeekAtLastError());
+    RAFT_CUBLAS_TRY(raft::linalg::detail::cublasgemm(cublas,
```
cjnolet: Are we able to use the raft linalg/gemm API for this? Or is there something special we're doing here that the gemm API doesn't cover?

cpp/src/cluster/gmm_impl.cuh
divyegala: We have a public GEMM https://github.com/NVIDIA/raft/blob/aa9a6d5083fb445cd1948e7e47053abb7736cb8f/cpp/include/raft/linalg/gemm.cuh#L221-L226 and GEMV https://github.com/NVIDIA/raft/blob/aa9a6d5083fb445cd1948e7e47053abb7736cb8f/cpp/include/raft/linalg/gemv.cuh#L240-L245

Is there something missing from these APIs?

cpp/src/cluster/gmm_impl.cuh
Intron7: Switched to the public gemm caller

cpp/src/cluster/gmm_impl.cuh
Intron7: Switched to the public gemm caller

cpp/src/cluster/gmm_impl.cuh
Intron7: switched this one to gemm

cpp/include/cuvs/cluster/gmm.hpp
```diff
@@ -0,0 +1,348 @@
...
+ * @param[out]   lower_bound     Per-sample average log-likelihood of the best
+ *                               fit.
+ * @param[out]   n_iter          Number of EM iterations of the best fit.
+ * @param[out]   converged       Whether the best fit converged within
+ *                               ``params.tol``.
+ * @param[in]    warm_start      Use the incoming weights/means/covariances as
+ *                               the single initialization.
+ */
+template <typename T>
+void fit(raft::resources const& handle,
```
cjnolet, line 167: Instead of putting templates for computational functions in the header files, please just use direct overloads. The extern template just isn't consistent with the other APIs in cuVs, and it makes it more challenging to read overall.

cpp/src/cluster/gmm_double.cu
```diff
@@ -0,0 +1,51 @@
...
+ */
+
+#include "gmm.cuh"
+
+#include <cuvs/cluster/gmm.hpp>
+#include <cuvs/core/export.hpp>
+
+namespace cuvs::cluster::gmm {
+
+template CUVS_EXPORT void fit<double>(raft::resources const&,
```
cjnolet: As mentioned in the prior comment, please just define the functions with the concrete types up front. No need for templates at all (we designed RAFT this way because it's header only, but since cuVS is a standard C++ library with compiled binaries, we don't need to use the templates at all.

cpp/src/cluster/gmm_impl.cuh
```diff
@@ -0,0 +1,1348 @@
...
+template <typename T>
+cusolverStatus_t potrf_batched(cusolverDnHandle_t h,
+                               cublasFillMode_t uplo,
+                               int nn,
+                               T* Aarray[],
+                               int lda,
+                               int* infoArray,
+                               int batch,
+                               cudaStream_t stream);
+template <>
```
cjnolet, line 80: The good thing is that this should be largely copy/paste into the raft cusolver wrappers, so it won't require a ton of work- just need to get the PR opened for this in RAFT and then change the include and call the functions in here. I know the wrappers on the raft side are in the detail API... we can clean that part up later. But we should at least get these in the cusolver_wrappers in RAFT so we can centralize them.

cpp/src/cluster/gmm_impl.cuh
```diff
@@ -0,0 +1,1348 @@
...
+    precision_cholesky_full_batched<T>(handle, covariances, prec_chol, d, K, ws);
+    detail::log_det_full_kernel<T>
+      <<<dim3(K), dim3(REDUCE_BLOCK), 0, stream>>>(prec_chol, d, K, log_det);
+  } else if (ct == covariance_type::TIED) {
+    precision_cholesky_one<T>(handle, covariances, prec_chol, d, ws);
+    detail::log_det_tied_kernel<T>
+      <<<dim3(1), dim3(REDUCE_BLOCK), 0, stream>>>(prec_chol, d, K, log_det);
+  } else {
+    size_t total = cov_elems(ct, d, K);
+    T min_var    = thrust::reduce(thrust::cuda::par.on(stream),
```
cjnolet: Can we use raft::reduce here? It's better to call raft functions so we can centralize calls to libraries we don't have direct control over as much as possible.

cpp/src/cluster/gmm_impl.cuh
```diff
@@ -0,0 +1,1348 @@
...
+}
+
+// ===========================================================================
+// EM driver
+// ===========================================================================
+template <typename T>
+T mean_device(raft::resources const& handle, const T* v, int n)
+{
+  cudaStream_t stream = raft::resource::get_cuda_stream(handle);
+  T sum = thrust::reduce(thrust::cuda::par.on(stream), v, v + n, T(0), thrust::plus<T>());
```
cjnolet: Same here- please use raft::reduce.

Intron7: done

cpp/src/cluster/gmm_impl.cuh
Intron7: done

cpp/src/cluster/gmm_impl.cuh
Intron7: I'll added a TODO here. I'll create a follow up raft PR

cpp/src/cluster/gmm_double.cu
Intron7: done

cpp/include/cuvs/cluster/gmm.hpp
Intron7: done

cjnolet (approved):

PR #2249, opened 2026-06-17
Use new `top_1_nn` primitive in KMeans

divyegala:
This PR adds infrastructure built on top of existing JIT LTO architecture to generate kernels using `cutile-python` at build time, and embed them in the C++ library to make them callable from C++.

Merge https://github.com/NVIDIA/cuvs/pull/2548 first.


cpp/tests/CMakeLists.txt
```diff
@@ -120,7 +120,7 @@ ConfigureTest(
 ConfigureTest(
   NAME CLUSTER_TEST
   PATH cluster/kmeans.cu cluster/kmeans_balanced.cu cluster/kmeans_find_k.cu cluster/linkage.cu
-       cluster/connect_knn.cu cluster/spectral.cu
+       cluster/connect_knn.cu cluster/spectral.cu cluster/soa_unpack_trace.cu
```
ahendriksen: `cluster/soa_unpack_trace.cu` appears to be uncommitted?

cpp/src/distance/detail/fused_distance_nn/cutile/fused_1nn_kernel.py
```diff
@@ -0,0 +1,182 @@
...
+            if metric_code == METRIC_INNER_PRODUCT:
+                score = -accumulator
+            else:
+                b_norm = ct.load(
+                    B_norm, index=(n,), shape=(tn,), padding_mode=zero_pad
+                )
+                if metric_code == METRIC_L2_EXPANDED:
+                    # The A norm is constant across centroids. Reduce
+                    # 0.5 * ||y||^2 - dot(x, y), then recover full L2 once.
+                    score = (0.5 * b_norm)[None, :] - accumulator
```
huuanhhuyn, line 131: I guess for L2, `b_norm`and `a_norm` below are not yet squared? Because for cosine, we dropped the squared root on `b_norm`

cpp/src/distance/detail/fused_distance_nn/cutile/export_fused_1nn.py
```diff
@@ -0,0 +1,271 @@
...
+        stride_lower_bound_incl=(0, None),
+        alias_groups=(),
+        may_alias_internally=False,
+        stride_constant=(None, 1),
+        stride_divisible_by=(
+            _elem_stride_divisible_for_tma(elem_dtype)
+            if require_tma_friendly_pitch
+            else (1, 1)
+        ),
+        shape_divisible_by=(1, 1),
```
huuanhhuyn: Should we make `shape_divisible_by=(2, 1)` for `sm_80` and `sm_89` when `elem_dtype` is float16 and for strict ABI? This should already improve those sm's for 13.4 and float16

cpp/src/distance/detail/fused_distance_nn/cutile/fused_1nn_kernel.py
```diff
@@ -0,0 +1,182 @@
...
+                else:
+                    # Defer the A-norm division until after selecting the
+                    # winning centroid.
+                    score = -(accumulator / b_norm[None, :])
```
huuanhhuyn: Nit: I guess `accumulator / (-b_norm)[None, :]` might reduce slightly the number of flops as  `accumulator` has the shape [tm, tn] while `b_norm` has [tn]

cpp/src/cluster/detail/kmeans_balanced.cuh
```diff
@@ -99,55 +99,118 @@ inline std::enable_if_t<std::is_floating_point_v<MathT>> predict_core(
...
-      raft::linalg::map(handle,
-                        raft::make_const_mdspan(minClusterAndDistance.view()),
-                        raft::make_device_vector_view<LabelT, IdxT>(labels, n_rows),
-                        raft::compose_op<raft::cast_op<LabelT>, raft::key_op>());
+      auto nearest_dist =
+        raft::make_device_mdarray<MathT, IdxT>(handle, mr, raft::make_extents<IdxT>(n_rows));
+
+      if constexpr (std::is_same_v<LabelT, IdxT>) {
+        auto labels_view = raft::make_device_vector_view<IdxT, IdxT>(labels, n_rows);
+        cuvs::cluster::kmeans::detail::minClusterAndDistanceCompute<MathT, IdxT>(
```
tarang-jain: One small thing to fix along the way: we can use the public API here: `min_cluster_and_distance`

cpp/src/distance/detail/fused_distance_nn/cutile/export_fused_1nn.py
divyegala: If this will be fixed in `13.5` anyway, I think we can just wait and document the requirement to make sure we don't have too many conditionals lying around.

huuanhhuyn: The fix for `float16` is already available in `13.4` by adding `shape_divisible_by=(2, 1)`. `13.5` includes the fix for `float32` without additional code change. 

Please feel free to decide. I guess we decided to tolerate a performance drop for `Ampere` and `L40s` until an update to `13.5`.

cpp/src/distance/detail/fused_distance_nn/cutile/fused_1nn_kernel.py
```diff
@@ -0,0 +1,182 @@
...
+    if matrix_layout not in ("strict", "relaxed"):
+        raise ValueError(f"Unsupported matrix_layout {matrix_layout!r}")
+
+    acc_dtype = ct.float32
+    idx_dtype = _idx_dtype(index_type)
+    out_dist_dtype = ct.float16 if data_type == "half" else ct.float32
+    core_shape = (tile_m, tile_n)
+    best_shape = (tile_m, 1)
+
+    @ct.kernel(occupancy=ct.ByTarget(sm_100=2, sm_120=2))
```
huuanhhuyn: May you please remove `sm_100=2`? This would harm performance on B100. I overlooked it.

cpp/src/distance/detail/fused_distance_nn/cutile/fused_1nn_kernel.py
divyegala: The kernel receives squared norms for L2, but not for cosine

cpp/src/distance/detail/fused_distance_nn/cutile/export_fused_1nn.py
divyegala: Captured here https://github.com/NVIDIA/cuvs/issues/2431

huuanhhuyn: It wouldn't be much additional conditions. `shape_divisible_by` could be just a flipped version of `stride_divisible_by` - i.e. if `shape_divisible_by=(A,1)` then `shape_divisible_by=(1,A)`.

This would benefit perfs on `sm_80` till `sm_89` for the strict ABI.

Nit: In case you decide to add it, `_elem_stride_divisible_for_tma` and `require_tma_friendly_pitch` should include `ldgsts` in the namings for clarity? `TMA` for `sm_90` and above and `LDGSTS`  for `sm_80` and above

cpp/src/cluster/kmeans.cuh
```diff
@@ -331,30 +335,69 @@ void min_cluster_distance(raft::resources const& handle,
...
+      auto batch_cost    = raft::make_device_scalar<DataT>(handle, DataT{0});
+      auto centroids_i32 = raft::make_device_matrix_view<const DataT, int>(
+        centroids.data_handle(), static_cast<int>(n_clusters), static_cast<int>(n_features));
+      const IndexT max_batch_rows = max_i32 / n_clusters;
+
+      for (IndexT offset = 0; offset < n_samples; offset += max_batch_rows) {
+        const int batch_rows = static_cast<int>(std::min(max_batch_rows, n_samples - offset));
+        auto X_i32           = raft::make_device_matrix_view<const DataT, int>(
+          X.data_handle() + offset * n_features, batch_rows, static_cast<int>(n_features));
+
```
tarang-jain, line 381: I think this leads to numerical overflows. Later on to compute the matrix we want this product for resizing in kmeans_common.cuh
`L2NormBuf_OR_DistBuf.resize(dataBatchSize * centroidsBatchSize, stream);`

This can overflow. I ran into an overflow while I was trying to profile e2e after merging your branch.

tarang-jain (changes requested):

cpp/src/cluster/kmeans.cuh
divyegala: Can you provide a stack trace?

conda/environments/all_cuda-133_arch-aarch64.yaml
```diff
@@ -15,8 +15,10 @@ dependencies:
 - cuda-nvrtc-dev
 - cuda-nvtx-dev
 - cuda-profiler-api
+- cuda-tileiras
```
huuanhhuyn, line 18: is `tileiras`13.3 (maybe 13.3.36) ?

- afaik the alloyDB customer aims to use CAGRA on L40s for its performance-to-cost reason. A jump from 13.3 to 13.5 would deliver 2x-4x performance for K-Means predict on L40s and Ampere.
- Looking at the cuda toolkit release history, I guess 13.5 will only be released after 3-6 months?

Should we bake 13.5 to cuvs already now? E.g. Can we upload 13.5 to a public RAPISAI artifactory and install it from there? Once 13.5 is publicly released, we could switch back to the public conda package.

cpp/src/distance/detail/fused_distance_nn/cutile/fused_1nn_cutile_matrix.json
```diff
@@ -0,0 +1,535 @@
...
+        "cc_major": 8,
+        "cc_minor": 6,
+        "arch_tag": "cutile_arch_8_6"
+      },
+      {
+        "output_format": "cubin",
+        "artifact_ext": "cubin",
+        "artifact_basename": "@data_type@_@index_abbrev@_@abi_abbrev@_@gpu_code@",
+        "register": "cubin",
+        "gpu_code": "sm_89",
```
huuanhhuyn: Do we still need sm_89 cubin or only sm_86? 
(The same question for line 176)

cpp/src/distance/detail/fused_distance_nn/cutile/fused_1nn_cutile_matrix.json
```diff
@@ -0,0 +1,535 @@
...
+        "cc_major": 8,
+        "cc_minor": 0,
+        "arch_tag": "cutile_arch_8_0"
+      },
+      {
+        "output_format": "cubin",
+        "artifact_ext": "cubin",
+        "artifact_basename": "@data_type@_@index_abbrev@_@abi_abbrev@_@gpu_code@",
+        "register": "cubin",
+        "gpu_code": "sm_86",
```
huuanhhuyn: Shouldn't sm_86 belong to the next cell? This tile shape is [optimal ](https://gitlab-master.nvidia.com/Devtech-Compute/2026.06-flash-kmeans/-/blob/wip-a100/tune_shapes/best_shapes/summary.md?ref_type=heads)for sm_80 only
```
sm_80   relaxed (128x128x32)           strict (128x128x128, occ=2)
sm_86   relaxed (128x128x32, occ=2)    strict (128x128x32, occ=2)
```

cpp/src/distance/detail/fused_distance_nn/cutile/export_fused_1nn.py
huuanhhuyn: I checked that `shape_divisible_by=(1, 4)` for fp32 and `shape_divisible_by=(1,8)` for fp16 only enforces the second axis (K) to be divisible - i.e. within the current strict ABI

Therefore, it'd be free to already add it here and we will automatically receive 2x-4x perf on Ampere and L40s once tileiras 13.5 is baked into cuvs.

huuanhhuyn (commented):
Many thanks @divyegala for the comprehensive PR!

I double-verified the functional correctness of the kernel from this PR. The kernel also matches one-to-one the performance with the one from gitlab.

conda/environments/all_cuda-133_arch-aarch64.yaml
huuanhhuyn: Nvm. The target GPU for AlloyDB is RTX Pro 6000.

divyegala: cuvs 26.10 will be released with tileiras 13.4, which should bring a huge jump in performance from 13.3. I do not think it's the right move to bake in 13.5 kernels in cuvs at the moment, as then our end users will have no way to build the kernels from source. We cannot provide pre-release 13.5 artifacts anywhere until NVIDIA officially releases them.

cpp/src/distance/detail/fused_distance_nn/cutile/export_fused_1nn.py
divyegala: I thought this was only a stop-gap for 13.4 and we wouldn't need the extra hints for 13.5?

huuanhhuyn: No, both the hint and 13.5 tileiras are needed to unlock the performance jump (`LDGSTS.E.128` instead of `LDG`).

conda/environments/all_cuda-133_arch-aarch64.yaml
huuanhhuyn: It looks that 13.3.36 already performs on par with 13.4 (probably cutile also merged (some of) the features to the latest of 13.3). Maybe we are lucky to get the jump right after your merge. . Although I haven't thoroughly verified this.

cpp/src/distance/detail/fused_distance_nn/cutile/export_fused_1nn.py
divyegala: Okay got it, I will add it

cpp/src/cluster/kmeans.cuh
tarang-jain: Shared offline!

cpp/src/distance/detail/fused_distance_nn/cutile/export_fused_1nn.py
huuanhhuyn: Thank you.

cpp/src/distance/detail/fused_distance_nn/cutile/fused_1nn_kernel.py
```diff
@@ -0,0 +1,189 @@
...
+            best_idx = ct.where(update, n * tn + curr_idx, best_idx)
+
+        if metric_code == METRIC_INNER_PRODUCT:
+            out_dist = -best_dist
+        else:
+            a_norm = ct.load(
+                A_norm, index=(bidm,), shape=(tm,), padding_mode=zero_pad
+            )[:, None]
+            if metric_code == METRIC_L2_EXPANDED:
+                out_dist = a_norm + 2.0 * best_dist
```
dantegd, line 155: Should we clamp the reconstructed L2 distance to the same nonnegative/self-neighbor behavior as the existing expanded-L2 epilog before applying sqrt? 

A_norm/B_norm are reduced separately from the TF32 MMA, so when a centroid is copied from a data row (like in KMeans++ is a concrete path) the algebraically zero distance can land slightly below zero. L2SqrtExpanded then writes NaN, and L2Expanded can feed a negative value into seeding or inertia. The existing distance_ops/l2_exp.cuh has an explicit roundoff clamp for this case.

cpp/src/distance/detail/fused_distance_nn.cuh
```diff
@@ -49,36 +47,77 @@ void fusedDistanceNNImpl(OutT* min,
...
+                                                                           k,
+                                                                           workspace,
+                                                                           cutlass_redOp,
+                                                                           pairRedOp,
+                                                                           sqrt,
+                                                                           false,
+                                                                           cutlass_kvp_scratch,
+                                                                           stream);
       break;
+    case cuvs::distance::DistanceType::InnerProduct: break;
```
dantegd: Could we route InnerProduct to a real non-cuTile implementation when cuTile isn't built or available? 

On CUDA 12, try_fused_1nn_tile is compiled as false, so this function initializes the scratch output to (0, max), hits this empty case, and then unpacks that untouched sentinel. The new input_fp32_fused test includes InnerProduct unconditionally, so the supported CUDA 12 configuration returns incorrect results. What do you think about selecting the unfused path upstream for InnerProduct when cuTile is unavailable, or implementing the fallback here?

c/src/cluster/kmeans.cpp
```diff
@@ -282,10 +372,24 @@ extern "C" cuvsError_t cuvsKMeansPredict(cuvsResources_t res,
 {
   return cuvs::core::translate_exceptions([=] {
     auto dataset = X->dl_tensor;
+    const bool use_int64_index =
+      kmeans_predict_uses_int64_index(X, centroids, labels, params->n_clusters);
     if (dataset.dtype.code == kDLFloat && dataset.dtype.bits == 32) {
-        _predict<float>(res, *params, X, sample_weight, centroids, labels, normalize_weight, inertia);
+      if (use_int64_index) {
+        _predict<float, int64_t, int64_t>(
```
dantegd: Should we decouple the internal index type from the caller's labels dtype here? use_int64_index can be true solely because X or centroids exceeds the int32 span, even though validate_kmeans_labels_dtype accepted an int32 labels tensor. This branch then instantiates LabelsT=int64_t, so from_dlpack rejects the accepted int32 buffer, hierarchical prediction reaches the same combination but explicitly rejects non-int32 labels. 

What do you think about dispatching the index and label choices separately, or using an int64 temporary and casting into the caller's labels buffer?

cpp/src/distance/detail/fused_distance_nn/cutile/fused_1nn_tile.cu
```diff
@@ -0,0 +1,285 @@
...
+                        cuvs::distance::DistanceType metric,
+                        bool is_sqrt,
+                        void* index_workspace,
+                        cudaStream_t stream)
+{
+  if (!cuvs::detail::jit_lto::cutile_launch_available_on_current_device()) { return false; }
+  static_assert(std::is_same_v<IdxT, int> || std::is_same_v<IdxT, int64_t>);
+
+  constexpr int strict_pitch_elements = 16 / sizeof(DataT);
+  const bool use_strict_abi           = k % strict_pitch_elements == 0;
```
dantegd, line 328: Both the strict and relaxed signatures declare base_addr_divisible_by=16 for matrices, norms, indices, and distances, but this selection only checks row pitch. Valid RAFT subviews or DLPack byte offsets can therefore launch under a false compiler assumption; the int64 loop below also makes xn + max_i32 and nearest_dist + max_i32 non-16-byte-aligned on its second batch. [cuTile documents](https://docs.nvidia.com/cuda/cutile-python/compilation.html#kernel-signatures) violating an exported signature assumption as undefined behavior: “Launching the exported kernel with an array that doesn’t satisfy this assumption would then result in undefined behavior.”

Should we validate base alignment for every non-null array before launching, or export a base-alignment-1 fallback? What do you think about making alignment part of eligibility before choosing either ABI?

cpp/src/cluster/detail/kmeans_balanced.cuh
```diff
@@ -186,14 +247,24 @@ auto calc_minibatch_size(const raft::resources& handle,
...
-        // fusedL2NN needs a mutex and a key-value pair for each row.
-        mem_per_row += sizeof(int);
-        mem_per_row += sizeof(raft::KeyValuePair<IdxT, MathT>);
-      } else {
-        // unfused path needs a full GEMM output (distance matrix row).
-        mem_per_row += sizeof(MathT) * n_clusters;
+    case distance::DistanceType::L2SqrtExpanded:
+    case distance::DistanceType::InnerProduct: {
+      switch (use_fused<MathT, IdxT, IdxT>(handle, n_rows, n_clusters, dim, metric)) {
+        case FusedDistancePath::FusedCutile:
```
dantegd: Could we include all allocations performed by predict_core in mem_per_row? Even for FusedCutile with int32 indices, predict_core always allocates nearest_dist[n_rows]; when LabelT != IdxT it also allocates nearest_idx, and float L2/cosine cuTile materializes TF32 norms. Returning n_rows when the current estimate is zero bypasses the stated 512 MiB cap and can create full-size temporaries or OOM on the exact large-data path this batching heuristic is meant to protect. What do you think about charging at least nearest_dist, plus temporary indices/norms when applicable, before deriving the batch size?

dantegd (changes requested):
Could we make the path selection and workspace allocation agree on whether the cuTile launch is guaranteed to succeed? When minClusterDistanceCompute.cu selects FusedCutile, it doesn’t allocate the CUTLASS KVP scratch or mutex workspace. But if try_fused_1nn_tile() returns false, cpp/src/distance/detail/fused_distance_nn.cuh falls through to CUTLASS and expects those buffers to exist. That leaves us without a valid fallback if launcher construction fails, even though the try_ API suggests falling back is supported. Would it make sense to build or probe the launcher before committing to FusedCutile, or keep the fallback workspace allocated for as long as this path can still return false?

cpp/src/distance/detail/fused_distance_nn.cuh
divyegala: The unfused path does not support InnerProduct for some reason. Let me see if we can support or just throw for now.

c/src/cluster/kmeans.cpp
divyegala: Yes, far better way to do it.

cpp/src/distance/detail/fused_distance_nn/cutile/fused_1nn_tile.cu
divyegala: Oh I did not think at all about subviews or offsets, good catch. For now, let's just validate base pointer alignment. I can add to the follow-up issue to validate 1-byte alignment and what kind of cubin is produced by cutile?

cpp/src/cluster/detail/kmeans_balanced.cuh
```diff
@@ -99,55 +100,115 @@ inline std::enable_if_t<std::is_floating_point_v<MathT>> predict_core(
...
+                                                                     workspace);
+        raft::copy(
+          handle, raft::make_device_vector_view<LabelT, IdxT>(labels, n_rows), nearest_idx.view());
+      }
       break;
     }
     case cuvs::distance::DistanceType::InnerProduct: {
-      // TODO: pass buffer
-      rmm::device_uvector<MathT> distances(n_rows * n_clusters, stream, mr);
+      if (uses_fused_distance_nn(
```
dantegd: Could we make this decision using the same pointer-aware eligibility check as min_cluster_and_distance()? On CUDA 13, use_fused() can enter this branch before checking the actual X, centroid, and labels pointers. If any of those rejects cuTile, min_cluster_and_distance() falls back through use_legacy_fused(); for InnerProduct that means Unfused, which eventually calls pairwise_distance_kmeans() and throws because it only supports L2 metrics. CUDA 12 instead reaches the GEMM+argmin path below. What do you think about resolving final cuTile eligibility here and taking this existing GEMM fallback whenever it isn’t eligible?

cpp/src/cluster/detail/kmeans_balanced.cuh
```diff
@@ -164,36 +225,69 @@ inline std::enable_if_t<std::is_floating_point_v<MathT>> predict_core(
...
+        if constexpr (!std::is_same_v<LabelT, IdxT>) { mem_per_row += sizeof(IdxT); }
+      }
+      if (metric != distance::DistanceType::InnerProduct) {
+        // predict may need a minibatch-sized input-norm buffer before entering predict_core.
+        mem_per_row += sizeof(MathT);
+      }
+
+      switch (fused_path) {
+        case FusedDistancePath::FusedCutile:
+          // Conservatively budget the fallback in case the eventual pointer-aware probe fails.
```
dantegd: Could we budget the actual architecture-specific fallback here? On Blackwell—and on Hopper when both the row and cluster counts are below the legacy fused threshold—a pointer-aware cuTile rejection selects Unfused. That path allocates a batch-by-cluster distance matrix plus KVP output, but this case only charges int + KVP per row. Because predict_core() passes zero batch limits, the next allocation can exceed the 512 MiB target by a large factor. What do you think about budgeting the maximum of the cuTile scratch and the use_legacy_fused() scratch rather than assuming every fallback is CUTLASS?

cpp/include/cuvs/detail/jit_lto/tileir_compat.hpp
```diff
@@ -0,0 +1,108 @@
...
+ * True when a cuTile launch may be attempted for the given device: cuTile is enabled, the runtime
+ * is CUDA 13+, and either compatible same-family SASS exists (no driver JIT required) or the
+ * driver can JIT the embedded TileIR bytecode fallback.
+ */
+#if CUVS_CUTILE_ENABLED
+inline bool cutile_launch_available_for_arch(int cc_major, int cc_minor, int driver_version)
+{
+  if (!runtime_cuda13_or_newer()) { return false; }
+  if (has_embedded_cubin_for_arch(cc_major, cc_minor)) { return true; }
+  return tileir_fallback_available(driver_version);
```
dantegd, line 96: I wonder if we shoukd  gate the TileIR fallback on cuTile-supported compute capabilities? [RAPIDS CUDA 13 still supports Turing/SM75](https://docs.rapids.ai/platform-support/), while [cuTile requires compute capability 8.x through 12.x](https://docs.nvidia.com/cuda/cutile-python/quickstart.html#prerequisites). On SM75 with a new enough driver, this returns true, so KMeans selects FusedCutile and attempts to load TileIR instead of keeping the supported legacy CUDA path. What do you think about checking the supported architecture range before reporting availability?

cpp/include/cuvs/detail/jit_lto/cutile_module.hpp
```diff
@@ -0,0 +1,94 @@
...
+  }
+  return std::nullopt;
+}
+
+inline std::shared_ptr<rtcx::algorithm_launcher> load_cutile_launcher(
+  const CutileModuleImage& image, const std::string& kernel_symbol)
+{
+  cudaLibrary_t library{};
+  RAFT_CUDA_TRY(
+    cudaLibraryLoadData(&library, image.data, nullptr, nullptr, 0, nullptr, nullptr, 0));
```
dantegd: Should we preserve the nullable try_get_launcher() contract for expected module compatibility and JIT failures? 

If I'm not mistaken, RAFT_CUDA_TRY here turns errors such as cudaErrorNoKernelImageForDevice, cudaErrorUnsupportedPtxVersion, and cudaErrorJitCompilerNotFound into exceptions, so can_launch_fused_1nn_tile() never returns false and KMeans cannot take its legacy fallback. 

What do you think about treating expected compatibility/JIT errors as a failed build, unloading the library if kernel lookup fails, while still propagating genuine resource failures such as allocation errors?

cpp/src/cluster/detail/kmeans_balanced.cuh
```diff
@@ -99,55 +100,115 @@ inline std::enable_if_t<std::is_floating_point_v<MathT>> predict_core(
...
-        0,  // batch_samples (unused for fused reduction)
-        0,  // batch_centroids (unused for fused reduction)
-        workspace);
-
-      // Copy keys to output labels
-      raft::linalg::map(handle,
-                        raft::make_const_mdspan(minClusterAndDistance.view()),
-                        raft::make_device_vector_view<LabelT, IdxT>(labels, n_rows),
-                        raft::compose_op<raft::cast_op<LabelT>, raft::key_op>());
+      auto nearest_dist =
```
dantegd: Could we keep native KVP storage when cuTile isn’t selected? On CUDA 12 this always allocates nearest_dist—and nearest_idx when LabelT differs, then minClusterAndDistanceCompute() allocates another full KVP because it was given SoA outputs. 

The old path used the KVP output directly, so this adds avoidable per row scratch to every legacy L2/cosine balanced prediction; the hidden KVP also uses the current RMM resource rather than mr. 

What do you think about deciding the path first and using minClusterAndDistanceComputeKvp() for legacy paths, as process_batch() does?

cpp/src/cluster/detail/kmeans_common.cuh
```diff
@@ -685,66 +788,102 @@ __device__ void check_convergence(raft::device_scalar_view<const DataT> clusteri
...
+    compute_centroid_adjustments(handle,
+                                 batch_data,
+                                 batch_weights,
+                                 nearest_idx,
+                                 static_cast<IndexT>(centroid_sums.extent(0)),
+                                 centroid_sums,
+                                 weight_per_cluster,
+                                 batch_workspace,
+                                 /*reset_sums=*/false);
+    auto weighted_dist = raft::make_device_vector<DataT, IndexT>(handle, n_samples);
```
dantegd: We can  we avoid materializing weighted_dist on the cuTile path, no? 

The KVP branch just below feeds a transform iterator over distance × weight into computeClusterCostFromIterator(), and the same approach works with nearest_dist here. As written, every cuTile Lloyd batch carries an extra n_samples * sizeof(DataT) allocation. What do you think about using the same iterator pattern in both branches?

cpp/src/distance/detail/fused_distance_nn/cutile/export_fused_1nn.py
```diff
@@ -0,0 +1,296 @@
...
+    LDGSTS instead of LDG. Tail tiles remain masked in the kernel.
+
+    Odd D or general layouts need a separate relaxed export profile.
+    """
+    return ArrayConstraint(
+        elem_dtype,
+        ndim=2,
+        index_dtype=index_dtype,
+        stride_lower_bound_incl=(0, None),
+        alias_groups=(),
```
dantegd: Question: are callers allowed to use overlapping read-only dataset and centroid views? I couldn’t find a non-aliasing precondition on the KMeans APIs, while [an empty alias_groups sequence tells cuTile that this array cannot alias any other parameter](https://docs.nvidia.com/cuda/cutile-python/compilation.html). If that overlap is supported, passing the same storage for X and centroids invokes the exported kernel outside its declared signature. Would common alias groups for the read only inputs or an explicit eligibility check and documented precondition make the contract clearer?v

dantegd (changes requested):

cpp/src/distance/detail/fused_distance_nn/cutile/export_fused_1nn.py
divyegala: They are not, but when is this a realistic path? This is akin to adding `__restrict__` on each input pointer, right? When would centroids and X be the exact same pointer?

cpp/include/cuvs/cluster/kmeans.hpp
```diff
@@ -838,10 +847,10 @@ void predict(raft::resources const& handle,
  */
 void predict(raft::resources const& handle,
              const kmeans::params& params,
-             raft::device_matrix_view<const float, int64_t> X,
-             std::optional<raft::device_vector_view<const float, int64_t>> sample_weight,
-             raft::device_matrix_view<const float, int64_t> centroids,
-             raft::device_vector_view<int64_t, int64_t> labels,
+             raft::device_matrix_view<const float, int> X,
```
tfeher, line 850: Why do we need `int` as matrix extent type? I see that kmeans.hpp had such overloads already. But is there any reason to keep them, instead of just having `int64_t` matrix extents?

cpp/src/distance/detail/fused_distance_nn/cutile/fused_1nn_kernel.py
```diff
@@ -0,0 +1,191 @@
...
+                a = ct.load(
+                    A, index=(bidm, k), shape=(tm, tk), padding_mode=zero_pad
+                ).astype(dtype)
+                b_T = ct.load(
+                    B,
+                    index=(k, n),
+                    shape=(tk, tn),
+                    padding_mode=zero_pad,
+                    order=(1, 0),
+                ).astype(dtype)
```
huuanhhuyn, line 119: From `sm_80` to `sm_89`, the default rounding mode (`RN`) is not supported by the hardware. It should be `RA` for those GPUs. 
`ct.load().astype(dtype, rounding_mode=ct.RoundingMode.RA`)

In case this PR is merged after 13.4. Could you please add it? This change accounts for comprehensive perf improvements on those architectures.

cpp/src/cluster/detail/kmeans_common.cuh
```diff
@@ -685,66 +826,103 @@ __device__ void check_convergence(raft::device_scalar_view<const DataT> clusteri
...
-  int batch_centroids_param,
-  raft::device_vector_view<raft::KeyValuePair<IndexT, DataT>, IndexT> minClusterAndDistance,
-  raft::device_vector_view<const DataT, IndexT> L2NormBatch,
-  rmm::device_uvector<DataT>& L2NormBuf_OR_DistBuf,
-  rmm::device_uvector<char>& workspace,
-  raft::device_matrix_view<DataT, IndexT> centroid_sums,
-  raft::device_vector_view<DataT, IndexT> weight_per_cluster,
-  raft::device_scalar_view<DataT> clustering_cost,
-  rmm::device_uvector<char>& batch_workspace)
+void process_batch(raft::resources const& handle,
```
tfeher: Since we are refactoring this, can we switch to `raft::device_vector_view` instead of `rmm::device_uvector&`?

cpp/src/cluster/detail/kmeans_common.cuh
```diff
@@ -57,29 +59,101 @@
...
+/** Native result representation selected by fused 1-NN. */
+enum class Fused1nnOutputLayout : std::uint8_t {
+  /** Separate index and distance arrays (cuTile). */
+  Soa,
+  /** Native key/value-pair array (CUTLASS/SIMT and the unfused reducer). */
+  Kvp,
+};
+
+/** Norm representation required by the resolved fused-1NN implementation. */
+enum class Fused1nnNormPolicy : std::uint8_t {
```
tfeher, line 77: Is norm policy used? I only find references where we set its value, but I do not see it used.

cpp/src/cluster/detail/kmeans_common.cuh
```diff
@@ -57,29 +59,101 @@
...
+ * KMeans uses the result layout and byte counts to reuse its existing raw buffers. The fused
+ * implementation owns execution of the resolved path.
+ */
+template <typename IndexT>
+struct Fused1nnRequirements {
+  FusedDistancePath path{};
+  Fused1nnOutputLayout output_layout{};
+  Fused1nnNormPolicy norm_policy{};
+  size_t result_bytes{};
+  size_t result_alignment{};
```
tfeher, line 94: Is `result_alignment` used / planned to be used?

cpp/src/cluster/detail/kmeans_common.cuh
```diff
@@ -57,29 +59,101 @@
...
+  size_t distance_offset{};
+  size_t workspace_bytes{};
+  size_t workspace_alignment{};
```
tfeher, line 97: Is `workspace_alignment` actually used?

cpp/src/cluster/detail/kmeans_common.cuh
```diff
@@ -57,29 +59,101 @@
...
+  IndexT sample_tile{};
+  IndexT centroid_tile{};
```
tfeher, line 99: These are the M and N dims for the GEMM op, right? Could you add a comment clarifying that?

cpp/src/cluster/detail/kmeans.cuh
```diff
@@ -1084,49 +1079,86 @@ void kmeans_predict(raft::resources const& handle,
...
-    raft::make_device_scalar_view(clusterCostD.data()),
-    raft::value_op{},
-    raft::add_op{});
-
-  raft::linalg::map(
-    handle, labels, raft::key_op{}, raft::make_const_mdspan(minClusterAndDistance.view()));
+  const auto requirements = get_fused_1nn_requirements(
+    handle, X, centroids, metric, pams.batch_samples, pams.batch_centroids);
+
+  if (requirements.output_layout == Fused1nnOutputLayout::Soa) {
```
tfeher, line 1088: We have repeated dispatch like this. It would be great if we can avoid this. 

Kmeans consumers already want `labels` and `dist` arrays. The KVP branch exists because CUTLASS/SIMT write `raft::KeyValuePair`. Can we keep this as an implementation detail, not part of the `minClusterAndDistanceCompute` API  (by moving the associated map operations deeper in the CUTLASS/SIMT and unfused call chain)?

tfeher (changes requested):
Hi Divye, here is the first batch of my comments.

The different implementations make the dispatch of the `fused_1nn` operation overly complex. This issue existed before this PR, now it becomes even more complex.

Ideally we would only have cutile, and cublas + minreduce as fallback. But I understand that we need to keep the previous fused implementations to avoid perf regression when cuTile is not available. Still, we should try some refactoring to make the call chain simpler.

cpp/src/distance/detail/fused_distance_nn/cutile/fused_1nn_kernel.py
huuanhhuyn: 13.4 has been released.

cpp/src/cluster/detail/kmeans_common.cuh
```diff
@@ -676,7 +703,7 @@ __device__ void check_convergence(raft::device_scalar_view<const DataT> clusteri
  * @param[in]     batch_samples_param  Batch-samples param forwarded to minClusterAndDistanceCompute
  * @param[in]     batch_centroids_param Batch-centroids param forwarded to
  *                                      minClusterAndDistanceCompute
- * @param[inout]  minClusterAndDistance Work buffer [batch_size]
+ * @param[inout]  output_storage       Reusable backend-native assignment output storage
```
tarang-jain, line 711: Can we add to the docs what this buffer is and what size it is set to?

tarang-jain (changes requested):

KyleFromNVIDIA (approved):

cpp/src/distance/top_1_nn.cuh
```diff
@@ -35,8 +39,208 @@ struct Top1nnOutputTypes {
...
+  static constexpr bool supported                   = true;
+  static constexpr Top1nnOutputLayout output_layout = Top1nnOutputLayout::KeyValuePair;
+
+  struct key_op {
+    __host__ __device__ IdxT operator()(const key_value_type& value) const { return value.key; }
+  };
+
+  struct value_op {
+    __host__ __device__ DistT operator()(const key_value_type& value) const { return value.value; }
+  };
```
tfeher, line 63: Can't we just use `raft::key_op()` and `raft::value_op()`?

cpp/src/cluster/detail/kmeans_common.cuh
```diff
@@ -685,66 +717,49 @@ __device__ void check_convergence(raft::device_scalar_view<const DataT> clusteri
...
+                   raft::device_matrix_view<const DataT, IndexT> batch_data,
+                   raft::device_vector_view<const DataT, IndexT> batch_weights,
+                   raft::device_matrix_view<const DataT, IndexT> centroids,
+                   cuvs::distance::DistanceType metric,
+                   int batch_samples_param,
+                   int batch_centroids_param,
+                   rmm::device_uvector<char>& output_storage,
+                   raft::device_vector_view<const DataT, IndexT> L2NormBatch,
+                   rmm::device_uvector<DataT>& L2NormBuf_OR_DistBuf,
+                   rmm::device_uvector<char>& workspace,
```
tarang-jain, line 730: I did not realize that we had two workspaces here `workspace` and `batch_workspace`. Is this a bug?

cpp/src/cluster/detail/kmeans_balanced.cuh
```diff
@@ -411,17 +415,132 @@ void predict(const raft::resources& handle,
...
+    native_half ? max_minibatch_size : 0, stream, mem_res);
+  rmm::device_uvector<int> native_indices(native_half ? max_minibatch_size : 0, stream, mem_res);
+  rmm::device_uvector<char> native_workspace(0, stream, mem_res);
+  bool native_centers_ready     = false;
   const MathT* dataset_norm_ptr = nullptr;
   auto cur_dataset_ptr          = cur_dataset.data();
   for (IdxT offset = 0; offset < n_rows; offset += max_minibatch_size) {
     IdxT minibatch_size = std::min<IdxT>(max_minibatch_size, n_rows - offset);
 
+    if constexpr (native_half) {
```
tfeher, line 439: It would be easier to read if this new block goes into a separate function. E.g. `predict_core_half`

tfeher (approved):
Thanks Divye for the PR! I have just some cosmetic suggestions, these don't need to hold the PR. In general the PR looks good to me.

tarang-jain (approved):
Pre-approving. My comments were mostly lightweight and can even be addressed later.

cpp/src/cluster/detail/kmeans_common.cuh
divyegala: This predates this PR

tfeher (approved):
Thanks Divye for the update, LGTM!

dantegd (approved):

KyleFromNVIDIA (approved):

PR #2250, opened 2026-06-22
Switch Product Quantization VQ to mean() when n_centers = 1

lowener:
When running a PQ preprocessing operation, the VQ option can be used to act as a way to mean-center the dataset.
This proposed change will enable to do that mean-centering operation faster by using a direct call to `raft::stats::mean` instead of running the expectation-maximization steps.


tarang-jain (approved):

cjnolet:
> this proposed change will enable to do that mean-centering operation faster by using a direct call to raft::stats::mean

Thanks for the PQ @lowener. Can you please provide some benchmarks here to demonstrate the difference in the perf for this change?

cjnolet (approved):

lowener:
The speedup of course depend on the shape of the problem.
Building on a dataset of 1Mx1024, the gain is of 149ms in average. 
When used for a whole pq::build call, using `pq_bits=8`, `pq_dim=256` the speedup for the `build()` function is of
- 14.6% with `n_points_per_pq_cluster = 512` (131k points used for PQ, all points used for VQ)
- 4.4% with `n_points_per_pq_cluster = 3900` (all points of the dataset used for PQ, all points used for VQ)

cjnolet (approved):

PR #2255, opened 2026-06-23
Add missing memory estimates

huuanhhuyn:
**Add missing memory usage estimates:**

- IVF-PQ extend phase takes more memory than search phase for OpenAI 5M dataset
- Graph optimization debuging allocates large host memory when raft debug mode is enabled
image

- When cagra param "attach_dataset_on_build" is enabled, the graph is allocated on the device memory
image

- When data type is other than float (e.g. int8), cagra allocates addition kmeans trainset temporary buffer on large workspace
image


cpp/src/neighbors/detail/cagra/cagra_helpers.cpp
```diff
@@ -84,14 +84,57 @@ std::tuple<size_t, size_t, size_t, size_t> optimize_workspace_size(size_t n_rows
...
+                      + n_rows * sizeof(uint32_t)               // in_edge_count
+                      + graph_degree * sizeof(uint32_t);        // hist
+
+  size_t total_host       = mst_host + combine_host + debug_host_size;
   size_t total_host_fixed = mst_host_fixed + combine_host_fixed;
   size_t total_dev        = std::max(prune_dev, rev_dev + combine_dev);
   size_t total_dev_fixed  = std::max(prune_dev_fixed, combine_dev_fixed);
 
   return std::make_tuple(total_host, total_dev, total_host_fixed, total_dev_fixed);
 }
```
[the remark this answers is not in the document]
huuanhhuyn: Added closing brace

cpp/src/neighbors/detail/cagra/cagra_helpers.cpp
```diff
@@ -84,14 +84,57 @@ std::tuple<size_t, size_t, size_t, size_t> optimize_workspace_size(size_t n_rows
...
+  size_t n_rows_placeholder  = n_rows + ivf_pq::kIndexGroupSize * n_clusters;
+  size_t placeholder_dev     = n_rows_placeholder * bytes_per_row;
+  size_t n_rows_resize_lists = n_rows + kSpecAlignMax * n_clusters;
+  size_t resize_lists_dev    = n_rows_resize_lists * bytes_per_row;
+
+  // Placeholder freed before resize_list
+  size_t device_size = std::max(placeholder_dev, resize_lists_dev);
+
+  return device_size + workspace_size;
+}
```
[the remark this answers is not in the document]
huuanhhuyn: Added RAFT_EXPECTS

cpp/src/neighbors/detail/cagra/cagra_helpers.cpp
```diff
@@ -84,14 +84,60 @@ std::tuple<size_t, size_t, size_t, size_t> optimize_workspace_size(size_t n_rows
   }
   size_t combine_dev = combine_dev_fixed;
 
-  size_t total_host       = mst_host + combine_host;
+  size_t debug_host_size = 0;
+  if (raft::default_logger().should_log(rapids_logger::level_enum::debug)) {
```
achirkin, line 88:
```suggestion
  if (raft::default_logger().should_log(rapids_logger::level_enum::debug)) {
  // cagra::detail::graph::optimize() allocates extra memory to calculate
  // graph metrics when debug logging is enabled
```

cpp/src/neighbors/detail/cagra/cagra_helpers.cpp
```diff
@@ -84,14 +84,60 @@ std::tuple<size_t, size_t, size_t, size_t> optimize_workspace_size(size_t n_rows
...
+  constexpr size_t kSpecAlignMax           = 1024;
+
+  size_t n_rows     = dataset.extent(0);
+  size_t dim        = dataset.extent(1);
+  size_t pq_dim     = params.build_params.pq_dim;
+  size_t pq_bits    = params.build_params.pq_bits;
+  size_t rot_dim    = raft::round_up_safe<size_t>(dim, pq_dim);
+  size_t n_clusters = params.build_params.n_lists;
+
+  RAFT_EXPECTS(pq_dim > 0, "pq_dim should not be 0");
```
achirkin: nitpick: I don't think this check is necessary - it doesn't guard against anything dangerous below (no divisions) and its valid values are checked in IVF-PQ during build anyway.

cpp/src/neighbors/detail/cagra/cagra_helpers.cpp
```diff
@@ -84,14 +84,60 @@ std::tuple<size_t, size_t, size_t, size_t> optimize_workspace_size(size_t n_rows
...
+  // wastes n_cluster * alignment_size
+  size_t n_rows_placeholder  = n_rows + ivf_pq::kIndexGroupSize * n_clusters;
+  size_t placeholder_dev     = n_rows_placeholder * bytes_per_row;
+  size_t n_rows_resize_lists = n_rows + kSpecAlignMax * n_clusters;
+  size_t resize_lists_dev    = n_rows_resize_lists * bytes_per_row;
+
+  // Placeholder freed before resize_list
+  size_t device_size = std::max(placeholder_dev, resize_lists_dev);
+
+  return device_size + workspace_size;
```
achirkin: I think we need to be careful to not account for the same memory multiple times. As far as I remember, we add the whole workspace size as used somewhere above in the call chain. The intuition behind that is that the algorithms are always free to use all of the workspace size (e.g. by changing the batch size). Please make sure your changes are consistent with the rest of the logic in that regard.

achirkin (changes requested):
Hi @huuanhhuyn , I have some concerns about estimating the workspace memory, (see below). Otherwise, looks good.

cpp/src/neighbors/detail/cagra/cagra_helpers.cpp
huuanhhuyn: The AI review suggested adding this check.

cpp/src/neighbors/detail/cagra/cagra_helpers.cpp
huuanhhuyn: I don't think we have any duplicated estimate for the placeholder and resize lists within our `cagra_build_mem_usage()` call chain.  

The estimate would be the upper-bound for the real allocation because the `max_batch_size` could be shrinked further on the real path and the alignment "waste" will be less. The memory resource tracking shows that the over estimation is not too significant (e.g. 280MiB vs 230MiB).

huuanhhuyn: Do you think I should add the shrinking logics based on the total free mem? This would improve the over-estimation and it will still be the upper bound afterwards.

With the current version, if the smallest free mem is 1 GiB and the largest dim is 4000 then the max_batch_size could be over-estimated by approx. 4x (the real batch size is shrinked to 16k).

huuanhhuyn:
@achirkin Thanks for your review. Could you please look at it again?

cpp/src/neighbors/detail/cagra/cagra_helpers.cpp
```diff
@@ -137,6 +190,15 @@ inline std::pair<size_t, size_t> ivf_pq_build_mem_usage(
...
   size_t kmeans_host_mem    = kmeans_indices_host + kmeans_pinned_host;
 
+  // Extend phase
+  size_t extend_gpu_mem = params.build_params.add_data_on_build
+                            ? ivf_pq_extend_mem_usage(dataset, params, dtype_size)
+                            : 0;
+
+  // Add graph to index on GPU
+  size_t attach_graph_gpu_mem =
+    attach_dataset_on_build ? n_rows * graph_degree * sizeof(uint32_t) : 0;
```
tfeher: - `attach_dataset_on_build` would copy the dataset to the GPU not the graph. In case of OOM we fall back to not attaching the dataset. Therefore it is ok to ignore `attach_dataset_on_build`.
- IIUC, the [graph is attached unconditionally](https://github.com/NVIDIA/cuvs/blob/ff0c8076bf2f8d354bc21039ebffa14a1c9128a3/cpp/src/neighbors/detail/cagra/cagra_build.cuh#L2331) (i.e. copied to the device). We should change this: in practice we can build a graph even if it does not fit GPU memory (as long as CPU mem is large enough). This graph would be later useful for hnsw search on the CPU.

tfeher (changes requested):
Thanks Huy for working on these improvements. Overall it looks good, but I have one comment.

cpp/src/neighbors/detail/cagra/cagra_helpers.cpp
huuanhhuyn: The estimate was removed for simplification as discussed

cpp/src/neighbors/detail/cagra/cagra_helpers.cpp
huuanhhuyn: If I understand the code correctly, we attach the graph from host to device [here](https://github.com/huuanhhuyn/cuvs/blob/b966399a7058faa5eeeabe9114247166d6124808/cpp/include/cuvs/neighbors/cagra.hpp#L710) when `attach_dataset_on_build ` is set to True and the knn graph is on host before the update.

I am not sure about the decision point when the graph is on the host and when on the device. For deep 10M, it was on the host.

So it indeed copies the graph conditionally. Or do I mistake it somewhere?

cpp/src/neighbors/detail/cagra/cagra_helpers.cpp
```diff
@@ -208,8 +220,7 @@ std::pair<size_t, size_t> cagra_build_mem_usage(raft::resources const& res,
                                                              dtype,
                                                              pq_params,
                                                              cparams.graph_degree,
-                                                             cparams.intermediate_graph_degree,
-                                                             cparams.guarantee_connectivity);
+                                                             cparams.intermediate_graph_degree);
```
achirkin: What's the reason for the removal of this parameter?

huuanhhuyn: It was by accident. I will bring it back otherwise the CI won't pass any way

achirkin (approved):
Thanks, @huuanhhuyn, for the updates!

cpp/src/neighbors/detail/cagra/cagra_helpers.cpp
tfeher: The `attach_dataset_on_build` flag only controls whether the [dataset is attached](https://github.com/NVIDIA/cuvs/blob/7f2e67d61a90947b5f81f9ffed952be0a4343f82/cpp/src/neighbors/detail/cagra/cagra_build.cuh#L2333-L2349) to the index.

The subsequent `update_graph` call would attach the graph unconditionally to the index.

`cagra::build` defines `cagra_graph` as a host matrix https://github.com/NVIDIA/cuvs/blob/7f2e67d61a90947b5f81f9ffed952be0a4343f82/cpp/src/neighbors/detail/cagra/cagra_build.cuh#L2259 

Therefore `update_graph` will always do a graph copy to device.

This means that including the `graph_size` in the GPU mem estimate is correct in the current state of the code.

So you could consider this a follow up request: we need to change this, we do not want to copy the graph to the device when the goal is to create an HNSW index.

cpp/src/neighbors/detail/cagra/cagra_helpers.cpp
```diff
@@ -137,6 +150,9 @@ inline std::pair<size_t, size_t> ivf_pq_build_mem_usage(
   size_t kmeans_pinned_host = 2 * pinned_rows * dim * dtype_size;  // two staging double-buffers
   size_t kmeans_host_mem    = kmeans_indices_host + kmeans_pinned_host;
 
+  // Add graph to index on GPU
+  size_t create_index_gpu_mem = n_rows * graph_degree * sizeof(uint32_t);
+
```
huuanhhuyn, line 155: Add a comment explaining why we intentionally ignore attached dataset size (fallback path when insufficient memory)

tfeher (approved):
LGTM, thanks Huy!

Issue #2100, opened 2026-05-16
[CI] CLUSTER_TEST kmeans flaky


aamijar:
Observing a flaky test failure:
https://github.com/rapidsai/cuvs/actions/runs/25956131520/job/76302995776#step:12:1083

https://github.com/rapidsai/cuvs/actions/runs/25984850719/job/76380196382#step:12:1060

aamijar:
I was able to reproduce this failure with the following

# Environment
RTX6000 pro
CUDA 13.2

# Steps
```
./CLUSTER_TEST --gtest_filter='KmeansFitBatchedTests/KmeansFitBatchedTestF.Result/4' --gtest_repeat=50 --gtest_break_on_failure
```

tarang-jain:
This has been fixed. Closing this one!

aamijar:
Hi @tarang-jain, I think we may still have this issue (5 days ago main nightlies):

https://github.com/NVIDIA/cuvs/actions/runs/35426896363/job/105854344135#step:13:3521

PR #2256, opened 2026-06-23
[CI] Update Convergence Criteria to Fix Flaky KMeans Test

tarang-jain:
Addresses the KMeans flaky test in https://github.com/rapidsai/cuvs/issues/2100. Since we bit match centroids, we limit the number of iterations to ensure that both the OOC and in-memory run the same number of iters. The flaky test is due to numerical differences arising out of atomic operations (perhaps due to different ordering or adding several batch sums in the OOC case versus a single atomic op in the in-memory case) -- `raft::reduce_rows_by_key`. 

UPDATES 08/25/2026:
Rather than clamping the number of iters, we fix the convergence criteria.
rtxpro6000:
main: 9/50 runs were failing.
After this change: 100/100 runs pass.


cpp/tests/cluster/kmeans.cu
```diff
@@ -381,6 +381,9 @@ class KmeansFitBatchedTest : public ::testing::TestWithParam<KmeansBatchedInputs
     params.tol                 = testparams.tol;
     params.rng_state.seed      = 1;
     params.oversampling_factor = 0;
+    // Limit the number of iterations to ensure same number of iterations for reference and batched
+    // code paths.
+    params.max_iter = 3;
```
cjnolet: This fix doesn't seem particularly robust, and appears to me like it'll reduce the coverage for the tests, right? Default max_iters for k-means is much higher. How do we test in a way where we can validate its ability to stop early without having to reduce the max-iters to the point where we are essentially not testing the early stopping at all?

tarang-jain: I think the point of the test when we are bit matching centroids is to ensure that the accumulation of partial sums in the OOC setting in a single iteration matches the in-memory setting. The test fails when there is an edge case -- the variation due to floating point precision in the arithmetic is just enough to push the difference in centroids between consecutive iterations to fall below the convergence criteria -- in-memory converges one iteration sooner in than OOC.

cjnolet (approved):

PR #2257, opened 2026-06-23
Add KvikIO to Improve I/O Throughput

julianmi:
Motivation:
- Performance: CAGRA-ACE uses intermediate files when the dataset, graph, and intermediate objects do not fit into memory. Using direct file I/O and GPUDirect Storage (GDS) improves I/O throughput. I observed a 1.17x end-to-end speedup for BIGANN-1B on L40S.
- Reduce memory pressure: IVF-PQ uses pinned memory in k-means through `raft::sample_rows` in each partition. ACE uses large files leading to high page-cache pressure and memory fragmentation. Direct file I/O skips the page cache and resolves the potentially failing pinned memory allocations.
- Disk-heavy serialize and deserialize routines are affected similarly.

Changes:
- Add KvikIO-backed bulk I/O for ACE/HNSW disk paths.
- Use KvikIO futures for concurrent reads and explicit async result handling.
- Align ACE `.npy` data regions for direct I/O and GDS-friendly transfers.
- Harden file I/O helpers with better validation, ownership, and short-read/write checks.
- Support for GDS for CAGRA and IVF-PQ serialize and deserialize routines. The `std::istream`/`std::ostream` APIs are unchanged.

Package impact: 
- This PR adds one direct runtime dependency: KvikIO (`libkvikio` for `conda` and `libkvikio-cu12`/`libkvikio-cu13` for wheels).
- KvikIO is approximately 0.5 MiB compressed on conda and 2.6-2.8 MiB as a wheel. Wheel installations add `cuda-pathfinder` and `cuFile`, totaling approximately 3.7-4.2 MiB of additional downloads.
- A clean x86_64 `conda` solve adds 14 packages totaling approximately 16.1 MiB compressed, primarily ICU (12.1 MiB) through KvikIO’s `libcurl` dependency chain.
- Existing RAFT/RMM and `rapids-logger` packages are reused.


conda/recipes/libcuvs/recipe.yaml
```diff
@@ -75,6 +75,12 @@ cache:
       - libnvjitlink-dev
       - librmm =${{ minor_version }}
       - libraft-headers =${{ minor_version }}
+      - libkvikio =${{ minor_version }}
+      # libcufile (GPUDirect Storage backend that kvikio compiles against): CUDA 13 on all arches,
+      # CUDA 12 on x86_64 only.
+      - if: cuda_major == "13" or linux64
+        then:
+          - libcufile-dev
```
jameslamb: It should be possible to depend on `libcufile` unconditionally. RAPIDS moved its CUDA 12 floor to CUDA 12.2 specifically for that.

`libkvikio` does that:

https://github.com/rapidsai/kvikio/blob/bdb788f45ef191384a294ecef3312ea2db35a2c7/conda/recipes/libkvikio/recipe.yaml#L74

Can you please explain why this limitation was added? If it's leftover from older RAPIDS references, please remove these guards and comments throughout this PR and make this an unconditional dependency.

python/libcuvs/pyproject.toml
```diff
@@ -20,6 +20,7 @@ license = "Apache-2.0"
 requires-python = ">=3.11"
 dependencies = [
     "cuda-toolkit[cublas,curand,cusolver,cusparse,nvrtc]==13.*",
+    "libkvikio==26.8.*,>=0.0.0a0",
```
jameslamb: RAPIDS wheels runtime `dlopen()` their RAPIDS shared-library dependency.

Add `libkvikio` here: https://github.com/rapidsai/cuvs/blob/71306ec8e76f6c104768cfaba301ca2dcba780bd/python/libcuvs/libcuvs/load.py#L38-L44

I don't believe `rmm` or `raft` depend on it or that it depends on either of them so it can be in any order (I think).

python/libcuvs/pyproject.toml
```diff
@@ -81,6 +82,7 @@ regex = "(?P<value>.*)"
 build-backend = "scikit_build_core.build"
 requires = [
     "cmake>=4.0",
+    "libkvikio==26.8.*,>=0.0.0a0",
```
jameslamb: Please update the script we use when cutting new release branches: https://github.com/rapidsai/cuvs/blob/71306ec8e76f6c104768cfaba301ca2dcba780bd/ci/release/update-version.sh#L107

It now needs `libkvikio` entries.

conda/recipes/libcuvs/recipe.yaml
jameslamb: This also may be totally unnecessary for this recipe... I don't see any direct uses of cufile headers in this PR. If this was just to satisfy `libkvikio`'s transitive dependency on cuFile, leave it out... `libkvikio` conda packages already carry a runtime dependency on `libcufile-dev`.

jameslamb (commented):
I left a few small initial comments from a `packaging-codeowners` perspective. In general, not opposed to having `kvikio` be a dependency of `cuvs` if cuVS maintainers are ok with it.

I'd be happy to review again once cuVS maintainers have more thoroughly gone through this and given their opinions.

cpp/src/neighbors/detail/hnsw.hpp
```diff
@@ -813,44 +818,47 @@ void serialize_to_hnswlib_from_disk(raft::resources const& res,
...
-    }
+    // Issue the three reads concurrently through kvikio (its threadpool parallelizes each), then
+    // wait for all to complete.
+    auto graph_future = graph_kv.pread(graph_buf.data_handle(), graph_bytes, graph_offset);
+    auto dataset_future =
+      dataset_kv.pread(dataset_buf.data_handle(), dataset_bytes, dataset_offset);
+    auto label_future         = label_kv.pread(label_buf.data_handle(), label_bytes, label_offset);
+    const size_t graph_read   = graph_future.get();
+    const size_t dataset_read = dataset_future.get();
+    const size_t label_read   = label_future.get();
```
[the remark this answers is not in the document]
julianmi: Done.

cpp/src/util/file_io.cpp
```diff
@@ -22,28 +23,19 @@ void read_large_file(const file_descriptor& fd,
...
+  // kvikio selects GPUDirect Storage (cuFile) for device destinations on a GDS-capable system, and
+  // the POSIX + threadpool backend (with O_DIRECT when available) otherwise. The destination may be
+  // host or device memory; kvikio detects which.
+  kvikio::FileHandle handle(path, "r");
+  const size_t bytes_read = handle.pread(dest_ptr, total_bytes, file_offset).get();
+  RAFT_EXPECTS(bytes_read == total_bytes,
+               "Incomplete read from %s: expected %zu bytes, got %zu",
+               path.c_str(),
+               total_bytes,
+               bytes_read);
```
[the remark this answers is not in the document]
julianmi: Done.

conda/recipes/libcuvs/recipe.yaml
julianmi: Thanks, I've missed that this was fixed in CUDA 12.2. I've also removed the outdated `libcufile-dev` dependency.

python/libcuvs/pyproject.toml
julianmi: Applied. I've added it first in case `rmm` or `raft` might use `kvikio` in the future.

python/libcuvs/pyproject.toml
julianmi: Done.

cpp/src/util/file_io.cpp
```diff
@@ -22,28 +72,20 @@ void read_large_file(const file_descriptor& fd,
...
-                 bytes_read,
-                 file_pos);
-
-    bytes_remaining -= chunk_size;
-    offset += chunk_size;
-  }
+  // kvikio selects GPUDirect Storage (cuFile) for device destinations on a GDS-capable system, and
+  // the POSIX + threadpool backend (with O_DIRECT when available) otherwise. The destination may be
+  // host or device memory; kvikio detects which.
+  auto handle = detail::open_kvikio_file_for_ace_io(path, "r", dest_ptr);
```
huuanhhuyn, line 170: Many thanks Julian for this fix! 

Should we apply your fix to cagra serialization (and perhaps ivf_pq, vamana, scann) where the io transfer is large enough (GB range)? For cagra, it seems we have two places at [serialize](https://github.com/julianmi/cuvs/blob/7bba838b0031c1d89e5f8bceb22de0cbf0451e80/cpp/src/neighbors/detail/cagra/cagra_serialize.cuh#L96) and [serialize_to_hnswlib](https://github.com/julianmi/cuvs/blob/7bba838b0031c1d89e5f8bceb22de0cbf0451e80/cpp/src/neighbors/detail/cagra/cagra_serialize.cuh#L249)

The wiki10M issue happens on the non-ace path. This PR covers currently the ACE.

huuanhhuyn (changes requested):

cpp/include/cuvs/util/file_io.hpp
```diff
@@ -197,78 +252,180 @@ std::pair<file_descriptor, size_t> create_numpy_file(const std::string& path,
...
                       const size_t total_bytes,
                       const uint64_t file_offset);
 
+/**
+ * @brief Sequential std::ostream backed by kvikio.
+ *
+ * Ordinary stream output is staged into a large host buffer and written to disk through kvikio,
+ * which bypasses the page cache via O_DIRECT when supported (and falls back to buffered POSIX
+ * writes otherwise). Device buffers passed to write_device() use GPUDirect Storage when available.
+ * Because this is a std::ostream it can be passed anywhere an ostream is expected (e.g. the
```
huuanhhuyn: Could we have a unit test that verifies the statement "it can be passed anywhere an ostream is expected"? The unit test should test inherited public methods from `std::ostream` e.g. `operator<<()`, `is_open()` , `seekp()` etc. It should test also the construction of `kvikio_ofstream` from an `std::ostream` object.

Alternatively, we could explicitly delete them and enable individual ones when we actually need to.

cpp/src/util/file_io.cpp
julianmi: This patch's main target was CAGRA-ACE given how large the file typically are. I think it makes sense to extend this to the other serializers as well. I've added a patch and am happy to hear other maintainers' opinions on this change.

We might also look at deserialization if we decide on adding KvikIO.

cpp/include/cuvs/util/file_io.hpp
julianmi: Thanks, this is probably a bit too broad of an statement. I clarified it and added tests.

cpp/src/util/file_io.cpp
julianmi: I've added KvikIO support for CAGRA and IVF-PQ deserialization routines.

cpp/src/neighbors/ivf_list.cuh
huuanhhuyn, line 169: I am curious when we still want to have non-GDS (de)serialization. If GDS is always the preferred way, should we remove the non-GDS leftovers?

julianmi: We still need the non-GDS path for the public `std::istream`/`std::ostream` APIs, which support stringstreams, pipes, and custom stream buffers. Filename-based operations use KvikIO with GDS or POSIX I/O fallbacks. I expect the filename-based operations to be the main use case where one wants to deserialize an index on disk.

cpp/src/util/file_io.cpp
huuanhhuyn: Thanks. Do we plan to support it for serialization?

julianmi: I've checked routines that interact with files and added KvikIO support. This patch has become large. Let's see how the review goes. We might need to split this patch.

cpp/CMakeLists.txt
```diff
@@ -1632,6 +1639,7 @@ SECTIONS
               $<$<BOOL:${CUVS_NVTX}>:CUDA::nvtx3>
               $<COMPILE_ONLY:nvidia::cutlass::cutlass>
               $<COMPILE_ONLY:cuco::cuco>
+              $<BUILD_LOCAL_INTERFACE:kvikio::kvikio>
```
dantegd: Should we preserve KvikIO in the installed static target’s transitive link interface? 

cuvs_static now contains references to compiled kvikio::* symbols, but $<BUILD_LOCAL_INTERFACE:...> evaluates to empty when the target is consumed from another build system. Installing libkvikio supplies the DSO, but it does not add it to a downstream executable’s final link line, so consumers of the installed cuvs::cuvs_static target can get unresolved symbols when an affected archive member is selected, right?

cpp/src/util/file_io.cpp
```diff
@@ -22,28 +87,20 @@ void read_large_file(const file_descriptor& fd,
   RAFT_EXPECTS(total_bytes > 0, "Total bytes must be greater than 0");
   RAFT_EXPECTS(dest_ptr != nullptr, "Destination pointer must not be nullptr");
   RAFT_EXPECTS(fd.is_valid(), "File descriptor must be valid");
+  const std::string path = fd.get_path();
+  RAFT_EXPECTS(!path.empty(), "File descriptor must have an associated path for kvikio I/O");
```
dantegd: The inode/device validation looks good for the path replacement race, but should we also retain the previous POSIX pread path for pathless descriptors, and likewise retain pwrite in the write helper?

file_descriptor(int) and these bulk helpers are installed public APIs if I'm not mistaken, and descriptor only callers worked before this PR and the new test at  file_io_test.cpp now checks that both operations throw. That breaks raw external FDs and memfd-style descriptors in a PR labeled non-breaking, or we can make it breaking of course. What do you think?

cpp/include/cuvs/util/file_io.hpp
```diff
@@ -27,40 +28,118 @@
...
+
+  fd_streambuf(const fd_streambuf&)            = delete;
+  fd_streambuf& operator=(const fd_streambuf&) = delete;
+
+  fd_streambuf(fd_streambuf&& other) noexcept
+    : fd_(std::exchange(other.fd_, -1)),
+      buffer_(std::move(other.buffer_)),
+      buffer_size_(std::exchange(other.buffer_size_, 0))
+  {
+    setg(buffer_.get(), buffer_.get(), buffer_.get());
```
dantegd: I think we need to preserve the unread portion of the buffer when moving the stream.  After a buffered read, fd_ has already advanced past the full 8 KiB buffer, so resetting the get area here causes any unread bytes to be skipped. It looks like move assignment has the same issue. Would it make sense to add a test that reads a prefix, moves the stream, and then verifies the remaining contents?

cpp/src/neighbors/detail/cagra/cagra_build.cuh
```diff
@@ -1385,20 +1444,86 @@ auto build_ace(raft::resources const& res, const index_params& params, DatasetVi
...
+                ::cuvs::neighbors::cagra::detail::build_from_device_matrix<T, IdxT>(
+                  res, sub_index_params, sub_dataset_dev->as_dataset_view());
+              used_device_read = true;
+              return sub_index_t{std::in_place_type<device_sub_index_t>, std::move(direct_index)};
+            } catch (const std::bad_alloc&) {
+              RAFT_LOG_WARN(
+                "ACE: partition %lu did not fit in device memory for a direct (GDS) read; "
+                           "falling back to a host read",
+                partition_id);
+            } catch (const raft::logic_error&) {
```
dantegd: Could we narrow this catch, or at least log the original exception?

dantegd (changes requested):
The integration looks well designed, the pr looks great overall.

cpp/CMakeLists.txt
julianmi: Good catch, thanks. `cuvs_static` now retains KvikIO as a link-only dependency, and KvikIO is registered in the static build/install export sets. The shared target continues to keep it private.

cpp/src/util/file_io.cpp
julianmi: Yes, that might be helpful in the future. I've added a POSIX `pread`/`pwrite` path for pathless descriptors. I've also extended the tests.

cpp/include/cuvs/util/file_io.hpp
julianmi: Agreed. Move construction and assignment now transfer the underlying `std::streambuf` get area, preserving unread buffered bytes. Both operations have regression coverage.

cpp/src/neighbors/detail/cagra/cagra_build.cuh
julianmi: The fallback now catches only `std::bad_alloc,` which includes RMM allocation failures, and logs the original exception.

dantegd (approved):
lgtm pending CI

huuanhhuyn (approved):
Thank you @julianmi for the features!

cpp/src/util/file_io.cpp
```diff
@@ -54,28 +184,251 @@ void write_large_file(const file_descriptor& fd,
...
+std::istream& kvikio_file_reader::stream() { return impl_->stream(); }
+
+void kvikio_file_reader::read_device(void* data, size_t size) { impl_->read_device(data, size); }
+
+// std::streambuf that stages output into a large buffer and writes full buffers to disk through
+// kvikio at an increasing file offset. The trailing partial buffer is written on sync()/close().
+class kvikio_ofstream::sbuf : public std::streambuf {
+ public:
+  sbuf(const std::string& path, size_t cap)
+    : path_(path), handle_(path, "w"), buffer_(std::max<size_t>(cap, kNumpyDataAlignment))
```
jamxia155: `handle_` construction can throw on file-open failure and propagate to the ctor of `sbuf`. In such cases, the `if (!of) RAFT_FAIL(...)` guards that are usually present after instantiation an `sbuf` becomes dead code and the user just sees a raw exception.

Recommend using a function try block to catch any exceptions within the `sbuf` ctor, including from the initializer list:
```
  sbuf::sbuf(const std::string& path, size_t cap)
  try : path_(path), handle_(path, "w"), buffer_(std::max<size_t>(cap, kNumpyDataAlignment))
  {
    RAFT_EXPECTS(buffer_.size() <= static_cast<size_t>(std::numeric_limits<int>::max()),
                 "kvikio_ofstream buffer size must fit in std::streambuf::pbump");
    setp(buffer_.data(), buffer_.data() + buffer_.size());
  }
  catch (const std::exception& e) {
    RAFT_FAIL("Cannot open file %s for writing: %s", path.c_str(), e.what());
  }
```
Following this recommendation, the `if (!of) RAFT_FAIL(...)` guards at call sites can then be removed.

cpp/src/neighbors/detail/cagra/cagra_build.cuh
```diff
@@ -688,127 +741,111 @@ void ace_reorder_and_store_dataset(
...
-                 total_partition_size);
-
-  RAFT_EXPECTS(static_cast<size_t>(sub_dataset.extent(0)) == total_partition_size,
-               "sub_dataset rows (%lu) must match total partition size (%lu)",
-               sub_dataset.extent(0),
-               total_partition_size);
-  RAFT_EXPECTS(static_cast<size_t>(sub_dataset.extent(1)) == dataset_dim,
-               "sub_dataset columns (%lu) must match dataset dimensions (%lu)",
-               sub_dataset.extent(1),
-               dataset_dim);
```
jamxia155, line 812: I understand that these checks are not needed at the existing call sites, but future callers might be using `ace_load_partition_dataset_from_disk` in other situations where these conditions are not guaranteed. Should we pass in the sub-dataset extents to preserve these checks?

cpp/src/neighbors/detail/dataset_serialize.hpp
```diff
@@ -261,6 +295,53 @@ auto deserialize_device_dense(raft::resources const& res, std::istream& is)
...
+      reader.read_device(packed.data_handle(), static_cast<std::size_t>(rows) * row_bytes);
+      raft::copy_matrix(storage.data_handle() + first_row * metadata.stride,
+                        metadata.stride,
+                        packed.data_handle(),
+                        metadata.dim,
+                        metadata.dim,
+                        rows,
+                        stream);
+      raft::resource::sync_stream(res);
+    }
```
jamxia155: There is an opportunity here to overlap `read_device` and `copy_matrix` with double-buffering. Is this on a path hot enough for the extra work?

cpp/include/cuvs/util/file_io.hpp
```diff
@@ -27,40 +28,132 @@
...
     setg(buffer_.get(), buffer_.get(), buffer_.get() + n);
     return traits_type::to_int_type(*gptr());
   }
 
+  pos_type seekoff(off_type off, std::ios_base::seekdir dir, std::ios_base::openmode which) override
+  {
+    if (fd_ == -1 || (which & std::ios_base::in) == 0) { return pos_type(off_type(-1)); }
+
+    int whence = SEEK_SET;
+    if (dir == std::ios_base::cur) {
```
jamxia155, line 76: There's a subcase worth considering: if `off == 0`, we should report position without discarding buffered readahead.
```
    if (dir == std::ios_base::cur) {
      if (off == 0) {
        // Pure tellg(): report position without discarding buffered readahead.
        off_t position = ::lseek(fd_, 0, SEEK_CUR);
        if (position == static_cast<off_t>(-1)) { return pos_type(off_type(-1)); }
        return pos_type(position - static_cast<off_t>(egptr() - gptr()));
      }
```

cpp/cmake/thirdparty/get_kvikio.cmake
```diff
@@ -0,0 +1,43 @@
...
+    BUILD_EXPORT_SET cuvs-static-exports
+    INSTALL_EXPORT_SET cuvs-static-exports
+    CPM_ARGS
+    EXCLUDE_FROM_ALL TRUE
+    GIT_REPOSITORY https://github.com/${PKG_FORK}/kvikio.git
+    GIT_TAG ${PKG_PINNED_TAG}
+    GIT_SHALLOW TRUE
+    SOURCE_SUBDIR cpp
+    # Force KvikIO checkout to get public headers.
+    PATCH_COMMAND ""
```
jameslamb: What is the purpose of this? I'm not sure what's meant by "Force KvikIO checkout to get public headers".

Looking through https://docs.rapids.ai/api/rapids-cmake/stable/command/rapids_cpm_find/#command:rapids_cpm_find I see this note:

> _A `PATCH_COMMAND` will always trigger usage of `CPMAddPackage()` instead of `CPMFindPackage()`, unless `BUILD_PATCH_ONLY` is specified. This is true even if the patch command is empty._

Are you trying to use this to force a re-download of cuVS instead of finding it from installed packages? If so, why?

cpp/cmake/thirdparty/get_kvikio.cmake
```diff
@@ -0,0 +1,43 @@
...
+    OPTIONS "KvikIO_BUILD_EXAMPLES OFF" "KvikIO_REMOTE_SUPPORT OFF"
```
jameslamb:
```suggestion
    OPTIONS
      "KvikIO_BUILD_BENCHMARKS OFF"
      "KvikIO_BUILD_EXAMPLES OFF"
      "KvikIO_BUILD_NSYS_PLUGIN OFF"
      "KvikIO_REMOTE_SUPPORT OFF"
```

Looking through https://github.com/rapidsai/kvikio/blob/9af39e27161d680083b8c309b9040f8c944d68ad/cpp/CMakeLists.txt#L50-L55, I think there are more things that could be turned off for this usage, to save build time and reduce cuVS's exposure to changes in kvikio.

python/libcuvs/libcuvs/load.py
```diff
@@ -32,24 +33,35 @@ def _load_wheel_installation(soname: str):
     return None
 
 
-def load_library():
-    """Dynamically load libcuvs.so and its dependencies"""
+def _load_dependency(module_name: str):
```
jameslamb: I'm confused by this refactoring.

Why is this PR doing anything more than just adding `libkvikio` here following the same patterns used for `libraft` and `librmm`?

Keeping these consistent across the RAPIDS libraries is really helpful for minimizing maintenance burden.

If there wasn't a strong reason for this refactoring and it was just personal preference, please put everything back the way it was and just add `libkvikio` where needed.

jameslamb (changes requested):
This is looking ok to me from a packaging perspective.  I left a few comments that I'd like to see addressed before merging, so leaving a blocking review.

Adding a new link in the dependency graph also requires some other followup, I've staged those PRs for you:

* https://github.com/rapidsai/workflows/pull/129
* https://github.com/rapidsai/devcontainers/pull/759

jamxia155 (commented):
Mostly looks good, leaving some minor suggestions.

cpp/cmake/thirdparty/get_kvikio.cmake
julianmi: Yes, I tried to force a full download after I saw missing-header failures in CI ([example](https://github.com/NVIDIA/cuvs/actions/runs/28093789542/job/83178402032)). I've reverted the change and it looks like CI is passing now. I'm not sure what changed though.

cpp/cmake/thirdparty/get_kvikio.cmake
julianmi: Thanks, I've applied these settings.

python/libcuvs/libcuvs/load.py
julianmi: This was another attempt of fixing CI issues using the `kvikio` wheel ([example](https://github.com/NVIDIA/cuvs/actions/runs/29899183288/job/88867820930)). I've reverted the changes and CI passes. Could this have been an issue upstream: https://github.com/rapidsai/kvikio/commit/93606c074f3d863a7052af25afae569f72cb3304?

cpp/src/util/file_io.cpp
julianmi: Good catch, thanks. I've applied this suggestion. It cleans up the code nicely as well..

cpp/src/neighbors/detail/cagra/cagra_build.cuh
julianmi: Done in 29a5edcdc.

cpp/src/neighbors/detail/dataset_serialize.hpp
julianmi: This is a great idea, thanks. I've experimented a bit with a 4 GiB padded payload with five alternating samples. The test machine didn't have GDS support though and falls back to direct I/O. I see a slight but consistent speedup for the deserialization of about 1%. This might be higher with GDS. I've added the double-buffering. Please have another look.

cpp/include/cuvs/util/file_io.hpp
julianmi: Applied in df87fb86f.

cpp/src/neighbors/detail/dataset_serialize.hpp
jamxia155: The solution seems correct and safe. Thanks for putting in the effort for that last 1%!

jamxia155 (approved):

python/libcuvs/libcuvs/load.py
jameslamb: I think it's helpful to include a little plaintext log (since those log links will disappear over time, and to make this findable from search)

```text
/pyenv/versions/3.11.15/lib/python3.11/site-packages/cuda/pathfinder/_dynamic_libs/search_steps.py:66: in raise_not_found
    raise DynamicLibNotFoundError(f'Failure finding "{self.lib_searched_for}": {err}\n{att}')
E   cuda.pathfinder._dynamic_libs.load_dl_common.DynamicLibNotFoundError: Failure finding "libcufile.so": No such file: libcufile.so*, No such file: libcufile.so*
```

yes definitely, this would have been fixed by https://github.com/rapidsai/kvikio/pull/1013

jameslamb (approved):
Build / packaging changes are looking good to me. Thanks for the thorough responses and for working with me on it.

jameslamb:
Seeing so many approvals, `/merge`-ing this at @julianmi 's request.

PR #2258, opened 2026-06-23
KMeans: Reuse Precomputed Norms for Inertia Computation

### LINKED ISSUE #2057: [FEA] cluster_cost can accept X norm
The `cluster_cost` could be modified to accept an optional X_norm to avoid recomputing it. That function also synchronize each time it is called, while it would be better to accumulate cluster_cost on device and transfer+sync once at the end.

tarang-jain:
Dataset norms are precomputed and cached during the kmeans training iterations. We need not recompute them during the final inertia computation.
Closes https://github.com/rapidsai/cuvs/issues/2057
This also prevents the sync stream for every batch in the OOC setting.
This changes the C++ API in the public header, so it is a breaking change.


fern/pages/cpp_api/cpp-api-neighbors-common.md
lowener, line 1: Is this file some leftover from a merge or is it intentional?

lowener (commented):
Looks good, can you just address the rest of the coderabbit comments, some seems pertinent.

fern/pages/cpp_api/cpp-api-neighbors-common.md
tarang-jain: I built the docs with build_docs.sh and this diff was generated. I'll remove this.

cjnolet:
@tarang-jain @lowener fern docs are static now, so if code or code docs are changed, we should expect them to generate new markdown. 

Now, you'll need to be careful to make sure the changes are correct.... but I'd expect to see docs changing as the code changes.

fern/pages/cpp_api/cpp-api-cluster-kmeans.md
```diff
@@ -1059,8 +1066,9 @@ std::optional<raft::device_vector_view<const double, int64_t>> sample_weight = s
 | `handle` | in | `const raft::resources&` | The raft handle |
 | `X` | in | `raft::device_matrix_view<const double, int64_t>` | Training instances to cluster. The data must be in row-major format. [dim = n_samples x n_features] |
 | `centroids` | in | `raft::device_matrix_view<const double, int64_t>` | Cluster centroids. The data must be in row-major format. [dim = n_clusters x n_features] |
-| `cost` | out | `raft::host_scalar_view<double>` | Resulting cluster cost |
```
cjnolet, line 1062: Don't we want to support both host and device here? Why just change this to device?

tarang-jain: The host API seems redundant. It basically calls the public device API and copies to host scalar + syncs the stream.

tarang-jain: I have marked it for deprecation.

viclafargue (commented):
Could direct C++ coverage for the new `X_norm` overload be added here or as a follow-up? At minimum, a test should verify that supplied squared norms produce the same cost as internally computed norms and that an invalid norm length is rejected.

Additionally, could reusing `d_norms`/`h_norm_cache` in the MG final-inertia path be considered as a follow-up? [It currently omits `X_norm`](https://github.com/NVIDIA/cuvs/blob/ab4b72dcf1a785cb86da2bcf4a7ed1c7153746ec/cpp/src/cluster/detail/kmeans_mg.cuh#L553-L558), so norms are recomputed for each batch, although correctness is unaffected.

tarang-jain:
> Could direct C++ coverage for the new X_norm overload be added here

Just added a c++ test.

lowener (approved):

viclafargue (approved):
Thanks! LGTM

tarang-jain:
nsys reveals that inertia computation in quite fast and negligible in comparison to the fused reduction which the primary bottleneck. In fact, the transfer of cached host norms to device might lead to additional overhead for transfers in each iteration, rather than recomputing them on device itself. So the changes in this PR are not justified, closing it.

Issue #2178, opened 2026-06-02
[FEA] Vamana indexing optimizations


cjnolet:

bkarsin:
The optimization progress has been delayed due to bugs introduced and performance benefits not meeting expectations. Thus far I have implemented and tested many optimizations (some more details in https://github.com/rapidsai/cuvs/issues/1757) and performance is between 5-40% faster than the baseline. In addition, I have added fp16 support that gets up to 2x speedup over fp32.  I am currently working on final rounds of optimization to try and get a bit more performance before submitting the PR.

Issue #1757, opened 2026-01-29
[FEA] Optimize Vamana index build kernels


bkarsin:
There are a number of perf optimizations that can improve the GPU Vamana index build in cuVS. This issue tracks these optimizations that may be applied with several PRs to limit scope and maintain stability.  Perf estimates for each optimization are rough and attempt to be conservative.
**Note: speedups are speculative and should not be relied on until optimizations are tested**

**High Impact / Priority Optimizations**

- [ ] Increase blockDim of `GreedySearch` and `RobustPrune` kernels from 32 to 128-256. Block size limits theoretical occupancy to 50%, so expect **1.2x-1.5x** kernel perf improvement.
- [ ] Parallelize all distance computations in `enqueue_all_neighbors` for improved parallelism. Could significantly reduce sync overheads on method that `GreedySearch` spends 30-40% of runtime on. Expect **1.4-1.6x** improvement to `GreedySearch`
- [ ] Reduce syncthreads required during merge loop of `RobustPrune`. Can reduce syncs by 10x during merge loop. `RobustPrune` spends 20% of time here, so estimate **1.3-1.4x** improvement of `RobustPrune`.
- [ ] Replace serial prefix sum execution with CUB parallel prefix sums. Minimal time is spent here, so roughly 5% E2E build time improvement.
- [ ] Optimize the `check_visited` data structure of `RobustPrune` (hash set or bitset). Currently a linear search, estimate a **1.2x** speedup to `GreedySearch`.

**Overall perf gain estimate**
- **2-2.5x** speedup of `GreedySearch`
- **1.5-2.2x** speedup of `RobustPrune`
Depending on the dataset, the time breakdown can vary between these kernels. On narrow vectors, `GreedySearch` dominates, while `RobustPrune` is more costly for wide vectors. 

Expected E2E perf gains: **1.6x-2.4x**

**Additional more challenging / speculative optimizations**

- [ ] Try to improve cache reuse of candidate-candidate distances computed during `RobustPrune`. Shared memory is a major occupancy limiter for these kernels, so this may not be feasible.
- [ ] Reduce synchronizations used by `PriorityQueue` structure by replacing them with warp primitives.

bkarsin:
Many of the optimization tried thus far were less impactful than estimated, and many different directions had to be tried that did not achieve performance speedup. Currently working on a final pass of optimizations and testing before submitted associated PR.

PR #2264, opened 2026-06-25
Vamana build optimization set and fp16 support

bkarsin:
Series of GPU Vamana build performance optimizations that addresses #2178 and #1757. Initial estimates from #1757 were not accurate, so many other optimizations were tried (some abandoned, some successful). This PR includes:

GreedySearch optimizations:
- Multi-warp blocks - increases occupancy of GreedySearch
- Reduce shared memory used per warp
- fp16 query approximation (along with fp16 support)

RobustPrune optimizations:
- Multi-warp block support (block size depends on graph degree)
- Caching accepted candidate vectors during occlusion loop
- Remove redundant merge of candidate lists and avoid syncs

General optimizations:
- Replace prefix_sums kernel with cub variant
- Re-use distances from GreedySearch in RobustPrune kernel
- fp16 support added

Together these optimizations give significant speedups across all configs with minimal recall variance compared to the current baseline. I benchmarked performance across a range of synthetic datasets and two real-world datasets. (**NOTE**: finishing benchmarks and will update tables below once they are all collected).

Synthetic dataset build tests (all 1M vector datasets)
<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:////Users/bkarsin/Library/Group%20Containers/UBF8T346G9.Office/TemporaryItems/msohtmlclip/clip.htm">
<link rel=File-List
href="file:////Users/bkarsin/Library/Group%20Containers/UBF8T346G9.Office/TemporaryItems/msohtmlclip/clip_filelist.xml">

</head>

<body link="#467886" vlink="#96607D">

  |   | RTX pro 6000 | RTX pro 6000 | H100 | H100 | L4 | L4
-- | -- | -- | -- | -- | -- | -- | --
TYPE | Dim/Degree | BASELINE | OPT | BASELINE | OPT | BASELINE | OPT
fp32 | 64/32 | 1.11 | 1.07 | 1.35 | 1.19 | 9.83 | 3.57
fp32 | 768 / 32 | 5.98 | 4.03 | 6.01 | 4.92 | 36.6 | 25.8
fp32 | 960 / 32 | 7.82 | 5.61 | 8.23 | 6.52 | 50.1 | 38.7
fp32 | 64/64 | 4.4 | 4.05 | 6.14 | 5.4 | 16.4 | 14.8
fp32 | 768/64 | 28.8 | 18.2 | 33.7 | 25.3 | 194 | 125
fp32 | 960/64 | 39.4 | 24.9 | 50.7 | 36.5 | 251 | 185
fp16 | 64/32 |   | 1.05 |   |   |   | 3.8
fp16 | 768 / 32 |   | 2.5 |   |   |   | 11.1
fp16 | 960 / 32 |   | 3.13 |   |   |   | 14.1
fp16 | 64/64 |   | 4.01 |   |   |   | 13.6
fp16 | 768/64 |   | 10.4 |   |   |   | 53.8
fp16 | 960/64 |   | 14.2 |   |   |   | 65.6
int8 | 64/32 | 1.06 | 0.966 | 1.69 | 1.74 | 3.5 | 3.35
int8 | 768 / 32 | 2.98 | 2.38 | 4.82 | 3.69 | 11.2 | 9.21
int8 | 960 / 32 | 3.85 | 3 | 6.39 | 4.32 | 15.1 | 12
int8 | 64/64 | 4.15 | 3.61 | 6.36 | 6.05 | 14.4 | 13.3
int8 | 768/64 | 12.9 | 9.74 | 22.4 | 14.9 | 51.7 | 39.7
int8 | 960/64 | 17.1 | 12.5 | 28.8 | 19.9 | 72.1 | 52.1

</body>

</html>

Also tested real-world BIGANN 10M (uint8 128D) and GIST (fp32 960D) datasets:
<html xmlns:v="urn:schemas-microsoft-com:vml"
xmlns:o="urn:schemas-microsoft-com:office:office"
xmlns:x="urn:schemas-microsoft-com:office:excel"
xmlns="http://www.w3.org/TR/REC-html40">

<head>

<meta name=ProgId content=Excel.Sheet>
<meta name=Generator content="Microsoft Excel 15">
<link id=Main-File rel=Main-File
href="file:////Users/bkarsin/Library/Group%20Containers/UBF8T346G9.Office/TemporaryItems/msohtmlclip/clip.htm">
<link rel=File-List
href="file:////Users/bkarsin/Library/Group%20Containers/UBF8T346G9.Office/TemporaryItems/msohtmlclip/clip_filelist.xml">

</head>

<body link="#467886" vlink="#96607D">

  |   | RTX pro 6000 | RTX pro 6000 | H100 | H100 | L4 | L4
-- | -- | -- | -- | -- | -- | -- | --
Dataset | deg / iters | BASELINE | OPT | BASELINE | OPT | BASELINE | OPT
BIGANN | 32 / 1.0 | 12.017 | 10.736 | 13.1152 | 12.58 | 40.3484 | 39.887
BIGANN | 32 / 2.0 | 25.942 | 23.229 | 32.0109 | 30.4 | 90.6138 | 89.994
BIGANN | 64 / 1.0 | 36.642 | 30.688 | 48.1399 | 43.65 | 126.248 | 120.394
BIGANN | 64 / 2.0 | 86.221 | 72.026 | 118.444 | 107.91 | 299.2 | 285.543
GIST (fp32) | 32 / 1.0 | 6.39807 | 4.9804 | 6.773 | 5.51 | 41.48 | 32.3789
GIST (fp32) | 32 / 2.0 | 16.2624 | 12.1026 | 19.0785 | 13.7675 | 103.197 | 76.84
GIST (fp32) | 64 / 1.0 | 22.9393 | 16.74 | 27.2307 | 20.1767 | 150.2 | 109.98
GIST (fp32) | 64 / 2.0 | 60.1051 | 41.452 | 71.1611 | 53.6482 | 380.185 | 262.1
GIST (fp16) | 32 / 1.0 |   | 2.597 |   | 3.80264 |   |  
GIST (fp16) | 32 / 2.0 |   | 6.176 |   | 8.692 |   |  
GIST (fp16) | 64 / 1.0 |   | 9.965 |   | 13.793 |   |  
GIST (fp16) | 64 / 2.0 |   | 23.95 |   | 34.1922 |   |  

</body>

</html>


cpp/src/neighbors/detail/vamana/vamana_build.cuh
```diff
@@ -84,6 +91,26 @@ __global__ void print_queryIds(void* query_list_ptr)
 
 #define KERNEL_TIMING (RAFT_LOG_ACTIVE_LEVEL <= RAPIDS_LOGGER_LOG_LEVEL_DEBUG)
 
+template <typename accT, typename IdxT>
+__global__ void gather_query_sizes(QueryCandidates<IdxT, accT>* query_list,
```
cjnolet: Do you have a sense for how much this is adding to the binary size? @divyegala should be able to explain how to see the deployed metrics for PRs.

jamxia155:
/ok to test [c055c60](https://github.com/NVIDIA/cuvs/pull/2264/commits/c055c606ac49d846ade8bbe9b38efe7d8c8a2688)

jamxia155:
/ok to test [df13b42](https://github.com/NVIDIA/cuvs/pull/2264/commits/df13b42a0b6c3d18361b87ad5069480d930b0b12)

python/cuvs/cuvs/tests/test_vamana.py
```diff
@@ -20,7 +22,7 @@ def _gen_data(shape, dtype):
     raise AssertionError("unexpected dtype in test helper")
 
 
-@pytest.mark.parametrize("dtype", [np.float32, np.int8, np.uint8])
+@pytest.mark.parametrize("dtype", [np.float32, np.float16, np.int8, np.uint8])
 def test_vamana_build_basic(dtype):
     n_rows, n_cols = 1000, 16
```
tarang-jain, line 27: Could we add at least one test at larger dims? The current new FP16 coverage uses only 16 dimensions, so I think we dont test `greedy_search_use_fp16_query_smem()`?

tarang-jain (changes requested):

python/cuvs/cuvs/tests/test_vamana.py
bkarsin: This is just for the python test. Added fp16 cpp tests (cpp/tests/neighbors/ann_vamana/test_half_uint32_t.cu) that cover a wide range of params.

jamxia155:
/ok to test [350ed89](https://github.com/NVIDIA/cuvs/pull/2264/commits/350ed892bcdeacb63cd08b8acc7da1fc2aa121b4)

jamxia155:
/ok to test [74caba9](https://github.com/NVIDIA/cuvs/pull/2264/commits/74caba996ba89747b6d16de55408e63ce9e7fba7)

jamxia155:
/ok to test [b8238f3](https://github.com/NVIDIA/cuvs/pull/2264/commits/b8238f33f95f1d7d9feb80de9650f1e79f8b4ae6)

cpp/tests/neighbors/ann_vamana.cuh
```diff
@@ -255,6 +255,13 @@ class AnnVamanaTest : public ::testing::TestWithParam<AnnVamanaInputs> {
       raft::random::normal(handle_, r, database.data(), ps.n_rows * ps.dim, DataT(0.1), DataT(2.0));
       raft::random::normal(
         handle_, r, search_queries.data(), ps.n_queries * ps.dim, DataT(0.1), DataT(2.0));
+    } else if constexpr (std::is_same_v<DataT, half>) {
+      // raft::random::normal requires std::is_floating_point (excludes half); use uniform,
+      // which supports half. uniformInt (the integer branch below) rejects half.
+      raft::random::uniform(
+        handle_, r, database.data(), ps.n_rows * ps.dim, DataT(0.1), DataT(2.0));
+      raft::random::uniform(
+        handle_, r, search_queries.data(), ps.n_queries * ps.dim, DataT(0.1), DataT(2.0));
```
jinsolp: I suggest that we just generate normal data and then cast that to fp16 instead of generating fp16 in uniform distribution

cpp/src/neighbors/detail/vamana/greedy_search.cuh
```diff
@@ -96,186 +110,196 @@ __global__ void GreedySearchKernel(
...
+                                             laneId);
         }
       } else {
-        // already visited
         continue;
       }
 
-      num_neighbors = degree;
-      __syncthreads();
+      num_neighbors[warpIdx] = degree;
```
jinsolp, line 271: maybe make only lane 0 write to this?

cpp/src/neighbors/detail/vamana/vamana_structs.cuh
```diff
@@ -226,6 +1008,16 @@ struct QueryCandidates {
...
+  {
+    for (int i = laneId; i < maxSize; i += 32) {
+      ids[i]   = raft::upper_bound<IdxT>();
+      dists[i] = raft::upper_bound<accT>();
+    }
+    if (laneId == 0) { size = 0; }
+  }
+
   // Checks current list to see if a node as previously been visited
   __inline__ __device__ bool check_visited(IdxT target, accT dist)
```
jinsolp: I believe this function is no longer used anywhere anymore. Should we delete it?

cpp/src/neighbors/detail/vamana/vamana_structs.cuh
```diff
@@ -170,16 +205,118 @@ __device__ SUMTYPE l2_ILP4(Point<T, SUMTYPE>* src_vec, Point<T, SUMTYPE>* dst_ve
...
+template <typename SUMTYPE>
+__device__ SUMTYPE l2_ILP4_half(Point<__half, SUMTYPE>* src_vec, Point<__half, SUMTYPE>* dst_vec)
+{
+  __half temp_dst[4] = {
+    __float2half(0.0f), __float2half(0.0f), __float2half(0.0f), __float2half(0.0f)};
+  __half partial_sum[4] = {
+    __float2half(0.0f), __float2half(0.0f), __float2half(0.0f), __float2half(0.0f)};
+  for (int i = threadIdx.x; i < src_vec->Dim; i += 4 * blockDim.x) {
+    temp_dst[0] = dst_vec->coords[i];
+    if (i + 32 < src_vec->Dim) temp_dst[1] = dst_vec->coords[i + 32];
```
jinsolp: Should we use `blockDim.x` instead of hardcoding to 32 here?

tarang-jain: I agree with this. We should not be hardcoding things because the kernel invocation can change (or the same kernel may be called somewhere else in the future).

cpp/src/neighbors/detail/vamana/greedy_search.cuh
```diff
@@ -96,186 +110,196 @@ __global__ void GreedySearchKernel(
   int max_queue_size,
   Node<accT>* topk_pq_mem)
 {
-  int n      = dataset.extent(0);
+  const int warpIdx = threadIdx.x / 32;
```
tarang-jain: here as well. Lets not hardcode 32. Lets use raft::WarpSize instead.

cpp/src/neighbors/detail/vamana/priority_queue.cuh
```diff
@@ -211,40 +211,36 @@ __host__ __device__ bool operator>(const Node<SUMTYPE>& first, const Node<SUMTYP
...
-*/
+// Warp-level version: no __syncthreads, uses laneId for single-thread ops and warp shuffle
 template <typename SUMTYPE>
-__inline__ __device__ void parallel_pq_max_enqueue(Node<SUMTYPE>* pq,
-                                                   int* size,
-                                                   const int pq_size,
-                                                   Node<SUMTYPE> input_data,
-                                                   SUMTYPE* cur_max_val,
-                                                   int* max_idx)
+__inline__ __device__ void parallel_pq_max_enqueue_warp(Node<SUMTYPE>* pq,
```
tarang-jain, line 234: Nit: we have macros to mark kernels: such as `_RAFT_INLINE`

cpp/src/neighbors/detail/vamana/priority_queue.cuh
```diff
@@ -283,34 +275,72 @@ __inline__ __device__ void parallel_pq_max_enqueue(Node<SUMTYPE>* pq,
       }
     }
 
-    if (threadIdx.x == 31) {
+    if (laneId == 31) {
```
tarang-jain: Again I am marking more places where things are hardcoded. We have raft variables to take care of this.

cpp/src/neighbors/detail/vamana/vamana_build.cuh
tarang-jain: I would argue that we can achieve these things with raft linalg primitives (such as raft::linalg::map rather than having to write a new kernel).

cpp/src/neighbors/detail/vamana/vamana_build.cuh
```diff
@@ -84,6 +91,26 @@ __global__ void print_queryIds(void* query_list_ptr)
...
+  for (int i = blockIdx.x * blockDim.x + threadIdx.x; i < count; i += blockDim.x * gridDim.x) {
+    edge_counts[i] = query_list[i].size;
+  }
+}
+
+template <typename accT, typename IdxT>
+__global__ void scatter_prefix_offsets(QueryCandidates<IdxT, accT>* query_list,
+                                       const int* edge_offsets,
+                                       int count)
+{
```
tarang-jain, line 101: same here: we should be able to achieve these with raft primitives.

cpp/src/neighbors/detail/vamana/vamana_structs.cuh
```diff
@@ -31,6 +34,37 @@ namespace cuvs::neighbors::vamana::detail {
...
+static constexpr int kGreedySearchFp16QuerySmemMinDim = 512;
+
+template <typename T>
+__host__ __device__ inline bool greedy_search_use_fp16_query_smem(int dim)
+{
+  return dim >= kGreedySearchFp16QuerySmemMinDim && !is_cuda_fp16_v<T>;
+}
+
+template <typename T>
+__host__ __device__ inline int greedy_search_query_smem_elem_size(int dim)
```
tarang-jain, line 62: Again, lets use our raft macros: RAFT_HOST_DEVICE_INLINE

cpp/src/neighbors/detail/vamana/priority_queue.cuh
tarang-jain: I wouldnt say these macros should block the PR from merging.

cpp/src/neighbors/detail/vamana/vamana_build.cuh
```diff
@@ -416,22 +468,23 @@ void batched_insert_vamana(
...
+                                    cs);
 
     // Get number of unique node destinations
     IdxT unique_dests =
       cuvs::sparse::neighbors::get_n_components(edge_dest.data_handle(), total_edges, stream);
 
     // Find which node IDs have reverse edges and their indices in the reverse edge list
-    thrust::device_vector<IdxT> edge_dest_vec(edge_dest.data_handle(),
-                                              edge_dest.data_handle() + total_edges);
+    RAFT_CUDA_TRY(cudaMemcpyAsync(thrust::raw_pointer_cast(edge_dest_vec.data()),
```
tarang-jain: raft::copy please!

tarang-jain (commented):
There are a lot of changes in this PR where I want to basically tell the exact same things -- using raft macros and raft primtives wherever possible. We should certainly avoid writing small kernels which can easily be a lambda for raft::linalg. It helps keeps things readable and binary size increases can become slightly more predictable. For example, earlier we have found that raft::linalg::map kernels generally add less to the binary size than thrust::for_each (or similar thrust primitives).

cpp/src/neighbors/detail/vamana/robust_prune.cuh
```diff
@@ -112,106 +138,173 @@ __global__ void RobustPruneKernel(
...
+      s_res_size = merged_size;
     }
+    __syncthreads();
 
     // If we need to prune at all...
-    if (res_size > degree) {
-      int accept_count = 0;
+    if (s_res_size > degree) {
+      if (threadIdx.x == 0) s_accept_count = 0;
+      __syncthreads();
```
tarang-jain, line 239: do we need to sync the whole block here? We do it already at line 234. I see that s_accept_count is set to zero and you might need the whole block to see this before moving forward, but how about all threads set that variable rather than just thread zero setting it? Let me know if my understanding is not correct here.

tarang-jain (changes requested):

cpp/src/neighbors/detail/vamana/greedy_search.cuh
```diff
@@ -96,186 +110,196 @@ __global__ void GreedySearchKernel(
...
     bool self_found = false;
-    // Remove self edges
-    for (int j = threadIdx.x; j < query_list[i].size; j += blockDim.x) {
-      if (query_list[i].ids[j] == query_vec->id) {
+    for (int j = laneId; j < query_list[i].size; j += 32) {
+      if (query_list[i].ids[j] == cur_query_id) {
         query_list[i].dists[j] = raft::upper_bound<accT>();
         query_list[i].ids[j]   = raft::upper_bound<IdxT>();
-        self_found             = true;  // Flag to reduce size by 1
+        self_found             = true;  // Flat to reduce size by 1
```
lowener:
```suggestion
        self_found             = true;  // Flag to reduce size by 1
```

cpp/src/neighbors/detail/vamana/vamana_structs.cuh
bkarsin: This is run by a single warp within the block, and uses a warp width stride. I can change it to a define macro if that helps?

cpp/tests/neighbors/ann_vamana.cuh
bkarsin: Fixed

cpp/src/neighbors/detail/vamana/vamana_structs.cuh
bkarsin: Removed

cpp/src/neighbors/detail/vamana/greedy_search.cuh
bkarsin: Replaced all warp width values with raft::WarpSize throughout

cpp/src/neighbors/detail/vamana/vamana_structs.cuh
bkarsin: Replaced with raft::WarpSize

cpp/src/neighbors/detail/vamana/priority_queue.cuh
bkarsin: Replace with raft::WarpSize-1

cpp/src/neighbors/detail/vamana/vamana_build.cuh
bkarsin: Fixed

cpp/src/neighbors/detail/vamana/robust_prune.cuh
bkarsin: All the threads in the block use the shared `s_accept_count` after this, and any thread may increment it. If we have all threads set it to 0 and no sync, we may lose later increments by a trailing thread/warp re-setting it to 0.

cpp/src/neighbors/detail/vamana/priority_queue.cuh
bkarsin: I can change this throughout, but I can't find the docs in raft on the different INLINE macros.

cpp/src/neighbors/detail/vamana/vamana_structs.cuh
bkarsin: Done

cpp/src/neighbors/detail/vamana/vamana_build.cuh
bkarsin: This scatter kernel is not so simple. I could do it but would need to re-work the code a bit more to use a strided mdspan or something. Can work on this if it's a blocker for the PR.

cpp/src/neighbors/detail/vamana/vamana_build.cuh
bkarsin: Done

cpp/src/neighbors/detail/vamana/greedy_search.cuh
bkarsin: Having just lane 0 write it would require a sync after. Seems simpler/cleaner to just have all threads write, since they have the same value.

cpp/src/neighbors/detail/vamana/greedy_search.cuh
```diff
@@ -74,18 +76,31 @@ __global__ void SortPairsKernel(void* query_list_ptr, int num_queries, int topk)
...
 
   Output - the id and dist lists in query_list_ptr will be updated with the nodes visited
            during the GreedySearch.
 **********************************************************************************************/
+// GreedySearchKernel launches 128 threads/block and requests a minimum blocks-per-SM
+// hint for occupancy. On SM 7.2 (Xavier) and SM 7.5 (Turing) the hardware caps at
+// 1024 threads/SM, so 12 * 128 = 1536 exceeds the limit and ptxas emits
+// ".minnctapersm out of range ... will be ignored". Use 8 (8 * 128 = 1024) on those
+// architectures and keep the tuned value of 12 elsewhere (>=1536 threads/SM).
+#if defined(__CUDA_ARCH__) && (__CUDA_ARCH__ == 720 || __CUDA_ARCH__ == 750)
```
tarang-jain, line 98: This macro can be used only in device code, right? For host code we can use `raft::resource::get_device_properties`. Fore example we have this elsewhere:
```
/**
 * @brief Returns true if the fused distance NN implementation should be used.
 *
 * On Ampere (SM <= 8.x) always use fused.
 * On Hopper (SM 9.x) use fused when m or n >= 4096.
 * On Blackwell (SM >= 10.x) use unfused.
 */
template <typename MathT, typename IdxT, typename LabelT>
bool use_fused(const raft::resources& handle, IdxT m, IdxT n, IdxT k)
{
  cudaDeviceProp prop;
  prop = raft::resource::get_device_properties(handle);
  if (prop.major <= 8) {
    // Use fused for Ampere or before
    return true;
  } else if (prop.major == 9 && (m >= 4096 || n >= 4096)) {
    // On Hopper if m, n are bigger than 4096, use fused
    return true;
  } else if (prop.major >= 10) {
    // On Blackwell onwards, use unfused
    return false;
  }
  return false;
}
```

tarang-jain (changes requested):

cpp/src/neighbors/detail/vamana/greedy_search.cuh
bkarsin: This is only used here to define the launch_bounds, since older hardware has a smaller max threads per SM, and using the smaller value on new hardware hurts performance. It's a compile-time parameter so it needs to be a macro, I believe. It looks like the use_fused example determines which kernel to use during runtime, though.

cpp/src/neighbors/detail/vamana/greedy_search.cuh
```diff
@@ -96,186 +111,195 @@ __global__ void GreedySearchKernel(
...
 
-      // computing distances between the query vector and neighbor vectors then enqueue in priority
-      // queue.
-      enqueue_all_neighbors<T, accT, IdxT>(
-        num_neighbors, query_vec, &dataset(0, 0), neighbor_array, heap_queue, dim, metric);
-
-      __syncthreads();
-
-    }  // End cand_q_size != 0 loop
+      enqueue_all_neighbors_warp(num_neighbors[warpIdx],
```
lowener, line 281: I think a syncwarp is needed before reading this value. Threads in the same warp could still be writing with `atomicMin` inside `num_neighbors[warpIdx]`.
Or if `laneId >= degree` some threads will skip the loop.

cpp/src/neighbors/detail/vamana/greedy_search.cuh
tarang-jain: I am saying that the variable `__CUDA_ARCH__` wouldnt be defined in host code.

bkarsin: Ah, I see. I added these macros because I was getting failed builds on 720 and 750 arches due to `launch_bounds__(128,12)` exceeding the limit. The error comes from ptxas, so `__CUDA_ARCH__` fixes it. Is `raft::resource::get_device_properties` visible to ptxas at compile time? If so and you prefer this method, I can change it.

tarang-jain: `raft::resource::get_device_properties` is the way to correctly fetch the CUDA architecture when not inside a device function / kernel.

bkarsin: Ok, done

cpp/src/neighbors/detail/vamana/robust_prune.cuh
```diff
@@ -80,24 +84,46 @@ __global__ void RobustPruneKernel(
...
   s_query.Dim    = dim;
   static __shared__ int prev_edges;
-  static __shared__ accT graphDist;
+  static __shared__ int s_accept_count;
+  static __shared__ int s_do_accept;
+  static __shared__ int s_res_size;
+
+  const int laneId    = threadIdx.x & 31;
+  const int warpId    = threadIdx.x >> 5;
+  const int num_warps = blockDim.x >> 5;
```
tarang-jain, line 117: Lets keep this simple. I have been advised by @tfeher in the past to not use bit arithmetic for simple division / mod operations, since the compiler is typically smart enough to automatically simplify these. So we can do again things like laneId = threadIdx.x % raft::WarpSize and warpId   = threadIdx.x / 32.

cpp/src/neighbors/detail/vamana/vamana_build.cuh
```diff
@@ -214,6 +247,63 @@ void batched_insert_vamana(
...
+  cub::DeviceScan::ExclusiveSum(nullptr,
+                                scan_temp_bytes,
+                                edge_counts.data_handle(),
+                                edge_offsets.data_handle(),
+                                max_batchsize + 1,
+                                cs);
+  auto scan_temp_storage = raft::make_device_mdarray<uint8_t>(
+    res, large_ws, raft::make_extents<int64_t>(std::max(scan_temp_bytes, size_t{1})));
+
+  thrust::device_vector<IdxT> edge_dest_vec(max_total_edges);
```
tarang-jain: Use raft::device_vector here

tarang-jain (approved):
I have not looked into the implementation itself (what the code inside the kernels does) and my review continues to be along the same direction as my previous comments -- to improve overall code quality and readability. Pointing out a couple of very small things.

cpp/src/neighbors/detail/vamana/vamana_build.cuh
cjnolet: We can do this as a follow-up. @bkarsin do you mind creating an issue and referencing it in a comment here just to get this merged?

cpp/src/neighbors/detail/vamana/robust_prune.cuh
tarang-jain: I think this can also be ignored @bkarsin . Its too much nit-picking.

cpp/src/neighbors/detail/vamana/vamana_build.cuh
bkarsin: Created issue #2493 detailing that this kernel should be replaced with RAFT primitives.

cpp/src/neighbors/detail/vamana/vamana_build.cuh
bkarsin: Done

jamxia155:
/ok to test [f0579f9](https://github.com/NVIDIA/cuvs/pull/2264/commits/f0579f989ad25c39c322df2c671ca7880cd70d5f)

python/cuvs/cuvs/neighbors/vamana/vamana.pyx
```diff
@@ -169,13 +169,14 @@ def build(IndexParams index_params, dataset, resources=None):
     struct controls the degree of the final graph.
 
     The following distance metrics are supported:
-        - L2Expanded
+        - L2Expanded (sqeuclidean)
+        - L2SqrtExpanded (l2 / euclidean distance)
```
lowener, line 173: Also update the C doc for `cuvsVamanaBuild` for the `L2SqrtExpanded` distance

cpp/src/neighbors/detail/vamana/vamana_structs.cuh
```diff
@@ -142,44 +177,150 @@ __device__ SUMTYPE l2_ILP4(Point<T, SUMTYPE>* src_vec, Point<T, SUMTYPE>* dst_ve
...
+  __half d = __hsub(s, t);
+  lane_sum = __hfma(d, d, lane_sum);
+}
+
+/* ILP helpers: accumulate (s-t)^2 into acc; operands must already be loaded */
+__device__ __forceinline__ void l2_half_fma_sq(__half& acc, __half s, __half t)
+{
+  __half d = __hsub(s, t);
+  acc      = __hfma(d, d, acc);
+}
```
lowener: No difference between the two function. Can we just define and use one?

cpp/src/neighbors/detail/vamana/vamana_structs.cuh
```diff
@@ -142,44 +177,150 @@ __device__ SUMTYPE l2_ILP4(Point<T, SUMTYPE>* src_vec, Point<T, SUMTYPE>* dst_ve
...
+    if (i + raft::WarpSize < src_vec->Dim) temp_dst[1] = dst_vec->coords[i + raft::WarpSize];
+
+    l2_half_fma_sq(partial_sum[0], src_vec[0].coords[i], temp_dst[0]);
+    if (i + raft::WarpSize < src_vec->Dim)
+      l2_half_fma_sq(partial_sum[1], src_vec[0].coords[i + raft::WarpSize], temp_dst[1]);
+  }
+  partial_sum[0] = __hadd(partial_sum[0], partial_sum[1]);
+
+  return l2_half_warp_reduce<SUMTYPE>(partial_sum[0]);
+}
```
lowener, line 271: Would it make sense to also provide an overload of `l2_half_fma_sq` with half2 type? I see that `__hfma2` and `__hsub2` exist.

lowener: It might benefit `l2_ILP4_half` below too

cpp/src/neighbors/detail/vamana/vamana_structs.cuh
```diff
@@ -197,12 +338,572 @@ __host__ __device__ SUMTYPE l2(const T* src, const T* dest, int dim)
...
+
+template <typename SUMTYPE>
+__device__ __forceinline__ void l2_fma_sq2(SUMTYPE& acc, float sx, float sy, float2 dst2)
+{
+  float dx = sx - dst2.x;
+  float dy = sy - dst2.y;
+  acc      = fmaf(dx, dx, acc);
+  acc      = fmaf(dy, dy, acc);
+}
```
lowener, line 548: This might also benefit from fsub2/fmaf2. The data loaded is already float2

cpp/src/neighbors/detail/vamana/vamana_structs.cuh
```diff
@@ -391,6 +1167,7 @@ __global__ void create_reverse_edge_list(void* query_list_ptr,
 
   for (int i = blockIdx.x * blockDim.x + threadIdx.x; i < num_queries;
        i += blockDim.x * gridDim.x) {
+    int read_idx   = i * query_list[i].maxSize;
```
lowener: unused

cpp/src/neighbors/detail/vamana/greedy_search.cuh
bkarsin: Fixed

cpp/src/neighbors/detail/vamana/greedy_search.cuh
bkarsin: Good catch, added a syncwarp here.

cpp/src/neighbors/detail/vamana/vamana_structs.cuh
bkarsin: Removed the `l2_half_accum` and just use `l2_half_fma_sq` throughout.

cpp/src/neighbors/detail/vamana/vamana_structs.cuh
bkarsin: Good point, though this is not a super simple change. Would have to choose what versions to replace with this (only ILP variants?) and deal with edge cases like odd dimensions. Would also require some re-benchmarking. To avoid making this PR even bigger, would adding at issue for this be better?

cpp/src/neighbors/detail/vamana/vamana_structs.cuh
bkarsin: Removed

cpp/src/neighbors/detail/vamana/vamana_structs.cuh
lowener: Okay then can you create that issue, and add a comment in the code linking to that issue for better tracking please?

bkarsin: Created issue #2510 and added comment in code that references it

cpp/src/neighbors/detail/vamana/vamana_structs.cuh
bkarsin: Created issue #2510 and added comment

python/cuvs/cuvs/neighbors/vamana/vamana.pyx
bkarsin: Done

lowener (approved):
LGTM

jamxia155:
/ok to test [882b781](https://github.com/NVIDIA/cuvs/pull/2264/commits/882b78151510fcefcd21070b778b86309a206261)

bdice (approved):

PR #2278, opened 2026-06-29
Fix `FastIntDiv` tparam

aamijar:
Add explicit template parameters for FastIntDiv, follow up to https://github.com/NVIDIA/raft/pull/3059.


cpp/tests/neighbors/ann_ivf_flat.cuh
```diff
@@ -1,5 +1,5 @@
 /*
- * SPDX-FileCopyrightText: Copyright (c) 2024-2025, NVIDIA CORPORATION.
+ * SPDX-FileCopyrightText: Copyright (c) 2024-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
```
cjnolet, line 2: Why is this adding "& Affiliates"?

aamijar: That is because of https://github.com/NVIDIA/cuvs/commit/627f2d764abe865b3c644c0e13ccebedc5c0f612. And because of this https://github.com/rapidsai/pre-commit-hooks/pull/120.

PR #2282, opened 2026-07-01
ACE: Harden npartitions Check

julianmi:
Improve the `npartitions` handling such that invalid cases are prevented.

| Requested `npartitions` | Behavior |
|--------|--------|
| 0 | Valid auto sentinel. Silently resolves to the existing two-partition starting point, then follows the current small-partition and memory heuristics. |
| 1 | Warns and resolves to 2, then follows the same heuristics. |
| 2 … `n_rows`  | Accepted. The quality heuristic may reduce an excessively high count to avoid tiny partitions. |
| > `n_rows` | Rejected before planning with cannot exceed dataset size. |
| Disk-mode memory adjustment raises count above `n_rows` | Rejected afterward with the “configured memory limit is unsatisfiable” error. |


tfeher (approved):
Thanks Julian, LGTM.

dantegd (approved):
lgtm python-codeowner approval

PR #2284, opened 2026-07-01
Rust: builder-based params, per-module errors

yan-zaretskiy:
Here we further rework the Rust bindings API to be more idiomatic, and also align closer with the underlying C++ namespacing. More specifically:
- Every param type is built with a validating builder instead of `new()` + chained `set_*`. Unset fields keep the C-library defaults; out-of-range values are rejected by `build()`.

```rust
// before
let params = IndexParams::new()?.set_graph_degree(64).set_intermediate_graph_degree(128);
// after
let params = IndexParams::builder().graph_degree(64).intermediate_graph_degree(128).build()?;
```
Thus, by the time the params instance is constructed, it is guaranteed to be valid.
- Resources stream is set at construction. Instead of a post-construction setter, we provide a custom stream via `unsafe { Resources::with_stream(stream) }`. The `Resources::new()` keeps the default stream. 
- Module layout mirrors C++ namespacing. ANN indexes moved under `cuvs::neighbors`.
- The crate-wide `cuvs::Error/cuvs::Result` are removed. Each module returns its own `thiserror` enum, e.g. `CagraError` or `IvfFlatError`. This enables individual index errors to be more granular, instead of creating a single error type encompassing all failure modes.


rust/cuvs/src/distance/mod.rs
```diff
@@ -3,34 +3,165 @@
...
+    /// L2 distance with square root (unexpanded form).
+    L2SqrtUnexpanded,
+    /// Inner product.
+    InnerProduct,
+    /// Chebyshev (L-infinity) distance.
+    Linf,
+    /// Canberra distance.
+    Canberra,
+    /// Generalized Minkowski (Lp) distance with exponent `p`.
+    LpUnexpanded(f32),
```
tgockel, line 42: This isn't a big deal, but might be worth noting -- the exponent `p` is not always honored. Over in the C code, this is only moved when `metric_arg` exists. Might be worth a comment...it's not a regression from the existing C code and I'm not sure how to address it without a major refactor.

rust/cuvs/src/neighbors/ivf_pq/mod.rs
```diff
@@ -0,0 +1,36 @@
...
+//! `AsDlTensor` / `AsDlTensorMut` traits; see the [`dlpack`](crate::dlpack)
+//! module for the tensor model and `examples/cagra.rs` for the same build/search
+//! workflow.
+
+mod index;
+mod params;
+
+pub use index::Index;
+pub use params::{
+    IndexParams, SearchParams, cudaDataType_t, cuvsIvfPqCodebookGen, cuvsIvfPqListLayout,
```
tgockel: You probably don't want to export these native `cudaDataType_t`, `cuvsIvfPqCodebookGen`, and `cuvsIvfPqListLayout` types publicly, since everything else is wrapped.

rust/cuvs/src/neighbors/cagra/params.rs
```diff
@@ -0,0 +1,497 @@
...
+    #[builder]
+    pub fn new(
+        pq_bits: Option<u32>,
+        pq_dim: Option<u32>,
+        vq_n_centers: Option<u32>,
+        kmeans_n_iters: Option<u32>,
+        vq_kmeans_trainset_fraction: Option<f64>,
+        pq_kmeans_trainset_fraction: Option<f64>,
+    ) -> Result<Self, CagraError> {
+        if let Some(bits) = pq_bits
```
tgockel, line 47: This applies to other `new`s in this file, too: Does the C library have a facility to handle validation of these FFI params for you? It might be worth using to avoid duplicating work.

rust/cuvs/src/neighbors/cagra/params.rs
```diff
@@ -0,0 +1,497 @@
...
+        thread_block_size: Option<usize>,
+        hashmap_mode: Option<HashMode>,
+        hashmap_min_bitlen: Option<usize>,
+        hashmap_max_fill_rate: Option<f32>,
+        num_random_samplings: Option<u32>,
+        rand_xor_mask: Option<u64>,
+    ) -> Result<Self, CagraError> {
+        let params = Self::try_new()?;
+
+        let effective_algo = algo.unwrap_or(unsafe { (*params.handle).algo.into() });
```
tgockel: Very minor, but prefer `.unwrap_or_else(|| unsafe { (*params.handle).algo.into() })` so you only deref in the `None` case. `unwrap_or` eagerly evaluates even when `algo` is `Some(_)`.

rust/cuvs/src/ivf_flat/index_params.rs
```diff
@@ -1,106 +0,0 @@
...
-use crate::distance_type::DistanceType;
-use crate::error::{Result, check_cuvs};
-use std::fmt;
-use std::io::{Write, stderr};
-
-pub struct IndexParams(pub ffi::cuvsIvfFlatIndexParams_t);
-
-impl IndexParams {
-    /// Returns a new IndexParams
-    pub fn new() -> Result<IndexParams> {
```
tgockel, line 15: Since this is deleted, the example over in the [README Rust API](https://github.com/NVIDIA/cuvs/blob/main/README.md#rust-api) that uses `IndexParams::new()` won't work any more. For Rust, there's a clever trick you can use to make sure your READMEs always compile:

```rust
#[cfg(doctest)]
#[doc = include_str!("../../README.md")]
pub struct ReadmeDocTests;
```

This picks up the triple-ticks and makes sure they compile (https://doc.rust-lang.org/rustdoc/write-documentation/documentation-tests.html). Rust is the assumed language, but the examples currently all specify `c`/`python`/`rust`/etc -- Rust doctest will ignore non-Rust languages.

There are some other things in `examples/rust` that use this same pattern that fail to compile. CI can catch this for you by running `cargo build --examples`

rust/cuvs/src/distance/mod.rs
yan-zaretskiy: In the C++ implementation the exponent `p` is read only when the `LpUnexpanded` metric is requested, and the only two places that don't reject this metric are `pairwise_distance` and `brute_force`. Both of them do read the value of `p`. So in this sense, the `p` is always honored by algos that support the Lp metric, even though the C API obscures this fact.

rust/cuvs/src/neighbors/cagra/params.rs
yan-zaretskiy: The internal C++ impl can validate *some* of those params. I'd rather not split the validation logic across FFI. It makes for cleaner errors and keeps it in one place. Ideally, I'd drop them all, but then we need to strengthen the C side validation.

tgockel: Understandable.

KyleFromNVIDIA (approved):

cjnolet (approved):

PR #2287, opened 2026-07-02
Revert cagra-hnswlib wrapper refactoring to allow overriding heuristics

achirkin:
Bring back the ability to override CAGRA build parameters via index config for CAGRA-HNSW wrapper.

PR https://github.com/NVIDIA/cuvs/pull/1719 attempted to simplify the wrapper logic by using `hnsw::build` helper.
The helper uses CAGRA-HNSW heuristics to set up optimal CAGRA build parameters based on the passed HNSW parameters. Unfortunately, this makes it impossible to modify individual CAGRA build parameters via the configs, which is necessary for benchmarking the performance of the heuristics themselves.
This PR reverts the relevant files to bring back the old behavior.

### DIFF
--- a/cpp/bench/ann/src/cuvs/cuvs_cagra_hnswlib_wrapper.h
+++ b/cpp/bench/ann/src/cuvs/cuvs_cagra_hnswlib_wrapper.h
@@ -1,5 +1,5 @@
 /*
- * SPDX-FileCopyrightText: Copyright (c) 2023-2026, NVIDIA CORPORATION.
+ * SPDX-FileCopyrightText: Copyright (c) 2023-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
  * SPDX-License-Identifier: Apache-2.0
  */
 #pragma once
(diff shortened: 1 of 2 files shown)


cpp/bench/ann/src/cuvs/cuvs_cagra_hnswlib_wrapper.h
```diff
@@ -85,9 +85,38 @@ void cuvs_cagra_hnswlib<T, IdxT>::build(const T* dataset, size_t nrow)
...
+  bps.cagra_params = [dataset_is_on_host, orig_cagra_params = bps.cagra_params](
+                       auto dataset_extents, auto metric) {
+    auto params                    = orig_cagra_params(dataset_extents, metric);
+    params.attach_dataset_on_build = !dataset_is_on_host;
+    return params;
+  };
+  cuvs_cagra<T, IdxT> cagra_wrapper{this->metric_, this->dim_, bps};
+
+  // build the CAGRA index
+  cagra_wrapper.build(dataset, nrow);
```
tfeher, line 102: Let's assume that the dataset is large and the user only provides `M` and `ef_construction` parameters. We would want the benchmark to fall back to ACE if needed. `hnsw::build()` enables such fallback, but I think `cagra::build()` build does not. Let's look for an solution where the benchmark can also provide such fall-back if the user does not override the build algorithm.

tfeher (changes requested):

PR #2290, opened 2026-07-02
Re-enable Python code coverage

josephine-wolf-oberholtzer:
Contributes to: https://github.com/rapidsai/build-infra/issues/367

We've finally found a `CODECOV_TOKEN` secret value and added it as a repo-scoped secret.

Let's turn code coverage back on for Python tests.

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -452,7 +452,6 @@ jobs:
     if: fromJSON(needs.changed-files.outputs.changed_file_groups).test_python_conda
     with:
       build_type: pull-request
-      run_codecov: false
       script: ci/test_python.sh
   rocky8-clib-standalone-build-matrix:
     needs: checks
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -68,7 +68,6 @@ jobs:
       build_type: ${{ inputs.build_type }}
       branch: ${{ inputs.branch }}
       date: ${{ inputs.date }}
-      run_codecov: false
       script: ci/test_python.sh
       sha: ${{ inputs.sha }}
   conda-java-tests:


.github/workflows/pr.yaml
```diff
@@ -452,7 +452,6 @@ jobs:
     if: fromJSON(needs.changed-files.outputs.changed_file_groups).test_python_conda
     with:
       build_type: pull-request
-      run_codecov: false
```
jameslamb, line 455: Since we're using `secrets: inherit` in this workflow, won't it still end up getting the `CODECOV_TOKEN` secret from the `rapidsai` org (the source of the workflow), referenced here?

https://github.com/rapidsai/shared-workflows/blob/700fd392ec991347a9627f08f27a87b6e4296bff/.github/workflows/conda-python-tests.yaml#L254

And if so won't we still hit these auth issues?

If I'm wrong about that and having an identically-named repo-scoped variable works, then great.

But if I'm right and that's how this works, I think we'll need to update `conda-python-tests.yaml` to accept a codecov token as an input in `secrets:` (see https://github.com/rapidsai/shared-workflows/blob/700fd392ec991347a9627f08f27a87b6e4296bff/.github/workflows/conda-python-tests.yaml#L72).

jameslamb (approved):
Approving so this can be `/merge`'d if it works, but I'm unsure if it'll be enough. Apologies if I've just misunderstood how `secrets: inherit` works.

.github/workflows/pr.yaml
josephine-wolf-oberholtzer: AFAIK, `secrets: inherit` is the totality of secrets this repo sees in the NVIDIA org, not the secrets the shared workflows see in the rapidsai org. If the opposite was true, then the conda uploads would have worked. But I had to punch those secrets into this repo by hand.

I've punched an explicit repo-scoped `CODECOV_TOKEN` into this repo's secrets config.

If the `conda-python-tests` jobs run their codecov steps successfully, then this will have worked.

If not, I'll punch the secrets into the jobs by hand. All good either way just for the knowledge.

josephine-wolf-oberholtzer: Side-note: I really wanna discourage / get rid of `secrets: inherit`.

josephine-wolf-oberholtzer: @jameslamb codecov upload works again 😄 : https://github.com/NVIDIA/cuvs/actions/runs/28605887156/job/84832413989?pr=2290

josephine-wolf-oberholtzer:
We have codecov.io data again:

Screenshot 2026-07-02 at 1 21 50 PM

.github/workflows/pr.yaml
jameslamb: > Side-note: I really wanna discourage / get rid of secrets: inherit.

same

> codecov upload works again 😄

great, ship it!

PR #2291, opened 2026-07-02
Support inner-product distance metric in IVF-RaBitQ

jamxia155:
IVF-RaBitQ previously supported L2 only. This PR adds InnerProduct support across all four search modes (LUT16, LUT32, QUANT4, QUANT8).

Approach: The per-vector RaBitQ factors are derived directly for the inner product rather than reconstructed from a squared-L2 estimate. Starting from `-<o,q> = -<q,c> - <o-c,c> - <o-c,q-c>`, everything that does not depend on the code folds into the per-vector factors at build time, and the query side contributes a single additive term `g_add = -<q,c>`. Concretely, `F_add` becomes `-<o-c,c> + ||o-c||*<o_bar,c>/<o_bar,o>`, and the rescale and error factors lose the factor of 2 that the squared-L2 expansion carries. The estimator kernel itself (`est = f_add + g_add + f_rescale * (ip + kBxSumq)`) is then byte-identical for both metrics.

Kernels emit the negated inner product, so the existing min-selection, block-sort queue, and per-query threshold are reused unchanged; the result is negated after select_k.

Note: an earlier revision of this PR recovered `<q,x>` from the squared-L2 estimate via `<q,x> = (||q||^2 + ||x||^2 - ||q-x||^2)/2`. That is algebraically the same estimator (expanding it cancels `||q||^2` and `||x||^2` against the terms inside `||q-c||^2` and `||o-c||^2`, and the error bounds coincide) but it required storing `||x||^2` per vector and applying a per-candidate correction at search time. The formulation above removes both. Thanks to @Stardust-SJF for pointing at [RaBitQ-Library](https://github.com/VectorDB-NTU/RaBitQ-Library), whose estimator derivation this follows.

Impact: `InnerProduct`'s marginal search-time cost over L2 is now zero: the same per-candidate loads, the same arithmetic, no extra per-vector storage. Index construction is roughly neutral: one FMA per dimension folded into an existing reduction loop, against a dropped full row-norm pass over the dataset.

Details:
- Metric plumbed through `index_params` to `index<IdxT>` to `IVFGPU`; build validates `{L2Expanded, InnerProduct}`. All four search modes support both.
- The three quantizer factor kernels switch on `cuvs::distance::DistanceType`, so adding a metric means adding a case rather than editing a branch. The default arm emits NaN rather than silently falling back to L2, so an unhandled metric fails loudly. The extra `<o-c,c>` reduction is guarded on the metric (uniform across the block) so L2 does not pay for it.
- Kernel params gain `d_g_add`, separating the estimator's additive term from `d_centroid_distances`, which continues to supply `g_error = ||q-c||` for both metrics. For L2 the two alias; for InnerProduct, `d_g_add holds -<q,c>`.
- Probe selection picks clusters by `argmax <q,c>` (mirroring ivf_pq's select_clusters with `alpha=-1`); that same GEMM result is `g_add`, so it is computed once and reused rather than recomputed. k-means clustering stays L2.
- The sign-aware atomic threshold min is selected at compile time via a `Signed` axis, retained only on the four block-sort fragments that update the threshold. `-<q,x>` is negative regardless of formulation, so the int-reinterpret `atomicMin` trick is not monotone there. The three non-block-sort emit fragments are metric-agnostic and carry no `Signed` axis, so L2 codegen is unchanged and total fragment count decreases.
- Serialization gains a leading version field and the metric enum; format version is 2. There is no per-vector norm blob. A v1 index is rejected at load with an explicit version-mismatch error, since its InnerProduct factors would be silently misinterpreted by this estimator.

Testing

- `var_metric()` sweeps all four search modes × k ∈ {10, 128} (block-sort and non-block-sort) × bits_per_dim ∈ {1, 4} (no-ex and with-ex), under a dedicated IvfRabitqInnerProduct instantiation: 48 cases across the three test bodies, all of which round-trip through serialize/deserialize.
- A `signed_data` flag feeds mean-zero data to those cases so inner products take both signs and exercise the sign-aware atomic threshold. Existing L2 cases are unchanged.
- Verified against the previous reconstruction-based implementation on identical data and parameters: 30 of 48 cases are bit-identical, and the remaining 18 differ by at most two neighbors out of 131072 (max Δrecall ~ 1.6e-5). All differences fall in the k=128 non-block-sort path, where top-k boundary ties are resolved differently by the changed arithmetic order.


Stardust-SJF:
Thanks, James, for supporting IP metric in IVF-RaBitQ! I have two minor comments below:

cpp/src/neighbors/ivf_rabitq/gpu_index/ivf_gpu.cu
```diff
@@ -1094,8 +1134,33 @@ void IVFGPU::PrepareClusterSearchInputs(
...
+                                                                  &alpha_ip,
+                                                                  initializer->GetCentroid(0),
+                                                                  num_padded_dim,
+                                                                  queries.data_handle(),
+                                                                  num_padded_dim,
+                                                                  &beta_ip,
+                                                                  searcher.get_g_add(),
+                                                                  num_centroids);
+    select_in = searcher.get_g_add();
+  }
```
Stardust-SJF: I feel that we may not need two GEMMs here. I suppose what we need is q*c^t. For different alphas, there is no need to redo the GEMM. Can we reduce them to one GEMM call?

cpp/src/neighbors/ivf_rabitq/gpu_index/ivf_gpu.cu
```diff
@@ -105,6 +119,16 @@ void IVFGPU::load_transposed(const char* filename)
...
+  // Format version is the first field.
+  std::uint32_t version = 0;
+  read_exact(&version, sizeof(std::uint32_t));
+  RAFT_EXPECTS(version == kSerializationVersion,
+               "ivf_rabitq::deserialize: serialization version mismatch (got %u, expected %u) "
+               "in: %s",
+               version,
+               kSerializationVersion,
+               filename);
+
```
Stardust-SJF, line 133: Border case: if the old version's file's num_vectors's low 32 bits equals to 2, it may pass the version validation.

cpp/src/neighbors/ivf_rabitq/gpu_index/ivf_gpu.cu
jamxia155: Great point, thanks! Refactored into 1 GEMM + 1 element-wise operation.

cpp/src/neighbors/ivf_rabitq/gpu_index/ivf_gpu.cu
jamxia155: This is in line with existing convention of including a light-weight consistency check on the index file, but not a comprehensive, error-proof one.

Stardust-SJF: I see the point, thanks for the explanation. The code looks good to me now.

PR #2297, opened 2026-07-04
Fix IVF extend list resize bug

qwertyforce:
## Description

Fixes a bug in IVF-Flat and IVF-SQ `extend()` where a list could receive new vectors but keep its old logical `list::size`.

The issue happens when extending a list that already has enough allocated capacity. Interleaved IVF list storage may be padded, so the old copy extent can be larger than the old logical size.

Example:

```text
old logical size: 100
old padded copy size: 128
new logical size: 120
```

Previously, `resize_list()` received only the padded value, `128`, and used it both for copying old storage and for deciding whether `list::size` needed to be updated. It checked `120 <= 128`, skipped the update, and left `list::size == 100` even though new vectors were inserted up to `120`.

This stale size can break the shared-list / copy-on-write logic in later extends and corrupt search results.

## Fix

Update the `make_list_extents` in IVF-Flat and IVF-SQ to round up the list extents on creation, so that the extents always carry the "physical" size, then remove the rounding up from other places, so that the `list::size` value is always the "logical" size. This mirrors the current IVF-PQ approach.

## Tests

Added regression tests for:

```text
IVF-Flat repeated extend from a shared base index with enough existing capacity
IVF-SQ in-place extend with enough existing capacity
```


achirkin:
old_used_size is the old logical size. old real size is checked via extends.

qwertyforce:
> old_used_size is the old logical size. old real size is checked via extends.

in ivf-flat extend, resize_list() receives  ```raft::Pow2<kIndexGroupSize>::roundUp(old_list_sizes[label])``` as an argument, right?

qwertyforce:
here is poc
```
// ivf_flat_extend_size_poc.cpp

#include <cuvs/neighbors/ivf_flat.hpp>

#include <raft/core/copy.cuh>
#include <raft/core/device_mdarray.hpp>
#include <raft/core/device_mdspan.hpp>
#include <raft/core/resource/cuda_stream.hpp>

#include <cstdint>
#include <iostream>
#include <numeric>
#include <optional>
#include <vector>

namespace ivf_flat = cuvs::neighbors::ivf_flat;

template <typename IndexT>
uint32_t visible_list_size(raft::resources const& handle, IndexT& index)
{
  auto stream = raft::resource::get_cuda_stream(handle);
  uint32_t h = 0;
  raft::copy(&h, index.list_sizes().data_handle(), 1, stream);
  raft::resource::sync_stream(handle);
  return h;
}

template <typename IndexT>
void print_sizes(char const* tag, raft::resources const& handle, IndexT& index)
{
  std::cout << tag << "\n"
            << "  index.list_sizes()[0] = " << visible_list_size(handle, index) << "\n"
            << "  list::size            = " << index.lists()[0]->get_size() << "\n"
            << "  capacity              = " << index.lists()[0]->indices_capacity() << "\n";
}

int main()
{
  raft::resources handle;
  auto stream = raft::resource::get_cuda_stream(handle);

  constexpr int64_t base_rows = 100;
  constexpr int64_t grow_rows = 20;
  constexpr int64_t rows      = base_rows + grow_rows;
  constexpr int64_t dim       = 4;

  std::vector<float> h_data(rows * dim);
  for (int64_t r = 0; r < rows; ++r) {
    for (int64_t c = 0; c < dim; ++c) {
      h_data[r * dim + c] = static_cast<float>(r + c);
    }
  }

  auto data = raft::make_device_matrix<float, int64_t>(handle, rows, dim);
  raft::copy(data.data_handle(), h_data.data(), h_data.size(), stream);

  ivf_flat::index_params params;
  params.n_lists                        = 1;
  params.metric                         = cuvs::distance::DistanceType::L2Expanded;
  params.add_data_on_build              = false;
  params.kmeans_trainset_fraction       = 1.0;
  params.adaptive_centers               = false;
  params.conservative_memory_allocation = false;

  auto all_view = raft::make_device_matrix_view<const float, int64_t, raft::row_major>(
    data.data_handle(), rows, dim);

  auto empty_index = ivf_flat::build(handle, params, all_view);

  auto base_view = raft::make_device_matrix_view<const float, int64_t, raft::row_major>(
    data.data_handle(), base_rows, dim);

  auto base_index = ivf_flat::extend(handle, base_view, std::nullopt, empty_index);
  raft::resource::sync_stream(handle);

  print_sizes("after base extend", handle, base_index);

  std::vector<int64_t> h_ids(grow_rows);
  std::iota(h_ids.begin(), h_ids.end(), base_rows);

  auto ids = raft::make_device_vector<int64_t, int64_t>(handle, grow_rows);
  raft::copy(ids.data_handle(), h_ids.data(), h_ids.size(), stream);

  auto grow_view = raft::make_device_matrix_view<const float, int64_t, raft::row_major>(
    data.data_handle() + base_rows * dim, grow_rows, dim);

  auto ids_view =
    raft::make_device_vector_view<const int64_t, int64_t>(ids.data_handle(), grow_rows);

  auto grown_index = ivf_flat::extend(
    handle,
    grow_view,
    std::make_optional<raft::device_vector_view<const int64_t, int64_t>>(ids_view),
    base_index);

  raft::resource::sync_stream(handle);

  print_sizes("after grow extend", handle, grown_index);

  auto visible  = visible_list_size(handle, grown_index);
  auto internal = grown_index.lists()[0]->get_size();

  if (visible == rows && internal == base_rows) {
    std::cout << "\nBUG: index-visible size is " << visible
              << " but internal list::size stayed " << internal << "\n";
    return 1;
  }

  std::cout << "\nNo mismatch observed\n";
  return 0;
}
```


after base extend
  index.list_sizes()[0] = 100
  list::size            = 100
  capacity              = 128
  
  
after grow extend
  index.list_sizes()[0] = 120
  list::size            = 100
  capacity              = 128

BUG: index-visible size is 120 but internal list::size stayed 100

achirkin:
Ok, sorry for closing this prematurely, this does seem legit. I'm not sure though if this internal get_size() is used where it could cause any problems. I'll look into this in the following days.

qwertyforce:
Thank you very much! I forgot to mention how i even stumbled  upon this - i got very strange Illegal memory access errors (i am using python wrapper for multiple extends + search + torch for calculating the vectors) and believe that this bug is related (IMA was in calc_chunk_indices_kernel)
But I couldnt distill my production code to a small reproducible poc (the IMA is also non-deterministic), so all i can do is either the one poc i sent, or https://gist.github.com/qwertyforce/65d9bb2d108701240fc9ef1bc952e685 (tldr: if you create 2 new indexes by extend from the base one (not inplace), child_b data can leak into child_a. child_a is not modified after child_b is created, but child_a search results change)

qwertyforce:
off-topic, figured out why i got IMA, https://github.com/NVIDIA/raft/pull/3041, different issue.
should I create an issue/PR in cuVS to speed up the process, or is this comment enough?

achirkin:
Everything that gets merged in the main branch before next burndown, gets automatically in the next release - the new release branch is forked from the main when burndown starts. Here's our schedule https://docs.rapids.ai/maintainers/

cpp/src/neighbors/ivf_list.cuh
```diff
@@ -66,13 +66,17 @@ CUVS_EXPORT void resize_list(raft::resources const& res,
...
+  // old_logical_size is the previous visible size from this index's list_sizes().
+  // old_used_size is the old allocation copy extent and may include padded slots
+  // required by interleaved list layouts.
   bool skip_resize = false;
   if (orig_list) {
     if (new_used_size <= orig_list->indices.extent(0)) {
-      auto shared_list_size = old_used_size;
-      if (new_used_size <= old_used_size ||
+      auto shared_list_size = old_logical_size;
+      if (new_used_size <= old_logical_size ||
```
huuanhhuyn: I guess there are 3 cases:
1. new_used_size <= old_logical_size -> no size update, no reallocation
2. old_logical_size < new_used_size <= old_used_size -> size update, no reallocation
3. new_used_size > old_used_size -> size update, reallocation 

If I understand it correctly, your current version will do both size update and reallocation for the 2nd case.

huuanhhuyn (commented):
Thank you for your contribution! I'd have 1 doubt.

cpp/src/neighbors/ivf_list.cuh
qwertyforce: >If I understand it correctly, your current version will do both size update and reallocation for the 2nd case.

Only if ```orig_list->size.compare_exchange_strong(shared_list_size, new_used_size)``` returns false

it returns false if orig_list->size was already mutated before this resize_list call ( if the current shared list size is not old_logical_size anymore)

cpp/tests/neighbors/ann_ivf_flat/test_float_int64_t.cu
```diff
@@ -19,4 +22,73 @@ TEST_P(AnnIVFFlatTestF_float, AnnIVFFlat)
...
+    data.data_handle() + base_rows * dim, grow_rows, dim);
+  auto grow_indices_view =
+    raft::make_device_vector_view<const int64_t, int64_t>(indices.data_handle(), grow_rows);
+  auto first_grown_index =
+    extend(handle,
+           grow_data_view,
+           std::make_optional<raft::device_vector_view<const int64_t, int64_t>>(grow_indices_view),
+           base_index);
+  raft::resource::sync_stream(handle);
+
```
huuanhhuyn, line 74:
```suggestion
  ASSERT_EQ(first_grown_index.lists()[0].get(), base_index.lists()[0].get());
```
I'd suggest to test that the first grown list reuses the base list without reallocation.

cpp/tests/neighbors/ann_ivf_flat/test_float_int64_t.cu
```diff
@@ -19,4 +22,73 @@ TEST_P(AnnIVFFlatTestF_float, AnnIVFFlat)
...
+    raft::make_device_matrix_view<const float, int64_t>(data.data_handle(), rows, dim);
+  auto empty_index = build(handle, params, all_data_view);
+
+  auto base_data_view =
+    raft::make_device_matrix_view<const float, int64_t>(data.data_handle(), base_rows, dim);
+  auto base_index = extend(handle, base_data_view, std::nullopt, empty_index);
+  raft::resource::sync_stream(handle);
+
+  ASSERT_EQ(base_index.lists()[0]->get_size(), base_rows);
+  ASSERT_GE(base_index.lists()[0]->indices_capacity(), rows);
```
huuanhhuyn:
```suggestion
  ASSERT_EQ(base_index.lists()[0]->indices_capacity(), raft::Pow2<kIndexGroupSize>::roundUp(rows));
```
I think this makes the test more expressive

cpp/tests/neighbors/ann_ivf_flat/test_float_int64_t.cu
```diff
@@ -19,4 +22,73 @@ TEST_P(AnnIVFFlatTestF_float, AnnIVFFlat)
...
+
+  std::vector<int64_t> host_indices(grow_rows);
+  std::iota(host_indices.begin(), host_indices.end(), base_rows);
+  auto indices = raft::make_device_vector<int64_t, int64_t>(handle, grow_rows);
+  raft::copy(indices.data_handle(), host_indices.data(), host_indices.size(), stream);
```
huuanhhuyn:
```suggestion
  auto indices = raft::make_device_vector<int64_t, int64_t>(handle, grow_rows);
  raft::linalg::range(indices.data_handle(), base_rows, rows, stream);
```
raft has a device built-in method for it.
We should `#include <raft/linalg/init.cuh>` and remove the `include <numeric>`

cpp/src/neighbors/ivf_list.cuh
huuanhhuyn: That's right. I misunderstood the logics.

cpp/tests/neighbors/ann_ivf_sq/test_float_int64_t.cu
```diff
@@ -1,17 +1,77 @@
...
+#include <vector>
+
 namespace cuvs::neighbors::ivf_sq {
 
 typedef AnnIVFSQTest<float, float, int64_t> AnnIVFSQTestF_float;
 TEST_P(AnnIVFSQTestF_float, AnnIVFSQ) { this->testAll(); }
 
 INSTANTIATE_TEST_CASE_P(AnnIVFSQTest, AnnIVFSQTestF_float, ::testing::ValuesIn(inputs));
 
+TEST(AnnIVFSQTest, ExtendInPlaceUpdatesListSizeWithinCapacity)
```
huuanhhuyn, line 23: I'd have the same reviews for this test.

cpp/tests/neighbors/ann_ivf_flat/test_float_int64_t.cu
```diff
@@ -19,4 +22,73 @@ TEST_P(AnnIVFFlatTestF_float, AnnIVFFlat)
...
+
+  std::vector<float> host_data(rows * dim);
+  for (int64_t row = 0; row < rows; row++) {
+    for (int64_t col = 0; col < dim; col++) {
+      host_data[row * dim + col] = static_cast<float>(row + col);
+    }
+  }
+
+  auto data = raft::make_device_matrix<float, int64_t>(handle, rows, dim);
+  raft::copy(data.data_handle(), host_data.data(), host_data.size(), stream);
```
huuanhhuyn:
```suggestion
  auto data = raft::make_device_matrix<float, int64_t>(handle, rows, dim);
  raft::matrix::fill(handle, data.view(), 0.0f);
```
I think we could initialize the data directly on device with zeros. We test the sizes instead of filled values. I assume the `build()` works fine with zeros.

After the changes, we should `#include <raft/matrix/init.cuh>` and remove `#include <vector>`

huuanhhuyn (changes requested):

huuanhhuyn:
Some suggestions to improve the tests. Otherwise LGTM! Thank you for the contribution!

qwertyforce:
fixed the tests

cpp/tests/neighbors/ann_ivf_sq/test_float_int64_t.cu
```diff
@@ -54,12 +50,11 @@ TEST(AnnIVFSQTest, ExtendInPlaceUpdatesListSizeWithinCapacity)
   raft::resource::sync_stream(handle);
 
   ASSERT_EQ(index.lists()[0]->get_size(), base_rows);
-  ASSERT_GE(index.lists()[0]->indices_capacity(), rows);
+  ASSERT_EQ(index.lists()[0]->indices_capacity(), raft::Pow2<kIndexGroupSize>::roundUp(rows));
+  auto* base_list = index.lists()[0].get();
```
huuanhhuyn, line 54: Why do we need to get the base_list pointer here? Can't we later get it at line 69?

huuanhhuyn:
@achirkin I can't trigger the full 85 CI checks. Can you trigger it or it is because of missing commit signing?

cpp/tests/neighbors/ann_ivf_sq/test_float_int64_t.cu
qwertyforce: We need this pointer there so we can check afterward that the extend didnt reallocate the list

also, its a little bit weird, but ivf-flat exposes both inplace and non-inplace extend, while ivf-sq exposes only inplace version
and in test that is in ivf-flat folder, we test non-inplace variant of extend and in ivf-sq we test inplace version

achirkin (dismissed):
I see the issue is real, but I still have the doubts about the chosen approach to fix it. ivf-list shouldn't be aware of what is "logical" size. I think a more appropriate fix for ivf-flat and ivf-sq would be to pad the extents to account for the interleaved storage. Please have a look at my draft https://github.com/qwertyforce/cuvs/pull/1

tfeher (approved):
Thanks @qwertyforce for the fix LGTM.

huuanhhuyn (approved):
Thanks @qwertyforce and @achirkin!

PR #2303, opened 2026-07-07
Fix filtered CAGRA search returning incorrect recall due to kHostMmap

rupakroynv:
## Description

Filtered CAGRA search returns incorrect recall when the benchmark harness passes the filter bitset using MemoryType::kHostMmap. This affects CUVS_CAGRA_ANN_BENCH when running filtered search workloads with a pre-computed filter bitset file.

### Root Cause

cuvs_cagra::get_preference() returned MemoryType::kHostMmap for both the dataset and the filter bitset. kHostMmap allocates a file-backed MAP_PRIVATE mmap where pages are not pre-faulted and  physical memory mappings.

Inside CAGRA's search dispatch (cagra.cuh), bitset_view::count() is called to estimate the filtering rate. That function wraps the bitset pointer in raft::make_device_vector_view and passes it to raft::popc, which launches a CUDA kernel that reads from the pointer assuming it is device memory. A host mmap pointer is not valid device memory. So, the kernel reads garbage or zeros, producing a wrong filtering_rate and causing CAGRA to return incorrect results that caused recall collapse to approximately the filter pass rate (~5%), regardless of itopk.

A secondary issue: the benchmark harness had no support for loading a pre-computed filter bitset from a file (filter_bitset_file in the JSON config). Only filtering_rate (random bitset generation) was supported. The filter_bitset_file field was silently ignored, causing filter_bitset() to return nullptr and CAGRA to run completely unfiltered searches.

### Fix

1. Decouple filter memory type from dataset memory type (ann_types.hpp, benchmark.hpp, cuvs_cagra_wrapper.h)

Add std::optional<MemoryType> filter_memory_type to algo_property so the filter bitset memory type can be set independently of the dataset. cuvs_cagra sets filter_memory_type = kDevice while keeping dataset_memory_type = kHostMmap. This preserves support for datasets larger than GPU memory while ensuring the filter bitset is explicitly cudaMemcpy'd to device memory before any GPU kernel accesses it. The filter_memory_type field defaults to nullopt (falls back to dataset_memory_type) so all other algorithms are unaffected.

2. Add filter_bitset_file support (conf.hpp, dataset.hpp, benchmark.hpp)

Parse filter_bitset_file from the JSON dataset block and load it into the dataset as a blob<bitset_carrier_type>. Previously only filtering_rate (random bitset) was supported; pre-computed filter files were silently ignored.

3. Skip redundant copy_with_padding() (cuvs_cagra_wrapper.h)

set_search_param() unconditionally called copy_with_padding(), which allocates a second full device copy of the dataset.   Skip copy_with_padding() when the data is already on device and no 16-byte alignment padding is needed.

4. Fix copy() crash (cuvs_cagra_wrapper.h)

sub_indices_ can be left in a corrupted state when the dataset update code path runs. For single-index CAGRA it is always logically empty, so reset it via placement new before invoking the copy constructor to prevent bad_array_new_length.

### Notes

- These changes apply to the ANN benchmark harness only; the CAGRA library API is unchanged.
- No existing unit tests cover filtered search through the benchmark harness memory-type path; this fix was validated via direct benchmark runs.

### DIFF
--- a/cpp/bench/ann/src/common/ann_types.hpp
+++ b/cpp/bench/ann/src/common/ann_types.hpp
@@ -9,6 +9,7 @@
 
 #include <memory>
 #include <nlohmann/json.hpp>
+#include <optional>
 #include <stdexcept>
 #include <string>
 #include <vector>
@@ -78,6 +79,8 @@ struct algo_property {
   MemoryType dataset_memory_type;
   // neighbors/distances should have same memory type as queries
   MemoryType query_memory_type;
+  // filter bitset memory type; defaults to dataset_memory_type if not explicitly set
+  std::optional<MemoryType> filter_memory_type{std::nullopt};
 };
 
 class algo_base {
--- a/cpp/bench/ann/src/common/benchmark.hpp
+++ b/cpp/bench/ann/src/common/benchmark.hpp
@@ -102,6 +102,9 @@ inline auto parse_algo_property(algo_property prop, const nlohmann::json& conf) 
   if (conf.contains("query_memory_type")) {
     prop.query_memory_type = parse_memory_type(conf.at("query_memory_type"));
   }
+  if (conf.contains("filter_memory_type")) {
+    prop.filter_memory_type = parse_memory_type(conf.at("filter_memory_type"));
+  }
   return prop;
 };
 
@@ -263,7 +266,8 @@ void bench_search(::benchmark::State& state,
     }
     try {
       a->set_search_param(*search_param,
-                          dataset->filter_bitset(current_algo_props->dataset_memory_type));
+                          dataset->filter_bitset(current_algo_props->filter_memory_type.value_or(
+                            current_algo_props->dataset_memory_type)));
     } catch (const std::exception& ex) {
       state.SkipWithError("An error occurred setting search parameters: " + std::string(ex.what()));
       return;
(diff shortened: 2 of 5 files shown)


jamxia155:
The benchmark was using `property.dataset_memory_type = MemoryType::kHostMmap` which allowed datasets that cannot fit in GPU memory, as long as they can fit on host. We lose that capability with the current change to `kDevice`. An improvement would be to de-couple the memory type of the dataset from the filter, e.g. `property.dataset_memory_type = MemoryType::kHostMmap` and a new `property.filter_memory_type = MemoryType::kDevice`.

The PR description suggests that the issue was incorrect results without ATS or performance degradation with ATS. Am I remembering things right that you were using a GH200 system that was ATS-enabled, but the issue you ran into was with correctness?

Aside: for formatting, please run `pre-commit run --files cpp/bench/ann/src/cuvs/cuvs_cagra_wrapper.h`.

rupakroynv:
@jamxia155 
>The benchmark was using property.dataset_memory_type = MemoryType::kHostMmap which allowed datasets that cannot fit in GPU memory, as long as they can fit on host. We lose that capability with the current change to kDevice. An improvement would be to de-couple the memory type of the dataset from the filter, e.g. property.dataset_memory_type = MemoryType::kHostMmap and a new property.filter_memory_type = MemoryType::kDevice.
>
Done
>The PR description suggests that the issue was incorrect results without ATS or performance degradation with ATS. Am I remembering things right that you were using a GH200 system that was ATS-enabled, but the issue you ran into was with correctness?
>
Updated the PR description. Its is a correctness issue with or without ATS support.
>Aside: for formatting, please run pre-commit run --files cpp/bench/ann/src/cuvs/cuvs_cagra_wrapper.h.
>
Done

PR #2305, opened 2026-07-08
wheels: build CUDA 13 wheels with latest CTK (13.3.0)

jameslamb:
Contributes to https://github.com/rapidsai/build-planning/issues/268

Part of a RAPIDS-wide initiative to build wheels using v13.3.0 of the CUDA Toolkit, and to generally get RAPIDS back to the pattern of always building against its latest supported CTK.

As of https://github.com/rapidsai/shared-workflows/pull/590, RAPIDS builds CUDA 13 wheels with v13.3 of the CTK. This updates the project's `nvidia-nvjitlink` floor to match that.


jameslamb:
2 wheel test jobs timed out after 6 hours, though one was a CUDA 12.9 job so I suspect it's not related to this PR:

https://github.com/NVIDIA/cuvs/actions/runs/28916932499/job/85792220262?pr=2305

bdice (approved):

jameslamb:
If we get to the point where only `conda-cpp-tests` (which can take 2+ hours) is still running and everything else has passed, I'll admin-merge this.

All CI will run here because I touched `dependencies.yaml`, but these changes have 0 effect on anything outside of wheels.

And I want this merged + new packages built on `main` before merging https://github.com/rapidsai/cugraph/pull/5575, to hopefully get out new cuGraph builds built with new-enough RAFT and cuVS ... in `cugraph-gnn`, we are still dealing with the fallout from the RAFT ABI break a few days ago, I think because a new-enough build of that entire combination hasn't yet been published: https://github.com/rapidsai/cugraph-gnn/pull/487

jameslamb:
Only conda-based CI is left here, and this change wouldn't affect it at all. Cancelling CI to save resources, admin-merging to keep the migration moving.

PR #2307, opened 2026-07-08
Add missing HNSW bindings

julianmi:
Add the missing HNSW bindings and cleans up the changes in https://github.com/NVIDIA/cuvs/pull/1719.

Implements feature https://github.com/NVIDIA/cuvs/issues/2156. More work might be needed to better fit the recall curve provided by HNSW. See https://github.com/NVIDIA/cuvs/pull/2031. CC @achirkin


java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/HnswIndexImpl.java
```diff
@@ -97,6 +118,13 @@ public void close() {
    */
   @Override
   public SearchResults search(HnswQuery query) throws Throwable {
+    synchronized (lifecycleLock) {
```
imotov: If I understand the intent correctly, this synchronized block was added to ensure a clean shutdown. However, it also appears to serialize all searches, preventing concurrent execution. If we want to enforce shutdown semantics here, a read-write lock may be a better fit. That said, at least in Lucene this concern is usually delegated to the caller. The caller typically tracks in-flight requests via reference counting (or a similar mechanism) and ensures that shutdown does not proceed until all outstanding requests have completed. So adding it here might duplicate an effort in some systems.

julianmi: Thank you and sorry that I missed this comment. I ran into an issues with a test that I've added and added the synchronization. I've reverted this change.

dantegd (approved):
lgtm, but it'd be a good idea to also update fern/pages/cuvs_bench/param_tuning.md?. The cuvs_hnsw section currently describes builds as always using ACE, but it would be good to reflect that the C++ HNSW build path automatically uses an in-memory CAGRA build when sufficient memory is available and falls back to ACE otherwise. This PR exposes that existing behavior through the bindings, so it would be good for the tuning guide to reflect it. What do you think?

julianmi:
> lgtm, but it'd be a good idea to also update fern/pages/cuvs_bench/param_tuning.md?. The cuvs_hnsw section currently describes builds as always using ACE, but it would be good to reflect that the C++ HNSW build path automatically uses an in-memory CAGRA build when sufficient memory is available and falls back to ACE otherwise. This PR exposes that existing behavior through the bindings, so it would be good for the tuning guide to reflect it. What do you think?

Good catch. I updated the `cuvs_hnsw` tuning section to describe the automatic in-memory CAGRA/ACE selection and clarified that the ACE tuning parameters apply only when ACE is selected.

imotov (approved):

mythrocks (approved):
Approving. @imotov has reviewed this.

We need to get @imotov on the code-owners list.

tfeher (approved):
LGTM

PR #2311, opened 2026-07-09
Use librtcx for JIT-LTO

arhag23:
Removes reusable JIT-LTO logic as this is now maintained in librtcx.


cpp/src/neighbors/ivf_sq/detail/jit_lto_kernels/scan_planner.hpp
```diff
@@ -1,11 +1,11 @@
 /*
- * SPDX-FileCopyrightText: Copyright (c) 2025-2026, NVIDIA CORPORATION.
+ * SPDX-FileCopyrightText: Copyright (c) 2025-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
  * SPDX-License-Identifier: Apache-2.0
  */
 
 #pragma once
 
-#include <cuvs/detail/jit_lto/AlgorithmPlanner.hpp>
+#include <AlgorithmPlanner.hpp>
```
bdice: We need these in some kind of subdirectory, not on the top level include path.

bdice (changes requested):

cpp/src/neighbors/ivf_sq/detail/jit_lto_kernels/scan_planner.hpp
arhag23: If this looks okay https://github.com/rapidsai/librtcx/pull/13 then we can merge that and fix the include paths in this PR.

conda/recipes/libcuvs/recipe.yaml
```diff
@@ -256,6 +257,7 @@ outputs:
         - libcurand
         - libcusolver
         - libcusparse
+        - zstd-static =1.5.7
```
divyegala, line 260: Does rtcx always pull this? It should be part of `dependencies.yaml` then, no? If it's a build time dependency, it will be needed for local and wheel builds also.

arhag23: For build, rtcx will just grab it from source and build it. Since rtcx is a static library, consumers also need to have `libzstd_static` as well but when building from source CMake takes care of this. With cuVS also providing a static library, users of that will need `libzstd_static` as well. For cuDF this issue is invisible since they do the same thing with libzstd but only provide a shared library.

cpp/src/neighbors/ivf_sq/detail/jit_lto_kernels/scan_planner.hpp
arhag23: @bdice do the style changes in that PR look okay for you? The header will then end up under the `rtcx` subdirectory.

divyegala (approved):

bdice (approved):

PR #2313, opened 2026-07-10
All-neighbors graph on host memory

### LINKED ISSUE #1903: [FEA] Improve batched `all-neighbors` given CPU indices/distances

jinsolp:
Closes https://github.com/NVIDIA/cuvs/issues/1903

This PR adds support for CPU kNN graph for batched all-neighbors. Passing host matrices for the indices and distances avoids having to fit the entire graph on GPU. Small perf gap and expected peak GPU memory reductions below:

results

### DIFF
--- a/c/include/cuvs/neighbors/all_neighbors.h
+++ b/c/include/cuvs/neighbors/all_neighbors.h
@@ -1,5 +1,5 @@
 /*
- * SPDX-FileCopyrightText: Copyright (c) 2025-2026, NVIDIA CORPORATION.
+ * SPDX-FileCopyrightText: Copyright (c) 2025-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
  * SPDX-License-Identifier: Apache-2.0
  */
 
--- a/c/include/cuvs/neighbors/all_neighbors.h
+++ b/c/include/cuvs/neighbors/all_neighbors.h
@@ -28,8 +28,10 @@ extern "C" {
  * provide the dataset on host.
  *
  * Notes:
- * - Outputs (indices, distances, core_distances) are expected to be on device memory.
- * - Host variant accepts host-resident dataset; device variant accepts device-resident dataset.
+ * - A host-resident dataset accepts either host- or device-resident outputs (indices, distances,
+ *   core_distances); a device-resident dataset requires device-resident outputs. All provided
+ *   outputs must share the same memory space.
+ * - Host-resident outputs never materialize the full [num_rows x k] graph on the GPU.
  * - For batching, `overlap_factor < n_clusters` must hold.
  * - When `core_distances` is provided, mutual-reachability distances are produced (see alpha).
  */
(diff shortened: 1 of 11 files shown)


cpp/src/neighbors/all_neighbors/all_neighbors_builder.cuh
```diff
@@ -604,32 +677,23 @@ struct all_neighbors_builder_brute_force : public all_neighbors_builder<T, IdxT>
...
         num_data_in_cluster,
-        this->k,
         cuvs::distance::is_min_close(bf_params.build_params.metric));
-    } else {
+    } else {  // full build: write directly to the device output in the sink
+      const auto& direct = std::get<typename global_graph_view<T, IdxT>::direct_t>(global.dest);
+      auto distances_view =
+        direct.second.has_value()
+          ? direct.second.value()
+          : raft::make_device_matrix<T, IdxT>(this->res, dataset.extent(0), this->k).view();
```
dantegd: What do you think about keeping an owning device_matrix alive here when distances weren’t requested? This view points into a temporary destroyed at the end of this statement, so the brute force search below receives a dangling pointer for indices-only builds. Could we retain the matrix in a local owner and add indices-only coverage?

cpp/src/neighbors/all_neighbors/all_neighbors.cuh
```diff
@@ -1,23 +1,68 @@
...
+
+  // 2. Check shape consistency
+  RAFT_EXPECTS(dataset.extent(0) == indices.extent(0),
+               "number of rows in dataset should be the same as number of rows in indices matrix");
+  if (distances.has_value()) {
+    RAFT_EXPECTS(indices.extent(0) == distances.value().extent(0) &&
+                   indices.extent(1) == distances.value().extent(1),
+                 "indices matrix and distances matrix has to be the same shape.");
+  }
+  if (core_distances.has_value()) {
```
dantegd, line 58: Should we also require core_distances.size() == dataset.extent(0) here? The host path uses this size as num_rows and indexes the distances matrix with it, while the reachability epilogue is indexed by dataset row. A mismatched C++ view can therefore cause out-of-bounds access. What do you think about validating the extent here?

c/include/cuvs/neighbors/all_neighbors.h
```diff
@@ -28,8 +28,10 @@ extern "C" {
  * provide the dataset on host.
  *
  * Notes:
- * - Outputs (indices, distances, core_distances) are expected to be on device memory.
- * - Host variant accepts host-resident dataset; device variant accepts device-resident dataset.
+ * - A host-resident dataset accepts either host- or device-resident outputs (indices, distances,
+ *   core_distances); a device-resident dataset requires device-resident outputs. All provided
+ *   outputs must share the same memory space.
+ * - Host-resident outputs never materialize the full [num_rows x k] graph on the GPU.
```
dantegd: full_build_host() still allocates full device indices and optional distances when n_clusters == 1, no? So the current wording overstates the memory reduction a little bit. 

What do you think about narrowing this guarantee to batched builds (n_clusters > 1)?

dantegd (changes requested):
PR looks great! Just had 3 comments and a broader concern: batched inner-product results can still be incorrect when similarities are negative if I'm not mistaken. reset_global_matrices uses std::numeric_limits<T>::min(), which is the smallest positive value, so the sentinel can rank ahead of valid negative inner products during the merge. I think switching this to lowest() and adding a centered-data test might be a good idea.

cpp/src/neighbors/all_neighbors/all_neighbors.cuh
jinsolp: Addressed in [1b5b6d9](https://github.com/NVIDIA/cuvs/pull/2313/commits/1b5b6d97807535a45974034bbfe8c1166a7774d6)

c/include/cuvs/neighbors/all_neighbors.h
jinsolp: Addressed in [4c90945](https://github.com/NVIDIA/cuvs/pull/2313/commits/4c90945665413681d5084f5909237cfc78439a6d)

cpp/src/neighbors/all_neighbors/all_neighbors_builder.cuh
jinsolp: Addressed in [c4ede85](https://github.com/NVIDIA/cuvs/pull/2313/commits/c4ede85862962dabdd721a12648e7c9ab7da45b2)

jinsolp:
/ok to test [d848e81](https://github.com/NVIDIA/cuvs/pull/2313/commits/d848e815d4989c76271f73304f4f21fbae884f18)

jinsolp:
/ok to test [ef7ea16](https://github.com/NVIDIA/cuvs/pull/2313/commits/ef7ea16682dbaacdb90e662c50ed1d354a4c3827)

dantegd (approved):

Issue #2292, opened 2026-07-02
[FEA] Out-of-core K-means improvements


cjnolet:
We have received the following feedback from users: 
1. cuVS 26.6.0 host-resident multi-GPU streaming is a meaningful improvement because it allows the dataset to remain in host RAM and stream to GPUs, instead of requiring the full dataset to fit in GPU memory.

2. The RAFT comms path performed close to the cuML MNMG baseline in a benchmark setup: 16.29s per iteration versus 14.64s per iteration for cuML MNMG, about +11.3%.

3. The SNMG path was much slower in the user's test: around 256s per iteration on 8 GPUs, compared with an estimated around 30s per iteration for RAFT comms on 8 GPUs.

4. The user suspects the SNMG gap is mainly from lack of compute-transfer overlap, OMP thread scheduling overhead, batch-level barriers, and memory management overhead.

5. Their profiling also suggests the shared CUTLASS fused distance kernel is significantly slower than their cuBLAS GEMM baseline for this large-K workload.

Suggestions from user:

- [ ] Add pinned/page-locked memory support for streaming, or use internally allocated pinned staging buffers, to avoid pageable-memory bounce copy.
- [ ] Add double buffering or pipelined H2D transfer in SNMG so the next batch can transfer while the current batch is computing.
- [ ] Expose a native Python API path for RAFT comms resources, instead of requiring a duck-typing bridge around pylibraft.common.Handle.
- [ ] Consider a cuBLAS GEMM plus argmin path for large-K distance computation, or otherwise optimize the current CUTLASS fused distance kernel.
- [ ] Improve API usability around return-type consistency, adaptive streaming_batch_size, and fit progress reporting.

viclafargue:
The RAFT and SNMG paths run the same batched multi-GPU KMeans code. However, while RAFT usually uses one process per GPU rank, so each rank has its own CUDA/RMM allocator state. SNMG uses one process with multiple OpenMP threads, so all GPU ranks share process-level allocator state.
Batched KMeans does many temporary device allocations while streaming host batches. Without a memory pool, the SNMG threads contend heavily on allocation/free. This gets much worse with more GPUs and smaller batch sizes.
[...]

PR #2323, opened 2026-07-16
Exposing opt-in double buffering for MG batched KMeans

viclafargue:
Partially answers #2292

- GPUs: 2xQuadro RTX 8000
- Input: float32 host memory; RMM pool: 90%
- Timing: median of 3 runs after one warmup per mode; max iterations: 5
- Speedup is single-buffer time / double-buffer time; values above 1 favor prefetch.
- Device input buffers are shown per rank as single → double; common KMeans workspace is excluded.

| Scenario | Host | Rows/GPU | Features | Clusters | Batch rows | Batches/GPU | Device input buffers/rank | Iterations | Single (s) | Double (s) | Speedup |
| :--- | :--- | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| Copy-dominated | pageable | 4,000,000 | 32 | 2 | 524,288 | 8 | 64 MiB → 128 MiB | 5 | 0.767 | 0.698 | 1.099x |
| Overlap-friendly | pinned | 4,000,000 | 32 | 2 | 524,288 | 8 | 64 MiB → 128 MiB | 5 | 0.594 | 0.491 | 1.209x |
| Compute-dominated | pinned | 4,000,000 | 32 | 16384 | 524,288 | 8 | 64 MiB → 128 MiB | 5 | 5.028 | 4.594 | 1.094x |
| Single batch | pinned | 4,000,000 | 32 | 32 | 4,000,000 | 1 | 488 MiB → 488 MiB | 5 | 0.172 | 0.178 | 0.970x |
| Large buffer | pinned | 8,000,000 | 128 | 2 | 4,000,000 | 2 | 1.91 GiB → 3.81 GiB | 5 | 4.726 | 4.234 | 1.116x |

Prefetch is beneficial but not universal: it gives up to 1.21× speedup when pinned H2D copies can overlap KMeans work, but only ~1.09× in copy- or compute-limited cases and no gain for a single batch.


c/include/cuvs/cluster/kmeans.h
```diff
@@ -199,6 +199,12 @@ struct cuvsKMeansParams {
    * or n_samples for device data.
    */
   int64_t init_size;
+
+  /**
+   * Whether host-resident multi-GPU KMeans should prefetch the next streaming
```
cjnolet:
```suggestion
   * Use prefetching to data latencies in multi-GPU batched k-means when trained on host-resident input data```

File: c/src/cluster/mg_kmeans.cpp [hunk 2]
```diff
@@ -38,6 +38,7 @@ cuvs::cluster::kmeans::params convert_params(const ParamsT& params)
   kmeans_params.batch_centroids      = params.batch_centroids;
   kmeans_params.init_size            = params.init_size;
   kmeans_params.streaming_batch_size = params.streaming_batch_size;
+  kmeans_params.streaming_batch_prefetch = params.streaming_batch_prefetch;
```
cjnolet: I really don't like the use of the term "streaming" here. It carries altogether too much baggage with it. Can we please just remove it here? I think we need to consider removing it from "streaming_batch_size" as well. Streaming to me means a k-means that runs through a streaming mechanism, meaning it can only go forward as an iterator and can restart from where it left off when more data is received. There are variants of k-means that satisfy this, but the batched variant we've implemented does not.

cpp/include/cuvs/cluster/kmeans.hpp
```diff
@@ -155,6 +155,16 @@ struct params : base_params {
    * Default: 0 (process all data at once).
    */
   int64_t streaming_batch_size = 0;
+
+  /**
+   * Whether host-resident multi-GPU KMeans should prefetch the next streaming
```
cjnolet: Please reword to not use words like "whether". The user can imply that, we don't need to say it. A more concise and clear way to state this would be to jump right into what the feasture accomplishes for the user:

"Hide latencies for host-resident inputs in multi-gpu K-means by prefetching the next batch while the current batch is processing. This is done asychronously using different CUDA streams, enabling overlap of host to device (let's not use H2D in public facing documentation, because we can't expect users to know what that means) transfer with the computation by allocating a second device batch buffer on each NCCL rank. In practice this can moderate speedups of up to "XX" over processing batches in serial".

cpp/tests/cluster/kmeans_mg.cu
```diff
@@ -281,6 +283,16 @@ const std::vector<KmeansSNMGInputs<float>> snmg_inputsf = {
...
+  {1000,
+   32,
+   5,
+   0.0001f,
+   kmeans_weight_mode::none,
+   128,
+   1,
+   cuvs::cluster::kmeans::params::Array,
+   20,
+   true},
```
cjnolet, line 296: Should definitely improve the coverage here- what ahppens when there's only 1 batch? what happens when the batch size is not a multiple of the total number of vectors?

python/cuvs/cuvs/cluster/kmeans/kmeans.pyx
```diff
@@ -94,13 +94,21 @@ cdef class KMeansParams:
         increases.
 
         Default: 0 (process all data at once).
+    streaming_batch_prefetch : bool
+        Whether host-resident multi-GPU KMeans should use a second device
```
cjnolet: Please follow suggestions above about the wording of descriptions.

cjnolet (changes requested):
Thanks for giving this a first pass. I'd also like for @tfeher and the devtech team to take a look at this as well. Also, please post benchmarks on here.

c/src/cluster/mg_kmeans.cpp
tarang-jain: The problem with `batch_size` is that we have two other parameters -- `batch_centroids` and `batch_samples`. So `batch_size` can be confusing. Perhaps we should reconsider the names of those two parameters then? Or maybe think of some other prefix for `batch_size`?

cjnolet: What are the meanings of batch samples vs batch size?

tarang-jain: the tile sizes for the distance computation (local reductions are done on this tile) -- i.e. `batch_samples` * `batch_centroids` distances are computed at once. So I was saying that we are already doing what the first part of the flash-kmeans paper mentions (at least that was my understanding)

tarang-jain:
```
  /**
   * batch_samples and batch_centroids are used to tile 1NN computation which is
   * useful to optimize/control the memory footprint
   * Default tile is [batch_samples x n_clusters] i.e. when batch_centroids is 0
   * then don't tile the centroids
   *
   * NB: These parameters are unrelated to streaming_batch_size, which controls how many
   * samples to transfer from host to device per batch when processing out-of-core
   * data.
   */
  int batch_samples = 1 << 15;

  /**
   * if 0 then batch_centroids = n_clusters
   */
  int batch_centroids = 0;

```

cjnolet: If one is for tiling then let's call it like it is, that's a start in the right direction. The longer we make the names, the more a user has to remember.

Let's trying and think of how we can make this easier to use. "Streaming" is not a good word because it has too many other meanings. 

Out-of-core buffered (or batched) is more aligned with the actual purpose but that's also pretty verbose.

There's really 2 levels of batching going on here- one is to control intermediate computation of the 1nn itself, the other is to control the device footprint of each "batching" of data vectors from host. 

Having 3 options with batch in the name is super confusing. But I think we can alleviate this (and avoid having to break existing apis) if we change batch to device_buffer just for the non-tiled version. 

So that would be device_buffer_samples and device_buffer_prefetch. That seems to be more indicative of "I'm buffering up some number of samples on device".


But above and beyond this change- what would be the negative impact to just making prefetch always on? I can imagine the gpu utilization would fill up with work before the memory would fill up with batches, which would suggest to me that there's a buffer buffer size that can remain pretty constant for most gpus.

viclafargue: Prefetch is not universally beneficial: it only improves throughput when the GPU can overlap host-to-device copies with meaningful computation. If either copies or compute dominate completely, the second stream and buffer provide little or no gain.

Making it unconditional would also make the memory tradeoff unconditional. Every applicable rank would reserve space for two device batches instead of one, even for workloads where overlap cannot help. That extra allocation can be large enough to cause an out-of-memory failure for configurations that otherwise fit.

cjnolet: Let's get some benchmarks attached to this PR so that we can evaluate. Again , I suspect the actual device utilization is likely to cap long before the memory does, which would indicate that even if there's no gain in performance, so long as there is not a drop in performance, the simplicity for the user could be justified in the default. But let's back that up with numbers.

cjnolet: BTW I'm not necessarily say we should or shouldn't make it the default, but I am saying we should look into the impact and provide real evidence either way (concrete numbers / formula always helps but benchmarks should ultimately guide us).

tarang-jain: If we are renaming the streaming_batch_size param, we must do it quick because its in the C layer. If we dont change it this release, it'll be locked user experience for the next 6 months due to ABI stability.

cjnolet: Agree we should try to get this out in time for the release, but just in case it doesn't, we can always add the new parameter and handle it accordingly underneath until we are able to remove the old. Think about it like a deprecation.

cjnolet: Just in case, I suppose we could always open up a smaller PR to do just the rename. Let's see how things go next week.

tarang-jain: > I suppose we could always open up a smaller PR to do just the rename

I created this: https://github.com/NVIDIA/cuvs/pull/2328

cpp/src/cluster/detail/kmeans_mg.cuh
```diff
@@ -140,7 +142,18 @@ void mnmg_fit(
...
 
-  auto stream     = comms.stream();
+  auto stream = comms.stream();
+  std::unique_ptr<rmm::cuda_stream> data_copy_stream_owner;
+  rmm::cuda_stream_view data_copy_stream{stream};
+  bool enable_data_prefetch = false;
+  if constexpr (!data_on_device) {
+    if (params.streaming_batch_prefetch) {
+      data_copy_stream_owner =
+        std::make_unique<rmm::cuda_stream>(rmm::cuda_stream::flags::non_blocking);
```
tarang-jain, line 152: Rather than creating the prefetch stream, lets just get it from the handle:
`cuvs::spatial::knn::detail::utils::get_prefetch_stream(handle);`

tarang-jain: Also i dont think we need an explicit argument for opt-in prefetch. If an extra stream is supplied in the pool, we simply use it for prefetch, else we dont prefetch. This has been the pattern in IVF-PQ, IVF-Flat and others.

tarang-jain (changes requested):

PR #2325, opened 2026-07-17
CAGRA ACE build: avoid rounding down graph degree during sub-index build

achirkin:
Passing `graph_degree / 2` to `from_hnsw` heuristics to get `graph_degree` back as a result unintentionally rounds down odd degrees, which corrupts subsequent copy operations during ACE index build.
We workaround this issue by assigning the graph_degree explicitly on top of the heuristics.


achirkin:
Closing in favor of https://github.com/NVIDIA/cuvs/pull/2335

PR #2328, opened 2026-07-20
[KMeans] Rename streaming_batch_size param

tarang-jain:
This PR renames the `streaming_batch_size` param for kmeans to:
1. Better convey the meaning of the param -- link to discussion: https://github.com/NVIDIA/cuvs/pull/2323#discussion_r3599476090
2. Avoid confusion with the already existing `batch_samples` and `batch_centroids` params

The idea is that the `streaming` word can be misleading. We are "buffering up" a batch of samples onto the device.

Merge after #2329


jinsolp (commented):
Can we call it `n_device_buffer_samples` to make it obvious that it's the number of samples?

cpp/include/cuvs/cluster/kmeans.hpp
```diff
@@ -113,7 +113,7 @@ struct params : base_params {
    * Default tile is [batch_samples x n_clusters] i.e. when batch_centroids is 0
    * then don't tile the centroids
    *
-   * NB: These parameters are unrelated to streaming_batch_size, which controls how many
+   * NB: These parameters are unrelated to device_buffer_samples, which controls how many
    * samples to transfer from host to device per batch when processing out-of-core
```
cjnolet: Instead of "controls how many samples to tranfer from host to device per batch", maybe say something like "specifies the number of training vectors that get buffered on device at a time when the training vectors are passed in on host".

tarang-jain:
>Can we call it n_device_buffer_samples to make it obvious that it's the number of samples?

I think @jinsolp lets not make param names even longer. `device_buffer_samples` is already pretty informative.

cpp/include/cuvs/cluster/kmeans.hpp
tarang-jain: Done!

cjnolet (approved):

PR #2331, opened 2026-07-20
Update Comparison Branch for ABI Stability Git Workflow

tarang-jain:
When the release branch is created, this check must compare against the release branch and not main

### DIFF
--- a/.github/workflows/check-c-abi.yaml
+++ b/.github/workflows/check-c-abi.yaml
@@ -62,10 +62,10 @@ jobs:
         run: |
           echo "⚠️ No baseline artifacts found, extracting from release/26.08 branch..."
           echo "This is slower but ensures we always have a baseline for comparison."
-          git worktree add ../cuvs-main origin/release/26.08
+          git worktree add ../cuvs-release-26.08 origin/release/26.08
           mkdir -p baseline
           check-c-abi extract \
-            --header-path ../cuvs-main/c/include \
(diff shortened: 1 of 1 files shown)


cjnolet (approved):

.github/workflows/check-c-abi.yaml
```diff
@@ -46,40 +46,40 @@ jobs:
...
 
-      - name: Extract baseline ABI from main branch (fallback 2)
+      - name: Extract baseline ABI from release/26.08 branch (fallback 2)
         if: steps.download-merge-base.outcome == 'failure' && steps.download-main.outcome == 'failure'
         run: |
-          echo "⚠️ No baseline artifacts found, extracting from main branch..."
+          echo "⚠️ No baseline artifacts found, extracting from release/26.08 branch..."
           echo "This is slower but ensures we always have a baseline for comparison."
-          git worktree add ../cuvs-main main
+          git worktree add ../cuvs-main origin/release/26.08
```
cjnolet: I think maybe we should change this from cuvs-main to cuvs-26.08.

cjnolet: I suspect we would want to have these separate so that main and the release branch can coexist without overwriting each other? I'm not sure if this is a cached copy, but better to be safe than sorry.

tarang-jain: this is just the name of a directory, but yes I get your point.

msarahan (approved):

PR #2337, opened 2026-07-21
cuvs: add package version retrieval function

QuerthDP:
Add a function that allows to retrieve the current version of a `cuvs` Rust package.

Also add a simple unit test for the function.

### DIFF
--- a/rust/cuvs/src/lib.rs
+++ b/rust/cuvs/src/lib.rs
@@ -26,3 +26,18 @@ pub mod vamana;
 pub use dlpack::{AsDlTensor, AsDlTensorMut, DLPackError, DLTensorView, DLTensorViewMut, DType};
 pub use error::{Error, Result};
 pub use resources::Resources;
+
+/// Returns the version of this `cuvs` crate.
+pub fn version() -> &'static str {
+    env!("CARGO_PKG_VERSION")
+}
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+
+    #[test]
+    fn test_version() {
+        assert_eq!(version(), env!("CARGO_PKG_VERSION"));
+    }
+}


QuerthDP:
I cannot assign labels myself.
I would assume the labels would be `improvement` (?) and `non-breaking` since I'm adding a new function here so this does not break any existing user code (except for wildcard imports).

QuerthDP:
@cjnolet how do we proceed with it? Would you consider such feature to be added? We want to use this library as a GPU VS POC and engine (crate) version is required in our API.

yan-zaretskiy:
What is the use case you're after? I don't see value in wrapping `env!("CARGO_PKG_VERSION");` into a function. With that said, I do plan to expose the actual C library version getter, via `ffi::cuvsVersionGet(&mut major, &mut minor, &mut patch)`.

rust/cuvs/src/lib.rs
```diff
@@ -26,3 +26,18 @@ pub mod vamana;
 pub use dlpack::{AsDlTensor, AsDlTensorMut, DLPackError, DLTensorView, DLTensorViewMut, DType};
 pub use error::{Error, Result};
 pub use resources::Resources;
+
+/// Returns the version of this `cuvs` crate.
+pub fn version() -> &'static str {
+    env!("CARGO_PKG_VERSION")
+}
```
yan-zaretskiy, line 33: I would suggest instead to wrap the underlying FFI function like so.
```rust
/// Returns the cuVS library version as `(major, minor, patch)`.
pub fn version() -> Result<(u16, u16, u16), LibraryError> {
    let mut major: u16 = 0;
    let mut minor: u16 = 0;
    let mut patch: u16 = 0;

    // SAFETY:
    // - All three pointers are valid, aligned `u16` locals.
    let status = unsafe { ffi::cuvsVersionGet(&mut major, &mut minor, &mut patch) };
    check_cuvs(status)?;
    Ok((major, minor, patch))
}
```

yan-zaretskiy (changes requested):

QuerthDP:
> What is the use case you're after?

Return engine (cuVS crate) version via the `/info` endpoint in our service.

QuerthDP:
> I do plan to expose the actual C library version getter, via `ffi::cuvsVersionGet(&mut major, &mut minor, &mut patch)`.

When could it be done and released?

QuerthDP:
Any progress here?

yan-zaretskiy:
> Any progress here?

I pushed a PR with my proposed implementation [here](https://github.com/NVIDIA/cuvs/pull/2409). It wraps the underlying C API for querying the .so version of libcuvs and the test expresses our FFI versioning policy.

QuerthDP:
> > Any progress here?
> 
> I pushed a PR with my proposed implementation [here](https://github.com/NVIDIA/cuvs/pull/2409). It wraps the underlying C API for querying the .so version of libcuvs and the test expresses our FFI versioning policy.

Thanks, looks good. 
Let's close this once you merge.

PR #2345, opened 2026-07-22
Add CAGRA build heuristics independent of HNSW parameters

achirkin:
Split the `cagra::index_params::from_hnsw_params` into a core `cagra::index_params::from_dataset` and a thin wrapper ``cagra::index_params::from_hnsw_params`. This solves two problems:
1. Avoid using the HNSW parameters that do not make sense for CAGRA
2. Allow HNSW parameter heuristics change without affecting CAGRA in future

The new parameter defaults are selected to yield exactly the same configurations at old call sites of `from_hnsw_params`


cpp/include/cuvs/neighbors/cagra.hpp
```diff
@@ -223,6 +224,51 @@ struct index_params : cuvs::neighbors::index_params {
...
+   *
+   * This is the main CAGRA build heuristic: it chooses between NN-descent and IVF-PQ based on the
+   * dataset size and tunes their parameters based on the target intermediate graph degree and the
+   * requested build quality. It returns the `graph_build_params` variant only; the caller is
+   * responsible for setting `graph_degree` / `intermediate_graph_degree`.
+   *
+   * @param dataset The shape of the input dataset
+   * @param intermediate_graph_degree The intermediate (kNN) graph degree the build should target
+   * @param metric The distance metric to search
+   * @param build_quality Higher values increase the build quality (and cost) up to a point.
```
tfeher, line 240: What is the range of the values the user could provide here?

cpp/include/cuvs/neighbors/cagra.hpp
```diff
@@ -223,6 +224,51 @@ struct index_params : cuvs::neighbors::index_params {
...
+  /**
+   * @brief Create CAGRA index parameters heuristically tuned for a dataset.
+   *
+   * Returns default CAGRA `index_params` with `graph_build_params` selected by
+   * `graph_build_heuristic` for the given dataset.
+   *
+   * @param dataset The shape of the input dataset
+   * @param graph_degree Degree of the output graph.
+   * @param metric The distance metric to search
+   * @param build_quality Higher values increase the build quality (and cost) up to a point.
```
tfeher, line 258: Please document valid range for build_quality. Also it would make sense to give a hint for `graph_degree`. E.g multiples of 32 is a good practice.

tfeher (changes requested):
Thanks Artem, it makes sense to separate the dataset shape dependent CAGRA heuristics to general CAGRA heuristic and HNSW specific ones. The PR looks good to me, apart from two small docstring issues.

cpp/include/cuvs/neighbors/cagra.hpp
achirkin: Any. The default is 7, which keep the previous behavior non-broken. I added the docstring.

cjnolet (changes requested):
This is a relatively small PR in comparison to the dataset changes (that we absolutely need merged for this release).

If we keep merging (lower priority) changes that depend on the current dataset APIs, we aren't going to be able to merge the datasets api changes in time for the release.

Let's at least wait until @HowardHuang1's PR is merged for this one. Given how close it is to the release, I'd almost suggest we proactively push this to 26.10.

achirkin:
@cjnolet I understand, this PR surely can wait.
FIY, it shouldn't overlap with the dataset API PR in any way, it only adds two functions that consume extents (not an mdspan, mdarray, or a new dataset type): `caga::index_params::from_dataset` and `cagra::index_params::graph_build_heuristic`.

cjnolet:
Thanks @achirkin. I see now, this is expecting the shape information individually, instead of through an mdspan/dataset.

@imotov if not too much trouble, can you merge this in your heuristic PR and work from that? I agree with Artem here that the change seems not too terrible on the Java side. Will help us streamline a bit more.

cjnolet (approved):

tfeher (approved):
Thanks Artem, LGTM.

PR #2346, opened 2026-07-22
Use batched unfused 1-NN for KMeans on Blackwell

viclafargue:
An investigation of large-scale KMeans failures on Blackwell revealed that the unfused 1-NN path bypasses KMeans distance batching and materializes the complete n_samples × n_clusters distance matrix. This can require hundreds of gigabytes of temporary workspace and may overflow int32 indexing.

This change batches the unfused 1-NN implementation on Blackwell, restoring bounded workspace usage. It also renames the dispatch booleans to make the metric-capability and implementation-selection decisions clearer.

### DIFF
--- a/cpp/src/cluster/detail/minClusterDistanceCompute.cu
+++ b/cpp/src/cluster/detail/minClusterDistanceCompute.cu
@@ -1,5 +1,5 @@
 /*
- * SPDX-FileCopyrightText: Copyright (c) 2025-2026, NVIDIA CORPORATION.
+ * SPDX-FileCopyrightText: Copyright (c) 2025-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
  * SPDX-License-Identifier: Apache-2.0
  */
 
@@ -27,15 +27,15 @@ void minClusterAndDistanceCompute(
   int batch_centroids,
   rmm::device_uvector<char>& workspace)
 {
-  cudaStream_t stream = raft::resource::get_cuda_stream(handle);
-  auto n_samples      = X.extent(0);
-  auto n_features     = X.extent(1);
-  auto n_clusters     = centroids.extent(0);
-  bool is_fused       = metric == cuvs::distance::DistanceType::L2Expanded ||
-                  metric == cuvs::distance::DistanceType::L2SqrtExpanded ||
-                  metric == cuvs::distance::DistanceType::CosineExpanded;
+  cudaStream_t stream           = raft::resource::get_cuda_stream(handle);
+  auto n_samples                = X.extent(0);
+  auto n_features               = X.extent(1);
+  auto n_clusters               = centroids.extent(0);
+  const bool can_use_fused_path = metric == cuvs::distance::DistanceType::L2Expanded ||
+                                  metric == cuvs::distance::DistanceType::L2SqrtExpanded ||
+                                  metric == cuvs::distance::DistanceType::CosineExpanded;
 
-  if (is_fused) {
+  if (can_use_fused_path) {
     L2NormBuf_OR_DistBuf.resize(n_clusters, stream);
     auto centroidsNorm =
       raft::make_device_vector_view<DataT, IndexT>(L2NormBuf_OR_DistBuf.data(), n_clusters);
@@ -51,10 +51,10 @@ void minClusterAndDistanceCompute(
     raft::KeyValuePair<IndexT, DataT> initial_value(0, std::numeric_limits<DataT>::max());
     raft::matrix::fill(handle, minClusterAndDistance, initial_value);
 
-    bool should_use_fused =
+    const bool use_fused_path =
       use_fused<DataT, IndexT, IndexT>(handle, n_samples, n_clusters, n_features);
 
-    if (should_use_fused) {
+    if (use_fused_path) {
       workspace.resize((sizeof(int)) * n_samples, stream);
 
       cuvs::distance::fusedDistanceNNMinReduce<DataT, raft::KeyValuePair<IndexT, DataT>, IndexT>(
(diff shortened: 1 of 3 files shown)


tfeher (changes requested):
I don't think this is not the right solution. Missing the tiling logic was an oversight in #1768. It was actually [discussed](https://github.com/NVIDIA/cuvs/pull/1768/changes#r3194057962), and was thought to be addressed, but apparently there is another call path where this is not working.

I think the correct solution would add a simple 1D tiling along n_samples. Wouldn't that solve the problem?

tarang-jain:
> I think the correct solution would add a simple 1D tiling along n_samples. Wouldn't that solve the problem?

A 1D tiling would also have to be done on the centroids right?

tfeher:
> I think the correct solution would add a simple 1D tiling along n_samples

Having said that, this code path will be reworked in https://github.com/NVIDIA/cuvs/pull/2249, and what we do here will be deprecated. Therefore we should not develop anything complicated to fix this.

tfeher:
> A 1D tiling would also have to be done on the centroids right?

Depends on what are typical parameter ranges we expect this to work on. If the target use case would not work with tiling only on n_samples, then I would just go for the simpler solution proposed by Victor.

tarang-jain:
>Depends on what are typical parameter ranges we expect this to work on. If the target use case would not work with tiling only on n_samples, then I would just go for the simpler solution proposed by Victor.

The behabior of the API is to acknowledge tile parameters on both the centroids and the samples. That said, kmeans_balanced might have different requirements. Changing the use_fused function could mean affecting the behavior there as well, which may not be intended for balanced kmeans.

cpp/src/cluster/detail/kmeans_common.cuh
```diff
@@ -76,8 +76,8 @@ bool use_fused(const raft::resources& handle, IdxT m, IdxT n, IdxT k)
     // On Hopper if m, n are bigger than 4096, use fused
     return true;
   } else if (prop.major >= 10) {
-    // On Blackwell onwards, use unfused
-    return false;
+    // On Blackwell onwards, use fused
+    return true;
```
tarang-jain: Here, we should think about the effect on balanced kmeans. Avoidance of tiling there may have be intentional. Then if balanced and regular kmeans are expected to behave differently, we must solve this problem in another way -- kmeans should be able to override this function and force the use of fused kernel always.

viclafargue:
Balanced KMeans uses the same assignment function but does not pass the regular KMeans tiling parameters. Instead, it bounds memory through its own outer minibatching and calls the assignment with `batch_samples = 0` and `batch_centroids = 0`. Its use of the unfused Blackwell path was therefore intentional (normally executes as a single inner tile).

My latest commit fix restores Blackwell to the shared unfused heuristic and makes regular KMeans’s unfused assignment honor `batch_samples` and `batch_centroids`, avoiding the full distance-matrix allocation.

achirkin (commented):
Thanks for the PR! The allocation/batch sizing seems right. I'd suggest to add some tests to cover specifically this codepath - it's not clear whether it is ever tested in the CI with the current test suite.

viclafargue:
Added a test that exercises no batching, sample-only batching, centroid-only batching, and combined batching for both 32-bit and 64-bit indices. It validates labels and inertia on every GPU and verifies that batching reduces device-memory allocations.
We use `predict` because `minClusterAndDistanceCompute` is an internal, non-exported template and cannot be linked directly from the test executable.

achirkin (approved):
Thank for the new tests! LGTM

cpp/tests/cluster/kmeans_predict_batching.cu
```diff
@@ -0,0 +1,156 @@
...
+    !detail::use_fused<float, int, int>(handle, test_n_samples, test_n_clusters, test_n_features);
+
+  for (std::size_t i = 1; i < batch_configs.size(); ++i) {
+    auto config      = batch_configs[i];
+    auto int_bytes   = run_predict_with_batching<int>(handle, config);
+    auto int64_bytes = run_predict_with_batching<int64_t>(handle, config);
+
+    if (uses_unfused_path) {
+      // Before the batching fix every configuration allocated the full
+      // samples-by-centroids matrix.
```
tarang-jain: Can we get rid of this comment? We dont need to make note of something in the code that was broken earlier and is already fixed here.

tarang-jain (approved):
Thanks. LGTM except one small comment.

cpp/tests/cluster/kmeans_predict_batching.cu
viclafargue: Done

cpp/src/cluster/detail/minClusterDistanceCompute.cu
```diff
@@ -74,26 +74,66 @@ void minClusterAndDistanceCompute(
...
+              [cIdx] __device__(KeyValueT current, KeyValueT batch) {
+                batch.key += cIdx;
+                return batch.value < current.value ? batch : current;
+              },
+              raft::make_const_mdspan(minClusterAndDistanceView),
+              batchMinView);
+          }
+          cIdx += nc;
+        }
+        dIdx += ns;
```
tfeher, line 135: Do we have a specific reason to keep these separate from the loop header? It would be more readable to keep it in the header.

tfeher (approved):
Thanks Victor, LGTM!

cpp/src/cluster/detail/minClusterDistanceCompute.cu
viclafargue: Moving these increments into the headers would indeed be clearer. Let's address this change as part of the follow-up that we will need to revise the tiling system in light of the introduction of the new 1-NN with cuTile (#2249).

PR #2352, opened 2026-07-22
Fast CAGRA Index Merge

landrumb:
This PR implements the Fastener graph merge operation. This PR supports merging for `float`, `half`, `int8`, and `uint8` dtypes, and euclidean distances.

There was originally specialization to use int8 GEMM for the integer types, but I ran into portability issues on Ada and it turns out that using the same f32 path for both is simpler and not substantially slower. Currently investigating switching the unified path to TF32 to use tensor cores.

The core logic resides in `cagra_merge_scaffold.cuh`. 

This PR adds 7.64 MiB to `libcuvs.so`, a 2.94% increase.

image
(H100)
image
image
image
image
image


cpp/src/neighbors/detail/cagra/cagra_merge_scaffold.cuh
```diff
@@ -0,0 +1,1299 @@
...
+  std::vector<uint32_t> ends_host;
+  starts_host.reserve((partitions.memberships.size() + leaf_size - 1) / leaf_size);
+  ends_host.reserve(starts_host.capacity());
+
+  int64_t covered = 0;
+  for (auto const& range : partitions.ranges) {
+    RAFT_EXPECTS(range.start == covered && range.end > range.start &&
+                   range.end <= static_cast<int64_t>(partitions.memberships.size()),
+                 "Fastener partition ranges must compactly cover all memberships");
+    for (int64_t start = range.start; start < range.end; start += leaf_size) {
```
dantegd: I think there’s a connectivity edge case here. Since the sorts are stable, rows within a partition can remain grouped by their source index. Slicing an oversized partition into consecutive chunks can then produce leaves containing rows from only one input, and the leaf kernel skips every same-origin pair.

For example, identical inputs whose row counts are multiples of leaf_size can end up with every chunk aligned to a single origin, leaving the original graphs disconnected. Could we either keep splitting or deterministically mix origins before slicing? A regression test for that case would also be helpful.

cpp/src/neighbors/detail/cagra/cagra_merge.cuh
```diff
@@ -175,4 +204,236 @@ index<T, IdxT> merge(raft::resources const& handle,
...
+      handle, raft::make_const_mdspan(dataset.view()), preflight.offsets, scaffold_params);
+    return merge_scaffold::append_to_input_graphs<T, IdxT>(
+      handle, indices, preflight.offsets, raft::make_const_mdspan(scaffold.view()));
+  }();
+
+  {
+    raft::common::nvtx::range<cuvs::common::nvtx::domain::cuvs> scope("cagra::merge/append/sort");
+    cagra::detail::graph::sort_knn_graph_device_inplace(
+      handle, params.metric, raft::make_const_mdspan(dataset.view()), merged_graph.view());
+    merged_graph = merge_scaffold::cap_sorted_graph(
```
dantegd, line 466: Are we capping this a little too early? If an input already has graph_degree local neighbors that are closer than its scaffold neighbors, this nearest prefix cap can remove every cross index edge before graph::optimize() gets a chance to consider it.

It also leaves the optimizer with the same input and output degree, so there’s no candidate headroom for robust pruning. Was wondering if maybe we keep up to max(intermediate_graph_degree, graph_degree) candidates here, bounded by the available candidate width, and let the optimizer produce the final graph_degree?

cpp/src/neighbors/detail/cagra/cagra_merge.cuh
```diff
@@ -175,4 +204,236 @@ index<T, IdxT> merge(raft::resources const& handle,
...
+
+  auto preflight = preflight_fastener<T, IdxT>(params, merge_params, indices, row_filter);
+  if (!preflight.eligible) {
+    if (merge_params.algo == cagra::merge_algo::AUTO) {
+      return merge_rebuild(handle, params, indices, row_filter);
+    }
+    RAFT_FAIL("FASTENER cagra::merge is unsupported: %s", preflight.reason.c_str());
+  }
+
+  return merge_fastener(handle, params, merge_params, indices, preflight);
```
dantegd: Should AUTO fall back to rebuild if Fastener runs out of memory? Preflight can validate shapes and limits, but it can’t know whether the consolidated dataset, temporary GEMM buffers, and graphs will all fit at runtime.

The previous merge path could fall back to host memory, so changing the existing overload to AUTO means some merges that previously succeeded may now fail. I think explicit FASTENER should still fail, but AUTO retrying the rebuild path would preserve the existing behavior. What do you think?

cpp/src/neighbors/detail/cagra/cagra_merge.cuh
```diff
@@ -175,4 +204,236 @@ index<T, IdxT> merge(raft::resources const& handle,
...
+  if (!merge_scaffold::leaf_gemm_supported<T>(result.dim, merge_params.leaf_size)) {
+    return reject("dataset dimension exceeds the leaf GEMM workspace limit");
+  }
+  if (rows > std::numeric_limits<uint32_t>::max() / spill) {
+    return reject("combined rows times the configured spill width must fit in uint32_t");
+  }
+  if (params.graph_degree == 0 || static_cast<uint64_t>(params.graph_degree) >= rows) {
+    return reject("graph_degree must be positive and smaller than the combined row count");
+  }
+  if (static_cast<uint64_t>(params.graph_degree) > max_input_degree + scaffold_degree) {
```
dantegd, line 352: I think one limit is missing from preflight: the graph passed to sort_knn_graph_device_inplace() can’t exceed kMaxSortDegree (1024). For example, an input degree of 1024 plus any scaffold currently passes preflight and then fails later in the sorter. In AUTO mode, that also means we miss the rebuild fallback.

Could we include max_input_degree + scaffold_degree <= kMaxSortDegree in the eligibility check?

cpp/tests/neighbors/ann_cagra/test_merge_fastener.cu
```diff
@@ -0,0 +1,965 @@
...
+    merge_params fastener;
+    fastener.algo               = merge_algo::FASTENER;
+    input.physical_merge_params = fastener;
+  }
+  return inputs;
+}
+}  // namespace
+
+typedef AnnCagraIndexMergeTest<float, float, std::uint32_t> AnnCagraFastenerMergeRecallTest;
+TEST_P(AnnCagraFastenerMergeRecallTest, DefaultFloatMergeSearchRecallAgainstBruteForce)
```
dantegd, line 1231: Could we add recall coverage for at least one non-float dtype and one merge with more than two inputs? The current dtype tests show that the output graph is structurally valid, but only the small two-part float case verifies that the merged graph is actually useful for search. That feels important if AUTO is going to become the default.

cpp/src/neighbors/detail/cagra/cagra_merge_scaffold.cuh
```diff
@@ -0,0 +1,1299 @@
...
+                   split_params const& params,
+                   split_context& context,
+                   rmm::device_uvector<uint32_t>& keys,
+                   rmm::device_uvector<partition_membership>& memberships)
+{
+  auto stream = raft::resource::get_cuda_stream(res);
+  auto dim    = dataset.extent(1);
+
+  // Cut every parent in the bucket into fixed-height tiles
+  int tile_rows = context.assignment_tile_rows;
```
dantegd: Could we derive tile_rows from the dimension and padded leader count instead of always using 2048? 

Preflight only checks the leaf-GEMM shape, while assignment needs roughly (tile_rows + padded_leaders) * dim + tile_rows * padded_leaders float elements. Some configurations therefore pass preflight and later fail the workspace assertion here if I'm not mistaken. Could we reduce the tile height per bucket and reject only when the leader matrix plus even one point row cannot fit?

cpp/src/neighbors/detail/cagra/cagra_merge_scaffold.cuh
```diff
@@ -0,0 +1,1299 @@
...
+
+  carry_parents(res, parents, plan, keys, memberships);
+  auto buckets = bucket_split_parents(plan, params);
+  for (size_t bucket_index = 0; bucket_index < buckets.size(); ++bucket_index) {
+    if (buckets[bucket_index].empty()) { continue; }
+    assign_bucket(res,
+                  dataset,
+                  parents,
+                  buckets[bucket_index],
+                  1 << bucket_index,
```
dantegd, line 742: Curious, have you measured how much work is going to padded leaders here? A parent just over a power-of-two boundary could nearly doubles the gathered leader storage, dot-product GEMM, distance materialization, and select_k width.

cpp/src/neighbors/detail/cagra/cagra_merge_scaffold.cuh
```diff
@@ -0,0 +1,1299 @@
...
+      dim,
+      parents.memberships.data(),
+      device_tiles.data(),
+      static_cast<int>(batch_size),
+      tile_rows,
+      tile_points.data());
+    RAFT_CUDA_TRY(cudaGetLastError());
+
+    int leader_blocks = strided_grid_size(static_cast<int64_t>(batch_size * leader_elements));
+    manyway_gather_tile_leaders_kernel<<<leader_blocks, THREADS_PER_BLOCK, 0, stream>>>(
```
dantegd, line 645: It looks like every tile of the same parent gathers the same leader matrix again. For large parents split across many 2048-row tiles, that can add substantial conversion and memory traffic, especially when the leader count is close to the tile height. Could we gather each parent’s leaders once and reuse them across its tiles, or did benchmarking show that the added bookkeeping costs more?

cpp/src/neighbors/detail/cagra/cagra_merge_scaffold.cuh
```diff
@@ -0,0 +1,1299 @@
...
+  uint32_t const* leader_ids,
+  partition_membership const* input_memberships,
+  assignment_tile const* tiles)
+{
+  int64_t linear = static_cast<int64_t>(blockIdx.x) * blockDim.x + threadIdx.x;
+  int64_t stride = static_cast<int64_t>(blockDim.x) * gridDim.x;
+  int64_t total  = static_cast<int64_t>(batch_size) * tile_rows * padded_leaders;
+  for (; linear < total; linear += stride) {
+    int leader     = static_cast<int>(linear % padded_leaders);
+    int row        = static_cast<int>((linear / padded_leaders) % tile_rows);
```
dantegd: This is a small thing, but wanted to ask: for dimensions that aren’t naturally 16-byte aligned, this allocates and copies a second full consolidated dataset while all input datasets are still alive. Would it be practical to consolidate directly into aligned storage and pass a row stride through the scaffold/sort kernels? That could materially lower peak memory for int8/uint8 and other unaligned dimensions.

cpp/src/neighbors/detail/cagra/cagra_merge_scaffold.cuh
```diff
@@ -0,0 +1,1299 @@
...
+      source.extent(0),
+      source.extent(1),
+      graph.data_handle(),
+      graph_degree,
+      base_degree,
+      scaffold_degree,
+      static_cast<uint32_t>(offsets[part]));
+    RAFT_CUDA_TRY(cudaGetLastError());
+  }
+  raft::resource::sync_stream(res);
```
dantegd: Is this synchronization needed? The helper has no local temporary buffers consumed by the copy kernels, and the caller immediately sorts the returned graph on the same resource stream. Stream ordering should already preserve that dependency

dantegd (changes requested):
Overall, I really like the direction and PR! Reusing the partition graphs, adding a lightweight cross-partition scaffold, and then running the existing CAGRA optimization is such a great way to make merge substantially cheaper than a rebuild. The split between preflight, consolidation, scaffold construction, and graph assembly also makes the implementation fairly easy to follow.

One API  concern: the C API still calls the existing C++ overload without merge_params,  so it needs updating, right?

landrumb:
The Java API is still lagging, but is a bit more to convert. Could add to this PR or defer, up to @dantegd.

cpp/src/neighbors/detail/cagra/cagra_merge.cuh
landrumb: The rationale here was that if you had no cross-input edges that were closer than the edges you started with, you probably didn't need them. I think I evaluated this and found that it saved time on optimize without a real impact on recall, but I'll check again for you and add another comment with numbers.

cpp/src/neighbors/detail/cagra/cagra_merge_scaffold.cuh
landrumb: I compared against always doing non-batched GEMM, which was substantially slower, and batching with individual sizes, which was moderately faster but iirc added more LoC than felt worth it.

cpp/src/neighbors/detail/cagra/cagra_merge_scaffold.cuh
landrumb: | Dataset | Arm | Merge ms | Δ | Recall@12 | Δ | Rep spread |
|---|---|---:|---:|---:|---:|---:|
| wiki-1m | baseline | 545.6 | — | 0.991683 | — | 1.1e-4 |
| wiki-1m | roundrobin | 546.6 | +0.18% | 0.991875 | +0.000192 | 4.1e-5 |
| wiki-1m | resplit | 857.7 | +57.2% | 0.992083 | +0.000400 | 2.5e-5 |
| openai-2m | baseline | 2271.5 | — | 0.958383 | — | 7.5e-5 |
| openai-2m | roundrobin | 2272.1 | +0.03% | 0.958758 | +0.000375 | 8.4e-5 |
| openai-2m | resplit | 3292.8 | +45.0% | 0.958158 | −0.000225 | 1.2e-4 |
| yfcc-10m | baseline | 2744.5 | — | 0.981367 | — | 7.5e-5 |
| yfcc-10m | roundrobin | 2734.6 | −0.36% | 0.983583 | +0.002216 | 3.4e-5 |
| yfcc-10m | resplit | 3915.5 | +42.7% | 0.982783 | +0.001416 | 4.2e-5 |

Good catch here, the interaction between concatenating the datasets and stable sorting here didn't occur to me. Experiments show that:
- 6% of leaves are single origin in the current design on wiki-1M, so this is happening, although even with round-robin you get like 5-10% of those still just because some single origin leaves happen for organic geometric reasons
- round-robin slightly bumps recall with an impact on build time that seems to be within noise
- further geometric splitting until you have leaves all at or below the max leaf size is expensive and doesn't seem amazing for recall.

Also trying not splitting them further and just treating them as big leaves, which I don't think will work but want to rule out before committing to round robin.

landrumb: | Dataset | Arm | Merge ms | Δ | Recall@12 | vs baseline | vs roundrobin | Spread |
|---|---|---:|---:|---:|---:|---:|---:|
| wiki-1m | baseline | 514.5 | — | 0.991775 | — | +0.000008 | 2.5e-5 |
| wiki-1m | roundrobin | 516.2 | +0.35% | 0.991767 | −0.000008 | — | 1.0e-4 |
| wiki-1m | big512 | 542.1 | +5.37% | 0.992825 | +0.001050 | +0.001058 | 8e-6 |
| wiki-1m | big1024 | 554.0 | +7.68% | 0.993075 | +0.001300 | +0.001308 | 1.7e-5 |
| openai-2m | baseline | 2203.4 | — | 0.958417 | — | −0.000308 | 6.7e-5 |
| openai-2m | roundrobin | 2206.6 | +0.15% | 0.958725 | +0.000308 | — | 1.6e-5 |
| openai-2m | big512 | 2273.0 | +3.16% | 0.960408 | +0.001991 | +0.001683 | 1.0e-4 |
| openai-2m | big1024 | 2287.2 | +3.80% | 0.960958 | +0.002541 | +0.002233 | 4.2e-5 |
| yfcc-10m | baseline | 2419.1 | — | 0.981433 | — | −0.002125 | 5.9e-5 |
| yfcc-10m | roundrobin | 2430.8 | +0.49% | 0.983558 | +0.002125 | — | 1.2e-4 |
| yfcc-10m | big512 | 2528.8 | +4.54% | 0.984625 | +0.003192 | +0.001067 | 1.7e-5 |
| yfcc-10m | big1024 | 2594.4 | +7.25% | 0.985058 | +0.003625 | +0.001500 | 6.7e-5 |

Treating leaves of size up to 512 (default max leaf size is 256) as normal leaves that don't need round robin gives slightly higher recall but has some added complexity and takes slightly longer to do. For now, I'll stick to round robin for simplicity but let me know if you think this is a worthwhile tradeoff.

cpp/src/neighbors/detail/cagra/cagra_merge.cuh
landrumb: The "headroom" case below is `min(max(intermediate_graph_degree, graph_degree), candidate_width)`, which ultimately ends up passing up to 88 points to optimize.
| Dataset | Mode | Merge ms | Δ | Recall@12 | Δ | Spread |
|---|---|---:|---:|---:|---:|---:|
| wiki-1m | default | 512.9 | — | 0.991842 | — | 2.5e-5 |
| | headroom | 597.8 | **+16.56%** | 0.991792 | −0.000050 | 3.4e-5 |
| openai-2m | default | 2201.9 | — | 0.958742 | — | 7.5e-5 |
| | headroom | 2402.6 | **+9.12%** | 0.957100 | **−0.001642** | 8.3e-5 |
| yfcc-10m | default | 2418.7 | — | 0.983567 | — | 2.5e-5 |
| | headroom | 3259.6 | **+34.76%** | 0.982217 | **−0.001350** | 3.4e-5 |

We get larger build times with the same or very slightly worse recall.

cpp/src/neighbors/detail/cagra/cagra_merge.cuh
landrumb: Good idea, made this change

cpp/src/neighbors/detail/cagra/cagra_merge_scaffold.cuh
landrumb: This seems to perform ~the same but adds a small amount of complexity to do the caching.
| Dataset | Arm | Merge ms (median) | Range | Δ |
|---|---|---:|---:|---:|
| wiki-1m | baseline | 517.0 | 506.0–518.8 | — |
| wiki-1m | cached | 515.7 | 509.5–518.7 | −0.25% |
| openai-2m | baseline | 2214.8 | 2199.9–2216.9 | — |
| openai-2m | cached | 2220.0 | 2208.8–2221.4 | +0.23% |
| yfcc-10m | baseline | 2422.0 | 2405.9–2429.0 | — |
| yfcc-10m | cached | 2427.8 | 2417.4–2432.1 | +0.24% |

landrumb:
Currently failing because the wheel builds are too big. Not sure there's much I can do about this.

cpp/src/neighbors/detail/cagra/cagra_merge_scaffold.cuh
```diff
@@ -0,0 +1,1371 @@
...
+inline constexpr int THREADS_PER_BLOCK       = 256;
+inline constexpr int MAX_STRIDED_GRID_BLOCKS = 1 << 20;
+/** Warps per block in the one-row-per-warp kernels; their launch geometry derives from this. */
+inline constexpr int ROW_WARPS_PER_BLOCK = 4;
+
+/** Grid size for the grid-stride kernels: blocks of THREADS_PER_BLOCK covering `items`, capped so
+ * oversized workloads loop within resident threads instead. */
+inline auto strided_grid_size(int64_t items) -> int
+{
+  return static_cast<int>(std::min<int64_t>((items + THREADS_PER_BLOCK - 1) / THREADS_PER_BLOCK,
```
achirkin: nitpick
```suggestion
  return static_cast<int>(std::min<int64_t>(raft::div_rounding_up_safe(items, THREADS_PER_BLOCK),
```

cpp/src/neighbors/detail/cagra/cagra_merge_scaffold.cuh
```diff
@@ -0,0 +1,1371 @@
...
+  split_context(rmm::cuda_stream_view stream, int64_t rows)
+    : norms(static_cast<size_t>(rows), stream)
+  {
+  }
+
+  rmm::device_uvector<float> norms;
+  int assignment_tile_rows    = ASSIGNMENT_TILE_ROWS;
+  size_t gemm_workspace_bytes = GEMM_WORKSPACE_BYTES;
+  uint64_t seed               = DETERMINISTIC_SEED;
+};
```
achirkin, line 102: Let's try to switch to raft primitives here: `raft::device_vector<float>` as the buffer and `raft::resources` as an argument.
Then, we need to consider: does this context fit into the bounded workspace memory or does it grow with the problem size unbounded? Depending on the answer, use `raft::get_workspace_resource_ref` or `raft::get_large_workspace_resource_ref`.
By doing this, we let the user control where the working memory for the algorithm resides, enable better accounting (via tools like `raft::memory_tracking_resources`), and prepare for the dry run execution (feature soon to be enabled in raft).

cpp/src/neighbors/detail/cagra/cagra_merge_scaffold.cuh
```diff
@@ -0,0 +1,1371 @@
...
+struct partition_range {
+  uint32_t key  = 0;
+  int64_t start = 0;
+  int64_t end   = 0;
+};
+
+struct partition_set {
+  rmm::device_uvector<partition_membership> memberships;
+  std::vector<partition_range> ranges;
+};
```
achirkin, line 122: same as above:
```suggestion
  raft::device_vector<partition_membership> memberships;
  raft::host_vector<partition_range> ranges;
```

cpp/src/neighbors/detail/cagra/cagra_merge_scaffold.cuh
```diff
@@ -0,0 +1,1371 @@
...
+  if (row < rows) { memberships[row] = {static_cast<uint32_t>(row), uint16_t{0}, uint16_t{0}}; }
+}
+
+/** Represent the full dataset as one identity partition without copying any vectors. */
+inline auto make_root_partition(raft::resources const& res, int64_t rows) -> partition_set
+{
+  auto stream = raft::resource::get_cuda_stream(res);
+  partition_set root{rmm::device_uvector<partition_membership>(static_cast<size_t>(rows), stream),
+                     {{uint32_t{0}, int64_t{0}, rows}}};
+  int blocks = static_cast<int>((rows + THREADS_PER_BLOCK - 1) / THREADS_PER_BLOCK);
```
achirkin: nitpick
```suggestion
  int blocks = raft::div_rounding_up_safe<int>(rows, THREADS_PER_BLOCK);
```

cpp/src/neighbors/detail/cagra/cagra_merge_scaffold.cuh
```diff
@@ -0,0 +1,1371 @@
...
+                                             b_stride,
+                                             &beta,
+                                             out,
+                                             CUDA_R_32F,
+                                             a_rows,
+                                             out_stride,
+                                             batch_count,
+                                             GEMM_COMPUTE_TYPE,
+                                             CUBLAS_GEMM_DEFAULT));
+}
```
achirkin: Do we have a wrapper for this in raft? Would be nice to add. Also, is there a cublaslt alternative? this would be preferred - we try to migrate to cublaslt-only, because it allows safer multithreading.

cpp/src/neighbors/detail/cagra/cagra_merge_scaffold.cuh
```diff
@@ -0,0 +1,1371 @@
...
+  size_t batch_capacity = std::max<size_t>(
+    1, std::min<size_t>(tiles.size(), context.gemm_workspace_bytes / bytes_per_batch));
+
+  rmm::device_uvector<assignment_tile> device_tiles(batch_capacity, stream);
+  rmm::device_uvector<float> tile_points(batch_capacity * point_elements, stream);
+  rmm::device_uvector<float> tile_leaders(batch_capacity * leader_elements, stream);
+  rmm::device_uvector<float> tile_dots(batch_capacity * dot_elements, stream);
+  rmm::device_uvector<uint32_t> tile_leader_ids(batch_capacity * padded_leaders, stream);
+  rmm::device_uvector<float> selected_distances(batch_capacity * selected_elements, stream);
+  rmm::device_uvector<int> selected_leaders(batch_capacity * selected_elements, stream);
```
achirkin: Try to move to raft::device_vector and use the workspace memory resources (the batch is bounded, right?).

cpp/src/neighbors/detail/cagra/cagra_merge.cuh
```diff
@@ -177,4 +207,261 @@ index<T, IdxT> merge(raft::resources const& handle,
...
+  merged_index.update_graph(handle, std::move(optimized_graph));
+  if (params.attach_dataset_on_build) {
+    // CAGRA search assumes 16-byte row alignment (vectorized loads). Consolidate remains dense for
+    // the scaffold; attach through make_aligned_dataset so unaligned dims are padded.
+    merged_index.update_dataset(handle, make_aligned_dataset(handle, std::move(dataset), 16));
+  } else {
+    using ds_idx_type = typename index<T, IdxT>::dataset_index_type;
+    merged_index.update_dataset(
+      handle, std::make_unique<cuvs::neighbors::empty_dataset<ds_idx_type>>(preflight.dim));
+  }
```
achirkin: I guess, we should move the dataset consolidation into a separate place, perhaps as a part of the dataset API. CC @HowardHuang1 to discuss what would be the place for dataset merging under the new API and who should own it?

cpp/src/neighbors/detail/cagra/cagra_merge_scaffold.cuh
```diff
@@ -0,0 +1,1371 @@
...
+  size_t dot_elements      = static_cast<size_t>(tile_rows) * padded_leaders;
+  size_t selected_elements = static_cast<size_t>(tile_rows) * params.fanout;
+  size_t bytes_per_batch =
+    (point_elements + leader_elements + dot_elements + selected_elements) * sizeof(float) +
+    static_cast<size_t>(padded_leaders) * sizeof(uint32_t) + selected_elements * sizeof(int) +
+    sizeof(assignment_tile);
+  RAFT_EXPECTS(bytes_per_batch <= context.gemm_workspace_bytes,
+               "Fastener assignment workspace is too small for one tile");
+  size_t batch_capacity = std::max<size_t>(
+    1, std::min<size_t>(tiles.size(), context.gemm_workspace_bytes / bytes_per_batch));
```
achirkin: I'd suggest to size the batch_capacity based on the raft workspace size that can be queried at runtime, rather than based on the compile-time constant.

achirkin (changes requested):
Thanks for this extensive contribution! Since the datataset API changes are in, the main request is to rethink where we allocate the new, consolidated dataset during merge (see below).
Then, please try to refactor the code to use raft mdarrays and mdspans in place of RMM and use the workspace/large_workspace resources for all temporary allocations - this will help us to keep track of memory requirements (which seem to be plenty).

cpp/src/neighbors/detail/cagra/cagra_merge_scaffold.cuh
landrumb: I've gone through and I think this is done correctly now, let me know if I misunderstand the usage of the raft allocs. Is there a way I should be testing preparedness for dry run execution?

cpp/src/neighbors/detail/cagra/cagra_merge_scaffold.cuh
landrumb: I can't find a wrapper for it in raft, but I think it could be done with the right configuration of `cublasLtMatmul`, which uses a bunch of descriptors I would need to understand a little better before trying to use. Do you want me to switch to that?

landrumb: switched to the `cublasLtMatmul`

cpp/src/neighbors/detail/cagra/cagra_merge.cuh
landrumb: I can make an issue for this, but could we leave it here for now?

cpp/src/neighbors/detail/cagra/cagra_merge_scaffold.cuh
landrumb: I think this was incidentally fixed by the dataset API.

cpp/tests/neighbors/ann_cagra/test_merge_fastener.cu
landrumb: added 
- `Uint8TwoWayMergeSearchRecall`: non-float (uint8_t) recall
- `FloatFourWayMergeSearchRecall`: more than two inputs
- `Uint8FourWayMergeSearchRecall`: covers both

cpp/src/neighbors/detail/cagra/cagra_merge.cuh
landrumb: Adding offline context to the thread:
The dataset merging logic now present is just copying vectors into storage that's now provided by the caller per interface changes that were introduced with the dataset API. The rebuild version of this takes a bitmap to filter them, and the fastener version doesn't.

cpp/src/neighbors/detail/cagra/cagra_merge_scaffold.cuh
```diff
@@ -0,0 +1,1524 @@
...
+           aligned(capacity * leader_elements * sizeof(float)) +
+           aligned(capacity * dot_elements * sizeof(float)) +
+           aligned(capacity * static_cast<size_t>(padded_leaders) * sizeof(uint32_t)) +
+           aligned(capacity * selected_elements * sizeof(float)) +
+           aligned(capacity * selected_elements * sizeof(int)) +
+           aligned(capacity * dot_elements * sizeof(float)) +
+           aligned(capacity * dot_elements * sizeof(int));
+  };
+
+  RAFT_EXPECTS(assignment_workspace_bytes(1, 1) <= workspace_bytes,
```
dantegd: I think there’s still a gap in the AUTO fallback here. Preflight only checks the leaf workspace, so a configuration can pass preflight but fail this minimum assignment-workspace check when the padded leader matrix is larger. RAFT_EXPECTS throws raft::logic_error, which isn’t caught by the std::bad_alloc fallback in cagra_merge.cuh. 

Could we either cover this minimum shape in preflight or make this failure fall back under AUTO? It would also be helpful to test a case where one leaf fits but one assignment row does not if it's possible.

cpp/src/neighbors/detail/cagra/cagra_merge_scaffold.cuh
```diff
@@ -0,0 +1,1524 @@
...
+{
+  auto stream = raft::resource::get_cuda_stream(res);
+  // Memberships are dataset-sized, so they come from the large workspace rather than the bounded
+  // one; the single root range is host metadata.
+  auto memberships = raft::make_device_mdarray<partition_membership, int64_t>(
+    res, raft::resource::get_large_workspace_resource_ref(res), raft::make_extents<int64_t>(rows));
+  auto ranges = raft::make_host_vector<partition_range, int64_t>(res, 1);
+  ranges(0)   = partition_range{uint32_t{0}, int64_t{0}, rows};
+  partition_set root{std::move(memberships), std::move(ranges)};
+  int blocks = raft::div_rounding_up_safe<int>(static_cast<int>(rows), THREADS_PER_BLOCK);
```
dantegd: Might be a good idea to keep rows as int64_t through the division and cast only the resulting block count? 

Preflight permits combined row counts above int_max, but this cast overflows before the much smaller grid size is calculated.

cpp/include/cuvs/neighbors/cagra.hpp
```diff
@@ -3429,16 +3439,75 @@ void serialize_to_hnswlib(
...
- * All input indices must use the same `DatasetViewT` (dense padded or standard device views).
+ * All input indices must use the same `DatasetViewT` (dense padded or standard device views),
+ * and must share one row stride.
+ *
+ * @param[in] res RAFT resources used for the merge.
+ * @param[in] params Parameters for the returned CAGRA index.
+ * @param[in] indices CAGRA indices to merge.
+ * @param[in] merged_dataset Caller-owned storage for the consolidated dataset. Size it with
+ * `cuvs::neighbors::cagra::detail::merged_dataset_size`.
+ * @param[in] row_filter Optional row filter. Any filter selects rebuild in AUTO and is rejected by
```
dantegd: can users actually use this sizing instruction from the installed C++ API?

merged_dataset_size() is defined under cpp/src and isn’t declared in the public header. Could we either expose an appropriate public helper?

dantegd (changes requested):
Have only the small concerns, besides it things look solid! 

Also, it looks like a few working artifacts may have slipped into the PR that need to be removed before merge: PR_2352_REVIEW_WORK.md, pipnn.pdf and  FASTENER_PR_PLAN.md, right?

cpp/src/neighbors/detail/cagra/cagra_merge_scaffold.cuh
landrumb: Good point, changed.

cpp/src/neighbors/detail/cagra/cagra_merge_scaffold.cuh
landrumb: Moved it to preflight and added the test

cpp/include/cuvs/neighbors/cagra.hpp
landrumb: Looking more at it I'm not super convinced this is a useful enough helper to expose. It just gives the number of rows in the output, which is either a popcount of the filter or the sum of the inputs, and leaves the user to extract the dimension and stride to figure out the size of the alloc. I think it makes more sense to either surface a helper that gives the size and/or shape of the allocation or surface no helper at all.

cpp/include/cuvs/neighbors/cagra.hpp
```diff
@@ -3429,16 +3439,75 @@ void serialize_to_hnswlib(
...
+struct merge_params {
+  merge_algo algo        = merge_algo::AUTO;
+  uint32_t levels        = 2;
+  uint32_t root_fanout   = 2;
+  uint32_t lower_fanout  = 3;
+  double leader_fraction = 0.02;
+  uint32_t max_leaders   = 1024;
+  uint32_t leaf_size     = 256;
+  uint32_t leaf_degree   = 4;
+};
```
julianmi, line 3461: We might want to add a section in the documentation on these parameters and their typical values.

c/src/neighbors/cagra.cpp
```diff
@@ -151,7 +152,8 @@ static void merge_indices_for_layout(
       auto owner    = std::make_unique<owner_t>(std::move(matrix), dim);
       auto view     = owner->as_dataset_view();
       auto merged_idx =
-        cuvs::neighbors::cagra::merge(*res_ptr, params_cpp, index_ptrs, view, row_filter);
+        cuvs::neighbors::cagra::merge(
```
julianmi, line 155: Does this `std::bad_alloc` catch need to `RAFT_FAIL` when using the FASTENER algorithm to match the C++ behavior? It looks to me like we would rebuild similarly to using AUTO.

cpp/src/neighbors/detail/cagra/cagra_merge.cuh
```diff
@@ -192,4 +203,344 @@ cuvs::neighbors::cagra::index<T, IdxT, DatasetViewT> merge(
...
+  RAFT_EXPECTS(static_cast<int64_t>(merged_dataset.dim()) == preflight.dim,
+               "merged_dataset dimension (%u) must equal the input dimension (%ld)",
+               unsigned(merged_dataset.dim()),
+               long(preflight.dim));
+
+  auto const output_const_view = merged_dataset.view();
+  auto* destination            = const_cast<T*>(output_const_view.data_handle());
+  {
+    raft::common::nvtx::range<cuvs::common::nvtx::domain::cuvs> scope("cagra::merge/consolidate");
+    // Zero first so the pad columns are defined: the sorter treats the padded width as the
```
julianmi: Can we check if we need zeroing (`stride == dim`) at all and if so only zero the padded columns?

cpp/src/neighbors/detail/cagra/cagra_merge_scaffold.cuh
```diff
@@ -0,0 +1,1557 @@
...
+                          build_params const& params) -> raft::device_matrix<uint32_t, int64_t>
+{
+  auto stream             = raft::resource::get_cuda_stream(res);
+  int64_t rows            = dataset.extent(0);
+  int leaf_size           = static_cast<int>(params.leaf_size);
+  int leaf_degree         = static_cast<int>(params.leaf_degree);
+  auto const& memberships = leaves.partitions->memberships;
+
+  // Prefill every slot with its own row id, leaf KNN overwrites the slots it fills
+  auto graph = raft::make_device_mdarray<uint32_t, int64_t>(
```
julianmi, line 906: Could `build_leaf_neighbors()` write directly into the scaffold suffix of the final candidate graph?

cpp/src/neighbors/detail/cagra/cagra_merge_scaffold.cuh
```diff
@@ -0,0 +1,1557 @@
...
+  __shared__ uint32_t leaf_origins[MAX_LEAF_SIZE];
+  for (int i = threadIdx.x; i < leaf_n; i += blockDim.x) {
+    records[i]      = memberships[start + static_cast<uint32_t>(i) * leaf_stride];
+    leaf_origins[i] = origins[records[i].id];
+  }
+  __syncthreads();
+
+  int u = threadIdx.x;
+  if (u >= leaf_n) { return; }
+  double top_d[MAX_LEAF_DEGREE];
```
julianmi: Why do we promote to `double` the `float` Gram matrix. Does a FP64 distance calculation improve recall?

cpp/src/neighbors/detail/cagra/cagra_merge_scaffold.cuh
```diff
@@ -0,0 +1,1557 @@
...
+                                                           uint32_t const* scaffold,
+                                                           int64_t source_rows,
+                                                           int64_t source_degree,
+                                                           uint32_t* destination,
+                                                           int64_t destination_degree,
+                                                           int64_t base_degree,
+                                                           int64_t scaffold_degree,
+                                                           uint32_t offset)
+{
+  int64_t row = static_cast<int64_t>(blockIdx.x) * blockDim.x + threadIdx.x;
```
julianmi: Did you benchmark this launch configuration against assigning one warp per row? This might help coalesce memory accesses.

cpp/src/neighbors/detail/cagra/cagra_merge_scaffold.cuh
```diff
@@ -0,0 +1,1557 @@
...
+  constexpr int WARPS_PER_BLOCK = THREADS_PER_BLOCK / raft::WarpSize;
+  int lane                      = threadIdx.x % raft::WarpSize;
+  int warp                      = threadIdx.x / raft::WarpSize;
+  int64_t row                   = static_cast<int64_t>(blockIdx.x) * WARPS_PER_BLOCK + warp;
+  if (row >= rows) { return; }
+
+  int64_t input_base  = row * input_degree;
+  int64_t output_base = row * output_degree;
+  int selected        = 0;
+  for (int64_t tile = 0; tile < input_degree && selected < output_degree; tile += raft::WarpSize) {
```
julianmi: Did you profile this kernel? It looks to me like the inner loop is unbalanced and we might see a lot of branch divergence. If this is performance relevant, we might consider a hash set per warp or similar.

landrumb: Just profiled and it's 0.5-2.1% of total merge time on different datasets. I think probably not worth the complexity right now.

cpp/src/neighbors/detail/cagra/cagra_merge_scaffold.cuh
landrumb: One warp per row is substantially faster, good call, changed.

cpp/src/neighbors/detail/cagra/cagra_merge_scaffold.cuh
landrumb: Not sure, probably some historical thing I've forgotten. Fixed now.

cpp/src/neighbors/detail/cagra/cagra_merge_scaffold.cuh
landrumb: good idea, changed this and it was surprisingly clean to do so. Makes no impact on performance that I can see but saves a small amount of peak memory.

cpp/src/neighbors/detail/cagra/cagra_merge_scaffold.cuh
julianmi: Yes, let's leave it for now.

cpp/src/neighbors/detail/cagra/cagra_merge_scaffold.cuh
```diff
@@ -0,0 +1,1553 @@
...
+  float alpha = 1.0f;
+  float beta  = 0.0f;
+
+  using matmul_descriptor = std::unique_ptr<std::remove_pointer_t<cublasLtMatmulDesc_t>,
+                                            decltype(&cublasLtMatmulDescDestroy)>;
+  using matrix_layout     = std::unique_ptr<std::remove_pointer_t<cublasLtMatrixLayout_t>,
+                                            decltype(&cublasLtMatrixLayoutDestroy)>;
+
+  cublasLtMatmulDesc_t operation_raw = nullptr;
+  RAFT_CUBLAS_TRY(cublasLtMatmulDescCreate(&operation_raw, GEMM_COMPUTE_TYPE, CUDA_R_32F));
```
achirkin: Please destroy the descriptors and matrix layouts after you use them, (otherwise, this is a memory leak currently).
Have a look at how it's one in raft: https://github.com/NVIDIA/raft/blob/main/cpp/include/raft/linalg/detail/cublaslt_wrappers.hpp

I think,  the best solution though would be to create a tiny PR in raft to add the corresponding wrapper/helper for the configuration you need and then use it here.

cpp/src/neighbors/detail/cagra/cagra_merge_scaffold.cuh
```diff
@@ -0,0 +1,1553 @@
...
+  size_t min_batch_capacity = 1;
+  size_t max_batch_capacity = tiles.size();
+  while (min_batch_capacity < max_batch_capacity) {
+    size_t candidate = min_batch_capacity + (max_batch_capacity - min_batch_capacity + 1) / 2;
+    if (tile_bytes(candidate, static_cast<size_t>(tile_rows)) <= workspace_bytes) {
+      min_batch_capacity = candidate;
+    } else {
+      max_batch_capacity = candidate - 1;
+    }
+  }
```
achirkin, line 593: Do you expect the workspace should fit reasonably-sized batches in any expected configuration?
In some edge cases - just to protect us from complete failure - we fallback to the large workspace resource. Do you think we should do the same here or not?

achirkin (commented):
Thanks for the refactoring! One problem with the switch to cublasLtMatmul (see below).

cpp/src/neighbors/detail/cagra/cagra_merge_scaffold.cuh
landrumb: For pretty much any expected situation it should fit many batches. The current implementation just reduces the batch size as needed and fails if a singe-entry batch doesn't fit, but I think it would be hard to run into this problem without severely constricting the workspace, so really more a question about the workspace being reasonable than the configuration being reasonable. If we're assuming the workspace is reasonable (73 MB, which is the tile size with the default number of leaders and 4096 dim, seems like the largest I'd reasonably expect, and even if the workspace was only 73 MB, it would work with half-sized tiles), this shouldn't be an issue. I feel like making a crazy small workspace seems like a user choice suggesting they would explicitly reject using the large workspace resource, so I think we should leave it.

cpp/src/neighbors/detail/cagra/cagra_merge_scaffold.cuh
landrumb: Added raft PR for this: https://github.com/NVIDIA/raft/pull/3106

cpp/include/cuvs/neighbors/cagra.hpp
landrumb: Removed mention and discussed requirements per slack thread.

cpp/include/cuvs/neighbors/cagra.hpp
landrumb: Added in https://github.com/NVIDIA/cuvs/pull/2352/commits/78296cafcf71769a2dac4852bfc3a5eee0fd1290

c/include/cuvs/neighbors/cagra.h
```diff
@@ -1034,6 +1061,27 @@ CUVS_EXPORT cuvsError_t cuvsCagraMerge(cuvsResources_t res,
...
+ * @brief Merge multiple CAGRA indices with explicit merge parameters.
+ *
+ * @param[in] res cuvsResources_t opaque C handle
+ * @param[in] params cuvsCagraIndexParams_t parameters for the output index
+ * @param[in] merge_params cuvsCagraMergeParams_t parameters controlling the merge algorithm, or
+ *                         NULL to use AUTO defaults
+ * @param[in] indices Array of input cuvsCagraIndex_t handles to merge
+ * @param[in] num_indices Number of input indices
+ * @param[in] filter Filter that can be used to filter out vectors from the merged index
+ * @param[out] output_index Output handle initialized with `cuvsCagraIndexCreate`
```
dantegd, line 1081: Could we carry the merged_dataset documentation from cuvsCagraMerge() into this new entry point? 

It’s currently missing from the parameter list, and there’s an important behavioral difference worth mentioning: for an unfiltered merge, AUTO and REBUILD can use the existing host fallback when device allocation fails, while explicit FASTENER reports the failure. That affects both allocation expectations and the dataset lifetime contract.

cpp/include/cuvs/neighbors/cagra.hpp
```diff
@@ -3429,16 +3439,76 @@ void serialize_to_hnswlib(
...
+ * All input indices must use the same `DatasetViewT` (dense padded or standard device views),
+ * and must share one row stride.
+ *
+ * @param[in] res RAFT resources used for the merge.
+ * @param[in] params Parameters for the returned CAGRA index.
+ * @param[in] indices CAGRA indices to merge.
+ * @param[in] merged_dataset Caller-owned storage for the consolidated dataset. Should have as many
+ * rows as the combined datasets, will (after the application of `row_filter`, if present) and have
+ * the same dimension and stride as the input datasets.
+ * @param[in] row_filter Optional row filter. Any filter selects rebuild in AUTO and is rejected by
```
dantegd: I think this sentence got a little tangled while removing the helper reference. 

A wording like: “Must have one row per retained vector after applying row_filter, and the same dimension and stride as the input datasets” would be nice

dantegd (changes requested):
Just have two docstring comments, otherwise things look great to me!

cpp/include/cuvs/neighbors/cagra.hpp
landrumb: fixed

c/include/cuvs/neighbors/cagra.h
landrumb: Done, and added the AUTO and REBUILD note only on the params one because that distinction isn't exposed by the other function.

dantegd (approved):

cpp/src/neighbors/detail/cagra/cagra_merge_scaffold.cuh
landrumb: PR merged and change applied in https://github.com/NVIDIA/cuvs/pull/2352/commits/4bd095e193cdfd4ff807cf906ef73caf160b7001

cpp/bench/ann/src/cuvs/cuvs_ann_bench_param_parser.h
```diff
@@ -416,6 +416,37 @@ void parse_build_param(const nlohmann::json& conf,
...
+  }
+  if (conf.contains("fastener_max_leaders")) {
+    param.merge_params.max_leaders = conf.at("fastener_max_leaders");
+  }
+  if (conf.contains("fastener_leaf_size")) {
+    param.merge_params.leaf_size = conf.at("fastener_leaf_size");
+  }
+  if (conf.contains("fastener_leaf_degree")) {
+    param.merge_params.leaf_degree = conf.at("fastener_leaf_degree");
+  }
```
achirkin: A nitpick: use `collect_conf_with_prefix(conf, "fastener_")` to decouple the fastener params from build params, same as we do with compression params a few lines below.

achirkin (approved):
Thanks, LGTM!

PR #2357, opened 2026-07-23
enforce 'yamllint' checks

jameslamb:
## Description

Contributes to https://github.com/rapidsai/build-planning/issues/305

Proposes enforcing `yamllint` checks here. My primary motivation is to catch correctness issues in `dependencies.yaml` files, like duplicate entries silently resolving to the last one or indentation mistakes leading to filters being ignored.

But this also has some side benefits for consistency, which makes it a bit easier to write automation.


python/cuvs_bench/cuvs_bench/config/algos/faiss_cpu_hnsw_flat.yaml
```diff
@@ -5,6 +5,6 @@ groups:
       M: [16, 32]
       efConstruction: [32, 64, 128, 256, 512]
     search:
-      bounded_queue: [False, True]
-      check_relative_distance: [False, True]
+      bounded_queue: [false, true]
```
jameslamb, line 8: `yamllint` caught these.

YAML 1.1 treats `True` (capital `T`) as a boolean `true`.

YAML 1.2 only supports `true` / `false` and treats `True` as the string `"True"`. See https://yaml.org/spec/1.2.2/#10212-boolean

This maybe hasn't bitten us yet because a lot of YAML parsers still respect the 1.1 spec (example: https://github.com/yaml/pyyaml/issues/116), but it's a nice bit of future-proofing to just use the values supported by both versions.

.pre-commit-config.yaml
```diff
@@ -2,205 +2,216 @@
 # SPDX-License-Identifier: Apache-2.0
 
 repos:
-      - repo: https://github.com/pre-commit/pre-commit-hooks
```
jameslamb, line 5: This diff looks huge because there was inconsistent indentation used throughout. Indentation matters for correctness in YAML. I didn't find any cases where the wrong indentation led to an actual issue, but still think it'd be good to standardize on one... so I chose 2 spaces, which lots of other RAPIDS projects use for this file.

In the future, if someone uses a different indentation here `yamllint` with catch it and complain, so it should help us stay consistent.

.pre-commit-config.yaml
```diff
@@ -2,205 +2,216 @@
...
+    rev: v1.38.0
+    hooks:
+      - id: yamllint
+        additional_dependencies: [pyyaml]
+        exclude: |
+          (?x)^(
+            conda/environments/.*|
+            cpp/[.]clang-format$|
+            cpp/[.]clang-tidy$
+          )
```
jameslamb, line 210: It's hard to spot in this big diff, so just calling out that here's the new `yamllint` config I'm proposing introducing.

Everything else in this file's diff is just whitespace changes.

.yamllint.yaml
```diff
@@ -0,0 +1,37 @@
...
+  comments-indentation: disable
+  document-end: disable
+  document-start: disable
+  key-duplicates:
+    forbid-duplicated-merge-keys: true
+  line-length: disable
+  truthy:
+    allowed-values: ['false', 'true']
+    # having problematic value in keys is rare... and also
+    # GitHub Actions' choie of 'on:' triggers this check
```
jameslamb:
```suggestion
    # GitHub Actions' choice of 'on:' triggers this check
```

bdice (approved):

cjnolet (approved):

PR #2322, opened 2026-07-15
Fix missing trailing barrier in ivf_rabitq blockReduceSum

hhy3:

PR #2362, opened 2026-07-24
Fuse RaBitQ block reductions into one templated blockReduceSumN<N>

Stardust-SJF:
Thanks @hhy3 for finding and fixing the bug! This PR removes the barrier between block reductions by replacing the per-value `blockReduceSum` (and PR #2322's helper-internal barrier) with a fused `blockReduceSumN<N>` that reduces all N factor components in a single pass over a shared[N][32] memory space. Each reduction group is now one call, the space is never reused back-to-back, so the shared-memory reuse race is removed. Also, it fixes another potential shared-memory race in the `compute_best_rescale_parallel` function.

By the way, it will also slightly improve the performance for the quantization process.


jamxia155:
/ok to test [22172a1](https://github.com/NVIDIA/cuvs/pull/2362/commits/22172a1bceff8fd0506d2121c1c6b88588dae465)

tfeher:
/ok to test [a2c5747](https://github.com/NVIDIA/cuvs/pull/2362/commits/a2c5747771bb661d12c25159e7c04aad59972ed2)

PR #2366, opened 2026-07-27
Revert "Update DLPack to 1.0 (#2354)"

bdice:
This reverts commit 350a3b0ffd48eba860d48a7fb564bb9fe77580cb.

This update introduced conflicts with cuDF, which is already in code freeze for 26.08. I have a proposal in https://github.com/rapidsai/build-planning/issues/308 to update to DLPack 1.x, which can occur in 26.10 or later because it is not an ABI breaking change (which was the original reason this targeted 26.08, because 26.08 is allowed to break ABI).


trxcllnt (approved):

robertmaynard (approved):

trxcllnt:
cc: @divyegala

jameslamb (approved):

PR #2369, opened 2026-07-28
Fix ``cuvs-bench`` opensearch recall and save results

### LINKED ISSUE #2358: [BUG] ef_search search parameter failed to sweep correctly
**Description**
In cuvs_bench, OpenSearchBackend.search() keeps neighbors only from the last ef_search in a sweep, so recall is computed for just that one point — every other sweep point is dropped and the recall-vs-QPS curve collapses to a single dot.

The loop records per-point timing but no per-point recall and overwrites last_neighbors every iteration ([opensearch.py#L960-L969](https://github.com/rapidsai/cuvs/blob/e626144ba3fa536ecd4c75f4e95fefe6c056403a/python/cuvs_bench/cuvs_bench/backends/opensearch.py#L960-L969)). SearchResult has a single recall/neighbors field, so the orchestrator computes recall once, from those last-point neighbors ([orchestrator.py#L265-L274](https://github.com/rapidsai/cuvs/blob/e626144ba3fa536ecd4c75f4e95fefe6c056403a/python/cuvs_bench/cuvs_bench/orchestrator/orchestrator.py#L265-L274)). 

TLDR: 
For both built-in parameter sweep presets (--groups) in config/algos/opensearch_faiss_hnsw.yaml, i.e. test and base, both yield only one data point (value of ef_search) per build (value of m). No pareto frontiers built for each value of m. 

**Steps to Reproduce**
With OpenSearch running and an opensearch_faiss_hnsw index built, run the test group (ef_search: [10, 20]):
```
BENCH_GROUPS=test DATASET=sift-128-euclidean K=10 \
docker compose -f docker-compose.multinode.yml --profile client run --rm bench
```
Two sweep points go in, one comes out: ef_search=20   recall@10 0.9967   QPS 2045.4
 Omitted 1 timing-only search rows without recall.

The results JSON has one entry and the plot is a single point. (base, 4 m × 10 ef_search, likewise collapses from 40 points to 4)

**Expected Behavior**
Each ef_search should get its own recall (from that point's neighbors), so an M×E sweep produces M×E points and actual frontier curves.

**Environment Details**
- Environment location: Cloud (AWS EC2), Docker
- Method of cuVS install: from source, the Dockerfile does git clone https://github.com/rapidsai/cuvs.git (@ e626144) and adds python/cuvs_bench to the image
- docker compose --profile client build bench then ... run --rm bench

See [https://github.com/jrbourbeau/cuvs/blob/9cf5c45ea07d884e07242f4727c027354e5232b0/deploy/AWS_MULTINODE_SINGLE_ROLE_SETUP.md](url)

**Additional context**
Fix: compute recall inside the sweep loop per ef_search and store it per entry in per_search_param_results, rather than keeping only the last point's neighbors.

jrbourbeau:
This PR includes updates to close https://github.com/NVIDIA/cuvs/issues/2358 and also address the opensearch backend not saving persistent results.

Primary changes are:

1. Backend `search` methods now return a list of results that are passed to the orchestrator where recall can be computed for each result separately (xref https://github.com/NVIDIA/cuvs/issues/2358#issuecomment-5064187759)
2. All backends now explicitly return in-memory build / search results which are then saved to CSV files. `--data-export` flag is deprecated since this always happens now. 

cc @jnke2016 

Closes https://github.com/NVIDIA/cuvs/issues/2358


python/cuvs_bench/cuvs_bench/tests/test_opensearch.py
```diff
@@ -503,12 +512,42 @@ def test_build_and_search(self, live_backend):
...
 
-        search_result = live_backend.search(dataset, [idx], k=k)
+        search_result = live_backend.search(dataset, [idx], k=k)[0]
         assert search_result.success
         assert search_result.recall == 0.0
         assert search_result.queries_per_second > 0
         assert search_result.neighbors.shape == (10, k)
-        assert len(search_result.metadata["per_search_param_results"]) == 1
+
+    def test_recall_is_computed_for_each_search_parameter(self, live_backend):
```
jrbourbeau, line 521: Note: this test makes sense for all the backends. In the future we should probably have this as an e2e test we run for each backend.

jnke2016 (approved):
Thanks @jrbourbeau , this looks great. Converting search() to return a list of SearchResults (one per search-param combo) and letting the orchestrator compute recall per result cleanly fixes #2358 , and it keeps recall centralized across all backends. I ran the bench unit tests and it passes. LGTM

jrbourbeau (commented):
Thanks @jnke2016! 

@NVIDIA/cuvs-python-codeowners does anyone have bandwidth to take a look at this?

cjnolet (approved):

PR #2376, opened 2026-07-29
[WIP] Preload cudart before cuvs native library load to prevent downstream class load errors

### LINKED ISSUE #2300: [BUG] CuVSProvider.java needs to load cudart to prevent provider load error
I encountered a class load failure with behavior with cuv-lucene while trying GPU acceleration feature in Apache Solr 10. Turns out that in the flow to get the CuVSProvider instance, precisely at this line:  https://github.com/apache/solr/blob/releases/solr/10.0.0/solr/modules/cuvs/src/java/org/apache/solr/cuvs/GpuMetricsService.java#L154

the call runs into a class loader exception and renders future calls to CuVSProvider in that error state. 

RCA pointed to "libcudart.so" not being loaded and I submitted the following fix for the same in Solr : https://github.com/apache/solr/pull/4328

However in my opinion the same fix really belongs in CuVSProvider.provider() method (https://github.com/NVIDIA/cuvs/blob/main/java/cuvs-java/src/main/java/com/nvidia/cuvs/spi/CuVSProvider.java#L196) to load the system library so that the provider is available upon the static method call by upper layers ( cuvs-java --> cuvs-lucene --> Solr). 

cuvs-lucene is doing a similar explicit load in https://github.com/NVIDIA/cuvs-lucene/blob/main/src/main/java/com/nvidia/cuvs/lucene/Utils.java#L145 

**Note**: Solr 10 uses lucene-cuvs version 25.10.0  which is where I observed the issue. I am working on tests to see if the issue also reproduces on cuvs main. But I see that the code for that part has not changed so in all likelihood the issue is still relevant.

rahulgoswami:
Fixes: https://github.com/NVIDIA/cuvs/issues/2300

Issue happens because libcudart.so (a dependency of libcuvs.so (or libcuvs_c.so?)) is not available during the class loading sequence unless explicitly loaded via System.loadLibrary() call  . The symbol lookup space (where Java looks for dependent native libraries) gets defined in one of the header class files (To-Do: add exact header) that gets generated as a Panama FFM binding. And this loading sequence looks for libcuvs.so but not for libcudart.so.

As opposed to the initial proposal in the GH issue of making the loadLibrary() call in CvVSProvider.provider() call, Claude suggested it might be better to introduce it in the library load sequence in NativeDependencyLoader.java instead and I went with that approach.

To-Do: Write up a more detailed description of the root cause and the solution.


rahulgoswami:
Thanks @cjnolet . I have updated the branch to the current main and run the pre-commit hook for formatting changes.

cjnolet (approved):

PR #1846, opened 2026-02-24
[REVIEW] New Dataset API Clarifying Ownership

HowardHuang1:
### Overview

Addressing #1574 and #1571. 

Replaced strided_dataset with padded_dataset class. Added support all the way up to CAGRA code. 

### Old class structure (Classes + Inheritance):

Screenshot 2026-06-23 at 7 06 37 PM

### New Class Structure (ContainerType Tags + Composition):
Screenshot 2026-06-23 at 7 07 23 PM

Inheritance is removed entirely and all dataset types are on the same level of the inheritance tree.

3 Levels:
1) Storage
Screenshot 2026-07-06 at 11 03 41 AM

2) Container tags
Screenshot 2026-07-06 at 11 03 11 AM

3) Public dataset/view aliases
Screenshot 2026-06-23 at 7 02 34 PM

### Ownership
The index and cagra::build / cagra::index do not own raw vector storage, they only take views. 
- callers (or the C merged holder) must keep backing memory alive for as long as the index is used.

The old code had a type-erased std::unique_ptr<dataset_view<...>>, i.e. non-owning view handles. The new code uses templates on the index type which determines the type of dataset_view the index holds.

### ACE v.s. non-ACE paths on Host

ACE path copies datasets that can't entirely fit in CPU memory in chunks onto GPU memory by calling make_padded_dataset. This is 1x memory on CPU and 1x memory on GPU.

### Return types:
Used mainly to maintain lifetime of dataset.

cuvs_cagra_c_api_lifetime_holder
- unique_ptr<vpq_dataset> vpq_owner
- unique_ptr<dataset> padded_dataset_owner
- raft::device_matrix dataset
- cagra::index idx
It is a single C++ struct in cagra.cpp that groups the real cagra::index with any extra heap-owned things the C API had to create so the index’s non-owning views stay valid. 

### Miscellaneous: Extend Serialize Deserialize
Will fill in later

### Factories:
- make_device_padded_dataset_view
- make_host_padded_dataset_view
- make_device_padded_dataset
- make_host_padded_dataset
- make_vpq_dataset 
    - in pq.hpp and pq.cu
- make_merged_dataset
    - in cagra.hpp

### Places where make_padded_dataset/view are called internally (not by user):
Host non-ACE path
- cpp/src/neighbors/cagra_build_inst.cu.in
- cagra_from_host_padded in cpp/src/neighbors/iface/iface.hpp
- c/src/neighbors/cagra.cpp

Tiered CAGRA
[...]

PR #2384, opened 2026-07-31
Fix benchmark regressions from #1846

achirkin:
Fix a few issues introduced to the benchmark executables limiting the functionality:

 - Reinstate the dropped support for CAGRA-Q (compression_* parameters)
 - Avoid multiple copies of the data and the graph on device in various code paths 
 - Avoid copying the data if it is device-accessible
 - Make logical merge reuse one buffer slot per split and free it before reallocating, instead of appending new split copies on every set_search_dataset call.
 - Make multi-GPU CAGRA benchmarks use the padded index and dataset view, which CAGRA search requires.
 - Revert copy() to the shallow copy-constructor form used before, so all shared state is forwarded to per-thread wrappers.

Library changes (non-breaking):
- Modified the implementation of `cuvs::make_vpq_dataset` to bring the ability to quantize from host data without copying the whole dataset and to quantize from fp16 data (the relevant symbols were hidden).


cjnolet:
@achirkin its too late for anything to be merged into 26.08. You'll need to retarget to main.

achirkin:
Sure. I'll just finish it up today and let you decide whether to re-target and whether to optionally backport the fixes later if necessary.

cpp/bench/ann/src/cuvs/cuvs_cagra_wrapper.h
```diff
@@ -65,6 +66,69 @@ inline void maybe_log_cagra_persistent_concurrency_hint(bool persistent_search)
...
+    buffers.emplace_back(raft::make_device_matrix<T, int64_t>(res, 0, 0));
+  }
+}
+
+/**
+ * Turn a host-resident index into a searchable device index carrying only the graph.
+ *
+ * The graph is already in device memory, owned by `host_index`; the returned index only views it,
+ * so the caller must keep `host_index` alive. The dataset is attached later, by `set_search_param`
+ * or `set_search_dataset`.
```
tfeher, line 115: Could you explain why is this function needed? What does it mean "host_index"? An index that stores the graph in host memory? But the description says "The graph is already in device memory"

cpp/bench/ann/src/cuvs/cuvs_cagra_wrapper.h
```diff
@@ -210,200 +282,83 @@ class cuvs_cagra : public algo<T>, public algo_gpu {
...
-      *dataset_ = std::move(padded->data_);
-      index_    = std::make_shared<index_type>(std::move(ace_index));
+    if (dataset_is_on_host) {
+      // Construct the graph straight from host memory: cagra::build streams the rows in batches
+      // (and dispatches to ACE when it is configured), so the dataset is not uploaded here at all.
+      // The single device copy needed for search is made later, by set_search_param.
+      host_index_ = std::make_shared<host_index_type>(cuvs::neighbors::cagra::build(
+        handle_, host_params, cuvs::neighbors::make_host_standard_dataset_view(dataset_view_host)));
+      index_ =
+        std::make_shared<index_type>(detail::to_graph_only_index<T, IdxT>(handle_, *host_index_));
```
tfeher, line 306: What is the differnece between` host_index_ `and `index_`? Isn't `host_index_` already  a graph_only index? (due to `add_dataset_on_build=false`?

cpp/bench/ann/src/cuvs/cuvs_cagra_wrapper.h
```diff
@@ -417,14 +372,57 @@ void cuvs_cagra<T, IdxT>::build(const T* dataset, size_t nrow)
...
+               "cagra: compression_* (CAGRA-Q) requires the L2Expanded metric.");
+  auto rows = static_cast<int64_t>(nrow);
+  // The codebooks are trained on the device; upload the rows first if the GPU cannot read them.
+  auto staging = raft::make_device_matrix<T, int64_t>(handle_, 0, 0);
+  auto src = raft::make_device_matrix_view<const T, int64_t, raft::row_major>(dataset, rows, dim_);
+  if (!detail::is_device_accessible(dataset)) {
+    staging = raft::make_device_matrix<T, int64_t>(handle_, rows, dim_);
+    raft::copy(staging.data_handle(),
+               dataset,
+               static_cast<size_t>(rows) * dim_,
```
tfeher: If this is called on the whole dataset, then we might run OOM. We should have a vpq quantization method that takes host input data.

tfeher (dismissed):
Thanks Artem, most of it looks good, but I have some questions.

cpp/bench/ann/src/cuvs/cuvs_cagra_wrapper.h
achirkin: That means the new `cuvs::neighbors::cagra::host_standard_index<T, IdxT>` type - so it assumes the dataset is on the host (template parameter), while the graph can still be anything.

cpp/bench/ann/src/cuvs/cuvs_cagra_wrapper.h
achirkin: host_index_ (host_standard_index) is introduced as a workaround in this PR alongside index_ (device_padded_index, searchable) to support various input dataset types with minimum number of copies.

cpp/bench/ann/src/cuvs/cuvs_cagra_wrapper.h
cjnolet: This is the result of the dataset type being added to the template. Ive already talked to @HowardHuang1  and expressed my preference for a different naming here. We are going to work on improving this in 26.10 for sure.

cjnolet (dismissed):

cpp/bench/ann/src/cuvs/cuvs_cagra_wrapper.h
achirkin: Fixed by restoring the from-host functionality in the library.

cjnolet (approved):

PR #2412, opened 2026-08-04
Factor CAGRA graph kernels into shared translation unit

landrumb:
The graph merge PR (#2352) uses some of the kernels defined by CAGRA, and binary size takes a hit from them not appearing in a TU where they can be shared by both CAGRA and Fastener.

### DIFF
--- a/cpp/src/neighbors/detail/cagra/graph_shared.cuh
+++ b/cpp/src/neighbors/detail/cagra/graph_shared.cuh
@@ -1,0 +1,46 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+#pragma once
+
+#include <cuvs/core/export.hpp>
+#include <cuvs/distance/distance.hpp>
+
+#include <raft/core/device_mdspan.hpp>
+#include <raft/core/resources.hpp>
+
+#include <cuda_fp16.h>
+
+#include <cstdint>
+
+namespace cuvs::neighbors::cagra::detail::graph {
+
+// kern_sort capacity is WarpSize * numElementsPerThread; largest specialization uses 32.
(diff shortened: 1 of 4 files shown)


landrumb (commented):
/ok to test

cpp/src/neighbors/detail/cagra/graph_shared.cuh
```diff
@@ -0,0 +1,46 @@
...
+
+#include <cstdint>
+
+namespace cuvs::neighbors::cagra::detail::graph {
+
+// kern_sort capacity is WarpSize * numElementsPerThread; largest specialization uses 32.
+inline constexpr uint64_t kMaxSortDegree = 32 * 32;
+
+#define CUVS_DECL_CAGRA_GRAPH_SORT(DataT)                              \
+  CUVS_EXPORT void launch_sort_knn_graph(raft::resources const& res,   \
```
divyegala: I do not thing the export is needed as it's an internal function only, so we can just hide the symbol within the .so

cpp/src/neighbors/detail/cagra/graph_shared.cuh
```diff
@@ -0,0 +1,46 @@
...
+CUVS_DECL_CAGRA_GRAPH_SORT(float);
+CUVS_DECL_CAGRA_GRAPH_SORT(half);
+CUVS_DECL_CAGRA_GRAPH_SORT(int8_t);
+CUVS_DECL_CAGRA_GRAPH_SORT(uint8_t);
+
+#undef CUVS_DECL_CAGRA_GRAPH_SORT
+
+/** Run the existing CAGRA optimizer through one compiled instantiation instead of rematerializing
+ *  its reverse-graph, prune, merge, and MST kernels in every Fastener dtype TU. */
+CUVS_EXPORT void optimize_device_graph(
```
divyegala: Same as above

divyegala (approved):
Just one suggestion, thanks!

Issue #2405, opened 2026-08-03
[FEA][C API][Language Wrappers] Recover Cagra-Q path in C API and downstream language wrappers


HowardHuang1:
- Cagra-Q path in C API layer and downstream language wrappers was accidentally disabled when compression params was removed.
- Need to expose make_vpq_dataset() factory + rebind compressed dataset on uncompressed index functions at C API and downstream language wrappers. C++ API already has this.

PR #2413, opened 2026-08-04
Recover cagra-q path in C API and downstream language wrappers

HowardHuang1:
Addresses issue #2405.


c/include/cuvs/neighbors/cagra.h
```diff
@@ -255,6 +255,24 @@ CUVS_EXPORT cuvsError_t cuvsCagraCompressionParamsCreate(cuvsCagraCompressionPar
...
+ * `cuvsCagraUpdateDataset`. Caller owns the returned dataset and must keep it alive while any
+ * index uses it. Metric for subsequent search must remain `L2Expanded`.
+ *
+ * @param[in] res cuvs resources
+ * @param[in] source_dataset device-padded dataset (owning or view)
+ * @param[in] params VPQ compression params; NULL selects defaults
+ * @param[out] vpq_dataset newly allocated owning VPQ dataset handle
+ * @return cuvsError_t
+ */
+CUVS_EXPORT cuvsError_t cuvsDatasetMakeVpq(cuvsResources_t res,
```
cjnolet: Let's drop the V from this and call it `cuvsDatasetMakePq`.

c/include/cuvs/neighbors/cagra.h
```diff
@@ -580,21 +598,25 @@ CUVS_EXPORT cuvsError_t cuvsCagraIndexGetDataset(cuvsCagraIndex_t index, DLManag
...
- * the index is converted and its opaque handle is rebound to a search-ready
- * device-padded index. Caller retains ownership of
- * \p device_padded_dataset and must keep it alive while \p index uses it.
+ * Caller retains ownership of \p dataset and must keep it alive while \p index uses it.
  *
- * @param[in] res             cuvsResources_t opaque C handle
- * @param[in] device_padded_dataset owning or non-owning device-padded dataset handle
- * @param[inout] index        CAGRA index handle
+ * @param[in] res      cuvsResources_t opaque C handle
+ * @param[in] dataset  device-padded or owning device VPQ_F16 dataset handle
```
cjnolet: Hmm. I don't know that i like this naming. VPQ_F16... that's cryptic... and why only F16? Is that really the only data type that's supported?

Let's keep the naming contention here: `device-padded or `pq`. Also- there's no reason we should need to specify "owning device" here... that should be opaque to the user (they just create a dataset in the c layer and they pass it in and we worry about the view creation under the hood).

cpp/include/cuvs/neighbors/cagra.hpp
```diff
@@ -4598,6 +4634,35 @@ auto attach_dataset(
...
+ * @brief Attach a device VPQ (f16 codebook) dataset and return a search-ready `vpq_f16_index`.
+ *
+ * Builds a new VPQ-typed index by copying the graph from `idx`. Caller owns `vpq_dataset` storage.
+ *
+ * If `idx` is already a `vpq_f16_index`, call `idx.update_device_dataset_same_layout(res,
+ * vpq_dataset)` directly.
+ */
+template <typename T, typename IdxT, typename IndexViewT>
+  requires cuvs::neighbors::ann_dataset_view<IndexViewT>
+auto attach_dataset(raft::resources const& res,
```
cjnolet: Please stop naming this attach_dataset. The only function that should be exposed to the outside world is `update_dataset`. Since we're not rushing to get this into 26.08, we need to fix this.

cpp/include/cuvs/preprocessing/quantize/pq.hpp
```diff
@@ -267,32 +270,33 @@ template <typename T>
...
+ * auto vpq = cuvs::preprocessing::quantize::pq::make_device_vpq_dataset(res, vpq_params,
+ * padded.view()); auto vpq_idx = cuvs::neighbors::cagra::attach_dataset(res, idx,
+ * vpq.as_dataset_view());
  * @endcode
  */
 template <typename SrcT>
-[[nodiscard]] auto make_vpq_dataset(raft::resources const& res,
-                                    cuvs::neighbors::vpq_params const& params,
-                                    SrcT const& src)
+[[nodiscard]] auto make_device_vpq_dataset(raft::resources const& res,
```
cjnolet: Please rename- `pq_dataset`. We don't need the `vpq` in the name.

go/cagra/cagra.go
```diff
@@ -21,11 +21,22 @@ type PaddedDataset struct {
 	dataset C.cuvsDataset_t
 }
 
+// Owning VPQ dataset handle for CAGRA-Q search.
+type VpqDataset struct {
```
cjnolet: Why is this a separate struct? Also.. please rename ALL instances of `vpq_dataset` to `pq_dataset`.

cjnolet: It was a mistake that it was ever named vpq to begin with- I understand it does a `vector quant` on top of the pq, but that's a feature, not a defining name.

cpp/include/cuvs/neighbors/cagra.hpp
```diff
@@ -4548,6 +4548,42 @@ auto convert_standard_to_padded_index(
...
 
+/**
+ * @brief Convert a dense (non-VPQ) CAGRA index into a VPQ f16 index and attach a VPQ dataset.
+ *
+ * Copies graph/source-indices from `src` into a new `vpq_f16_index` and attaches `vpq_dataset`.
+ * Caller owns `vpq_dataset` storage for the lifetime of the returned index.
+ */
+template <typename T, typename IdxT, typename IndexViewT>
+  requires cuvs::neighbors::ann_dataset_view<IndexViewT>
+auto convert_dense_to_vpq_f16_index(
```
aamijar: This function seems unnecessary. We shouldn't be copying the cagra graph just to change the template type.
I am making the change to `update_dataset` to do the move instead of copy. We should be able to reuse `update_dataset` for this.

c/tests/neighbors/ann_cagra_c.cu
```diff
@@ -2008,3 +2009,109 @@ TEST(CagraC, SearchMultiPartitionMultiKernelRejected)
   }
   cuvsResourcesDestroy(res);
 }
+
+TEST(CagraC, BuildAttachVpqSearch)
+{
+  // CAGRA-Q smoke test: dense build → MakeVpq → UpdateDataset(VPQ) → Search.
```
aamijar: Thanks for adding the tests to C api and language wrappers. I don't see tests in cpp though. We should add that.

c/include/cuvs/core/dataset.h
```diff
@@ -20,7 +20,9 @@ extern "C" {
  */
 typedef enum {
   CUVS_DATASET_LAYOUT_STANDARD = 0,
-  CUVS_DATASET_LAYOUT_PADDED   = 1
+  CUVS_DATASET_LAYOUT_PADDED   = 1,
+  /** Device VPQ storage with f16 codebooks (CAGRA-Q search dataset). */
+  CUVS_DATASET_LAYOUT_VPQ_F16  = 2
```
aamijar:
```suggestion
  CUVS_DATASET_LAYOUT_PQ  = 2
```

python/cuvs/cuvs/tests/test_cagra.py
```diff
@@ -226,6 +226,36 @@ def test_cagra_build_from_dataset_handle(
...
+    dataset = generate_data((n_rows, n_cols), np.float32)
+    dataset_device = device_ndarray(dataset)
+
+    index = cagra.build(
+        cagra.IndexParams(metric="sqeuclidean"),
+        dataset_device,
+    )
+    padded = make_device_padded_dataset(dataset_device)
+    compression = cagra.CompressionParams(pq_bits=8, pq_dim=8)
+    vpq = cagra.make_vpq_dataset(padded, compression_params=compression)
```
aamijar: Are we saying the user needs to do `make_device_padded_dataset` before `make_vpq_dataset`? Can you just call `make_vpq_dataset` by itself?

aamijar (changes requested):
Adding my review below:

aamijar:
This PR is marked a a breaking change, can you clarify what is breaking here?

c/include/cuvs/neighbors/cagra.h
HowardHuang1: VPQ_F32 is not supported right now in the cagra-q search kernels. 

Source vectors can be f32/f16/int8/uint8 but the codebook math type stored for search is forced to half.

compute_distance_vpq-impl.cuh has this line: 
`static_assert(std::is_same_v<CODE_BOOK_T, half>, "Only CODE_BOOK_T = `half` is supported now");`

c/include/cuvs/neighbors/cagra.h
HowardHuang1: Just wanted to clarify my understanding of make_vpq_dataset() factory. There is a PQ step and a VQ step but the VQ step is optional and run only when the flag bool use_vq = true. Is that why we are renaming the make_vpq_dataset() factory to make_pq_dataset()? So we will have pq dataset, rabitq dataset, bbq dataset, and sq dataset but no standalone vpq dataset? We will treat vpq as a special condition of pq dataset?

cjnolet: The VQ is not a proper VQ. It's misnamed. It's coarse centroid assignments. But we don't need to include that in the name.

cpp/include/cuvs/preprocessing/quantize/pq.hpp
```diff
@@ -267,32 +270,33 @@ template <typename T>
...
+                                           cuvs::neighbors::vpq_params const& params,
+                                           SrcT const& src)
   -> cuvs::neighbors::device_vpq_dataset<half, int64_t>
```
achirkin: Please incorporate the relevant changes from https://github.com/NVIDIA/cuvs/pull/2384 (if it doesn't get in first) to bring back the functionality to quantize host data and fp16 data.

achirkin: NB: the #2384 is merged, so feel free to adjust if required (but please ensure both device and host code paths are covered here and in the benchmarks).

c/include/cuvs/neighbors/cagra.h
HowardHuang1: @achirkin's new PR #2384 may have fixed this to support VPQ_F32. Could you confirm @achirkin ?

cpp/include/cuvs/neighbors/cagra.hpp
HowardHuang1: This attach_dataset() is not meant to be called publicly at the C++ API layer. It is an internal helper that is called within the user-callable public facing update_dataset() which is the single entry point for users to update/attach any sort of dataset. C API and language wrappers already have this implementation but due to mutability guarantees in C++, we didn't get to implementing a unified update_dataset() at the C++ API layer in 26.08.

Am planning to unify it under a single public facing update_dataset() at C++ API layer with separate overloads with different mutability guarantees wrapping attach_dataset() and update_device_dataset_same_layout() in a separate PR.

cjnolet: If it's not meant to be called publicly then put it in src/

We also should not have templates for free functions in header files in include

HowardHuang1: Got it, will address this in a separate PR!

cjnolet: Please address it here . This is being introduced in this PR. I would prefer to fix it here.

cpp/include/cuvs/neighbors/cagra.hpp
cjnolet: This should not be in the include directory.

cjnolet (changes requested):

c/include/cuvs/neighbors/cagra.h
achirkin: No, we only support FP16 codebooks. I think it could make sense to add 8-bit support for improved performance, but FP32 codebooks are not very practical (we loose more on PQ encoding than FP32-precision of the residuals can likely yield).

cpp/include/cuvs/neighbors/cagra.hpp
aamijar: Hi @HowardHuang1, now that https://github.com/NVIDIA/cuvs/pull/2427 has been merged, you should be able to simplify a lot of this PR.

c/include/cuvs/neighbors/cagra.h
```diff
@@ -255,6 +255,24 @@ CUVS_EXPORT cuvsError_t cuvsCagraCompressionParamsCreate(cuvsCagraCompressionPar
  */
 CUVS_EXPORT cuvsError_t cuvsCagraCompressionParamsDestroy(cuvsCagraCompressionParams_t params);
 
+/**
+ * @brief Train an owning device VPQ (f16 codebook) dataset from a device-padded source.
```
cjnolet: Please please please stop using VPQ. That is not a valid term. Do not use it anywhere.

cpp/include/cuvs/neighbors/cagra.hpp
HowardHuang1: Thanks Anupam!

c/include/cuvs/core/dataset.h
```diff
@@ -20,7 +20,9 @@ extern "C" {
  */
 typedef enum {
   CUVS_DATASET_LAYOUT_STANDARD = 0,
-  CUVS_DATASET_LAYOUT_PADDED   = 1
+  CUVS_DATASET_LAYOUT_PADDED   = 1,
+  /** Device PQ storage with f16 codebooks (CAGRA-Q search dataset). */
+  CUVS_DATASET_LAYOUT_PQ_F16 = 2
```
cjnolet: I'm honestly still not sure I'm on board w/ this naming. I don't think we need the F16 here, do we?

c/include/cuvs/neighbors/cagra.h
```diff
@@ -282,6 +282,24 @@ CUVS_EXPORT cuvsError_t cuvsCagraCompressionParamsCreate(cuvsCagraCompressionPar
  */
 CUVS_EXPORT cuvsError_t cuvsCagraCompressionParamsDestroy(cuvsCagraCompressionParams_t params);
 
+/**
+ * @brief Train an owning device PQ (f16 codebook) dataset from a device-padded source.
```
cjnolet: Let's remove the f16. That seems like an implementation detail.

c/include/cuvs/neighbors/cagra.h
```diff
@@ -282,6 +282,24 @@ CUVS_EXPORT cuvsError_t cuvsCagraCompressionParamsCreate(cuvsCagraCompressionPar
...
+ *
+ * @param[in] res cuvs resources
+ * @param[in] source_dataset device-padded dataset (owning or view)
+ * @param[in] params PQ compression params; NULL selects defaults
+ * @param[out] pq_dataset newly allocated owning PQ dataset handle
+ * @return cuvsError_t
+ */
+CUVS_EXPORT cuvsError_t cuvsDatasetMakePq(cuvsResources_t res,
+                                          cuvsDataset_t source_dataset,
+                                          cuvsCagraCompressionParams_t params,
```
cjnolet: This seems a bit odd- How do these params differ from `PqParams`?  Also, why is this inside cagra.h? This should be inside `preprocessing/pq.h`

I've mentioned many times before that the datasets should not be coupled to index types.

c/src/neighbors/cagra.cpp
```diff
@@ -52,7 +52,13 @@ struct cuvs_cagra_c_api_index_lifetime_holder {
 /** Owns how to delete co-located index storage; `cuvsCagraIndex::addr` points here. */
 struct sg_cagra_c_api_index_box {
   void* index_ptr;
-  enum class dataset_layout : uint8_t { device_padded, device_standard, host_padded, host_standard } layout;
+  enum class dataset_layout : uint8_t {
+    device_padded,
+    device_standard,
+    host_padded,
+    host_standard,
+    device_pq_f16
```
cjnolet: Please remove the f16 from all naming. This info should be carried along in config info, not in the name of the type itself.

cjnolet (changes requested):

c/src/neighbors/cagra.cpp
```diff
@@ -529,37 +544,44 @@ static void make_host_standard_dataset_view(raft::resources*,
...
-          cuvs::neighbors::cagra::update_dataset(*res_ptr, std::move(idx), padded_view);
-        auto* holder =
-          new cuvs_cagra_c_api_index_lifetime_holder<T, view_t>{std::move(padded_idx)};
-        destroy_sg_cagra_c_api_box(index->addr);
-        index->addr = 0;
-        bind_index_lifetime_holder_to_C_index<T, view_t>(index, index->dtype, holder);
-      });
+  with_dataset_view<owner_t, view_t>(source_dataset, [&](auto const& padded_view) {
+    auto pq =
+      cuvs::preprocessing::quantize::pq::make_device_pq_dataset(*res_ptr, ps, padded_view.view());
```
aamijar, line 573: Calling .view() on padded_view is removing the "dataset wrapper" and giving us the underlying mdspan storage.
This will cause a bug. Look at this branch here. What will happen is that we will not enter the if branch (raft::mdspan doesn't have .view() or .dim()) and we won't make the copy from padded to contiguous dense (standard). See some related issue I opened here https://github.com/NVIDIA/cuvs/issues/2567 and https://github.com/NVIDIA/cuvs/issues/2569.

https://github.com/NVIDIA/cuvs/blob/7886253f1c37d23ab3e4abab5857beaa28d2261c/cpp/include/cuvs/preprocessing/quantize/pq.hpp#L306-L332

```suggestion
      cuvs::preprocessing::quantize::pq::make_device_pq_dataset(*res_ptr, ps, padded_view);
```

aamijar (changes requested):

HowardHuang1:
/ok to test [b0542ae](https://github.com/NVIDIA/cuvs/pull/2413/commits/b0542aeb9d78dc541b52f871e409084104666ef2)

HowardHuang1:
/ok to test [d4fe20f](https://github.com/NVIDIA/cuvs/pull/2413/commits/d4fe20fe1cd49cb015d0d2da20c540216e8280f1)

HowardHuang1:
/ok to test [a59d945](https://github.com/NVIDIA/cuvs/pull/2413/commits/a59d94526c3060e85c7d7a3dc1eb4f75cbc78548)

c/include/cuvs/preprocessing/quantize/pq.h
```diff
@@ -1,12 +1,14 @@
...
+ * SPDX-FileCopyrightText: Copyright (c) 2025-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
  * SPDX-License-Identifier: Apache-2.0
  */
 
 #pragma once
 
 #include <cuvs/cluster/kmeans.h>
 #include <cuvs/core/c_api.h>
+#include <cuvs/core/dataset.h>
+#include <cuvs/neighbors/cagra.h>
```
cjnolet: Why are you bringing CAGRA dependencies into the preprocessing namespace? Please remove this

c/include/cuvs/preprocessing/quantize/pq.h
```diff
@@ -221,6 +223,24 @@ CUVS_EXPORT cuvsError_t cuvsProductQuantizerGetEncodedDim(cuvsProductQuantizer_t
...
+ *
+ * @param[in] res cuvs resources
+ * @param[in] source_dataset device-padded dataset (owning or view)
+ * @param[in] params PQ compression params; NULL selects defaults
+ * @param[out] pq_dataset newly allocated owning PQ dataset handle
+ * @return cuvsError_t
+ */
+CUVS_EXPORT cuvsError_t cuvsDatasetMakePq(cuvsResources_t res,
+                                          cuvsDataset_t source_dataset,
+                                          cuvsCagraCompressionParams_t params,
```
cjnolet: Again- why? This is cagra-specific, has nothing to do with pq....

c/src/neighbors/cagra.cpp
cjnolet: A view of a dataset should be a dataset_view, not an mdspan. Let's just be very very clear about that- a dataset view should ALWAYS be its own construct and should never be simplified to just an mdspan.

HowardHuang1:
/ok to test [07c209e](https://github.com/NVIDIA/cuvs/pull/2413/commits/07c209e8a8bf0f7c1da1b9abfabae772c1711d96)

HowardHuang1:
/ok to test [aaf7b5c](https://github.com/NVIDIA/cuvs/pull/2413/commits/aaf7b5cd82e211ff831e1d87d8b2023b1748c85a)

HowardHuang1:
/ok to test [71bdef3](https://github.com/NVIDIA/cuvs/pull/2413/commits/71bdef3234d5b51cc99883ce26e6568d97960b4c)

HowardHuang1:
/ok to test [cc0742a](https://github.com/NVIDIA/cuvs/pull/2413/commits/cc0742ab18960d85ac58b50e9fd1fbe3a79288c0)

PR #2416, opened 2026-08-05
Fix CAGRA-ACE regressions from #1846

julianmi:
Fix CAGRA-ACE after the dataset-view changes in #1846. The change preserves the typed host dataset and its logical dimension, restores batched host partition builds without pre-materializing an extra device-padded partition, and attaches the original non-owning dataset view to in-memory indexes when requested.

Addresses https://github.com/NVIDIA/cuvs/pull/1846#discussion_r3685764576. CC @tfeher


tfeher (approved):
Thanks @julianmi for the fix, LGTM!

cpp/src/neighbors/detail/cagra/cagra_build.cuh
```diff
@@ -1402,12 +1423,14 @@ auto build_ace(raft::resources const& res,
...
-      auto sub_dataset_dev = cuvs::neighbors::make_device_padded_dataset(
-        res, raft::make_const_mdspan(sub_dataset.view()));
-      auto sub_index = ::cuvs::neighbors::cagra::detail::build_from_device_matrix<T, IdxT>(
-        res, sub_index_params, sub_dataset_dev->as_dataset_view());
+      // Keep the partition host-resident so IVF-PQ and NN-descent can consume it in batches.
+      // Iterative CAGRA uploads and pads the partition inside build_from_host_matrix.
+      auto sub_dataset_view = cuvs::neighbors::make_host_standard_dataset_view(
+        raft::make_const_mdspan(sub_dataset.view()));
+      auto sub_index = ::cuvs::neighbors::cagra::detail::build_from_host_matrix<T, IdxT>(
+        res, sub_index_params, sub_dataset_view);
```
aamijar: Are we able to use the public API cagra::build instead of the detail namespace?

aamijar (changes requested):

cpp/src/neighbors/detail/cagra/cagra_build.cuh
julianmi: Thanks, I switched to the public API.

aamijar (approved):

cjnolet (approved):

PR #2419, opened 2026-08-06
[DO NOT MERGE] kNN Decoding method for synthetic data generator

### LINKED ISSUE #2225: [FEA] BSDG: Investigate build times of synthetic data <-> real data
BSDG (Billion Scale Data Generator)

Need to compare the index build times of synthetic data <-> real data

jinsolp:
The kNN Decoding method for matching the build time & search behavior for the synthetic data generator.
Should not be merged. PR made for future bookkeeping purposes. Putting it in draft because it doesn't need reviews at the moment.

Eventually this should be blended in to the code in `cuvs_bench.synthesize_datasets`

Closes https://github.com/NVIDIA/cuvs/issues/2225


python/cuvs_bench/cuvs_bench/synthesize_dataset/knn_decode_synthesizer/build_vamana.py
```diff
@@ -0,0 +1,147 @@
+#!/usr/bin/env python
```
jinsolp: This is just a helper file I used to call vamana build with timing

PR #2420, opened 2026-08-06
fix(update-version): don't blanket replace the word main with release branch in README

gforsyth:
Hotfix for 26.08 release.

2 weeks ago this commit NVIDIA/cuvs@dd84ab5#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R159-R167 added some helpful new documentation to the `README.md` in cuvs.

It's a rust example and so it uses fn main() in the example code. update-version.sh looks everywhere for branch references that should be updated to release/26.08 and it did that, which then broke the rust doctests.

So, we fix update-version to be a little more measured in its replacements, then we have to cut a hotfix release on top of 26.08 for cuvs to get the rust artifacts published.


gforsyth:
I'm going to admin merge this after the rust builds complete

gforsyth:
Merging

PR #2421, opened 2026-08-06
fix(update-version): don't blanket replace the word main with release branch in README.md

gforsyth:
Hotfix for 26.08 release.

2 weeks ago this commit NVIDIA/cuvs@dd84ab5#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R159-R167 added some helpful new documentation to the `README.md` in cuvs.

It's a rust example and so it uses fn main() in the example code. update-version.sh looks everywhere for branch references that should be updated to release/26.08 and it did that, which then broke the rust doctests.

So, we fix update-version to be a little more measured in its replacements, then we have to cut a hotfix release on top of 26.08 for cuvs to get the rust artifacts published.

This PR is for making sure this commit also lands in `main` so this doesn't happen again.

xref #2420

### DIFF
--- a/ci/release/update-version.sh
+++ b/ci/release/update-version.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
-# SPDX-FileCopyrightText: Copyright (c) 2020-2026, NVIDIA CORPORATION.
+# SPDX-FileCopyrightText: Copyright (c) 2020-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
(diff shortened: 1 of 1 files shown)


cjnolet (approved):

jameslamb (approved):
oy

cjnolet (approved):

jameslamb:
This was failing the "Recently Updated" check but since it only touches `update-version.sh`, I'm confident in just admin-merging it.

PR #2423, opened 2026-08-06
[REVIEW] Add device-memory overload for cagra::helpers::optimize

mnorris11:
Disclaimer: AI mostly generated, but validated on a 8x H100 host. Let me know if I should delete the lengthy AI comments or if this PR is silly and there's better workarounds.  We will add a hack in Faiss at https://github.com/facebookresearch/faiss/pull/5500 until we can merge this PR / import the latest cuVS version (probably around 26.10? or later?).

1. Please write a description in this text box of the changes that are being
   made.

Problem
--
`cagra::helpers::optimize` currently accepts host matrices only, so a caller that already holds its k-NN graph in device memory -- for example the output of `all_neighbors::build` -- must copy the graph to host, optimize, and copy back.

The device implementation already exists. `graph::optimize` is templated on both mdspan accessors and `make_reverse_graph_gpu` has an `is_device_accessible` fast path (graph_core.cuh:826), and `batch_load_iterator` switches to a zero-copy `kPassthrough` mode for device accessors. That code is unreachable today because `detail::optimize` in cagra_build.cuh erases the caller's accessor:

    using g_accessor_internal =
      raft::host_device_accessor<cuda::std::default_accessor<internal_IdxT>,
                                 raft::memory_type::host>;

so `graph::optimize` is only ever instantiated with host accessors. With host accessors the reverse-graph phase degrades into `graph_degree` separate host column gathers, each with its own H2D copy and a full stream synchronisation.

This change
--

- propagates the caller's memory types through `detail::optimize` instead of erasing them, and makes `new_graph` accessor-generic there and in `cagra::optimize`;
- adds a `device_matrix_view` overload of `cagra::helpers::optimize` to the public API;
- exposes the existing `guarantee_connectivity` flag on the public overloads, which previously could not be reached from outside.

Both public overloads keep their existing signatures via a defaulted argument, so this is source compatible.

Measured on 100M x 128d vectors (graph degree 32, intermediate degree 32) on 8x H100, as part of a multi-GPU CAGRA build: the optimize step goes from 119.1s to 1.65s, a 72x reduction, with recall unchanged. 

2. Please ensure that you have written units tests for the changes made/features
   added.

Tests: adds a device-to-device case and a case asserting the device overload produces the same graph as the host overload for the same input.

3. If you are closing an issue please use one of the automatic closing words as
   noted here: https://help.github.com/articles/closing-issues-using-keywords/

None I think?


cpp/include/cuvs/neighbors/cagra.hpp
```diff
@@ -4656,10 +4656,45 @@ std::pair<size_t, size_t> cagra_build_mem_usage(raft::resources const& res,
...
+ * @brief Optimize a KNN graph into a CAGRA graph without leaving device memory.
+ *
+ * Same as the host overload, but both graphs stay in device memory. Pruning,
+ * reverse-graph construction and the final merge all run on device with no host
+ * staging. This avoids copying the graph out and back, and avoids serialising the
+ * reverse-graph phase into `graph_degree` separate host gathers, each with its own
+ * H2D copy and stream synchronisation.
+ *
+ * Prefer this overload when the k-NN graph is already on device, for example the
+ * output of `all_neighbors::build`.
```
mfoerste4: This is indeed very verbose - please align this with the host overload description.

cpp/include/cuvs/neighbors/cagra.hpp
```diff
@@ -4656,10 +4656,45 @@ std::pair<size_t, size_t> cagra_build_mem_usage(raft::resources const& res,
...
  * @param[in] knn_graph Input KNN graph on host [n_rows, k_in]
  * @param[out] new_graph Output CAGRA graph on host [n_rows, k_out]
+ * @param[in] guarantee_connectivity Run the MST pass so the pruned graph is guaranteed
+ *            to be connected
  */
 void optimize(raft::resources const& handle,
               raft::host_matrix_view<uint32_t, int64_t, raft::row_major> knn_graph,
-              raft::host_matrix_view<uint32_t, int64_t, raft::row_major> new_graph);
+              raft::host_matrix_view<uint32_t, int64_t, raft::row_major> new_graph,
+              bool guarantee_connectivity = false);
```
mfoerste4, line 4798: Is this parameter really required for your workflow? The mst path should be functionally correct, but is not optimized wrt. speed and memory consumption. I would rather not see it in the public API. @achirkin , what is your opinion on this?

cpp/src/neighbors/detail/cagra/cagra_build.cuh
```diff
@@ -1941,22 +1941,34 @@ void build_knn_graph(
...
-    reinterpret_cast<internal_IdxT*>(new_graph.data_handle()),
-    new_graph.extent(0),
-    new_graph.extent(1));
-
+  // Propagate the caller's memory types rather than erasing them to
+  // `memory_type::host`. Erasing them makes the device-resident branches
+  // inside graph::optimize unreachable -- the `is_device_accessible` path in
+  // make_reverse_graph_gpu, and the zero-copy `kPassthrough` mode of
+  // batch_load_iterator -- so a caller that already holds the graph on device
+  // is forced through host staging.
```
mfoerste4: Comment can be removed

mfoerste4 (changes requested):
@mnorris11 , thanks for the PR. Besides the lengthy AI comments you already mentioned the PR looks good to me. I would like to wait for other peoples comments on exposing the mst graph option in its current state though. CC @cjnolet

cpp/include/cuvs/neighbors/cagra.hpp
achirkin: I don't think this parameter will be a big problem since the default is false, but I agree unless you sure you need this feature it will be easier for us to support the API without it.

cjnolet: I do absolutely want the guarantee_connectivity to be exposed to the public. The reason this parameter was introduced was not for testing, but rather because we came to realization that some types of neighborhood graphs, such as the RNG graph which underlines assumptions in Vamana, implicitly assume the graph is connected. This can obviously greatly impact the navigability since NSG-style graphs tend to start with an all-neighbors knn graph (which is not guaranteed to be connected).

cpp/src/neighbors/detail/cagra/cagra_build.cuh
mnorris11: Done!

cpp/include/cuvs/neighbors/cagra.hpp
mnorris11: Done!

mfoerste4 (approved):
LGTM!

cjnolet (approved):

tarang-jain (approved):

tarang-jain:
please run
`pre-commit run --all-files`
to fix the style errors

jamxia155:
/ok to test [d24d528](https://github.com/NVIDIA/cuvs/pull/2423/commits/d24d528b4f10d9e108cb8643fee694811186ef45)

jamxia155:
/ok to test [bca2484](https://github.com/NVIDIA/cuvs/pull/2423/commits/bca248442b6bae69cc13b8b4e5bb58d2cb6a33f7)

jamxia155:
/ok to test [4006696](https://github.com/NVIDIA/cuvs/pull/2423/commits/4006696cf1c7c1a01aa92304383cbba65fe81533)

jamxia155:
/ok to test [444794b](https://github.com/NVIDIA/cuvs/pull/2423/commits/444794bdd83508808ba6cdb2f97d0eee011e6f75)

jamxia155:
/ok to test [e072b5f](https://github.com/NVIDIA/cuvs/pull/2423/commits/e072b5f6c4c604966ba8d25c3ec92606bd9b65d3)

PR #2426, opened 2026-08-07
fix knn_merge_parts for inner product

qwertyforce:
By default, when merging the results of sharded multigpu index, knn_merge_parts keeps K smallest values. But inner product is not a distance, it is a measure of similarity. Therefore results are wrong for IP.
In this PR we are adding an additional overload, that receives an argument select_min, we keep backward compatibility and add a new test


divyegala (commented):
Hi @qwertyforce, thank you for the PR! I think this change will severely affect the binary size of cuVS, can you please provide a before-after measurement? Alternatively, we could run an element-wise operation to negate the inner product distances before running the merge kernel.

divyegala (approved):
Thanks for the contribution!

viclafargue (approved):
Great catch! Thank you for identifying the issue and following through with a PR to fix it.

@divyegala There indeed is an other independently existing issue in the direct merge path that would deserve some work. I will directly open up a PR to fix this.

**EDIT:** Here is the PR : https://github.com/NVIDIA/cuvs/pull/2441.

PR #2430, opened 2026-08-11
Enable security scanning via NVIDIA/security-workflows suite

gmanal:
Onboards cuVS to NVIDIA's centrally maintained security scanning via [`NVIDIA/security-workflows`](https://github.com/NVIDIA/security-workflows), pinned to [`v0.3.0`](https://github.com/NVIDIA/security-workflows/releases/tag/v0.3.0).
Two layers, no new tooling for this repo to maintain:
* **Local (advisory)** — `secret-scan-trufflehog` pre-commit hook, self-installing, pinned by tag like the repo's other hooks.
* **CI (enforcing)** — the security *suite*: one pinned reference fanning out to the Pulse secret scan (fails on verified/live secrets) and CodeQL SAST.

Both scans are enabled explicitly in the new workflow, and every scan in the suite is opt-in — scanners added upstream later won't switch themselves on here.

## Changes

* `.github/workflows/security-suite.yml` — on pushes to `main`, `release/*`, and copy-pr-bot `pull-request/[0-9]+`, matching `build.yaml`/`pr.yaml`. Uses this repo's permissions style (`permissions: {}` at workflow level, granted on the job) and the `linux-amd64-cpu4` label already used by `telemetry-summarize`.
* `.pre-commit-config.yaml` — add the hook, `rev: v0.3.0`.

**SAST scope: `python` + `actions`.** Both run under CodeQL `build-mode: none`, so no toolchain and no added build time. `actions` earns its place: 9 workflows, two on `pull_request_target`.

### How I tested this
* `pre-commit run` on both changed files — passes, including `zizmor`, `yamllint`, `verify-copyright`.
* `pre-commit run --all-files secret-scan-trufflehog` — clean across the tree. (`v0.3.0` excludes TruffleHog's Lob detector, which reads `test_` + 35 chars as an API key and flagged pytest names as *verified* secrets elsewhere — hence this pin, not an earlier one.)


.github/workflows/security-suite.yml
```diff
@@ -0,0 +1,46 @@
...
+    with:
+      enable-secret-scan: true
+      enable-sast-scan: true
+      secret-runs-on: linux-amd64-cpu4
+      # Set the policy explicitly so enforcement can't drift with upstream defaults.
+      #   unverified — fail on verified/live secrets; warn on unverified  [default]
+      #   strict     — fail on any finding (verified or unverified)
+      #   all        — warn only; never fail the job on findings
+      secret-failure-policy: unverified
+      sast-languages: '["python","actions"]'
```
jameslamb:
```suggestion
      sast-languages: '["actions","python"]'
```

Let's alphabetize this, please.

Also, I couldn't find documentation on the available options for this at https://github.com/NVIDIA/security-workflows ... are there more?

This repo has Bash, C++, Go, Java, and Rust code too.

.pre-commit-config.yaml
```diff
@@ -2,6 +2,13 @@
 # SPDX-License-Identifier: Apache-2.0
 
 repos:
+  # Runs first so a leaked credential blocks the commit before any formatter runs.
+  # Self-installing: pre-commit downloads a pinned, checksum-verified trufflehog into
+  # the hook environment on first use. CI enforces the same class of finding via Pulse.
+  - repo: https://github.com/NVIDIA/security-workflows
```
jameslamb, line 6:
```suggestion
  # Runs first so a leaked credential blocks the commit before any formatter runs.
  - repo: https://github.com/NVIDIA/security-workflows
```

Let's remove this comment that explains how `pre-commit` works, please.

.github/workflows/security-suite.yml
jameslamb, line 1: Please add this file to all the `changed-files` lists in https://github.com/NVIDIA/cuvs/blob/main/.github/workflows/pr.yaml (preserve alphabetical order).

We shouldn't need to trigger cuVS's expensive GPU-using CI on PRs that only change this file.

.github/workflows/security-suite.yml
```diff
@@ -0,0 +1,46 @@
+# SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+# SPDX-License-Identifier: Apache-2.0
+#
+# CI security scanning via the NVIDIA/security-workflows suite: Pulse secret scan + CodeQL SAST.
+# Pulse runs on Linux nv-gha-runners.
+# Pinned to security-workflows v0.3.0.
+
```
jameslamb:
```suggestion
```

Let's please remove these comments that don't add information (they are obvious from the code) and are likely to become out of date as this configuration changes over time.

jameslamb (changes requested):
Left some comments for your consideration.

.pre-commit-config.yaml
gmanal: done

.github/workflows/security-suite.yml
gmanal: done

.github/workflows/security-suite.yml
gmanal: done

.github/workflows/security-suite.yml
gmanal: added java and rust to the scan list.
Go can't be analyzed buildless and would force a build mode Rust rejects; C++ is mostly CUDA, which CodeQL cannot parse at all.

jameslamb: > _Go can't be analyzed buildless and would force a build mode Rust rejects_

For my own understanding, in case we add this to more repos.... what does this statement mean? Can you provide some links or restate it?

gmanal: CodeQl Supports different modes for different set of languages.
https://docs.github.com/en/code-security/reference/code-scanning/codeql/build-options-for-compiled-languages
For this I'll need to make changes to workflow to support language specific mode ( which is currently missing, and used none mode be default).
I am working on change and with it we should be able to enable any language in list of sast-languages.

gmanal: @jameslamb , have made the necessary changes.
https://github.com/NVIDIA/security-workflows/tree/main/.github/workflows#sast--sast-scan-codeqlyml
please review and let me know for any further questions.
Go onboarding requires building the code. 
We will enable it in separate MR.

gmanal: @jameslamb , whats the next step here? Could you please review.

.github/workflows/security-suite.yml
```diff
@@ -0,0 +1,42 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+name: security suite
+
+on:
+  push:
+    branches:
+      - "main"
+      - "release/*"
+      - "pull-request/[0-9]+"
```
jameslamb: Are you expecting that failures in this job should block PR CI?

If so, this should be updated to be callable with `workflow_call` and threaded into `pr.yaml` appropriately (find to keep it as a separate workflow here instead of as an inline job there, so it can also be run with `workflow_dispatch`).

Especially in the list that describes what has to pass for CI:

https://github.com/NVIDIA/cuvs/blob/399c9fedf8ffcf1e0639a6ea966b90808539baee/.github/workflows/pr.yaml#L11

jameslamb (changes requested):
One more question for your consideration, please.

Sorry this is dragging on so long. Having a link to prior discussion or some of these answers upfront in the PR description would have avoided some review cycles here, I think. The PR description is very detailed about what this code is but doesn't answer any the questions "why do this?" and "what is the intended usage?".

`@` me for another review when you've addressed my question.

.github/workflows/security-suite.yml
gmanal: Yes — done. `security-suite.yml` is now callable via `workflow_call` and threaded into `pr.yaml` + `pr-builder`'s `needs`, so a verified (live) secret fails the required check.

.github/workflows/pr.yaml
```diff
@@ -445,6 +453,14 @@ jobs:
       issues: write
       pull-requests: read
     uses: ./.github/workflows/check-c-abi.yaml
+  security-suite:
+    needs: telemetry-setup
```
jameslamb:
```suggestion
```

This job does not need `telemetry-setup`. It can start immediately when the run starts.

.github/workflows/security-suite.yml
jameslamb: Great, thanks.

jameslamb (approved):
Thanks for working through the review comments, no further concerns from me.

PR #2433, opened 2026-08-11
[KMeans] Configure Workspace for Large Batch Sizes

tarang-jain:
RAFT limits the regular workspaces to one-fourth the available GPU memory. If a batch size larger than that is used, fall back to the large workspace to avoid OOM.

Merge this after #2484 

Experiment to check managed vs pool MR to check if this selectively choosing the workspace makes sense. The idea is that the workspace MR is mostly a pool MR and the large MR is a managed MR. We show the difference in throughput for data transfer from host to pool and managed workspace.

H100 + Intel(R) Xeon(R) Platinum 8480CL

| Mode    | E2E Time (s) | kmeans_fit Time (s) | Average H2D Throughput (GB/s) |
|---------|-------------:|--------------------:|----------------------:|
| Managed | 217.838      | 126.545             | 51.59                 |
| Pool    | 208.082      | 118.157             | 56.71                 |


viclafargue (approved):
LGTM

cpp/src/cluster/detail/kmeans.cuh
```diff
@@ -696,15 +697,27 @@ void kmeans_fit(
...
   rmm::device_uvector<char> batch_workspace(device_buffer_samples, stream);
 
+  auto batch_memory = raft::resource::get_workspace_resource_ref(handle);
+  if constexpr (!data_on_device) {
+    size_t batch_staging_bytes =
+      static_cast<size_t>(device_buffer_samples) * static_cast<size_t>(n_features) * sizeof(DataT);
+    if (weight_ptr != nullptr) {
+      batch_staging_bytes += static_cast<size_t>(device_buffer_samples) * sizeof(DataT);
+    }
+    if (batch_staging_bytes > raft::resource::get_workspace_free_bytes(handle)) {
```
cjnolet, line 719: Why not just always use the large workspace? I think doing this conditionally creates an additional challenge for user debugging that we could avoid if we just use the same workspace resources all the time. Will let @achirkin comment here too.

tarang-jain: Thats a good point, especially since we do expect batches to be quite large ( > 20 GB or so per batch).

achirkin: The user may set up the large workspace to use a slower memory than the normal workspace (e.g. managed memory vs device pool - a setup we recommend and also set in benchmarks).
Therefore, please view this as an optimization: if access to arrays allocated via this resource is the bottleneck, we should keep it; otherwise, it's ok to use the large workspace by default.
In this case, we're talking about batching, so my understanding is having small enough batches is a normal behavior, whereas the switch to the large workspace is an edge case to make the algorithm not fail if there's not enough memory.

tarang-jain: I updated the PR desc to show the throughput with a managed and pooled workspace respectively. I am in favor of keeping this PR as is: use the large MR only when the required bytes exceed the workspace.

tarang-jain:
While the difference in H2D time exists, it is marginal (if at all) when double buffering is enabled. We avoid complex memory estimation and always use double buffering. This can be closed and revisted if future benchmarks reveal something new.

PR #2437, opened 2026-08-12
Replace O(N*degree^2) dedup with per-thread bit-packed bitset

jamxia155:
The graph shrink step in `GNND::build()` copied the NN-descent output while removing duplicate and self-referencing neighbor IDs using a nested scan: for each of the N nodes, each of the `node_degree` candidates was checked against all already-placed entries, giving `O(N*degree^2)` CPU work that scales poorly with graph degree and dataset size.

Replace with a per-thread bit-packed "seen" array (1 bit per dataset row), making the check O(1) and the whole pass `O(N*degree)`. Bits are cleared immediately after each row finishes, so no per-row reset is needed, and the extra memory is a small, nrow-independent fraction of what the step already allocates.

Tested on 48-core AMD + H100, up to 20M rows / 1024 degrees against the original loop, a sort-based variant, a hashmap variant, and a GPU kernel; this wins every configuration tested (1.8x-8.5x over the next best), though the margin vs. GPU narrows with `nrow` as the per-thread array outgrows cache.


cpp/src/neighbors/detail/nn_descent.cuh
```diff
@@ -1727,48 +1815,45 @@ void GNND<Data_t, Index_t>::build(Data_t* data,
...
+    const int graph_degree = static_cast<int>(graph_.node_degree);
+
+    // Upload h_graph (InternalID_t = int32, same size as Index_t) to device.
+    static_assert(sizeof(InternalID_t<Index_t>) == sizeof(int),
+                  "dedup_graph_kernel assumes InternalID_t is int-sized");
+    const size_t h_graph_elems = static_cast<size_t>(nrow_) * graph_degree;
+    rmm::device_uvector<int> d_in(h_graph_elems, stream);
+    RAFT_CUDA_TRY(cudaMemcpyAsync(
+      d_in.data(), graph_.h_graph, h_graph_elems * sizeof(int), cudaMemcpyHostToDevice, stream));
+
```
tarang-jain: This PR brings the entire set of host candidates into device memory. The number of candidates per host can be very large. Earlier we were only capping the degree of on device graph buffers to DEGREE_ON_DEVICE (=32) it seems.

Secondly, I don't see why we were even doing it in this way on host in the first place i.e. comparing every single pair. We already call `sort_lists()`, so the list of candidates is sorted by distances, right? Since we are sorting pairs, the same candidate should technically appear adjacent. Then we should just do a single scan through each candidate list. Or are you saying that there could be differences (for example the epilog is asymmetric maybe?) which means the same candidate can appear at non-adjacent places in the sorted list?

tarang-jain (changes requested):

cpp/src/neighbors/detail/nn_descent.cuh
jamxia155: Thanks for taking a look, Tarang. I'll look into a streaming approach to keep the memory footprint bounded.

As for the de-dup implementation, the original implementation sorted the candidates by distance so a de-dup based on ID cannot be achieved with a linear scan. However, it is possible to first sort by ID, dedup, then sort again by distance. I will run some tests and see if that closes the gap with the GPU port.

tarang-jain: > the original implementation sorted the candidates by distance

It sorts the pairs. So for candidates with identical distance, the same candidate would be placed adjacent, right? I think std::sort does this for tuples by moving on to the next element if the first one is identical.

tarang-jain: But we need to think about cases when the same candidate can have two different distances for example small numerical differences in distances (while adding reverse edges). Is that possible? I would be quite surprised though if there was asymmetry like that.

jamxia155: The possibility for the same candidate with two different distances arises from a somewhat different scenario (all line numbers refer to [cpp/src/neighbors/detail/nn_descent.cuh](https://github.com/NVIDIA/cuvs/blob/4c720a29829d25057efccc28a3f110a1d6226f66/cpp/src/neighbors/detail/nn_descent.cuh) (permalink):
- 1239-1241: `init_random_graph()` seeds each row's initial candidate list segment-by-segment, independently, with no cross-segment uniqueness check. The same `id` value from line 1240 can happen to end up in multiple segments.
- 1246: the distances are initialized as `std::numeric_limits::max()`.
- 1312: `update_graph()` computes `seg_idx = new_neighb_id.id() % num_segments`; 1316: `insert_to_ordered_list()` is called using this one `seg_idx` only. If the same `id` exists in another segment, its initial value never gets overwritten.

jamxia155: I benchmarked a few different implementations (original, sort-based, hashmap-based, bitset-based, GPU `__ballot_sync`-based): the bitset-based approach on CPU won in all configurations tested. So I'm switching this PR to a CPU implementation after all.

Of course, the perf comparison depends on the relative capabilities of the CPU and GPU (node I tested on had 48 logical cores + H100 PCI-E) but the burden of proof to switch to GPU is also higher than the current CPU optimization proposed.

cpp/src/neighbors/detail/nn_descent.cuh
```diff
@@ -1727,47 +1727,59 @@ void GNND<Data_t, Index_t>::build(Data_t* data,
...
+      }
+
+      // Unset every bit this row touched so the array is back to all-zero for the next row this
+      // thread processes. Safe to run over the full row rather than tracking exactly which
+      // entries were newly set: clearing an already-clear bit is a no-op, and a duplicate value
+      // that slipped through via the fallback above (only possible if a row exhausts its retry
+      // budget) shares its bit with wherever it was legitimately set earlier in this same row.
+      for (size_t k = 0; k < build_config_.node_degree; k++) {
+        size_t idx = static_cast<size_t>(output_neighbor_list_ptr[k]);
+        seen_bits[idx >> 6] &= ~(uint64_t{1} << (idx & 63));
```
tarang-jain: Can we not directly set this to zero?
So `seen_bits[idx >> 6] = 0`?

tarang-jain (commented):
The bitset approach looks great. Just one small comment.

cpp/src/neighbors/detail/nn_descent.cuh
jamxia155: Most astute! Updated.

tarang-jain (approved):

PR #2476, opened 2026-08-17
Fix GPU and Java-heap OOM when building large vector indexes

nvzm123:
## Summary

This ports [NVIDIA/cuvs-lucene#173](https://github.com/NVIDIA/cuvs-lucene/pull/173) into `java/cuvs-lucene` following the move into the cuVS monorepo.

The scope is limited to the accelerated-HNSW memory-pressure change:

1. Float, binary-quantized, and scalar-quantized accelerated-HNSW construction uses native host-backed cuVS matrices instead of retaining a complete device-backed input matrix.
2. Ordinary FLOAT32 force merges count the live vectors yielded by Lucene and replay them directly into one native host matrix, avoiding a complete intermediate `List<float[]>`.
3. Higher-layer construction can read selected rows from that native matrix without recreating the complete dataset on the Java heap.
4. Empty and one-vector ordinary merge results are handled before allocating the non-trivial build matrix.

The resulting codec still uses GPU CAGRA construction followed by Lucene CPU HNSW search.

The original work superseded [NVIDIA/cuvs-lucene#141](https://github.com/NVIDIA/cuvs-lucene/pull/141). The related increase of `MAX_HNSW_LAYERS` to 99 is already present on the current monorepo `main` and is not part of this diff.

## Merge and ownership safety

The memory change includes the safety required by its new allocation and replay path:

1. `CuVSMatrix.Builder` is closeable. The built-in builders release untransferred storage when closed, transfer ownership only after a successful `build()`, reject use after close or build, and allow repeated `close()` calls.
2. A default no-op `close()` preserves compatibility for providers compiled against the earlier builder interface.
3. The first merge pass counts the live vectors actually returned by Lucene rather than relying on source-segment metadata.
4. The replay pass rejects both fewer and more vectors than the first pass observed and closes the builder on read, population, validation, or build failure.
5. The existing public `AcceleratedHNSWUtils.createMultiLayerHnswGraph(...)` list-based descriptor remains available; the native-matrix overload used internally is package-private.

## Scope boundaries

- `CuVS2510GPUSearchCodec` remains device-backed and is otherwise unchanged by this PR.
- Binary- and scalar-quantized force merges still materialize their existing heap lists; streaming those paths is deferred to a follow-up.
- Upper-layer selection and remapping behavior is unchanged. The default remains one HNSW layer.
- Broader native-resource cleanup, GPU-search fallback policy, malformed-graph validation, and upper-layer correctness hardening are deferred to a follow-up.

## Validation

Fresh exact-source results for this branch:

- `cuvs-java`: 5 unit tests and 111 integration tests; 0 failures, 0 errors, and 1 inherited skip
- `cuvs-lucene`: 344 tests; 0 failures, 0 errors, and 30 inherited skips
- Three-level reopen-and-search coverage for float, binary-quantized, and scalar-quantized accelerated HNSW
- Deleted, sparse, zero-live-vector, and one-live-vector force-merge coverage
- Merge replay underflow, overflow, cleanup, and suppressed-failure coverage
- Spotless, generated-documentation idempotence, and `git diff --check`

Deep1B 1M, 96 dimensions, using fresh index directories:

- Float, binary-quantized, and scalar-quantized builds each flushed four 250,000-vector segments and force-merged them into one three-level accelerated-HNSW segment.
- For every format, validation reopened the retained index and checked `checkIntegrity()`, 1,000,000 documents and vectors, a complete unique source-ID domain, graph level membership and nesting, duplicate-free adjacency, finite search results, and ground-truth overlap.

These Deep1B runs are correctness sanity checks. Page-cache eviction was unavailable in the validation container, so the recorded timings are not performance evidence.


java/cuvs-lucene/src/main/java/com/nvidia/cuvs/lucene/AcceleratedHNSWUtils.java
```diff
@@ -77,95 +77,83 @@ public static GPUBuiltHnswGraph createSingleVectorHnswGraph(int size, int dimens
...
+      CuVSMatrix vectorDataset,
       int hnswLayers,
       CagraIndexParams params,
       QuantizationType quantization)
       throws Throwable {
 
+    int size = (int) vectorDataset.size();
     int M = Math.ceilDiv((int) adjacencyListMatrix.columns(), 2);
 
-    // Store all layers data
```
imotov, line 95: Why did you remove all comments from this method?

java/cuvs-lucene/src/main/java/com/nvidia/cuvs/lucene/Lucene99AcceleratedHNSWVectorsWriter.java
```diff
@@ -154,29 +158,48 @@ private void writeFieldInternal(FieldInfo fieldInfo, List<float[]> vectors) thro
...
+    if (size == 0) {
+      writeEmpty(fieldInfo, hnswMeta);
+      return;
+    }
+    if (size < 2) {
+      float[] buf = new float[fieldInfo.getVectorDimension()];
+      dataset.getRow(0).toArray(buf);
+      writeSingleVectorGraph(fieldInfo, List.of(buf));
+      return;
+    }
```
imotov: The code duplication with lines 153-160 that preceded this call looks strange.

java/cuvs-lucene/src/main/java/com/nvidia/cuvs/lucene/Lucene99AcceleratedHNSWVectorsWriter.java
```diff
@@ -273,13 +296,24 @@ private void writeSingleVectorGraph(FieldInfo fieldInfo, List<float[]> vectors)
...
+   * occurs when force-merging large segments.
    */
   private void vectorBasedMerge(FieldInfo fieldInfo, MergeState mergeState) throws IOException {
     try {
-      List<float[]> dataset =
-          createListFromMergedVectors(
-              KnnVectorsWriter.MergedVectorValues.mergeFloatVectorValues(fieldInfo, mergeState));
+      FloatVectorValues mergedVectors =
+          KnnVectorsWriter.MergedVectorValues.mergeFloatVectorValues(fieldInfo, mergeState);
+      int size = mergedVectors.size();
```
imotov: I don't think this size is correct when merging segments with deletes. It's computed up front as `sum(sub.values.size())` over the raw per-segment values, while the iterator below is driven by `DocIDMerger` and skips deleted docs - so the matrix is allocated for more rows than we actually fill, and the tail stays zeroed.

`dataset.size()` then still reports the inflated count, so CAGRA builds over the phantom zero vectors and we write an HNSW graph with more nodes than flatVectorsWriter wrote vectors causing all sorts of issue. 

`TestAcceleratedHNSWDeletedDocuments#testVectorSearchWithDeletedDocuments` catches it on most seeds, e.g. `-Dtests.seed=496998C`.

java/cuvs-lucene/src/main/java/com/nvidia/cuvs/lucene/Lucene99AcceleratedHNSWVectorsWriter.java
```diff
@@ -154,29 +158,48 @@ private void writeFieldInternal(FieldInfo fieldInfo, List<float[]> vectors) thro
       writeSingleVectorGraph(fieldInfo, vectors);
       return;
     }
-    try {
-      CuVSMatrix dataset =
-          Utils.createFloatMatrix(
-              vectors, fieldInfo.getVectorDimension(), getCuVSResourcesInstance());
+    CuVSMatrix dataset = Utils.createFloatMatrix(vectors, fieldInfo.getVectorDimension());
+    writeFieldInternal(fieldInfo, dataset);
```
imotov: I think that would leak `dataset` for trivial cases such as `dataset.size() < 2` since nobody takes ownership of dataset as in the non-trivial case.

java/cuvs-lucene/src/main/java/com/nvidia/cuvs/lucene/AcceleratedHNSWUtils.java
nvzm123: Oops, I only intended to remove the superfluous comments -- I restored the useful comments, corrected two inaccurate descriptions, and documented the matrix overload’s non-obvious width and allocation behavior. Thanks!

java/cuvs-lucene/src/main/java/com/nvidia/cuvs/lucene/Lucene99AcceleratedHNSWVectorsWriter.java
nvzm123: That did look a bit strange, thanks. I extracted the zero- and one-vector cases into writeTrivialField(...). Both the flush/list path and merge path call it before allocating a matrix, while writeNonTrivialField(...) now contains only the shared non-trivial build path.

java/cuvs-lucene/src/main/java/com/nvidia/cuvs/lucene/Lucene99AcceleratedHNSWVectorsWriter.java
nvzm123: Good catch, that's correct, thank you! The raw per-segment size was not a safe allocation count when DocIDMerger filtered deleted documents. The merge now counts exactly the live vectors yielded by MergedVectorValues, allocates that many rows, and validates that a fresh replay produces the same count. Underflow and overflow fail explicitly and release untransferred storage. I also added deterministic deleted-and-sparse force-merge coverage, zero/one-live-vector cases, and focused replay-mismatch tests.

java/cuvs-lucene/src/main/java/com/nvidia/cuvs/lucene/Lucene99AcceleratedHNSWVectorsWriter.java
nvzm123: Fixed, thank you. Zero- and one-vector fields are now handled before creating either a host matrix or merge builder, so those cases allocate no dataset. For non-trivial merges, the builder is try-with-resources and releases its matrix if replay fails before ownership transfer.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/spi/JDKProvider.java
```diff
@@ -622,7 +622,8 @@ public CuVSMatrix newMatrixFromArray(byte[][] vectors) {
     return dataset;
   }
 
-  private abstract static class MatrixBuilder<T extends CuVSMatrixInternal> {
+  private abstract static class MatrixBuilder<T extends CuVSMatrixInternal>
+      implements AutoCloseable {
```
imotov: I'm not quite clear on what this is for. Could you clarify?

java/cuvs-lucene/src/main/java/com/nvidia/cuvs/lucene/Lucene99AcceleratedHNSWVectorsWriter.java
```diff
@@ -139,44 +142,47 @@ public KnnFieldVectorsWriter<?> addField(FieldInfo fieldInfo) throws IOException
...
+    try {
+      int size = (int) dataset.size();
       CagraIndexParams params =
           CagraIndexParamsFactory.create(acceleratedHNSWParams, dataset.size(), dataset.columns());
-
       CagraIndex cagraIndex =
           CagraIndex.newBuilder(getCuVSResourcesInstance())
               .withDataset(dataset)
               .withIndexParams(params)
               .build();
```
imotov, line 177: Wouldn't this leak as well?

java/cuvs-lucene/src/main/java/com/nvidia/cuvs/lucene/Lucene99AcceleratedHNSWVectorsWriter.java
```diff
@@ -139,44 +142,47 @@ public KnnFieldVectorsWriter<?> addField(FieldInfo fieldInfo) throws IOException
...
    * @throws IOException
    */
   private void writeFieldInternal(FieldInfo fieldInfo, List<float[]> vectors) throws IOException {
-    if (vectors.size() == 0) {
-      writeEmpty(fieldInfo, hnswMeta);
-      return;
-    }
-    if (vectors.size() < 2) {
-      writeSingleVectorGraph(fieldInfo, vectors);
+    if (writeTrivialField(fieldInfo, vectors.size())) {
```
imotov, line 153: That seems to still leak as I mentioned before I think.

java/cuvs-lucene/src/main/java/com/nvidia/cuvs/lucene/LuceneAcceleratedHNSWBinaryQuantizedVectorsWriter.java
jamxia155: If dataset size is 1, `dataset` is not closed and gets leaked.

java/cuvs-lucene/src/main/java/com/nvidia/cuvs/lucene/LuceneAcceleratedHNSWScalarQuantizedVectorsWriter.java
jamxia155: Same issue as flagged in `java/cuvs-lucene/src/main/java/com/nvidia/cuvs/lucene/LuceneAcceleratedHNSWBinaryQuantizedVectorsWriter.java`

java/cuvs-lucene/src/main/java/com/nvidia/cuvs/lucene/Lucene99AcceleratedHNSWVectorsWriter.java
```diff
@@ -139,44 +142,47 @@ public KnnFieldVectorsWriter<?> addField(FieldInfo fieldInfo) throws IOException
...
       return;
     }
-    try {
-      CuVSMatrix dataset =
-          Utils.createFloatMatrix(
-              vectors, fieldInfo.getVectorDimension(), getCuVSResourcesInstance());
+    CuVSMatrix dataset = Utils.createHostFloatMatrix(vectors, fieldInfo.getVectorDimension());
```
jamxia155, line 156: This now creates a host matrix from `vectors` instead of a device matrix so that's a tradeoff between GPU and host memory? Also should this be counted in `ramBytesUsed()`?

java/cuvs-lucene/src/main/java/com/nvidia/cuvs/lucene/Lucene99AcceleratedHNSWVectorsWriter.java
```diff
@@ -273,19 +289,75 @@ private void writeSingleVectorGraph(FieldInfo fieldInfo, List<float[]> vectors)
...
-      List<float[]> dataset =
-          createListFromMergedVectors(
-              KnnVectorsWriter.MergedVectorValues.mergeFloatVectorValues(fieldInfo, mergeState));
-      writeFieldInternal(fieldInfo, dataset);
+      int size = countMergedVectors(fieldInfo, mergeState);
+      if (writeTrivialField(fieldInfo, size)) {
+        return;
+      }
+      FloatVectorValues mergedVectors =
+          KnnVectorsWriter.MergedVectorValues.mergeFloatVectorValues(fieldInfo, mergeState);
```
jamxia155: `countMergedVectors` already computes `mergedVectors` using `mergeFloatVectorValues`, but we compute `mergedVectors` again here.

PR #2480, opened 2026-08-18
[WIP] Vpq dataset serialisation

irina-resh-nvda:
**Main functionality:** 
A PQ-compressed dataset could be built and searched but not saved: serialize static_asserts on any dataset view that is not dense, so compressed rows were requantized on every run and an index over them could not be persisted at all.

This restores the writer in the dataset serializer and exposes it twice: `pq::serialize / pq::deserialize` for a `device_vpq_dataset<half, int64_t>`, over a stream or a filename and taking the dataset before its destination, so a dataset can be compressed once offline and reloaded; and `cagra::serialize / deserialize` for `vpq_f16_index` under a new `device_pq` dataset kind, so the compressed rows can travel inside the index file. One kind covers both codebook types, since the payload records its own.

`out_dataset` stays optional, as everywhere else: leaving it out loads the graph and skips whatever rows the file holds, for a caller who keeps their own PQ-quantized dataset.

Since an index names its dataset kind in its type,` cagra::read_serialized_header` reports a file's `dtype` and kind, and `cagra::owning_dataset_for_index_t` names the dataset an index deserializes into.

A standalone .vpq carries the preamble the CAGRA index files already use, a `NumPy` `dtype` prefix then a version, so a future change to the encoded row layout fails cleanly instead of being misread.

**Also here:** 
`MG_C_TEST` is registered only when `BUILD_MG_ALGOS` is on, where it previously failed to link. The C API is otherwise untouched: it still reports a compressed index file as an invalid kind, since `cuvsDatasetLayout_t` has no layout for one, and the single case added to its dispatch is unreachable and there only because `-Wswitch` is an error.

Docs: the search overload documentation says what those overloads take instead of naming `vpq_f16_index`, "CAGRA-Q" is gone from the code this PR writes, and `make_vpq_dataset` and `pq::serialize` now state their memory cost.

**Tests:** 
`PREPROCESSING_TEST` gains `PqSerialization`, which round-trips the `pq_bits`, `pq_dim` combinations a CAGRA search accepts, comparing both raw bytes and decoded vectors, and covers the refusals: an empty stream, a foreign `dtype` prefix, a version from the future, a truncated payload and a null out-parameter.

`NEIGHBORS_ANN_CAGRA_FLOAT_UINT32_TEST` gains `CagraPqSerialize`, which saves an index over compressed rows, loads it back, and checks that it returns the same neighbours with the dense rows gone. It also covers saving the graph on its own, loading the graph alone and reattaching rows afterwards, skipping a dense dataset through the same path, reading the header before loading, and refusing to load a compressed index as a dense one.

# CAGRA over a PQ-compressed dataset: example workflows

Example workflows

## Common start

```cpp
auto compressed = pq::make_vpq_dataset(res, vpq_params, dense);   // dense -> compressed, on device
auto graph_src  = cagra::build(res, ip, make_device_standard_dataset_view(dense));
cagra::vpq_f16_index<float> idx{res, L2Expanded, compressed.as_dataset_view(), graph_src.graph()};
// where the iterative build over compressed rows exists, the last two lines are one:
//   cagra::build(res, ip, compressed.as_dataset_view())
```

## Path 1 — rows are inside the index file

```cpp
cagra::serialize(res, "index.bin", idx);                  // include_dataset defaults to true
// ---- later, elsewhere
cagra::vpq_f16_index<float> loaded{res};
std::unique_ptr<cagra::owning_dataset_for_index_t<decltype(loaded)>> rows;
cagra::deserialize(res, "index.bin", &loaded, &rows);     // rows come out of the file
cagra::search(res, sp, loaded, queries, nbrs, dists);     // `rows` must outlive `loaded`
```

## Path 2 — rows kept separately as a `.vpq` (compress once, sweep many)

```cpp
pq::serialize(res, compressed, "base.vpq");                             // the input format
cagra::serialize(res, "index.bin", idx, /* include_dataset */ false);   // graph only
// ---- later, elsewhere
std::unique_ptr<cuvs::neighbors::device_vpq_dataset<half, int64_t>> rows;
pq::deserialize(res, "base.vpq", &rows);
cagra::vpq_f16_index<float> loaded{res};
cagra::deserialize(res, "index.bin", &loaded);                          // no out_dataset: skipped
loaded.update_device_dataset_same_layout(res, rows->as_dataset_view());
cagra::search(res, sp, loaded, queries, nbrs, dists);
```

## Notes

- If you don't know which file you've been handed:
  `auto h = cagra::read_serialized_header(res, "index.bin");` gives `h.dtype` and `h.dataset_kind`
  (`serialized_dataset_kind::device_pq` for a compressed one) before you name the index type.
- Path 2's graph-only load also works on a file written by path 1: the rows in it are skipped.
- An index never owns its rows. Whatever holds them — `compressed`, `rows` — has to outlive every
  index that views them, and a graph-only index cannot answer a query until something attaches rows.
- The C API cannot load path 1's files yet: `cuvsDatasetLayout_t` has no compressed layout, so its
  header check rejects the kind. Path 2's graph-only file is fine there.

### DIFF
--- a/c/src/neighbors/cagra.cpp
+++ b/c/src/neighbors/cagra.cpp
@@ -15,8 +15,11 @@
 #include <raft/core/copy.hpp>
 #include <raft/core/error.hpp>
 #include <raft/core/mdspan_types.hpp>
+#include <raft/core/numpy_serializer.hpp>
 #include <raft/core/resource/cuda_stream.hpp>
+#include <raft/core/numpy_serializer.hpp>
 #include <raft/core/resources.hpp>
+#include <raft/core/serialize.hpp>
 
 #include "../core/exceptions.hpp"
 #include "../core/interop.hpp"
@@ -971,40 +974,47 @@ struct serialized_cagra_header {
   cuvs::neighbors::cagra::serialized_dataset_kind dataset_kind;
 };
 
-/**
- * What the file holds, in the terms the C entry points dispatch on.
- *
- * The reading is the C++ API's: `cagra::read_serialized_header` parses the same preamble, validates
- * the format version and the dataset kind, and leaves this with nothing to do but restate the
- * element type in DLPack terms.
- */
 static auto read_serialized_header(cuvsResources_t res, const char *filename)
     -> serialized_cagra_header {
   auto res_ptr = reinterpret_cast<raft::resources *>(res);
-  auto const header = cuvs::neighbors::cagra::read_serialized_header(
-      *res_ptr, std::string(filename));
-  DLDataType dtype{.code = 0, .bits = 0, .lanes = 1};
-  switch (header.dtype) {
-    case CUDA_R_32F:
-      dtype.code = kDLFloat;
-      dtype.bits = 32;
-      break;
-    case CUDA_R_16F:
-      dtype.code = kDLFloat;
-      dtype.bits = 16;
-      break;
-    case CUDA_R_8I:
-      dtype.code = kDLInt;
-      dtype.bits = 8;
-      break;
-    case CUDA_R_8U:
-      dtype.code = kDLUInt;
-      dtype.bits = 8;
-      break;
-    default:
-      RAFT_FAIL("Unsupported dtype in file %s", filename);
+  std::ifstream is(filename, std::ios::in | std::ios::binary);
+  if (!is) {
+    RAFT_FAIL("Cannot open file %s", filename);
   }
-  return {dtype, header.dataset_kind};
+
+  char dtype_string[4]{};
+  if (!is.read(dtype_string, sizeof(dtype_string))) {
+    RAFT_FAIL("Invalid or truncated index header in file %s", filename);
+  }
+
+  auto const dtype = raft::numpy_serializer::parse_descr(
+      std::string(dtype_string, sizeof(dtype_string)));
+  DLDataType output_dtype{
+      .code = 0, .bits = static_cast<uint8_t>(dtype.itemsize * 8), .lanes = 1};
+  if (dtype.kind == 'f' && dtype.itemsize == 4) {
+    output_dtype.code = kDLFloat;
+  } else if (dtype.kind == 'e' && dtype.itemsize == 2) {
+    output_dtype.code = kDLFloat;
+  } else if (dtype.kind == 'i' && dtype.itemsize == 1) {
+    output_dtype.code = kDLInt;
+  } else if (dtype.kind == 'u' && dtype.itemsize == 1) {
+    output_dtype.code = kDLUInt;
+  } else {
+    RAFT_FAIL("Unsupported dtype in file %s", filename);
+  }
+
+  auto const version = raft::deserialize_scalar<int>(*res_ptr, is);
+  auto const dataset_kind_raw =
+    raft::deserialize_scalar<std::uint32_t>(*res_ptr, is);
+  RAFT_EXPECTS(
+      version == cuvs::neighbors::cagra::cagra_serialization_version,
+      "serialization version mismatch, expected %d, got %d",
+      cuvs::neighbors::cagra::cagra_serialization_version, version);
+  using kind = cuvs::neighbors::cagra::serialized_dataset_kind;
+  RAFT_EXPECTS(dataset_kind_raw <= static_cast<std::uint32_t>(kind::host_standard),
+               "Invalid serialized dataset kind %u in file %s",
+               dataset_kind_raw, filename);
+  return {output_dtype, static_cast<kind>(dataset_kind_raw)};
 }
 
 template <typename Fn>
@@ -1048,12 +1058,10 @@ void dispatch_serialized_dataset_kind(
       fn.template operator()<
           cuvs::neighbors::device_padded_dataset_view<T, int64_t>>();
       break;
-    case serialized_kind::device_pq:
-      // A recognised file the C API has no index layout for, as opposed to an unreadable one.
-      // cuvsDatasetLayout_t covers standard and padded only, and every C entry point dispatches
-      // on that layout, so there is nothing here to hand a PQ-compressed index to yet.
-      RAFT_FAIL("File holds a PQ-compressed dataset, which the C API has no dataset "
-                "layout for; load it through the C++ API");
+    // Unreachable: read_serialized_header rejects this kind before the dispatch, since
+    // cuvsDatasetLayout_t has no PQ-compressed layout to hand back. Listed only because the switch
+    // is exhaustive and -Wswitch is an error. Delete it when the C API gains the layout.
+    case serialized_kind::device_pq: break;
   }
 }
 
--- a/c/src/neighbors/cagra.cpp
+++ b/c/src/neighbors/cagra.cpp
@@ -1048,11 +1048,11 @@ void dispatch_serialized_dataset_kind(
       fn.template operator()<
           cuvs::neighbors::device_padded_dataset_view<T, int64_t>>();
       break;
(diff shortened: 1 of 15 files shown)


c/src/neighbors/cagra.cpp
```diff
@@ -1063,6 +1058,12 @@ void dispatch_serialized_dataset_kind(
       fn.template operator()<
           cuvs::neighbors::device_padded_dataset_view<T, int64_t>>();
       break;
+    case serialized_kind::device_vpq_f16:
+      // A recognised file the C API has no index layout for, as opposed to an unreadable one.
+      // cuvsDatasetLayout_t covers standard and padded only, and every C entry point dispatches
+      // on that layout, so there is nothing here to hand a VPQ index to yet.
+      RAFT_FAIL("File holds a VPQ-compressed (CAGRA-Q) dataset, which the C API has no dataset "
+                "layout for; load it through the C++ API");
```
cjnolet: This is not an acceptable way to handle things tha are not yet exposed to C. Folks use C because they have to (for C ABI stability guarantees, for example). We can't just throw an error and forward them to C++

cpp/include/cuvs/neighbors/cagra.hpp
```diff
@@ -2274,16 +2264,19 @@ enum class serialized_dataset_kind : std::uint32_t {
   host_padded = 3,
   /** Host-resident dataset using its standard row layout. */
   host_standard = 4,
+  /** Device-resident VPQ-compressed dataset with f16 codebooks (CAGRA-Q). */
+  device_vpq_f16 = 5,
```
cjnolet: Since we're adding new things, please rename to `device_pq`. VPQ is not a proper term it accidentally made its way into the public APIs (which was an oversight). Also- why the "f16"? Can we drop that?

cpp/include/cuvs/neighbors/cagra.hpp
```diff
@@ -2834,6 +2827,106 @@ void deserialize(raft::resources const& handle,
...
+ * the dense dataset it was compressed from and without retraining the codebooks. As everywhere
+ * else, the index holds a view: `deserialize` returns the owning dataset through `out_dataset`,
+ * which the caller has to keep alive for as long as the index is used.
+ *
+ * Unlike the dense overloads, `out_dataset` is required. Nothing can be searched in a VPQ index
+ * whose rows were dropped, so there is no use for a graph-only load, and asking for one is an
+ * error rather than a silently unusable index. For the same reason `include_dataset = false`
+ * produces an index that only `update_dataset` can make searchable again.
+ */
+void serialize(raft::resources const& handle,
```
cjnolet, line 2914: Order of argument should be handle, params (if applicable), input, output.

cpp/include/cuvs/neighbors/cagra.hpp
```diff
@@ -2834,6 +2827,106 @@ void deserialize(raft::resources const& handle,
...
+ *
+ * The compressed rows travel with the index, so that a deserialized index can be searched without
+ * the dense dataset it was compressed from and without retraining the codebooks. As everywhere
+ * else, the index holds a view: `deserialize` returns the owning dataset through `out_dataset`,
+ * which the caller has to keep alive for as long as the index is used.
+ *
+ * Unlike the dense overloads, `out_dataset` is required. Nothing can be searched in a VPQ index
+ * whose rows were dropped, so there is no use for a graph-only load, and asking for one is an
+ * error rather than a silently unusable index. For the same reason `include_dataset = false`
+ * produces an index that only `update_dataset` can make searchable again.
```
cjnolet: Please use proper doxygen tags to denote which arguments are inputs and outputs.

cpp/include/cuvs/neighbors/cagra.hpp
```diff
@@ -2834,6 +2827,106 @@ void deserialize(raft::resources const& handle,
                  std::unique_ptr<cuvs::neighbors::device_standard_dataset<uint8_t, int64_t>>*
                    out_dataset = nullptr);
 
+/* vpq_f16_index overloads (CAGRA-Q).
```
cjnolet: `vpq_f16_index` doesn't mean anything to a user reading the docs. Please opt to use more descriptive summaries for comments. Please also don't call this "cagra-q" anymore. It's just cagra with quantized vectors (the vector quantization should be completely decoupled from the graph).

cpp/include/cuvs/neighbors/cagra.hpp
```diff
@@ -2834,6 +2827,106 @@ void deserialize(raft::resources const& handle,
...
+ */
+void serialize(raft::resources const& handle,
+               const std::string& filename,
+               const cuvs::neighbors::cagra::vpq_f16_index<float>& index,
+               bool include_dataset = true);
+
+void deserialize(
+  raft::resources const& handle,
+  const std::string& filename,
+  cuvs::neighbors::cagra::vpq_f16_index<float>* index,
```
cjnolet, line 2939: Please make the output dataset optional. The graph does not need to be deserialized with the vectors and we should not enforce this. Also- we would like to not require the dataset be known up front wen the user is deserializing a graph that might happen to have a dataset. Goal is to have it deserialize to _some_ dataset owned by the user, but the user shouldn't have to care about its concrete type. 

In the C layer, we play to make this even more abstract.

cpp/include/cuvs/neighbors/cagra.hpp
cjnolet: I see the other functions did not follow the proper order. Let's at least make sure the new functions are in the proper order (we can have someone update the others in a follow-up).

cpp/src/preprocessing/quantize/pq.cu
```diff
@@ -76,6 +83,55 @@ CUVS_INST_VPQ_BUILD(uint8_t);
...
+  std::ofstream os(filename, std::ios::out | std::ios::binary | std::ios::trunc);
+  RAFT_EXPECTS(os.good(), "pq::serialize: cannot open %s for writing", filename.c_str());
+  serialize(res, os, dataset);
+}
+
+void deserialize(raft::resources const& res,
+                 std::istream& is,
+                 std::unique_ptr<cuvs::neighbors::device_vpq_dataset<half, int64_t>>* out_dataset)
+{
+  RAFT_EXPECTS(out_dataset != nullptr, "pq::deserialize: out_dataset must not be null");
```
cjnolet, line 113: We can't do this- the whole point in decoupling the dataset from the index is to satisfy use-cases where the user has already pq quantized their dataset and just wants to construct a graph. They should not be forced to store the pq vectors with the index.

cpp/include/cuvs/neighbors/cagra.hpp
irina-resh-nvda: f16 for consistency with how its defined elsewhere (e.g. cagra_search_inst.cu.in)
I will rename all terms my branch introduces, but will leave the larger vpq->pq rename for the follow up work

c/src/neighbors/cagra.cpp
irina-resh-nvda: @aamijar Can you help me with the C part?

cpp/include/cuvs/neighbors/cagra.hpp
irina-resh-nvda: Done.
I also updated the description of this PR, the use case you mention is now outlined in path 2

cpp/src/preprocessing/quantize/pq.cu
irina-resh-nvda: This function deserializes the pq dataset itself

PR #2481, opened 2026-08-18
Optimize cuvs-lucene CAGRA_HNSW index build; add example

jamxia155:
This PR replaces [cuVS-Lucene #195](https://github.com/NVIDIA/cuvs-lucene/pull/195) as cuVS-Lucene has been merged into cuVS.

This PR builds on the out-of-core host-streaming refactor from #2476 by @nvzm123. This PR subsumes #2476 to avoid stacking the PRs.

On top of #2476, this PR adds the following.

**Improvements**

- Native flat buffering: stream vectors directly into a native host matrix during indexing instead of buffering them as a heap List<float[]>, cutting peak host memory from ~2x to ~1x and eliminating the per-vector matrix-assembly copy. Opt-in; requires all input vectors to be indexed in the original order, i.e. not supported for sorted, merged, or filtered index segments; not currently supported for segments with quantized fields.
- Parallelized CAGRA-to-HNSW graph conversion: materializing the CAGRA adjacency into on-heap NeighborArrays was a serial per-node loop; parallelized under the existing `writerThreads` knob.
- Parallelized level-0 HNSW graph serialization: level-0 (all N nodes) is now delta/VInt-encoded in parallel across memory-bounded waves of threads, then concatenated to the `IndexOutput` in node order according to on-disk format in the serial path.

**New example: OptimizedCagraHnswBuildExample**

A reference pattern for building a large accelerated HNSW index with all ingest- and build-side optimizations, showcasing:

- Streaming, prefetched, bounded-memory ingestion: open the source file once, read it front-to-back in large sequential chunks; hold at most two chunks in memory; fill the next chunk on a background thread while the ingest thread drains the current one, hiding disk read behind indexing; unpack into a caller-reused float[] allocation (no per-vector allocation, safe because Lucene copies the value eagerly inside `addDocument`).
- Native flat buffering sized exactly to each segment's vector count via `withNumInputVectors`, avoiding the heap-buffered assembly copy.
- Overlapping multi-segment build:
    - Default: K single-segment passes appended to one directory; peak host memory is one slice (N/K).
    - Overlapped: a bounded pool (`PIPELINE_DEPTH`) builds segments concurrently into their own directories with the GPU commit serialized on a semaphore (ingest overlaps a prior segment's GPU commit), then combines the finished per-segment indexes by hardlinking their files into the final directory via `HardlinkCopyDirectoryWrapper` + `addIndexes` (no bulk copy of vector data). Peak host memory is up to `PIPELINE_DEPTH * N / K`.
- An `enableRMMAsyncMemory()` call (with a note that it must not be used with CPU-only codecs) to opt-in to RMM-managed memory resources, plus exposing the primary tuning knobs (withMaxConn, withBeamWidth, withCuvsDistanceType, withWriterThreads).


java/cuvs-lucene/src/main/java/com/nvidia/cuvs/lucene/AcceleratedHNSWUtils.java
```diff
@@ -235,56 +243,149 @@ private static CuVSMatrix buildCagraGraphForSubset(
...
+                    return null;
+                  }));
         }
-        // Write the size after duplicates are removed
-        vectorIndex.writeVInt(actualSize);
-        // Write de-duplicated neighbors
-        for (int i = 0; i < actualSize; i++) {
-          vectorIndex.writeVInt(scratch[i]);
+        for (Future<?> f : futures) {
+          f.get();
```
imotov: If one of these tasks throws  InterruptedException or an ExecutionException  the remaining tasks in that wave keep running to completion even after this method exists. I think it might be prudent to call shutdownNow() in case of an exception so we interrupt all other threads and call shutdown() only in the case of a success.

java/cuvs-lucene/src/main/java/com/nvidia/cuvs/lucene/NativeFlatVectorsWriter.java
```diff
@@ -0,0 +1,202 @@
...
+ *       https://github.com/apache/lucene/blob/releases/lucene/10.2.0/lucene/core/src/java/org/apache/lucene/codecs/lucene99/Lucene99FlatVectorsWriter.java
+ * </ul>
+ *
+ * <p><b>On a Lucene upgrade:</b> re-verify this class against the new version's two files above. The
+ * constants are package-private in Lucene (hence re-declared here), and the file format is validated
+ * on read via {@code CodecUtil.checkIndexHeader} (codec name + version range) plus the fixed meta
+ * layout — so if the new version bumps {@code VERSION_CURRENT}, renames a codec, or changes the meta
+ * field order, files written here will be silently incompatible and the stock
+ * {@code Lucene99FlatVectorsReader} will reject or misread them. Concretely:
+ *
```
imotov, line 52: This doesn't sound quite right to me. I think the guidance should be the opposite: once this is written, we should *not* touch these constants.

Whatever we write with the current `Lucene99FlatVectorsFormat` v0 layout stays readable through the last Lucene 11.x release, since Lucene guarantees back-compat for one major version and `backward_codecs` keeps the v0 read path alive. So we can keep upgrading `lucene-core` while leaving this serialization exactly as it is.

The constants aren't really ours to sync anyway. The codec names and extensions are a handshake with the reader. `checkIndexHeader` string-matches the name and the extensions pick the files, so Lucene can't change them without breaking every existing index. And `VERSION_CURRENT` is the version *we emit*, not Lucene's. Bumping it while leaving `writeField`/`writeMeta` untouched is the single change most likely to actually break things: we'd stamp a new version number onto old-layout bytes, and the reader would misparse them silently rather than reject them.

When we do decide to move to a newer format, it won't be a constants update - we'll have to re-examine the whole binary serialization. So it's safer to leave the constants pinned until we're ready to do that.

What this paragraph should say instead, I think: on upgrade the only thing to verify is that the stock reader still accepts what we write - which `TestNativeFlatVectorsWriterRoundTrip` already does, against whatever reader is on the classpath.

java/cuvs-lucene/src/test/java/com/nvidia/cuvs/lucene/TestNativeFlatVectorsWriterFormatConstants.java
```diff
@@ -0,0 +1,46 @@
...
+ * {@code lucene-core} jar actually on the classpath (its manifest {@code Specification-Version},
+ * not a parse of {@code pom.xml}'s text), so a Lucene upgrade fails this test immediately rather
+ * than silently writing {@code .vec}/{@code .vemf} files the stock
+ * {@code Lucene99FlatVectorsReader} may no longer read correctly.
+ *
+ * <p>This intentionally fires on <em>any</em> version change, not just ones that actually alter
+ * the format — per the class javadoc, every upgrade needs a manual re-verification pass against
+ * the new version's {@code Lucene99FlatVectorsFormat}/{@code Lucene99FlatVectorsWriter} sources.
+ */
+public class TestNativeFlatVectorsWriterFormatConstants {
```
imotov: I don't think this test based on my previous comment on NativeFlatVectorsWriter.

java/cuvs-lucene/src/test/java/com/nvidia/cuvs/lucene/TestNativeFlatVectorsWriterRoundTrip.java
```diff
@@ -0,0 +1,135 @@
...
+                + " Lucene99FlatVectorsReader",
+            dataset[Integer.parseInt(id)],
+            roundTripped,
+            0f);
+        seen++;
+      }
+      assertEquals("did not visit every vector", numDocs, seen);
+    }
+  }
+}
```
imotov, line 279: I think we could use a couple more tests here:
- The same round-trip test with 10000 docs, but with only a random 8000 of them carrying the vector field. Cardinality above `IndexedDISI.MAX_ARRAY_LENGTH` (4095) selects the DENSE encoding path.
- The same round-trip test with 1000 docs and a random 800 carrying the field, which stays under that threshold and takes the SPARSE path instead.
- A test with more than one vector field, or a tweak to one of the existing tests, so we cover the boundary between consecutive per-field records in `.vemf`.

java/cuvs-lucene/src/main/java/com/nvidia/cuvs/lucene/AcceleratedHNSWUtils.java
```diff
@@ -235,56 +243,149 @@ private static CuVSMatrix buildCagraGraphForSubset(
...
+          final int slot = t;
+          if (subStart >= subEnd) {
             continue;
           }
-          scratch[actualSize++] = nnodes[i] - nnodes[i - 1];
+          futures.add(
+              pool.submit(
+                  () -> {
+                    ByteBuffersDataOutput buffer = new ByteBuffersDataOutput();
+                    int[] scratch = new int[graph.maxConn() * 2];
```
imotov: `graph.maxConn()` is not trivial. Might makes sense to extract it outside of the loop.

java/cuvs-lucene/examples/src/main/java/com/nvidia/cuvs/lucene/examples/OptimizedCagraHnswBuildExample.java
```diff
@@ -0,0 +1,657 @@
...
+    // disable RAM-based flushing so nothing flushes before commit, and forbid merges. This is what
+    // makes native flat buffering valid. Order matters: enable the doc-count trigger BEFORE
+    // disabling
+    // the RAM trigger, since Lucene rejects a config where both are disabled at once.
+    IndexWriterConfig config =
+        new IndexWriterConfig()
+            .setCodec(codec)
+            .setUseCompoundFile(false)
+            .setMaxBufferedDocs(Math.max(2, size + 1))
+            .setRAMBufferSizeMB(IndexWriterConfig.DISABLE_AUTO_FLUSH)
```
imotov: This is a pretty strong requirement that might not be acceptable by Solr and Elasticsearch.

imotov (commented):
Sorry for the delay. I wanted to better understand the logic and explore a couple of alternatives.

java/cuvs-lucene/src/main/java/com/nvidia/cuvs/lucene/AcceleratedHNSWUtils.java
jamxia155: Updated as suggested. Thanks.

java/cuvs-lucene/src/main/java/com/nvidia/cuvs/lucene/NativeFlatVectorsWriter.java
jamxia155: You're right that `Lucene99FlatVectorsFormat` should retain compatibility for the major version. Revised the tests/documentation and removed unnecessary tripwire tests.

java/cuvs-lucene/src/test/java/com/nvidia/cuvs/lucene/TestNativeFlatVectorsWriterFormatConstants.java
jamxia155: Removed.

java/cuvs-lucene/src/test/java/com/nvidia/cuvs/lucene/TestNativeFlatVectorsWriterRoundTrip.java
jamxia155: Added tests as suggested. Thanks!

java/cuvs-lucene/src/main/java/com/nvidia/cuvs/lucene/AcceleratedHNSWUtils.java
jamxia155: Great spot! Extracted outside loop here and in one more instance.

java/cuvs-lucene/examples/src/main/java/com/nvidia/cuvs/lucene/examples/OptimizedCagraHnswBuildExample.java
jamxia155: Updated documentation to highlight the prerequisites for using the native flat buffering optimization. Now framed as an opportunistic optimization that is constrained by multiple pre-conditions.

java/cuvs-lucene/src/main/java/com/nvidia/cuvs/lucene/CagraIndexParamsFactory.java
```diff
@@ -18,6 +25,121 @@ public class CagraIndexParamsFactory {
 
   private CagraIndexParamsFactory() {}
 
+  /**
+   * Translation of the internal logic found in the constructor of {@code struct ivf_pq_params} in
+   * cuVS's {@code cpp/include/cuvs/neighbors/ivf_pq.hpp}.
```
imotov: Ok, I am really confused now. We just removed this in https://github.com/NVIDIA/cuvs-lucene/pull/177/changes#diff-b2d360e425d107d219d1a101fb52f6fdcb1643229dfda96c40dae7e01904c899L31 in order to delegate this logic to cuVS, we do we resurrect it again here?

examples/java/src/main/java/com/nvidia/cuvs/lucene/examples/OptimizedCagraHnswBuildExample.java
```diff
@@ -0,0 +1,669 @@
+/*
```
cjnolet, line 1: Let's please move this `examples/java`. cuvs lucene is a library.

java/cuvs-lucene/examples/pom.xml
```diff
@@ -92,6 +92,11 @@
         <artifactId>lucene-backward-codecs</artifactId>
         <version>10.2.0</version>
     </dependency>
+    <dependency>
```
cjnolet: `examples/java` is going to need to build its own jar (just how the examples for the other languages build their own artifacts)

examples/java/README.md
```diff
@@ -32,3 +32,17 @@ To run the Index and Search on GPU example do:
 ```sh
 mvn clean install && java -Djava.util.logging.config.file=src/main/resources/logging.properties -cp target/examples-26.10.0-jar-with-merged-services.jar com.nvidia.cuvs.lucene.examples.IndexAndSearchonGPUExample
 ```
+
+To run the optimized CAGRA-HNSW build example (reference pattern for efficiently building an
```
cjnolet, line 36: Please move to examples/java

examples/java/src/main/java/com/nvidia/cuvs/lucene/examples/OptimizedCagraHnswBuildExample.java
jamxia155: Moved.

java/cuvs-lucene/src/main/java/com/nvidia/cuvs/lucene/Lucene99AcceleratedHNSWVectorsWriter.java
```diff
@@ -93,6 +109,13 @@ public Lucene99AcceleratedHNSWVectorsWriter(
     this.flatVectorsWriter = flatVectorsWriter;
     this.infoStream = state.infoStream;
     this.acceleratedHNSWParams = acceleratedHNSWParams;
+    this.numInputVectors = acceleratedHNSWParams.getNumInputVectors();
+    this.nativeMode = numInputVectors > 0;
```
imotov: When `nativeMode` here is `true` this writer doesn't actually follow the contract of KnnVectorsWriter faithfully, which means it shouldn't be used by anything but CagraHnswBulkIndexWriter. So, I think instead of branching inside more universal Lucene99AcceleratedHNSWVectorWriterit should be a separate class implementing KnnVectorsWriter, preferably package private is it can only be instantiated within the limits of CagraHnswBulkIndexWriter. I would also remove getNumInputVectors() from acceleratedHNSWParams while at it. I don't think we need to do it as part of this PR, so if you don't have time to get to it, please open an issue for it and assign it to me, so we don't lose it.

java/cuvs-lucene/src/main/java/com/nvidia/cuvs/lucene/CagraHnswBulkIndexWriter.java
```diff
@@ -0,0 +1,660 @@
...
+      buildOverlapped(fbinPath, config, callback, chunkSizeMB, slices, dim);
+    } else {
+      try (FbinVectorSource source = new FbinVectorSource(fbinPath, chunkSizeMB)) {
+        buildSequential(source, config, callback, slices);
+      }
+    }
+  }
+
+  private static void ensureRMMEnabled() {
+    if (RMM_ENABLED.compareAndSet(false, true)) {
```
imotov: That looks like a race condition. The second and the following threads that will reach this point are not going to wait for the first thread to actually set RMM. Instead of setting it here, I would rather check that it is already set and bail out with an error if it is not.

java/cuvs-lucene/src/main/java/com/nvidia/cuvs/lucene/FieldWriter.java
```diff
@@ -23,18 +25,64 @@ public class FieldWriter extends KnnFieldVectorsWriter<Object> {
...
+  private final int numInputVectors;
+
+  private final boolean nativeBuffering;
+  private final CuVSMatrix.Builder<CuVSHostMatrix> hostMatrixBuilder;
+  private final DocsWithFieldSet nativeDocsWithField;
+  private int nativeCount;
+  private CuVSHostMatrix builtMatrix;
+
   @SuppressWarnings("unchecked")
   public FieldWriter(
```
imotov, line 32: I think this should be also a separate NativeFieldWriter it is just too different from the "normal" one.

java/cuvs-lucene/src/main/java/com/nvidia/cuvs/lucene/Lucene99AcceleratedHNSWVectorsWriter.java
```diff
@@ -213,6 +280,37 @@ public void flush(int maxDoc, DocMap sortMap) throws IOException {
...
+              + " numInputVectors vectors were added; see AcceleratedHNSWParams.Builder"
+              + "#withNumInputVectors.");
+    }
+    FieldInfo fieldInfo = fieldData.fieldInfo();
+    try {
+      CuVSHostMatrix dataset = fieldData.getHostMatrix();
+      nativeFlat.writeField(fieldInfo, dataset, maxDoc, fieldData.getDocsWithFieldSet());
+      writeFieldInternal(fieldInfo, dataset);
+    } finally {
+      fieldData.releaseNativeBuffer();
```
imotov: This might not be the most reliable place to release the native buffer. It looks like if we have wrong numInputVectors or sort key or any other error situation, we never reach this point.

java/cuvs-lucene/src/main/java/com/nvidia/cuvs/lucene/FbinVectorSource.java
```diff
@@ -0,0 +1,213 @@
...
+    current = next;
+  }
+
+  @Override
+  public void get(int index, float[] dst) throws IOException {
+    if (index < 0 || index >= windowSize) {
+      throw new IndexOutOfBoundsException(
+          "Index " + index + " out of bounds [0, " + windowSize + ")");
+    }
+    if (index < nextExpectedRelative) {
```
imotov, line 172: VectorSource docs say "index must be greater than **or equal** to the index passed to the previous call, if any". That seems a bit inconsistent.

imotov (commented):
Looks much better! Thanks! Left a few comments.

java/cuvs-lucene/src/main/java/com/nvidia/cuvs/lucene/FbinVectorSource.java
jamxia155: Fixed docs (code was correct).

java/cuvs-lucene/src/main/java/com/nvidia/cuvs/lucene/Lucene99AcceleratedHNSWVectorsWriter.java
jamxia155: Add a `releaseAllNativeBuffers()` backstop to `close()`.

java/cuvs-lucene/src/main/java/com/nvidia/cuvs/lucene/FieldWriter.java
jamxia155: Split as suggested.

java/cuvs-lucene/src/main/java/com/nvidia/cuvs/lucene/Lucene99AcceleratedHNSWVectorsWriter.java
jamxia155: Split as suggested.

java/cuvs-lucene/src/main/java/com/nvidia/cuvs/lucene/CagraHnswBulkIndexWriter.java
jamxia155: This is indeed a race condition and even worse, it goes against our existing convention of leaving RMM enablement as an opt-in in application code. Sticking to the convention naturally resolves this as the typical pattern is to enable RMM before any GPU resources are created.

Also updated the documentation for `enableRMM*` and `resetRMMPooledMemory` methods to clearly advise against self-enable/reset within classes.

examples/java/README.md
imotov: @cjnolet there is no examples/java at the moment - https://github.com/NVIDIA/cuvs/tree/main/examples. Java examples currently live in two places: [`java/examples`](https://github.com/NVIDIA/cuvs/tree/main/java/examples) for generic cuVS examples and [`java/cuvs-lucene/examples`](https://github.com/NVIDIA/cuvs/tree/main/java/cuvs-lucene/examples) for Lucene-specific examples. I have it on my [to-do list](https://github.com/NVIDIA/cuvs/issues/2288) to move them into /examples/java in the near future, but I don't think we should do it as part of this commit since it would (at least temporarily) introduce a third place where Java examples can be found.

jameslamb (approved):
Approving for CI/packaging codeowners groups (CI/packaging changes look minimal and non-controversial).

imotov (approved):
LGTM, just waiting for #2543 to be merge so we don't spread examples all over the place.

Issue #2482, opened 2026-08-18
[BUG] cuvsDatasetMakePadded rejects an already-aligned device tensor


imotov:
**Describe the bug**

`cuvsDatasetMakePadded` fails when handed a device tensor whose rows already sit at CAGRA's required stride. The refusal makes sense since it avoids a redundant copy. However, it is not documented in the API, and the C API exposes no way to predict it.

**Steps/Code to reproduce bug**

1. Allocate a contiguous float32 device matrix whose dimension is a multiple of 4 — e.g. 64 rows × 128 dims. (Contiguous means `rowStride == columns`, which for 4-byte elements is already the 16-byte-aligned width CAGRA requires.)
2. Wrap it in a `DLManagedTensor` with device type `kDLCUDA`.
3. Call `cuvsDatasetMakePadded(res, tensor, CUVS_DATASET_MEM_TYPE_DEVICE, &out)`.
4. Repeat with 127 dims instead of 128 — it succeeds. The boundary is `dim % 4 == 0` for float32.

**Expected behavior**

The dataset contents are copied into newly allocated padded storage, as documented at https://github.com/NVIDIA/cuvs/blob/2140532c5274dfbd9ba1d18c7bbdac15cc7ea93a/c/include/cuvs/core/dataset.h#L62-L63

**Actual behavior**

`CUVS_ERROR`, with `cuvsGetLastErrorText()` returning:

```
RAFT failure at cpp/include/cuvs/neighbors/common.hpp:1162:
source is device and stride is already correct.
Use make_device_padded_dataset_view() to get a view instead.
```

**Environment details**

Not environment specific. Present on `main` and `release/26.08`:
https://github.com/NVIDIA/cuvs/blob/2140532c5274dfbd9ba1d18c7bbdac15cc7ea93a/cpp/include/cuvs/neighbors/common.hpp#L1161-L1166

**Additional context**

The Python workaround is probably the clearest signal that the API is missing something:
https://github.com/NVIDIA/cuvs/blob/2140532c5274dfbd9ba1d18c7bbdac15cc7ea93a/python/cuvs/cuvs/common/dataset.pyx#L73-L89

It matches on the error text, so any rewording of that message turns the workaround into a crash. Rust, Go, and Java have no equivalent workaround and propagate the failure.

Might be related: #2402 (Dataset API C++, C, and language wrapper follow-up), #2394 (Remove dataset views from language wrappers).

PR #2483, opened 2026-08-18
Fix silent brute force fallback at aligned vector dimensions

imotov:
cuvsDatasetMakePadded rejects a device matrix already at CAGRA's required stride, so every field whose dimension is a multiple of four failed its CAGRA build and was silently indexed as brute force. Pick the factory that matches the layout, and log the fallback so the next one is visible.

See #2482


java/cuvs-lucene/src/main/java/com/nvidia/cuvs/lucene/CuVS2510GPUVectorsWriter.java
```diff
@@ -206,6 +206,14 @@ private void writeFieldInternal(FieldInfo fieldInfo, List<float[]> vectors) thro
...
         } catch (Throwable t) {
           // Fallback to brute force in a few cases, for now.
+          // Log it to make it more obvious that this is what is happening.
+          info(
+              infoStream,
+              COMPONENT,
+              "CAGRA build failed for field \""
+                  + fieldInfo.name
+                  + "\", falling back to a brute force index: "
+                  + t);
```
mythrocks, line 216: 👍

mythrocks (approved):
Generally 👍.  A design consideration suggested for the future.

PR #2484, opened 2026-08-18
[FEA] Truly Decoupled Double Buffering for SG KMeans

tarang-jain:
Screenshot 2026-08-18 at 4 44 05 PM
Profile:
H100 + Intel(R) Xeon(R) Platinum 8480CL

Dataset mem type = pinned memory. This speeds up training by a lot. nsys gave throughput for H2D transfers around 50 GB/s. This is a 114 GB dataset. The screenshot above shows the overlap between the two streams with prefetch.

| Metric | Baseline (2M batch) | Prefetch (2M batch) | Baseline (4M batch) |
|---|---:|---:|---:|
| profile_mode | baseline | prefetch | baseline |
| dataset | base_falcon_1024_30M.fbin | base_falcon_1024_30M.fbin | base_falcon_1024_30M.fbin |
| rows | 30,000,000 | 30,000,000 | 30,000,000 |
| dimensions | 1,024 | 1,024 | 1,024 |
| clusters | 2,048 | 2,048 | 2,048 |
| batch_rows | 2,000,000 | 2,000,000 | 4,000,000 |
| requested_iterations | 20 | 20 | 20 |
| pinned_load_seconds | 66.3182 | 59.2441 | 60.1955 |
| initialization_seconds | 0.0169335 | 0.0170245 | 0.0172878 |
| fit_seconds | 101.064 | 59.8278 | 101.737 |
| fit speedup vs. 4M | **1.007x** | **1.700x** | **1.000x** |
| e2e_seconds | 168.422 | 120.27 | 162.921 |
| e2e speedup vs. 4M | **0.967x** | **1.355x** | **1.000x** |
| iterations | 20 | 20 | 20 |
| inertia | 2.14326e+07 | 2.14325e+07 | 2.14325e+07 |

Also the use of pinned memory shows > 50 GB/s of throughput for H2D transfers. This gets much lower with regular host mem.
Screenshot 2026-08-18 at 5 03 57 PM


cpp/src/cluster/detail/kmeans.cuh
```diff
@@ -686,25 +688,58 @@ void kmeans_fit(
...
   rmm::device_uvector<char> batch_workspace(device_buffer_samples, stream);
 
-  auto data_batches = cuvs::spatial::knn::detail::utils::make_batch_load_iterator<DataT>(
-    handle, X.data_handle(), n_samples, n_features, device_buffer_samples, stream);
+  auto batch_memory = raft::resource::get_workspace_resource_ref(handle);
+  if constexpr (!data_on_device) {
+    size_t batch_staging_bytes =
+      static_cast<size_t>(device_buffer_samples) * static_cast<size_t>(n_features) * sizeof(DataT);
+    if (weight_ptr != nullptr) {
+      batch_staging_bytes += static_cast<size_t>(device_buffer_samples) * sizeof(DataT);
```
viclafargue: The decision to go for a large workspace should account for double buffering.

```
const size_t buffer_count = enable_batch_prefetch && device_buffer_samples < n_samples ? 2 : 1;
size_t batch_staging_bytes = buffer_count * static_cast<size_t>(device_buffer_samples) * static_cast<size_t>(n_features) * sizeof(DataT);
if (weight_ptr != nullptr)
    batch_staging_bytes += buffer_count * static_cast<size_t>(device_buffer_samples) * sizeof(DataT);
```

viclafargue (commented):
Great work, the reported speedup looks promising! To make the changes easier to review and validate, could we keep this PR focused on double-buffered prefetching? The workspace-resource selection, final-inertia optimization, and cluster-cost refactoring could be handled in separate follow-up PRs, allowing each change to be tested and benchmarked independently.

cpp/src/cluster/detail/kmeans.cuh
```diff
@@ -686,25 +688,58 @@ void kmeans_fit(
...
+    }
+    if (batch_staging_bytes > raft::resource::get_workspace_free_bytes(handle)) {
+      batch_memory = raft::resource::get_large_workspace_resource_ref(handle);
+    }
+  }
+
+  // Use the caller's stream pool for host-input staging. If no pool is configured, this falls
+  // back to the main stream and disables cross-stream prefetching.
+  auto [batch_copy_stream, enable_batch_prefetch] =
+    cuvs::spatial::knn::detail::utils::get_prefetch_stream(handle);
```
viclafargue, line 708: Using the caller’s stream pool follows the cuVS pattern, but makes double buffering difficult to discover. A boolean parameter would be clearer (see [#2323](https://github.com/NVIDIA/cuvs/pull/2323)), though API changes may be forbidden in upcoming releases. Could we instead always enable double buffering for multi-batch host input using an internal non-blocking stream?

tarang-jain: I am not opposed to a param for opt-in prefetch, but I am opposed to creating a stream on the user's behalf for 2 reasons:
1. The raft handle should be the way for the user to control streams and scheduling work on the GPU. Creating one internally takes this control away from the user.
2. It sends the wrong message about memory footprint, which will now be 2x the memory footprint they compute using the device_buffer_samples. I also need to update docs for device_buffer_samples for this.

cpp/src/cluster/detail/kmeans.cuh
tarang-jain: Yes I will do this and move it to the workspace PR. We can merge that PR after this one.

cpp/src/cluster/detail/kmeans.cuh
viclafargue: Ok, but let's document clearly in the public OOC KMeans API that a user would have set up a stream pool to enable double buffering. Also, we would need to check if setting up a stream pool does not have unforeseen effects in primitives that KMeans is using?

tarang-jain: Updated the docs. Stream creation is cheap. Stream pool does not have unforeseen effects on any other KMeans primitives. We follow the same pattern as IVF and CAGRA.

PR #2361, opened 2026-07-24
Brings the Java build in line with other clients.

imotov:
Fixes several build issues that previously allowed improperly formatted files and files with incorrect copyright headers to pass CI.
- Replaces the Java-specific copyright check with the standard pre-commit check.
- Adds Java and Maven dependencies to the all configuration.
- Moves formatting and Spotless checks into pre-commit.
- Fixes formatting in files that were not properly formatted.
- Removes the remaining headers_h_1 imports.

Fixes #2131

PR #2485, opened 2026-08-18
Make run_spotless.sh more verbose and strict

imotov:
Some developers ran into issues with spotless-fmt skipping formatting silently on their local machines and then failing in CI. This PR now fails if there are any modified Java files and prints a warning if there are no changed Java files and Maven is not present.

Follow-up for #2361

### DIFF
--- a/ci/checks/run_spotless.sh
+++ b/ci/checks/run_spotless.sh
@@ -5,19 +5,36 @@
 # pre-commit hook wrapper that runs 'spotless:apply' to format the Java sources of every Maven
 # project under java/.
 #
-# Most cuvs contributors do not work on the Java client and do not have Maven installed. For them
-# (running outside CI without Maven) this skips gracefully, so that 'pre-commit run --all-files'
-# does not require every contributor to install Maven. In CI, Maven is expected to be available and
-# its absence is treated as an error.
+# Most cuvs contributors do not work on the Java client and do not have Maven installed. For them,
+# running 'pre-commit run --all-files' matches every Java source file in the repo regardless of
+# whether they touched any of it, so this skips gracefully when Maven is missing and there are no
+# actual local changes to Java sources. In CI, and for anyone who has actually modified Java
+# sources locally, Maven is expected to be available and its absence is treated as an error.
 
 set -euo pipefail
+
+# Keep these in sync with the spotless-fmt hook's 'files'/'exclude' entries in
+# .pre-commit-config.yaml.
+JAVA_SRC_PATTERN='^java/(cuvs-java|cuvs-lucene)/([^/]+/)?src/.*\.java$'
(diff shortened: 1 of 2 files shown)


ci/checks/run_spotless.sh
```diff
@@ -5,19 +5,36 @@
...
     echo "spotless-fmt: 'mvn' is required in CI but was not found on PATH." >&2
     exit 1
   fi
-  echo "spotless-fmt: skipping Java formatting ('mvn' not installed and not running in CI)." >&2
+  if java_sources_modified; then
+    echo "spotless-fmt: 'mvn' is required to format modified Java sources but was not found on PATH." >&2
+    exit 1
+  fi
+  echo "spotless-fmt: 'mvn' was not found on PATH and no Java sources were modified, skipping Java formatting." >&2
   exit 0
```
jamxia155, line 38: Nit: logically these 2 lines belong in an else branch after `if java_sources_modified`.

ci/checks/run_spotless.sh
```diff
@@ -5,19 +5,36 @@
...
 
 set -euo pipefail
 
+# Keep these in sync with the spotless-fmt hook's 'files'/'exclude' entries in
+# .pre-commit-config.yaml.
+JAVA_SRC_PATTERN='^java/(cuvs-java|cuvs-lucene)/([^/]+/)?src/.*\.java$'
+JAVA_SRC_EXCLUDE='.*/panama/.*'
+
+java_sources_modified() {
+  git status --porcelain --untracked-files=all -- java/cuvs-java java/cuvs-lucene |
```
jamxia155, line 22: How is this supposed to flag changed files in this PR? Wouldn't the output just be empty for a fresh checkout without modifications?

ci/checks/run_spotless.sh
imotov: If I do that and somebody takes java code, modifies it, commits it and runs `pre-commit run --all-files` after that they will not know why their code fails in CI. Because pre-commit will simply say "Passed".

ci/checks/run_spotless.sh
imotov: This will only work for changed files that have been staged. Unfortunately, I don't have a good way to check for changed Java files that have already been committed. That's why I placed the warning you commented on above where it is.

jamxia155: Gotcha. This looks good then.

jameslamb (approved):
So if I understand it, this makes it like "if you're touching a Java file, you have to have `maven` installed for pre-commit to work"?

Fair enough, not going to get in the way of that. You might face pushback from folks who make C++ changes and mechanically apply changes to the Java bindings based on those, but definitely worth experimenting with.

Issue #2440, opened 2026-08-12
[FEA][C++][C][Language Wrappers] Rename vpq to pq for datasets, factories, and params


HowardHuang1:
VPQ doesn't exist. This was bad naming. 

VQ is a very primitive method that is not used for vector search.

use_vq param should be renamed "train_coarse"

Only PQ is used for vector search.
PQ is kmeans cluster assignment on each fragment of the original vector after the original vector is split into fragments.

PR #2494, opened 2026-08-20
Rename vpq to pq and use_vq to train_coarse for datasets, factories, and params

HowardHuang1:
Addresses #2440.

VPQ was never a real, named technique — it's just PQ, optionally preceded by a coarse centroid-assignment step. That coarse step also isn't true vector quantization (true VQ replaces a vector with its centroid ID; this assigns a centroid label alongside the still-intact PQ-encoded residual, i.e. an IVF-style coarse assignment). Both names were misleading, hence this rename.

Rules followed:
- "vpq" does not describe anything real and is removed everywhere, public and private. Public API surface must not contain it (hard requirement); private usage isn't required to change, but we removed it there too for consistency rather than leave a subset of the confusing name behind.
- "VQ"/"use_vq" as a *toggle name* is misleading (implies real VQ is happening) and is renamed to `train_coarse`. "VQ" as a *quantity name* (vq_n_centers, vq_codebook, vq_labels, train_vq()/predict_vq()) is left alone — it's real, understood terminology for the coarse-centroid concept itself, not the part that was wrong.
- Known, not-yet-implemented follow-up (explicitly out of scope for this change): `train_coarse` should be forced `false` for CAGRA and only ever `true` for IVF-style use, since the coarse step only makes sense when building an IVF index. SCANN and VAMANA already set `train_coarse = false` at their own call sites; CAGRA's own compression entry point (`make_pq_dataset()` / `pq_build()`) still hardcodes `true` and doesn't expose the field at all on `cuvs::neighbors::pq_params`. CAGRA also does not yet reject a PQ dataset built with coarse centroids enabled. Both are behavioral changes, not naming, and are left for a follow-up PR.

Renamed (types, functions, files):
- vpq_params -> pq_params; vpq_f16_index/vpq_f32_index -> pq_f16_index/pq_f32_index
- vpq_dataset(_view/_container), device_vpq_dataset(_view), host_vpq_dataset(_view) -> pq_dataset equivalents; is_vpq_dataset(_v) -> is_pq_dataset(_v)
- dataset_view_kind::vpq_f16/vpq_f32 and the is_{device,host,}_vpq_{f16,f32}_* predicate family -> pq equivalents
- make_vpq_dataset() factory -> make_pq_dataset(); vpq_train_from_rows -> pq_train_from_rows; vpq_build/vpq_build_half -> pq_build/pq_build_half; vpq_codebooks field -> pq_codebooks
- compute_distance_vpq* files/functions/JIT-LTO symbols (dispatch_cagra_vpq_*, is_vpq, query_type_tag_vpq_t, codebook_tag_vpq_t, vpq_descriptor_spec, vpq_smem_value_config) -> pq equivalents; files vpq_dataset.cuh/vpq_utils.cuh/compute_distance_vpq* -> pq*
- use_vq (bool) -> train_coarse, across the C++ struct member, C struct field, the C API/ABI getter cuvsProductQuantizerGetUseVq -> cuvsProductQuantizerGetTrainCoarse, the Python kwarg/property, and the Rust FFI binding
- Two private helper functions, initially collapsed into ambiguous names by the mechanical vpq->pq substitution (both ended up called "pq_params" despite converting between two distinct structs), renamed for clarity by destination-namespace: to_vpq_params -> to_neighbors_pq_params (-> cuvs::neighbors::pq_params) make_pq_params_from_vpq -> to_quantize_pq_params (-> cuvs::preprocessing::quantize::pq::params)
- pq_codebooks: `pq_codebooks` on `quantizer<T>` is PQ's codebook plus an optional VQ codebook when `train_coarse` is set (empty otherwise). Clarified this in the struct's doc comment, in cpp/include/cuvs/preprocessing/quantize/pq.hpp and the mirrored fern doc.

Left unchanged (real, non-misleading terminology, not in scope):
- vq_n_centers, vq_kmeans_trainset_fraction, max_train_points_per_vq_cluster (fields)
- vq_codebook, vq_labels, vq_code_book, vq_centers (fields/locals)
- train_vq()/predict_vq() (private detail:: helpers performing the actual coarse training)
- cpp/include/cuvs/preprocessing/quantize/pq.hpp / pq.cu — a separate, already-correctly-named standalone PQ preprocessing feature; never had "vpq" in its own naming

Breaking changes (all reviewed/approved):
- C++ API: CAGRA's pq_* type/alias family (source-level break; zero C API/C ABI surface touched, confirmed no C-layer symbol referenced "Vpq")
- C API + C ABI: cuvsProductQuantizerGetUseVq -> cuvsProductQuantizerGetTrainCoarse (the one change ci/check_c_abi flags as breaking)
- Python: use_vq kwarg/property hard-renamed to train_coarse, no back-compat alias
- Rust: cuvs-sys bindings regenerated to match (bindgen output verified byte-identical to the hand-edit)


cpp/include/cuvs/neighbors/cagra.hpp
```diff
@@ -1731,11 +1731,11 @@ void search(raft::resources const& res,
  * @param[in] sample_filter an optional device filter function object that greenlights samples
  * for a given query. (none_sample_filter for no filtering).
  *
- * @note FP32 VPQ search is declared for ABI stability but fails at runtime until implemented.
+ * @note FP32 PQ search is declared for ABI stability but fails at runtime until implemented.
```
aamijar, line 1734: Wait, why do we have this function if it is known to fail? We are guaranteeing ABI stability at C layer not C++ layer.

aamijar (changes requested):

c/src/preprocessing/quantize/pq.cpp
```diff
@@ -299,14 +299,14 @@ extern "C" cuvsError_t cuvsProductQuantizerGetEncodedDim(cuvsProductQuantizer_t
   });
 }
 
-extern "C" cuvsError_t cuvsProductQuantizerGetUseVq(
-  cuvsProductQuantizer_t quantizer, bool* use_vq)
+extern "C" cuvsError_t cuvsProductQuantizerGetTrainCoarse(
```
cjnolet, line 302: Sorry to do this again- can we maybe use the term "use_coarse" instead of "train coarse"? I think that might make this make more sense.

cjnolet: Hopefully it's a quick search and replace across the codebase.

PR #2497, opened 2026-08-21
Lucene: Merge CAGRA indexes on the GPU instead of rebuilding from vectors

### LINKED ISSUE #1078: [FEA] `merge()` Java API for Tiered Index

imotov:
mergeOneField now hands the segments' CAGRA indexes to the cuVS merge API, falling back to the vector based merge when the merged rows cannot be lined up with the merged flat vectors. Deleted documents are dropped through a row, which needs a new CagraIndex.merge overload taking a BitSet. Also cleans up some merge overloads on CuVSProvider and CagraIndex.

Closes #1078


java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/CagraIndexImpl.java
```diff
@@ -221,11 +221,14 @@ private IndexReference build(CagraIndexParams indexParameters, CuVSMatrixInterna
   private static MemorySegment createCagraIndex() {
     try (var localArena = Arena.ofConfined()) {
       MemorySegment indexPtrPtr = localArena.allocate(cuvsCagraIndex_t);
-      // cuvsCagraIndexCreate gets a pointer to a cuvsCagraIndex_t, which is defined as a pointer to
+      // cuvsCagraIndexCreate gets a pointer to a cuvsCagraIndex_t, which is defined
+      // as a pointer to
```
mythrocks: Nit:  Inadvertent whitespace change?  Formatting seems off here.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/CagraIndexImpl.java
```diff
@@ -221,11 +221,14 @@ private IndexReference build(CagraIndexParams indexParameters, CuVSMatrixInterna
...
       // cuvsCagraIndex.
-      // It's basically an "out" parameter: the C functions will create the index and "return back"
+      // It's basically an "out" parameter: the C functions will create the index and
+      // "return back"
       // a pointer to it: (*index = new cuvsCagraIndex{};
-      // The "out parameter" pointer is needed only for the duration of the function invocation (it
+      // The "out parameter" pointer is needed only for the duration of the function
+      // invocation (it
```
mythrocks: Nit: Inadvertent whitespace change? Formatting seems off here.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/CagraIndexImpl.java
```diff
@@ -923,69 +935,150 @@ public static CagraIndex.Builder newBuilder(CuVSResources cuvsResources) {
...
       for (int i = 0; i < indexes.length; i++) {
         CagraIndexImpl indexImpl = (CagraIndexImpl) indexes[i];
         indexesSegment.setAtIndex(
             ValueLayout.ADDRESS, i, indexImpl.cagraIndexReference.getMemorySegment());
+        if (rowFilter != null) {
+          mergedRowCount += indexImpl.size();
+        }
+      }
+      if (rowFilter != null) {
+        if (rowFilter.length() > mergedRowCount) {
```
paul-aiyedun, line 976: Can we update the `rowFilter` Javadoc to explain what happens when  `rowFilter.length() < mergedRowCount`?

java/cuvs-lucene/src/main/java/com/nvidia/cuvs/lucene/CuVS2510GPUVectorsWriter.java
```diff
@@ -391,92 +392,261 @@ static int distFuncToOrd(VectorSimilarityFunction func) {
...
+      }
+      for (int ord = liveRows.nextSetBit(0); ord >= 0; ord = liveRows.nextSetBit(ord + 1)) {
+        rowFilter.set(rowCount + ord);
+      }
+      rowCount += fieldEntry.count();
+      survivingCount += liveCount;
+      readers.add(cuvsReader);
+    }
+    boolean filtered = survivingCount != rowCount;
+    // Merging a single index is only worth it when the filter has rows to drop; without one the
```
paul-aiyedun, line 488: What does the fallback/vector based merge do in this case? Does it involve a host to device copy, and if so, is that better than doing the merge on GPU?

java/cuvs-lucene/src/main/java/com/nvidia/cuvs/lucene/CuVS2510GPUVectorsWriter.java
```diff
@@ -391,92 +392,261 @@ static int distFuncToOrd(VectorSimilarityFunction func) {
...
+      }
+    }
+    return liveRows;
+  }
+
+  /**
+   * Reports why this field cannot go through the cuVS merge API and has to fall back to the vector
+   * based merge.
+   */
+  private void reportNotMergeable(FieldInfo fieldInfo, String reason) {
```
paul-aiyedun: Nit: Consider `logNotMergeable`. When first looking at where this call was made, it was not clear what the side effect of this method call would be (e.g. like throwing a exception).

java/cuvs-java/src/test/java/com/nvidia/cuvs/CagraBuildAndSearchIT.java
```diff
@@ -1007,6 +1007,259 @@ public void testMergingIndexes() throws Throwable {
...
+                  .withIndexParams(indexParams)
+                  .build();
+          CagraIndex index2 =
+              CagraIndex.newBuilder(resources)
+                  .withDataset(vector2)
+                  .withIndexParams(indexParams)
+                  .build()) {
+
+        // No device dataset is attached, so cuVS refuses to merge these.
+        for (int attempt = 0; attempt < 5; attempt++) {
```
paul-aiyedun, line 1218: Why do we do this multiple time?

mythrocks (approved):
I'm mostly 👍 on the change.

It appears to have a change in the `CuVSProvider` public interface.  But that is acceptable.

paul-aiyedun (commented):
Changes overall look good to me. I just had a few questions.

java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/CagraIndexImpl.java
imotov: Great idea! Done.

java/cuvs-lucene/src/main/java/com/nvidia/cuvs/lucene/CuVS2510GPUVectorsWriter.java
imotov: Doing it on GPU is not an option at the moment: https://github.com/NVIDIA/cuvs/blob/399c9fedf8ffcf1e0639a6ea966b90808539baee/cpp/src/neighbors/detail/cagra/cagra_merge.cuh#L240 In this case Java API basically just reflects C++ API.

java/cuvs-lucene/src/main/java/com/nvidia/cuvs/lucene/CuVS2510GPUVectorsWriter.java
imotov: 👍

java/cuvs-java/src/test/java/com/nvidia/cuvs/CagraBuildAndSearchIT.java
imotov: It runs a few times to confirm that a failed merge doesn't corrupt the input indexes - they should still report their row counts and remain valid inputs for a subsequent successful merge. 5 is probably overkill. 2 iterations would prove the same point that failures don't accumulate.

paul-aiyedun (approved):
Looks good to me.

Issue #2326, opened 2026-07-19
[FEA] Add support for Quantized Datasets in NN-descent


lowener:
This feature request is to add to `nn-descent::build()` the support of BBQ/RabitQ Dataset. This is is part of adding support of Quantized dataset for CAGRA.
Related issues:
-  #2175
- #2315
- #1801 
- #851

Issue #2175, opened 2026-06-02
[FEA] BBQ Dataset


cjnolet:
BBQ or Better BInary Quantization is based on RaBitQ and it's the current algorithm supported in Lucene and Elasticsearch. We are being asked by multiple Lucene users to accept a "dataset" of vectors in the CAGRA build() API. The challenge is that the BBQ algorithm requires some additional state to be carried along for correction codes, which need to be considered during search. This means the dataset needs to be specialized for a BBQ quantizer.

Issue #2162, opened 2026-06-02
[FEA] Lucene BBQ preprocessing API


cjnolet:
Lucene has had several iterations of different quantization routines over the years, with one of the most recent routines being BBQ, which is inspired heavily by concepts from RaBitQ. We should support all the routines we need to support so that we can be fully compatible.

PR #2506, opened 2026-08-25
Support of BBQ Dataset View

lowener:
This PR introduce the structures used for the BBQ API. Related PR #2175
Those structures will be used for the support of compressed dataset in CAGRA and NN-Descent (#2326), as well as the BBQ Preprocessing Quantizer (#2162)

The quantizer views, dataset types and aliases are in the `preprocessing/quantize/bbq.hpp` to avoid circular dependencies

### DIFF
--- a/cpp/include/cuvs/neighbors/common.hpp
+++ b/cpp/include/cuvs/neighbors/common.hpp
@@ -308,6 +308,15 @@ using standard_dataset_owning_storage =
 template <typename ViewT, typename DataT, typename IdxT>
 using standard_dataset_view_storage = dense_row_major_dataset_view_storage<ViewT, DataT, IdxT>;
 
+template <typename T, typename IdxT, typename Accessor>
+using dense_owning_vector = std::conditional_t<Accessor::is_device_accessible,
+                                               raft::device_vector<T, IdxT>,
+                                               raft::host_vector<T, IdxT>>;
+
+template <typename T, typename IdxT, typename Accessor>
+using dense_view_vector = std::conditional_t<Accessor::is_device_accessible,
+                                             raft::device_vector_view<T, IdxT>,
+                                             raft::host_vector_view<T, IdxT>>;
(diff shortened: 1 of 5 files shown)


cpp/include/cuvs/neighbors/common.hpp
```diff
@@ -308,6 +308,15 @@ using standard_dataset_owning_storage =
...
 
+template <typename T, typename IdxT, typename Accessor>
+using dense_owning_vector = std::conditional_t<Accessor::is_device_accessible,
+                                               raft::device_vector<T, IdxT>,
+                                               raft::host_vector<T, IdxT>>;
+
+template <typename T, typename IdxT, typename Accessor>
+using dense_view_vector = std::conditional_t<Accessor::is_device_accessible,
+                                             raft::device_vector_view<T, IdxT>,
+                                             raft::host_vector_view<T, IdxT>>;
```
achirkin, line 319: This is a funny one :)

```suggestion
template <typename T, typename IdxT, typename Accessor>
using vector = raft::mdarray<T, raft::vector_extent<IdxT>, raft::layout_c_contiguous,  Accessor>

template <typename T, typename IdxT, typename Accessor>
using vector_view = raft::mdspan<T, raft::vector_extent<IdxT>, raft::layout_c_contiguous,  Accessor>
```

1. Don't use host/device aliases as the base type - just use the accessor itself on the raw type
2. We have an established naming scheme for these: scalar/vector/matrix/mdarray and corresponding {object}_view
3. In the end, they really belong to raft rather than cuvs

You can have a look an mimic the definitions at  https://github.com/NVIDIA/raft/blob/main/cpp/include/raft/core/device_mdarray.hpp and https://github.com/NVIDIA/raft/blob/main/cpp/include/raft/core/device_mdspan.hpp

tarang-jain (approved):

lowener:
Closing this as PR #2654 was merged instead.

Issue #2401, opened 2026-08-03
[FEA][C API] Migrate merge() Dataset API to have users pass in concatenated dataset + integer offset.


HowardHuang1:
Migrate merge() contract the same way we did with extend().

Users should be responsible with concatenating the 2 datasets into 1 merged dataset and provide that single merged dataset + integer offset. Merge() should then merge graph only.

Remove merged_storage struct.

Think about how to pre-calculate merged storage size after bitset filter:
https://github.com/NVIDIA/cuvs/pull/1846#discussion_r3639389310

PR #2507, opened 2026-08-26
Migrate CAGRA merge() to caller-concatenated dataset + offsets contract

HowardHuang1:
Addresses issue #2401. 

merge() previously allocated the merged dataset buffer and copied each input index's rows into it internally, hiding memory allocation and copies from the caller. This migrates it to the same explicit caller-owned-buffer contract already used by extend(): callers concatenate every input index's dataset (applying any row filter) into a single buffer themselves, and pass per-index offsets marking where each index's rows start. merge() now only builds/merges the graph and binds the returned index to that buffer.

A new merged_dataset_offsets() helper is added for the bitset-filtered case, where the caller can't otherwise derive per-index surviving row counts; unfiltered callers can compute offsets directly (cumulative index sizes) without it.

- C++: cagra::merge() and Fastener/rebuild internals updated; new cagra::merged_dataset_offsets().
- C API: cuvsCagraMerge/cuvsCagraMergeWithParams updated to match; new cuvsCagraMergedDatasetOffsets.
- Java: CagraIndex.merge() public API updated to the same contract (breaking); real in-repo consumer (cuvs-lucene) and tests updated.
- Rust, Python, Go: new merge wrappers added (none existed before), matching the C API's contract and modeled on each language's existing extend()/update_dataset() conventions.

Also fixes an unrelated pre-existing build break in cpp/src/core/bloom_filter.cu (cuco::default_filter_policy renamed to cuco::bloom_filter_policy in the pinned cuCollections version), needed to get any build of libcuvs compiling in this environment.


aamijar (commented):
Hi Howard, as discussed offline, let's consider the purpose of this merge PR change.
We already enforce that the user must pre allocate and pass in the merged_dataset storage. The merge() api will populate that storage.
I'm a little confused as to what we are trying to achieve in this PR.

HowardHuang1:
/ok to test [a14a1da](https://github.com/NVIDIA/cuvs/pull/2507/commits/a14a1da93f135f7b0a1897b8b0bad997ec06f1ae)

HowardHuang1:
This PR is more about memory savings. It is also meant to mirror how we altered the extend API. Merge should now be merging graph only. 

There are 2 cases: unfiltered and filtered merge. 
- Unfiltered merge: Peak memory is the same in main and this PR.
- Filtered merge: The memory win is the filtered path.

On main: 
1. You allocate GPU storage for surviving rows (final_rows).
2. Merge allocates a second GPU staging buffer for the unfiltered concat (all rows of every input).
3. It copies every input dataset into that staging buffer (deleted/filtered rows included).
4. It applies the bitset and copies survivors into your smaller buffer.
5. It builds the graph on that smaller buffer.
- Peak GPU memory is roughly unfiltered concat + filtered result, plus CSR/index scratch for the bitset. The large staging buffer is merge’s, not yours, and it exists only to throw most of it away.

This PR:
- No intermediate staging buffer is created. 
1. Concat is done on CPU by the caller
2. You allocate only the surviving filtered rows on GPU
- Peak GPU memory is just filtered concat. No 2 buffers exist at the same time.


extend() | merge() (this PR)
-- | --
Caller fills extended_dataset = old \|\| new | Caller fills merged_dataset = index0 \|\| index1 \|\| … (already filtered)
new_start_row marks the split | offsets mark each index’s split
Library does not copy dataset rows | Library does not copy dataset rows
Graph work + update_dataset to bind the view | Graph work + update_dataset to bind the view
Caller owns lifetime | Caller owns lifetime

Hope that helps!

HowardHuang1:
/ok to test [ab5fe77](https://github.com/NVIDIA/cuvs/pull/2507/commits/ab5fe7769aff1c46888343dbaead129f008f67ff)

HowardHuang1:
/ok to test [a8e86d7](https://github.com/NVIDIA/cuvs/pull/2507/commits/a8e86d7a8b5a103fa35760118fce0302a3ea67e4)

c/include/cuvs/neighbors/cagra.h
```diff
@@ -1105,6 +1141,7 @@ CUVS_EXPORT cuvsError_t cuvsCagraMergeWithParams(cuvsResources_t res,
                                                  size_t num_indices,
                                                  cuvsFilter filter,
                                                  cuvsDataset_t merged_dataset,
+                                                 const int64_t* offsets,
```
cjnolet, line 1149: This breaks ABI. We can't change C functions (add parameters, etc..).

PR #2511, opened 2026-08-26
Delete stale Panama bindings instead of generating on top of them

imotov:
`jextract` overwrites the files it generates but never removes them, so a binding for a symbol that was renamed or dropped from the C headers survives in `java/cuvs-java/src/main/java22/com/nvidia/cuvs/internal/panama/` and keeps compiling into the jar. The stale class then disagrees with `libcuvs_c.so`, producing confusing errors that go away only after deleting the directory by hand.

Makes `java/panama-bindings/generate-bindings.sh` remove the target package directory before invoking `jextract`. Also adds the two directories `java/build.sh` produces — `java/internal/build` and the generated `panama` package — to `BUILD_DIRS`, so `./build.sh clean` leaves the tree pristine without requiring a subsequent build.


jameslamb (approved):
Yay `bash` arrays instead of space-delimited strings! Approved.

cjnolet (approved):

PR #2515, opened 2026-08-27
Lucene: preserve throwable types in accelerated vector formats

### LINKED ISSUE #2474: Lucene: Propagate `Error` subclasses unchanged instead of indiscriminately wrapping in `RuntimeException`
Various locations in the code base currently catch all exceptions and throw `RuntimeException`, e.g. https://github.com/NVIDIA/cuvs-lucene/blob/4db50a1903ca28e3783b2c2d2c1e58290a974b90/src/main/java/com/nvidia/cuvs/lucene/Lucene99AcceleratedHNSWVectorsFormat.java#L93-L94

@imotov [weighed in on this issue](https://github.com/NVIDIA/cuvs-lucene/pull/133#discussion_r3591284068):

> I think, for now, this can be replaced with [Utils.handleThrowable](https://github.com/NVIDIA/cuvs-lucene/blob/main/src/main/java/com/nvidia/cuvs/lucene/Utils.java#L36). More generally, though, I don't think this is a great pattern. We shouldn't indiscriminately wrap everything in a RuntimeException, as that prevents applications using this library from handling serious conditions such as OutOfMemoryError or AssertionError appropriately. In general, it's better to propagate Error subclasses unchanged and only wrap exceptions that are actually intended to be converted into unchecked exceptions.

shaunakkapur:
## Summary

Replace the broad RuntimeException wrappers in fieldsWriter and fieldsReader across the plain, scalar quantized, and binary quantized accelerated vector formats with Utils.handleThrowable.

This preserves IOException, RuntimeException, and Error instances instead of discarding their type, cause, and stack information. It also documents and tests the shared Error propagation contract.

Add focused helper and vector format regression coverage. The call site tests verify that IOException and RuntimeException instances are propagated unchanged, while the helper test verifies exact Error identity.

## Testing

* TestUtilsThrowableHandling: 4 tests passed
* TestAcceleratedHNSWVectorsFormatThrowableHandling: 3 tests passed
* Spotless check passed
* git diff check passed

The call site tests were also run against the previous wrappers and correctly failed for IOException and RuntimeException identity.

Closes #2474


java/cuvs-lucene/src/main/java/com/nvidia/cuvs/lucene/Lucene99AcceleratedHNSWVectorsFormat.java
```diff
@@ -91,7 +91,8 @@ public KnnVectorsWriter fieldsWriter(SegmentWriteState state) throws IOException
             acceleratedHNSWParams.getNumMergeWorkers(),
             new TaskExecutor(acceleratedHNSWParams.getMergeExec()));
       } catch (Exception e) {
-        throw new RuntimeException(e.getMessage());
+        Utils.handleThrowable(e);
+        throw new AssertionError("handleThrowable always throws"); // unreachable
```
imotov: I don't think we need these assertions. You already have assertThrows in the test, and it ensures that we don't swallow an exception in these methods.

imotov (commented):
Looks good to me. Left one comment.

java/cuvs-lucene/src/main/java/com/nvidia/cuvs/lucene/Lucene99AcceleratedHNSWVectorsFormat.java
shaunakkapur: Fixed, I adjusted Utils.handleThrowable() to return a RuntimeException instead of void, so callers can just write throw Utils.handleThrowable(e) and the extra unreachable AssertionError line is no longer needed.

imotov (approved):

PR #2516, opened 2026-08-27
Lucene: define and validate CAGRA parameter boundaries

### LINKED ISSUE #2450: Lucene: Define and implement parameter boundaries
We expose the following parameters via our `cuvs-lucene` API.

Build Index
- CAGRA:
   - cuVS writer threads
   - CAGRA Intermediate graph degree
   - CAGRA Graph degree

Searching
- CAGRA ItopK
- CAGRA Search width

We need to have a valid boundary defined for each of the above parameters, and at least have basic checks in place at the `cuvs-lucene` level, so that only valid parameters are eventually passed to the lower-level cuVS API via the `cuvs-Java` layer.

Ideally, I want this defined and checks in place before we let customers use the API.

shaunakkapur:
## Summary

Define and document boundaries for all five Lucene CAGRA parameter families: writer threads, intermediate graph degree, graph degree, iTopK, and search width.

Validate individual build parameter ranges when public builder setters are called. Validate at build time that graphDegree does not exceed intermediateGraphDegree, independent of setter order.

Validate search parameters when GPUKnnFloatVectorQuery is constructed, including the native SINGLE_CTA iTopK limit and the numeric safety limit for search width.

Add focused minimum, maximum, invalid range, cross field relationship, and search algorithm coverage.

## Testing

* TestAcceleratedHNSWParams
* TestGPUSearchParams
* TestGPUKnnFloatVectorQueryParameters
* TestSegmentMaxConnConsistency.testSingleVectorSegmentDoesNotUseConfiguredGraphDegree
* 29 tests passed with zero failures, errors, or skips
* Spotless check passed
* git diff check passed

Closes #2450

### DIFF
--- a/java/cuvs-lucene/src/main/java/com/nvidia/cuvs/lucene/AcceleratedHNSWParams.java
+++ b/java/cuvs-lucene/src/main/java/com/nvidia/cuvs/lucene/AcceleratedHNSWParams.java
@@ -31,11 +31,9 @@ public class AcceleratedHNSWParams {
     CUSTOM
   }
 
-  /*
-   * TODO: Update boundaries for all parameters when a consensus is reached.
-   * Issue: https://github.com/rapidsai/cuvs-lucene/issues/99
-   */
+  /** Bounds for the public CAGRA and HNSW build parameters. */
   public static final int MIN_WRITER_THREADS = 1;
+
   public static final int MAX_WRITER_THREADS = 512;
   public static final int MIN_INT_GRAPH_DEG = 2;
   public static final int MAX_INT_GRAPH_DEG = 512;
--- a/java/cuvs-lucene/src/main/java/com/nvidia/cuvs/lucene/AcceleratedHNSWParams.java
+++ b/java/cuvs-lucene/src/main/java/com/nvidia/cuvs/lucene/AcceleratedHNSWParams.java
@@ -560,7 +560,7 @@ public class AcceleratedHNSWParams {
                 + MAX_GRAPH_DEG
                 + "]");
       }
-      if (graphdegree > intermediateGraphDegree) {
+      if (strategy == Strategy.CUSTOM && graphdegree > intermediateGraphDegree) {
         throw new IllegalArgumentException(
             "graphDegree must not be greater than intermediateGraphDegree.");
       }
--- a/java/cuvs-lucene/src/main/java/com/nvidia/cuvs/lucene/AcceleratedHNSWParams.java
+++ b/java/cuvs-lucene/src/main/java/com/nvidia/cuvs/lucene/AcceleratedHNSWParams.java
@@ -340,6 +340,7 @@ public class AcceleratedHNSWParams {
      * @return instance of {@link Builder}
      */
     public Builder withWriterThreads(int writerThreads) {
+      validateRange("writerThreads", writerThreads, MIN_WRITER_THREADS, MAX_WRITER_THREADS);
       this.writerThreads = writerThreads;
       return this;
     }
@@ -354,6 +355,8 @@ public class AcceleratedHNSWParams {
      * @return instance of {@link Builder}
      */
     public Builder withIntermediateGraphDegree(int intermediateGraphDegree) {
+      validateRange(
+          "intermediateGraphDegree", intermediateGraphDegree, MIN_INT_GRAPH_DEG, MAX_INT_GRAPH_DEG);
       this.intermediateGraphDegree = intermediateGraphDegree;
       return this;
     }
@@ -368,6 +371,7 @@ public class AcceleratedHNSWParams {
      * @return instance of {@link Builder}
      */
     public Builder withGraphDegree(int graphDegree) {
+      validateRange("graphdegree", graphDegree, MIN_GRAPH_DEG, MAX_GRAPH_DEG);
       this.graphdegree = graphDegree;
       return this;
     }
@@ -516,6 +520,13 @@ public class AcceleratedHNSWParams {
     public Builder withHnswHeuristicType(HnswHeuristicType hnswHeuristicType) {
       this.hnswHeuristicType = hnswHeuristicType;
       return this;
+    }
+
+    private static void validateRange(String name, int value, int min, int max) {
+      if (value < min || value > max) {
+        throw new IllegalArgumentException(
+            name + " not in valid range. Valid range: [" + min + ", " + max + "]");
+      }
     }
 
     /**
--- a/java/cuvs-lucene/src/main/java/com/nvidia/cuvs/lucene/AcceleratedHNSWParams.java
+++ b/java/cuvs-lucene/src/main/java/com/nvidia/cuvs/lucene/AcceleratedHNSWParams.java
@@ -371,7 +371,7 @@ public class AcceleratedHNSWParams {
      * @return instance of {@link Builder}
      */
     public Builder withGraphDegree(int graphDegree) {
-      validateRange("graphdegree", graphDegree, MIN_GRAPH_DEG, MAX_GRAPH_DEG);
+      validateRange("graphDegree", graphDegree, MIN_GRAPH_DEG, MAX_GRAPH_DEG);
       this.graphdegree = graphDegree;
       return this;
     }
@@ -559,6 +559,10 @@ public class AcceleratedHNSWParams {
                 + ", "
                 + MAX_GRAPH_DEG
                 + "]");
+      }
+      if (graphdegree > intermediateGraphDegree) {
+        throw new IllegalArgumentException(
+            "graphDegree must not be greater than intermediateGraphDegree.");
       }
       if (hnswLayers < MIN_HNSW_LAYERS || hnswLayers > MAX_HNSW_LAYERS) {
         throw new IllegalArgumentException(
--- a/java/cuvs-lucene/src/main/java/com/nvidia/cuvs/lucene/CuVS2510GPUVectorsReader.java
+++ b/java/cuvs-lucene/src/main/java/com/nvidia/cuvs/lucene/CuVS2510GPUVectorsReader.java
@@ -501,9 +501,14 @@ public class CuVS2510GPUVectorsReader extends KnnVectorsReader {
         CagraSearchParams searchParams;
         if (knnCollector instanceof GPUPerLeafCuVSKnnCollector) {
           GPUPerLeafCuVSKnnCollector collector = (GPUPerLeafCuVSKnnCollector) knnCollector;
+          int effectiveITopK = Math.max(collector.getiTopK(), topK);
+          // topK may have been raised above the value validated at query construction time (see
+          // GPUKnnFloatVectorQuery.validateSearchParameters), e.g. by the filter-cardinality bump
+          // above. Re-validate against the final value actually sent to native CAGRA.
+          GPUKnnFloatVectorQuery.validateSingleCtaItopk(effectiveITopK, collector.getSearchAlgo());
           searchParams =
               new CagraSearchParams.Builder()
-                  .withItopkSize(Math.max(collector.getiTopK(), topK))
+                  .withItopkSize(effectiveITopK)
                   .withSearchWidth(collector.getSearchWidth())
                   .withThreadBlockSize(collector.getThreadBlockSize())
                   .withMaxIterations(collector.getMaxIterations())
--- a/java/cuvs-lucene/src/main/java/com/nvidia/cuvs/lucene/GPUKnnFloatVectorQuery.java
+++ b/java/cuvs-lucene/src/main/java/com/nvidia/cuvs/lucene/GPUKnnFloatVectorQuery.java
@@ -77,14 +77,19 @@ public class GPUKnnFloatVectorQuery extends KnnFloatVectorQuery {
   /**
    * Largest intermediate-result count representable by the public Java API.
    *
-   * <p>This is a representational limit only, not a guarantee that every value up to this bound
-   * is supported for every search. Native CAGRA sizes internal traversal hash tables from a
-   * combination of itopk_size, search_width, max_iterations, and (for MULTI_CTA, which a normal
-   * one-query {@code AUTO} search resolves to) the graph degree and dataset size — none of which
-   * are all known at query-construction time. Combinations that exceed native CAGRA's hash-table
-   * capacity are rejected by native CAGRA itself with a clear exception (see {@link
-   * Utils#handleThrowable}); this class does not attempt to replicate that dataset- and
-   * algorithm-dependent sizing logic.
+   * <p>This is a representational limit only. It is emphatically not a supported maximum: values
+   * anywhere near it are rejected by native CAGRA in practice. Native CAGRA sizes internal
+   * traversal hash tables from a combination of itopk_size, search_width, max_iterations, and
+   * (for MULTI_CTA, which a normal one-query {@code AUTO} search resolves to) the graph degree
+   * and dataset size, none of which are all known at query-construction time, so this class does
+   * not attempt to replicate that sizing logic.
+   *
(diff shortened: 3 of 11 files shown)


java/cuvs-lucene/src/test/java/com/nvidia/cuvs/lucene/TestSegmentMaxConnConsistency.java
```diff
@@ -131,6 +131,7 @@ public void testSingleVectorSegmentDoesNotUseConfiguredGraphDegree() throws Exce
         new AcceleratedHNSWParams.Builder()
             .withStrategy(AcceleratedHNSWParams.Strategy.HEURISTIC)
             .withMaxConn(MAX_CONN)
+            .withIntermediateGraphDegree(256) // ignored under HEURISTIC
```
dantegd: Doesn’t this added setter show that the new relationship check is being applied to values that heuristic mode ignores?

For AcceleratedHNSWParams, heuristic mode derives both degrees from maxConn and beamWidth. For GPUSearchParams, it derives the intermediate degree from the graph degree. The configured pair therefore isn’t the pair eventually passed to cuVS.

As a result, a previously valid configuration such as graphDegree=256 with the default intermediate degree now needs a second ignored setter just to build. What do you think about enforcing graphDegree <= intermediateGraphDegree only for CUSTOM, or checking the actual derived values instead?

java/cuvs-lucene/src/main/java/com/nvidia/cuvs/lucene/GPUKnnFloatVectorQuery.java
```diff
@@ -117,13 +135,37 @@ public GPUKnnFloatVectorQuery(
...
     this.threadBlockSize = threadBlockSize;
     this.maxIterations = maxIterations;
     this.searchAlgo = searchAlgo;
   }
 
+  private static void validateSearchParameters(
+      int iTopK, int searchWidth, int k, CagraSearchParams.SearchAlgo searchAlgo) {
+    validateRange("iTopK", iTopK, MIN_ITOPK, MAX_ITOPK);
+    validateRange("searchWidth", searchWidth, MIN_SEARCH_WIDTH, MAX_SEARCH_WIDTH);
+    int effectiveITopK = Math.max(iTopK, k);
```
dantegd: I think we may be validating effectiveITopK too early here. The filtered per-segment path later increases topK to min(k + 10, filterCardinality) and uses that value when creating the native search parameters.

For example, k=503 and iTopK=512 pass this check, but a filter accepting at least 513 vectors sends 513 downstream. Native CAGRA aligns that to 544 and rejects it for SINGLE_CTA.

What do you think about centralizing this check and running it after the fallback adjusts topK, or rerouting that case to another algorithm?

java/cuvs-lucene/src/main/java/com/nvidia/cuvs/lucene/GPUKnnFloatVectorQuery.java
```diff
@@ -71,6 +71,24 @@
...
+  public static final int MAX_SINGLE_CTA_ITOPK = 512;
+
+  /** Smallest supported number of CAGRA search entry points. */
+  public static final int MIN_SEARCH_WIDTH = 1;
+
+  /**
+   * Largest search width that keeps CAGRA's result buffer within its unsigned 32-bit indexing
+   * limit at the maximum graph degree and aligned {@link #MAX_ITOPK}.
+   */
+  public static final int MAX_SEARCH_WIDTH = 4_194_303;
```
dantegd, line 112: I wonder if we should reconsider how these upper bounds are derived? MAX_SEARCH_WIDTH appears to protect the unsigned result-buffer calculation, but that isn’t the limiting calculation for the MULTI_CTA path that a normal one-query AUTO search uses.

MULTI_CTA sizes its hash from max(searchWidth, ceil(iTopK / 32)) * max(32, maxIterations). With MAX_SEARCH_WIDTH, that requires at least a 28-bit hash table, while native CAGRA hard-limits it to 25. MAX_ITOPK is even larger and would require a 32-bit table, beyond the range of the native 1U << bitlen sizing.

Since the current test only constructs the Java query, it doesn’t exercise these native constraints. What do you think about validating after the algorithm is resolved and adding a test that actually creates the native search plan?

dantegd (changes requested):
Great to see this PR! Just had a few comments

java/cuvs-lucene/src/test/java/com/nvidia/cuvs/lucene/TestSegmentMaxConnConsistency.java
shaunakkapur: Good point! The relationship is now only enforced under CUSTOM, since HEURISTIC derives the degrees itself. I removed the workaround setter from TestSegmentMaxConnConsistency and added tests covering both the CUSTOM rejection and the HEURISTIC no-op case.

java/cuvs-lucene/src/main/java/com/nvidia/cuvs/lucene/GPUKnnFloatVectorQuery.java
shaunakkapur: I think you are right, this was checked prematurely. I pulled the SINGLE_CTA check into a shared helper and called it again in CuVS2510GPUVectorsReader right before the native params are built, so it sees the post-filter value. Also added a GPU test reproducing your exact case (k=503, iTopK=512, filter cardinality ≥513) that asserts the error names SINGLE_CTA and the 512/513 values.

java/cuvs-lucene/src/main/java/com/nvidia/cuvs/lucene/GPUKnnFloatVectorQuery.java
shaunakkapur: Agreed, these were presented as more universal than they are. The oversized searchWidth test now goes through a normal one-query AUTO search and asserts the native failure names the hash_bitlen 25-bit limit, and the README/javadoc now say only the minimum of 1 and the SINGLE_CTA max of 512 are real native limits.

One open item: I couldn't add equivalent coverage for MAX_ITOPK. With iTopK = Integer.MAX_VALUE the native call doesn't return an error, it leaks and stays stuck past the test's own timeout, so asserting on it isn't CI-safe. I documented that in the test. Since the real limit depends on the resolved algorithm and other settings, would you rather we enforce an algorithm-specific limit in the Lucene API, or fix the native calculation to reject it safely first?

dantegd (approved):

java/cuvs-lucene/src/main/java/com/nvidia/cuvs/lucene/GPUKnnFloatVectorQuery.java
```diff
@@ -127,13 +167,56 @@ public GPUKnnFloatVectorQuery(
...
     this.searchWidth = searchWidth;
     this.threadBlockSize = threadBlockSize;
     this.maxIterations = maxIterations;
     this.searchAlgo = searchAlgo;
   }
 
+  private static void validateSearchParameters(
+      int iTopK, int searchWidth, int k, CagraSearchParams.SearchAlgo searchAlgo) {
+    validateRange("iTopK", iTopK, MIN_ITOPK, MAX_ITOPK);
+    validateRange("searchWidth", searchWidth, MIN_SEARCH_WIDTH, MAX_SEARCH_WIDTH);
```
imotov: Why not validate `maxIterations` and `threadBlockSize` here? I think negative maxIterations might be pretty bad.

java/cuvs-lucene/src/main/java/com/nvidia/cuvs/lucene/CuVS2510GPUVectorsReader.java
```diff
@@ -508,9 +508,14 @@ public void search(String field, float[] target, KnnCollector knnCollector, Bits
         CagraSearchParams searchParams;
         if (knnCollector instanceof GPUPerLeafCuVSKnnCollector) {
           GPUPerLeafCuVSKnnCollector collector = (GPUPerLeafCuVSKnnCollector) knnCollector;
+          int effectiveITopK = Math.max(collector.getiTopK(), topK);
+          // topK may have been raised above the value validated at query construction time (see
+          // GPUKnnFloatVectorQuery.validateSearchParameters), e.g. by the filter-cardinality bump
+          // above. Re-validate against the final value actually sent to native CAGRA.
+          GPUKnnFloatVectorQuery.validateSingleCtaItopk(effectiveITopK, collector.getSearchAlgo());
```
imotov: I don't think we should throw here. This is an internal detail and will confuse users, since it might or might not happen depending on how the index is built. You could build the index and it works; you index a few more records and it fails; you force-merge and it works again. In other words, that might lead to a very frustrating situation where it fails sporadically and only in production. This cap is more of a heuristic than a requirement, so I would suggest keeping the cap in place and getting rid of the exception.

java/cuvs-lucene/src/main/java/com/nvidia/cuvs/lucene/AcceleratedHNSWParams.java
```diff
@@ -538,6 +547,10 @@ private void validate() throws IllegalArgumentException {
                 + MAX_GRAPH_DEG
                 + "]");
       }
+      if (strategy == Strategy.CUSTOM && graphdegree > intermediateGraphDegree) {
```
imotov: Not sure about this one. We kind of [accept it ignore](https://github.com/NVIDIA/cuvs/blob/main/cpp/src/neighbors/detail/cagra/cagra_build.cuh#L173) in other layers so this might break things that actually work today. @dantegd what do you think?

java/cuvs-lucene/src/main/java/com/nvidia/cuvs/lucene/GPUKnnFloatVectorQuery.java
```diff
@@ -127,13 +167,56 @@ public GPUKnnFloatVectorQuery(
...
+      throw new IllegalArgumentException(
+          "effective iTopK must not exceed "
+              + MAX_SINGLE_CTA_ITOPK
+              + " for SINGLE_CTA search, but was "
+              + effectiveITopK
+              + ".");
+    }
+  }
+
+  private static void validateRange(String name, int value, int min, int max) {
```
imotov: Could you move this into some utility class?

java/cuvs-lucene/src/main/java/com/nvidia/cuvs/lucene/GPUSearchParams.java
imotov: It looks like you copied some of these validations into setters while leaving another copy here. Validating in setters is good since we fail early, but spreading the same logic across multiple places is not as good, since it's more difficult to figure out what is covered and what is not. Usage-wise, I think all these `withSomething()` calls followed by `build()` will be on the same line anyway, so failing earlier doesn't buy us much in this case while giving range checkers a good name and moving all these validations into `validate()` might be a cleaner way of handling this.

java/cuvs-lucene/src/main/java/com/nvidia/cuvs/lucene/AcceleratedHNSWParams.java
```diff
@@ -31,11 +31,9 @@ public static enum Strategy {
     CUSTOM
   }
 
-  /*
-   * TODO: Update boundaries for all parameters when a consensus is reached.
-   * Issue: https://github.com/rapidsai/cuvs-lucene/issues/99
-   */
+  /** Bounds for the public CAGRA and HNSW build parameters. */
```
imotov: This now a javadoc comment for MIN_WRITER_THREADS. Might be confusing.

java/cuvs-lucene/src/main/java/com/nvidia/cuvs/lucene/AcceleratedHNSWParams.java
```diff
@@ -31,11 +31,9 @@ public static enum Strategy {
...
   public static final int MIN_WRITER_THREADS = 1;
+
```
imotov: Stray new line?

imotov (commented):
Left a couple of comments.

java/cuvs-lucene/src/main/java/com/nvidia/cuvs/lucene/GPUKnnFloatVectorQuery.java
shaunakkapur: Oh good catch, I now validate both when the query is constructed.

java/cuvs-lucene/src/main/java/com/nvidia/cuvs/lucene/CuVS2510GPUVectorsReader.java
shaunakkapur: That makes sense, I removed the exception and now just cap the over-fetch at 512.

java/cuvs-lucene/src/main/java/com/nvidia/cuvs/lucene/AcceleratedHNSWParams.java
shaunakkapur: I justdropped the check so native clamping works as it does today (@dantegd  let me know if you'd prefer otherwise).

java/cuvs-lucene/src/main/java/com/nvidia/cuvs/lucene/GPUKnnFloatVectorQuery.java
shaunakkapur: Got it, I moved it into a shared ParameterValidation class.

java/cuvs-lucene/src/main/java/com/nvidia/cuvs/lucene/AcceleratedHNSWParams.java
shaunakkapur: Fixed.

java/cuvs-lucene/src/main/java/com/nvidia/cuvs/lucene/AcceleratedHNSWParams.java
shaunakkapur: Fixed this.

java/cuvs-lucene/src/main/java/com/nvidia/cuvs/lucene/GPUSearchParams.java
shaunakkapur: That makes sense, I removed the setter checks so all validation now lives in validate().

java/cuvs-lucene/src/main/java/com/nvidia/cuvs/lucene/GPUKnnFloatVectorQuery.java
```diff
@@ -127,13 +168,53 @@ public GPUKnnFloatVectorQuery(
...
+    validateSingleCtaItopk(Math.max(iTopK, k), searchAlgo);
+  }
+
+  /**
+   * Validates the caller's effective iTopK against the SINGLE_CTA algorithm's limit.
+   *
+   * @param effectiveITopK the itopk_size value about to be sent to native CAGRA
+   * @param searchAlgo the CAGRA search algorithm the query will run under
+   */
+  static void validateSingleCtaItopk(int effectiveITopK, CagraSearchParams.SearchAlgo searchAlgo) {
```
imotov: This can now be private right?

imotov (approved):
Left a minor comment. Otherwise LGTM.

java/cuvs-lucene/src/main/java/com/nvidia/cuvs/lucene/GPUKnnFloatVectorQuery.java
shaunakkapur: Yep, it’s only used within the query class now. Made it private.

PR #2519, opened 2026-08-28
Dry run: preprocessing::quantize::scalar

achirkin:
Tweak scalar quantizer to support execution in dry run mode.

### DIFF
--- a/cpp/tests/test_utils.cuh
+++ b/cpp/tests/test_utils.cuh
@@ -1,5 +1,5 @@
 /*
- * SPDX-FileCopyrightText: Copyright (c) 2018-2026, NVIDIA CORPORATION.
+ * SPDX-FileCopyrightText: Copyright (c) 2018-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
  * SPDX-License-Identifier: Apache-2.0
  */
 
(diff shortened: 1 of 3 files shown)


mfoerste4 (approved):
LGTM!

cpp/tests/test_utils.cuh
```diff
@@ -323,4 +327,96 @@ inline std::vector<float> read_csv(std::string filename, bool skip_first_n_colum
...
+ * allocations, then runs it for real with all six memory resources wrapped in
+ * statistics adaptors to record actual peak usage. Compares the predicted and
+ * actual peaks according to the specified @p behavior.
+ *
+ * @tparam Action callable with signature void(raft::resources const&)
+ */
+template <typename Action>
+void execute_with_dry_run_check(raft::resources const& res,
+                                Action&& action,
+                                alloc_behavior behavior,
```
tfeher, line 349: What is the expected usage of this parameter?

tfeher (changes requested):
Hi Artem, thanks for the PR, it looks good, but I have one question regarding the usage of the test util.

cpp/tests/test_utils.cuh
achirkin: This harness is a verbatim copy from raft test utils. You can specify there where:
a. NO_ALLOCATIONS: you don't expect any allocations. Both dry run and real run should yield all zeroes.
b. ARGUMENT_DRIVEN: you expect the exact size of allocations is fully determined by the input arguments. Dry run and real run should return exact same sizes.
c. DATA_DRIVEN: the allocation size depends on data. Dry run sizes must never be less than real run sizes.

tfeher: Ok. Would be great to add this as a docstring to the `alloc_behaviour` (both in raft and here) in a follow up.

tfeher (approved):

PR #2526, opened 2026-08-29
[ANN_BENCH] Remove Stream Pool Creation from Individual Wrappers

tarang-jain:
- remove stream pool creation from individual wrappers


cpp/bench/ann/src/cuvs/cuvs_cagra_wrapper.h
```diff
@@ -679,7 +679,7 @@ void cuvs_cagra<T, IdxT>::search_base(
         raft::resources composite_handle(handle_);
         size_t n_streams = cagra_indices.size();
         raft::resource::set_cuda_stream_pool(composite_handle,
-                                             std::make_shared<rmm::cuda_stream_pool>(n_streams));
+                                             make_non_blocking_stream_pool(n_streams));
```
achirkin: I'm a bit surprised how this stream pool setting has appeared in all wrapper code in the first place.
Please remove all these repeated set_cuda_stream_pools from the wrapper files and put it exactly once either in the shared_raft_resources or in configured_raft_resources object (where more appropriate).

cpp/bench/ann/src/cuvs/cuvs_ann_bench_utils.h
```diff
@@ -31,6 +32,13 @@
 
 namespace cuvs::bench {
 
+/** Create streams that do not implicitly synchronize with the default stream. */
+inline auto make_non_blocking_stream_pool(size_t n_streams)
+  -> std::shared_ptr<rmm::cuda_stream_pool>
+{
+  return std::make_shared<rmm::cuda_stream_pool>(n_streams, rmm::cuda_stream::flags::non_blocking);
+}
```
achirkin: Once we reduce the stream pool setting to just one place, I think we won't need this helper function anymore.

achirkin (changes requested):
Thanks, I think the change makes perfect sense and can only improve the performance. We don't need any extra stream syncs, because the replaced managed memory was used as a large workspace resource (which is async by contract).
Just please update the description with the explanation of the problem you wanted to solve and do a small cleanup (see below).

cpp/bench/ann/src/cuvs/cuvs_cagra_wrapper.h
tarang-jain: I got rid of this behavior in the other ivf wrappers. But for cagra it seems this particular pool is a special case, for the composite index.

cpp/bench/ann/src/cuvs/cuvs_ann_bench_utils.h
```diff
@@ -132,6 +134,8 @@ class configured_raft_resources {
   /** Default constructor creates all resources anew. */
   configured_raft_resources() : configured_raft_resources{std::make_shared<shared_raft_resources>()}
   {
+    raft::resource::set_cuda_stream_pool(
+      *res_, std::make_shared<rmm::cuda_stream_pool>(1, rmm::cuda_stream::flags::non_blocking));
```
tarang-jain: Add a prefetch stream in the default constructor.

cpp/bench/ann/src/common/util.hpp
```diff
@@ -1,12 +1,16 @@
...
  */
 #pragma once
 
 #include "ann_types.hpp"
 #include "cuda_stub.hpp"  // cuda-related utils
 
+#ifndef BUILD_CPU_ONLY
+#include <rmm/cuda_stream.hpp>
+#endif
+
```
achirkin, line 13: Please check if all relevant components compile with all combinations of CUVS_ANN_BENCH_SINGLE_EXE and BUILD_CPU_ONLY flags (it's not checked in CI atm).

achirkin (approved):
Thanks for the updates, LGTM!
But why did you undo the async managed resource change? I would expect it to slightly improve performance during build.

cpp/bench/ann/src/common/util.hpp
tarang-jain: There was just one problem -- looking for and linking raft::raft in CPU_ONLY mode which i fixed below with the TARGET_NAME_IF_EXISTS guard.

bdice (approved):

PR #2527, opened 2026-08-29
Use JIT-LTO kernel handles through an API the toolkit actually provides

drzraf:
Use JIT-LTO kernel handles through an API the toolkit actually provides

Part 1 of 2. On its own **this commit changes nothing observable: it is a prerequisite for building cuVS with a CUDA toolkit older than 12.8**, and that becomes possible only in combination with https://github.com/rapidsai/librtcx/pull/17 - "Support CUDA toolkits older than 12.8 via the driver API"

Neither is useful without the other, and either may merge first. The split follows the code: since https://github.com/NVIDIA/cuvs/pull/2311 the JIT-LTO plumbing lives in librtcx, which loads the linked image (`cudaLibraryLoadData` / `cudaLibraryGetKernel` / `cudaLibraryUnload`) and launches it (`cudaLaunchKernelExC`); cuVS keeps only what it does with the resulting handle. Both halves hit the same wall -- the runtime library-management API arrived in CUDA 12.8 -- but they are in different repositories, so they are fixed separately. Without librtcx#17 every cuVS translation unit that includes the rtcx launcher still fails with

> rtcx/algorithm_launcher.hpp:24: error: identifier "cudaLibrary_t" is undefined

(this commit does not change that)

The cuVS half. cuVS queries and sets attributes on the `cudaKernel_t` handles that rtcx hands out for run-time-linked kernels. Both things it does with those handles are CUDA 12.8 features:

* Passing a `cudaKernel_t` where a `"const void* func"` entry point is expected.
  The runtime documents this from 12.8 onwards -- "If the specified function
  does not exist, then it is assumed to be a `cudaKernel_t` and used as is",
  which appears in the 12.8 documentation of `cudaFuncGetAttributes`,
  `cudaFuncSetAttribute` and `cudaOccupancyMaxActiveBlocksPerMultiprocessor` and in
  no earlier version. On an older toolkit the handle is not a valid entry
  address and the call fails with `cudaErrorInvalidDeviceFunction`.

* `cudaKernelSetAttribut`eForDevice, which does not exist before 12.8 in either the headers or libcudart.

The driver API has always taken the equivalent `CUkernel` / `CUfunction`, and `cudaKernel_t` is a `typedef` for `struct CUkern_st*`, i.e. exactly `CUkernel`, so the handles carry over unchanged. Add `src/util/jit_kernel_compat.hpp`, which forwards to the runtime API on 12.8+ and to `cuKernelSetAttribute`, `cuFuncGetAttribute` and `cuOccupancyMaxActiveBlocksPerMultiprocessor` otherwise, and route the six call sites through it. On **12.8+ every wrapper compiles down to the original runtime call** and `CUDA::cuda_driver` is not linked, so nothing changes for the toolkits cuVS supports today.

Two details worth recording:

* The runtime's primary context is forced to exist before any driver call,
  because `cuKernelGetFunction` resolves against the current context. Skipping
  this yields `CUDA_ERROR_INVALID_CONTEXT` from an otherwise correct sequence.

* Driver statuses are translated to the nearest runtime status so that existing
  `RAFT_CUDA_TRY` diagnostics stay meaningful.

In `launchConfigGenerator` the choice between the two occupancy APIs is an `if constexpr` on the argument type, because that helper is also called with genuine `__global__` function pointers, for which the runtime API is correct.

The IVF-PQ kernel selector needed one further change. It asserted that a failed `cudaFuncSetAttribute` was also observable through `cudaGetLastError()`, which only holds for the runtime API; the driver path never sets the runtime's sticky error. The intent -- skip a kernel candidate that cannot get the shared memory it wants -- is preserved, and the sticky error is cleared explicitly.

Verified with CUDA 12.6.85: with https://github.com/rapidsai/librtcx/pull/17 applied (`-DCPM_rtcx_SOURCE=...`),
`libcuvs.so` builds and links for **sm_50**, and CAGRA build and search run correctly on a Maxwell device. Without it, the same tree fails only inside the rtcx headers, which is what _"part 1 of 2"_ means in practice.

Note that dependencies.yaml still lists CUDA 12.2 and 12.5 in its `cuda_version` matrix, so the toolkit range the project declares and the one it can actually build have disagreed since https://github.com/NVIDIA/cuvs/pull/1405.

(Opus 5)

### DIFF
--- a/cpp/CMakeLists.txt
+++ b/cpp/CMakeLists.txt
@@ -1532,6 +1532,19 @@ if(NOT BUILD_CPU_ONLY)
                                  ${CUVS_CUSPARSE_DEPENDENCY} ${CUVS_CURAND_DEPENDENCY}
   )
 
+  # Before CUDA 12.8 the runtime rejects a cudaKernel_t where a kernel entry point is expected, and
+  # cudaKernelSetAttributeForDevice does not exist at all, so the JIT-LTO kernel handles are queried
+  # and configured through the equivalent driver-API entry points instead (see
+  # src/util/jit_kernel_compat.hpp). That needs the driver stub at link time.
+  set(CUVS_CUDA_DRIVER_DEPENDENCY "")
+  if(CUDAToolkit_VERSION VERSION_LESS 12.8)
+    set(CUVS_CUDA_DRIVER_DEPENDENCY CUDA::cuda_driver)
+    message(
+      STATUS
+        "cuVS: CUDA ${CUDAToolkit_VERSION} does not accept kernel handles in the runtime API; using the driver API for JIT-LTO kernel attributes"
+    )
+  endif()
+
   if(NOT cuvs_compile_mode STREQUAL "static_only")
     add_library(cuvs SHARED $<TARGET_OBJECTS:cuvs_objs>)
     add_library(cuvs::cuvs ALIAS cuvs)
@@ -1571,6 +1584,7 @@ if(NOT BUILD_CPU_ONLY)
              $<$<BOOL:${CUVS_NVTX}>:CUDA::nvtx3>
       PRIVATE $<TARGET_NAME_IF_EXISTS:OpenMP::OpenMP_CXX> $<COMPILE_ONLY:nvidia::cutlass::cutlass>
               $<COMPILE_ONLY:cuco::cuco> CUDA::nvJitLink CUDA::nvrtc rtcx::rtcx
+              ${CUVS_CUDA_DRIVER_DEPENDENCY}
     )
 
     # ensure CUDA symbols aren't relocated to the middle of the debug build binaries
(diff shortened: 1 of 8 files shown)


bdice:
> Note that dependencies.yaml still lists CUDA 12.2 and 12.5 in its cuda_version matrix, so the toolkit range the project declares and the one it can actually build have disagreed since https://github.com/NVIDIA/cuvs/pull/1405.

We require CUDA 12.9+ to build this library. We can support earlier CUDA versions at runtime, back to 12.2, but this project does not aim to support earlier compilers. The project also requires Volta or newer architectures at runtime.

drzraf:
I commented about planned-obsolescence in https://github.com/NVIDIA/cuvs/pull/2529#issuecomment-5466807374

Given that pre-Volta architecture support is:
1. Possible
2. Provide actual performance benefit to their owners
3. while incurring very limited disruption to the codebase


Would you please mind elaborating the who & why of this CUDA/Volta+ architecture restriction, so that it could appear justified rather than arbitrary?


Technical arguments would also be welcomed.

Statement like _The project also requires ..._ describe the current situation that these PRs precisely intend to change, as such these can't rationally be considered arguments.


Thank you

PR #2529, opened 2026-08-29
Support pre-Volta (Maxwell/Pascal) GPU architectures

drzraf:
**Support pre-Volta (Maxwell/Pascal) GPU architectures**

Context: https://github.com/rapidsai/librtcx/pull/17
Includes/applies on top of https://github.com/NVIDIA/cuvs/pull/2527
To consider: https://github.com/NVIDIA/cuvs/pull/2528

cuVS could not be compiled for anything older than sm_70. Building with `--gpu-arch="50;61"` failed in `<cuda/semaphore>`, and several further blockers were hiding behind that first error.

Introduce a CMake-generated `cuvs/detail/arch_config.hpp` that exposes the architecture range of the build (`CUVS_MIN_CUDA_ARCH`, `CUVS_CUTLASS_ENABLED`, ...). Unlike `__CUDA_ARCH__` it is visible in the host pass too, so it can gate includes and template instantiations consistently; unlike a global -D it does not invalidate every object file when the architecture list changes.

Compile-time blockers addressed:

* `CUTLASS`. `cuda::binary_semaphore` hard-errors below sm_70 and the reduced-vec output iterator needs `cg::binary_partition`. The `CUTLASS` kernels are only ever dispatched to on sm_80+, so **below sm_70 the whole path is compiled out and the pre-existing SIMT/JIT fallbacks are used**. This also cuts build time.

* `__match_any_sync` (sm_70) in the sparse SpMV kernel. Replaced with a helper in `util/arch_compat.cuh` that uses the intrinsic where available and otherwise a ballot-based emulation whose loop trip count is warp-uniform, which is required on pre-Volta hardware.

* `atomicAdd_block` (sm_60). Falls back to the device-scoped atomic.

* `atomicAdd(double*, bool)` in bin_distance. Below sm_60 the native double overload does not exist, leaving only RAFT's `atomicAdd(T*, T)` template, whose deduction then fails. Fixed with an explicit cast.

* cuco's block-scoped `atomics (atom.cas.cta.*, sm_60)` in the sparse hash SpMV strategy, and the system-scoped atomics dynamic batching uses for its host/device handshake. Neither has a meaningful pre-Pascal emulation, so the device code is compiled out and both paths now fail with an explicit error instead of silently misbehaving. **The `dense_smem` SpMV strategy is unaffected**.

* `__half/__half2` arithmetic (sm_53) in the CAGRA VPQ distance, the faiss-select comparators and the l_inf distance op. Added `util/half_compat.cuh`; pre-Pascal targets widen to fp32, which is also a small accuracy improvement.

* `atomicAdd(double*, double)` and the `atomicAdd_block` family are only declared for `__CUDA_ARCH__ >= 600`, so in an older device pass they do not exist and every header that calls them unqualified fails to compile -- RAFT's `strided_reduction.cuh` (instantiated with double by cuVS) and CUB's `agent_histogram.cuh` among them. Added `util/atomic_compat.cuh`, which supplies `atomicAdd(double*)` as a 64-bit CAS loop comparing bit patterns rather than values, so a stored NaN cannot make it spin forever, plus `atomicAdd_block` overloads forwarding to the device-scoped atomics. It expands to nothing on sm_60+ and in the host pass. The calls live in templates but their argument types are fundamental, so ADL contributes nothing and only declarations visible at the definition context are considered: the header must be included before any RAFT/CUB header, which is recorded in a comment at each of the four translation units that need it.

* `raft::sqrt(__half)` lowers to hsqrt, an FP16 ALU instruction requiring sm_53, and its RAFT overload static_asserts below that. Added `cuvs::util::sqrt_op`, which widens __half to fp32 on such targets and **forwards to `raft::sqrt` otherwise**, and used it in the two element-wise passes over possibly-fp16 data.

Fixes for RAFT, CUB and cuco headers all live in cuVS: those are CPM-fetched, so patching them in place would be invisible to a reviewer, lost on a version bump and untestable in CI. cuVS instead makes the missing declarations visible before the dependency header is parsed.

Runtime blockers addressed:

* JIT-LTO fragments were pinned to a 70-real baseline regardless of `CMAKE_CUDA_ARCHITECTURES`. LTO-IR is forward but not backward compatible, so the baseline now tracks the oldest architecture being built for; otherwise nvJitLink cannot link CAGRA/IVF search kernels for an older device.

* NN-descent was disabled entirely below sm_70, so CAGRA had no NN-descent graph build. Only the WMMA kernel needs tensor cores; the SIMT kernel uses nothing newer than sm_50 and its static shared memory stays under the 48 KiB per-block limit for every supported dtype. Lower its floor to sm_50 and fall back to it on pre-Volta devices instead of throwing.

Opus 5

Benchmarked on SIFT with a Maxwell device (940MX) :

| recall@10 | GPU/CPU (thermally throttled), run 1 | GPU/CPU (cold), run 2 |
|---|---|---|
| 0.99 | 4.1× | 2.5× |
| 0.999 | 2.6× | 2.6× |

=> **2.5–4.1× at recall 0.99**


drzraf:
ping 😉

Issue #2395, opened 2026-08-03
[FEA][C++ API] Abstract common functions into dataset/dataset_view structs which currently do nothing + Reuse mdarray/mdspan to simplify Dataset API


HowardHuang1:
Address @Artem’s C++ changes here:
https://github.com/NVIDIA/cuvs/pull/1846#discussion_r3553946083 

2 tasks:
1. dataset and dataset_view struct contain no information. 

```
template <typename ContainerType, typename DataT, typename IdxT, typename Accessor>
struct dataset;

template <typename ContainerType, typename DataT, typename IdxT, typename Accessor>
struct dataset_view;
```
 
Only concrete dataset types like padded_dataset, standard_dataset, and vpq_dataset have information. However, that information is replicated for each of the concrete datasets even though the code logic is the same. This can be abstracted out and moved one level up to reduce code redundancy. In particular: view() function and n_rows() function are same regardless of concrete dataset type and can be abstracted one level up into the dataset struct. This makes it so that later we can just have operations on some basic common dataset struct features without specifying concrete dataset type and only dispatch on actual concrete dataset type later in the nested calls.

2. We have:

```
template <typename DataT, typename IdxT, typename Accessor>
using dense_owning_matrix = std::conditional_t<Accessor::is_device_accessible,
                                               raft::device_matrix<DataT, IdxT, raft::row_major>,
                                               raft::host_matrix<DataT, IdxT, raft::row_major>>;

template <typename DataT, typename IdxT, typename Accessor>
using dense_view_matrix =
  std::conditional_t<Accessor::is_device_accessible,
                     raft::device_matrix_view<const DataT, IdxT, raft::row_major>,
                     raft::host_matrix_view<const DataT, IdxT, raft::row_major>>;
```

Have dense dataset use implementation of mdarray because it's just the same thing. For compressed dataset this would be something different, perhaps several mdarrays for codebooks and any additional info. We have a rhombus of indirection where we are first converting from concrete dataset to dense dataset and then to underlying mdarray when we can just directly go from concrete dataset to mdarray and get rid of the middle layer. The dense_dataset layer of indirection is unecessary. Get rid of the conditional. 

This may involve removing host or device accessor to get rid of indirection in dense dataset. Just pass that host or device accessor directly to mdarray as shown here through a comment on the BBQ dataset PR: https://github.com/NVIDIA/cuvs/pull/2506#discussion_r3877889073

PR #2530, opened 2026-08-30
Reuse raft::mdarray/mdspan directly in dense dataset storage

HowardHuang1:
Addresses #2395. 

This PR reworks our original container based structure with container type tags and allows us to abstract more shared functionality one level up from the concrete dataset types and move them into the dataset and dataset_view structs. Shared functions like as_dataset_view(), n_rows(), dim() are all abstracted one level up and a new dictionary_view to handle vpq (and all future compressed dataset type) codebooks is introduced in this PR.

dense_owning_matrix/dense_view_matrix (and the VPQ codebook/code matrix aliases) picked between raft::device_matrix and raft::host_matrix via std::conditional_t, even though those are themselves just aliases for raft::mdarray/raft::mdspan with the exact accessor already computed as Accessor. Point the aliases at raft::mdarray/raft::mdspan directly instead.

dense_row_major_dataset_owning_storage/_view_storage also wrapped their matrix/view as a field and hand-forwarded view()/data_handle(), which raft::mdarray/raft::mdspan already provide natively. They now inherit from the matrix/view type instead, so those forwards are no longer needed.


aamijar (commented):
Hi Howard, from our discussion offline, let's think of how the naming should look like for our two dataset view methods.

Current PR:
```cpp
dataset.data_view() returns the raft_device_matrix_view type
dataset.as_dataset_view() returns the dataset_view type
```

Proposed:
```
dataset.as_matrix_view() returns raft_device_matrix_view type
dataset.view() returns dataset_view type
```

What do you think about the proposed naming? Open to suggestions.

aamijar (commented):
Hi Howard, I see in this PR you are introducing some structure called dictionary? Is that necessary? We should make things as simple as possible.

PR #2533, opened 2026-08-31
Bound CAGRA hash table sizing loops so they cannot spin forever

### LINKED ISSUE #2523: [BUG] CAGRA: calc_hashmap_params() never terminates for large itopk_size (hash_bitlen limit is checked after the sizing loop)
**Describe the bug**

`search_plan_impl::calc_hashmap_params()` sizes its hash tables with loops of the form

```cpp
hash_bitlen = min_bitlen;
while (max_traversed_nodes > hashmap::get_size(hash_bitlen) * max_fill_rate) {
  hash_bitlen += 1;
}
RAFT_EXPECTS(hash_bitlen <= 25, "hash_bitlen cannot be largen than 25 (32M)");
```

(`cpp/src/neighbors/detail/cagra/search_plan.cuh`, lines 285-288 for the MULTI_CTA traversed-node table.)

The bound check runs **after** the loop, and the loop has no upper bound of its own. `hashmap::get_size()` is

```cpp
// cpp/src/neighbors/detail/cagra/hashmap.hpp:23
RAFT_INLINE_FUNCTION uint32_t get_size(const uint32_t bitlen) { return 1U << bitlen; }
```

`1U << bitlen` is undefined behaviour once `bitlen >= 32` (shift >= width of `unsigned int`). In practice on x86 the shift count is masked to 5 bits, so `get_size(32)` wraps to `1`, `get_size(33)` to `2`, and so on. `hash_bitlen` is declared `int64_t`, so it keeps incrementing.

The result: whenever the requested table would need `bitlen >= 32`, the loop condition can never become false again and `calc_hashmap_params()` **spins forever on the host**. The `RAFT_EXPECTS` that was supposed to reject the request is never reached. This is a hang, not an allocation failure or a clean error.

*1. Root cause, standalone , no GPU or cuVS needed:*

```cpp
#include <cstdio>
#include <cstdint>
// verbatim from cpp/src/neighbors/detail/cagra/hashmap.hpp:23
uint32_t get_size(const uint32_t bitlen) { return 1U << bitlen; }

int main() {
  for (uint32_t b = 29; b <= 35; b++) printf("get_size(%2u) = %11u\n", b, get_size(b));

  double   max_fill_rate      = 0.5;
  uint64_t max_traversed_nodes = 1168750000ULL;   // itopk_size ~1.1e9 under MULTI_CTA
  int64_t  hash_bitlen        = 11;
  for (int guard = 0; guard < 80; guard++) {
    if (!(max_traversed_nodes > get_size(hash_bitlen) * max_fill_rate)) {
      printf("loop TERMINATED at hash_bitlen=%ld\n", (long)hash_bitlen);
      return 0;
    }
    hash_bitlen += 1;
  }
  printf("loop DID NOT TERMINATE (hash_bitlen=%ld and climbing)\n", (long)hash_bitlen);
}
```

```
$ g++ -O0 -o getsize_wrap getsize_wrap.cpp && ./getsize_wrap
get_size(29) =   536870912
get_size(30) =  1073741824
get_size(31) =  2147483648
get_size(32) =           1     <-- wraps
get_size(33) =           2
get_size(34) =           4
get_size(35) =           8
loop DID NOT TERMINATE (hash_bitlen=91 and climbing)
```

*2. End to end, through `cuvs-lucene`:*

A single-query `AUTO` search (which resolves to `MULTI_CTA`) over a 200-vector, 32-dim index, varying only `itopk_size`:

```java
// cuvs-lucene; single segment, 200 vectors x 32 dims. Only itopk_size varies.
Codec codec = TestUtil.alwaysKnnVectorsFormat(new CuVS2510GPUVectorsFormat());
float[][] dataset = generateDataset(random(), 200, 32);
try (Directory dir = newDirectory()) {
  try (IndexWriter w = new IndexWriter(dir, new IndexWriterConfig().setCodec(codec))) {
    for (float[] v : dataset) {
      Document d = new Document();
      d.add(new KnnFloatVectorField("vector", v, EUCLIDEAN));
      w.addDocument(d);
    }
  }
  try (DirectoryReader reader = DirectoryReader.open(dir)) {
    IndexSearcher searcher = new IndexSearcher(reader);
    int k = 5;
    int iTopK = 1_100_000_000;   // 1_000_000_000 errors cleanly; 1_100_000_000 hangs
    searcher.search(
        new GPUKnnFloatVectorQuery(
            "vector", dataset[0], k, /*filter=*/ null, iTopK, /*searchWidth=*/ 8,
            /*threadBlockSize=*/ 0, /*maxIterations=*/ 0, CagraSearchParams.SearchAlgo.AUTO),
        k);
  }
}
```

| `itopk_size` | predicted `max_traversed_nodes` | vs `2^30` | observed |
| --- | --- | --- | --- |
| 500,000,000 | 531,250,000 | under | clean `RAFT_EXPECTS` failure at `search_plan.cuh:288`, 77 ms |
| 1,000,000,000 | 1,062,500,000 | under | clean `RAFT_EXPECTS` failure, 20 ms |
| 1,100,000,000 | 1,168,750,000 | over | **hangs**, no error, still running at a 60 s timeout |
| 2,147,483,647 | 2,281,701,376 | over | **hangs**, no error |

The transition sits exactly where the wrap predicts. `max_traversed_nodes = max(search_width, ceildiv(itopk_size, 32)) * max(32, max_iterations)`; the loop can still terminate while the required `bitlen <= 31`, i.e. while `max_traversed_nodes <= 2^30`, and hangs above it.

The stuck thread stays `RUNNABLE` inside `cuvsCagraSearchMultiPartition` and is not recoverable by the caller's timeout:

```
1) Thread[id=107, ..., state=RUNNABLE, ...]
     at com.nvidia.cuvs.internal.panama.headers_h.cuvsCagraSearchMultiPartition(headers_h.java:27697)
     at com.nvidia.cuvs.internal.MultiPartitionCagraSearchImpl.search(MultiPartitionCagraSearchImpl.java:185)
```

**Expected behavior**

An `itopk_size` that requires a hash table beyond the supported size should be rejected with the existing `RAFT_EXPECTS` error, exactly as it already is for the `bitlen` 26-31 range. It should never hang.

**Suggested direction**

Bound the loop rather than only checking afterwards, e.g. stop at the documented maximum and fail there:

```cpp
while (hash_bitlen <= 25 && max_traversed_nodes > hashmap::get_size(hash_bitlen) * max_fill_rate) {
  hash_bitlen += 1;
}
RAFT_EXPECTS(hash_bitlen <= 25, "...");
```

Making `get_size()` safe for `bitlen >= 32` (returning `uint64_t`, or asserting on the input) would also remove the undefined behaviour, though the loop still needs its own bound to fail fast.

Worth noting the same shape appears in the other sizing loops in `calc_hashmap_params()` (lines 270, 303, 343): each increments the bitlen unbounded and checks the limit only afterwards. Line 285 is the one confirmed here. Lines 303 and 343 size from user-controlled `itopk_size`/`search_width`, so they look reachable the same way for the non-MULTI_CTA algorithms, though I have not reproduced those. Line 270 sizes from `graph_degree` only, so in practice it stays far below the wrap. Probably worth fixing together.

 - Environment location: Bare-metal
 - Method of cuVS install: from source, VERSION `26.10.00`
 - GPU: NVIDIA RTX PRO 6000 Blackwell Server Edition

**Additional context**

Found while adding parameter-boundary validation to `cuvs-lucene` (#2516). Because the correct upper bound depends on the resolved algorithm, `max_iterations`, graph degree and dataset size, the Java layer does not try to derive it; it relies on native CAGRA rejecting unsupported combinations, which works for the `bitlen` 26-31 range but hangs above it.

Thanks to @dantegd for pointing at `calc_hashmap_params()` and the `1U << bitlen` shift range as the likely cause , this issue confirms that diagnosis and pins the threshold.

shaunakkapur:
Closes #2523.

## Problem

`search_plan_impl::calc_hashmap_params()` sizes its hash tables with loops that grow a bit length until the requested table fits, and check the supported maximum only *after* the loop:

```cpp
hash_bitlen = min_bitlen;
while (max_traversed_nodes > hashmap::get_size(hash_bitlen) * max_fill_rate) {
  hash_bitlen += 1;
}
RAFT_EXPECTS(hash_bitlen <= 25, "hash_bitlen cannot be largen than 25 (32M)");
```

`hashmap::get_size()` is `1U << bitlen`, which is undefined once `bitlen` reaches 32. In practice the shift count is masked to 5 bits, so `get_size(32)` wraps to `1`, `get_size(33)` to `2`, and so on. `hash_bitlen` is `int64_t`, so for any request needing a table larger than `2^31` entries the loop condition stays permanently true, the bit length grows without bound, and the `RAFT_EXPECTS` is never reached. The search hangs on the host rather than returning an error.

`check_params()` only caps `itopk_size` at 1024 for `SINGLE_CTA`, so `MULTI_CTA` and `MULTI_KERNEL` are the two algorithms that can reach the sizing loops with an oversized request. A normal single-query `AUTO` search resolves to `MULTI_CTA`.

## Fix

Bound each of the four sizing loops by the same maximum its post-loop check already uses, so an oversized request exits the loop and raises the existing error instead of spinning:

```cpp
while (hash_bitlen <= max_hash_bitlen_mc &&
       max_traversed_nodes > hashmap::get_size(hash_bitlen) * max_fill_rate) {
  hash_bitlen += 1;
}
RAFT_EXPECTS(hash_bitlen <= max_hash_bitlen_mc, "hash_bitlen cannot be largen than 25 (32M)");
```

Each limit is hoisted into a local constant so the loop bound and the check cannot drift apart, which is the failure mode this bug came from.

All four loops are covered (lines 270, 285, 303, 343). Only 285, 303 and 343 are reachable with user controlled sizes; 270 is bounded by `graph_degree` in practice and is bounded here for consistency.

## Behaviour

No change for any request that was previously accepted. A request that previously hung now raises the error it should always have raised. A request that previously errored still errors.

## Testing

New regression test `cpp/tests/neighbors/ann_cagra/bug_issue_2523_hashmap_bitlen.cu`, registered under `NEIGHBORS_ANN_CAGRA_TEST_BUGS`, covering `MULTI_CTA` oversized, `MULTI_KERNEL` oversized, and valid `itopk_size` on all three algorithms.

Note that a regression here resurfaces as a **test timeout** rather than a failed assertion, since the failure mode is a hang. This is called out in the test file header.

I also checked that the test actually catches this. With the fix reverted and nothing else changed, `MultiCtaOversizedItopkThrows` hangs and gets killed at the 300s timeout (exit 124) instead of failing an assertion. With the fix back in, it passes in about 1.6s.

Verified on an RTX PRO 6000 Blackwell:

- `itopk_size = 1,100,000,000` previously hung indefinitely (confirmed past a 60s timeout, with the thread stuck in native code). It now raises the `hash_bitlen` error in a few seconds.
- Full `NEIGHBORS_ANN_CAGRA_TEST_BUGS` suite: 10/10 passed, including the five pre-existing reproducers.
- A standalone harness replicating the old and new loop logic across all four loops and 40 powers of two: identical results on every in-range input (136 cases), equivalent error signalling where the old loop already errored (104 cases), and clean errors in place of 88 cases where the old loop did not terminate. Zero mismatches.
- `pre-commit` hooks pass on the changed files.

## Notes for reviewers

- I did not change `hashmap::get_size()` itself. Making it safe for `bitlen >= 32` (returning `uint64_t`, or asserting on its input) would remove the undefined behaviour at the source, but it is used in device code and by callers that assign to `uint32_t`, so it is a wider change. Bounding the loops fixes the reachable defect locally. Happy to follow up on `get_size()` separately if you would prefer that hardening too.
- The copyright header line in `search_plan.cuh` was rewritten by the `verify-copyright` pre-commit hook to the canonical form. That is unrelated to the fix but the hook applies it to any touched file.

### DIFF
--- a/cpp/tests/neighbors/ann_cagra/bug_issue_2523_hashmap_bitlen.cu
+++ b/cpp/tests/neighbors/ann_cagra/bug_issue_2523_hashmap_bitlen.cu
@@ -1,0 +1,135 @@
+/*
+ * SPDX-FileCopyrightText: Copyright (c) 2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
+ * SPDX-License-Identifier: Apache-2.0
+ */
+
+#include <gtest/gtest.h>
+
+#include "../cagra_padded_build_helpers.cuh"
+#include <cuvs/neighbors/cagra.hpp>
+
+#include <raft/core/device_mdarray.hpp>
+#include <raft/core/device_resources.hpp>
+#include <raft/core/error.hpp>
+#include <raft/random/rng.cuh>
+
+#include <cstdint>
+#include <utility>
+
+namespace cuvs::neighbors::cagra {
+
+/**
+ * @brief Regression test for issue #2523: hash table sizing must not loop forever.
+ *
+ * The hash tables in search_plan_impl::calc_hashmap_params() are sized by loops that grow
+ * a bit length until the requested table fits, with the supported maximum checked only
+ * after the loop. hashmap::get_size() is `1U << bitlen`, which is undefined once bitlen
+ * reaches 32 and in practice wraps back to a small value, so a request that needs a table
+ * larger than 2^31 entries made the loop condition permanently true. The bit length grew
+ * without bound, the post-loop RAFT_EXPECTS was never reached, and the search hung on the
+ * host instead of failing.
+ *
+ * check_params() only caps itopk_size at 1024 for SINGLE_CTA, so MULTI_CTA and MULTI_KERNEL
+ * are the two algorithms that can reach the sizing loops with an oversized request.
+ *
+ * These searches must now raise the existing "hash_bitlen cannot be larger than ..." error
+ * rather than hanging. A regression reappears as a test timeout, not a failed assertion.
+ */
+class cagra_hashmap_bitlen_no_hang_test : public ::testing::Test {
+ public:
+  using data_type  = float;
+  using index_type = uint32_t;
+
+ protected:
+  // Large enough that the traversed-node hash table would need more than 2^31 entries,
+  // which is what previously drove the sizing loop past the wrap point.
+  constexpr static size_t oversized_itopk = 1'100'000'000;
+  constexpr static size_t valid_itopk     = 64;
+
+  constexpr static int64_t n_dataset = 1000;
+  constexpr static int64_t n_dim     = 32;
+  constexpr static int64_t n_queries = 4;
+  constexpr static int64_t k         = 10;
+
+  void SetUp() override
+  {
+    dataset.emplace(raft::make_device_matrix<data_type, int64_t>(res, n_dataset, n_dim));
+    queries.emplace(raft::make_device_matrix<data_type, int64_t>(res, n_queries, n_dim));
+    neighbors.emplace(raft::make_device_matrix<index_type, int64_t>(res, n_queries, k));
+    distances.emplace(raft::make_device_matrix<data_type, int64_t>(res, n_queries, k));
+
+    raft::random::RngState r(1234ULL);
+    raft::random::uniform(
+      res, r, dataset->data_handle(), n_dataset * n_dim, data_type(-1), data_type(1));
+    raft::random::uniform(
+      res, r, queries->data_handle(), n_queries * n_dim, data_type(-1), data_type(1));
+
+    cagra::index_params index_params;
+    index_params.graph_degree              = 32;
+    index_params.intermediate_graph_degree = 64;
+
+    padded_.emplace(res, raft::make_const_mdspan(dataset->view()));
+    index_.emplace(cagra::build(res, index_params, padded_->view));
+    raft::resource::sync_stream(res);
+  }
+
+  void TearDown() override
+  {
+    index_.reset();
+    padded_.reset();
+    dataset.reset();
+    queries.reset();
+    neighbors.reset();
+    distances.reset();
+    raft::resource::sync_stream(res);
+  }
+
+  void search_with(cagra::search_algo algo, size_t itopk_size, uint32_t max_iterations)
+  {
+    cagra::search_params search_params;
+    search_params.algo           = algo;
+    search_params.itopk_size     = itopk_size;
+    search_params.search_width   = 8;
+    search_params.max_iterations = max_iterations;
+
+    cagra::search(res,
+                  search_params,
+                  *index_,
+                  raft::make_const_mdspan(queries->view()),
+                  neighbors->view(),
+                  distances->view());
+    raft::resource::sync_stream(res);
+  }
+
+  raft::resources res;
+  std::optional<cuvs::neighbors::test::padded_device_matrix_for_cagra<data_type>> padded_{};
+  std::optional<cagra::index<data_type, index_type>> index_         = std::nullopt;
+  std::optional<raft::device_matrix<data_type, int64_t>> dataset    = std::nullopt;
+  std::optional<raft::device_matrix<data_type, int64_t>> queries    = std::nullopt;
+  std::optional<raft::device_matrix<index_type, int64_t>> neighbors = std::nullopt;
+  std::optional<raft::device_matrix<data_type, int64_t>> distances  = std::nullopt;
+};
+
+// MULTI_CTA sizes the shared traversed-node table from itopk_size and search_width.
+// This is the path confirmed to hang before the fix.
+TEST_F(cagra_hashmap_bitlen_no_hang_test, MultiCtaOversizedItopkThrows)
+{
+  EXPECT_THROW(search_with(cagra::search_algo::MULTI_CTA, oversized_itopk, 0), raft::exception);
+}
+
+// MULTI_KERNEL reaches the small-hash loop first and then the normal-hash loop.
+// max_iterations is pinned so the sizing inputs do not depend on the auto-derived value.
+TEST_F(cagra_hashmap_bitlen_no_hang_test, MultiKernelOversizedItopkThrows)
+{
+  EXPECT_THROW(search_with(cagra::search_algo::MULTI_KERNEL, oversized_itopk, 32), raft::exception);
+}
+
+// Guards against the bound rejecting requests that were previously accepted.
+TEST_F(cagra_hashmap_bitlen_no_hang_test, ValidItopkStillSucceeds)
+{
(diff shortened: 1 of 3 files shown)


cpp/tests/neighbors/ann_cagra/bug_issue_2523_hashmap_bitlen.cu
```diff
@@ -0,0 +1,135 @@
...
+
+// MULTI_KERNEL reaches the small-hash loop first and then the normal-hash loop.
+// max_iterations is pinned so the sizing inputs do not depend on the auto-derived value.
+TEST_F(cagra_hashmap_bitlen_no_hang_test, MultiKernelOversizedItopkThrows)
+{
+  EXPECT_THROW(search_with(cagra::search_algo::MULTI_KERNEL, oversized_itopk, 32), raft::exception);
+}
+
+// Guards against the bound rejecting requests that were previously accepted.
+TEST_F(cagra_hashmap_bitlen_no_hang_test, ValidItopkStillSucceeds)
```
dantegd, line 132: Could we add a case where falling back to the normal hash succeeds? The valid case stays in the small hash, and the oversized case falls back but throws.
With this fixture and MULTI_KERNEL, setting max_iterations = 32 and itopk_size = 3840 lands exactly on the small-hash limit. Bumping it to 3872 should work with hash mode AUTO and fail with SMALL. What do you think?

dantegd (approved):
One follow-up: max_visited_nodes in detail/cagra/search_plan.cuh still narrows to uint32_t, so a large request can wrap around and look small. 

This predates the fix, so no need to block on it. What do you think about tracking it separately and softening the claim that every previously hanging request now throws? Checked the arithmetic and code path, but haven’t run this case.

cpp/tests/neighbors/ann_cagra/bug_issue_2523_hashmap_bitlen.cu
shaunakkapur: Added the boundary case with MULTI_KERNEL and max_iterations=32. It checks that 3840 succeeds with both modes, while 3872 succeeds with AUTO and throws with SMALL.

shaunakkapur:
That makes sense I confirmed with a planner-only reproduction and filed [#2684 ](https://github.com/NVIDIA/cuvs/issues/2684) to track it separately. I also narrowed the PR description so it no longer claims every previously hanging request now throws

PR #2538, opened 2026-09-01
Optimize transfer/compute overlap in out-of-core KMeans

viclafargue:
## Summary

This PR improves data-transfer and compute overlap for host-resident out-of-core KMeans using a cyclic two-buffer pipeline.

- Batch 1 follows batch 0 on the copy stream while batch 0 compute starts as soon as its transfer completes.
- Subsequent transfers overlap computation of the current batch.
- Buffer recycling continues across iteration boundaries, allowing batch 0 of the next pass to be prefetched during the previous pass.
- Device-resident inputs remain zero-copy.

## Implementation

- Adds a private KMeans batch loader with explicit staged, acquired, and reusable buffer states.
- Uses CUDA events so compute waits for H2D completion and a buffer cannot be overwritten until all of its consumers have been submitted.
- Applies the same dependency-driven scheduling to every batch, without special APIs or states for the first two batches.
- Uses persistent device scratch to avoid per-batch deallocation and its potential device-wide synchronization.
- Computes final inertia through the regular batched assignment and reduction pipeline, accumulates it on device, and copies only the final result to host.
- Leaves the shared ANN batch iterator unchanged.

## Benchmark under similar configuration

10 GiB pinned-host FP32 dataset (10,485,760 × 256), 2,560 clusters, three iterations, ten 1 GiB out-of-core batches, and 131,072-sample assignment tiles.

| Metric | `main` | PR |
|---|---:|---:|
| Median runtime | 1.7795 s | **0.7962 s** |
| Speedup | 1.00× | **2.23×** |
| Effective bulk throughput | 22.48 GiB/s | **50.24 GiB/s** |
| Profiled GPU span | 1773.12 ms | **785.65 ms** |
| H2D time | 752.29 ms | 752.38 ms |
| Kernel time | 1003.08 ms | 760.97 ms |
| H2D/kernel overlap | 0.00 ms | **727.91 ms** |
| H2D overlapped by kernels | 0.0% | **96.75%** |
| Kernels overlapped by H2D | 0.0% | **95.66%** |

## Profile

Main branch :
profile_main

This PR :
profile_pr

This PR (multi-GPU) :
multi_gpu_profile


cpp/src/cluster/detail/kmeans.cuh
```diff
@@ -686,25 +689,56 @@ void kmeans_fit(
...
+  std::optional<rmm::cuda_stream> owned_batch_copy_stream;
+  if constexpr (!data_on_device) {
+    // Host KMeans must pipeline even when a minimal caller (including the Python binding) did not
+    // install a RAFT stream pool. Keep the stream owner before the iterators so their destructors
+    // drain all event dependencies before the stream itself is destroyed.
+    if (handle.has_resource_factory(raft::resource::resource_type::CUDA_STREAM_POOL) &&
+        raft::resource::get_stream_pool_size(handle) >= 1) {
+      batch_copy_stream = raft::resource::get_stream_from_stream_pool(handle);
+    } else {
+      owned_batch_copy_stream.emplace(rmm::cuda_stream::flags::non_blocking);
```
tarang-jain: Lets not create a prefetch stream on our own. It is solely controlled by the raft resource. The resource should give full control of concurrent streams to the user.

cpp/src/cluster/detail/kmeans_batch_loader.cuh
```diff
@@ -0,0 +1,355 @@
...
+template <typename DataT, typename IndexT>
+class kmeans_batch_loader<DataT, IndexT, true> {
+ public:
+  kmeans_batch_loader(raft::resources const& res,
+                      DataT const* source,
+                      IndexT n_rows,
+                      IndexT row_width,
+                      IndexT batch_size,
+                      rmm::cuda_stream_view copy_stream,
+                      rmm::device_async_resource_ref mr)
```
tarang-jain, line 98: hardcoding rmm::device_async_resource_ref wont give us problems? What if the currently set MR is not an async resource?

cpp/src/cluster/detail/kmeans_batch_loader.cuh
```diff
@@ -0,0 +1,355 @@
...
+                   positions_[slot] == batch.position_ && buffer_ptrs_[slot] == batch.data(),
+                 "KMeans attempted to release a batch that is not active");
+    return slot;
+  }
+
+  void queue_h2d(DataT* dst, std::size_t pos)
+  {
+    auto const& batch = batches_[pos];
+    const auto bytes  = batch.size * row_width_ * sizeof(DataT);
+    RAFT_CUDA_TRY(cudaMemcpyAsync(
```
tarang-jain: raft::copy()

tarang-jain: Correct me if I am wrong but the main new addition here in comparison to the ANN batch loader is the recycling at the final slots, right? So prefetch continues for the next iteration. Having an entirely new batch loader makes me a bit uneasy. Prefetch should not be algorithm specific. Can we combine this with the existing ANN batch loader (this can be done in a follow up)?

tarang-jain (changes requested):

cpp/src/cluster/detail/kmeans_batch_loader.cuh
viclafargue: `device_async_resource_ref` does not require a `cudaMallocAsync` allocator. It only requires the stream-aware MR interface. An allocator may still synchronize internally, which could affect performance but not correctness. A truly synchronous-only MR would need an adapter before RAFT could pass it here. The main requirement is that the referenced MR outlive the loader.

cpp/src/cluster/detail/kmeans_batch_loader.cuh
viclafargue: You're right that cyclic recycling across iteration boundaries is the main addition. However, the KMeans loader also uses CUDA events instead of per-batch host synchronization and supports multiple input partitions. The current ANN loader cannot provide this without broader changes to its iterator/writeback contract. I agree the mechanism is not inherently KMeans-specific, so we can extract a shared generic loader in a follow-up.

c/include/cuvs/core/c_api.h
```diff
@@ -211,6 +211,16 @@ CUVS_EXPORT cuvsError_t cuvsMultiGpuResourcesDestroy(cuvsResources_t res);
...
 CUVS_EXPORT cuvsError_t cuvsMultiGpuResourcesSetMemoryPool(cuvsResources_t res, int percent_of_free_memory);
+
+/**
+ * @brief Set a CUDA stream pool on all devices managed by the multi-GPU resources
+ *
+ * @param[in] res cuvsResources_t opaque C handle for multi-GPU resources
+ * @param[in] num_streams Number of CUDA streams in each device's pool
+ * @return cuvsError_t
+ */
+CUVS_EXPORT cuvsError_t cuvsMultiGpuResourcesSetStreamPool(cuvsResources_t res,
```
tarang-jain, line 241: again -- I am not entirely opposed to this because I see how it makes things convenient. But why are we providing an API for this specifically? SNMG resources or raft resources with nccl comms are not special in the way stream pools should be set, right? In general I think such APIs to directly modify raft resources, should be in raft and not cuvs. We can add them here and then they have to stick around until we are allowed to break the ABI.

cpp/src/cluster/detail/kmeans.cuh
```diff
@@ -686,25 +688,49 @@ void kmeans_fit(
...
+  auto batch_inertia      = raft::make_device_scalar<DataT>(handle, DataT{0});
+  auto batch_cost         = raft::make_device_scalar<DataT>(handle, DataT{0});
+  auto h_inertia          = raft::make_pinned_scalar<DataT>(handle, DataT{0});
 
   rmm::device_uvector<char> batch_workspace(device_buffer_samples, stream);
 
-  auto data_batches = cuvs::spatial::knn::detail::utils::make_batch_load_iterator<DataT>(
-    handle, X.data_handle(), n_samples, n_features, device_buffer_samples, stream);
+  auto batch_mr          = data_on_device ? raft::resource::get_workspace_resource_ref(handle)
+                                          : raft::resource::get_large_workspace_resource_ref(handle);
```
tarang-jain, line 704: if the data is already device accessible, why are we even using the workspace?

tarang-jain: Furthermore, if the large workspace is managed, the transfer speed is slightly slower (I found 55 GB/s versus 48 GB/s). I agree that the batchsize is not bounded here (it can be as large as the dataset) but I would argue for falling back to the large workspace only at the breaking point where allocating from the regular workspace is not possible. That calculation can get complicated to account for whether or not weights are present, so I'll tag @achirkin for some ideas.

cpp/src/cluster/detail/kmeans.cuh
```diff
@@ -758,6 +784,18 @@ void kmeans_fit(
     }
   };
 
+  auto prefetch_batch = [&](std::size_t batch_pos) {
+    (void)data_batches.prefetch(batch_pos);
+    if (weight_batches.has_value()) { (void)weight_batches->prefetch(batch_pos); }
+  };
+
+  bool input_pipeline_started = false;
+  auto start_input_pipeline   = [&] {
```
tarang-jain: Can we make this function part of the batch loader class, so starting the pipeline explicitly becomes part of the batch loader? It makes it more readable. Overall there has been an increase in these lambdas all over the codebase (not the most readable I think)

cpp/src/cluster/detail/kmeans.cuh
```diff
@@ -830,53 +864,44 @@ void kmeans_fit(
...
-        wt_it = weight_batches->begin();
-      }
-      for (const auto& data_batch : data_batches) {
-        IndexT cur_batch_size = static_cast<IndexT>(data_batch.size());
-        const DataT* wt_data  = nullptr;
-        if (wt_it.has_value()) {
-          wt_data = (**wt_it).data();
-          ++(*wt_it);
+      for (std::size_t batch_pos = 0; batch_pos < data_batches.num_batches(); ++batch_pos) {
+        const auto data_batch = data_batches.acquire(batch_pos);
```
tarang-jain: Lets stick to the terminology of the ann loader (prefetch instead of acquire)

cpp/src/cluster/detail/kmeans_batch_loader.cuh
```diff
@@ -0,0 +1,357 @@
...
+  auto operator=(kmeans_batch_loader const&) -> kmeans_batch_loader& = delete;
+  kmeans_batch_loader(kmeans_batch_loader&&)                         = delete;
+  auto operator=(kmeans_batch_loader&&) -> kmeans_batch_loader&      = delete;
+
+  ~kmeans_batch_loader() noexcept
+  {
+    if (!batches_.empty()) {
+      RAFT_CUDA_TRY_NO_THROW(cudaStreamSynchronize(raft::resource::get_cuda_stream(*res_)));
+    }
+    RAFT_CUDA_TRY_NO_THROW(cudaStreamSynchronize(copy_stream_));
```
tarang-jain: Please use raft APIs for these. I tend to find these, but these should not need to be pointed out anymore. You can make rules for agents to use raft APIs wherever possible.

cpp/src/cluster/detail/kmeans_common.cuh
```diff
@@ -446,6 +446,50 @@ EXTERN_TEMPLATE_MIN_CLUSTER_DISTANCE(double, int)
 
 #undef EXTERN_TEMPLATE_MIN_CLUSTER_DISTANCE
 
+/**
+ * @brief Compute the optionally weighted sum of distances to the nearest centroid.
+ *
+ * Unlike minClusterAndDistanceCompute, this path does not calculate cluster labels.
+ */
+template <typename DataT, typename IndexT>
+void cluster_cost(
```
tarang-jain: why do we have this new function? Its not used anywhere, right?

tarang-jain (changes requested):

cpp/src/cluster/detail/kmeans_batch_loader.cuh
tarang-jain: https://github.com/NVIDIA/cuvs/issues/2581 I created this and assigned you @viclafargue

cpp/src/cluster/detail/kmeans.cuh
tarang-jain: Actually since this is unbounded (only bounded by dataset size) lets just stick to large_mr @viclafargue. Since compute will overlap, its not a big deal.

cpp/src/cluster/detail/kmeans.cuh
```diff
@@ -927,46 +958,89 @@ void kmeans_fit(
...
-                                            batch_sw);
-
-        iter_inertia += batch_cost;
+        auto l2_norm_view = raft::make_device_vector_view<const DataT, IndexT>(
+          L2NormBatch.data_handle(), cur_batch_size);
+        auto min_cad_view =
+          raft::make_device_vector_view<raft::KeyValuePair<IndexT, DataT>, IndexT>(
+            minClusterAndDistance.data_handle(), cur_batch_size);
+
+        minClusterAndDistanceCompute<DataT, IndexT>(handle,
```
irina-resh-nvda: Why do we use this instead of minClusterDistanceCompute? Do we need the cluster labels anywhere?

c/include/cuvs/core/c_api.h
viclafargue: The resources inside of the `device_resources_snmg` helper can be indeed be accessed individually. However, for convenience a user would in practice set memory or stream pools from utility functions. I agree that cuVS should not implement setting a stream pool on every resources. This is a temporary solution as I am aiming to get this be implemented in RAFT (see https://github.com/NVIDIA/raft/pull/3149). Would be glad to get your review on this.

That said the feature should be accessible from cuVS C and Python API. This is where the `cuvsResources_t` and `MultiGpuResources` wrappers come into play. We could also consider making all of cuVS resources come from RAFT, but this would be out of the scope of this PR.

tarang-jain: Okay! The problem with adding C APIs is that they have to stay for up to 6 months before we can deprecate them. If cuVS python API does not have a way to set the stream pool on the resource, this function makes sense. We have to be able to support async H2D in python.

cpp/src/cluster/detail/kmeans_common.cuh
viclafargue: This was to avoid device-wide syncs during the final inertia calculation. Removed it for now.

cpp/src/cluster/detail/kmeans.cuh
viclafargue: The final inertia can be computed with the optimized cuTile path, but it has lower precision. I reverted back to calling the `cluster_cost` function.

cpp/src/cluster/detail/kmeans.cuh
```diff
@@ -56,6 +58,50 @@ namespace cuvs::cluster::kmeans::detail {
...
+  rmm::device_uvector<char>& workspace;
+};
+
+template <typename DataT, typename IndexT>
+void cluster_cost(
+  raft::resources const& handle,
+  raft::device_matrix_view<const DataT, IndexT> X,
+  raft::device_matrix_view<const DataT, IndexT> centroids,
+  raft::device_scalar_view<DataT> cost,
+  cluster_cost_workspace<DataT, IndexT>& scratch,
```
tarang-jain: why overcomplicate this function? cluster_cost_workspace is not used anywhere else, right? We dont use this pattern at other places in the codebase. Just pass the arrays directly.

tarang-jain (changes requested):

python/cuvs/cuvs/common/resources.pyx
```diff
@@ -89,6 +91,31 @@ cdef class Resources:
...
+
+        Parameters
+        ----------
+        percent_of_free_memory : int
+            Percentage of free device memory to allocate for the pool.
+        """
+        check_cuvs(cuvsResourcesSetMemoryPool(
+            self.c_obj, percent_of_free_memory))
+
+    def set_stream_pool(self, num_streams=1):
```
aamijar: So on the python side, the user is required to setup a stream pool to get the compute/H2D optimization?
Shouldn't users be getting this optimization by default and not need to know about setting up stream pool?

tarang-jain:
In addition to splitting this PR into the C++ and C/Python components, @viclafargue can you make sure to work on top of https://github.com/NVIDIA/cuvs/pull/2249? This PR should not be a regression to that because it might be making some changes to `minClusterAndDistanceCompute`.

irina-resh-nvda (approved):

python/cuvs/cuvs/common/resources.pyx
viclafargue: I had the same question initially, but the decision was made to require users to configure a stream pool explicitly. I’ve made sure this requirement is clearly documented. The C and Python API changes have been moved to #2642.

cpp/src/cluster/detail/kmeans.cuh
```diff
@@ -684,27 +723,48 @@ void kmeans_fit(
...
   rmm::device_uvector<char> batch_workspace(device_buffer_samples, stream);
 
-  auto data_batches = cuvs::spatial::knn::detail::utils::make_batch_load_iterator<DataT>(
-    handle, X.data_handle(), n_samples, n_features, device_buffer_samples, stream);
+  auto batch_mr          = data_on_device ? raft::resource::get_workspace_resource_ref(handle)
+                                          : raft::resource::get_large_workspace_resource_ref(handle);
+  auto batch_copy_stream = raft::resource::get_cuda_stream(handle);
+  if constexpr (!data_on_device) {
+    if (handle.has_resource_factory(raft::resource::resource_type::CUDA_STREAM_POOL) &&
+        raft::resource::get_stream_pool_size(handle) >= 1) {
```
tarang-jain: we have a function to do exactly this: `get_prefetch_stream` in the detail namespace. Lets reuse that.

cpp/src/cluster/detail/kmeans_batch_loader.cuh
```diff
@@ -0,0 +1,365 @@
...
+    : kmeans_batch_loader(res,
+                          std::vector<kmeans_input_partition<DataT, IndexT>>{{source, n_rows}},
+                          row_width,
+                          batch_size,
+                          copy_stream,
+                          mr)
+  {
+  }
+
+  kmeans_batch_loader(raft::resources const&,
```
tarang-jain, line 111: Looks like its not even using the memory resource. I want to reassert my earlier comment about not using the regular workspace MR if the data is on device. Lets only use the large MR. Correct my understanding if I am missing something.

cpp/src/cluster/detail/kmeans_batch_loader.cuh
```diff
@@ -0,0 +1,365 @@
...
+                      std::vector<kmeans_input_partition<DataT, IndexT>> const& partitions,
+                      IndexT row_width,
```
tarang-jain: why not just store mdspans? We can avoid maintaining row_width_ and n_rows.

cpp/src/cluster/detail/kmeans_batch_loader.cuh
```diff
@@ -0,0 +1,365 @@
...
+};
+
+/** A contiguous KMeans input batch accessible from the main CUDA stream. */
+template <typename DataT>
+class kmeans_batch {
+ public:
+  [[nodiscard]] auto data() const noexcept -> DataT const* { return data_; }
+  [[nodiscard]] auto size() const noexcept -> std::size_t { return size_; }
+  [[nodiscard]] auto offset() const noexcept -> std::size_t { return offset_; }
+  [[nodiscard]] auto partition() const noexcept -> std::size_t { return partition_; }
```
tarang-jain: these methods should be inside the descriptor and the batch should just hold a reference to the descriptor.

cpp/src/cluster/detail/kmeans_batch_loader.cuh
```diff
@@ -0,0 +1,365 @@
...
+
+ private:
+  template <typename, typename, bool>
+  friend class kmeans_batch_loader;
```
tarang-jain, line 70: rather than friend-ing why not make the constructor public?

cpp/src/cluster/detail/kmeans_batch_loader.cuh
```diff
@@ -0,0 +1,365 @@
...
+ * Read-only batch loader used only by KMeans.
+ *
+ * The device specialization is a zero-copy view. The host specialization below owns the
+ * two-buffer, cyclic H2D pipeline needed by out-of-core KMeans.
+ */
+template <typename DataT, typename IndexT, bool DataOnDevice>
+class kmeans_batch_loader;
+
+template <typename DataT, typename IndexT>
+class kmeans_batch_loader<DataT, IndexT, true> {
```
tarang-jain, line 96: why do we even have this device loader? We are not even batching when the data is device accessible. We can straight away iterate over the partitions. It just creates additional maintenance overhead -- prefetch stream that wont be used, a second slot that wont be used etc. Lets just avoid using this altogether? process_batch can continue to consume mdspans, right? We can conditionally acquire a device mdspan for the host prefetch route.

tarang-jain (changes requested):

python/cuvs/cuvs/cluster/kmeans/kmeans.pyx
```diff
@@ -263,9 +275,17 @@ def fit(
...
-    >>> X_host = np.random.random((10_000_000, 128)).astype(np.float32)
+    >>> import cupyx
+    >>> from cuvs.common import Resources
+    >>> X_host = cupyx.empty_pinned((10_000_000, 128), dtype=np.float32)
+    >>> _ = np.random.default_rng().random(
+    ...     X_host.shape, dtype=np.float32, out=X_host)
     >>> params = KMeansParams(n_clusters=1000, device_buffer_samples=1_000_000)
-    >>> centroids, inertia, n_iter = fit(params, X_host)
+    >>> resources = Resources()
+    >>> resources.set_memory_pool(80)  # Percentage of currently free device memory
```
tarang-jain: these python APIs are no longer exposed in this PR no? The docs need to be fixed.

python/cuvs/cuvs/cluster/mg/kmeans/kmeans.pyx
```diff
@@ -76,6 +85,24 @@ def fit(
...
+    >>> from cuvs.cluster.mg import kmeans
+    >>> from cuvs.common import MultiGpuResources
+    >>> X = cupyx.empty_pinned((10_000_000, 128), dtype=np.float32)
+    >>> _ = np.random.default_rng().random(
+    ...     X.shape, dtype=np.float32, out=X)
+    >>> params = KMeansParams(n_clusters=1000,
+    ...                       device_buffer_samples=1_000_000)
+    >>> resources = MultiGpuResources()
+    >>> resources.set_memory_pool(80)
+    >>> resources.set_stream_pool(1)
```
tarang-jain: same comment as above.

tarang-jain (changes requested):

cpp/src/cluster/detail/kmeans_batch_loader.cuh
viclafargue: I switched both paths to always pass the large workspace resource. In the device-resident case, the selected loader specialization is zero-copy and does not allocate staging buffers, so the supplied resource is unused. For host-resident data, the large workspace resource is used to allocate the staging buffers.

cpp/src/cluster/detail/kmeans_batch_loader.cuh
viclafargue: `process_batch` can consume the mdspan directly, but the device loader is a lightweight pass-through, matching the previous ANN loader behavior. It does not allocate staging buffers or a second slot, and its prefetch operations are no-ops. Keeping it avoids adding separate host/device logic across the single- and multi-GPU loops, so I’d prefer to leave it as is.

tarang-jain (approved):

PR #2541, opened 2026-09-01
Enable ARM (aarch64) support for cuvs-java and cuvs-lucene

### LINKED ISSUE #1236: cuvs-java: support linux ARM
This issue has been filed in order to add support for ARM to the cuvs-java provider.  

Currently the provider implementation has the following check:
` if (Runtime.version().feature() > 21 && isLinuxAmd64())  { throw UOE }`

Since cuVS supports ARM, in theory we can just build and test on ARM, and if all goes well then just update `isLinuxAmd64` to allow `aarch64`.

imotov:
The C++ core already builds and is tested on both amd64 and arm64; cuvs-java had a hardcoded amd64-only gate that cuvs-lucene inherited.

Java bytecode is portable, but cuvs-java isn't pure Java — it uses the Panama FFM API to call into native libcuvs_c.so, and those bindings are generated by jextract, which bakes struct/function ABI layouts (computed by parsing the C headers with clang for the host's target triple) into the generated .class files at build time. Reusing an amd64-generated binding on aarch64 should work, since no arch-conditional logic in the jextract-visible header surface changes any type layout, struct offset, or function signature between x86_64 and aarch64 Linux.

We therefore expect the plain (no-native-bundled) jar to be arch-portable in practice. To verify that, the aarch64 cuvs-lucene CI job installs the amd64-built jar and runs cuvs-lucene's full test suite against it on aarch64 hardware with an aarch64-native libcuvs_c.so.

Closes #1236.


.github/workflows/build.yaml
```diff
@@ -182,15 +181,17 @@ jobs:
...
       build_type: ${{ inputs.build_type || 'branch' }}
       branch: ${{ inputs.branch }}
-      arch: "amd64"
+      arch: "${{ matrix.ARCH }}"
       date: ${{ inputs.date }}
       container_image: "rapidsai/ci-conda:26.10-cuda${{ matrix.CUDA_VER }}-${{ matrix.LINUX_VER }}-py${{ matrix.PY_VER }}"
       script: "ci/build_java.sh"
-      artifact-name: "cuvs-java-cuda${{ matrix.CUDA_VER }}"
+      # amd64 is the official published artifact; keep its name unqualified and only suffix other arches so they don't collide with it.
+      artifact-name: "cuvs-java-cuda${{ matrix.CUDA_VER }}${{ matrix.ARCH != 'amd64' && format('-{0}', matrix.ARCH) || '' }}"
```
jameslamb: What do you mean here by "official published artifact" and what specifically would break if we appended the architecture unconditionally to it?

I think this is just the name of a GitHub Actions artifact used for passing files around between CI runs, not an identifier that makes it to Maven Central or anything.

jameslamb: Actually, maybe here's my confusion. Seeing this in the PR description:

> _We therefore expect the plain (no-native-bundled) jar to be arch-portable in practice_

Is that "plain jar" what's built by `build_java.sh`? And if so, shouldn't this job continue to just be run on `amd64`?

jameslamb (commented):
Implementation looks good to me but I'm a bit confused about the goal, left some questions. Sorry in advance if they are naive / ignorant 😅

.github/workflows/build.yaml
imotov: @jameslamb  Yep, that's a tricky one. So if you have any ideas on how to do it better, I could definitely use them. The problem is that part of cuvs-java is generated by Panama from C headers. Theoretically, that part is platform-dependent. In practice, though, in our case it's the same for both architectures. So on one hand, publishing two basically identical jars for different architectures seems silly; on the other hand, we should make sure that 1) it builds on arm, 2) it works on arm, and 3) we can use the amd64-built jar on arm - i.e. that our practical assumption that the jars are functionally the same actually holds.

jameslamb: Let me rephrase... if "we can use the amd64-build jar on arm" is true, why does "it builds on arm" matter?

Do we have users who are building these jars from source in arm64 environments?

imotov: I don't think any users build it from source on arm64 today, since there is a check preventing cuvs-java from running on arm64, which this PR removes. So until this commit is merged, nobody can use it on arm64 anyway. However, I think after this PR is merged it will useful to build on arm64 to catch breaking changes on that platform. Maybe the right answer is to build and test fully on arm64, and then separately test that arm64 works with the amd64 build?

imotov: I have removed building cuvs-java and cuvs-lucene on arm64 and instead added a couple of scripts to just run the tests with jars downloaded from amd64 build.

jameslamb: Ok great, I like it! It'll be good to have that tested in CI.

jameslamb (approved):
Thanks for the changes, I'm very happy with where this ended up 😊

mythrocks (approved):
👍

PR #2543, opened 2026-09-02
Move Java example projects under examples/java/

### LINKED ISSUE #2448: Lucene: Add `examples` option to build.sh
This will be consistent with other rapids libraries while making it much easier for the user.

### LINKED ISSUE #2542: Consolidate cuvs-java and cuvs-lucene examples in examples/java directory
At the moment most of the examples live in the `examples` directory in the root of the project. So, it makes sense to move `cuvs-java` and `cuvs-lucene` examples into `examples/java` as well.

imotov:
Move java/examples and java/cuvs-lucene/examples to examples/java/cuvs-java and examples/java/cuvs-lucene, matching the existing examples for C, C++ and Go.

Add a --build-java-examples flag to build.sh, java/build.sh, and java/cuvs-lucene/build.sh that compiles the examples against the jar just built, and wire it into ci/build_java.sh/ci/build_lucene.sh so CI catches drift between the examples and the cuvs-java/cuvs-lucene API. Fix the examples themselves, which had silently drifted from the current API (builder signatures now require CuVSResources, destroyIndex() -> close(), CagraQuery needs a CuVSMatrix, and CAGRA search now requires an explicit device-padded dataset).

Closes #2448
Closes #2542


.github/workflows/pr.yaml
```diff
@@ -159,6 +159,11 @@ jobs:
           - '!ci/test_wheel_cuvs.sh'
           - '!ci/validate_wheel.sh'
           - '!docs/**'
+          - '!examples/**'
+          - 'examples/build.sh'
+          - 'examples/c/**'
+          - 'examples/cmake/**'
+          - 'examples/cpp/**'
```
jameslamb, line 169:
```suggestion
          - '!examples/.gitignore'
          - '!examples/java'
```

I think it's be a bit simpler to keep these list as only exclusions (`!`) instead of mixing inclusions and exclusions. In my opinion, that makes them easier to reason about. Could you apply similar changes through the rest of the `changed-files` diff in this PR please?

jameslamb (approved):
approved with a minor suggestion

examples/java/cuvs-java/src/main/java/com/nvidia/cuvs/examples/HnswExample.java
```diff
@@ -108,8 +113,8 @@ public static void main(String[] args) throws Throwable {
       if (hnswIndexFile.exists()) {
         hnswIndexFile.delete();
       }
-      index.destroyIndex();
-      hnswIndex.destroyIndex();
+      index.close();
+      hnswIndex.close();
```
mythrocks, line 117: Yikes, thank you for fixing this.  We should have done that when we fixed #1252.  Sorry I missed that.

mythrocks (approved):
👍 Thank you, this is much more streamlined.

PR #2547, opened 2026-09-03
Keep the Fern API reference in sync with the sources

### LINKED ISSUE #2546: [DOC] Stale API Docs
## Report incorrect documentation

The cuVS API reference pages under `fern/pages/*_api/` are generated by `fern/scripts/generate_api_reference.py`, but in practice it is rarely run. To make things worse, individual generated pages get hand-edited (or maybe agent-edited), and CI can't catch the resulting drift because `fern/build_docs.sh` regenerates before running `fern check`, so it always validates freshly-generated output rather than what's committed.

I just ran the generator, and noticed that it touched 61 pages, adding 102 symbols that had never been documented and removing 33 that no longer exist in the source. At the moment the docs had drifted roughly five weeks.

imotov:
The generated API reference under `fern/pages/*_api/` had drifted roughly five weeks behind the code, because `fern/scripts/generate_api_reference.py` is rarely run and CI can't detect it - `fern/build_docs.sh` regenerates *before* running `fern check`, so it validates freshly-generated output rather than what's committed.

This PR consists of two commits:

- **`7e0b8ee97` Add fern docs generation to pre-commit** - adds a `fern-api-reference` hook that regenerates the pages and fails if they differ from what's committed, so drift can't accumulate again (CI picks this up for free via `ci/check_style.sh`). Also fixes four generator bugs it surfaced: the nav losing `pages/api_reference.md`, `"Neighbors IVF RaBitQ"` being rewritten to `"Rabitq"`, `private` members appearing under "Public Members", and Java nav entries reading `"Com Nvidia cuVS Cuvsdevicematrix"` instead of `CuVSDeviceMatrix`. Adds `--quiet` so the hook's output stays readable.
- **`76f998d77` Update all stale documents** - this commit is huge but it contains only the resulting regeneration.

Fixes #2546


jameslamb (approved):
love this pattern. As long as it doesn't take too long to long to run locally, approed.

cjnolet (approved):

PR #2548, opened 2026-09-03
Generate cubins from `cutile` at build time and embed them in C++ DSO

divyegala:
This takes advantage of the existing JIT LTO infrastructure and builds on top of it.

### DIFF
--- a/cpp/cmake/modules/compute_matrix_product.cmake
+++ b/cpp/cmake/modules/compute_matrix_product.cmake
@@ -8,6 +8,9 @@
 include_guard(GLOBAL)
 
 function(cuvs_find_build_python output_var)
+  # cuTile is a build dependency. In conda builds, it is installed in BUILD_PREFIX while CMake's
+  # default search can resolve the host interpreter from PREFIX instead. Use the build prefix so
+  # configure-time matrix expansion and build-time kernel exports see the cuTile package.
   if(DEFINED ENV{BUILD_PREFIX})
     set(Python_ROOT "$ENV{BUILD_PREFIX}")
   endif()
--- a/cpp/cmake/modules/compute_matrix_product.cmake
+++ b/cpp/cmake/modules/compute_matrix_product.cmake
@@ -1,11 +1,22 @@
 # =============================================================================
 # cmake-format: off
-# SPDX-FileCopyrightText: Copyright (c) 2025-2026, NVIDIA CORPORATION.
+# SPDX-FileCopyrightText: Copyright (c) 2025-2026, NVIDIA CORPORATION & AFFILIATES. All rights reserved.
 # SPDX-License-Identifier: Apache-2.0
 # cmake-format: on
 # =============================================================================
 
 include_guard(GLOBAL)
+
+function(cuvs_find_build_python output_var)
+  if(DEFINED ENV{BUILD_PREFIX})
+    set(Python_ROOT "$ENV{BUILD_PREFIX}")
+  endif()
+  find_package(Python REQUIRED COMPONENTS Interpreter)
+  set(${output_var}
+      "${Python_EXECUTABLE}"
+      PARENT_SCOPE
+  )
+endfunction()
 
 function(compute_matrix_product output_var)
   set(options)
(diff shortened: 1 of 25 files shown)


cpp/cmake/modules/compute_matrix_product.cmake
```diff
@@ -1,32 +1,45 @@
...
+  if(DEFINED ENV{BUILD_PREFIX})
+    set(Python_ROOT "$ENV{BUILD_PREFIX}")
+  endif()
+  find_package(Python REQUIRED COMPONENTS Interpreter)
+  set(${output_var}
+      "${Python_EXECUTABLE}"
+      PARENT_SCOPE
+  )
+endfunction()
+
```
KyleFromNVIDIA, line 23: Why has this function been added? What changed to make the plain `find_package()` insufficient?

cpp/cmake/modules/generate_cutile_kernels.cmake
```diff
@@ -0,0 +1,287 @@
...
+
+  compute_matrix_product(matrix_product MATRIX_JSON_FILE "${_CUTILE_MATRIX_JSON_FILE}")
+
+  string(JSON len LENGTH "${matrix_product}")
+  math(EXPR last "${len} - 1")
+
+  # cmake-lint: disable=C0103,E1120
+  foreach(i RANGE "${last}")
+    string(JSON matrix_json_entry GET "${matrix_product}" "${i}")
+    process_cutile_matrix_entry(
```
KyleFromNVIDIA, line 254: Please update `cpp/cmake/config.json` to recognize this function.

cpp/cmake/modules/generate_cutile_kernels.cmake
```diff
@@ -0,0 +1,287 @@
...
+  endif()
+  if(NOT _CUTILE_KERNEL_PYTHON)
+    message(FATAL_ERROR "generate_cutile_kernels: KERNEL_PYTHON is required")
+  endif()
+
+  _cutile_kernels_setup(
+    MATRIX_JSON_FILE "${_CUTILE_MATRIX_JSON_FILE}" OUTPUT_DIRECTORY "${_CUTILE_OUTPUT_DIRECTORY}"
+  )
+  if(NOT _CUTILE_SETUP_OK)
+    generate_cutile_kernels_stub()
```
KyleFromNVIDIA: This has no effect. The newly-set `CUVS_CUTILE_ENABLED` will be lost as soon as this function returns.

KyleFromNVIDIA (changes requested):

cpp/cmake/modules/compute_matrix_product.cmake
divyegala: `find_package` looks in `PREFIX` which is the `host` conda environment and not the build one. Adding hints to that function did not help.

cpp/include/cuvs/detail/jit_lto/TileAlgorithmPlanner.hpp
```diff
@@ -0,0 +1,70 @@
...
+#include <vector>
+
+#include <rtcx/algorithm_launcher.hpp>
+
+namespace cuvs::detail::jit_lto {
+
+struct TileLauncherCache {
+  std::shared_mutex mutex;
+  std::unordered_map<std::string, std::shared_ptr<rtcx::algorithm_launcher>> launchers;
+  std::unordered_set<std::string> build_failed;
```
tarang-jain: Question (for my own understanding): why do we need to keep track of failed builds here? We are not doing that in the caches for the other algos, right?
A failed compilation would be permanently abandoned i.e. it wont be attempted again?

divyegala: Yes, this is specifically to make sure we don't attempt to keep loading `cutile` when it has already determined that it is not available to avoid repeated calls to the cuda runtime. Between no availability for CUDA 12 or some devices on CUDA 13 (sm75), this helps to quickly exit.

cpp/cmake/modules/generate_cutile_kernels.cmake
```diff
@@ -0,0 +1,276 @@
...
+    return()
+  endif()
+
+  cuvs_find_build_python(Python3_EXECUTABLE)
+
+  find_program(
+    CUTILE_BIN2C
+    NAMES bin2c
+    PATHS ${CUDAToolkit_BIN_DIR} REQUIRED
+  )
```
KyleFromNVIDIA, line 52: CMake 4.3 adds its own `bin2c` implementation. When we require 4.3, I would like to use it.

KyleFromNVIDIA (approved):

cpp/include/cuvs/detail/jit_lto/TileAlgorithmPlanner.hpp
```diff
@@ -0,0 +1,74 @@
...
+namespace cuvs::detail::jit_lto {
+
+struct CutileRuntimeCapabilities;
+
+struct TileLauncherCache {
+  std::shared_mutex mutex;
+  std::unordered_map<std::string, std::shared_ptr<rtcx::algorithm_launcher>> launchers;
+  // Cache expected compatibility misses so an unsupported module is not loaded on every call.
+  // Unexpected CUDA errors are raised rather than inserted here.
+  std::unordered_set<std::string> unavailable_launchers;
```
tarang-jain, line 29: Oh this is better! Correct my understanding but maybe we could skip iterating over those unavailable combos in the first place rather than doing unordered_set + find here? But approving!

tarang-jain (approved):

cpp/include/cuvs/detail/jit_lto/TileAlgorithmPlanner.hpp
divyegala: There might be some unknown load failures even after the unavailable combos have been filtered through

PR #2555, opened 2026-09-04
Expose Vamana build and serialize through cuvs-java

shaunakkapur:
## Description

Adds a public `cuvs-java` API for building and serializing Vamana indexes using the generated Panama bindings that already exist.

The API follows the existing CAGRA provider pattern:

- `VamanaIndex` exposes build, dimensions, and DiskANN serialization.
- `VamanaIndexParams` exposes the native build parameters with Java-side validation.
- `VamanaIndexImpl` owns the native handle and preserves the lifetime of a dataset view.
- `CuVSProvider`, `JDKProvider`, and `UnsupportedProvider` expose the new builder.
- The Vamana documentation gains Java examples.

`serialize(Path filePrefix, boolean includeDataset)` intentionally accepts a path prefix because the native call can write the graph at that prefix and the dataset at `<prefix>.data`.

The integration tests cover supported host and device datatypes, both native L2 metrics, dimensions, serialization layout, parameter validation, and resource ownership. This includes the float16 and device-backed paths added by the current native Vamana implementation.

## Scope

This PR is build-and-serialize only. It does not add Vamana search, deserialization, `sector_aligned`, AiSAQ layout changes, or `cuvs-lucene` integration because those capabilities are not exposed by the current Vamana C API or belong to later work.

## Testing

All 19 tests in `VamanaBuildAndSerializeIT` pass in CI on both CUDA 12.9 and 13.3, alongside the existing unit tests, the Lucene build, and the C++ matrix. The suite covers float32 and unsigned byte builds, float16 on a device-backed matrix, both serialization modes, the DiskANN file layout assertions, `L2SqrtExpanded`, parameter validation against the native `RAFT_EXPECTS` checks, dataset ownership, and lifecycle.

Also verified locally against a `libcuvs.so` built from current `main`:

- `mvn verify`, 19 of 19 plus the existing unit tests
- `mvn spotless:check` and `ci/checks/run_spotless.sh`, no rewrites
- `javadoc:javadoc` with `doclint all,-missing`
- `cuvs-lucene` `test-compile` against the installed jar

An earlier run failed in `conda-cpp-build` on a RAFT and ucxx mismatch in the build environment, in `raft/comms/detail/std_comms.hpp`. That was unrelated to this PR, which contains no C++ or CMake changes, and it has since been resolved upstream.


java/cuvs-java/src/main/java/com/nvidia/cuvs/spi/CuVSProvider.java
```diff
@@ -136,6 +136,18 @@ BruteForceIndex.Builder newBruteForceIndexBuilder(CuVSResources cuVSResources)
...
       throws UnsupportedOperationException;
 
+  /**
+   * Creates a new VamanaIndex Builder.
+   * <p>
+   * This is a {@code default} method rather than an abstract one so that
+   * providers written against an earlier version of this interface keep
+   * compiling.
+   */
+  default VamanaIndex.Builder newVamanaIndexBuilder(CuVSResources cuVSResources)
```
imotov, line 146: I think it might be the best not make this default. I just realize that by making this default we are missing actually needed override in FilterCuVSProvider.

fern/pages/java_api/index.md
```diff
@@ -18,6 +18,8 @@ These pages are generated from the Java source files in `java/cuvs-java/src/main
 - [LibraryException](/api-reference/java-api-com-nvidia-cuvs-libraryexception)
 - [RowView](/api-reference/java-api-com-nvidia-cuvs-rowview)
 - [SynchronizedCuVSResources](/api-reference/java-api-com-nvidia-cuvs-synchronizedcuvsresources)
+- [VamanaIndex](/api-reference/java-api-com-nvidia-cuvs-vamanaindex)
+- [VamanaIndexParams](/api-reference/java-api-com-nvidia-cuvs-vamanaindexparams)
```
imotov, line 22: These two should probably go to the Nearest Neighbors section. Might require updating with this list https://github.com/NVIDIA/cuvs/blob/279febcd516acd31a8e680004816f2e3c4a07faf/fern/scripts/generate_api_reference.py#L3034-L3041

imotov (approved):
LGTM

Issue #2537, opened 2026-08-31
[BUG] ivf_rabitq needs seralize & deserilize steps before the index is searchable


tfeher:
**Describe the bug**

If I run `ivf_rabitq::build()` followed by `ivf_rabitq::search()`, it will not return correct neighbors. In contrast, other indices work correctly: an index reutrned by `build()` is searchable (assuming it fits memory).

To get correct results from the `ivf_rabitq` index we need to save to file and load again. This step is included in cuvs-bench as well as the unit tests.

**Steps/Code to reproduce bug**
In the following example if `save_load` is true, then recall would be correct, otherwise 0.

```c++
    cuvs::neighbors::ivf_rabitq::index_params index_params;
    index_params.n_lists      = n_lists;
    index_params.bits_per_dim = bits_per_dim;

    auto index = cuvs::neighbors::ivf_rabitq::build(
      res, index_params, raft::make_const_mdspan(d_dataset.view()));

    std::string save_path = "ivf_rabitq_minimal.ibin";
    if (save_load) {
      cuvs::neighbors::ivf_rabitq::serialize(res, save_path, index);
      cuvs::neighbors::ivf_rabitq::deserialize(res, save_path, &index);
    }

    cuvs::neighbors::ivf_rabitq::search_params search_params;
    search_params.n_probes = n_probes;
    cuvs::neighbors::ivf_rabitq::search(res,
                                        search_params,
                                        index,
                                        raft::make_const_mdspan(d_queries.view()),
                                        d_neighbors.view(),
                                        d_distances.view());
```

**Expected behavior**
It is expected that we can search right after build, without saving/loading the dataset.

**Additional context**

Full example code to illustrate the bug
https://gist.github.com/tfeher/77f9e956f4bc66747fb5be0996961c54

Usage
```
./RABITQ_MINIMAL 10000 128 64 8 8 rabitq 0
./RABITQ_MINIMAL 10000 128 64 8 8 rabitq 1
```

PR #2521, opened 2026-08-28
Migrate stream APIs from rmm::cuda_stream_view to cuda::stream_ref

bdice:
## Summary

Track the coordinated migration of stream APIs and call sites from `rmm::cuda_stream_view` to CCCL's `cuda::stream_ref`. This propagates `cuda::stream_ref` through RMM containers and memory resources, RAFT resource and handle APIs, downstream C++ interfaces, Python/Cython bindings, benchmarks, tests, and documentation.

This migrates affected cuVS API signatures and internal call sites while extracting raw stream handles only where CUDA, generated/JIT, or legacy APIs require them.

Depends on rapidsai/rmm#2372 and NVIDIA/raft#3129.

Tracked in rapidsai/build-planning#318.

## Migrations

- Pass `cuda::stream_ref` through stream pools, resource accessors, conditionals, and downstream APIs without converting to `rmm::cuda_stream_view`
- Use `cuda::stream_ref` constructions for default/legacy/per-thread streams
  - `rmm::cuda_stream_default` ➡️ `cuda::stream_ref{cudaStream_t{cudaStreamDefault}}`
  - `rmm::cuda_stream_legacy` ➡️ `cuda::stream_ref{cudaStreamLegacy}`
  - `rmm::cuda_stream_per_thread` ➡️ `cuda::stream_ref{cudaStreamPerThread}`
- Use `.get()` when calling an API that requires a raw `cudaStream_t`, including CUDA runtime, library, CUB, and legacy API boundaries (previously `rmm::cuda_stream_view` used `value()`)
- Use `.sync()` when synchronizing a `cuda::stream_ref` (previously `rmm::cuda_stream_view` used `synchronize()`)
- Update Cython declarations and call sites to pass stream references directly where supported

PR #2557, opened 2026-09-05
Adopt CUDA stream compatibility accessors

bdice:
## Summary

Use the `get()` and `sync()` compatibility aliases added in [RMM #2537](https://github.com/rapidsai/rmm/pull/2537). These spellings are shared by `rmm::cuda_stream_view` and `cuda::stream_ref`.

This preserves existing stream types and public APIs while extracting mechanical accessor updates from the broader [stream migration](https://github.com/rapidsai/build-planning/issues/318). It is independently buildable without [RMM #2372](https://github.com/rapidsai/rmm/pull/2372) and leaves the migration PR focused on actual type and signature changes.

This updates raw CUDA, library, kernel-launch, and legacy API boundaries throughout cuVS while preserving current stream types. Changes that require RAFT to return `cuda::stream_ref` remain in [cuVS #2521](https://github.com/NVIDIA/cuvs/pull/2521).


bdice:
Validated this preparation branch against RMM `upstream/main`: the local `-j0` build completed and the focused IVF-PQ test passed. The dependent migration PR #2521 has been rebased onto this branch.

bdice:
Added commit 62aaa352 to cover the final two debug-only `RAFT_CHECK_CUDA` raw-stream boundaries in Vamana and IVF-PQ. This preparation head builds successfully with `build-cuvs -j0` against RMM `upstream/main` and the RAFT preparation branch.

divyegala (approved):

PR #2564, opened 2026-09-08
wheels: get 'tileiras' from system install, not 'cuda-toolkit' wheels

jameslamb:
Follow-up to https://github.com/NVIDIA/cuvs/pull/2548

Contributes to https://github.com/rapidsai/build-planning/issues/324

* switches to wheel builds and `pip` devcontainers getting `tileiras` from the system-installed CTK, not `cuda-toolkit` wheels.

### Benefits of the `tileiras` change

* one less version to need to remember to update when we bump the CTK version we build against
* consistent with how we get NVCC from the system install, not wheels
* removes a source of patching in DLFW builds (where a system CTK is always preferred to one provided by wheels)

### How I tested this

relied on CI


dependencies.yaml
```diff
@@ -429,23 +429,14 @@ dependencies:
           - libcusolver-dev
           - libcusparse-dev
           - libnvjitlink-dev
+  # NOTE: 'tileiras' is needed too, but in CI that's provided by a system-installed CTK.
```
KyleFromNVIDIA: Please update this comment to clarify that we are deliberately going against the "Installing from PyPI" advice at https://pypi.org/project/cuda-tile/.

KyleFromNVIDIA (changes requested):

bdice (approved):

dependencies.yaml
jameslamb:
```suggestion
  # NOTE: 'tileiras' is needed too, but in CI that's provided by a system-installed CTK, so
  #              we do not use the [tileiras] extra recommended at https://pypi.org/project/cuda-tile/
```

ok.

KyleFromNVIDIA (approved):

PR #2565, opened 2026-09-08
Fix CMake discovery for wheel dependencies

bdice:
Adds wheel prefix discovery and propagates KvikIO through the installed cuVS CMake target to fix cuML builds in https://github.com/NVIDIA/cuml/pull/8573.

Part of rapidsai/build-planning#325.


jameslamb (approved):
These changes look like exactly what we want, thanks for fixing the `kvikio` linking and sorry I missed that in my review of #2257 .

I'm not in the `cuvs-cmake-codeowners` group, but approving anyway to show my support for this.

cjnolet (approved):

PR #105, opened 2024-05-08
Determine why tests in some PRs are failing

robertmaynard:

PR #2568, opened 2026-09-09
cuTile smoke test: add sm_121 fragment for GB10, document a driver JIT-link gap

zbrad:
## Summary

GB10 (compute capability 12.1, DGX Spark) rejects `cutile_smoke`'s only embedded cubin (`sm_120`) at `cudaLibraryGetKernel` with `cudaErrorNoKernelImageForDevice`, despite matching `find_compatible_cubin_fragment()`'s same-major/lower-minor forward-compatibility rule. Confirmed empirically on real GB10 hardware that this "forward compatible within a major family" assumption doesn't hold here.

- Adds an exact-match `cutile_arch_12_1` / `sm_121` fragment.
- `sm_121a` (the family-conditional target) isn't reachable yet — `tileiras` in the currently-published `cuda-tile` doesn't support any `-a` suffixed `--gpu-name` target (confirmed by testing `sm_90a`/`sm_100a`/`sm_120a`/`sm_121a`, all rejected identically); filed as [NVIDIA/cutile-python#105](https://github.com/NVIDIA/cutile-python/issues/105).
- Corrects `find_compatible_cubin_fragment()`'s doc comment, which stated cubins are forward compatible across minor revisions within a major family — not true for this device.

## A second, independent finding

Fixing the above surfaced a separate problem: actually *launching* any cuTile-compiled kernel on this GB10 box fails with `cudaErrorJitCompilerNotFound`. Traced via `strace` to a real, confirmed cause: `libnvidia-gpucomp.so`, needed at kernel-launch time to complete a runtime relocation cuTile's cubins carry (`.rela.nv.constant4` against `___NV_TILE_LAUNCH...`, absent from a plain nvcc-compiled cubin through the identical `cudaLibraryLoadData`/`cudaLibraryGetKernel`/`cudaLaunchKernelExC` sequence — that path works fine), isn't shipped by the driver branch tested against (580.173.02).

Since `is_expected_cutile_unavailable()` already treats `cudaErrorJitCompilerNotFound` as an expected-unavailable condition for the *load* step, `LaunchesCompatibleCubin` now also tolerates it at launch time (skip, not fail) for consistency. A new `CutileSmoke.RequiresJitLinkCapableDriver` test deliberately does **not** tolerate it, so this gap stays visible rather than silently skipped — it's expected to start passing once a driver ships the missing component, which is the signal to fold its guarantee back into `LaunchesCompatibleCubin` and delete both.

## Testing

- `CutileSmoke.ResolvesEveryEmbeddedArchitecture` — updated for the new fragment, verified passing.
- `CutileSmoke.LaunchesCompatibleCubin` — was failing (`ASSERT_NE(launcher, nullptr)`), now resolves the launcher correctly (confirms the arch fix); skips cleanly on this driver due to the second issue above.
- `CutileSmoke.RequiresJitLinkCapableDriver` (new) — fails as expected/documented on this driver branch.
- Rest of the GB10 gtest suite (34/35 binaries, ~8000+ tests) run clean against this change; the one other failure (`CLUSTER_TEST`, a KMeans++ random-subsample tolerance flake) is unrelated and reproducibly flaky independent of this change.

Assisted-by: Claude


cpp/tests/detail/jit_lto/cutile_smoke.cu
```diff
@@ -85,51 +153,55 @@ TEST(CutileSmoke, LaunchesCompatibleCubin)
...
+// LaunchesCompatibleCubin above -- this is expected to FAIL on any driver
+// branch missing libnvidia-gpucomp.so (confirmed absent from the pinned
+// 580.173.02 branch via strace; only present in an untested 595.45.04
+// package cached locally, not installed). Excluded from tuned/full_test.sh's
+// release gate (see that script) precisely so it stays red and visible
+// instead of silently passing or silently being skipped -- once a driver
+// upgrade ships the missing component, this test starts passing on its own,
+// which is the signal to fold its guarantee back into LaunchesCompatibleCubin
+// and delete this test and the skip branch above.
+TEST(CutileSmoke, RequiresJitLinkCapableDriver)
```
divyegala: This test needs to be opt-in, otherwise it will fail on machines that have an embedded cubin but their drivers are not JIT capable

divyegala (changes requested):
Thank you for catching this bug! Could you please re-target to `release/26.10` branch?

cpp/tests/detail/jit_lto/cutile_smoke.cu
```diff
@@ -85,51 +168,59 @@ TEST(CutileSmoke, LaunchesCompatibleCubin)
...
+    // failure, so a legitimately-missing driver component surfaces here as a
+    // generic runtime_error rather than a typed cudaError_t we could re-check
+    // against is_expected_cutile_unavailable() (which already treats
+    // cudaErrorJitCompilerNotFound as expected-unavailable, but only for the
+    // *load* step). Match on the same error name for the launch step: some
+    // driver branches don't ship the component cuTile's cubins need to
+    // complete their runtime relocation at launch time -- see
+    // DISABLED_RequiresJitLinkCapableDriver below, which tracks this
+    // explicitly.
+    if (std::string(e.what()).find("cudaErrorJitCompilerNotFound") != std::string::npos) {
```
dantegd, line 184: Could we handle this consistently for both eager and lazy module loading? With CUDA_MODULE_LOADING=EAGER, cudaLibraryLoadData can return cudaErrorJitCompilerNotFound, which try_get_launcher() turns into nullptr; the earlier ASSERT_NE then fails before we reach this skip. With LAZY, the same error is handled here.

cpp/include/cuvs/detail/jit_lto/cutile_module.hpp
```diff
@@ -30,8 +30,13 @@ struct CutileModuleImage {
 /**
  * Selects the newest compatible cubin in the device's compute-capability major family.
  *
- * CUDA cubins are forward compatible across minor revisions within a major family, so an SM 8.9
- * device can load SM 8.6 SASS and an SM 12.1 device can load SM 12.0 SASS.
+ * CUDA cubins are *usually* forward compatible across minor revisions within a major family (an
```
dantegd, line 33: This is more of a limitation of cuTile artifactos or in general?

cpp/include/cuvs/detail/jit_lto/cutile_arch_tags.hpp
```diff
@@ -39,6 +39,19 @@ struct cutile_arch_12_0 {
   static constexpr int cc_minor = 0;
 };
 
+/**
+ * GB10 (DGX Spark, compute capability 12.1) does not accept the base sm_120
+ * cubin as forward-compatible SASS -- cudaLibraryGetKernel rejects it with
+ * cudaErrorNoKernelImageForDevice despite same-major/lower-minor. This exact
+ * cc_major.cc_minor target is required. See NVIDIA/cutile-python#105 for why
+ * this isn't sm_121a (cuTile's tileiras doesn't support family-conditional
```
dantegd, line 47: [Issue #105](https://github.com/NVIDIA/cutile-python/issues/105) shows that tileiras rejects the a suffix, but I don’t see whether sm_120f was tried. Could we update the wording and mention sm_120f if it was also tested?

dantegd (changes requested):
Couple of general questions:

- The PR description says this test should start passing once the driver issue is fixed, but it’s now DISABLED_ and I couldn’t find any CI invocation that enables disabled tests. Could we describe it as a manual check and link a tracking issue, or add a non-gating GB10 job if it’s meant to signal when the issue is fixed?
- This also makes the current availability contract a little misleading. tileir_compat.hpp says compatible SASS needs no driver JIT, and TileAlgorithmPlanner.cpp can cache a launcher before its first dispatch, but this PR shows that launch can still fail because a driver-side component is missing. Since there aren’t production callers yet (right?), could we at least update the docs and track a structured late-unavailable path as a follow-up?

divyegala:
@zbrad we have decided to not host a specific `sm121` cubin. However, the bug should now be resolved and `sm121` will now use the old non-cutile backend.

We appreciate your contribution, and please feel free to close the PR or propose a different plan.

zbrad:
> @zbrad we have decided to not host a specific `sm121` cubin. However, the bug should now be resolved and `sm121` will now use the old non-cutile backend.
> 
> We appreciate your contribution, and please feel free to close the PR or propose a different plan.

Just to be clear, the "cutiles will not host sm121" means that the entire NVIDIA DGX Spark series will not be able to use cutiles?  So long as that is stated in requirements, that covers disclosure.  Could you also clarify if that means cutiles will not be supported on the upcoming RTX laptops?  I think they share the same architecture, but I don't have visibility into that.

divyegala:
> Just to be clear, the "cutiles will not host sm121" means that the entire NVIDIA DGX Spark series will not be able to use cutiles? So long as that is stated in requirements, that covers disclosure. Could you also clarify if that means cutiles will not be supported on the upcoming RTX laptops? I think they share the same architecture, but I don't have visibility into that.

Yes, at least temporarily. We are working on a solution to support cutile on DGX Spark and RTX laptops ASAP. The only solution available at the moment without embedding a cubin is to use driver JIT. Is there a specific reason you'd like to use cutile-backed KMeans on DGX Spark? KMeans will still be available via cutlass otherwise.

PR #2579, opened 2026-09-09
brute_force_prefiltered test warp boundary but with less input

robertmaynard:
For the longest running test items in brute_force_prefiltered.cu we can do a tweaks to improve runtime:
* Use smaller input dim values which aren't a multiple of the vector width to still validate serial trailing loads 
* Increase the sparsity but keep about the `kSddmmMaxSelectivity` threshold  

These combine together to remove 20 minutes from `NEIGHBORS_TEST` .


tfeher (approved):
LGTM, thanks @robertmaynard!

PR #2584, opened 2026-09-10
docs-build: switch to RTX Pro 6000, add changed-files filter

jameslamb:
Contributes to https://github.com/rapidsai/build-planning/issues/333

Updates to hopefully reduce how often `docs-build` jobs get stuck waiting for CI runners.

* switches those jobs to RTX Pro 6000 runners
* uses shared `docs-build` matrix to set job details

Also noticed that there's a `docs_build` group in the `changed-files` configuration, but it isn't used to limit when docs builds runs. This adds that as well.

### DIFF
--- a/.github/workflows/pr.yaml
+++ b/.github/workflows/pr.yaml
@@ -33,6 +33,7 @@ jobs:
       - go-build-matrix
       - go-build
       - docs-build
+      - docs-build-matrix
       - wheel-build-libcuvs
       - wheel-build-cuvs
       - wheel-tests-cuvs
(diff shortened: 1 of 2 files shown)


.github/workflows/pr.yaml
```diff
@@ -746,11 +755,15 @@ jobs:
       pull-requests: read
     secrets: inherit # zizmor: ignore[secrets-inherit]
     uses: rapidsai/shared-workflows/.github/workflows/custom-job.yaml@release/26.10
+    if: fromJSON(needs.changed-files.outputs.changed_file_groups).build_docs
```
jameslamb, line 758: This looked like an omission, given that there's configuration for it:

https://github.com/NVIDIA/cuvs/blob/29e1101b856e7d48f53887b5a5097a9f0a3360e9/.github/workflows/pr.yaml#L86-L89

This fixes that, so some types of PRs that don't affect the docs will not have a docs CI job.

.github/workflows/pr.yaml
```diff
@@ -737,7 +738,15 @@ jobs:
...
       script: "ci/build_go.sh"
+  docs-build-matrix:
+    permissions:
+      contents: read
+    uses: rapidsai/shared-workflows/.github/workflows/compute-matrix.yaml@release/26.10
+    with:
+      build_type: pull-request
+      matrix_name: docs-build
   docs-build:
+    needs: [docs-build-matrix]
```
jameslamb, line 749: I was surprised to not see any conda packages needed for the docs builds, but looks like that was an intentional change in #2067

gforsyth (approved):

bdice (approved):

PR #2600, opened 2026-09-12
Remove unimplemented cagra search apis

aamijar:
The cagra search APIs with vpq f32 are not implemented and result in a runtime error. Same goes for cagra search and MG cagra search on standard datasets.
Instead users should get a compile/link time error. We should simply remove the apis and the vpq_f32_index type which doesn't have any use cases currently.

Update: We will keep the cagra search and MG cagra search on standard dataset, and simply implement the path where if the alignment meets cagra's requirements we can call the search function. However, if it doesn't we can give an error.


cpp/include/cuvs/neighbors/cagra.hpp
```diff
@@ -1443,74 +1439,6 @@ void search(raft::resources const& res,
...
-// device_standard_index overloads (uint32_t neighbor indices)
-void search(raft::resources const& res,
-            cuvs::neighbors::cagra::search_params const& params,
-            const cuvs::neighbors::cagra::device_standard_index<float, uint32_t>& index,
-            raft::device_matrix_view<const float, int64_t, raft::row_major> queries,
-            raft::device_matrix_view<uint32_t, int64_t, raft::row_major> neighbors,
-            raft::device_matrix_view<float, int64_t, raft::row_major> distances,
-            const cuvs::neighbors::filtering::base_filter& sample_filter =
-              cuvs::neighbors::filtering::none_sample_filter{});
-void search(raft::resources const& res,
```
tarang-jain, line 1455: we can leave the device_standard_index APIs as is. We can create the padded view ourselves if needed. If the dim is not aligned, only then we can error out.

tarang-jain (changes requested):

cpp/include/cuvs/neighbors/cagra.hpp
aamijar: Alright I'll change it so that we call the search kernel if its already aligned, instead of always erroring out.

aamijar: Addressed in https://github.com/NVIDIA/cuvs/pull/2600/commits/9d02b04814b64bee5882ee6275486db48b102a52

PR #2609, opened 2026-09-15
Update Doxygen to 1.18.0

vuule:
## Description

Depends on #2636, which enables the docs build needed to verify this upgrade in CI.

This PR updates Doxygen to 1.18.0 and refreshes generated dependency files.

The Doxyfile update removes options that are obsolete in Doxygen 1.18 and preserves GitHub-compatible Markdown heading identifiers. Additional repository-specific documentation warning fixes may be required.


bdice (commented):
I don't think this project uses Doxygen anymore, it uses Fern for its docs. We can probably clean this up and remove Doxygen support instead.

vuule:
superseded by #2649

PR #2636, opened 2026-09-16
Fix docs-build changed-files dependency

vuule:
Declares changed-files as a docs-build dependency so the build_docs output is available when GitHub evaluates the job condition. Without this dependency, documentation builds are skipped even when build_docs is true.


bdice (approved):

bdice (approved):

vuule:
@bdice I think that another test trigger comment is needed here

PR #2642, opened 2026-09-17
Add memory and stream pool configuration to resource APIs

viclafargue:
Add memory and CUDA stream pool configuration to single- and multi-GPU resource APIs, including Python bindings and documentation.


cpp/include/cuvs/cluster/kmeans.hpp
```diff
@@ -148,9 +148,14 @@ struct params : base_params {
    * Only used by the batched (host-data) code path and ignored by
    * device-data overloads.
    *
+   * Inputs spanning multiple batches are automatically double-buffered. Budget about
```
tarang-jain, line 151: Is this true? Even if a pool is not provided, we are doubling the memory usage? Can we stick to one buffer for this case?

PR #2653, opened 2026-09-18
Parallelize bounded HNSW graph post-processing

nvzm123:
## Summary

This isolates the bounded post-ingest HNSW optimizations previously developed alongside #2481 and #2594. GPU CAGRA construction remains serialized; `writerThreads` now also controls bounded CPU work after construction.

## Changes

- Bounds effective workers by the requested count, available processors, and available work.
- Materializes eligible device-backed adjacency layers from one temporary host copy when the layer has at least 65,536 nodes and the copy is at most 4 GiB; larger device layers retain the serial path.
- Serializes level-zero neighbors in parallel 64 MiB waves while committing encoded bytes in deterministic node order.
- Preserves serial upper-layer writing and the existing on-disk format.
- Propagates `writerThreads` through ordinary, binary-quantized, and scalar-quantized accelerated writers.
- Preserves task-order failure suppression, interruption status, executor termination, and native-host cleanup.
- Documents the expanded `writerThreads` behavior and memory bounds.

## Validation

- `mvn -o spotless:check install`: 352 tests, 0 failures/errors, 30 skipped
- Deep1B 1M one-segment sanity run with `writerThreads=1`: 1,000,000 documents/vectors and 32,000,000 validated arcs
- Deep1B 1M one-segment sanity run with `writerThreads=16`: 1,000,000 documents/vectors and 32,000,000 validated arcs
- Deep1B 1M four-segment sanity run with `writerThreads=16`: 1,000,000 documents/vectors, four leaves, and 32,000,000 validated arcs
- All three retained indexes passed assertion-enabled `CheckIndex -level 2` and completed 980 measured queries after 20 warmups
- API reference generation was idempotent; `git diff --check` passed

The Deep1B runs are functional sanity checks, not controlled performance measurements.

## Integration note

This branch is based directly on `main` and contains only the post-ingest optimization. If #2476 lands first, this change needs a semantic rebase so that #2476's newer ownership, validation, and upper-layer fixes are retained.


fern/pages/user_guide/lucene.md
```diff
@@ -231,7 +231,9 @@ Both `AcceleratedHNSWParams` and `GPUSearchParams` default to a `HEURISTIC` stra
 - For the accelerated HNSW codecs, set `maxConn` and `beamWidth`, the HNSW parameters you would tune on the CPU. cuVS derives graph degrees and the build algorithm from them. These two values also configure the CPU fallback writer, so one setting covers both paths.
 - For the GPU search codec, set `buildQuality`. Higher values spend more build time for a higher-quality graph. This codec also passes `graphDegree` into the heuristic, so leave it at its default unless you intend to cap the graph.
 
-Increasing `writerThreads` raises index build concurrency. The accelerated HNSW codecs default to a single writer thread, while the GPU search codec defaults to 32.
+`writerThreads` controls native cuVS writer concurrency for both parameter types. For the accelerated HNSW codecs, it also sets the maximum number of CPU threads used to materialize and serialize the finished graph; the default of one keeps this post-build processing serial. The GPU search codec continues to use the setting only for native writer concurrency and defaults to 32.
```
jamxia155, line 234: `writerThreads` previously controlled threading behavior in cuVS CAGRA while the new usage is for after CAGRA has been built i.e. on the Lucene side. I would prefer to use a separate parameter, perhaps `graphSerializationThreads`.

java/cuvs-lucene/src/main/java/com/nvidia/cuvs/lucene/GPUBuiltHnswGraph.java
```diff
@@ -41,47 +55,197 @@ public class GPUBuiltHnswGraph extends HnswGraph {
...
+  }
+
+  /**
+   * Materializes disjoint node ranges concurrently. Each thread writes its own slots of {@code
+   * neighbors} and its own {@link NeighborArray} instances, so no synchronization is needed; {@code
+   * source} must be a host matrix (stateless {@code getRow}).
+   */
+  private static void fillNeighborArrayParallel(
+      CuVSMatrix source, NeighborArray[] neighbors, int size, int numThreads) throws IOException {
+    ExecutorService pool = Executors.newFixedThreadPool(Math.max(1, numThreads - 1));
```
jamxia155, line 212: It feels weird to spin up a thread pool, do some work, then shut it down all in the same method--aren't these pools supposed to amortize the thread creation overhead over repeated usage? Also, would we be in trouble if we have concurrent flushes each creating a separate thread pool? We might have to use a bounded thread pool that persists over the appropriate stage.

java/cuvs-lucene/src/main/java/com/nvidia/cuvs/lucene/GPUBuiltHnswGraph.java
```diff
@@ -41,47 +55,197 @@ public class GPUBuiltHnswGraph extends HnswGraph {
...
 
+  /** Node count below which parallel materialization is not worth the thread overhead. */
+  static final int PARALLEL_MIN_NODES = 1 << 16;
+
+  /**
+   * Maximum temporary native-host copy used to make a device adjacency safe for concurrent reads.
+   * Larger device matrices retain serial row access instead of risking a full-matrix native-memory
+   * spike on top of the Java {@link NeighborArray} representation.
+   */
+  static final long MAX_PARALLEL_GRAPH_COPY_BYTES = 4L << 30;
```
jamxia155, line 126: Can we use a higher cutoff? At 4GB, larger datasets (e.g. the 100M dataset you've benchmarked) probably have to fall back to the serial path.

java/cuvs-lucene/src/main/java/com/nvidia/cuvs/lucene/AcceleratedHNSWUtils.java
```diff
@@ -237,54 +337,153 @@ private static CuVSMatrix buildCagraGraphForSubset(
...
-          // Sorting step helps here
-          if (nnodes[i - 1] == nnodes[i]) {
+      writeLevelSerial(
+          graph, vectorIndex, level, sortedNodes, offsets[level], countOnLevel0, maxConn);
+    }
+    return offsets;
+  }
+
+  /** Node count below which parallel level-0 serialization is not worth the overhead. */
+  static final int PARALLEL_MIN_NODES = 1 << 16;
```
jamxia155, line 375: To avoid duplicating this definition, can we just use `GPUBuiltHnswGraph.PARALLEL_MIN_NODES` in this file?

java/cuvs-lucene/src/main/java/com/nvidia/cuvs/lucene/AcceleratedHNSWUtils.java
```diff
@@ -237,54 +337,153 @@ private static CuVSMatrix buildCagraGraphForSubset(
    */
   public static int[][] writeGraph(GPUBuiltHnswGraph graph, IndexOutput vectorIndex)
       throws IOException {
-    // write vectors' neighbors on each level into the vectorIndex file
+    return writeGraph(graph, vectorIndex, 1);
+  }
+
+  public static int[][] writeGraph(GPUBuiltHnswGraph graph, IndexOutput vectorIndex, int numThreads)
```
jamxia155, line 343: Should we keep this package-private for now and only make them public later as needed?

java/cuvs-lucene/src/main/java/com/nvidia/cuvs/lucene/GPUBuiltHnswGraph.java
```diff
@@ -41,47 +55,197 @@ public class GPUBuiltHnswGraph extends HnswGraph {
    */
   public GPUBuiltHnswGraph(
       int size, int dimensions, List<int[]> layerNodes, List<CuVSMatrix> layerAdjacencies) {
+    this(size, dimensions, materializeSerial(size, layerNodes, layerAdjacencies));
+  }
+
+  /** Builds a graph while materializing adjacency rows with the requested number of threads. */
+  public GPUBuiltHnswGraph(
```
jamxia155, line 62: Should we keep this package-private for now and only make them public later as needed?

PR #2658, opened 2026-09-18
Harden cuVS Lucene matrix ownership, fallback, and graph construction

nvzm123:
## Summary

This is an incremental follow-up to #2476. It keeps the memory-focused change small while centralizing related lifecycle, failure-handling, graph-integrity, and bounded-memory quantized-merge work for later decomposition.

Until #2476 merges, GitHub may show its prerequisite changes in this PR's Files view as well. The change unique to this PR is the final hardening commit layered directly on top of #2476.

### Pre-existing defects and lifecycle hardening

1. **Unbuilt matrix cleanup.** Matrix builders allocate native storage during construction. Writer, conversion, and upper-layer construction call sites changed here close builders when population or `build()` fails, preventing untransferred storage from becoming unreachable.

2. **Primary and suppressed failure preservation.** Shared lifecycle helpers retain the operation failure as primary, attach cleanup failures as suppressed, and close a successfully returned matrix if builder cleanup subsequently fails.

3. **Matrix conversion cleanup.** `CuVSDeviceMatrix.toHost()` and `CuVSHostMatrix.toDevice()` close a newly allocated destination if the copy fails. A destination-close failure is suppressed onto the copy failure.

4. **Device-builder initialization order.** CUDA stream acquisition happens before device-matrix allocation. If stream acquisition fails, no device allocation has occurred without a cleanup handle.

5. **Native error classification.** CUDA and cuVS status failures are reported as `LibraryException`. The asynchronous CUDA wrapper checks the returned status outside the foreign-call invocation catch so a native status failure is not converted into an `AssertionError`.

6. **GPU-search CAGRA fallback policy.** The writer no longer catches every `Throwable` and silently substitutes a brute-force index. Fallback is limited to a clean `LibraryException` during CAGRA construction, before persistence begins.

7. **Persistence and cleanup boundary.** CAGRA fallback is rejected after the serializer is invoked, after the output position changes, or when cleanup or another suppressed failure occurs. These cases fail indexing instead of risking mixed or incomplete output.

8. **GPU-search CAGRA ownership.** Cleanup accounts explicitly for the builder, original device matrix, CAGRA index, and padded dataset view or copy. The existing device-backed input is used directly, and resources close in dependency order while preserving cleanup failures.

9. **GPU-search brute-force ownership.** The brute-force dataset and index are both closed on success and failure. The private helper accepts `CuVSDeviceMatrix`, making its device-only requirement explicit.

10. **Accelerated-HNSW ownership.** Float, binary, and scalar writers close their dataset if CAGRA construction fails and otherwise transfer it to the successfully built index according to the existing CAGRA ownership contract.

11. **Temporary graph-matrix cleanup.** Single-vector adjacency matrices, upper-layer subset datasets, subset indexes, and remapped adjacency matrices close after eager graph materialization or on failure.

12. **Upper-layer ordinal remapping.** Negative native sentinels are preserved and later skipped. Valid subset-local ordinals are translated to absolute ordinals. Positive ordinals outside the subset are rejected instead of becoming self-references.

13. **Malformed adjacency rows.** Null, short, and long rows are rejected before they can create missing entries or implicit node-0 edges.

14. **Graph-integrity validation.** `GPUBuiltHnswGraph` validates adjacency data type, row count, degree, global ordinal bounds, duplicate upper nodes, zero-sized graphs, empty upper layers, nesting between successive levels, and upper-layer neighbor membership. Upper-node arrays are cloned so later caller mutation cannot alter the graph.

15. **Trivial quantized fields.** One-vector binary and scalar fields are handled before allocating the non-trivial native build matrix.

16. **Touched-path arithmetic and shape checks.** Dataset sizes, matrix widths, row counts, vector dimensions, count increments, and byte-width calculations use explicit validation and exact narrowing or arithmetic where these paths require an `int`.

### Safety required by the new two-pass paths

17. **Quantized replay-count validation.** Binary and scalar merge paths verify that the second iterator produces exactly the count observed while collecting quantization statistics. Both underflow and overflow fail before an incorrectly sized matrix can be used.

### New bounded-memory and lookup work

18. **Binary-quantized force merges.** The first pass collects centroid statistics. The second pass quantizes each row into one reusable byte buffer and copies it directly into one native host matrix, avoiding complete float-vector and quantized-vector heap lists for accelerated graph construction.

19. **Scalar-quantized force merges.** The first pass collects per-dimension extrema. The second pass uses one reusable encoded row and one native host matrix, with the same exact replay-count checks.

20. **Scalar flush conversion.** Signed-to-unsigned conversion reuses one row buffer instead of constructing a second `List<byte[]>`.

21. **Upper-layer subset construction.** Each upper layer copies only its selected rows into a native subset matrix using a reusable row buffer. It no longer constructs a heap two-dimensional subset array.

22. **Upper-layer lookup.** Internally generated sorted upper-node lists use binary search. The public constructor continues to accept valid unsorted lists and uses linear lookup for those inputs.

### Contracts and documentation

23. **CAGRA dataset ownership.** The existing behavior is documented and tested: the caller owns a supplied matrix until `build()` succeeds; a successful build transfers ownership to the returned index; a failed build leaves ownership with the caller.

24. **Generated API documentation.** Relevant generated pages are refreshed, newly added lifecycle and test seams are kept out of public-member listings, and affected `AcceleratedHNSWUtils` inline-code rendering is corrected.

## Compatibility and behavior changes

- No codec version constant or on-disk file layout changes.
- Valid graph inputs remain supported, including unsorted upper-node arrays.
- Malformed `GPUBuiltHnswGraph` inputs that were previously tolerated, silently rewritten, or allowed to fail later are rejected at construction. This includes zero-sized graphs, empty or non-nested upper levels, duplicate or out-of-range nodes, invalid row shapes, and upper neighbors outside their level.
- GPU-search CAGRA failures outside the narrow pre-persistence native-construction case fail indexing instead of silently falling back to brute force.
- Native CUDA and cuVS status failures now retain the `LibraryException` subtype.
- Quantization formulas are intentionally preserved. Golden tests cover binary output, reusable-buffer clearing, scalar negative values, and constant dimensions.
- The default accelerated-HNSW configuration still builds one layer. Upper-layer corrections apply when callers opt into additional layers.

## Explicit limitations

- This work relies on the closeable-builder contract introduced by #2476. Third-party builders that inherit its compatibility no-op `close()` cannot receive built-in cleanup guarantees until their provider implements `close()`.
- Low-level native-handle and OpenMP-state failure paths inside `CagraIndexImpl.build()` are unchanged.
- Validation and exact arithmetic are limited to the matrix, graph, and writer paths touched here; this is not a provider-wide overflow or malformed-input audit.
- Scalar quantization's existing all-negative-maximum initialization behavior is deliberately preserved rather than changed in the same patch.
- Quantized merges deliberately use two passes because their statistics depend on the complete live-vector set.
- The bounded-memory changes do not imply constant memory: construction still retains the full native input matrix and native resources needed for the current graph build.
- The GPU-search codec was already device-backed; this PR does not introduce a new direct-device mode.
- No performance claim is made from the sanity validation.

## Validation

Fresh exact-source results for this branch:

- `cuvs-java`: 5 unit tests and 120 integration tests; 0 failures, 0 errors, and 1 inherited skip
- `cuvs-lucene`: 389 tests; 0 failures, 0 errors, and 30 inherited skips
- Focused graph and remapping validation: 13 tests passed
- Three-level reopen-and-search coverage passed for float, binary-quantized, and scalar-quantized accelerated HNSW
- GPU writer coverage exercises aligned and unaligned CAGRA datasets, brute force, combined payloads, fallback boundaries, output-position guards, and cleanup ordering
- Spotless, generated-documentation idempotence, and `git diff --check`

### Deep1B 1M sanity validation

Fresh index directories and the same 1,000,000-vector, 96-dimensional source were used for:

1. Float accelerated HNSW with four source segments force-merged into one three-level segment
2. Binary-quantized accelerated HNSW with the same merge topology
3. Scalar-quantized accelerated HNSW with the same merge topology
4. GPU CAGRA
5. GPU brute force
6. Combined GPU CAGRA and brute-force payloads

Validation reopened every retained index and checked, as applicable, exact document/vector cardinality, complete unique source-ID coverage, `checkIntegrity()`, payload and loaded-handle types, graph level membership and nesting, duplicate-free adjacency, finite search results, and ground-truth overlap. The combined index additionally exercised direct CAGRA routing at `k=64` and brute-force routing at `k=1025`. The brute-force run's top-64 result set exactly matched the ground-truth top-64 set for all 980 measured queries. For the hardening float-HNSW index, a raw-bit comparison of all 1,000,000 vectors across 96 dimensions found no mismatch between each stored source ID and its corresponding FBIN row.

Lucene's default `TieredMergePolicy` may reorder non-adjacent source segments during a force merge. The validated invariant is preservation of each source ID and its association with the corresponding vector under Lucene's selected merge order, not monotonic post-merge document IDs.

These Deep1B runs are correctness sanity checks. Page-cache eviction was unavailable in the validation container, so their timings are not performance evidence.


nvzm123:
just for documentation

PR #2662, opened 2026-09-19
Fix Fern parsing of enums

lowener:
Fern parser was not removing comments before splitting over commas leading to invalid documentation.


aamijar (approved):
Good catch!

PR #2664, opened 2026-09-21
FIX IVF-RabitQ: allow searching after building without serialize round-trip

### LINKED ISSUE #2537: [BUG] ivf_rabitq needs seralize & deserilize steps before the index is searchable

achirkin:
Make the index searchable right after build:
- Change the in-memory built bitcode layout
- Ensure the centroid norms are computed

closes https://github.com/NVIDIA/cuvs/issues/2537

breaking change: serialized data layout is changed


cpp/tests/neighbors/ann_ivf_rabitq.cuh
tfeher, line 381: This one does not search, right? Do we have any test that does search after serialization?

tfeher (changes requested):
Thanks Artem for the fix! Could you confirm that we still test searching after serialize/deserialize? Otherwise it looks good.

cpp/tests/neighbors/ann_ivf_rabitq.cuh
achirkin: `this->run()` template takes a build lambda as an argument, calls it, then performs the search and evaluation. So all the tests macros defined here do the search.

tfeher (approved):
Thanks for the clarification, LGTM!

cpp/src/neighbors/ivf_rabitq/gpu_index/quantizer_gpu.cu
```diff
@@ -307,7 +308,7 @@ __global__ void pack_and_compute_factors_kernel(
       int bit = d_bin_XP[row * D + block_id * 32 + i];
       cur |= ((uint32_t)bit << (31 - i));
     }
-    d_packed_code[row * blocks_per_point + block_id] = cur;
+    d_packed_code[block_id * N + row] = cur;
```
Stardust-SJF, line 311: Thanks for the fix! This looks correct. By the way, do you think it would be better to make it coalesced (even though the write volume can be small)?

cpp/src/neighbors/ivf_rabitq/gpu_index/quantizer_gpu.cu
```diff
@@ -307,7 +308,7 @@ __global__ void pack_and_compute_factors_kernel(
       int bit = d_bin_XP[row * D + block_id * 32 + i];
```
Stardust-SJF, line 308: I just noticed that this part can be uncoalesced; maybe we can get it fixed in another PR.

